PR #3803, opened 2022-02-23
Fix gas fee on education wizard and related JSON parse error

### LINKED ISSUE #3626: Gas fee number does not load
**Describe the bug**
In the gas fee education wizard, specifically in this location:
https://github.com/MetaMask/metamask-mobile/blob/d9757b7e16e6469c85452702f415dc2e19b30608/app/components/Views/GasEducationCarousel/index.js#L153

it should display the text (in English):
> Current average gas fee: `{fee}`

The problem is, the `{fee}` part does not load correctly (that's what I'm assuming) and it just does not show, because it is initialized as `null` and that gets rendered.

**Screenshots**

**To Reproduce**
1. Open the right-side app drawer
2. Click on <kbd>Add funds</kbd>
3. Hit <kbd>Buy ETH</kbd>

_**Note**: this will only work on fresh app installs, since the alert only shows up once._

**Expected behavior**
To be honest, I have no idea how the gas price is actually retrieved, but I would guess it can fail if it's retrieved over the network. This means it would be good to have a loading / error symbol/texting about the status of this information.

 - Device: iPhone Xs
 - OS: iOS 15 Developer Beta 2
 - App Version v3.8.0

**Severity**

- Is this tech debt?

### LINKED ISSUE #3759: JSON Parse error: Unrecognized token '<'
While going  through the on ramp flow and use apple pay as your payment method a user gets: `JSON Parse error: Unrecognized token '<'`. See sentry logs with the error: https://sentry.io/organizations/metamask/issues/2964111352/?query=is%3Aunresolved

see [here](http://recordit.co/nLFhswwkXB)

to reproduce:
on an iPhone device tap on the buy button
on the purchase method view, select Apple Pay as your payment method 
Slide through the educational carousel. Notice the error appears while sliding through the carousel. 

**Severity**

- Is this tech debt?

Fatxx:
**Description**

Replace old Codefi API endpoint with GasController method to fetch gas fee estimates.

**Checklist**

**Issue**

Resolves #3759 and #3626


cortisiko (approved):
🌮 🌮 🌮

app/util/custom-gas/index.js
```diff
@@ -105,59 +105,25 @@ export function parseWaitTime(min) {
...
+		const { GasFeeController } = Engine.context;
+		const { gasFeeEstimates } = await GasFeeController.fetchGasFeeEstimates({ shouldUpdateState: false });
 
-	if (average === fast && average === safeLow) {
-		throw new Error('Api returned same gas prices');
+		return {
+			averageGwei: gasFeeEstimates.estimatedBaseFee,
+			fastGwei: gasFeeEstimates.high.suggestedMaxFeePerGas,
+			safeLowGwei: gasFeeEstimates.low.suggestedMaxFeePerGas,
+		};
```
wachunei, line 122: Is this function called always in a EIP1559 network only? If not I'd suggest to add different cases on `gasEstimateType` property of `GasFeeState` since it might happen that `gasFeeEstimates` is `EthGasPriceEstimate`, `LegacyGasPriceEstimate`, etc, instead of always `GasFeeEstimates` ([this is what I mean](https://github.com/MetaMask/controllers/blob/main/src/gas/GasFeeController.ts#L144-L183))

andrepimenta (approved):
LGTM

app/util/custom-gas/index.js
wachunei: @Fatxx thanks for considering this, changes look good 👌

Issue #2829, opened 2021-06-19
Lack of full ENS namespace support


brantlymillegan:
**Describe the bug**
In MetaMask Mobile on Android, I can't send to the ENS name brantly.art. I *am* able to send to brantly.eth and brantly.xyz. This indicates to me that you have a hardcoded list of suffixes in order to recognize an ENS name. This is incorrect. The ENS suffixes namespace expands over time (soon to expand to include all DNS suffixes, and the DNS suffix namespace itself expands over time). The correct implementation of ENS will attempt to look-up any dot-separated name on ENS.

**Screenshots**
![1](https://user-images.githubusercontent.com/30162762/122648230-37b3e200-d0f6-11eb-9050-6896340911ae.jpeg)

**To Reproduce**
_Steps to reproduce the behavior_
1. Go to MetaMask Mobile on Android
2. Try to send something to "brantly.art" - nothing happens

**Expected behavior**
I should be able to send to any ENS name. Right now, MetaMask mobile recognizes ENS names incorrectly such that it excludes some ENS names.

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S21+
 - OS: latest Android
 - App Version: v2.4.0 (746)

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**
This is a problem now, and will soon be a very big problem at ENS greatly expands DNS namespace integration.

- How critical is the impact of this bug on a user?
Arbitrary excludes some ENS names from working, user confusion

- Is this visible to all users?
Yes

- Is this tech debt?
It's simply due to the ENS integration being wrong. A simple fix will future this forever: simply attempt to look up any dot-separated name on ENS

sethkfman:
@omnat I wanted to bring this issue to your attention. Lets discuss early next week. Thx

brantlymillegan:
@sethkfman @omnat hey, just checking in on this. any update?

gantunesr:
Fixed in #3825

PR #3825, opened 2022-02-28
Fix issue #2829 - Lack of full ENS namespace support

tommasini:
**Description**

The ENS name verification was hardcoded and in metamask-extension they have an ens verification already implemented that fix this with a regex expression

Now we can find an account with the name "brantly.art" 

**Checklist**

**https://github.com/MetaMask/metamask-mobile/issues/2829**

Resolves #


gantunesr (approved):
Great work @tommasini. LGTM

PR #3836, opened 2022-03-02
Do not pin `google-play-deploy` job in Bitrise

rickycodes:
**Description**

Right now we have an issue with the `google-play-deploy` job in Bitrise. It looks like it's trying to pull down a version of `bundletool` that does not exist

image

There have been so many releases the asset is no longer up on GitHub and without this `.jar` file the upload fails.

Not pining this means it now runs the latest version of the job which pulls down a more recent version of `bundletool` (which uploads to play store successfully):

https://app.bitrise.io/build/03653d69-fabf-4944-a2c1-dc4e2c52e0bc


sethkfman (approved):
LGTM

PR #3840, opened 2022-03-02
[Improvement] Collect code coverage

owencraston:
**Description**

- Enable jest code coverage collection
- This is valuabe as a benchmark but also as a debug tool
Screen Shot 2022-03-02 at 1 51 34 PM

Screen Shot 2022-03-02 at 1 57 56 PM

- [We can configure this further if needed](https://jestjs.io/docs/configuration)
- this will create a local directory of the test coverage report that is git ignored for now but can be used as a developer tool

## Pros/Cons of adding this change
- the main con of collecting code coverage is that it slows down the speed that tests get ran
- In order to put a number on the cost of adding this I collected some test run times from `main` and this branch. Here is a summary of the data.
- Ran on my brand new Macbook pro m1 with 16Gb of ram (milage may vary 😆 )

#### With Cache
| Average run time without test coverage | Average run time with test coverage | Delta (%) |
|----------------------------------------|-------------------------------------|-----------|
| 42.6 seconds                           | 59.3 seconds                        | -39.2     |

#### Pre Cache
| Run time without test coverage | Run time with test coverage | Delta (%) |
|--------------------------------|-----------------------------|-----------|
| 81 seconds                     | 148 seconds                 | -82.7     |

I will leave a comment with my raw data/notes when I was doing this analysis so people can fact check if they like. 

#### Conclusion/Questions
For most day to day tasks involving running tests, the cached results are going to be what you are using due to the fact that you will most likely be running the tests many times. Even still the cached results are about 40% slower when collecting code coverage. The question then becomes `do we feel that having our test suite run 40% slower is worth measuring our test coverage/enforcing better testing in the future`?

**Checklist**

**Issue**

Resolves #???

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,7 @@ buck-out/
 *.jsbundle
 
 # testing
-coverage
+tests
 
 # app-specific
 /android/app/src/main/assets/InpageBridgeWeb3.js
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,13 +10,10 @@ const config = {
 		'node_modules/(?!(@react-native|react-native|rn-fetch|redux-persist-filesystem|@react-navigation|@react-native-community|@react-native-masked-view|react-navigation|react-navigation-redux-helpers|@sentry))',
 	],
 	snapshotSerializers: ['enzyme-to-json/serializer'],
-	collectCoverage: false,
-	collectCoverageFrom: [
-		'<rootDir>/app/**/*.{js,jsx}',
-		'!<rootDir>/node_modules/',
-		'!<rootDir>/app/util/*.{js,jsx}',
-		'!<rootDir>/app/entry*.js',
-	],
+	collectCoverage: true,
+	coveragePathIgnorePatterns: ['/node_modules/', '__mocks__', '<rootDir>/e2e/'],
+	coverageReporters: ['text-summary', 'lcov'],
+	coverageDirectory: '<rootDir>/tests/coverage',
 };
 
 // eslint-disable-next-line import/no-commonjs
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,7 +10,8 @@ const config = {
 		'node_modules/(?!(@react-native|react-native|rn-fetch|redux-persist-filesystem|@react-navigation|@react-native-community|@react-native-masked-view|react-navigation|react-navigation-redux-helpers|@sentry))',
 	],
 	snapshotSerializers: ['enzyme-to-json/serializer'],
-	collectCoverage: true,
+	// This is an environment variable that can be used to execute logic only in development
+	collectCoverage: process.env.NODE_ENV !== 'production',
 	coveragePathIgnorePatterns: ['/node_modules/', '__mocks__', '<rootDir>/e2e/'],
 	coverageReporters: ['text-summary', 'lcov'],
 	coverageDirectory: '<rootDir>/tests/coverage',


jest.config.js
```diff
@@ -10,13 +10,10 @@ const config = {
 		'node_modules/(?!(@react-native|react-native|rn-fetch|redux-persist-filesystem|@react-navigation|@react-native-community|@react-native-masked-view|react-navigation|react-navigation-redux-helpers|@sentry))',
 	],
 	snapshotSerializers: ['enzyme-to-json/serializer'],
-	collectCoverage: false,
-	collectCoverageFrom: [
```
owencraston, line 14: I think its easier to ignore the places  you DON'T want coverage from vs specifying all the places you do. Also I don't see any value in collecting coverage from third party node modules so I opted to exclude it.

owencraston:
### Raw Data/Notes
- test 1 with cov
    -  148 seconds on the jest clock
    - 155 seconds on my timer
-  test 2 with cov
    -  77 seconds on my clock
    -  71 seconds on jest timer  
        
        
-  test 3 with code cov
    
    
    
    -  53 seconds on the jest timer
    -  56 seconds on my clock
-  test 4 with code coverage
    
    
    
    -  54 seconds on the jest timer
    -  58 seconds on my timer
-  cleared the jest cache
-  test 1 without code cov
    
    
    
    -  81 seconds on the jest timer
    -  87 seconds my clock
-  test 2 without code coverage
    
    
    
    -  46.8 seconds on
    -  53 seconds on my clock
-  test 3 without code coverage
    
    
    
    -  41.5 seconds on jest timer
    -  46 seconds on my clock
-  test 4 without code coverage
    
    
    
    -  39.5 seconds on jets timer
    -  43 on my clock
    

### Analysis

- Pre cache
    - 148 with test coverage
    - 81 seconds without test coverage
    - ((81-148)/81)*100 = **-82.7160493827**
    - so roughly 83% slower when collecting test coverage before a cached test
- post cache
    - With coverage average time
        - (71 + 53 + 54)/3 = **59.3333333333 seconds**
    - without coverage average run time
        - (46.8 + 41.5 + 39.5) = **42.6**
    - delta
        - ((42.6-59.3)/42.6)*100 = **-39.2018779343**

jest.config.js
```diff
@@ -10,13 +10,10 @@ const config = {
...
-		'<rootDir>/app/**/*.{js,jsx}',
-		'!<rootDir>/node_modules/',
-		'!<rootDir>/app/util/*.{js,jsx}',
-		'!<rootDir>/app/entry*.js',
-	],
+	collectCoverage: true,
+	coveragePathIgnorePatterns: ['/node_modules/', '__mocks__', '<rootDir>/e2e/'],
```
sethkfman, line 14: I would suggest adding `'<rootDir>/app/**/*test.{ts,js}',`

owencraston: @sethkfman  do you mean you want these ignored? or you want them to be tested? The config here defines what files we are ignoring when running the test suite.

.gitignore
```diff
@@ -59,7 +59,7 @@ buck-out/
 *.jsbundle
 
 # testing
-coverage
+tests
```
owencraston, line 62: @rickycodes with this folder being git ignored the actual result would not be uploaded to github. If we want to use this file in the future we would just need to remove this line.

jest.config.js
sethkfman: Yes, my thought is that we have a lot of `.test.ts` files embedded in the app directory.

owencraston: Ah I see what you mean. You are saying we do not want to collect coverage on the test files as it will lower our coverage %. We do not actually need to do this since the `.test.x` files are what is used to determine if a file is tested so we they will never have code coverage collected on "themselves" ...if that makes sense. I believe if it did work like this, it would cause an infinite loop since you would `need to test the test file which would then lead to that new test not being tested and you'd have to just keep testing test files`.

sethkfman (approved):
LGTM

rickycodes (approved):

Issue #3870, opened 2022-03-08
Token Detection Broken on Main branch


Cal-L:
### **Description**

Token detection seems to have stopped working on the Main branch after v4.2.2 release

### **Technical Details**

-  Back track to see what broke

### **Acceptance Criteria**

-  Token detection should work

gantunesr:
Fixed in #3871

PR #3871, opened 2022-03-09
[FIX] Fix Token Detection

Cal-L:
**Description**

#3870 

**Checklist**

**Issue**

Resolves #???


Fatxx (approved):
LGTM

gantunesr:
@Cal-L the issue was the patch or web3? I upgrade the controllers patching it the same you did for the RN upgrade, I'm also upgrading Web3 in the controllers. Will this bug appear again in that case?

Cal-L:
Issue was both actually. The reason why I patched was because I had a different version of web3 library. It wasn't necessary so I just ended up removing all of it for now. Controllers still work the same as before

PR #3877, opened 2022-03-10
Add OSS attribution

rickycodes:
**Description**

This is an example of what keeping an automated up-to-date list of attributions in source _could_ look like.

This adds `oss-attribution-generator` which can be run to generate a simple `attributions.txt` file. it also adds aforementioned generated file to source and introduces a ci job that will fail if the attributions aren't up-to-date (a dependency has been added or removed).

Finally, the app has been updated to point to this new file instead of linking to the outdated one on the website:

image 

https://user-images.githubusercontent.com/675259/157584566-09ca5386-e53e-42a5-b0fe-a3040c59b113.mp4


package.json
```diff
@@ -90,7 +92,9 @@
     "@metamask/contract-metadata": "^1.30.0",
     "@exodus/react-native-payments/validator": "^13.7.0",
     "react-devtools-core": "4.22.1",
-    "simple-get": "4.0.1"
+    "simple-get": "4.0.1",
+    "oss-attribution-generator/**/underscore": "1.12.1",
+    "oss-attribution-generator/**/deep-extend": "0.5.1"
```
rickycodes, line 97: this kinda sucks.

it hasn't been updated in 4 years so there's some vuls that I've patched via resolution. we could explore scripting this ourselves, im sure it wouldn't be too hard to fork this and update it to suit our needs (for example we could probably drop the json file?).

rickycodes:
this was minimal amount of effort, but I think it's a good start. maybe we should not keep the json file? we could also just but the attributions at the root so it's easier to find instead of this default folder?

andreahaku (approved):
LGTM

wachunei (approved):

wachunei (approved):

PR #3878, opened 2022-03-10
Upgrade controllers v26.0.0 and Swaps controller to v6.6.0

gantunesr:
**Description**

Upgrade @metamask/controllers from v25.1.0 to v26.0.0

> **BREAKING**: Fetch and return token image as part of getDetails calls on ERC721Standard and ERC1155Standard (https://github.com/MetaMask/controllers/pull/702)

> This change is breaking because it requires that the AssetsContractController (on which the ERC721Standard and ERC1155Standard are instantiated) be passed a listener for onPreferencesStateChange from the PreferencesController so that it can use the user's preferred IPFSGateway to fetch any images hosted on IPFS. Consumers will have to pass onPreferencesStateChange in an options object (first arg) to the AssetsContractController constructor when initializing.

Reference: https://github.com/MetaMask/controllers/releases/tag/v26.0.0


jpcloureiro (approved):
lgtm

PR #3900, opened 2022-03-16
Fix crash on sign message popup

Fatxx:
**Description**

Safely delete property on a new object instead of deleting the original one which further causes the object to be `undefined` and resulting in a crash.

**Checklist**

**Issue**

Resolves #???


jpcloureiro (approved):
lgtm

andrepimenta (approved):

PR #3921, opened 2022-03-19
Keystone Integration - fix ui issues with iOS

soralit:
**Description**
This PR fixed some UI issues appear in iOS.

**Checklist**

**Issue**

Resolves #???

### DIFF
--- a/app/components/Nav/Main/RootRPCMethodsUI.js
+++ b/app/components/Nav/Main/RootRPCMethodsUI.js
@@ -389,8 +389,10 @@ const RootRPCMethodsUI = (props) => {
 	);
 
 	const renderQRSigningModal = () => {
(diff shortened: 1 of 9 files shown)


soralit:
For the react-native-modal performs different behavior between iOS and Android. The global QR signing modal is not workable in iOS. iOS can only show one Modal with same level at one time, other modals must wrapper in the upper one to be a children. 
So I add the QRSigningModal to every needed places which are: `Signing Dapp Tx`, `Approve Account`, `Sign message` and limit the usage of the global modal to only show when there is a `swap tx` / `MetaMask tx` / `tx in history`.

gantunesr (commented):
Hey @soralit, this LGTM, I want to verify some use cases before merging this PR.

PR #3950, opened 2022-03-24
Keystone Integration: Fix an app crash issue and an ui issue on iOS

soralit:
**Description**

This PR fixed a crash issue when signing dapp transaction with Keystone. 
Also fix that the speed up transaction QR code cannot shown on iOS platform.

**Checklist**

**Issue**

Resolves #???

### DIFF
--- a/app/components/Views/Approval/index.js
+++ b/app/components/Views/Approval/index.js
@@ -96,40 +96,49 @@ class Approval extends PureComponent {
 		navigation.setOptions(getTransactionOptionsTitle('approval.title', navigation, {}, colors));
 	};
 
-	componentDidMount = () => {
-		this.updateNavBar();
-	};
-
 	componentDidUpdate = () => {
 		this.updateNavBar();
 	};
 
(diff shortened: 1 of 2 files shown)


app/components/Views/Approval/index.js
```diff
@@ -96,40 +96,49 @@ class Approval extends PureComponent {
...
+					KeyringController.cancelQRSignRequest();
+				} else {
+					Engine.context.TransactionController.cancelTransaction(transaction.id);
+				}
+				Engine.context.TransactionController.hub.removeAllListeners(`${transaction.id}:finished`);
+				AppState.removeEventListener('change', this.handleAppStateChange);
+				this.clear();
+			}
+		} catch (e) {
+			if (e) {
```
soralit, line 119: For the cancelTransaction will call twice when user swipe up their phone to make app inactive, it will throw an error but the value is undefined.

gantunesr (approved):

PR #3967, opened 2022-03-29
Fix/2830 enter many names

### LINKED ISSUE #2830: Immediately resolves name as soon as I enter ".eth", but this excludes the ability to enter many names
**Describe the bug**
When entering an ENS name into a Recipient field, as soon as I enter a possible ENS suffix, e.g. ".eth" or ".xyz", the app immediately resolves the name and doesn't let me keep typing. This is a problem because subdomains exist in ENS. E.g. the name evan.ethereum.eth exists, but when I try to type it in as soon as I get to "evan.eth..." it resolves the name and doesn't let me keep typing. This would also be a problem if I tried to send to "brantly.xyz.eth", the app cuts me off as soon as i typed in "brantly.xyz", even though "brantly.xyz.eth" and "brantly.xyz" are two different names with different owners and records.

**Screenshots**
_If applicable, add screenshots or links to help explain your problem_

**To Reproduce**
1. Go to MetaMask Mobile on Android
2. Tap "Send"
3. Try to type in to the Recipient field the ENS name "evan.ethereum.eth"
4. You'll see that it cuts you off as soon as you type in "evan.eth..."

**Expected behavior**
You should be able to type in any name, and the app should never stop you from typing in more. As you type in, could present possible names below for you to select, but it shouldn't stop you from typing more.

 - Device: Samsung Galaxy S21+
 - OS: latest Android
 - App Version: v2.4.0 (746)

**Severity**

Could lead to lost funds

potentially everyone

yes

- Is this tech debt?
no, you just have a wrong implementation of ENS

tommasini:
**Description**

The input text is not longer transformed in a formatted text with was impossible to keep writing.
Now it is a text input with a check icon that tell us that the address it is valid.

**Checklist**

**Issue**

Resolves #2830 
Resolves the fact that we could not write many names like "evan.ethereum.ether" and now we can access that addresses too with this solution.

Small demo with the final result

https://user-images.githubusercontent.com/46944231/160585472-64d17744-04c8-47ed-93d6-d1ed27c8c660.mp4


app/components/Views/SendFlow/AddressInputs/index.js
```diff
@@ -212,62 +263,90 @@ export const AddressTo = (props) => {
...
+							autoFocus
 							onBlur={onInputBlur}
 							onSubmitEditing={onSubmit}
 							value={toSelectedAddress}
 							testID={'txn-to-address-input'}
 							keyboardAppearance={themeAppearance}
 						/>
 					</View>
-					{!!onScan && (
+					{!toSelectedAddress?.length > 0 ? (
```
gantunesr, line 274: Can you move this condition to a variable with a clear name so its easier to read and understand the logic?

app/util/address/index.js
```diff
@@ -28,10 +28,10 @@ export function renderShortAddress(address, chars = 4) {
 	return `${checksummedAddress.substr(0, chars + 2)}...${checksummedAddress.substr(-chars)}`;
 }
 
-export function renderSlightlyLongAddress(address, chars = 4) {
+export function renderSlightlyLongAddress(address, chars = 4, initialChars = 20) {
 	if (!address) return address;
 	const checksummedAddress = toChecksumAddress(address);
-	return `${checksummedAddress.substr(0, chars + 20)}...${checksummedAddress.substr(-chars)}`;
+	return `${checksummedAddress.substr(0, chars + initialChars)}...${checksummedAddress.substr(-chars)}`;
 }
```
gantunesr, line 35: Can you add unit testing to this method?

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -347,6 +347,9 @@ class SendFlow extends PureComponent {
 						networkAddressBook[checksummedToSelectedAddress].name) ||
 					(identities[checksummedToSelectedAddress] && identities[checksummedToSelectedAddress].name);
 			} else {
+				toAddressName = await doENSReverseLookup(toSelectedAddress, network);
+
+				toAddressName = toAddressName === undefined ? toSelectedAddress : toAddressName;
```
gantunesr, line 352: Would this work too?

```suggestion
				toAddressName = toAddressName || toSelectedAddress;
```

gantunesr (changes requested):
Hey @tommasini, this is looking good. Left a few comments and also we have a provider that takes care of the translations, so you only need to add the english text and they take care of the rest, so you can revert those changes.

app/components/Views/SendFlow/AddressInputs/index.js
tommasini: Yes of course! Thank you!

app/util/address/index.js
tommasini: Of course! Done it! Thank you!

app/components/Views/SendFlow/SendTo/index.js
tommasini: Yes, tested and implemented! Thank you!

tommasini:
With @jpcloureiro, we solve the [issue](https://github.com/MetaMask/metamask-mobile/issues/4005) , upgrading the "to" input performance in the last commit of this branch

chrisleewilcox:
Not able to checkout/pull branch.  Getting an invalid error....
image

tommasini:
Closing this PR and opening a new one from the remote metamask mobile app repo, can access [here](https://github.com/MetaMask/metamask-mobile/pull/4089)

PR #3975, opened 2022-03-29
Split CI into multiple jobs

rickycodes:
**Description**

This splits our previously single job into _multiple_ jobs. we do this by caching `yarn setup` in the first `setup` job and then utilise that cache in downstream jobs.

Additionally, I've made one small tweak where the node version used in the action containers is plucked from the `.npmrc` file in the project. if/when we upgrade nodejs we won't have to update the fixed version here.

new pipeline looks like this: 
image

you can also look at the last workflow run for this pull request :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,31 +1,26 @@
-name: CI
+name: ci
 on: [push,pull_request]
 jobs:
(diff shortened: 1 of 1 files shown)


wachunei (approved):
LGTM

PR #3980, opened 2022-03-30
Patch network specific asset modal (Token detection)

blackdevelopa:
**Description**

Patched network specific assets education modal to include token detection (mainnet)

**Checklist**

**Issue**

Resolves #???

### DIFF
--- a/app/components/Nav/Main/MainNavigator.js
+++ b/app/components/Nav/Main/MainNavigator.js
@@ -83,6 +83,11 @@ const WalletTabHome = () => (
 			component={RevealPrivateCredential}
 			options={RevealPrivateCredential.navigationOptions}
 		/>
+		<Stack.Screen
+			name="ExperimentalSettings"
+			component={ExperimentalSettings}
+			options={ExperimentalSettings.navigationOptions}
+		/>
 	</Stack.Navigator>
 );
 
--- a/app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
+++ b/app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
@@ -1,7 +1,8 @@
-import React, { memo } from 'react';
+import React, { memo, useCallback } from 'react';
 import { View, Text, Linking, StyleSheet } from 'react-native';
 import { strings } from '../../../../../locales/i18n';
 import { useNavigation } from '@react-navigation/native';
+import { MAINNET } from '../../../../constants/network';
 import { useAppThemeFromContext, mockTheme } from '../../../../util/theme';
 
 const createStyles = (colors: {
@@ -39,46 +40,44 @@ interface DescriptionProps {
 	clickableText: string | undefined;
 	number: number;
 	onClose: () => void;
-	mainnetTokenDetection: boolean;
-	isMainnet: boolean;
+	isMainnetTokenDetectionEnabled: boolean;
+	network: string;
 }
 
 const Description = (props: DescriptionProps) => {
-	const { description, clickableText, number, onClose, mainnetTokenDetection, isMainnet } = props;
+	const { description, clickableText, number, onClose, isMainnetTokenDetectionEnabled, network } = props;
 	const { colors } = useAppThemeFromContext() || mockTheme;
 	const styles = createStyles(colors);
 	const navigation = useNavigation();
 
-	const handlePress = () => {
+	const handlePress = useCallback(() => {
 		if (number === 2) {
 			Linking.openURL(strings('network_information.learn_more_url'));
 		} else {
 			onClose();
 			navigation.navigate('AddAsset', { assetType: 'token' });
 		}
-	};
+	}, [navigation, onClose, number]);
 
-	const handleNavigation = () => {
+	const handleNavigation = useCallback(() => {
 		onClose();
 		navigation.navigate('ExperimentalSettings');
-	};
+	}, [navigation, onClose]);
 
 	return (
 		<View style={styles.descriptionContainer}>
 			<View style={styles.contentContainer}>
 				<Text style={styles.numberStyle}>{number}.</Text>
 				<Text style={styles.description}>
-					<Text>
-						{description} {''}
-					</Text>
+					<Text>{`${description} `}</Text>
 					{clickableText && (
 						<>
-							{!mainnetTokenDetection && isMainnet && (
+							{!isMainnetTokenDetectionEnabled && network === MAINNET && (
 								<>
 									<Text style={styles.link} onPress={handleNavigation}>
-										{strings('network_information.enable_token_detection')}
-									</Text>{' '}
-									{strings('network_information.or')}{' '}
+										{`${strings('network_information.enable_token_detection')} `}
+									</Text>
+									{`${strings('network_information.or')} `}
 								</>
 							)}
 							<Text style={styles.link} onPress={handlePress}>
--- a/app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
+++ b/app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
@@ -1,6 +1,7 @@
 import React, { memo } from 'react';
 import { View, Text, Linking, StyleSheet } from 'react-native';
 import { strings } from '../../../../../locales/i18n';
+import { useNavigation } from '@react-navigation/native';
 import { useAppThemeFromContext, mockTheme } from '../../../../util/theme';
 
 const createStyles = (colors: {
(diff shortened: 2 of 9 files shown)


app/components/Nav/Main/MainNavigator.js
```diff
@@ -82,6 +82,11 @@ const WalletTabHome = () => (
 			component={RevealPrivateCredential}
 			options={RevealPrivateCredential.navigationOptions}
 		/>
+		<Stack.Screen
+			name="ExperimentalSettings"
```
owencraston, line 86: I know no one else if doing this but I really suggest turning these navigation keys being constants.

app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
```diff
@@ -37,35 +38,52 @@ interface DescriptionProps {
 	description: string;
 	clickableText: string | undefined;
 	number: number;
-	navigation: any;
 	onClose: () => void;
+	mainnet_token_detection_off: boolean;
```
owencraston, line 42: Nitpick: This should be camel case. `mainnetTokenDetection`.
I also don't think the `_off` is necessary, instead you should do the inverse and simply tiggle it on/off.

app/components/UI/NetworkInfo/index.test.tsx
```diff
@@ -12,15 +13,16 @@ const initialState = {
 	engine: {
 		backgroundState: {
 			NetworkController: {
-				provider: { type: 'mainnet', rpcTarget: 'http://10.0.2.2:8545' },
+				provider: { type: MAINNET, rpcTarget: 'http://10.0.2.2:8545' },
```
owencraston, line 16: should we not make this rpcTarget a constant similar to what you have done with `MAINNET`?

app/components/UI/NetworkInfo/index.tsx
```diff
@@ -108,19 +108,22 @@ interface NetworkInfoProps {
...
 		onClose,
 		ticker,
+		isTokenDetectionEnabled,
 		networkProvider: { nickname, type, ticker: networkTicker, rpcTarget },
-		navigation,
 	} = props;
 	const { colors } = useAppThemeFromContext() || mockTheme;
 	const styles = createStyles(colors);
 
+	const mainnet_token_detection_on = type === MAINNET && isTokenDetectionEnabled;
```
owencraston, line 124: These are redundant. you could just have one constant called `mainNetTokenDetection = type === MAINNET && isTokenDetectionEnabled` and then in the places you used `mainnet_token_detection_off` you could just use `!mainNetTokenDetection`

owencraston: why do we have a constant for mainnet token detection but not for the other networks?

owencraston:
Here we can see when token detection is on, the asset education modal messaging is different, indicating that tokens should be auto imported. 
https://user-images.githubusercontent.com/22918444/160901309-887b2236-24f1-42cb-84f6-1f50614706c8.mov
- From a UI perspective everything looks good, just left a few comments about the code

app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
blackdevelopa: Valid. Updated. Thank you

app/components/UI/NetworkInfo/index.test.tsx
```diff
@@ -12,15 +13,16 @@ const initialState = {
...
 			},
-			PreferencesController: { frequentRpcList: ['http://10.0.2.2:8545'] },
+			PreferencesController: { useStaticTokenList: true, frequentRpcList: ['http://10.0.2.2:8545'] },
```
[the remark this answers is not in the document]
owencraston: I just realized these were tests. My bad

app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
```diff
@@ -37,35 +38,53 @@ interface DescriptionProps {
...
 		if (number === 2) {
 			Linking.openURL(strings('network_information.learn_more_url'));
 		} else {
 			onClose();
-			navigation.push('AddAsset', { assetType: 'token' });
+			navigation.navigate('AddAsset', { assetType: 'token' });
 		}
 	};
 
+	const handleNavigation = () => {
```
owencraston, line 61: this should also be wrapped in a useCallback

app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
```diff
@@ -37,35 +38,53 @@ interface DescriptionProps {
...
+	};
+
 	return (
 		<View style={styles.descriptionContainer}>
 			<View style={styles.contentContainer}>
 				<Text style={styles.numberStyle}>{number}.</Text>
 				<Text style={styles.description}>
-					<Text>{description}</Text>
+					<Text>
+						{description} {''}
```
owencraston, line 72: why the extra empty string here?

owencraston: If you need the extra space here i'd recommend using string interpolation. 
```
`{${description} `
```

app/components/UI/NetworkInfo/index.tsx
```diff
@@ -108,19 +108,32 @@ interface NetworkInfoProps {
...
 		onClose,
 		ticker,
+		isTokenDetectionEnabled,
 		networkProvider: { nickname, type, ticker: networkTicker, rpcTarget },
-		navigation,
 	} = props;
 	const { colors } = useAppThemeFromContext() || mockTheme;
 	const styles = createStyles(colors);
 
+	let mainnetTokenDetection = false;
```
owencraston, line 124: instead of this declarative approach we can clean it up by using a factory function. Luckily react has a useMemo function which calculates and stores values and will also memoize them for us. 
Something like this should work...
```
const mainnetTokenDetection = useMemo(() => {
	if (type === MAINNET && isTokenDetectionEnabled) {
		return true;
	}
	return false;
}, []);
```
the `isMainnet` variable since its just a variable storing information we already have in `type === MAINNET`

app/components/UI/NetworkInfo/InfoDescription/infoDescription.tsx
```diff
@@ -37,35 +39,51 @@ interface DescriptionProps {
 	description: string;
 	clickableText: string | undefined;
 	number: number;
-	navigation: any;
 	onClose: () => void;
+	isMainnetTokenDetectionEnabled: boolean;
+	network: string;
```
owencraston, line 44: I suggest we add a type to this file [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/constants/network.js) that would look like...
`export type Network = MAINNET | ROPSTEN | KOVAN | RINKEBY | GOERLI | RPC | NO_RPC_BLOCK_EXPLORER | PRIVATENETWORK;`

and then within this file we can the prop be ...
```
network:Network;
```

and the same would go for the [Network info](https://github.com/MetaMask/metamask-mobile/blob/9ad7d610b77a11365940f988ccb9b28c1ba303af/app/components/UI/NetworkInfo/index.tsx#L101)

owencraston: This file will need to be converted to `.ts` to do this but it should be very simple

owencraston (approved):

chrisleewilcox:
Getting the following error when running `yarn setup`...

```
**ERROR** Failed to apply patch for package @metamask/design-tokens at path
  
    node_modules/@metamask/design-tokens

  This error was caused because @metamask/design-tokens has changed since you
  made the patch file for it. This introduced conflicts with your patch,
  just like a merge conflict in Git when separate incompatible changes are
  made to the same piece of code.

  Maybe this means your patch file is no longer necessary, in which case
  hooray! Just delete it!

  Otherwise, you need to generate a new patch file.

  To generate a new one, just repeat the steps you made to generate the first
  one.

  i.e. manually make the appropriate file changes, then run 

    patch-package @metamask/design-tokens

  Info:
    Patch file: patches/@metamask+design-tokens+1.4.2.patch
    Patch was made for version: 1.4.2
    Installed version: 1.5.1

---
patch-package finished with 1 error(s).
4. Create xcconfig files...
```

chrisleewilcox:
The modal is not translated for other languages....
image

chrisleewilcox (approved):
Looks good to merge.  Translation can be handle in a different issue.

PR #3987, opened 2022-03-30
add new resolutions for plist, minimist and remove minimist exclude

rickycodes:
**Description**

This fixes `yarn audit:ci` by adding two new resolutions. I'm also removing the skip for `minimist` since we don't require it any longer.


sethkfman (approved):
LGTM

PR #4003, opened 2022-03-31
Split CI into multiple jobs (#3975)

rickycodes:
* split ci into multiple jobs

* only run ci on pushes to main branch

**Description**

add ci split to release PR


wachunei (approved):
LGTM

PR #4068, opened 2022-04-11
Update `crowdin/github-action`

rickycodes:
Update `crowdin/github-action` to `9237b4cb361788dfce63feb2e2f15c09e2fe7415` (latest)

This action was updated in our allow list and needs to be updated here as well.

same as: https://github.com/MetaMask/metamask-extension/pull/14381

Also removing `base_url` same as: https://github.com/MetaMask/metamask-extension/pull/14364


andreahaku (approved):
LGTM

Issue #3952, opened 2022-03-24
[UI] Keystone - Connect Wallet Flow review


holantonela:
REF 
https://www.figma.com/file/7dE8BHjcoF7CYDfvKKz2Wx/MMM-%3C%3E-Keystone-UI-Review?node-id=0%3A1

General

- [x] Close icon is misaligned

Select Account 

- [x] The input needs `8px` margin. Also, the default checkbox component does not span the whole height. Please use it.
- [x] We need `16px` of Y spacing between the list and the Prev/Next navigation buttons.
- [x] Typography should be in 16px for title and 14 px for content
- [x] Spacing between checkbox and content must be 16px
- [x] optimize the order of the content: name of the account as title, address subtitle 
- [x] align cell with the title: "select an account"
- [x] grey background when cell is selected is not necessary and it doesn't look good
- [x] Size of icons must be between 32px or 24 px

holantonela:
Updated Table UI

![image](https://user-images.githubusercontent.com/87079938/159959097-67dfbd5e-b230-4f12-b7f1-6b60e3cac5fb.png)

PR #4070, opened 2022-04-11
[5.3] [IMPROVEMENT] Update SelectQRAccounts UI

gantunesr:
**Description**

This PR updates the style of the `SelectQRAccounts` component and refactors the HOC `withQRHardwareAwareness`

**Use Case**

GIVEN I'm importing my accounts from my Keystone to MMM
WHEN I select an account or more
THEN the account(s) should be imported to MMM
 
**Checklist**

**Issue**

Progresses #3952


app/components/UI/QRHardware/withQRHardwareAwareness.tsx
```diff
@@ -9,51 +9,42 @@ const withQRHardwareAwareness = (
...
 			const { KeyringController } = Engine.context as any;
 			KeyringController.getQRKeyringState().then((store: any) => {
-				this.keyringState = store;
-				this.keyringState.subscribe(this.subscribeKeyringState);
+				keyringState.current = store;
+				keyringState.current.subscribe(subscribeKeyringState);
 			});
-		}
+			return () => {
+				if (keyringState) {
```
soralit, line 32: Should this be `keyringState.current`?

gantunesr: Yep

plasmacorral:
Why do we have specific formatting for the iphone X in app/components/Views/ConnectQRHardware/SelectQRAccounts/index.tsx and app/components/Views/ConnectQRHardware/index.tsx, @soralit?

plasmacorral:
This is how the screen is presented on my Pixel 3a

And for reference how this looked in earlier builds: 

The way we are wrapping the public address on a screen of this size is not ideal.  I would defer to design but it might be nice to display Keystone 0 on this screen to be consistent with how they will be presented in the account menu after import.

plasmacorral:
Just following up on my comment above:

> ...wrapping the public address on a screen of this size...

 I now see that it is the Eth balance appearing below the public address (not the address wrapping to a new line).   

Should we preserve the unit "ETH" to make this more clear, or was that removed intentionally?

gantunesr:
> Should we preserve the unit "ETH" to make this more clear, or was that removed intentionally?

It was a bug, its already fixed in the last commit

> I must defer to design, but it might be nice to display "Keystone 0" on this screen to be consistent with how account names will be presented in the account menu after import

I added the text "QR Hardware" as its showed in the Wallet View. What do you think @holantonela and @plasmacorral?

In the following screenshots you can see the evidence,

drawing

holantonela:
Thanks for updating this screen folks! I'm looping @SaraCheikh for a final review. 

Please, don't merge until we review it in design.

SaraCheikh:
1. Next/previous buttons**: these buttons look outdated, if we can't replace them with arrow buttons, we should at least improve the spacing:

- Spacing between the arrow and the text should be 8px and 32px between buttons, to improve visual harmony and prevent the user from pressing the wrong button. 

reference:
Screenshot 2022-04-25 at 14 29 29

2. Header: the introduction of the icon in the header only in this flow gives the feeling of being in another interface. For consistency, I would remove it and use the same header that we use on other screens. If for some major reason, it is absolutely necessary to use this header, I would suggest the following changes: 

- the size of the icon should be a maximum of 40x40
- Align the title with the close button instead of aligning it to the icon, for the user to directly relate the action of closing with the title of the screen. 
- the spacing between the icon and the title should be no more than 16px. At the moment there is exaggerated spacing between these elements. 

reference:
Screenshot 2022-04-25 at 14 46 23

gantunesr:
@holantonela 

| Light Mode | Dark Mode |
|------------|-----------|
| Screen Shot 2022-04-27 at 00 27 42 | Screen Shot 2022-04-27 at 00 26 52 |

app/components/UI/QRHardware/withQRHardwareAwareness.tsx
```diff
@@ -9,51 +9,42 @@ const withQRHardwareAwareness = (
...
-  class QRHardwareAwareness extends Component<any, any> {
-    state: {
-      QRState: IQRState;
-    } = {
-      QRState: {
-        sync: {
-          reading: false,
-        },
-        sign: {},
+  const QRHardwareAwareness = (props: any) => {
```
owencraston, line 12: The naming convention of this file is a little weird. Perhaps just call it `QRHardwareAwareness.tsx`?

owencraston: Can we better type these props?

app/components/Views/ConnectQRHardware/AccountDetails/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+const AccountDetails = (props: IAccountDetailsProps) => {
+  const { colors } = useAppThemeFromContext() || mockTheme;
+  const styles = createStyle(colors);
+  const navigation = useNavigation();
+  const { item, provider } = props;
+  const defaultTicker = 'ETH';
+
+  const toEtherscan = (address: string) => {
```
owencraston, line 55: This should probably be wrapped in a useCallback

app/components/Views/ConnectQRHardware/AccountDetails/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+        name={'external-link'}
+        onPress={() => toEtherscan(item.address)}
+        style={styles.linkIcon}
+        color={colors.text.default}
+      />
+    </View>
+  );
+};
+
+export default AccountDetails;
```
owencraston, line 89: This should probably be wrapped in a `React.memo(...)`

app/components/Views/ConnectQRHardware/SelectQRAccounts/index.tsx
```diff
@@ -144,23 +118,12 @@ const SelectQRAccounts = (props: ISelectQRAccountsProps) => {
...
-            <Text style={styles.address}>
-              {renderFromWei(item.balance)} {provider.ticker}
-            </Text>
-            <Icon
-              size={18}
-              name={'external-link'}
-              onPress={() => toEtherscan(item.address)}
-              color={colors.text.default}
-            />
+            <AccountDetails item={item} provider={provider} />
```
owencraston, line 126: Beautiful, I love to see this componentization.

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -282,18 +279,18 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
     <Fragment>
```
owencraston, line 279: whats the use of a fragment here vs a regular view?

gantunesr: AFAIK, `Fragment` is used only to wrap components, you can't render and add styles to a `Fragment`. So I think it's appropriate to this case.

[Reference](https://reactjs.org/docs/fragments.html)

app/components/UI/QRHardware/withQRHardwareAwareness.tsx
gantunesr: > The naming convention of this file is a little weird. Perhaps just call it QRHardwareAwareness.tsx

https://github.com/MetaMask/metamask-mobile/pull/4244

> Can we better type these props?

I'm having some problems to declare the props in an acceptable way since every component that uses this HOC will have their own props interface. We can discuss this further tomorrow.

app/components/Views/ConnectQRHardware/AccountDetails/index.tsx
gantunesr: Good observation https://github.com/MetaMask/metamask-mobile/pull/4070/commits/5f541feed3d79dda80439b71c27f30e9645f5bdc

app/components/Views/ConnectQRHardware/AccountDetails/index.tsx
gantunesr: In a second inspection I think there's no need for this method to be in this component, we could move it up to the parent and pass it in the props or use it as a helper method

app/components/UI/QRHardware/withQRHardwareAwareness.tsx
owencraston: I didn't realize this was a HOC. This does complicate things? Should we consider turning this into a hook for future scalability?

app/components/Views/ConnectQRHardware/AccountDetails/index.tsx
owencraston: true

app/components/Views/ConnectQRHardware/index.tsx
owencraston: I don't think I fully understand the use case of fragments yet but I will trust your research.

owencraston (approved):
Nice work. Only thing I would say is we should try and move `withQRHardwareAwareness` to a hook but i'd suggest we do this in another PR to reduce the scope. I didn't run the app so this is just a code review!

plasmacorral:
Android:



iOS:

app/components/UI/QRHardware/withQRHardwareAwareness.tsx
gantunesr: I agree with your comment down below, we can address this improvement in a separate OR

AlexJupiter:
Candidate for 5.3

PR #4098, opened 2022-04-13
fix crash when changing password

jpcloureiro:


andreahaku (approved):
LGTM even if we should understand what's not passing a valid address to the component

blackdevelopa:
@jpcloureiro do you think if you rebased, it will fix the failing ci/audit?

jpcloureiro:
> LGTM even if we should understand what's not passing a valid address to the component

Yes, we should check where the `EthereumAddress` component is used and what address prop is being passed on.
Let's handle that on a new issue.

> @jpcloureiro do you think if you rebased, it will fix the failing ci/audit?

Not sure. Did github publish a new vulnerability in the last hours?

cortisiko (approved):
✅ ✅ ✅ ✅

PR #4106, opened 2022-04-14
bump detox version

cortisiko:
This is a version bump from 19.4.0 to 19.6.5

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "metamask",
-  "version": "4.4.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "metamask",
-  "version": "4.4.0",
```
andreahaku, line 3: this is not necessary I think. Surely not related to this PR :)

andreahaku (approved):
Other than the app version number, LGTM :)

PR #4124, opened 2022-04-19
remove async vuln from resolutions and skip instead

rickycodes:
The `async` resolution that was here previously was a little heavy handed and so we're undoing that and skipping this disclosure for now. The skip was already applied in a previous commit: https://github.com/MetaMask/metamask-mobile/commit/d9e6ea105bb11f0329b199742c4d8aa3294a32cd


gantunesr (approved):
LGTM

Issue #4107, opened 2022-04-14
[5.3] RangeError: currency is not a well-formed currency code


AlexHerman1:
**Description:**
App is mostly crashing when trying to perform actions such as swap/claim/stake on a dapp, the error does break the core flow of the app, but will go away upon closing and reopening the app.

This error is produced when fiat currency is set to some of the alternative currencies in the settings-->general-->currency conversion list.

**Screenshots**
![image](https://user-images.githubusercontent.com/38788573/163408347-0a943f93-7d72-49c6-a13b-4f67844a8e9d.png)

**To Reproduce**
1. Go to settings-->general-->currency conversion and choose USDT - tether
2. Go to the browser, navigate to pancakeswap.finance
3. Try to perform a swap transaction
4. App will crash and produce error, txn will not go through

**Expected behavior**
Users can set any currency conversion and still interact with dapps successfully without producing the "well-formed currency" error

**Smartphone (please complete the following information):**
 - Device: Android and Apple devices
 - OS: multiple
 - App Version: 4.4.0

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

Fatxx:
Related with #4057

PR #4131, opened 2022-04-20
[5.3] [FIX] Add method for crypto that are not in ISO4217

tommasini:
**Description**
 The problem was that if we have the currency conversion selected with one of the Crypto values options, some of them will break our app because they were not listed in the ISO4217 that it's the currency format followed by the [FormatNumber](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat)

It was also opened a ticket [here](https://github.com/mdn/content/issues/15153) to try to understand why some crypto codes are acceptable and others not.

Proposed solution
The proposed solution it's when we the currency conversion selected do not meet the iso standards we return the format _value currency_ for example "1 1ST"

**Code Impact**
Big code impact, the formatCurrency method is used in other files that are used in multiple components and that components are in major transactions screens / transactions modals.

**Screenshots or Videos**
https://recordit.co/ZG6nNRfvzB

Add relevant testing to ensure the bug is fixed
**Test Cases**
Case1:
- Go to settings
- Go to general
- Select 1ST - FirstBlood in the currency conversion option
- Go to browser
- Browse https://app.uniswap.org/
- Do one swap
- Watch the Transaction approval modal (In main would break here)

**Checklist**

**Issue**

Progresses #4107


app/util/confirm-tx.js
```diff
@@ -82,15 +82,41 @@ export function getTransactionFee({ value, fromCurrency = 'ETH', toCurrency, con
...
+		code === 'USDC' ||
+		code === 'USDT' ||
+		code === 'WINGS' ||
+		code === 'ZEC'
+	) {
+		return false;
+	}
+	return true;
+}
+
```
jpcloureiro, line 107: Proposal: create an array with the codes

```js
const NON_ISO4217_CRYPTO_CODES = [
  '1ST',
  'DASH',
  ...
];

```

app/util/confirm-tx.js
```diff
@@ -82,15 +82,41 @@ export function getTransactionFee({ value, fromCurrency = 'ETH', toCurrency, con
...
+	let formatedCurrency;
+	if (!isCryptoCodeInISO4217(upperCaseCurrencyCode)) {
+		formatedCurrency = `${Number(value)} ${upperCaseCurrencyCode}`;
+	} else {
+		formatedCurrency = new Intl.NumberFormat(I18n.locale, {
+			currency: upperCaseCurrencyCode,
+			style: 'currency',
+			// eslint-disable-next-line no-mixed-spaces-and-tabs
+		}).format(Number(value));
+	}
```
jpcloureiro, line 119: with the proposed array above you can reduce the code to something similar to this

```js
const formatedCurrency = NON_ISO4217_CRYPTO_CODES.includes(upperCaseCurrencyCode)
  ? `${Number(value)} ${upperCaseCurrencyCode}`;
  : new Intl.NumberFormat(I18n.locale, {
      currency: upperCaseCurrencyCode,
      style: 'currency',
	  // eslint-disable-next-line no-mixed-spaces-and-tabs
	}).format(Number(value));
```

jpcloureiro (changes requested):

app/util/confirm-tx.js
tommasini: Thank you for the proposal! It will be much better!!

jpcloureiro (approved):

PR #4134, opened 2022-04-20
Feature/remove bitmask

rickycodes:
**Description**

I noticed yesterday that a couple of high and critical vulns were being skipped somehow...

After further examination I realised that the bitmask we had here was unnecessary as it was meant to be used in conjunction with `yarn audit`. Running `improved-yarn-audit` with `--min-severity moderate` actually handles this for us and gives us the correct exit code `0` if there's only a `LOW` vulnerability.

To handle the advisories correctly I've skipped one that has no patched version: https://github.com/advisories/GHSA-gff7-g5r8-mg8m and bumped `axios` as well. The `axios` bump should be safe since it's `minor`.

I've also moved exclusions into `.iyarc` so they're a bit easier to manage.

Lastly, I added `--fail-on-missing-exclusions` to ensure that if we have any unessesairy exclusions ci will also fail in that case: https://www.npmjs.com/package/improved-yarn-audit#user-content-maintaining-advisory-exclusions

### DIFF
--- a/.iyarc
+++ b/.iyarc
@@ -1,0 +1,3 @@
+# improved-yarn-audit advisory exclusions
+GHSA-fwr7-v2mv-hh25
+GHSA-gff7-g5r8-mg8m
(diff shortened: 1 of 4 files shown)


sethkfman (approved):
LGTM

PR #4135, opened 2022-04-20
browser improvement

blackdevelopa:
**Description**

This PR correctly updates the browser address bar when the origin is changed

**Checklist**

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/185

### DIFF
--- a/app/components/Views/BrowserTab/index.js
+++ b/app/components/Views/BrowserTab/index.js
@@ -857,6 +857,12 @@ export const BrowserTab = (props) => {
 	 */
 	const onLoadStart = async ({ nativeEvent }) => {
 		const { hostname } = new URL(nativeEvent.url);
+
+		if (nativeEvent.url !== url.current) {
+			changeUrl({ ...nativeEvent }, 'end-promise');
+			changeAddressBar({ ...nativeEvent }, 'end-promise');
+		}
+
 		if (!isAllowedUrl(hostname)) {
 			return handleNotAllowedUrl(nativeEvent.url);
 		}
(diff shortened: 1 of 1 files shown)


app/components/Views/BrowserTab/index.js
```diff
@@ -857,6 +857,12 @@ export const BrowserTab = (props) => {
 	 */
 	const onLoadStart = async ({ nativeEvent }) => {
 		const { hostname } = new URL(nativeEvent.url);
+
+		if (nativeEvent.url !== url.current) {
+			changeUrl({ ...nativeEvent }, 'end-promise');
+			changeAddressBar({ ...nativeEvent }, 'end-promise');
+		}
+
```
jpcloureiro, line 865: I see that both `changeUrl` and `changeAddressBar` are called later inside this `onLoadStart` function. `changeUrl` is called unconditionally but `changeAddressBar` is called only on Android devices.
Should we maintain that logic?

I also noticed that both `changeUrl` and `changeAddressBar` accept a `type` parameter but they are not using it inside their body.

blackdevelopa: You are right. The type isn't used so I removed it. For the former, I might have to revert on it.

sethkfman (approved):
LGTM

chrisleewilcox:
```
Scenario: User taps on spoof link in browser
```
Given I have MetaMask mobile open on my device
And I am on the browser screen
And I have a tab open on unknown site
When I tap on spoof link to "https://opensea.io"
And I am redirected to "https://opensea.io"
And I tap back for the browser
Then spoof page is displayed
And browser address bar displays spoof page url
And "opensea.io" is not displayed
```
```

chrisleewilcox (approved):
LGTM!

PR #4138, opened 2022-04-20
Add SDK instance error

wachunei:
This PR adds `sdkError` to the context, this will be helpful to catch and report an error when creating an On Ramp SDK instance.


wachunei:
I'll merge this.

PR #4140, opened 2022-04-21
Upgrade on-ramp-sdk to v0.0.24

wachunei:
This PR upgrades on-ramp-sdk to v0.0.24
Changes:
* Upgrade on-ramp-sdk
* Refactor method calls
* Country was removed and now only region is persisted
* Added country support check when entering the flow
* Order processor now access SDK directly and instances `SDK.orders()` when needed, this means provider was removed from the order processor.

### DIFF
--- a/app/components/Nav/Main/MainNavigator.js
+++ b/app/components/Nav/Main/MainNavigator.js
@@ -305,6 +305,11 @@ const FiatOnRampAggregator = () => (
 			<Stack.Screen name="GetQuotes" component={GetQuotes} options={GetQuotes.navigationOptions} />
 			<Stack.Screen name="Checkout" component={CheckoutWebView} options={CheckoutWebView.navigationOptions} />
 			<Stack.Screen name="Region" component={Region} options={Region.navigationOptions} />
+			<Stack.Screen
+				name="Region_hasStarted"
+				component={Region}
+				options={{ ...Region.navigationOptions, animationEnabled: false }}
+			/>
 			<Stack.Screen name="TransactionDetails" component={TransactionDetails} />
 		</Stack.Navigator>
 	</FiatOnRampSDKProvider>
--- a/app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
+++ b/app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
@@ -91,19 +91,13 @@ const AmountToBuy = () => {
 		selectedChainId,
 	} = useFiatOnRampSDK();
 
-	const selectedFiatCurrency = selectedFiatCurrencyId || selectedRegion?.currency;
-
-	const [{ data: countries, isFetching: isFetchingCountries, error: errorCountries }, queryGetCountries] =
-		useSDKMethod({
-			method: 'getCountries',
-			onMount: false,
-		});
+	const [{ data: countries, isFetching: isFetchingCountries, error: errorCountries }] = useSDKMethod('getCountries');
 
 	const [{ data: dataTokens, error: errorDataTokens, isFetching: isFetchingDataTokens }] = useSDKMethod(
 		'getCryptoCurrencies',
 		selectedRegion?.id,
 		selectedPaymentMethod,
-		selectedFiatCurrency
+		selectedFiatCurrencyId || ''
 	);
 
 	const [{ data: defaultCurrency, error: errorDefaultCurrnecy, isFetching: isFetchingDefaultCurrency }] =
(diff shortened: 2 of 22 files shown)


app/components/Nav/Main/MainNavigator.js
```diff
@@ -305,6 +305,11 @@ const FiatOnRampAggregator = () => (
 			<Stack.Screen name="GetQuotes" component={GetQuotes} options={GetQuotes.navigationOptions} />
 			<Stack.Screen name="Checkout" component={CheckoutWebView} options={CheckoutWebView.navigationOptions} />
 			<Stack.Screen name="Region" component={Region} options={Region.navigationOptions} />
+			<Stack.Screen
```
alaahd, line 308: What is the point of using a new screen here?

wachunei: I does not have the animation when opening the flow and user has already clicked on "Get Started"

app/components/UI/FiatOnRampAggregator/Views/Region.js
```diff
@@ -18,47 +18,41 @@ import RegionAlert from '../components/RegionAlert';
...
-		[hideRegionModal, setSelectedCountry, setSelectedRegion]
+		[hideRegionModal, setSelectedRegion]
 	);
 
-	const handleUnsetRegion = useCallback(() => {
-		setSelectedRegion(undefined);
-	}, [setSelectedRegion]);
+	const updatedRegion = useMemo(() => {
+		if (!selectedRegion || !data) return null;
+		const allRegions = data.reduce((acc, region) => [...acc, region, ...(region.states || [])], []);
```
alaahd, line 45: Since you are flattening the list of regions, I think It would make sense to filter out regions with states as those regions can not be selected, the backend is flagging them as "unsupported" so I suggest you can do chain a filter to allRegions as following:
```suggestion
		const allRegions = data.reduce((acc, region) => [...acc, region, ...(region.states || [])], [])
		                                        .filter(({states}) => !states)
```
by doing that you create a flat list of leaf nodes only.

app/components/Nav/Main/MainNavigator.js
alaahd: Which animation exactly, as I remember, we don't have any animation in the region selection screen. also where exactly we are reading the flag `animationEnabled` ?

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
```diff
@@ -258,19 +244,35 @@ const AmountToBuy = () => {
...
 	useEffect(() => {
 		if (!isFetchingDataTokens && !errorDataTokens && dataTokens) {
-			setTokens(dataTokens);
-			setSelectedAsset(
-				dataTokens.some((a) => a.symbol === 'ETH') ? dataTokens.find((a) => a.symbol === 'ETH') : dataTokens[0]
+			const filteredTokens = dataTokens.filter(
+				(token) => Number(token.network?.chainId) === Number(selectedChainId)
 			);
+			setTokens(filteredTokens);
+			setSelectedAsset(filteredTokens.find((a) => a.address === NATIVE_ADDRESS) || dataTokens[0]);
```
alaahd, line 251: Great catch defaulting the network native currency!

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
```diff
@@ -258,19 +244,35 @@ const AmountToBuy = () => {
...
+		}
+	}, [errorDataTokens, isFetchingDataTokens, dataTokens, setSelectedAsset, selectedChainId]);
+
+	useEffect(() => {
+		if (!isFetchingGetPaymentMethods && !errorGetPaymentMethods && paymentMethods) {
+			const paymentMethod = paymentMethods.find((pm) => pm.id === selectedPaymentMethod);
+			if (!paymentMethod) {
+				setSelectedPaymentMethod(paymentMethods?.[0]?.id);
```
alaahd, line 259: Any idea why you are setting selected payment method again? we are already doing that within `PaymentMethodModal.js` I am kind of a bit unclear on why we need this logic here?

app/components/Nav/Main/MainNavigator.js
wachunei: With animation (Region screen slides in from the right): 

https://user-images.githubusercontent.com/1024246/165308821-ac00fce1-ceb2-44f4-ab6f-cfcbed4fdc93.mp4

Without animation (Region screen replaces Get Started): 

https://user-images.githubusercontent.com/1024246/165309100-ea40c56e-51c9-48c4-8cd9-74ce3da1b3b7.mp4

app/components/UI/FiatOnRampAggregator/Views/Region.js
wachunei: We just do this to find the selected region. By filtering out the "states" region we break the behavior of the modal selector.

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
wachunei: We need this because of this bug: https://www.notion.so/consensyscodefi/Amount-to-Buy-screen-Payment-method-empty-after-changing-Region-c2b5d2fa6b4745e096db633515b677c0

Payment methods can become unavailable and we need to fallback to a default. This is the same approach we're taking on currencies and cryptocurrencies

alaahd (approved):

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
```diff
@@ -243,14 +243,14 @@ const AmountToBuy = () => {
 
 	// side effect to load available crypto assets to purchase using SDK method: getCryptoCurrencies
 	useEffect(() => {
-		if (!isFetchingDataTokens && !errorDataTokens && dataTokens) {
+		if (!isFetchingDataTokens && !errorDataTokens && dataTokens && !selectedAsset) {
```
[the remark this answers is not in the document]
wachunei: Current behavior is when `chainId` changes, then `selectedAsset` becomes `null`.
So the behavior you are describing is not going to happen.

If in the future we want to keep `selectedAsset` value when parameters change (imagine because of a good reason) we need to add another effect, which is to check the selected asset is included in the `tokens`, otherwise remove it from selection.

```tsx
useEffect(() => {
  if (tokens && selectedAsset && !tokens.find((token) => token.address === selectedAsset.address)) {
    setSelectedAsset(null);
  }
}, [tokens, setSelectedAsset, selectedAsset]);
```

alaahd (approved):

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
```diff
@@ -158,10 +158,14 @@ const AmountToBuy = () => {
 	 * Select the default fiat currency if current selection is not available.
 	 */
 	useEffect(() => {
-		if (fiatCurrencies && !fiatCurrencies?.find?.((currency) => currency.id === selectedFiatCurrencyId)) {
-			setSelectedFiatCurrencyId(defaultFiatCurrency);
+		if (
+			fiatCurrencies &&
+			defaultFiatCurrency &&
+			!fiatCurrencies?.find?.((currency) => currency.id === selectedFiatCurrencyId)
```
alaahd, line 164: I prefer to use `.some` for boolean checking and it is safe to offload all optional chaining operators: 
`!fiatCurrencies.some.((c) => c.id === selectedFiatCurrencyId)`

alaahd (approved):

alaahd (approved):

alaahd (approved):

PR #4152, opened 2022-04-25
skeleton views for region, payment method and amount to buy screens

VikChawla:
**Description**

Added skeleton components for region, payment method and amount to buy screens

**Checklist**

**Issue**

Progresses #???


app/components/UI/FiatOnRampAggregator/Views/Region.js
```diff
@@ -64,10 +87,20 @@ const Region = () => {
 	if (isFetching || !data) {
 		return (
 			<ScreenLayout>
-				<ScreenLayout.Body></ScreenLayout.Body>
+				<ScreenLayout.Body>
```
alaahd, line 90: It would be better if we can offload styling and Skelton elements to standalone components, so you only import a single component when content is loading! this approach will help us to avoid polluting the code.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.js
```diff
@@ -11,13 +11,40 @@ import WebviewError from '../../../UI/WebviewError';
...
+	lineTwo: {
+		padding: 8,
+		marginHorizontal: 40,
+	},
+	boxMargin: {
+		marginVertical: 10,
+	},
 });
 
+const SkeletonPaymentOption = () => (
```
alaahd, line 31: offload all Skelton related components and styling into standalone component.

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
```diff
@@ -293,7 +331,36 @@ const AmountToBuy = () => {
 	) {
 		return (
 			<ScreenLayout>
-				<ScreenLayout.Body></ScreenLayout.Body>
+				<ScreenLayout.Body>
```
alaahd, line 334: Offload all Skelton related styling and JSX into standalone component, example would be: `SkeltonAmountToBuy.jsx`

wachunei (changes requested):
I agree with Alaa, but don't create ad-hoc components as files, just extract and componentize the skeleton views.
Some key indicators of acceptance criteria:
- ⚠️ No extra styling should be added or removed from a View because they're using a Skeleton component
- Take advantage of props! Imagine `<SkeletonText short|medium|long|longer small|tall>`, the same for the Icon/Amount Label/Pill
- Avoid styling as "lineOne", "lineTwo", etc. These components should be used for general purpose and should just be dropped in. Random example (notice no custom layout/styling):

```jsx
<Box>
  <ListItem.Content>
      <ListItem.Icon><SkeletonIcon large /></ListItem.Icon>
      <ListItem.Body><SkeletonText tall long /></ListItem.Body>
      <ListItem.Amounts><SkeletonText short /></ListItem.Amounts>
  </ListItem.Content>
  <ListItem.Actions><SkeletonPill medium /> <SkeletonPill medium /></ListItem.Actions>
</Box>
```

alaahd:
I feel we can still create ad-hoc components to ease reusability of these components and have the lowest amount of code pollution in main views, In Atomic Design Methodology, `SkeletonIcon` `SkeletonText` `SkeletonPill` will act as **atoms** and the components I am advocating for will act as **molecules** which group different atoms to form a more complex/reusable Skelton view.

wachunei:
Screen Shot 2022-04-26 at 08 25 21

That composition of components must explicitly be declared in these "is loading" rendering section and not in other files, since they are just composed atoms without any behavior or interaction.

alaahd:
So imagine you have 20 views with the same shape or same skelton! are you going to compose 20 times ? what if that Skelton need to be changed later due to UI changes ?

alaahd:
Ok perfect!

wachunei:
Superseded by https://github.com/MetaMask/metamask-mobile/pull/4207

Issue #3965, opened 2022-03-28
Deep link opening iOS AppStore even when MetaMask is installed


epshtielsl:
**Describe the bug**
Deep links are opening up the App Store and prompting users to install MetaMask, even when the app is already installed on the devicie.

**Screenshots**
https://community.metamask.io/t/deeplink-opens-appstore-when-app-installed/18199/12

**To Reproduce**
Open any app that utilizes deep links to integrate with MetaMask, i.e. dapps that utilize WalletConnect.  When you attempt to sign a transaction, or other requests, MetaMask will open briefly, then the Apple AppStore opens up.

**Expected behavior**
MetaMask opens and performs the task requested.

**Smartphone (please complete the following information):**
 - Device: iPhone 12 Pro
 - OS: iOS 15.3.1
 - App Version: 4.2.2

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

andreahaku:
Thanks @epshtielsl for opening the issue. We are investigating it and trying to find the reason for this behaviour. Thanks again.

rafaelaugustos:
any update?

zosim:
> any update?

yes also hoping for some news. Thanks.

thunderball1:
@andreahaku is there any update? this is happening for very long time

liho00:
same issue, izit IOS or IOS version issues?

andreahaku:
Did you try the latest version of the app (5.0.x) as most of the deep links related issue should be solved with it. Please let me know. Thanks!

liho00:
> Did you try the latest version of the app (5.0.x) as most of the deep links related issue should be solved with it. Please let me know. Thanks!

Same not working, here is the screen recording https://user-images.githubusercontent.com/77238199/165326244-c9471c12-f88b-4ff8-b207-80f410623c51.MOV

un7c0rn:
Same issue here, using Metamask version 5.0.1. The deeplink from WC is `https://metamask.app.link/wc`––is this to spec?

un7c0rn:
According to this we're not generating the right deep link:

# For deep links
[...]

PR #4167, opened 2022-04-27
[5.2][IMPROVEMENT] Adds support for 'dapp/' urls support on 'metamask://' and fixes DL opening to Apple Store

andreahaku:
**Description**

Adds support for 'dapp/' urls support on 'metamask://' scheme.

So that it'll be possible to use links like:
```metamask://dapp/google.com``` and have the same results of ```https://metamask.app.link/dapp/google.com```.

This is not a deep link so it won't check if the app is installed and if not, take the user to stores to download it.
This is only works if the app is already installed (otherwise the device doesn't know how to deal with ```metamask://``` links.

Reason to use this over Branch.io links is that it does not depend on it and so can overcome some of the limitations of them, like sometimes not opening when programmatically called from dapps (without a user click/tap).

It also fixes the issue related to deep links going to Apple store on iOS even if the app is installed.

**Checklist**

**Issue**

Progresses #3965


andrepimenta (approved):
LGTM! 👍

cortisiko (approved):
I was able to deep link using universal deep links and url scheme deep links on both iOS and Android. This guy is 🌮 🌮 🌮

Issue #3331, opened 2021-10-23
[5.2] Pressing "home" does not show favorites


danfinlay:
**Describe the bug**
No favorites are shown, even when the user has favorites

**Screenshots**
![IMG_8268](https://user-images.githubusercontent.com/542863/138572752-63deaa91-d46c-4cf0-b55c-7abc8f7c8cb6.jpg)

**To Reproduce**
1. Save some favorites
2. Tap the "home" icon in the bottom bar
3. Notice "no favorites yet".

**Expected behavior**
I expect to see favorites. I can find them by going into tabs and creating a new tab, that path correctly shows favorites.

**Smartphone (please complete the following information):**
 - Device: iPhone 11 Pro
 - OS: iOS 14.8
 - App Version 3.4.1

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

gantunesr:
Internal conversation:https://consensys.slack.com/archives/CBW7S9FSN/p1638820384468400?thread_ts=1638810018.460500&cid=CBW7S9FSN

gantunesr:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/MetaMask-Mobile-5f984938ddc0e4001d4b79cb/issues/MetaMask/metamask-mobile/3331?planning-poker) with ZenHub @Fatxx @plasmacorral

dynahpointch:
WHEN THIS WILL BE FIXED OPEN NOW SINCE TWO YEARS ON REDDIT AND OTHER FORUMS ?
Please when this will bi addressed in the next 50 years and fixed ?

AlexHerman1:
woot!

PR #4175, opened 2022-04-28
[5.2][Fix] Favourites not showing when home button is pressed in browser tab menu

Fatxx:
**Description**

Fix no favourites shown when "home" is pressed in the browser, scripts are now guaranteed to be present when the website is loaded.

This is blocked by https://github.com/MetaMask/dapps/pull/122 and should be tested and merged together.

**How to test**

- Make sure you check out both branches (app and dapp repo);
- On [this line](https://github.com/MetaMask/metamask-mobile/blob/fix/home-favorites/app/core/AppConstants.js#L32) the URL should match your app local URL either `http://localhost:3000/` or `http://${your-ip}:3000/`
- Also on [this line](https://github.com/MetaMask/metamask-mobile/blob/fix/home-favorites/app/components/Views/BrowserTab/index.js#L67) it should be changed accordingly like `localhost:3000` or `{your-ip}:3000`

**Checklist**

**Issue**

Progresses #3331


andrepimenta (changes requested):
We probably need to check if the user is on the homepage before we inject the favourites otherwise we will inject them into all websites. Does it need to be on RPCMethods?

Fatxx:
It should only inject on the "metamask dapps web app" since the event is triggered from there. 

Maybe not I just used the RpcMethod logic because we were already using it for other events but we can create a custom event for that.

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -510,6 +512,10 @@ export const getRpcMethodMiddleware = ({
 				res.result = true;
 			},
 
+			metamask_onAppMounted: async () => {
```
andrepimenta, line 515: The issue would be if a malicious dapp calls this rpc method, it would get the favourites from the user

andrepimenta (changes requested):

app/core/RPCMethods/RPCMethodMiddleware.ts
andrepimenta: If this rpc method is needed, then we should check if it's the homepage
`isHomepage()`

andrepimenta (changes requested):

app/core/RPCMethods/RPCMethodMiddleware.ts
Fatxx: Done

andrepimenta (approved):
LGTM!

chrisleewilcox (approved):
iOS 15.0 Simulator = Pass
iOS 15.5 iPhone 12 = Pass
Android 10 Pixel 3 Simulator = Pass
Android 12 Samsung A42 = Pass

Browser Favorites
Tap Home button shows Favorites = Pass
New Browser Tab show Favorites = Pass
Delete Favorites = Pass

PR #4183, opened 2022-04-28
[FIX] - audit dependency exclusion

sethkfman:
- added yarn audit exclusion 
- doesn't seem to impact the code base
- the issue references a patch for `node-fetch: 2.6.7` which we have in the resolutions in the package.json


gantunesr (approved):
lgtm

Issue #4149, opened 2022-04-25
[UI] Add screen to select hardware wallet


gantunesr:
### **Description**

Develop a new screen where the user can select which Hardware wallet they want to connect to MMM

### **Acceptance Criteria**

1. Selecting Hardware Device

GIVEN I'm using MMM to connect a hardware wallet
WHEN the flow starts
THEN I should have the option to select the hardware I want to connect
AND be directed to the correct flow 

### **References**

-   Figma design: https://www.figma.com/file/gw8z5Y9rT017ZKnc30M4dy/MetaMask-Mobile-ft-Ledger?node-id=0%3A1121

PR #4192, opened 2022-05-02
[IMPROVEMENT] Add screen to select hardware wallet

gantunesr:
**Description**

This task adds a screen to let the user select a screen during the connect hardware wallet flow.

**Checklist**

**Issue**

Progresses #4149


app/components/Nav/Main/MainNavigator.js
```diff
@@ -437,6 +438,16 @@ const SetPasswordFlow = () => (
   </Stack.Navigator>
 );
 
+const ConnectHardwareWalletFlow = () => (
+  <Stack.Navigator name="ConnectHardwareWallet">
+    <Stack.Screen
+      name="SelectHardwareWallet"
```
owencraston, line 444: I know this is not a pattern yet but I am going to advocate to keep screen names as constants so they can be references in multiple locations. This is just a suggestion and not a blocker

gantunesr: Good suggestion. Lets start a new branch to integrate this pattern

app/components/Views/ConnectHardware/SelectHardware/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+      getNavigationOptionsTitle(
+        strings('connect_hardware.title_select_hardware'),
+        navigation,
+        false,
+        colors,
+      ),
+    );
+  }, [navigation, colors]);
+
+  const navigateToConnectQRWallet = () => {
```
owencraston, line 95: I cant help but think we should try and figure out a way to create beta flags. I don't think we can merge this without hiding this some how since it might get people prematurely excited about ledger and then disappoint them. Let me know if there is a plan for something like this or how you planned to release this.

owencraston:
It seems we are using out of date ledger assets. Here are the most up to date ones: https://www.ledger.com/press

app/components/Views/ConnectHardware/SelectHardware/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+const ledgerLogoDarkImgPath = 'images/ledger-dark.png';
+const ledgerLogoDark = require(ledgerLogoDarkImgPath);
+
+// QR Hardware Logo
+const qrHardwareLogoLightImgPath = 'images/qrhardware-light.png';
+const qrHardwareLogoLight = require(qrHardwareLogoLightImgPath);
+
+const qrHardwareLogoDarkImgPath = 'images/qrhardware-dark.png';
+const qrHardwareLogoDark = require(qrHardwareLogoDarkImgPath);
```
owencraston, line 77: We should abstract this into as HardwareImageComponent that takes options and we can handle the dark/light mode logic all in one spot.

zoltanarvai (approved):
Looks good, we can upgrade the Ledger Logo

app/components/Views/ConnectHardware/SelectHardware/index.tsx
zoltanarvai: This would go into the feat/ledger-integration branch. So only gets merge into the final product when we have ledger support

gantunesr: I don't think there is a plan to integrate beta or feature flags.

This screen will be merged into `feat/ledger-integration` and released alongside the ledger integration. I think having flags will be useful in the future, but for this particular case they're not necessary.

app/components/Views/ConnectHardware/SelectHardware/index.tsx
gantunesr: Good observation, I'll sync with the design team to check if there are plans to develop that component. This pattern can be found in other part of the code too, we should update those with that abstraction.

gantunesr:
I'll merge this PR under the following conditions:

1. New PR to refactor the name of the screens as constants.
2. Develop a `Image` component to handle dark mode
3. Replace outdated Ledger logo

app/components/Views/ConnectHardware/SelectHardware/index.tsx
owencraston: Agreed, I was just wondering. I didn't realize this was going into a feature branch.

Issue #3880, opened 2022-03-10
Deep linking to the approval flow switches networks correctly but does not show approval modal


cortisiko:
**Describe the bug**
Basically If you deep link here: https://metamask.app.link/send/0x01BE23585060835E02B77ef475b0Cc51aA1e0709@4/approve?address=0x178e3e6c9f547A00E33150F7104427ea02cfc747&uint256=5e8

The network updates correctly but the approval modal does not appear. If you try interacting with the app nothing happens. You would need to kill the app and relaunch to start back interacting with the app. You would also need to deep link again for the approval modal to appear. 

**Screen recording**
[see recording](http://recordit.co/HQB8MEF0O1)

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

tommasini:
@cortisiko In this last main branch update it's working fine with me, can you re-check if it keeps happening with you please?

Fatxx:
@chrisleewilcox can also re-check

cortisiko:
@tommasini @Fatxx I can confirm this is still happening. See recording [here](https://github.com/MetaMask/metamask-mobile/issues/3880). This was on the v5.1 RC candidate based off on main. I used my physical iPhone device running iOS 14.8. I also tested on my physical iPhone 11 running iOS 15.2 and I still encountered the issue.

PR #4204, opened 2022-05-03
Shows the modal when changes from a different network

tommasini:
**Description**
The problem was when we press a deeplink the network changes but the modal doesn't appear, because the state of redux was not updated yet with the chain id.

**Proposed Solution**
We update the transaction metadata with the chain id that comes with the url

**Code Impact**
Low

**Test Cases**
Case 1 (with chain id on the deep link):
 * Is in a different network
 * open the deeplink
 * the app will change the network to the chain id network on the link 
 * The modal will open
 Case 2 (without chain id on the deep link):
* Open the deeplink
* will open the modal in the current network

**Screenshots/Recordings**

 

**Checklist**

**Issue**

Progresses #3880


tommasini:
Closing this, the issue is already solved

PR #4248, opened 2022-05-06
made quick amounts dymanic

VikChawla:
**Description**

_Write a short description of the changes included in this pull request._

**Checklist**

**Issue**

Progresses #???

### DIFF
--- a/app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
+++ b/app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
@@ -207,6 +207,15 @@ const AmountToBuy = () => {
     selectedPaymentMethodId,
   );
 
+  const [{ data: limits, error: errorLimits, isFetching: isFetchingLimits }] =
+    useSDKMethod(
+      'getLimits',
+      selectedRegion?.id,
+      selectedPaymentMethodId,
+      selectedAsset?.id ? selectedAsset.id : '/currencies/crypto/1/eth',
+      selectedFiatCurrencyId,
+    );
+
   /**
    * * Defaults and validation of selected values
    */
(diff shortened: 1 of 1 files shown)


app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.js
```diff
@@ -634,17 +645,10 @@ const AmountToBuy = () => {
...
-                  { value: 300, label: '$300' },
-                  { value: 400, label: '$400' },
-                  // eslint-disable-next-line no-mixed-spaces-and-tabs
-                ]
-              : []
-          }
+          amounts={limits?.quickAmounts.map((limit) => ({
+            value: limit,
+            label: currentFiatCurrency?.denomSymbol + limit.toString(),
+          }))}
```
wachunei, line 651: Add a default empty array so `QuickAmounts` does not crash.
```suggestion
          amounts={limits?.quickAmounts.map((limit) => ({
            value: limit,
            label: currentFiatCurrency?.denomSymbol + limit.toString(),
          })) ?? []}
```

VikChawla: All other options except for US, Pounds, and Euros will return an empty quickAmounts array. Should I still make that fallback?

VikChawla: > Add a default empty array so `QuickAmounts` does not crash.

All other options except for US, Pounds, and Euros will return an empty quickAmounts array. Should I still make that fallback?

wachunei: Oh nice, then we don't 👌

wachunei (approved):

PR #4249, opened 2022-05-06
patch cross-fetch instead of skipping

rickycodes:
**Description**

this patches `cross-fetch` in a "non-breaking" manor. once a new release of `controllers` is cut and included we can delete the first resolution here since that's being handled by: https://github.com/MetaMask/controllers/pull/817. we can remove the second one if we determine it's safe to bump `react-native-gesture-handler` to latest.


sethkfman (approved):
LGTM

PR #4269, opened 2022-05-10
[5.3] Show amount being approved by default on approval screens

blackdevelopa:
**Description**

This PR should show the proposed approval/custom spend limit on the approval screen

**Checklist**

**Issue**

Progresses #4148


sethkfman (approved):
LGTM

sethkfman:
@blackdevelopa if this ticket is ready. I would move it into QA

blackdevelopa:
> @blackdevelopa if this ticket is ready. I would move it into QA

I am actually waiting on Saya for a design update on the approval screen before moving this to QA

plasmacorral:
Currently hash a2ddea4f85c1c211438f3278edd3b01f9aaa34cd is not showing these updates when using the MetaMask swaps button on Binance network in Android 11, 12 or iOS 15.5.

Screen Shot 2022-06-06 at 12 54 00 PM

Please make these changes in the swaps flow as well.

plasmacorral:
One more note here: In this branch the "proposed approval limit" is showing as a long number with commas that includes a whole bunch of zeros. (please forgive my terribly out of focus screenshot, but you can at least see all the zeros)
Screen Shot 2022-06-07 at 10 53 32 AM

Typically in production, this proposed approval limit value is the numeric equivalent of ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff which is equal to 115792089237316195423570985008687907853269984665640564039457584007913129639935.

plasmacorral:
I am still not seeing the change to more prominently feature the approval amount when using MetaMask swaps on Polygon/Binance chains.  
<div style="display: inline-block">

</div>

blackdevelopa:
Adding this comment for the Swap flow and removing the `QA issues found` label. https://consensys.slack.com/archives/C03BGLXT03E/p1654878051963769?thread_ts=1654184068.226649&cid=C03BGLXT03E

plasmacorral:
Android 12
Pixel 3a

Screen Shot 2022-06-10 at 6 11 08 PM

It seems that the part of the modal where this new information is presented is not scroll-able.  This actually serves to blind android users operating in split screen mode.  With this change as tested in d1e56723df86b01ffce9c1b98bd698ad6053526d we actually no longer present even a mention of the spending allowance amount or provide an option to edit within the approval modal.  The exception being, that if the user taps "View details" they can view but not edit the token spending allowance.

While the experience on the split screen using pancakeswap.finance is tough to say the least, but I found it surprising how many android users were recently encountering the app closure in android split screen. While this may be a relatively small segment of users, the implications for them here creates what I believe to be a merge blocker.  We should not further obfuscate the spending approval for this small segment of users in our effort to expand visibility and provide informed consent around spending approvals.

A possible solution would be to support scrolling for the entire approval modal, rather than just the bottom as we do today.

omnat:
Per the discussion, let's not consider this issue as a release blocker; i.e., the issue of approval amount not visible when user is in split screen mode. 

@sethkfman will write a ticket for upcoming sprint to disable split screen mode unless user is on a tablet (which is when split screen is likely more common and real estate not an issue for hiding consent visibility). 

cc @plasmacorral @jakehaugen @gantunesr

plasmacorral:
This is a non-blocking observation that may merit further investigation or may simply be a matter of my connectivity challenges today.

Android 11
Samsung A515f

https://recordit.co/oA3w4FoPyE
Screen Shot 2022-06-14 at 6 44 20 PM


I have been experiencing connectivity issues today, but this "Give permission to access your [missing {{token symbol}} value]?"  hung for nearly 3 seconds before the Token and spending approval amount was resolved. 

This was for a a non-native asset that had not yet been imported into MetaMask. 

Reproduction steps 1-14 cited in my earlier comments were used to get here.

PR #4277, opened 2022-05-11
Keystone: Fix swap transaction 

soralit:
**Description**
This PR fixed the situation that when user want to swap assets but exceeded the spend limit, if user rejected the approve transaction, the swap transaction will still appears which is a lot of confusing. 

**Issue**

new dependency: async-mutex. 

I add this dependency because UI will handle the 2 incoming transactions immediately so it is impossible to block and cancel the second swap transaction in UI layer. So I add the mutex and check the state in transactionController to see whether the second transaction was canceled.

### DIFF
--- a/app/components/Nav/Main/RootRPCMethodsUI.js
+++ b/app/components/Nav/Main/RootRPCMethodsUI.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect, useCallback, useMemo } from 'react';
 
 import { StyleSheet, Alert, InteractionManager } from 'react-native';
 import PropTypes from 'prop-types';
@@ -23,12 +23,12 @@ import WalletConnect from '../../../core/WalletConnect';
 import {
   getMethodData,
   TOKEN_METHOD_TRANSFER,
-  APPROVE_FUNCTION_SIGNATURE,
-  decodeApproveData,
   getTokenValueParam,
   getTokenAddressParam,
   calcTokenAmount,
   getTokenValueParamAsHex,
+  isApproveTransaction,
+  isSwapTransaction,
 } from '../../../util/transactions';
 import { BN } from 'ethereumjs-util';
 import Logger from '../../../util/Logger';
@@ -57,6 +57,7 @@ import { mockTheme, useAppThemeFromContext } from '../../../util/theme';
 import withQRHardwareAwareness from '../../UI/QRHardware/withQRHardwareAwareness';
 import QRSigningModal from '../../UI/QRHardware/QRSigningModal';
 import { networkSwitched } from '../../../actions/onboardNetwork';
+import { Mutex } from 'async-mutex';
 
 const hstInterface = new ethers.utils.Interface(abi);
 
@@ -67,6 +68,7 @@ const styles = StyleSheet.create({
   },
 });
 const RootRPCMethodsUI = (props) => {
+  const swapMutex = useMemo(() => new Mutex(), []);
   const { colors } = useAppThemeFromContext() || mockTheme;
   const [showPendingApproval, setShowPendingApproval] = useState(false);
   const [signMessageParams, setSignMessageParams] = useState({ data: '' });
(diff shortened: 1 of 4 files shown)


app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -222,9 +224,61 @@ const RootRPCMethodsUI = (props) => {
...
+        state: { transactions },
+      } = TransactionController;
+      const currentIndex = transactions.findIndex(
+        ({ id }) => id === transactionMeta.id,
+      );
+      if (currentIndex < 0) return;
+      const nextTransaction = transactions[currentIndex + 1];
+      if (!nextTransaction) return;
+      if (!isSwapTransaction(nextTransaction)) return;
+      await TransactionController.cancelTransaction(nextTransaction.id);
```
gantunesr, line 268: I'm not sure about this logic

- I see that if the next transaction is not a swap transaction this method will return, but the following transaction could be the swap transaction that should be cancelled
- I think it shouldn't be necessary to call `cancelTransaction`, we should avoid submitting the transaction to the TransactionController in the first place. We can have a meeting to explore this option

gantunesr (commented):
Hey @soralit, I left some comments, let me know what you think

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -222,9 +224,61 @@ const RootRPCMethodsUI = (props) => {
...
+        to &&
+        (swapsUtils.isValidContractAddress(props.chainId, to) ||
+          (data &&
+            data.substr(0, 10) === APPROVE_FUNCTION_SIGNATURE &&
+            decodeApproveData(data).spenderAddress?.toLowerCase() ===
+              swapsUtils.getSwapsContractAddress(props.chainId)))
+      );
+    },
+    [props.chainId],
+  );
```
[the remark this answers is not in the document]
soralit: That's cool.

app/components/Nav/Main/RootRPCMethodsUI.js
soralit: I'm trying to find a way to avoid adding the second transaction. But it would be hard to do without touching too much codes.  @gantunesr 
The problem is the two transactions are added together immediately after user swipe to swap. Codes here: https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/Swaps/QuotesView.js#L879

soralit: About the first question I think it would be hard to identify the target transactions ideally. In practice, it wouldn't has much opportunities for user to insert a non-swap transaction between these two transaction. 

I think the best way is to modify the TransactionController to add the two transactions together in one function call to make sure that they are added in order. 
Another way is that we can create a new data model called transactions group for such situations.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -222,9 +224,61 @@ const RootRPCMethodsUI = (props) => {
...
+      return (
+        transactionMeta.origin === process.env.MM_FOX_CODE &&
+        to &&
+        (swapsUtils.isValidContractAddress(props.chainId, to) ||
+          (data &&
+            data.substr(0, 10) === APPROVE_FUNCTION_SIGNATURE &&
+            decodeApproveData(data).spenderAddress?.toLowerCase() ===
+              swapsUtils.getSwapsContractAddress(props.chainId)))
+      );
+    },
```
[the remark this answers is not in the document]
soralit: Yes, forget to refactor the existing code.

app/components/Nav/Main/RootRPCMethodsUI.js
gantunesr: I see what you mean @soralit, we need to create an alternative flow in swaps give hardware wallet the time to approve both transaction in a more clean way. Let me sync with @AlexJupiter and @holantonela to see if we can develop these new flow for swaps

AlexJupiter: Okay, here's Swaps using Keystone on mobile: https://user-images.githubusercontent.com/7900773/168578894-a9eae0d1-3e6e-461f-89cf-f8929311a257.MP4

Here's using Swaps using Keystone on extension: https://user-images.githubusercontent.com/7900773/168580746-3b4c0ea5-b05f-4ea8-8e46-a7909454c8a1.mov

 @gantunesr @soralit can't we just do the same as extension on mobile? I'm unsure what's happening in the background, but it seems to be using animated QR codes here to perhaps combine these transactions?

gantunesr:
@soralit Please take a look at the PR https://github.com/MetaMask/metamask-mobile/pull/4475 and tell what you think, I was able to solve this issue without the need of cancelling the next transaction

gantunesr:
Closing in favor of [#4475](https://github.com/MetaMask/metamask-mobile/pull/4475)

PR #4285, opened 2022-05-12
Improve error in SDK and views. Clear error and data in useSDKMethod

wachunei:


app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.tsx
```diff
@@ -505,6 +505,14 @@ const AmountToBuy = () => {
    * * Derived values
    */
 
+  const isFetching =
```
alaahd, line 508: we can use `useMemo` to memorize this variable.

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
```diff
@@ -382,6 +382,16 @@ const GetQuotes = () => {
     );
   }
 
+  // Error while FetchingQuotes
```
alaahd, line 385: good catch reordering this.

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.tsx
wachunei: Not expensive enough, would not do it

PR #4157, opened 2022-04-26
Adjust language mapping

rickycodes:
**Description**

Opening this as a draft for now as it relies on #3804 

This PR updates the language mappings in the project to be compatible with Crowdin. It also deletes the old translations in favour of the new ones which are more comprehensive:

image

PR #4349, opened 2022-05-16
Fix subtitle mapping

rickycodes:
this broke partly because of #4157 and partly because we're using these keys for _not_ translations

this fixes it. I'll have to see if there's a way we can mark these keys to _not_ be translated


gantunesr (approved):

cortisiko (approved):
🌮 🌮 🌮

PR #4366, opened 2022-05-18
Fix order id retrieve in checkout

wachunei:
This PR upgrades SDK to version 1.0.3. This fixes a bug where `getOrderIdFromCallback` was throwing a NotFoundException

### DIFF
--- a/app/components/UI/FiatOnRampAggregator/Views/Checkout.tsx
+++ b/app/components/UI/FiatOnRampAggregator/Views/Checkout.tsx
@@ -16,7 +16,6 @@ import Engine from '../../../../core/Engine';
 import { toLowerCaseEquals } from '../../../../util/general';
 import { protectWalletModalVisible } from '../../../../actions/user';
 import {
-  callbackBaseUrl,
   processAggregatorOrder,
   aggregatorInitialFiatOrder,
 } from '../orderProcessor/aggregator';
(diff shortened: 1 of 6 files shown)


app/components/UI/FiatOnRampAggregator/Views/Checkout.tsx
```diff
@@ -91,6 +91,22 @@ const CheckoutWebView = () => {
           params?.provider.id,
           navState?.url,
         );
+
+        if (!orderId) {
+          const parsedUrl = parseUrl(navState?.url);
+          if (Object.keys(parsedUrl.query).length === 0) {
```
alaahd, line 97: I think we should check the query params in the callback URL before we inquire about order id.
```suggestion
          const orders = await SDK.orders();
          const parsedUrl = parseUrl(navState?.url);
          
          if (Object.keys(parsedUrl.query).length === 0) {
            // There was no query params in the URL to parse
            // Most likely the user clicked the X in Wyre widget
             navigation.dangerouslyGetParent()?.pop();
             return;
          } else {
           // It seems we have a search params coming as part of the callback url, lets try to map it with 
            order id using our SDK method
            const orderId = await orders.getOrderIdFromCallback(
                 params?.provider.id,
                 navState?.url,
             );
             if (!orderId) {
                 throw new Error(
                    `Order ID could not be retrieved. Callback was ${navState?.url}`,
                 );
             }
          }
```

wachunei: Fixed!

PR #4376, opened 2022-05-20
[5.2.0][FIX] M3 - Warning on legit address

tommasini:
**Description**
The warning of invalid address was showing before the address was verified but if it was on the address book was already verified.

**Proposed solution**
 Now the address that is on the address book will not be verified again, that way we will not have the delay of verifying the address again.

**Code Impact**
Low, the code will only affect how the SendTo screen works, it's the first screen of the send flow.

**Test Cases**
Case 1:
* In wallet view, press the send button
* In the Address to complete with (in mainnet):
  * ENS name
  * address
  * address from the address book
  * recents
  * from user accounts
* Press next and insert one amount
* In rinkeby network completed the transaction when the address was from
  *  insert normal address
  *  recent 
  *  address from the address book
Case 2:
 * Complete with an unknown address
 * Add the address to the Address book
Case 3:
 * Select the address from the address book
 * Clear address 
Case 4:
 * Select unkown address from recents (will appear the warning because needs to be verified)
 * Clear address
Case 5:
 * Write an unkown address  (will appear the warning because needs to be verified)
 * Clear address
Case 6:
* Write an ens name  (will appear the warning because needs to be verified)
* Clear ens name
Case 7:
* Scan address with QR code
* Clear address
* Scan again and do a transaction

**Checklist**

**Screenshots/Recordings**
Solution without the orange warning:
https://recordit.co/FCljUjFoxQ

**Issue**

Progresses #https://github.com/MetaMask/mobile-planning/issues/278

### DIFF
--- a/app/components/Views/SendFlow/SendTo/index.js
+++ b/app/components/Views/SendFlow/SendTo/index.js
@@ -358,8 +358,33 @@ class SendFlow extends PureComponent {
     });
     this.toggleFromAccountModal();
   };
+  /**
+   * This returns the address name from the address book or user accounts if the selectedAddress exist there
+   * @param {*} toSelectedAddress - Represents the value of what user writes on toAddress input
+   * @returns
+   */
+  getAddressNameFromBookOrIdentities = (toSelectedAddress) => {
+    const { addressBook, network, identities } = this.props;
+    const networkAddressBook = addressBook[network] || {};
 
-  validateAddress = async (toSelectedAddress) => {
+    const checksummedToSelectedAddress = toChecksumAddress(toSelectedAddress);
+
+    if (
+      networkAddressBook[checksummedToSelectedAddress] ||
+      identities[checksummedToSelectedAddress]
+    ) {
+      return (
+        (networkAddressBook[checksummedToSelectedAddress] &&
+          networkAddressBook[checksummedToSelectedAddress].name) ||
+        (identities[checksummedToSelectedAddress] &&
+          identities[checksummedToSelectedAddress].name)
+      );
+    }
+
+    return null;
+  };
+
+  validateAddressOrENSFromInput = async (toSelectedAddress) => {
     const { AssetsContractController } = Engine.context;
     const { addressBook, network, identities, providerType } = this.props;
     const networkAddressBook = addressBook[network] || {};
@@ -369,8 +394,7 @@ class SendFlow extends PureComponent {
       errorContinue,
       isOnlyWarning,
       confusableCollection,
-      toEnsAddressResolved,
-      isFromAddressBook;
+      toEnsAddressResolved;
     let [addToAddressToAddressBook, toSelectedAddressReady] = [false, false];
     if (isValidAddress(toSelectedAddress)) {
       const checksummedToSelectedAddress = toChecksumAddress(toSelectedAddress);
@@ -385,17 +409,10 @@ class SendFlow extends PureComponent {
           addToAddressToAddressBook = true;
         }
       } else if (
-        networkAddressBook[checksummedToSelectedAddress] ||
-        identities[checksummedToSelectedAddress]
+        !networkAddressBook[checksummedToSelectedAddress] &&
+        !identities[checksummedToSelectedAddress]
       ) {
-        toAddressName =
-          (networkAddressBook[checksummedToSelectedAddress] &&
-            networkAddressBook[checksummedToSelectedAddress].name) ||
-          (identities[checksummedToSelectedAddress] &&
-            identities[checksummedToSelectedAddress].name);
-        isFromAddressBook = true;
-      } else {
-        toAddressName = ens || toSelectedAddress;
+        toAddressName = toSelectedAddress;
         // If not in address book nor user accounts
         addToAddressToAddressBook = true;
       }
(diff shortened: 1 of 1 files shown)


app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -358,8 +358,35 @@ class SendFlow extends PureComponent {
     });
     this.toggleFromAccountModal();
   };
+  /**
+   * This returns the address name from the address book or user accounts if the selectedAddress exist there
+   * @param {*} toSelectedAddress - Represents the value of what user writes on toAddress input
```
gantunesr, line 363: Is `toSelectedAddress` a `string`? Add the correct data type.

```suggestion
   * @param {String} toSelectedAddress - Represents the value of what user writes on toAddress input
```

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -358,8 +358,35 @@ class SendFlow extends PureComponent {
...
+   * @returns
```
gantunesr, line 364: What's this method returning? `string | null`?

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -358,8 +358,35 @@ class SendFlow extends PureComponent {
...
+   */
+  getAddressNameFromBookOrIdentities = (toSelectedAddress) => {
+    if (!toSelectedAddress) return;
+
+    const { addressBook, network, identities } = this.props;
+    const networkAddressBook = addressBook[network] || {};
+
+    const checksummedToSelectedAddress = toChecksumAddress(toSelectedAddress);
```
gantunesr, line 372: Improve variable name
```suggestion
    const checksummedAddress = toChecksumAddress(toSelectedAddress);
```

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -358,8 +358,35 @@ class SendFlow extends PureComponent {
...
+      return (
+        (networkAddressBook[checksummedToSelectedAddress] &&
+          networkAddressBook[checksummedToSelectedAddress].name) ||
+        (identities[checksummedToSelectedAddress] &&
+          identities[checksummedToSelectedAddress].name)
+      );
+    }
 
-  validateAddress = async (toSelectedAddress) => {
+    return null;
```
gantunesr, line 386: I see there's some unnecessary validations in this method. For example, if `networkAddressBook[checksummedToSelectedAddress]` is falsy, you're going to evaluate it again in

```js
networkAddressBook[checksummedToSelectedAddress] && 
  networkAddressBook[checksummedToSelectedAddress].name) ||
```

Which is unnecessary since you already know that the value will be falsy. I recommend using a ternary operator to handle this logic in a more effective way. I think it would be something like this,

```js
return networkAddressBook[checksummedToSelectedAddress] ?
  networkAddressBook[checksummedToSelectedAddress].name :
  identities[checksummedToSelectedAddress] ?
    identities[checksummedToSelectedAddress].name :
    null
```

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -461,25 +480,46 @@ class SendFlow extends PureComponent {
...
-    this.setState({
-      toSelectedAddress,
-    });
+    const addressName =
+      this.getAddressNameFromBookOrIdentities(toSelectedAddress);
+
+    /**
+     * If the address is from addressBook or identities
+     * we do not need to validate since it was already validated before
+     */
```
gantunesr, line 504: Are you a 100% sure of this statement?

gantunesr (changes requested):

app/components/Views/SendFlow/SendTo/index.js
tommasini: Thank you, much better!

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -385,17 +403,10 @@ class SendFlow extends PureComponent {
...
-        toAddressName =
-          (networkAddressBook[checksummedToSelectedAddress] &&
-            networkAddressBook[checksummedToSelectedAddress].name) ||
-          (identities[checksummedToSelectedAddress] &&
-            identities[checksummedToSelectedAddress].name);
-        isFromAddressBook = true;
-      } else {
-        toAddressName = ens || toSelectedAddress;
+        toAddressName = toSelectedAddress;
         // If not in address book nor user accounts
```
gantunesr, line 410:
```suggestion
        // If not in the addressBook nor user accounts
```

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -461,25 +472,46 @@ class SendFlow extends PureComponent {
...
+        toSelectedAddress,
+        toSelectedAddressReady: true,
+        isFromAddressBook: true,
+        toSelectedAddressName: addressName,
+      });
+    } else {
+      this.validateAddressOrENSFromInput(toSelectedAddress);
+      /**
+       * Because validateAddressOrENSFromInput is asynchronous function
+       * we are setting the state here synchronously, so it does not block the UI
```
gantunesr, line 508:
```suggestion
       * Because validateAddressOrENSFromInput is an asynchronous function
       * we are setting the state here synchronously, so it does not block the UI
```

gantunesr (commented):
Some minor nits

plasmacorral:
https://github.com/MetaMask/mobile-planning/issues/278#issuecomment-1136134528

PR #4096, opened 2022-04-12
Feature/3157 token detection v2

Cal-L:
**Description**

This is the implementation for token detection v2 as designed in Figma - https://www.figma.com/file/2h15YfvbT2q4q1FOMJFPP1/Token-Detection-V2?node-id=0%3A1. This is dependent on a controller PR - https://github.com/MetaMask/controllers/pull/763. In this iteration, we introduce a concept of detected tokens, where users will have a choice to either import or ignore certain tokens that has been "detected" in their wallet.

**Acceptance Criteria**

- WIP

**Issue**
#3157

**Figma**
- https://www.figma.com/file/2h15YfvbT2q4q1FOMJFPP1/Token-Detection-V2?node-id=0%3A1

Issue #3157, opened 2021-09-21
Support token detection on custom networks


omnat:
### **Description**
This is MVP v2 of token detection where we add support for custom networks. Supported networks: Mainnet, Binance, Polygon, and Avalanche.

### **Technical Details**
- [ ] Enable Token Detection by default
- [ ] Update `What's new` screen to show carousel with Token Detection information
- [ ] Move Token Detection toggle to `Advanced Settings` screen
- [ ] Show search functionality on supported networks + Update banners
- [ ] Add tokens detected link below token list on `Wallet` screen
- [ ] Add tokens detected action sheet to display detected tokens + ability to ignore or add tokens
- [ ]  Extend action sheet to have ability to drill into associated token lists
- [ ]  Add token ignore modal that shows up the first time you uncheck/ignore a token
- [ ] Show token imported toast upon importing tokens from action sheet
- [ ] Will need to update the `AssetsContractController` to support switching between different networks
- [ ] Remove static token lists
- [ ] Update token detection logic to only control auto detection. This means that we will always be attempting to pull the dynamic token list regardless of toggle state.
- [ ] Add `Token Added` analytics event to importing tokens from action sheet
- [ ] Update `Tokens Added` analytics event to include `source` field

### **Acceptance Criteria**
- Auto token detection should work on supported networks
- Dynamic list should populate token list on search (supported networks listed above in description)
- Should have ability to ignore and/or import detected tokens on supported networks
- Upgrading app should display `Whats new` modal with token detection information
- Analytics: Reusing existing analytics event `Token Added` but should also include `source` field

### **References**
- Notion doc- https://www.notion.so/Improve-token-detection-v2-fb3ef7b7299c4ddb8e9481c2b22f1e87
- Deployed Addresses https://consensys.slack.com/archives/C021WBTTT63/p1642598375001100?thread_ts=1641949591.001800&cid=C021WBTTT63
Figma - https://www.figma.com/file/2h15YfvbT2q4q1FOMJFPP1/Token-Detection-V2?node-id=12%3A3663

### **TODOs**
- [ ] Need to deploy Avalanche contract + retrieve contract address
- [ ] Finalize token detection banner under add assets in Figma

omnat:
To add 
- Update copy to include Polygon & BSC ![image.png](https://images.zenhubusercontent.com/5c18e51bb423df7568ac3781/18faad02-accd-4583-8b74-4136534ec115)

[...]

PR #4388, opened 2022-05-24
[5.4] Feature/3157 token detection v2 merge

Cal-L:
**Description**

#3157

Video here - https://www.notion.so/5-4-0-8a73698dde22435ebf1cdbdcd934415f#9acb0b01e4624f54a634dbf1447e38e5

Original PR - #4096 

**Checklist**

**Screenshots/Recordings**

_If applicable, add screenshots or recordings to visualize the changes_

**Issue**

Progresses #???


gantunesr (approved):
LGTM

chrisleewilcox:
Adding remaining issues here.....

#1.  Token Import sheet does not scroll down (swipe up) 
- Wallet view > tap "{count} new tokens found in this account" with a count high enough to have tokens listed on the token sheet beyond what can fit on the token sheet for viewing.
- swipe up to scroll down

You will see that the token import sheet does not scroll down when you swipe up.  
https://recordit.co/fpw4EkS6vb

chrisleewilcox:
#2. PhishingController changes.

https://consensys.slack.com/archives/CBW7S9FSN/p1654528596008109?thread_ts=1654522908.344929&cid=CBW7S9FSN

chrisleewilcox:
#3. Avalanche network token search

https://consensys.slack.com/archives/C021WBTTT63/p1654632681012309

chrisleewilcox:
#3. Avalanche network token search

https://consensys.slack.com/archives/C021WBTTT63/p1654632681012309

chrisleewilcox:
#4. Token Import sheet buttons are doing more than expected, may cause confusion.

This may have been by design but it was confusing going through the account token detection flow.  Import button does more than just import tokens on the token sheet if not all tokens are selected to import.  Unselected tokens on the token sheet will also be hidden which I think will cause the confusion because there is also a hidden button that wasn't tapped.  I don't think the "Are you sure?" alert is sufficient in informing the user that unselected tokens will be hidden.

Scenario: Import multiple tokens detected but do not import all tokens detected
Given I am on Wallet view
And I have 7 tokens detected
When I tap on "7 new tokens found in this account"
And I select 4 tokens for import (or unselect 3 tokens)
And I tap "Import (4)" button
And I tap "Confirm" on alert
Then the 4 tokens selected are imported
And the 4 tokens selected are displayed on the wallet view token list
And the 4 tokens selected display the correct balance
And the 3 tokens not selected are not imported
And the 3 tokens selected are not hidden (FAIL: unselected tokens are hidden)
And "3 new tokens found in this account" link is displayed (FAIL: link is not displayed)

Cal-L:
> Adding remaining issues here.....
> 
> #1. Token Import sheet does not scroll down (swipe up) , cannot see all tokens detected
> 
> * Wallet view > tap "{count} new tokens found in this account" with a count high enough to have tokens listed on the token sheet beyond what can fit on the token sheet for viewing.
> * swipe up to scroll down
> 
> You will see that the token import sheet does not scroll down when you swipe up. https://recordit.co/fpw4EkS6vb

Pushed up fix

Cal-L:
> #3. Avalanche network token search.
> 
> https://consensys.slack.com/archives/C021WBTTT63/p1654632681012309

Discussed with Product - we will be keeping behavior as is for now 
https://consensys.slack.com/archives/G1L7H42BT/p1654619645424659?thread_ts=1650897225.469639&cid=G1L7H42BT

Cal-L:
> #4. Token Import sheet buttons are doing more than expected, may cause confusion.
> 
> This may have been by design but it was confusing going through the account token detection flow. Import button does more than just import tokens on the token sheet if not all tokens are selected to import. Unselected tokens on the token sheet will also be hidden which I think will cause the confusion because there is also a hidden button that wasn't tapped. I don't think the "Are you sure?" alert is sufficient in informing the user that unselected tokens will also be hidden.
> 
> Scenario: Import multiple tokens detected but do not import all tokens detected Given I am on Wallet view And I have 7 tokens detected When I tap on "7 new tokens found in this account" And I select 4 tokens for import (or unselect 3 tokens) And I tap "Import (4)" button And I tap "Confirm" on alert Then the 4 tokens selected are imported And the 4 tokens selected are displayed on the wallet view token list And the 4 tokens selected display the correct balance And the 3 tokens not selected are not imported And the 3 tokens selected are not hidden (FAIL: unselected tokens are hidden) And "3 new tokens found in this account" link is displayed (FAIL: link is not displayed)
> 
> 

This is currently working as intended, let's bring this up with Product and see what they think.

Cal-L:
> #2. PhishingController changes.
> 
> https://consensys.slack.com/archives/CBW7S9FSN/p1654528596008109?thread_ts=1654522908.344929&cid=CBW7S9FSN

Pending discussion with @sethkfman , will update as convo progresses.

chrisleewilcox:
#6. Not seeing any of the What's New content for TD.  I know there is another PR to deal with this but I want to have this as a placeholder for the What's New content.

chrisleewilcox:
#1. FIXED.

chrisleewilcox:
#3.  Avalanche import token search not showing any tokens except for WAVAX

I would like to discuss with product more on this issue. If we leave it this way then it is not aligned with Swaps token search which shows most of the supported tokens (and please do not fix Swaps token search to make it work like import token search).  Also, only WAVAX token will ever be detected with the current configuration so users receiving any other tokens on AVAX will not get detected, which makes Avalanche only capable or "1 new token found in this accoount".

Another issue is showing a warning that USDT is only on 1 confirmed list.  This can't be right and I think this is probably the root cause of understanding why leaving the current configuration as is will not convince users that Avalanche is supported by MMM.

bschorchit:
I think your comment is very on point.

The tricky part is we have very good reasons (mainly security) to add the 3 lists requirement.
As you mentioned, USDT currently only occurs in 1 list. Lowering the current requirement to meet this would allow any airdrop scam that manages to get its token added to a **single** list to be prompted and displayed to our users. Is this really something we want to risk?

I don't disagree at all of the importance of USDT and how inefficient this feature is for Avalanche at its current state. But I don't think the solution lies in us lowering our requirements, I'd rather see us engaging with the ecosystem and pushing for more updated lists for this network.

I really wonder why so many of those lists are failing to include USDT and other popular tokens.

That being said, I'm happy to be wrong here and hear suggestions that others might have.


Re: `Another issue is showing a warning that USDT is only on 1 confirmed list? Currently the #3 token in all of crypto behind BTC and ETH? This can't be right and I think this is probably the root cause of understanding why leaving the current configuration as is will not convince users that Avalanche is supported by MMM.`

Is this new behavior? I'd think swaps introduced it when they launched on Avalanche and this token detection update is not the cause of it.

chrisleewilcox:
I think the token list approach is probably not the appropriate logic or needs to be dug into more.  If USDT is not shown because our configurations due to security purposes then it doesn't seem we have the correct logic.  USDT should be shown even with this security concern. What are we not doing properly with our security settings in place but incorrectly filtering out USDT and all other tokens on Avalanche network?

chrisleewilcox:
#5. Styling issues with Keystone Hardware wallet Select an Account screen on Android device

Wallet view > tap Accounts icon > tap Connect Hardware wallet

prod:


branch:

omnat:
Very valid points @chrisleewilcox and speaks to the incompleteness of this feature. 

The solution is incomplete indeed if currently only 1 token passes the requirement of occurence on 3 lists. @bschorchit thoughts on removing support for Avalanche for now? We need to get this feature out and we can always iterate. I understand 3 lists was probably set as requirement for increased security, but I am with Chris, a network can't be considered "supported" from user point of view if only 1 token passes the requirement of 3 lists (also acknowledging that this can change as lists are dynamic in nature, and the token lists maintainers may update their lists for completeness). I am with you @bschorchit on not lowering the minimum list requirement for now. Let's be more cautious for user safety, and ship improvements in next iterations.

In future iterations, a more complete way to detect all tokens could be
1. [Make it easy to add a token after an action](https://www.notion.so/Suggest-adding-tokens-when-receiving-them-from-a-transaction-you-initiated-6098d4236ba84610875651026874d946).
2. Allow users to manage their lists - they could subscribe to a list they trust and adapt logic of relying on a single list.

@bschorchit @chrisleewilcox for this release, if we updated by removing Avalanche as "supported networks" from this feature launch, would this work from both your perspectives? I know it's not the best solution, but hopefully it prevents a bad UX where people expect more tokens to be detected on Avalanche, and still allows us to get this out the door which is a big improvement from what we currently have in production on other networks.

chrisleewilcox:
Thanks for your input @omnat .  I am more comfortable with removing saying we support for Avalanche at this point and keeping things as is for not until we hash out the details for supporting Avalanche network.  I don't think we even need to pull out any code for Avalanche network.  Just content saying we support Avalanche should be adjusted like in the What's New and other places until we work it out for Avalanche.

omnat:
Trying to balance

- not shipping bad UX
- not compromising on the user safety requirement of 3+ lists
- not delaying this feature release, so we can ship some value

omnat:
Sounds good @chrisleewilcox !

Let's see what @Cal-L and @bschorchit think about this path forward.

bschorchit:
This sounds good to me! I think we have 2 different proposals here:

1. Omna’s: turning token detection off for Avalanche
2. Chris’: keeping turning detection on for Avalanche, but not communicating it as a supported network

Which one are we moving forward with?

Cal-L:
> #5. Styling issues with Keystone Hardware wallet Select an Account screen on Android device
> 
> Wallet view > tap Accounts icon > tap Connect Hardware wallet
> 
> prod: 
> 
> branch: 

Change comes from main and was updated here - https://github.com/MetaMask/metamask-mobile/commit/66ed5697e61053448ed9872575ac7c0cbf35fe46

chrisleewilcox:
#5. Not issue, as designed.

chrisleewilcox:
#2.  Browser is now working again.

bschorchit:
Re [3](https://github.com/MetaMask/metamask-mobile/pull/4388#issuecomment-1150491893): Given that @chrisleewilcox reached out to the codefi team and they added more token-metadata providers for avalanche and [@Cal-L confirmed it's now working as expected](https://consensys.slack.com/archives/C021WBTTT63/p1655158999029859), if no one opposes to it, we should continue to treat Avalanche as a supported network for token detection.
cc: @omnat

gantunesr (approved):
LGTM

chrisleewilcox (approved):
LGTM!

PR #4394, opened 2022-05-25
Remove subtitles from translation files

rickycodes:
**Description**

Right now we have an issue with the subtitles. the translators keep trying to translate those keys:

image

This also caused an issue in regression and was patched in: https://github.com/MetaMask/metamask-mobile/pull/4349

This is because we're using the translations files for logic. Instead, I've moved the logic away from the JSON files and into the code.


app/util/video/index.ts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export function getSubtitleUri(language: string): string {
+  // eslint-disable-next-line no-prototype-builtins
+  if (subtitleMap.hasOwnProperty(language)) {
+    return `${path}${subtitleMap[language]}${ext}`;
+  }
+  // return english by default
+  return `${path}en${ext}`;
+}
```
rickycodes, line 26: this will all go away once we merge in recent translations where we've included these subtitle files. at that point we won't need a map and we can just load the file(s)

sethkfman (approved):
LGTM

PR #4418, opened 2022-05-30
[IMPROVEMENT] Address Verification

gantunesr:
**Description**

Improve address verification in SendFlow,

1. Allow user to input an address with all lower or upper case characters.
2. If the user inputs a mixed case address, it should be the correct checksum.  Else, the address should be flagged as invalid and not auto-checksum the address.
  - Example of incorrect checksum: `0x87187657b35f461d0ceEc338d9B8e944A193afe2`
  - Example of correct checksum: `0x87187657b35F461D0Ceec338d9b8E944a193aFE2`

**Checklist**

**Screenshots/Recordings**

| Bug | Fix |
|-------|-------------|
|||

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/288


Fatxx (changes requested):
Should we also add the same validation in `Views/Send/index.js` screen?

gantunesr:
@Fatxx I don't see any validation logic in `Views/Send/index.js` and we should stop using it in favor of `SendFlow`

Fatxx (approved):

cortisiko (approved):
I tested these Incorrect check sum addressed:

`0xa54D3c09E34aC96807c1CC397404bF2B98DC4eFb`
`0x87187657b35f461d0ceEc338d9B8e944A193afe2`

on the send flow and also on the add contact view. I also generated QR codes for the above addresses and tested with them. See QR codes [here](https://www.screencast.com/t/cdvWwa9f8) and [here](https://www.screencast.com/t/dxjZXdVNsoa8). I tested upper cased addresses and lower cased addresses. 


I tested scanning a keystone address QR code and I was able to successfully send to that address. 


Finally I did sanity checks on:
scanning a request QR code. The correct address received the funds
Scanning an extension QR code for a couple of my accounts. I had no issues 

QA passed! 🌮 🌮

PR #4421, opened 2022-05-30
Vault corruption recovery flow

### LINKED ISSUE #4645: Storage limit/corrupt data - reinstall mm
**Describe the bug**
Every few weeks I have to delete and reinstall mm on iPhone due to this corrupt local data that seems to occur after so much mm usage. This issue also seems to persist across other users as I hear about it in my chats sometimes. 

**Screenshots**
Attached
CD52B88C-D726-4137-BAF3-3CA330ECF027

**To Reproduce**
1. it occurs a few weeks of steady use after fresh install 

**Expected behavior**
_A clear and concise description of what you expected to happen_

 - Device: iPhone 11 Pro
 - OS: iOS 15.5
 - App Version [e.g. 1.0.0] - v5.3.0 (927)

**Severity**

- How critical is the impact of this bug on a user? It’s is a critical storage failure 
- Is this visible to all users? On multiple mobile users I have encountered 
- Is this tech debt?

owencraston:
**Description**

- We are building this feature off the work done in the [auth refactor](https://github.com/MetaMask/metamask-mobile/pull/5374)

### Why?
- Every so often the vault gets corrupted due to storage issues with the app. When this happens the users get locked out of MetaMask and in some cases can lose their wallet completely if they have no backed up their SRP. 
- We have tracked this error for a few months and it has occurred `67 000` times and has effected `over 2000 users`
    - [see sentry dashboard with this error](https://sentry.io/organizations/metamask/issues/3648079321/?query=is%3Aunresolved++Vault+Corruption+Error&referrer=issue-stream&statsPeriod=14d)
     Screenshot 2022-12-20 at 4 31 30 PM
- This NEEDS to be addressed. Apart from removing this issue altogether (not something we know how to do yet and will involve a massive refactor) this recovery flow is the best we can do now.

### What?
- listens for changes in the keyring controller and makes a backup stores in the [SecureKeychain](https://github.com/oblador/react-native-keychain)
- Adds the ability to reinitialize the engine with an existing keyring. Then engine will initialize the keyring controller with the backup if it is passed in, else it will create a new one. 
- When vault corruption error surfaces in [Login](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/Login/index.js#L426), we kick them into a vault recovery flow that recreates the engine and and authenticates them back into the app (or send them to login if they used password as a login method). 

### How?
This is a fairly complex fix so I will try to highlight how this is working....
- create a method to backup the vault.  see `app/core/backupVault.ts` method `backupVault`
- [listen](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-05c078ffc07c5a566f99774953d8ed1a8ff4e0cb3d0836cb27279bf559134acbR450) for changes on the KeyringController in the Engine and backup the vault when there is a change. See `app/core/Engine.js` `handleVaultBackup`
- modify the `Engine.init` to accept an [optional param](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-05c078ffc07c5a566f99774953d8ed1a8ff4e0cb3d0836cb27279bf559134acbR919) of the keyring state. 
- when we pass in a keyringState to the init call of the engine, [recreate the keyring controller with the backup](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-05c078ffc07c5a566f99774953d8ed1a8ff4e0cb3d0836cb27279bf559134acbR240).
- When the vault corruption error occurs in `Login/index.js`, start the recovery flow by calling [handleVaultCorruption](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-86a4727b0ac06050024c30141979a466b490a3bd8672dbd790643dd1be3f2bdcR404) which ...
    - gets the vault from backaup via [getVaultFromBackup()](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-86a4727b0ac06050024c30141979a466b490a3bd8672dbd790643dd1be3f2bdcR296)
    - verifies that the password the user typed into the password field can unlock the vault with [parseVaultValue](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-86a4727b0ac06050024c30141979a466b490a3bd8672dbd790643dd1be3f2bdcR300)
    - if they can, it is the correct password then we can get the [auth type and store the credentials](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-86a4727b0ac06050024c30141979a466b490a3bd8672dbd790643dd1be3f2bdcR306-R314)
        - this is needed in case they change the auth type from the previously used type (biometrics, remember me, password etc)
        - if the password storage is successful we will launch the recovery flow since we know we have valid credentials that can unlock the backup of the vault and we have the latest auth type. 
    - if the password is wrong we indicate this to the user
- The first screen in the recovery flow is `RestoreWallet`. This screen calls [EngineService.initializeVaultFromBackup()](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-fd04d31e4019fc6efd5c8a4b02f7404d583ae2c1b156f9990a9b195eaa7c5523R36).
    - [initializeVaultFromBackup](https://github.com/MetaMask/metamask-mobile/pull/4421/files#diff-03f8a503bb05c58fc3437bdd3f691809331949656c8b6c537c03646507b79ef6R97) gets the backup, destroys the old engine (not just resets it, removes the old instance), and then calls Engine.init with the backed up keyring state as the keyringState param in `Engine.init(state, keyringState)`
    - then we setup the controllers listeners on the new engine instance with `this.updateControllers` 
- On success we will navigate to the `WalletRestored` screen. Here we will attempt a login (using the previously stored credentials) by calling ```await Authentication.appTriggeredAuth(selectedAddress);
      navigation.navigate(Routes.ONBOARDING.HOME_NAV);```
      - if we don't have the credentials stored (the case for password login) we will simply navigate the user to Login where they can login normally. the vault/engine have been reinitialized at this point. 

### Acceptance criteria
- [ ] strings are translated
- [ ]  whenever a user experiences the Vault corruption error on Login and they have a valid password, they are successfully launched into the vault recovery flow
    - If the user has biometrics enabled, it will prompt them for biometrics before going into the recovery flow
- [ ] if the user has an incorrect password but tried to recover the vault they are informed that their password is incorrect
- [ ] When the user clicks `Restore wallet` the old engine is destroyed, a new engine is reinitialized with the vault backup and they are navigated to the `WalletRestored` screen.
- [ ] On `WalletRestored` there are three potential outcomes to pressing the `Continue to wallet` button...
    1. the user is using password and `goes back to login`
    2. the user is using biometrics/passcode and `we get the biometrics on the wallet restored success screen (faceID or passcode or fingerprint) and on success we log you in and you are navigated to the wallet view`
    3. the user is using remember me and `the user gets logged in automatically and is navigated to the wallet view`
 - [ ] // TODO figure out the acceptance criteria for `WalletResetNeeded` screen. This screen only appears if the engine does not successfully get reinitialized in `RestoreWallet`

## TODO
- [x] Metrics
- [x] Finalize strings and get them merged into main for translations so we can launch with translations
    - strings were merged in [this PR](https://github.com/MetaMask/metamask-mobile/pull/5465)
- [x] Finalize intended outcomes on `WalletResetNeeded` screen. 
- [x] Investigate waiting until the Engine is initialized and  controller subscription are completely done before navigating to success screen.
- [x] Write unit tests
    - [x] verify that the old engine instance does not match the new engine instance when calling `initializeVaultFromBackup`

**Checklist**

**Designs**

- [Designs](https://www.figma.com/file/S23y6s0krwA0eP34eydG1A/Mobile---Vault-Recovery?node-id=6%3A1953)

**Screenshots/Recordings**
_the version here has some minor UX issues around the loading state of the WalletRestored screen_

https://user-images.githubusercontent.com/22918444/208782013-ba73d217-ed95-489f-9a1b-9531faa2b0b4.mov

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/139
progresses https://github.com/MetaMask/metamask-mobile/issues/5420
Resolves https://github.com/MetaMask/metamask-mobile/issues/4645


app/store/index.js
```diff
@@ -13,8 +13,9 @@ import { migrations, version } from './migrations';
 import Logger from '../util/Logger';
 import EngineService from '../core/EngineService';
 import Device from '../util/device';
+import { getVaultFromBackup } from '../core/backupVault';
 
-const TIMEOUT = 40000;
+const TIMEOUT = 1;
```
owencraston, line 18: This is for testing purposes

app/components/Views/Login/index.js
```diff
@@ -285,6 +288,61 @@ class Login extends PureComponent {
...
+            );
+            navigation.navigate(...createRestoreWalletNavDetails());
+          } catch (e) {
+            Logger.log(
+              'Login handleVaultCorruption',
+              'Login Authentication.storePassword failed with the following error',
+              e,
+            );
+          }
+        } else {
```
sethkfman, line 323: Any thoughts on how to remove these repeated setState? Maybe return before the catch and remove the final else but keep the setState.

app/components/Views/RestoreWallet/WalletRestored.tsx
```diff
@@ -0,0 +1,103 @@
...
+    try {
+      await Authentication.appTriggeredAuth(selectedAddress);
+      navigation.navigate(Routes.ONBOARDING.HOME_NAV);
+    } catch (e) {
+      // we were not able to log in automatically so we will go back to login
+      navigation.navigate(Routes.ONBOARDING.LOGIN);
+    }
+  }, [navigation, selectedAddress]);
+
+  const backupSRPURL =
```
sethkfman, line 48: Move to URL constants

app/core/backupVault.ts
```diff
@@ -0,0 +1,70 @@
...
+  accessible: ACCESSIBLE.WHEN_UNLOCKED,
+};
+
+interface KeyRingBackupResponse {
+  success: boolean;
+  message: string;
+  state?: KeyringState;
+}
+
+export const backupVault = async (keyringState: KeyringState) => {
```
sethkfman, line 23: Can you add comments to each of these functions? It would be good to understand how they relate the keychain integration

app/store/index.js
```diff
@@ -47,6 +47,7 @@ const MigratedStorage = {
     } catch (error) {
       Logger.error(error, { message: 'Failed to set item' });
     }
+    // throw new Error('Failed async storage storage fetch.');
```
sethkfman, line 50: Remove if not needed

app/components/Views/RestoreWallet/WalletRestored.tsx
owencraston: This is temporary. I am still working with @jakehaugen on the intended behaviour. He wants to navigate to the BackupSRP page which is only possible if we have a successful login.

app/components/Views/RestoreWallet/WalletResetNeeded.tsx
```diff
@@ -0,0 +1,121 @@
...
+      // this does not work and we need to figure out what to do here
+      // navigation.navigate(Routes.ONBOARDING.IMPORT_FROM_SECRET_RECOVERY_PHRASE);
+    } catch (error) {
+      // TODO how should we handle this case?
+      setCreateNewWalletLoading(false);
+      console.log(
+        'WalletResetNeeded',
+        'there was an issue resetting the wallet',
+        error,
+      );
```
gantunesr, line 67: Remove before merge

app/components/Views/RestoreWallet/WalletResetNeeded.tsx
```diff
@@ -0,0 +1,121 @@
...
```
gantunesr, line 50: Remove before merge

app/components/Views/RestoreWallet/WalletResetNeeded.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const handleCreateNewWallet = useCallback(async () => {
+    setCreateNewWalletLoading(true);
+    try {
+      await resetWalletState();
+      // this does not work and we need to figure out what to do here
+      // navigation.navigate(Routes.ONBOARDING.IMPORT_FROM_SECRET_RECOVERY_PHRASE);
+    } catch (error) {
+      // TODO how should we handle this case?
+      setCreateNewWalletLoading(false);
```
gantunesr, line 45: Can you explain the situation here, I want to understand better the problem that you're pointing out in the comments

app/components/Views/Login/index.js
```diff
@@ -285,6 +288,61 @@ class Login extends PureComponent {
     return false;
   };
 
+  handleVaultCorruption = async (error) => {
+    const { navigation } = this.props;
+    console.log('vault/ vault error thrown: ', error);
```
gantunesr, line 293: Remove before merge

app/components/Views/Login/index.js
```diff
@@ -285,6 +288,61 @@ class Login extends PureComponent {
...
+        const vaultSeed = await parseVaultValue(
+          this.state.password,
+          keyringState.vault.toString(),
+        );
+        if (vaultSeed) {
+          // get authType
+          const authData = await Authentication.componentAuthenticationType(
+            this.state.biometryChoice,
+            this.state.rememberMe,
+          );
```
gantunesr, line 309: Why is it necessary to get the mnemonic here?

gantunesr, line 309: Would the method `KeyringController.validatePassword` also work here? That way you can validate the password without the need to decrypt the mnemonic

app/components/Views/Login/index.js
owencraston: we could create a passwordCheck function and abstract it in there?

app/components/Views/RestoreWallet/WalletResetNeeded.tsx
owencraston: We are still deciding what the best options are here and if we should have more than these two options...maybe a try again button? see [discussion on slack](https://consensys.slack.com/archives/C03BAF1DXCL/p1671567198279629) for more context.

app/components/Views/Login/index.js
owencraston: we could maybe use this in place of the parsedVault function call here. The only snag is that will be using the old KeyringController that is corrupt since the engine has not been reiniatliazed yet. The benefit of parseVaultValue is that it lets us pass in our backup as a param so we can validate the backup.

owencraston: the reason we are calling `componentAuthenticationType` ( and why I don't think we can replace it with `KeyringController.validatePassword`) is because its possible for a user to change their auth type before they enter the vault corruption flow. `componentAuthenticationType` will return to us what the desired auth type is based on priority and the given state values and then we can persist it in the call to storePassword below.

app/components/Views/RestoreWallet/WalletRestored.tsx
owencraston: After some investigation I realized it is impossible to navigate to the backup flow from here since the user is logged out. I think there is value in having some call to action here. Is there a better resource that I can link to? cc @sethkfman @gantunesr @AlexJupiter

owencraston: also should we have an analytics event for this?

app/components/Views/RestoreWallet/WalletResetNeeded.tsx
```diff
@@ -0,0 +1,110 @@
...
+      deviceMetaData,
+    );
+    navigation.navigate(Routes.MODAL.ROOT_MODAL_FLOW, {
+      screen: Routes.MODAL.DELETE_WALLET,
+    });
+  }, [deviceMetaData, navigation]);
+
+  const handleTryAgain = useCallback(async () => {
+    AnalyticsV2.trackEvent(
+      MetaMetricsEvents.VAULT_CORRUPTION_WALLET_RESET_NEEDED_TRY_AGAIN_BUTTON_PRESSED,
```
owencraston, line 55: @gantunesr should we add some way to determine if the user is going back to the restore screen from the try again button? that way we can tell if this try again button does anything at all. If so do you have suggestions on how to implement this?

gantunesr: I think with this events it should be enough, from the logic I can see that the user will navigate to the `VAULT_RECOVERY.RESTORE_WALLET`, and for this case you can also take a particular user and see an event pattern of this type,

```
Vault Corruption Wallet Reset Needed Try Again Button Pressed
Vault Corruption Restore Wallet Screen Viewed
Vault Corruption Restore Wallet Button Pressed
```

From that you can tell that the user previously pressed `Try Again` and ended up in the Restore View (multiple times if you see this pattern repeated). But if you want to improve it, I think the solution would be to persist an flag, set it to true when `Try Again` is pressed for the first time, and then use it as an argument in the event `Vault Corruption Restore Wallet Screen Viewed`. An alternative would be a counter for the times `Try Again` is pressed.

app/components/Views/RestoreWallet/WalletRestored.tsx
```diff
@@ -0,0 +1,121 @@
...
+  ScrollView,
+  Text as RNText,
+  Linking,
+} from 'react-native';
+import { strings } from '../../../../locales/i18n';
+import { createStyles } from './styles';
+import Text, {
+  TextVariants,
+} from '../../../component-library/components/Texts/Text';
+import StyledButton from '../../UI/StyledButton';
```
gantunesr, line 15: Same here,

> _Any reason on using StyledButton over the DS Button?_

app/components/Views/RestoreWallet/WalletRestored.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const deviceMetaData = useMemo(() => generateDeviceAnalyticsMetaData(), []);
+
+  useEffect(() => {
+    AnalyticsV2.trackEvent(
+      MetaMetricsEvents.VAULT_CORRUPTION_WALLET_SUCCESSFULLY_RESTORED_SCREEN_VIEWED,
+      deviceMetaData,
+    );
+  }, [deviceMetaData]);
+
+  const finishWalletRestore = useCallback(async () => {
```
gantunesr, line 50: nit
```suggestion
  const finishWalletRestore = useCallback(async (): Promise<void> => {
```

app/components/Views/RestoreWallet/WalletRestored.tsx
gantunesr: I think you can use the `SRP_GUIDE_URL` in the `constants/urls.ts` file. And I don't think an event here will add value.

app/components/Views/RestoreWallet/WalletRestored.tsx
```diff
@@ -0,0 +1,121 @@
...
+    } catch (e) {
+      // we were not able to log in automatically so we will go back to login
+      navigation.navigate(Routes.ONBOARDING.LOGIN);
+    }
+  }, [navigation, selectedAddress]);
+
+  const backupSRPURL =
+    'https://metamask.zendesk.com/hc/en-us/articles/360060826432-What-is-a-Secret-Recovery-Phrase-and-how-to-keep-your-crypto-wallet-secure';
+
+  const onPressBackupSRP = useCallback(async () => {
```
gantunesr, line 63: nit
```suggestion
  const onPressBackupSRP = useCallback(async (): void => {
```

app/components/Views/RestoreWallet/WalletRestored.tsx
```diff
@@ -0,0 +1,121 @@
...
+    Linking.openURL(backupSRPURL);
+  }, []);
+
+  const handleOnNext = useCallback(async () => {
```
gantunesr, line 67: nit
```suggestion
  const handleOnNext = useCallback(async (): void => {
```

app/core/EngineService.ts
```diff
@@ -1,9 +1,19 @@
...
 const UPDATE_BG_STATE_KEY = 'UPDATE_BG_STATE';
 const INIT_BG_STATE_KEY = 'INIT_BG_STATE';
 
+interface InitializeEngineResult {
+  success: boolean;
+  error?: string;
+}
+
+const VAULT_CREATION_ERROR = 'Error creating the vault';
+const NO_VAULT_IN_BACKUP_ERROR = 'No vault in backup';
```
gantunesr, line 15: Move to error constants file (`constants/error.ts`)

app/core/EngineService.ts
```diff
@@ -68,15 +82,47 @@ class EngineService {
...
           AppConstants.NETWORK_STATE_CHANGE_EVENT,
           update_bg_state_cb,
         );
     });
   };
+
+  /**
+   * Initialize the engine with a backup vault from the Secure KeyChain
+   *
+   */
```
gantunesr, line 98: Specify the return object in the commentary

gantunesr: For example,

```ts
  /**
   * METHOD DESCRIPTION
   *
   * @returns RETURN OBJECT DESCRIPTION
   */
```

app/core/backupVault.ts
```diff
@@ -0,0 +1,77 @@
+import { KeyringState } from '@metamask/keyring-controller';
+import Logger from '../util/Logger';
+import {
+  getInternetCredentials,
+  setInternetCredentials,
+  resetInternetCredentials,
+  Options,
+  ACCESSIBLE,
+} from 'react-native-keychain';
+
```
gantunesr, line 10: Add proper commentary to describe the methods in this file

app/core/backupVault.ts
```diff
@@ -0,0 +1,77 @@
...
+export async function getVaultFromBackup(): Promise<FetchKeyringFromBackupResponse> {
+  Logger.log(VAULT_BACKUP_KEY, 'getVaultFromBackup');
+  const credentials = await getInternetCredentials(VAULT_BACKUP_KEY);
+  if (credentials) {
+    return { success: true, vault: credentials.password };
+  }
+  return { success: false };
+}
+
+export const resetVaultBackup = async () => {
```
gantunesr, line 74: nit
```suggestion
export const resetVaultBackup = async (): Promise<void> => {
```

app/core/backupVault.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+    if (backupResult === false) {
+      Logger.log(VAULT_BACKUP_KEY, 'Vault backup failed');
+      const response: KeyringBackupResponse = {
+        success: false,
+        message: 'Vault backup failed',
+      };
+      return response;
+    }
+    Logger.log(VAULT_BACKUP_KEY, 'Vault successfully backed up');
```
gantunesr, line 46: We could remove this log since we're expecting this to be the happy path

app/core/backupVault.ts
```diff
@@ -0,0 +1,77 @@
...
+  Logger.log(VAULT_BACKUP_KEY, 'Unable to backup vault as it is undefined');
+  const response: KeyringBackupResponse = {
+    success: false,
+    message: 'Unable to backup vault as it is undefined',
+  };
+  return response;
+}
+
+export async function getVaultFromBackup(): Promise<FetchKeyringFromBackupResponse> {
+  Logger.log(VAULT_BACKUP_KEY, 'getVaultFromBackup');
```
gantunesr, line 66: We can also remove this one and add it before the last return to indicate that there was an error

app/core/backupVault.ts
```diff
@@ -0,0 +1,77 @@
...
+  const credentials = await getInternetCredentials(VAULT_BACKUP_KEY);
+  if (credentials) {
+    return { success: true, vault: credentials.password };
+  }
+  return { success: false };
+}
+
+export const resetVaultBackup = async () => {
+  Logger.log(VAULT_BACKUP_KEY, 'resetting vault backup');
```
gantunesr, line 75: We can also remove this one too

gantunesr (commented):
Left a lot of minor comments, the logic made sense according to previous discussion about this solution. Some observations,

- The file `app/core/backupVault.ts` is not tested, this is not a blocker for this PR and lets open a ticket to address it during this Q
- Feel free to add more comments explaining the logic and reasoning behind some methods, in a year from here other devs might wonder why it was necessary

app/components/Views/RestoreWallet/RestoreWallet.tsx
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable import/no-commonjs */
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { View, Image, ActivityIndicator } from 'react-native';
+import { strings } from '../../../../locales/i18n';
+import { createStyles } from './styles';
+import Text, {
+  TextVariants,
+} from '../../../component-library/components/Texts/Text';
+import StyledButton from '../../UI/StyledButton';
```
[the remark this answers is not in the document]
owencraston: The reason for this is that the DS Button (ButtonPrimary) does not take children or support a loading state. It can only render a label which comes in as string. The styled button allows me to conditionally render a loading indicator.

app/components/Views/RestoreWallet/WalletRestored.tsx
owencraston: ds does not support loading state. More into in the above comments.

app/components/Views/Login/index.js
```diff
@@ -298,6 +301,55 @@ class Login extends PureComponent {
...
+          backupResult.vault,
+        );
+        if (vaultSeed) {
+          // get authType
+          const authData = await Authentication.componentAuthenticationType(
+            this.state.biometryChoice,
+            this.state.rememberMe,
+          );
+          try {
+            await Authentication.storePassword(
```
gantunesr, line 322: Break out this nested try-catch

app/constants/error.ts
```diff
@@ -37,3 +37,14 @@ export const AUTHENTICATION_STORE_PASSWORD_FAILED =
 
 export const AUTHENTICATION_LOGIN_VAULT_CREATION_FAILED =
   'Authentication.loginVaultCreation was unable to recreate vault';
+
+// EngineService
+export const VAULT_CREATION_ERROR = 'Error creating the vault';
+export const NO_VAULT_IN_BACKUP_ERROR = 'No vault in backup';
+
+//backupVault
```
gantunesr, line 45: nit

```suggestion
// backupVault
```

app/core/backupVault.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * places the vault in react-native-keychain for backup
+ * @returns Promise<KeyringBackupResponse>
+  interface KeyringBackupResponse {
+    success: boolean;
+    error?: string;
+    vault?: string;
+  }
+ */
```
gantunesr, line 36: The return type is already specified in the code, you can just explain what it's returning on simple terms

app/core/backupVault.ts
```diff
@@ -0,0 +1,93 @@
...
+  return response;
+}
+
+/**
+ * retrieves the vault backup from react-native-keychain
+ * @returns Promise<KeyringBackupResponse>
+  interface KeyringBackupResponse {
+    success: boolean;
+    error?: string;
+    vault?: string;
```
gantunesr, line 75: Same here,

> _The return type is already specified in the code, you can just explain what it's returning on simple terms_

app/core/backupVault.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (credentials) {
+    return { success: true, vault: credentials.password };
+  }
+  Logger.error(VAULT_BACKUP_KEY, VAULT_FAILED_TO_GET_VAULT_FROM_BACKUP);
+  return { success: false, error: VAULT_FAILED_TO_GET_VAULT_FROM_BACKUP };
+}
+
+/**
+ * removes the vault backup from react-native-keychain
+ * @returns Promise<Void>
```
gantunesr, line 89: No need here since you're not returning a value inside the `Promise`

```suggestion
```

app/core/EngineService.ts
```diff
@@ -72,15 +86,52 @@ class EngineService {
...
   };
+
+  /**
+   * Initialize the engine with a backup vault from the Secure KeyChain
+   *
+   * @returns Promise<InitializeEngineResult>
+   *  InitializeEngineResult {
+        success: boolean;
+        error?: string;
+      }
```
gantunesr, line 106: Same here,

> _The return type is already specified in the code, you can just explain what it's returning on simple terms_

gantunesr (changes requested):
An observation,

- I'm pointing out again that `app/core/backupVault.ts` is un-tested, that's also the case for the `EngineService`. Maybe I missed this in the previous PR, but if you're going to modify this core files, it's always a good practice to add testing.

app/core/backupVault.ts
owencraston: I noticed that when I removed the interface the intenseness was not as descriptive which is why I added it. 

**With** the interface commented
Screenshot 2023-02-27 at 4 05 45 PM

**Without** the interface commented
Screenshot 2023-02-27 at 4 04 05 PM

app/core/EngineService.ts
owencraston: By adding this description we get better inline code documentation. This is from the file that is calling this method and we can see the exact response object.
Screenshot 2023-02-27 at 10 33 24 PM

owencraston:
@gantunesr I gave it a very good try on the unit testing but  I was unable to make any progress. I documented the need for testing in [this issue ](https://github.com/MetaMask/metamask-extension/issues/17938)as well as some of the issues I encountered. I hope that we can move this into QA without doing the tests today and that we can come back to this in the future.

gantunesr (changes requested):
LGTM. I'm flagging this PR as `request changes` for the piece of UI used to develop and test.

app/components/Views/Login/index.js
```diff
@@ -352,9 +407,10 @@ class Login extends PureComponent {
           strings('login.security_alert_desc'),
         );
         this.setState({ loading: false });
-      } else if (toLowerCaseEquals(error, VAULT_ERROR)) {
+      } else if (containsErrorMessage(error, VAULT_ERROR)) {
```
owencraston, line 410: I modified this from a strict equal to contains for two reasons...
1. Since the [Authentication Refactor](https://github.com/MetaMask/metamask-mobile/pull/5374) was merged our errors have become more nested. This shouldn't be a problem since I made sure to throw the original errors in the Authentication class but this adds an extra layer of protection in case things change within the Authentication logic in the future.
2. In order to help future proof this catch in the case that the authentication logic becomes more nested than it already is. A strict equal would fail if the error message was `'Error: Error: Cannot unlock without a previous vault.'` but this contains function would not. This extra error message is happens all the time when functions eagerly throw an error from another method.

app/components/Views/Login/index.js
```diff
@@ -352,13 +417,95 @@ class Login extends PureComponent {
           strings('login.security_alert_desc'),
         );
         this.setState({ loading: false });
-      } else if (toLowerCaseEquals(error, VAULT_ERROR)) {
-        const vaultCorruptionError = new Error('Vault Corruption Error');
```
owencraston, line 356: Moved this log into the `handleVaultCorruption` function.

app/components/Views/Login/index.js
```diff
@@ -352,13 +417,95 @@ class Login extends PureComponent {
...
-        Logger.error(vaultCorruptionError, strings('login.clean_vault_error'));
+      } else if (containsErrorMessage(error, VAULT_ERROR)) {
+        try {
+          await this.handleVaultCorruption();
+        } catch (e) {
```
owencraston, line 423: We also want to log the error and display the vault corruption error in the case that the our `handleVaultCorruption` has an uncaught error.

app/components/Views/Login/index.js
```diff
@@ -352,13 +417,95 @@ class Login extends PureComponent {
...
+    }
+  };
+
+  /* This method is identical to the one above, except for the call to userEntryAuth
+  This is a temporary test function to mimic the vault being corrupted
+  we will keep almost all the logic the same as onLogin except we call the purposely broken NEVER_MERGE_BROKEN_userEntryAuth method
+  this will throw the VAULT_ERROR and we can test the vault corruption flow
+  since the userEntryAuth is the one that will throw the VAULT_ERROR, this is the most realistic way to test the vault corruption flow
+*/
+  NEVER_MERGE_tempBreakTheVault = async () => {
```
owencraston, line 447: this needs to be deleted before merging, this method is exactly the same as the onLogin function with the exception that it makes a call to `await Authentication.NEVER_MERGE_BROKEN_userEntryAuth();` which eagerly throws an error. This is because the vault corruption error will come from the auth class so to best simulate how this recovery flow will work in prod I opted for this test method.

app/components/Views/Login/index.js
```diff
@@ -352,13 +417,95 @@ class Login extends PureComponent {
...
+        trackErrorAsAnalytics('Login: Invalid Password', error);
+
+        return;
+      } else if (error === PASSCODE_NOT_SET_ERROR) {
+        Alert.alert(
+          strings('login.security_alert_title'),
+          strings('login.security_alert_desc'),
+        );
+        this.setState({ loading: false });
+      } else if (containsErrorMessage(error, VAULT_ERROR)) {
```
owencraston, line 498: This block needs to match 1/1 the catch block in onLogin before we merge.

app/core/Authentication/Authentication.ts
```diff
@@ -407,6 +407,23 @@ class AuthenticationService {
     password = this.wipeSensitiveData();
   };
 
+  // This is purely to test the vault corruption flow and will need to be removed before merging
+  NEVER_MERGE_BROKEN_userEntryAuth = async (): Promise<void> => {
```
owencraston, line 411: Needs to be removed before merging

app/core/Authentication/AuthenticationError.ts
```diff
@@ -3,8 +3,8 @@ import { AuthData } from './Authentication';
 class AuthenticationError extends Error {
   public authData: AuthData;
   public customErrorMessage: string;
-  constructor(msg: string, customErrorMessage: string, authData: AuthData) {
-    super(msg);
+  constructor(message: string, customErrorMessage: string, authData: AuthData) {
```
owencraston, line 6: This is small and isn't 100% needed however in order to match the properties that exist in the standard Error class I changed this to `message`

olenapankina (changes requested):
**Tested on**
iPhone Xr iOS 16.1.1
Samsung Galaxy M21 Android 12, One UI Core version 4.1

**Issue_01:** User can swipe through vault recovery screens
- Restore needed-> Continue to wallet [4421_Issue_01_success flow.mov](https://user-images.githubusercontent.com/116747263/223225372-73843ed6-df58-4213-8ec6-31e7c1df89c5.mov)
- Restore needed-> New wallet needed [4421_Issue_01_failed flow.mp4](https://user-images.githubusercontent.com/116747263/223225544-cc88f7c1-375a-4668-8b9e-fc3db37cd606.mp4)

iOS - use gestures. Android - use native back button
**Expected result:** user should not be able to navigate between vault recovery screens. Android: user should not be able to navigate to the lock page using back button from vault recovery flow

**Issue_02:** password field is not cleared after restoring wallet
**Steps to reproduce:**
1. Set up password as auth
2. Force quit and relaunch the application
3. Enter password - click dev button
4. Finish success flow, enter the wallet
5. Put application in the background, wait for auto-lock time to pass (tested with 30s and immediate lock)
6. Open MetaMask
7. Observe password present in password entry field
8. Tap login button

**Actual result:** password present is password entry field, user is able to login with it [4421_Issue_02.mov](https://user-images.githubusercontent.com/116747263/223225687-195ff794-d91d-4546-b0d8-ae68eef8c9e2.mov)
**Expected result:** password should not be present in password entry after user successfully logged in
**Not able to reproduce on:** prod, auth refactor builds, after ordinary flow without vault recovery on vault recovery builds

**Issue_03:** User ends up on Lock page after tapping on 'Create a new wallet' at 'New wallet needed' screen
**Steps to reproduce:**
1. Launch the app
2. Enter vault recovery failed flow
3. At 'New wallet needed' screen tap at 'Create a new wallet' button
4. Tap 'Cancel' or observe screen in the background

**Actual result:** user ends up on Lock screen [4421_Issue_03.mp4](https://user-images.githubusercontent.com/116747263/223225821-f0d7f096-36a4-45ad-8dbf-82af997c7fb7.mp4)
**Expected result:** TBD, but I would expect user to stay in 'New wallet needed' screen

**Issue_04:** Android: New wallet needed flow: lock screen flashes in the background [4421_Issue_04.mov](https://user-images.githubusercontent.com/116747263/223225888-c146b3c0-eb92-43dc-a37f-a6a3b1e11e1c.mov)

app/components/Views/Login/index.js
owencraston: Before I c lean this up it matched 100%
Screenshot 2023-03-07 at 8 56 24 AM

app/component-library/components/Icons/Icon/Icon.types.ts
```diff
@@ -28,6 +28,8 @@ export enum IconSize {
   Md = '20',
   Lg = '24',
   Xl = '32',
+  // TODO finalize Icon size in Design system
```
owencraston, line 31: I did this based on @brianacnguyen instructions.

app/components/Views/Login/index.js
```diff
@@ -200,7 +205,7 @@ const PASSCODE_NOT_SET_ERROR = 'Error: Passcode not set.';
 const WRONG_PASSWORD_ERROR = 'Error: Decrypt failed';
 const WRONG_PASSWORD_ERROR_ANDROID =
   'Error: error:1e000065:Cipher functions:OPENSSL_internal:BAD_DECRYPT';
-const VAULT_ERROR = 'Error: Cannot unlock without a previous vault.';
+const VAULT_ERROR = 'Cannot unlock without a previous vault.';
```
gantunesr, line 208: A note for the future,

We should import this error from the KeyringController. Having the hardcoded string can create a new bug

gantunesr (approved):
LGTM

olenapankina:
**Issue_05**: Translations for vault recovery screens are not presented.

owencraston:
I pulled in the available translations from [here](https://github.com/MetaMask/metamask-mobile/pull/4262/files)

olenapankina (approved):
**Tested on** 
iPhone Xr iOS 16.1.1
Samsung Galaxy M21 Android 12, One UI Core version 4.1
commit e2713133

Issue 01-05 confirmed fixed.

[4421_Android_successful backup.mov](https://user-images.githubusercontent.com/116747263/224792936-55f12eee-3db5-4d1d-84bf-3aa7b7362aad.mov)
[4421_iOS_successful backup.mp4](https://user-images.githubusercontent.com/116747263/224793153-8a18b136-145e-4de4-a7d2-1801f519f4db.mp4)
[4421_Android_failed backup.mov](https://user-images.githubusercontent.com/116747263/224793379-55cd0556-8d76-43d3-9d07-cad00bab0537.mov)
[4421_iOS_failed backup.mp4](https://user-images.githubusercontent.com/116747263/224793460-1b936530-4487-42a6-801a-991627868e4a.mp4)

QA Passed :white_check_mark:

app/components/Views/Login/index.js
owencraston: deleted

PR #4426, opened 2022-05-31
[Fix] Refactor Ledger Message Sign Modal

owencraston:
**Description**
This PR achieved two things...
1. fixes the issue that @zoltanarvai had where the sign modal was not showing up after the MetaMask call to action was pressed
2. refactors the modal to use our latest pattern of being an individual screen. This forces components to navigate to this modal vs conditionally render. This is more declarative, allows for less global state (good for our storage) and is easier to follow where the component gets instantiated from. 

**Checklist**

**Screenshots/Recordings**

https://user-images.githubusercontent.com/22918444/171936154-de15e1f9-2638-47cd-9e78-3003010f25f2.MP4

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/292

I added another issue to track the [refactor of the Transaction Modal](https://github.com/MetaMask/mobile-planning/issues/293)


app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -460,22 +461,11 @@ const RootRPCMethodsUI = (props) => {
...
-        swipeDirection={'down'}
-        propagateSwipe
-        backdropOpacity={1}
-      >
-        <LedgerMessageSignModal />
-      </Modal>
-    ) : null;
+    if (ledgerSignMessageModalVisible && shouldRenderThisModal) {
+      props.navigation.navigate('LedgerConnectFlow', {
+        screen: Routes.LEDGER_MESSAGE_SIGN_MODAL,
```
zoltanarvai, line 466: I suppose we should somehow pass the finalise confirmation here.

owencraston: Yes but I removed the navigation logic from this component into the various Sign components (MessageSign, PersonalSign etc). You can see how I I achieve this [here](https://github.com/MetaMask/metamask-mobile/pull/4426/files#diff-3bf2de563e21a4d886263960ae41b980be1b061ecdc3e97e37299a8db5e736e7R169)

app/components/UI/LedgerModals/LedgerMessageSignModal.tsx
```diff
@@ -1,19 +1,39 @@
...
+  createNavigationDetails,
+  useParams,
+} from '../../../util/navigation/navUtils';
+import Routes from '../../../constants/navigation/Routes';
+import { useAppThemeFromContext, mockTheme } from '../../../util/theme';
+import { View } from 'react-native';
+
+export interface LedgerMessageSignModalParams {
+  messageParams: any;
+  onConfirmationComplete: (confirmed: boolean, rawSignature?: any) => void;
```
owencraston, line 16: @zoltanarvai I'd love to strongly types these instead of using `any`. Do you know if there are corresponding types for these params? especially the `type` param.

app/components/UI/LedgerModals/LedgerMessageSignModal.tsx
```diff
@@ -35,22 +55,31 @@ const LedgerMessageSignModal = () => {
...
 
   return (
-    <LedgerConfirmationModal
-      onConfirmation={executeOnLedger}
-      onRejection={onRejection}
-      deviceId={deviceId}
-    />
+    <ReusableModal ref={modalRef} style={styles.modal}>
+      <View style={styles.contentWrapper}>
+        <LedgerConfirmationModal
```
owencraston, line 76: The more I look at this component...the less I think its a modal. @zoltanarvai is there a case where we individually render this screen? If this is purely view logic, how do you guys feel about renaming it to `LedgerConfirmation` and then turning the modal instance of this into another component wrapped in `ReusableModal` like we have done in this component here? cc @gantunesr

app/components/UI/MessageSign/index.js
```diff
@@ -160,14 +156,14 @@ class MessageSign extends PureComponent {
...
-        type: 'signMessage',
-      });
+      this.props.navigation.navigate(
+        ...createLedgerMessageSignModalNavDetails({
+          messageParams: cleanMessageParams,
+          deviceId: ledgerKeyring.deviceId,
+          onConfirmationComplete: finalizeConfirmation,
+          type: 'signMessage',
+        }),
+      );
```
owencraston, line 166: @zoltanarvai why were we opening the ledger in these `Sign` components (Message, Personal, Sign) `AND` rendering it in the `RootRPCMethodsUI` component? It seems much cleaner to handle it here since we rely on the first modal being displayed.

owencraston: From my testing, only calling navigate in these sub components works as expected. Is there a use case I missed?

app/actions/modals/index.js
```diff
@@ -47,11 +47,6 @@ export const closeLedgerTransactionModal = () => ({
   type: 'CLOSE_LEDGER_TRANSACTION_ACTION',
 });
 
-export const openLedgerSignModal = (params) => ({
```
owencraston, line 50: 🔥

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -382,17 +380,16 @@ const RootRPCMethodsUI = (props) => {
...
 
-  const onSignAction = () => setShowPendingApproval(false);
+  const onSignAction = useCallback(() => {
+    setShowPendingApproval(false);
+  }, []);
 
   const toggleExpandedMessage = () =>
     setShowExpandedMessage(!showExpandedMessage);
 
   const renderSigningModal = () => (
```
owencraston, line 390: @gantunesr I think we should extract this logic into a generic `Sign` component where we passed the type as a prop. There is so much duplicate logic inside each specific sign type component that could be reduced into one function and passed a prop to the child view. If you look at the [finalizeConfirmation](https://github.com/MetaMask/metamask-mobile/blob/3650c6b77a9e9760da8845884e131ab34fa8f5fd/app/components/UI/PersonalSign/index.js#L138) this could likely be reduced to one function.

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -969,7 +969,6 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;
```
owencraston, line 972: Need to remove before merging

app/components/UI/DrawerView/index.js
```diff
@@ -552,8 +552,8 @@ class DrawerView extends PureComponent {
           'LockScreen',
         ].includes(route)
       ) {
-        // eslint-disable-next-line react/no-did-update-set-state
         this.state.showProtectWalletModal &&
+          // eslint-disable-next-line react/no-did-update-set-state
```
owencraston, line 556: @zoltanarvai @gantunesr what is the purpose of this set state inside this component did mount? Seems like risky business. I only edited it since my lint was failing but its unrelated to this change.

app/components/Nav/Main/RootRPCMethodsUI.js
owencraston, line 390: @gantunesr I think we should extract this logic into a generic `Sign` component where we passed the type as a prop. There is so much duplicate logic inside each specific sign type component that could be reduced into one function and passed a prop to the child view. If you look at the [finalizeConfirmation](https://github.com/MetaMask/metamask-mobile/blob/3650c6b77a9e9760da8845884e131ab34fa8f5fd/app/components/UI/PersonalSign/index.js#L138) this could likely be reduced to one function.

owencraston: I added it as a suggestion in our improvement doc.

app/components/UI/LedgerModals/LedgerMessageSignModal.tsx
zoltanarvai: The KeyringController is typed to treat this type as "any" ( return type of KeyringController.signTypedMessage, etc ). The keyring implementation of ledger actually returns this:
```
 Promise<{
    v: number;
    s: string;
    r: string;
  }>
    ```

app/components/UI/LedgerModals/LedgerMessageSignModal.tsx
zoltanarvai: Oh it is referring to the content only you are right, it does not need to be called Modal.

app/components/UI/MessageSign/index.js
zoltanarvai: I don't really mind, I guess I followed in a sense what was there already. the QRSigningModal was in RootRPCMethodsUI and didn't wanna put it to a third place.

app/components/UI/DrawerView/index.js
zoltanarvai: sorry I'm not sure, not something I've added.

PR #4442, opened 2022-06-03
[5.3] Skip to amount to buy when region is selected

wachunei:

This PR adds the skipping screen feature for on ramp aggregator.
- When user has not previously selected a region, they will land on the Region screen (current behavior).
- When user has previously selected a region, they will land on Amount to buy screen.
  - There is a small chance that the selected region changes to be unsupported, when that happens we redirect to Region screen and remove that selection.


app/components/Nav/Main/MainNavigator.js
```diff
@@ -441,6 +441,11 @@ const FiatOnRampAggregator = () => (
         component={Region}
         options={{ animationEnabled: false }}
       />
+      <Stack.Screen
```
alaahd, line 444: I do understand we need some routes here with disabled animations, those routes will still point to the same components but will be carrying a different behavior "disabled animation" I was thinking if you can avoid the creation of additional stacks by setting option dynamically before you `navigate` or `reset` some thing like : 
```suggestion
     navigation.setOptions({
      animationEnabled: false
    })
```
I don't see any major issue with this approach just asking if we can optimize one step further.

alaahd: As long as there is a way to dynamically set options specifically without affecting the global behavior.

app/components/UI/FiatOnRampAggregator/Views/GetStarted.tsx
```diff
@@ -88,12 +93,24 @@ const GetStarted: React.FC = () => {
 
   useEffect(() => {
     if (getStarted) {
-      navigation.reset({
-        index: 0,
-        routes: [{ name: Routes.FIAT_ON_RAMP_AGGREGATOR.REGION_HAS_STARTED }],
-      });
+      if (selectedRegion) {
```
alaahd, line 96: Do you think we should also check for the selected payment method in addition to selectedRegion?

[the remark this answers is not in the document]
wachunei: This is not part of the task

app/components/Nav/Main/MainNavigator.js
alaahd: Ok perfect. this was already a non-blocking feedback.

alaahd (approved):
LGTM!

alaahd (approved):

Cal-L:
Let's account for iPhones with notches on the bottom

image

Cal-L (approved):
Aside from my comment, LGTM!

PR #4443, opened 2022-06-03
[5.3] Fix wrong analytics property for region event

wachunei:

This PR fixes a wrongly mapped value to analytics event.


rickycodes (approved):
lgtm

PR #4453, opened 2022-06-06
Add React Testing Library setup and SendFlow tests

tommasini:
**Description**

Add test support for [react-testing-library](https://testing-library.com/docs/react-testing-library/intro), this should replace Enzyme in the future but for now we kept support for both. As Enzyme works with React Components (checking props and state), react-testing-library works with the actual outputted DOM which is more accurate to what the end user will actually see.

Alongside the initial setup we also added the first test cases for Send Flow screen.

**Checklist**

**Screenshots/Recordings**

_If applicable, add screenshots or recordings to visualize the changes_

**Issue**

Progresses #4465


Fatxx (commented):
Create test guideline to follow as standard for testing

gantunesr (commented):
LGTM. Left a comment to check if it's possible to reduce the number of lines

owencraston (approved):

PR #4469, opened 2022-06-07
Fix displaying notification when state changes

wachunei:

This PR fixes a bug where the update of an order is not being shown as notification


rickycodes (approved):
lgtm

PR #4472, opened 2022-06-07
[FIX] SRP Video Subtitles 

gantunesr:


rickycodes (commented):
This was handled by: https://github.com/MetaMask/metamask-mobile/pull/4394

I'll be opening a new PR which will include all the subtitles (as i mentioned earlier today).

we don't want to include these subtitles _in_ the translation files because the crowdin team keeps trying to translate them. that's the whole reason i removed these keys. (as discussed in #4394)

PR #4477, opened 2022-06-08
Bring in subtitles from crowdin

rickycodes:
This brings in the new subtitle files from #4262 

I need these files on main so the follow up PR (where we actually load these) can read them.


Fatxx (approved):
LGTM

andreahaku (commented):
LGTM

Issue #4473, opened 2022-06-08
Update theme hooks based on design-tokens architecture updates


georgewrmarshall:
### **Description**

Update to the theme hook based on the architecture updates to the design-tokens library

### **Technical Details**

-   TBC

### **Acceptance Criteria**

-   TBC

### **References**

-   https://github.com/MetaMask/design-tokens/pull/148

georgewrmarshall:
complete

PR #4481, opened 2022-06-08
Updating colors and typography to use theme objects

georgewrmarshall:
**Description**

Updating `colors` and `typography` objects to use the `lightTheme` and `darkTheme` objects from design tokens `v1.7`. The next step would be to add a `theme` object to the context so the root theme can be changed and passed through the provider instead of individual objects.

**Dependencies**
https://github.com/MetaMask/design-tokens/pull/148 has to be merged and package updated in mobile before this can be merged. ✅ 

**Checklist**

* ~[x] Tests are included if applicable~ NA
* ~[x] Any added code is fully documented~ NA

**Screenshots/Recordings**

**Issue**
https://github.com/MetaMask/metamask-mobile/issues/4473


app/util/theme/index.ts
```diff
@@ -65,47 +65,47 @@ export const useAppTheme = (): Theme => {
...
       setDarkStatusBar();
       break;
     default:
       // Default uses light theme
-      colors = colorTheme.light;
+      colors = lightTheme.colors;
       setLightStatusBar();
   }
 
-  return { colors, themeAppearance, typography };
```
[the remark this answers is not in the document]
georgewrmarshall: Oh great point! Updated here https://github.com/MetaMask/metamask-mobile/pull/4481/commits/77616f976430c0742162a1fe671abe7ed56a6f1a

app/util/theme/index.ts
```diff
@@ -60,48 +60,55 @@ export const useAppTheme = (): Theme => {
     AppThemeKey.light,
     AppThemeKey.dark,
   );
-  let colors: Colors;
+  let colors: Theme['colors'];
+  let typography: Theme['typography'];
```
georgewrmarshall, line 64: Is this a legit way to traverse types? 😅 Or show we define it in models and export `Colors` and `Typography` types there?  @Cal-L

jpcloureiro (approved):

app/util/theme/models.ts
```diff
@@ -1,15 +1,10 @@
-import { ThemeTypography } from '@metamask/design-tokens/dist/js/typography/types';
-
-// TODO: This should probably be defined from @metamask/design-token library
-export type Colors = any;
```
wachunei, line 4: Removing this type is breaking our `ts` files.
I think it should have been kept as `Colors: DesignTokenTheme['colors'];`.

We use them in the aggregator views:
Screen Shot 2022-06-28 at 10 38 42

Do we have type check enabled? This should have been caught at that stage.

PR #4495, opened 2022-06-14
Fix WalletConnect Icon on connect screen

andrepimenta:
**Description**

Use WalletConnect icon field on connect screen

How to test: Test multiple websites that use WalletConnect, they all should work and they all should show an Icon on the connect modal


cortisiko (approved):
I tested connecting to these dapps through walletconnect:

rariable
opensea
compound.finance
pancake swap
uniswap
olympusdao.finance

I tested opening the dapp in the native browser (chrome/Safari) then connecting via walletconnect

I also test opening the dapp on a desktop browser then scanning the WC QR code using the in app camera to connect to the dapp. 

In both scenarios I see the dapp icon on the connect modal. 

:white_check_mark: :white_check_mark:  This is QA pass!!

gantunesr (approved):

PR #4498, opened 2022-06-14
Add decimals optional prop to keypad in onramp amount view

wachunei:
This PR adds the `decimals` prop to `Keypad` component and uses as fallback when the currency is not found in the `CURRENCIES` list.


alaahd (approved):
LGTM!

gantunesr (approved):
LGTM

PR #4501, opened 2022-06-14
Add accesible false to pressable group preventing VoiceOver interaction

wachunei:
A `Pressable` wrapper was preventing individually selecting UI elements within. This disables its grouping at accessibility layer, unlocking nested UI elements to be accessible.


sethkfman (approved):
LGTM

PR #4528, opened 2022-06-17
Fix onramp error loop by providing payment method change

wachunei:
This PR fixes https://consensyssoftware.atlassian.net/browse/ONRAMP-90: A user remains in a loop after reaching a "no tokens available" error in the amount to buy.

### DIFF
--- a/app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.tsx
+++ b/app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.tsx
@@ -722,6 +722,8 @@ const AmountToBuy = () => {
       <ScreenLayout>
         <ScreenLayout.Body>
           <ErrorView
+            icon="info"
+            title={strings('fiat_on_ramp_aggregator.no_tokens_available_title')}
             description={strings(
               'fiat_on_ramp_aggregator.no_tokens_available',
               {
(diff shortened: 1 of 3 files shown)


app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.tsx
```diff
@@ -722,24 +722,27 @@ const AmountToBuy = () => {
       <ScreenLayout>
         <ScreenLayout.Body>
           <ErrorView
+            icon="info"
```
alaahd, line 725: non-blocking: I suggest we rename this prop to `type` instead of `icon`

locales/languages/en.json
```diff
@@ -1645,7 +1645,9 @@
     "no_providers_available": "No providers available",
     "try_different_amount_to_buy_input": "Try choosing a different payment method or try to increase or reduce the amount you want to buy!",
     "webview_received_error": "WebView received error status code: {{code}}",
-    "no_tokens_available": "There are currently no cryptocurrencies available to purchase on {{network}} in {{region}}",
+    "no_tokens_available_title": "No Tokens Available",
+    "no_tokens_available": "With the selected payment method, there are currently no tokens available to purchase on {{network}}",
```
alaahd, line 1649: I suggest to revise the message to "There are currently no tokens available to purchase on {{network}} with the selected payment method"

alaahd (approved):
LGTM!

app/components/UI/FiatOnRampAggregator/Views/AmountToBuy.tsx
wachunei: I avoid using `type` since it is a reserved word. Let's keep icon, as the other props also refer to elements of the view (icon, title, description, etc...)

locales/languages/en.json
wachunei: ^ @lorenzosantos

lorenzosantos: @alaahd Fantastic concise edit. My only build is to add a period at the end like so: "There are currently no tokens available to purchase on {{network}} with the selected payment method."

wachunei: Got it. will update.

PR #4534, opened 2022-06-19
feat: error handling and swap support

zoltana-ledger:
- Ledger Confirmation Flow now uses the Reusable Modal approach
- SWAP integration
- Retry now works
- If the device is busy we tell the user
- If the operation is not supported we tell the user (Typed data)
- Numerous bug fixes in edge case handling
- Wrapping missed approveTransaction calls with ledger flow
- ERC20 / NFT transfer works


zoltana-ledger:

PR #4541, opened 2022-06-20
Improved jumping while user tap the  Security & Privacy

helinyu:
while i tap the Security & Privacy of the settings list , it‘s alway Stuck . And i can't interact with it after


helinyu:

PR #4554, opened 2022-06-21
Allow for env variable to customise the simulator that gets launched on iOS

rickycodes:
**Description**

This allows the consumer to specify a different simulator via:

```
IOS_SIMULATOR="iPhone 13 Pro Max" yarn start:ios
```


wachunei (approved):
LGTM!

Issue #4077, opened 2022-04-12
[PS] Componentize Network Avatar


Cal-L:
### **Description**

Componentize the `NetworkAvatar` component.

### **Technical Details**

Suggestion
<img width="660" alt="image" src="https://user-images.githubusercontent.com/10508597/163512617-084a768c-dc33-41e6-a5a3-371ece87a1fb.png">

- [ ] Add component to Storybook
- [ ] Write unit test

### **Acceptance Criteria**

- [ ] Can view component in Storybook
- [ ] Unit test passes

### **References**

-  https://www.notion.so/Q2-Extension-IA-Nav-Design-System-Related-Tickets-9e403db21485420e91d1a6c64a3d7065#2852bb6e949f4451a1084f7866b7c3b4

Cal-L:
s-bold-md typography does not play well with XS size

PR #4580, opened 2022-06-23
[PS UI] Add NetworkAvatar component

jpcloureiro:
** Description **

#4077

With image

image

Storybook knobs

image

Without image

image

Without networkName

image


Cal-L (approved):
LGTM

Issue #4601, opened 2022-06-28
Fix missing Color type from theme


Cal-L:
### **Description**

Upgrading to design-tokens v1.7.0 broke the use of `Color` type in style sheets. This PR aims to fix those imports.

### **Technical Details**

-  Replace `Colors` with `ThemeColors` from @metamask/design-tokens library

Scenario: xxxx
- GIVEN a user is in x state
- WHEN a user does x
- AND a user does x
- THEN x should occur

### **References**

-   References go here.
-   Issue numbers. Links.
-   Slack threads.
-   Etc.

cortisiko:
closing since this issue was fixed in https://github.com/MetaMask/metamask-mobile/pull/4602

PR #4602, opened 2022-06-28
Fix color types

Cal-L:
**Description**

#4601 

How it looks now
image

**Screenshots/Recordings**

**Issue**

Progresses #???

**Checklist**

### DIFF
--- a/app/util/theme/models.ts
+++ b/app/util/theme/models.ts
@@ -1,4 +1,7 @@
 import { Theme as DesignTokenTheme } from '@metamask/design-tokens';
+import { ThemeColors } from '@metamask/design-tokens/dist/js/themes/types';
+
+export type Colors = ThemeColors;
 
 export enum AppThemeKey {
   os = 'os',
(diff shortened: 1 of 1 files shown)


wachunei:
I don't think we should import from nested files , ie: `import { ThemeColors } from '@metamask/design-tokens/dist/js/themes/types';`

Why not bring Colors type back from the theme utils, and export it from `Theme['colors']`?
If the export is moved in the `'@metamask/design-tokens` (let's say a file restructure occurred), a lot of instances have to be updated.
Exporting from utils is better abstraction with only one type to keep updated if the dependency changes.

jpcloureiro (commented):
~~Suggest on importing ` { Theme } from '../../../../util/theme/models'` and use the indexed access type `Theme['colors']`~~

~~This way we keep the `@metamask/design-tokens` imports only on `utils/theme`~~

@wachunei beat me to it 😄

wachunei (approved):
LGTM

PR #4604, opened 2022-06-28
Change quotes error to individual events

wachunei:
This PR sends individual events per provider that return a quote with error.


app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
```diff
@@ -328,69 +328,76 @@ const GetQuotes = () => {
...
+            totals.feeAmountRatio / quotesWithoutError.length,
+          provider_onramp_last:
+            quotesWithoutError.length > 1
+              ? quotesWithoutError[quotesWithoutError.length - 1]?.provider
+                  ?.name
+              : undefined,
         });
       }
+
+      if (quotes.length > quotesWithoutError.length) {
```
alaahd, line 386: I suggest we explicitly create and reference quotes with errors as following: 
```suggestion
      quotesWithError = quotes.filter(({ error }) => Boolean(error))
```
then we can check for the length directly 
```suggestion
     if (quotesWithError.length) { 
        quotesWithError.forEach((quote) =>
       // tracking code goes here.
     )}
```

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
```diff
@@ -328,69 +328,76 @@ const GetQuotes = () => {
       !isFetchingQuotes &&
       pollingCyclesLeft >= 0
     ) {
-      const quotesWihoutError: QuoteResponse[] = quotes
+      const quotesWithoutError: QuoteResponse[] = quotes
         .filter(({ error }) => !error)
```
alaahd, line 332: I suggest we use Boolean for dataset filtering. 
```suggestion
        .filter(({ error }) => !Boolean(error))
```

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
wachunei: This would go through the list every time and most of the times is not necessary.

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
wachunei: ! operator already casts to boolean, this is not necessary.

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
```diff
@@ -328,69 +328,76 @@ const GetQuotes = () => {
...
-      });
-
-      if (quotes.length > quotesWihoutError.length) {
-        trackEvent('ONRAMP_QUOTE_ERROR', {
-          provider_onramp_list: quotes
-            .filter(({ error }) => Boolean(error))
-            .map(({ provider }) => provider.name),
+      if (quotesWithoutError.length > 0) {
+        const totals = quotesWithoutError.reduce(
+          (acc, curr) => {
```
gantunesr, line 336: What's `acc`? I tell can't only from the variable name

gantunesr (commented):
LGTM. I'll wait for the other comments to be resolved before approving

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
wachunei: accumulator, it is a common way to name reduce functions arguments imo https://github.com/search?l=JavaScript&q=reduce%28%28acc%2C+curr%29+%3D%3E&type=Code

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
alaahd: Ok perfect.

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
alaahd: Resolved!

app/components/UI/FiatOnRampAggregator/Views/GetQuotes.tsx
alaahd: Resolved!

alaahd (approved):
LGTM! You just need to rebase with main.

gantunesr (approved):
LGTM

cortisiko:
Spoke with Lorenzo and he confirmed this was QA passed. See discussion here https://consensys.slack.com/archives/C039U5608K0/p1657634239853159?thread_ts=1657131054.888789&cid=C039U5608K0

PR #4649, opened 2022-07-11
[IMPROVEMENT] Url redirection

gantunesr:
**Description**

The description of this PR can be found in,

- https://github.com/MetaMask/mobile-planning/issues/312
- https://github.com/MetaMask/mobile-planning/issues/311

**Screenshots/Recordings**

Use case: General QR code reading
https://user-images.githubusercontent.com/17601467/178322039-92f6b571-97e2-4c9a-b6cb-29e48250515d.MP4

Use Case: Reading address
https://user-images.githubusercontent.com/17601467/178322082-50d46c76-23e9-4e4f-9953-9560f59a2741.MP4

**Issue**

Detailed in the section above.

**Checklist**


app/components/Views/QRScanner/index.tsx
```diff
@@ -81,9 +83,10 @@ interface Props {
  * View that wraps the QR code scanner screen
  */
 const QRScanner = ({ navigation, route }: Props) => {
-  const { onScanError, onScanSuccess, onStartScan } = route.params;
-  const mountedRef = useRef(true);
-  const shouldReadBarCodeRef = useRef(true);
+  const { onScanError, onScanSuccess, onStartScan, origin } = route.params;
```
owencraston, line 86: since this screen is only navigated to (vs being a child component) I would highly suggest that we migrate this file to leverage the typed navigation params pattern that way we can enforce the params we expect from this screen. We can also remove the navigation/routes from props. The code would look like this...
```
export interface QRScannerParams {
  onScanError: () => void;
  onScanSuccess: () => void;
  onStartScan: () => void;
  origin: typeof Routes;
}

export const createQRScannerNavDetails =
  createNavigationDetails<QRScannerParams>(Routes.QRScanner);
/**
 * View that wraps the QR code scanner screen
 */
const QRScanner = () => {
  const navigation = useNavigation();
  const { onScanError, onScanSuccess, onStartScan, origin } =
    useParams<QRScannerParams>();


```

Then the components that navigate to this screen would simply call `navigation.navigate(...createQRScannerNavDetails(function, function, function, origin));`

app/components/Views/SendFlow/SendTo/index.js
```diff
@@ -579,6 +579,7 @@ class SendFlow extends PureComponent {
           this.onToSelectedAddressChange(meta.target_address);
         }
       },
+      origin: Routes.SEND_FLOW.SEND_TO,
```
owencraston, line 582: by using the NavDetails pattern this would be enforced

app/components/Views/QRScanner/index.tsx
```diff
@@ -112,6 +150,25 @@ const QRScanner = ({ navigation, route }: Props) => {
         return;
       }
 
+      if (
+        (origin === Routes.SEND_FLOW.SEND_TO ||
```
owencraston, line 154: Thoughts on splitting this check logic into its own function? Then we could call it in order inside this callback? Something like `checkIfInvalidURL()`

app/components/Views/QRScanner/index.tsx
```diff
@@ -112,6 +150,25 @@ const QRScanner = ({ navigation, route }: Props) => {
...
+          origin === Routes.SETTINGS.CONTACT_FORM) &&
+        !isValidHexAddress(content)
+      ) {
+        showAlertForInvalidAddress();
+        end();
+        return;
+      }
+
+      if (/^(ftp|http|https):\/\/[^ "]+$/.test(content)) {
```
owencraston, line 163: same as above^. Thoughts on moving this into a function and calling it inside this callback?

owencraston (commented):
I tested the change and everything seems to be working as expected. NICE ✅. I just left a few comments on the code.

app/components/Views/QRScanner/index.tsx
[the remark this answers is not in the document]
gantunesr: What do you mean by "logic"? advantages? personal preference?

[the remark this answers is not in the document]
gantunesr: Do you see that one of those options brings more value to the code than the other two?

gantunesr: Already mentioned in a comment above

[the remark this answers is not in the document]
owencraston: By logic I just meant reasoning. If you say its personal preference I'm curious why it is the case. The only reason I ask is not because one is better in terms of the result it will give us but more that regex can make the code harder to reason about, especially when there is no descriptive method name. if you made this into a separate, descriptive method would greatly increase the readability of this logic.

owencraston (approved):
🚢 The code looks great. Thanks for implementing those suggestions I made. In my opinion this code is really clear and robust 💪

app/components/Views/QRScanner/index.tsx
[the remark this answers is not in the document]
gantunesr: I see your point, I have updated the code to use a more understandable logic for this method

app/components/Views/QRScanner/index.tsx
gantunesr: [099038a](https://github.com/MetaMask/metamask-mobile/pull/4649/commits/099038a48e9e157f514262dab5cf9913f3b068e1)

app/components/Views/QRScanner/index.tsx
gantunesr: [1a87f44](https://github.com/MetaMask/metamask-mobile/pull/4649/commits/1a87f4473921c677f94c417814d8326b8d32dd67)

app/components/Views/QRScanner/index.tsx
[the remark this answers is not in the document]
gantunesr: [1cf0c71](https://github.com/MetaMask/metamask-mobile/pull/4649/commits/1cf0c711bc2209dc437314a501ca3335db01bc6c) and [987b664](https://github.com/MetaMask/metamask-mobile/pull/4649/commits/987b6643206000f1f333a400056d2e266ae6e99d)

cortisiko (changes requested):
@gantunesr Testing is done. I left a comment with my test cases [here](https://github.com/MetaMask/mobile-planning/issues/312#issuecomment-1239992916). I am going to swing this back your way. Also, just a heads up in the below bug report when I say "in-app camera" I mean tapping on the camera icon in the top nav on the wallet view. 

**Issue 1:** 
This issue has two parts:

A). 
When I use the in-app camera to scan an address QR code I am prompted  with the modal letting me know that [I am about to visit an external link](http://recordit.co/9TNudavZM0 )

furthermore, IF I am on a custom network and I were to scan a mainnet address QR code, I am taken to the send flow with a brief error message ” Recipient address is invalid” before the address resolves itself. [View at the 28-second mark](http://recordit.co/b4a88Ekshu) 
B)
When I use the send flow QR code to scan a receive QR code [I am prompted with the unrecognized QR modal](http://recordit.co/5CIqGYcypA )

to reproduce:
have a production build of the app on device B
while on the wallet view tap on the receive button to open your address QR code
on device A, scan the QR code on device B

If I were to scan an address QR code in production I am able to proceed to the send flow without any issues. [If I were to scan the keystone receive QR code I am able to proceed to the send flow with no issues](http://recordit.co/fGFP10ydPj) 

**Issue 2** 
If I were to scan an invalid QR code using the in-app camera I am NOT prompted with [the invalid QR modal](http://recordit.co/vxxo0XL24I )

[But if I were to scan an invalid QR code by using the camera in the send flow, I am prompted with the correct error message](http://recordit.co/YDXlHf0yuv )

To reproduce:
Using the in app camera scan: https://www.screencast.com/t/nGFsbbVE 

**Issue 3**
Potential design issue: The Url redirect modal differs from that of the designs. See [here](https://www.screencast.com/t/JB7CZQBa0RA)

gantunesr:
@cortisiko 

- **Issue 1-A**: Fixed in the last commit ([c13b578](https://github.com/MetaMask/metamask-mobile/pull/4649/commits/c13b5784ae9668501c3eccdf058a187fcc1aa96c))
- **Issue 1-B**: Fixed in the last commit ([c13b578](https://github.com/MetaMask/metamask-mobile/pull/4649/commits/c13b5784ae9668501c3eccdf058a187fcc1aa96c))
- **Issue 2**: I think we need to consult @holantonela to know what should is the proper behaviour here. From what I understood, the warning modal only shows up in screens that the input should be an address.
- **Issue 3**: The component is not developed yet. It will be included in the next iteration.

holantonela:
> Issue 2: I think we need to consult @holantonela to know what should is the proper behaviour here. From what I understood, the warning modal only shows up in screens that the input should be an address.

Good point. We limit the scanning surface data in each camera view BUT i agree that if the QR code is invalid as malformed or unreadable, we should warn users consistently. 

Is it doable @gantunesr?

cortisiko (approved):
@gantunesr This PR is QA passed 🌮 🌮 🌮 

Test cases for this PR are covered [here](https://www.notion.so/QRScanner-feature-fa7c3023c080461884bd0026d69c3e1e) & [here](https://www.notion.so/importAccount-feature-Mike-cfe336dbe5ae47b6a956298b9eed2bee). I also confirmed that scanning keystone QR codes (keystone account receive QR code && transaction QR codes)  are not broken

PR #4670, opened 2022-07-13
[FEAT] Screenshot Warning

gantunesr:
**Description**

Improve the security around SRP reveal.

_**iOS specific**_. This PR adds a warning when the user takes a screenshot of their SRP or Private Key.

Changes:

- [ ] ~Introduction of `react-native-detector` v0.2.1 package~
- [x] Refactor of `RevealPrivateCredential` and move styles to a different file
- [x] Refactor of `ManualBackupStep1` and move styles to a different file

 it also refactors the component `RevealPrivateCredential` and `ManualBackupStep1`

**Screenshots/Recordings**

https://user-images.githubusercontent.com/17601467/178645859-b41bff01-b25b-4fa3-8d74-39a326c4894a.MP4

Note: You can observe a similar behaviour in the onboarding flow.

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/227

**Checklist**


owencraston:
Is the goal of this change to warn the user _before_ they take a screenshot or simply to warn them whenever they take one? Would it even be possible to prevent a screenshot when the user takes one?

https://user-images.githubusercontent.com/22918444/180300751-2862a9b2-edfb-4bc7-aae0-227208c20eef.MOV

app/util/analyticsV2.js
```diff
@@ -161,6 +161,10 @@ export const ANALYTICS_EVENTS_V2 = {
   ),
   ONRAMP_EXTERNAL_LINK_CLICKED: generateOpt('External Link Clicked'),
   ONRAMP_QUOTE_ERROR: generateOpt('On-ramp Quote Error'),
+  // Screenshots
+  SCREENSHOT_WARNING: generateOpt('screenshot_warning_seen'),
```
owencraston, line 165: nit pick: is this really a screen? Would `screenshot_warning_displayed` be more fitting?

package.json
```diff
@@ -200,6 +200,7 @@
     "react-native-confetti-cannon": "^1.5.0",
     "react-native-crypto": "2.1.2",
     "react-native-default-preference": "^1.4.3",
+    "react-native-detector": "0.2.1",
```
owencraston, line 203: what is the logic of using this package over [expo-screen-capture](https://www.npmjs.com/package/expo-screen-capture) which appears to more users and is better maintained?

app/util/testSetup.js
```diff
@@ -105,6 +104,9 @@ jest.mock('react-native-reanimated', () =>
 jest.mock('react-native-background-timer', () => 'RNBackgroundTimer');
 jest.mock('@react-native-community/async-storage', () => mockAsyncStorage);
 jest.mock('@react-native-community/cookies', () => 'RNCookies');
+jest.mock('react-native-detector', () => ({
+  addScreenshotListener: () => null,
```
owencraston, line 108: would it make more sense for this to be a void function?

app/components/Views/RevealPrivateCredential/index.js
```diff
@@ -1,229 +1,83 @@
...
+  Alert,
 } from 'react-native';
-import AsyncStorage from '@react-native-community/async-storage';
 import PropTypes from 'prop-types';
 import QRCode from 'react-native-qrcode-svg';
 import ScrollableTabView, {
   DefaultTabBar,
 } from 'react-native-scrollable-tab-view';
 import Icon from 'react-native-vector-icons/FontAwesome5';
+// eslint-disable-next-line import/no-unresolved
```
owencraston, line 21: why is this needed?

app/components/Views/RevealPrivateCredential/index.js
```diff
@@ -1,229 +1,83 @@
...
 import PreventScreenshot from '../../../core/PreventScreenshot';
-import SecureKeychain from '../../../core/SecureKeychain';
-import { fontStyles } from '../../../styles/common';
 import AnalyticsV2 from '../../../util/analyticsV2';
 import Device from '../../../util/device';
 import { strings } from '../../../../locales/i18n';
 import { isQRHardwareAccount } from '../../../util/address';
 import AppConstants from '../../../core/AppConstants';
-
-const createStyles = (colors) =>
```
owencraston, line 47: ❤️

gantunesr:
@owencraston in response to https://github.com/MetaMask/metamask-mobile/pull/4670#issuecomment-1191861490

The goal is just to warn the user about their action and state explicitly that we don't recommend. In iOS you can't prevent the screenshot.

app/util/testSetup.js
gantunesr: Not sure actually, but I have no preference so I'll just change it to void

package.json
gantunesr: Crazy, `expo-screen-capture` didn't show up in my research. I'll take a look at it and update the code if it brings more value than `react-native-detector`

app/util/analyticsV2.js
gantunesr: Agreed

package.json
owencraston: I don't feel too strongly about which package we use. My only thoughts are, I have used expo packages in the past and they are high quality and we have much less risk of the package being stale or deprecated as expo is an established project. The one caveat would be if expo requires us to add a lot of extra dependancies  it might not be worth it. I'd love to hear the results of your research.

owencraston:
@gantunesr thanks for clarifying. I see how that it's not possible on iOS. I am curious how we went about disabling screenshots on Android? is it something we disable only on one screen or is it app wide? if we are disabling it on one screen then we should try and consolidate the logic since these packages allow us to disable screenshots on android.

app/components/Views/RevealPrivateCredential/index.js
```diff
@@ -332,31 +124,107 @@ class RevealPrivateCredential extends PureComponent {
...
-  tryUnlock = () => {
-    this.setState({ isModalVisible: true });
+  const openSRPGuide = () => {
+    AnalyticsV2.trackEvent(AnalyticsV2.ANALYTICS_EVENTS.SCREENSHOT_WARNING);
+    Linking.openURL(SRP_GUIDE_URL);
   };
 
-  onPasswordChange = (password) => {
-    this.setState({ password });
+  const showScreenshotAlert = useCallback(() => {
```
owencraston, line 138: I would advocate turning the below (line 138-188) login into a hook so it could be reused and changed easily in the future.

app/components/Views/RevealPrivateCredential/index.js
```diff
@@ -332,31 +124,107 @@ class RevealPrivateCredential extends PureComponent {
...
+            AnalyticsV2.trackEvent(AnalyticsV2.ANALYTICS_EVENTS.SCREENSHOT_OK),
+        },
+      ],
+    );
+  }, [privateCredentialName]);
+
+  useEffect(() => {
+    updateNavBar();
+    InteractionManager.runAfterInteractions(() => {
+      PreventScreenshot.forbid();
```
owencraston, line 166: I am a bit confused, should we not be doing an android/ios check for this logic?

app/components/Views/RevealPrivateCredential/index.js
```diff
@@ -332,31 +124,107 @@ class RevealPrivateCredential extends PureComponent {
...
+    const credential = credentialName || route.params.privateCredentialName;
+    return credential === PRIVATE_KEY;
   };
 
-  copyPrivateCredentialToClipboard = async (privateCredentialName) => {
+  const navigateBack = () => {
+    navigation.pop();
+  };
+
+  const cancelLol = () => {
```
owencraston, line 199: giphy

owencraston: ^ I hope this wasn't too sassy 😄

app/components/Views/ManualBackupStep1/index.js
```diff
@@ -493,29 +285,75 @@ class ManualBackupStep1 extends PureComponent {
...
-    return (
-      <SafeAreaView style={styles.mainWrapper}>
-        <View style={styles.onBoardingWrapper}>
-          <OnboardingProgress currentStep={currentStep} steps={this.steps} />
-        </View>
-        {view === SEED_PHRASE
-          ? this.renderSeedphraseView()
-          : this.renderConfirmPassword()}
-      </SafeAreaView>
+  const showScreenshotAlert = useCallback(() => {
```
owencraston, line 293: would it be possible to extract the below logic (line 293-324)  into a hook that could be used in both this file and the `app/components/Views/RevealPrivateCredential/index.js` file? I also commented on that file about extracting the logic into a hook.

app/components/Views/RevealPrivateCredential/index.js
gantunesr: I'm getting a linter error in my machine, I'll remove and see if it also occurs in the workflow

gantunesr:
@owencraston regarding https://github.com/MetaMask/metamask-mobile/pull/4670#issuecomment-1197365885

> I am curious how we went about disabling screenshots on Android?

We created a native module to disable it. You can check it [here](https://github.com/MetaMask/metamask-mobile/blob/main/android/app/src/main/java/io/metamask/nativeModules/PreventScreenshot.java)

> Is it something we disable only on one screen or is it app wide?

Only SRP or Private Key related screens

> if we are disabling it on one screen then we should try and consolidate the logic since these packages allow us to disable screenshots on android.

I support this decision, but it would be out of scope for this development

app/components/Views/RevealPrivateCredential/index.js
gantunesr: PreventScreenshot already performs this check internally.

```
export default {
  forbid: isAndroid ? NativeModules.PreventScreenshot.forbid : () => true,
  allow: isAndroid ? NativeModules.PreventScreenshot.allow : () => true,
};
```

app/components/Views/RevealPrivateCredential/index.js
gantunesr: Lol good meme

gantunesr: It was because there were two `cancel` methods and I had to change the name of one to avoid errors, of course this is not the final name haha

app/util/analyticsV2.js
gantunesr: [7139321](https://github.com/MetaMask/metamask-mobile/pull/4670/commits/71393213a31e7efb46944534a4b36f0c39a68c33)

app/util/testSetup.js
gantunesr: What's your suggestion here? I tried changing it to void but I got some errors

app/components/Views/RevealPrivateCredential/index.js
gantunesr: [a9195dd](https://github.com/MetaMask/metamask-mobile/pull/4670/commits/a9195ddc1f0e18d90fb5ad44934261e4806327c5)

app/components/Views/RevealPrivateCredential/index.js
gantunesr: [26aa5dd](https://github.com/MetaMask/metamask-mobile/pull/4670/commits/26aa5dd69f0afc67abe1dc174aa119858d11697f)

app/components/Views/RevealPrivateCredential/index.js
gantunesr: For a second iteration I want to use a custom Alert component that our design system is building, I can try to move some logic to a hook but not the entire code. Wdyt?

app/components/Views/ManualBackupStep1/index.js
gantunesr: Refer to https://github.com/MetaMask/metamask-mobile/pull/4670#discussion_r932355236

app/util/testSetup.js
owencraston: It doesn't really matter. its just a nit pick. maybe try `jest.fn()`?

app/components/Views/RevealPrivateCredential/index.js
owencraston: Yeah I was mostly referring to the useEffect logic of adding/removing the listener. I think migrating it a hook and sharing it between the two screens in this PR would be a great step in componentizing this logic.

owencraston:
> > if we are disabling it on one screen then we should try and consolidate the logic since these packages allow us to disable screenshots on android.
> 
> I support this decision, but it would be out of scope for this development

Looking at the [expo package screen capture package](https://docs.expo.dev/versions/latest/sdk/screen-capture/#screencapturepreventscreencaptureasynckey), it seems like we could easily leverage this package to do both the screenshot prevention and the screenshot listener.

Also, slightly unrelated but it seems we can prevent screen recording on iOS for certain screens.
Screen Shot 2022-07-28 at 11 14 51 AM

package.json
owencraston: Looking a little closer, the[ expo library](https://docs.expo.dev/versions/latest/sdk/screen-capture/#api) allows us to listen for screenshots as well as prevent them. This could be a quick win to remove that native module. Since we need to add a library for this change anyway...its probably better that it does more than one thing for us.

This package seems a bit stale too 😏 
Screen Shot 2022-07-28 at 11 18 15 AM

app/util/testSetup.js
gantunesr: [4b31604](https://github.com/MetaMask/metamask-mobile/pull/4670/commits/4b316042c52764fb868ded121202edd6862feb31)

package.json
gantunesr: After reading about this expo dependencies and working on the configuration I think it's better to have a wider discussion with the team about the introduction of these dependencies and the repercussions it'll have in the project. For the moment the current dependency should be enough, its really small and with minimum risk. We can always develop a custom solution if this package becomes incompatible a some point or change to the expo package if the mobile team agrees to do so.

We can also include the Android screenshot solution to this next development to improve that development.

app/components/Views/RevealPrivateCredential/index.js
gantunesr: [8ec32b6](https://github.com/MetaMask/metamask-mobile/pull/4670/commits/8ec32b6462a71fadc16310ead53c8509e989fbe3) wdyt?

app/components/hooks/useScreenshotWarning.ts
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from 'react';
+import { addScreenshotListener } from 'react-native-detector';
+
+const useScreenshotWarning = (warning: () => void) => {
+  const [enabled, setEnabled] = useState<boolean>(false);
+  useEffect(() => {
+    const userDidScreenshot = () => {
+      if (enabled) {
+        warning();
```
owencraston, line 9: Looking at the the differences between the two `showScreenshotAlert` functions in the the two files this hook is being called. 

```
// app/components/Views/RevealPrivateCredential/index.js
const showScreenshotAlert = useCallback(() => {
    AnalyticsV2.trackEvent(AnalyticsV2.ANALYTICS_EVENTS.SCREENSHOT_WARNING);
    Alert.alert(
      strings('reveal_credential.screenshot_warning_title'),
      strings('reveal_credential.screenshot_warning_desc', {
        credentialName:
          privateCredentialName === PRIVATE_KEY
            ? strings('reveal_credential.private_key_text')
            : strings('reveal_credential.srp_abbreviation_text'),
      }),
      [
        {
          text: strings('reveal_credential.learn_more'),
          onPress: openSRPGuide,
          style: 'cancel',
        },
        {
          text: strings('reveal_credential.got_it'),
          onPress: () =>
            AnalyticsV2.trackEvent(AnalyticsV2.ANALYTICS_EVENTS.SCREENSHOT_OK),
        },
      ],
    );
  }, [privateCredentialName]);

// app/components/Views/ManualBackupStep1/index.js
const showScreenshotAlert = useCallback(() => {
	AnalyticsV2.trackEvent(AnalyticsV2.ANALYTICS_EVENTS.SCREENSHOT_WARNING);
    Alert.alert(
      strings('manual_backup_step_1.screenshot_warning_title'),
      strings('manual_backup_step_1.screenshot_warning_desc'),
      [
        {
          text: strings('reveal_credential.learn_more'),
          onPress: openSRPGuide,
          style: 'cancel',
        },
        {
          text: strings('reveal_credential.got_it'),
          onPress: () =>
            AnalyticsV2.trackEvent(AnalyticsV2.ANALYTICS_EVENTS.SCREENSHOT_OK),
        },
      ],
    );
  }, []);
  ```

Seeing here are there very little differences between these functions. Could we just bring this function into this hook so we don't need to repeat the code and this hook becomes more fully featured?

File: app/components/Views/ManualBackupStep1/index.js [hunk 11]
```diff
@@ -493,29 +285,68 @@ class ManualBackupStep1 extends PureComponent {
...
+      ],
     );
-  }
-}
+  }, []);
+
+  const [enableScreenshotWarning] = useScreenshotWarning(showScreenshotAlert);
+
+  useEffect(() => {
+    enableScreenshotWarning(view === SEED_PHRASE);
```
owencraston, line 316: what is the reason we need to check `view === SEED_PHRASE`?

app/components/hooks/useScreenshotWarning.ts
gantunesr: As I said in https://github.com/MetaMask/metamask-mobile/pull/4670#discussion_r932355236, the reason to not bring that logic inside the hook is because I'll change the `Alert` for a custom component in a next development after that component is merged into main

app/components/Views/ManualBackupStep1/index.js
gantunesr: Because I only want to enable the warning when the SRPi s revealed and not for all the views in that UX

owencraston (approved):
Thanks for taking those extra iterations. I didn't run the latest changes but the code looks good 🚢

cortisiko:
@gantunesr I did a quick run through to see how this feature works and I received a `[TypeError: undefined is not an object (evaluating '_this.props.passwordSet')]` when I attempted to reveal my SRP.

plasmacorral:
QA Build 972 crashes on Android 13, 12 and 11 when I attempt to access SRP with a device that uses PIN authenticaton with NO biometric auth factors present on the device.  I also noticed that when I attempt to reveal the private key, I get "Incorrect Password" even though I am using the correct password (not shown in video, but also easy to reproduce)

https://recordit.co/Uh75WoiAem

IOS 15.6.1 using a device with biometrics present (but not used by MetaMask) I get this error on QA build 972 when I attempt to reveal SRP:

gantunesr:
Copy updated according to https://github.com/MetaMask/mobile-planning/issues/227#issuecomment-1307108805 in commit [e951e16](https://github.com/MetaMask/metamask-mobile/pull/4670/commits/e951e16e61d342e0d00a834f9ad32a700674bfa0)

gantunesr (commented):
Note: Revert code version and build number before merging

cortisiko (approved):
This PR is 🌮 🌮 .

Scenarios covered outside of the AC mentioned [here](https://github.com/MetaMask/mobile-planning/issues/227#issue-1197515940):

- A user attempts to take a screenshot while his language is not set to English. Ensuring that the text in the warning alert defaults to English
- Regression on android. Nothing is broken there.

Issue #4308, opened 2022-05-13
Unable to re-add account nickname after making it an empty string


cortisiko:
**Describe the bug**
 If i were to remove my account nick name i.e. leave it blank, I am unable to add a new nickname to my account. 

see [recording](http://recordit.co/blCne5klOv )

**To Reproduce**
- launch the app
- create/import your wallet
- while on the wallet view
- tap and hold the account nickname to highlight it
- remove the default nickname
- attempt to add a new account nick name. Notice you are unable to do so

**Expected behavior**
You should be able to update an account nick name.

gantunesr:
Have in consideration the accounts that already have an empty string as name

cortisiko:
was fixed in https://github.com/MetaMask/metamask-mobile/pull/4684

PR #4684, opened 2022-07-18
Account nickname is always defined

tommasini:
**Description**
The issue is that when the nickname of the account would be empty, it would be non editable.

**Proposed Solution**
The proposed solution it is, when the name it's empty we will assume that the name of the account will be the last name of it.
Example: 
If it was "Account 2" after confirm the empty string on the account name, it will come back to be "Account 2"

Also its proposed, for the users that already have no nickname on their accounts we set the default name "Account" when the user opens the app. 

**Code Impact**
Low

**Test Cases**
Case1:
* Notice the account nickname
* Long press the account nickname
* Edit the name to an empty string
* Will came back to the last account nickname

**Screenshots/Recordings**

https://user-images.githubusercontent.com/46944231/179572818-7b5589f9-b820-48a9-bc10-692f2213229b.mp4

**Issue**

Progresses #4308 

**Checklist**


jakehaugen:
Looks great to me!

app/components/UI/AccountOverview/index.js
```diff
@@ -245,6 +245,13 @@ class AccountOverview extends PureComponent {
     InteractionManager.runAfterInteractions(() => {
       this.doENSLookup();
     });
+
+    //This is to fix the accounts which doesn't had name on their accounts
+    //If this is on Production for some time can be removed.
+    const { PreferencesController } = Engine.context;
+    if (!accountLabel) {
+      PreferencesController.setAccountLabel(selectedAddress, 'Account');
+    }
```
gantunesr, line 254: I think the code is self-explanatory, you can remove the comment

gantunesr (commented):
LGTM. Good work, just a small comment

app/components/UI/AccountOverview/index.js
tommasini: Okok! Doing it

gantunesr (approved):

cortisiko (changes requested):
@tommasini I found a fun edge case:

If I were to:
- long press to edit my account nickname
- remove the current name
- tap on the space bar to set ' ' as my account nickname
- upon doing so, I still experience the bug where I am unable to edit my account nickname after setting it to ' '

see [recording](http://recordit.co/2Ngj9SyMVH)

tommasini:
@cortisiko Really well catched, I was not considering the blank spaces!! Thank you Curtis! Awesome!!!

@gantunesr Could you give me another review please?

gantunesr (approved):

cortisiko (approved):
🌮 🌮 looks good @tommasini

PR #4686, opened 2022-07-18
Remove duplicated properties from trackEvent objects

wachunei:
**Description**

This PR removes duplicated object properties from `trackEvent` calls


Cal-L (approved):
LGTM

PR #4740, opened 2022-07-27
Fix duplicate property in on-ramp events definition

wachunei:
**Description**

This PR fixes a duplicate property definition error


Cal-L (approved):
LGTM

PR #4742, opened 2022-07-27
Add On-Ramp generic error view event

wachunei:
**Description**

This PR tracks the 'On-ramp Error' event when the Error View is displayed.

The error payload is: 

```ts
  ONRAMP_ERROR: {
    location: ScreenLocation;
    message: string;
    payment_method_id?: string;
    region?: string;
    currency_source?: string;
    currency_destination?: string;
  };
```


app/components/UI/FiatOnRampAggregator/components/ErrorView.tsx
```diff
@@ -80,15 +84,38 @@ function ErrorView({
...
 
   const ctaOnPressCallback = useCallback(() => {
     ctaOnPress?.();
   }, [ctaOnPress]);
 
+  useEffect(() => {
+    trackEvent('ONRAMP_ERROR', {
+      location,
+      message: description,
+      payment_method_id: selectedPaymentMethodId as string,
```
Cal-L, line 108: Is this type casted as `string` because it's potentially undefined and we can't `Number(undefined)`?

Cal-L (approved):
Just left a question. LGTM!

alaahd:
LGTM!

app/components/UI/FiatOnRampAggregator/components/ErrorView.tsx
wachunei: It is just to match the type, we know an undefined value would not affect the code so it is safe to cast it.

Issue #4561, opened 2022-06-21
On-chain NFTs do not render in the NFT tab


shanejonas:
**Describe the bug**
On-chain NFTs do not render in the NFT tab

**Screenshots**
<img width="423" alt="image" src="https://user-images.githubusercontent.com/364566/174907518-2ab3f530-469f-4c4d-bf4a-52c8e1646d84.png">

**To Reproduce**
(on rinkeby)
1. Go to https://oneclickdapp.com/scoop-sensor
2. Click on write
3. click mint
4. wait for tx
5. see collection show up but not the image

**Expected behavior**
I should see an image for on-chain SVGs

**Problem**
for on-chain SVGs `tokenURI(id)` returns base64 urls, they need to be parsed into JSON, then use the `image` key:
<img width="1217" alt="image" src="https://user-images.githubusercontent.com/364566/174907931-d79c2cb5-f8d5-4db7-9e66-8cc88531889a.png">

Here is what my mint site does to pull out the metadata: 
<img width="536" alt="image" src="https://user-images.githubusercontent.com/364566/175323691-b9f6a711-6900-44b8-a57e-12f4612be3ba.png">

tommasini:
Move this on to the icebox until we decide what to do.

What I concluded with my Android native research for implementing the remote svgi t's that to render a local svg file it's easier, but when we talk about having the SVG code coming remotely, it's not so easy, why:
 * We need to convert the svg code to vector code (with a file we could use the [shapeshifter](https://shapeshifter.design/)). 
 * Shapeshifer uses [paperJS](http://paperjs.org/features/#svg-import-and-export), for my understanding do not support some tags of the svg code (tested on the shapeshifter website)

shanejonas:
this is still a valid issue. MetaMask Mobile should support on-chain NFTs.

mastergaurang94:
Very much still a valid issue.

lampalork85:
Same problem on Chrome extension with On-Chain image of nft , the strange behavior is that in the preview of an NFT transfer, the image is displayed correctly :

![image](https://github.com/MetaMask/metamask-mobile/assets/38051081/8d4851d5-8ed7-4f54-b259-4e2429d6b27f)

 but not in the NFT panel where we see the default image from MM :
 
![image](https://github.com/MetaMask/metamask-mobile/assets/38051081/260a9f63-5a62-44ee-8b4f-ffff1090977e)

vrde:
[...]

PR #4757, opened 2022-07-29
Implementation of webview to support svg with style and animations

tommasini:
**Description**
The issue is that we can't render the SVG with the style tag with the react-native-svg library or image component of React native.

**Propose Solution**
Using a webview, we can render the SVG with the style and animation. This led to some security concerns.
So in order to try to eliminate the security concerns, we use a [library](https://github.com/cure53/DOMPurify) with version 2.3.10 (the latest version for now) to sanitize the SVG.
We forbidden the tag style on the html.
On webView we disabled opening automatically a window with javascript and the dom storage (dom storage only on android).

**Test Case**
In order to see how this sanitization algorithm works, we did some tests and added some possible malicious code to an SVG.
We render an SVG with javascript code and some attributes that could inject cross-site scripting and we initiate the sanitize function to clean it.
https://pastebin.com/i0tfrG6t
 To reproduce: 
 * Create an index.html file with the code on Pastebin
 * Open it on the browser
 * Inspect the svg
 * You can notice that some attributes are removed compared to the Pastebin 

We have this SVG:
`<svg width="329" height="329" id="svg" xmlns="http://www.w3.org/2000/svg" preserveAspectRatio="xMinYMin meet" viewBox="0 0 24 24"> <g><animate
attributeName="xlink:href" begin="0"
from="javascript:alert(1)" to="&"
/>` 

Transforms into

`<svg viewBox="0 0 24 24" preserveAspectRatio="xMinYMin meet" xmlns="http://www.w3.org/2000/svg" id="svg" height="329" width="329"> <g><animate begin="0">`

 

**Screenshots/Recordings**
<p align='center'> 
 
</p>

**Issue**

Progresses #4561 

**Checklist**

### DIFF
--- a/app/components/Base/RemoteImage/index.js
+++ b/app/components/Base/RemoteImage/index.js
@@ -1,6 +1,12 @@
 import React, { useMemo, useState } from 'react';
 import PropTypes from 'prop-types';
-import { Image, ViewPropTypes, View, StyleSheet } from 'react-native';
+import {
+  Image,
+  ViewPropTypes,
+  View,
+  StyleSheet,
+  PixelRatio,
+} from 'react-native';
 import FadeIn from 'react-native-fade-in-image';
 // eslint-disable-next-line import/default
 import resolveAssetSource from 'react-native/Libraries/Image/resolveAssetSource';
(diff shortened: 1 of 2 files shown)


tommasini:
Converting to a draft because needs more research

PR #4786, opened 2022-08-02
Turn on no-unused-expressions and react/no-unused-class-component-methods in .eslint.js

owencraston:
**Description**

- to add stricter lint rules that prevent unused code
- update the linter to check for the [react/no-unused-class-component-methods](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unused-class-component-methods.md) rule
- turn on [no-unused-expressions](https://eslint.org/docs/latest/rules/no-unused-expressions#:~:text=An%20unused%20expression%20which%20has,statement%20n%20%2B%3D%201%3B%20instead.)
- after these rules are enabled, went through and cleaned up all the unused props, expressions, variables and methods
- I also had to validate a few props since that rule got further enforced. 

**Reviewing**
- while these are just lint rules I want to be careful before merging as I did delete a lot of code and I think we should have at least a few reviews and a test of some key flows before merging.

**Checklist**

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -108,7 +108,10 @@ module.exports = {
       },
     ],
     'no-unsafe-negation': 2,
-    'no-unused-expressions': 2,
+    'no-unused-expressions': [
+      2,
+      { allowShortCircuit: true, allowTernary: true },
+    ],
     'no-use-before-define': [2, 'nofunc'],
     'no-useless-call': 2,
     'no-useless-computed-key': 2,
@@ -148,6 +151,8 @@ module.exports = {
     'react/no-string-refs': 2,
     'react/no-unused-prop-types': 2,
     'react/prefer-es6-class': 2,
+    'no-unused-vars': 2,
+    'react/no-unused-class-component-methods': 2,
     radix: 0,
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -108,7 +108,7 @@ module.exports = {
       },
     ],
     'no-unsafe-negation': 2,
-    'no-unused-expressions': 'off',
+    'no-unused-expressions': 2,
     'no-use-before-define': [2, 'nofunc'],
     'no-useless-call': 2,
     'no-useless-computed-key': 2,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -153,6 +153,7 @@ module.exports = {
     'react/prefer-es6-class': 2,
     'no-unused-vars': 2,
     'react/no-unused-class-component-methods': 2,
+    'react/display-name': 1,
     radix: 0,
   },
 };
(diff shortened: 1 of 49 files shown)


app/components/Views/Send/index.js
```diff
@@ -485,112 +424,6 @@ class Send extends PureComponent {
     this.state.mode === REVIEW && this.trackOnCancel();
   };
 
-  /**
```
owencraston, line 488: The amount of unused code in this file is concerning and makes me think we should be very diligent in the review/test.

app/components/Views/SendFlow/AddressList/index.js
```diff
@@ -207,10 +207,9 @@ class AddressList extends PureComponent {
       const initial = nameInitial
         ? nameInitial[0]
         : strings('address_book.others');
+      if (contact.isSmartContract && !this.props.onlyRenderAddressBook) return;
       if (Object.keys(addressBookTree).includes(initial)) {
         addressBookTree[initial].push(contact);
-      } else if (contact.isSmartContract && !this.props.onlyRenderAddressBook) {
```
owencraston, line 212: @blackdevelopa helped me make this change since it violated https://eslint.org/docs/latest/rules/no-unused-expressions

app/util/testSetup.js
```diff
@@ -93,9 +93,7 @@ jest.mock('react-native-keychain', () => ({
 }));
 jest.mock('react-native-share', () => 'RNShare');
 jest.mock('react-native-branch', () => ({
-  BranchSubscriber: () => {
```
owencraston, line 96: the tests still pass and this fixes the lint error

package.json
```diff
@@ -299,7 +299,7 @@
     "eslint-import-resolver-typescript": "^2.4.0",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-prettier": "^3.3.1",
-    "eslint-plugin-react": "7.16.0",
```
owencraston, line 302: I updated to get access to the new `react/no-unused-class-component-methods` rule

.eslintrc.js
```diff
@@ -148,6 +151,9 @@ module.exports = {
     'react/no-string-refs': 2,
     'react/no-unused-prop-types': 2,
     'react/prefer-es6-class': 2,
+    'no-unused-vars': 2,
+    'react/no-unused-class-component-methods': 2,
+    'react/display-name': 1,
```
owencraston, line 156: this came up a few times as an error and it does't seem like a huge deal so I felt a warning would be sufficient.

app/components/Views/Root/index.js
```diff
@@ -53,11 +49,15 @@ export default class Root extends PureComponent {
 const ConnectedRoot = () => {
   const theme = useAppTheme();
 
+  const errorHandler = (error, stackTrace) => {
```
owencraston, line 52: @andrepimenta was this what you meant when you referenced `this.errorHandler` below?

PR #4793, opened 2022-08-03
Dapp connect Warning

andrepimenta:
**Description**

Warn the user to update the app to a new version

**Screenshots/Recordings**
Simulator Screen Shot - iPhone 11 Pro - 2022-08-03 at 15 45 16

**Issue**

Progresses #???

**Checklist**


app/core/DeeplinkManager.js
```diff
@@ -239,7 +239,9 @@ class DeeplinkManager {
           // action is the first part of the pathname
           const action = urlObj.pathname.split('/')[1];
 
-          if (action === ACTIONS.WC && params?.uri) {
+          if (action === 'connect') {
```
owencraston, line 242: should we add this to the actions enum like we did for wallet connet?

app/core/DeeplinkManager.js
```diff
@@ -322,8 +324,9 @@ class DeeplinkManager {
       // For ex. go to settings
       case PROTOCOLS.METAMASK:
         handled();
-
-        if (url.startsWith('metamask://wc')) {
+        if (url.startsWith('metamask://connect')) {
```
owencraston, line 327: these should be constants. This logic could also be abstracted into a function that performs the correct alert. Thats not a blocker but would definitely be nice to get these keys into a constant/enum.

owencraston (approved):
Beautiful!

cortisiko:
@andrepimenta 
Lone issue: [on android when I attempt to deep link, I am taken to the browser](http://recordit.co/qQvF06YltY)

PR #4808, opened 2022-08-04
Add entry point in mobile app to labs dapp

meppsilon:
**Description**

This PR is about adding the entry point button that will link users to the Metalabs dapp in the Metamask Mobile browser.

There were a couple important decisions that I made:

I placed the button in both the Tokens tab and the CollectibleContracts tab. Initially, the button is supposed to sit below the list of tabs in the Wallet component and above the content within the tabs but I couldn't find a way to fit a UI element between the tabs list and the tabs content so I ended up placing the button in both of the components for the tab contents. That would allow it to show up for both tabs but it creates a weird behavior for the user because as they click through the tabs, all of the contents within the tabs will change but the entry buttons will remain the same, even though they slide in and out and they are technically different buttons.
I hardcoded metalabs.consensys.net as an approvedHost. The functionality we want is for the user to click the link and for the browser to open up on the metalabs page with the selected address already approved. This is the same functionality that we will be receiving from the extension. The solution of hardcoding the metalabs url as an approved host allowed for the accounts rpc requests from the metalabs dapp to automatically go through successfully. I considered another solution where I would check separately if the hostname was the same as the metalabs dapp url but I thought that solution was even more hardcoded than the previous and I didn't want to introduce more logic for approving an eth_accounts request that could possibly add confusing and overhead in the future. So I chose to add on top of the existing logic instead.

**Screenshots/Recordings**

https://www.figma.com/file/gi92L28Nr4PdUaLsTRhmkK/User-Flows---Master?node-id=106%3A14700

**Issue**

Progresses #???

**Checklist**


.js.env.example
```diff
@@ -18,3 +18,4 @@ export MM_FOX_CODE="EXAMPLE_FOX_CODE"
 # More info: https://github.com/MetaMask/metamask-mobile/issues/1984
 
 export MM_INFURA_PROJECT_ID="null"
+export METALABS_URL="https://metalabs.consensys.net"
```
gantunesr, line 21: We need to add this variable to our CI/CD workflow

locales/languages/en.json
```diff
@@ -281,6 +281,7 @@
     "collectible": "Collectible",
     "collectibles": "NFTs",
     "transactions": "TRANSACTIONS",
+    "goToPortfolio": "Go to portfolio",
```
gantunesr, line 284:
```suggestion
    "go_to_portfolio": "Go to portfolio",
```

app/components/UI/CollectibleContracts/index.js
```diff
@@ -253,6 +290,21 @@ const CollectibleContracts = ({
...
+          onPress={onPortfolioViewPress}
+          style={styles.metalabsPress}
+        >
+          <Image
+            style={styles.metalabsImage}
+            source={chartLineImg}
+            resizeMode={'contain'}
+          />
+          <Text centered numberOfLines={1} style={styles.metalabsText}>
+            {strings('wallet.goToPortfolio')}
```
gantunesr, line 304:
```suggestion
            {strings('wallet.go_to_portfolio')}
```

app/components/UI/Tokens/index.js
```diff
@@ -432,13 +460,38 @@ class Tokens extends PureComponent {
...
+            onPress={this.onPortfolioViewPress}
+            style={styles.metalabsPress}
+          >
+            <Image
+              style={styles.metalabsImage}
+              source={chartLineImg}
+              resizeMode={'contain'}
+            />
+            <Text centered numberOfLines={1} style={styles.metalabsText}>
+              {strings('wallet.goToPortfolio')}
```
gantunesr, line 491:
```suggestion
              {strings('wallet. go_to_portfolio')}
```

app/components/UI/CollectibleContracts/index.js
```diff
@@ -80,6 +81,32 @@ const createStyles = (colors) =>
...
+    metalabsLink: {
+      width: '100%',
+      borderBottomColor: colors.grey100,
+      borderBottomWidth: 1,
+      borderTopWidth: 1,
+      borderTopColor: colors.grey100,
+      marginBottom: 16,
+    },
+    metalabsText: {
+      color: colors.blue,
```
gantunesr, line 93: I think it's better to use the colors of our DS to make this feature Dark Mode compatible

app/components/UI/Tokens/index.js
```diff
@@ -105,6 +108,31 @@ const createStyles = (colors) =>
...
+    metalabsLink: {
+      width: '100%',
+      borderBottomColor: colors.grey100,
+      borderBottomWidth: 1,
+      borderTopWidth: 1,
+      borderTopColor: colors.grey100,
+    },
+    metalabsText: {
+      color: colors.blue,
+      textAlign: 'center',
```
gantunesr, line 120: Same

> I think it's better to use the colors of our DS to make this feature Dark Mode compatible

gantunesr (commented):
Hi @meppsilon, the PR looks good, I left some small comments. Can you add a recording of the feature for our QA team in the PR description? Thanks!

app/components/UI/CollectibleContracts/index.js
meppsilon: how do I use the colors from the DS?

.js.env.example
meppsilon: Let me know if you'd like me to set this up.

gantunesr (commented):
LGTM. I'll send this to our QA team

meppsilon:
@gantunesr One more thing, there is some push back on the designs of this feature so I might need to make some minor updates to the UI before we can merge. Functionality will remain the same.

cortisiko (changes requested):
hey @meppsilon ! A couple of things I noticed:


- I noticed when I tap on the portfolio link on the wallet view, the browser page [opens to the metamask homepage and nothing else](http://recordit.co/vPPO3G5EIE). Yes, I did add the `METALABS_URL` to my js.env file. 


- While on the tokens tab, [there is a missing translation for the link text ](https://www.screencast.com/t/gqXCXWAH9N )


- When I try to access the portfolio dapp while on the NFT’s tab I get: `TypeError: undefined is not an object (evaluating '_this.props.navigation')` See [recording](http://recordit.co/RB3iEPLm0i)

sethkfman:
@meppsilon Do you know the status of this ticket?

meppsilon:
@sethkfman It still under design review. I'm going to go ahead and close this PR for now because I think it is going to take some time before we release this feature.

PR #4305, opened 2022-05-13
Hide remember me

owencraston:
**Description**
- Remove remember me as an option for new users
- The key here is `new users`. Users who have already enabled remember me should not be affected by this change.
- This simply hides the front end that allows for remember me to be toggled on.
- Based on the feedback below, remember me will be something the user can turn on in settings and is now intended more for developers
- I also made a `LoginOptionsSwitch` that handles if what login option should be rendered into its own component since it was repeated in several files
- biometrics is always going to be the higher priority option for login. we will only show remember me as an option if its enabled in settings and biometrics are disabled

**Future improvements**
- Once the [auth refactor ships](https://github.com/MetaMask/metamask-mobile/pull/3755), we can move ALL of the biometric logic (state on weather to show biometric or not, handle errors etch) into this component. 
- The main blocker from doing this now is this [handleRejectedOsBiometricPrompt](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/ChoosePassword/index.js#L457) which calls [updateBiometryChoice](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/ChoosePassword/index.js#L557)
- the updateBiometryChoice function is repeated in each of these files

**Checklist**

* [x] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

**https://github.com/MetaMask/metamask-mobile/issues/4194**
Progresses https://github.com/MetaMask/metamask-mobile/issues/4194

PR #3875, opened 2022-03-09
Add authentication type to login metrics

owencraston:
**Description**

- Add login method as a metric we are tracking on login
- This is a precursor to removing `remember me` as we need track how many clients will be impacted

**🎩 Does it work?**

Screen Shot 2022-03-14 at 4 24 41 PM
Screen Shot 2022-03-14 at 4 23 12 PM

As you can see this property has been successfully [added in mix panel](https://mixpanel.com/project/1979229/view/3052429/app/lexicon#transformations/profile-properties/user-profiles/Authentication%20Type)

**Checklist**

* [x] There is a related GitHub issue
* [ ] Tests are included if applicable
* [x] Any added code is fully documented

**Issue**

Resolves https://github.com/MetaMask/mobile-planning/issues/170

PR #4824, opened 2022-08-09
[Fix] Fix password not being set as authentication type for login metrics

owencraston:
**Description**
- this issue builds off the work done in the [initial setup PR for login metrics](https://github.com/MetaMask/metamask-mobile/pull/3875)
- @plasmacorral discovered this bug while doing for https://github.com/MetaMask/metamask-mobile/pull/4305
- The authentication type analytics property of the user was never being set to password even if the user was using password as the method to login
- this side effect of this was that when a user disabled remember me, the data would not reflect that since it would fall back to password login
- call `Analytics.applyUserProperty(AUTHENTICATION_TYPE.PASSWORD);` inside the `SecureKeychain.resetGenericPassword` method
- why?
    - because this is the method we use to "reset" the login credentials to its default state...which is password
    - this is how we turn off remember me
    - this is how we turn off biometrics
    - this is what is called when you lock the wallet
    - placing the logic in here means we will never miss a case where the user is using password as their login option

**Issue**

Progresses #???

**Checklist**

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -160,8 +160,8 @@ android {
 		applicationId "io.metamask"
 		minSdkVersion rootProject.ext.minSdkVersion
 		targetSdkVersion rootProject.ext.targetSdkVersion
-		versionCode 939
-		versionName "5.5.0"
+		versionCode 941
+		versionName "5.5.1"
 		multiDexEnabled true
 		testBuildType System.getProperty('testBuildType', 'debug')
 		missingDimensionStrategy "minReactNative", "minReactNative46"
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -160,8 +160,8 @@ android {
 		applicationId "io.metamask"
 		minSdkVersion rootProject.ext.minSdkVersion
 		targetSdkVersion rootProject.ext.targetSdkVersion
-		versionCode 941
-		versionName "5.5.1"
+		versionCode 939
+		versionName "5.5.0"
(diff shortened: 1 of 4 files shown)


android/app/build.gradle
```diff
@@ -160,8 +160,8 @@ android {
 		applicationId "io.metamask"
 		minSdkVersion rootProject.ext.minSdkVersion
 		targetSdkVersion rootProject.ext.targetSdkVersion
-		versionCode 939
```
owencraston, line 163: must be removed before merging 🔥

bitrise.yml
```diff
@@ -323,10 +323,10 @@ app:
           PROJECT_LOCATION_IOS: ios
         - opts:
               is_expand: false
-          VERSION_NAME: 5.5.0
```
owencraston, line 326: must be removed before merging 🔥

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -825,7 +825,7 @@
 				CODE_SIGN_ENTITLEMENTS = MetaMask/MetaMaskDebug.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 939;
```
owencraston, line 828: must be removed before merging 🔥

sethkfman (approved):
LGTM

Issue #4814, opened 2022-08-08
Metamask displays wrong icon for transfer type transaction when changing to any language other than english


anaamolnar:
**Describe the bug**
When using Metamask with any language other than english, the icon corresponding to a transfer transaction gets replaced with the swaps icon.

**Screenshots**
![image](https://user-images.githubusercontent.com/82837486/183420674-68046654-a538-45a3-88f5-86f91b48a385.png)
![image](https://user-images.githubusercontent.com/82837486/183420718-ede5fbf8-919e-4602-8436-a4d59c452230.png)

**To Reproduce**
Steps to reproduce:
1. Perform a transfer (if you do not have an existent one in your transaction history)
2. Go to Settings->General
3. Change the language from English to another language of your choice
4. See the icon next to the transfer changes from the correct one the corresponding icon for swaps

**Expected behavior**
Icons should show correctly, depending on the transaction type no matter the language.

**Smartphone (please complete the following information):**
 - OS: Android
 - App Version 5.4.0

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

Devmenai:
> **Describe the bug**
> When using Metamask with any language other than english, the icon corresponding to a transfer transaction gets replaced with the swaps icon.
> 
> **Screenshots**
> ![image](https://user-images.githubusercontent.com/82837486/183420674-68046654-a538-45a3-88f5-86f91b48a385.png)
> ![image](https://user-images.githubusercontent.com/82837486/183420718-ede5fbf8-919e-4602-8436-a4d59c452230.png)
> 
> 
> **To Reproduce**
> Steps to reproduce:
> 1. Perform a transfer (if you do not have an existent one in your transaction history)
> 2. Go to Settings->General
> 3. Change the language from English to another language of your choice
> 4. See the icon next to the transfer changes from the correct one the corresponding icon for swaps
> 
> **Expected behavior**
> Icons should show correctly, depending on the transaction type no matter the language.
> 
> **Smartphone (please complete the following information):**
>  - OS: Android
>  - App Version 5.4.0
> 
> 
> -------------------------------------------------------------
> _to be added after bug submission by internal support / PM_
> **Severity**
> 
> 
> - How critical is the impact of this bug on a user?
> - Add stats if available on % of customers impacted
> - Is this visible to all users?
[...]

PR #4832, opened 2022-08-10
Sent transaction icon changes incorrectly when the language changes

tommasini:
**Description**
There was more than one issue, one of them was when we change the language, the icon of a transaction was not consistent, a sent icon was changed for one swipe icon. 
Also, we were comparing translations, and when we change the language, some of them stayed on the past language, because the component would re-render, but some util function was compiling faster than the new language rendered.

**Proposed Solution**
Comparing the action key and not the action key translation.
 [Technical] - Refactored logic for we do not compare translations anymore.

**Code Impact**
Low/Medium - Will affect every info of transactions, on TransactionsView and Asset.

**Test Cases**
Case 1:
 * Switch to Polygon Mainnet
 * Do a swap to USDT
 * Open Activity View
 * Open Asset View
 * Change language to Spanish
 * The icon and the translation should mean the same
Case 2:
 * Switch to Polygon Mainnet
 * Send a token to yourself 
 * Open Activity View
 * Open Asset View
 * Change language to Spanish
 * The icon and the translation should mean the same
Case 3:
 * Switch to Polygon Mainnet
 * Send MATIC to another wallet 
 * Open Activity View
 * Open Asset View
 * Change language to Spanish
 * The icon and the translation should mean the same
Case 4:
 * Switch to Polygon Mainnet
 * Send another token (Like USDT) to yourself and to another wallet
 * Open Activity View
 * Open Asset View
 * Change language to Spanish
 * The icon and the translation should mean the same
 Case 5:
 * Switch to rinkeby
 * Import a NFT
 * Send it to other wallet
 * Open Activity View
 * Change language to Spanish
 * The icon and the translation should mean the same

**Some expected English translations of transactions**
image

**Screenshots/Recordings**
https://recordit.co/eQPcABRTaA
<div display:'inline'>

</div>
<div display:'inline'>

</div>

**Issue**
https://github.com/MetaMask/metamask-mobile/issues/4814
Progresses #4184 

**Checklist**


app/components/UI/TransactionElement/utils.js
```diff
@@ -600,14 +600,23 @@ function decodeConfirmTx(args) {
     symbol = tokenList[renderTo].symbol;
   }
   let transactionType;
-  if (actionKey === strings('transactions.approve'))
+
+  const actionKeyLowerCase = actionKey.toLowerCase();
+
+  if (actionKeyLowerCase === strings('transactions.approve').toLowerCase())
```
blackdevelopa, line 606: Can we have `transactions.approve` in lowercase instead?

blackdevelopa (dismissed):

cortisiko (changes requested):
@tommasini it seems the issue still persists. I was able to reproduce by:

- heading to BSC
- performing a swap
- then I sent some BEP-20 tokens
- upon doing so, I viewed the transactions activity and notice the swaps logo is next to the send transaction

See transaction activity [here](https://www.screencast.com/t/rlsGYvwr0)
asset overview [here](https://www.screencast.com/t/Hw48DYiOhe)

tommasini:
I'm closing this PR since it's too old

PR #4860, opened 2022-08-15
Change skip to payment method in on-ramp flow

wachunei:
**Description**

This screen changes the behavior of skipping screen after the Fiat On Ramp flow has started. 
- Previous behavior was to skip directly to Amount to Buy screen
- New behavior is to skip to Payment Methods screen.

**Screenshots/Recordings**
Before:

https://user-images.githubusercontent.com/1024246/184681297-9efb9b9a-4da0-4475-be62-8cd596104350.mp4

After: 

https://user-images.githubusercontent.com/1024246/184681314-465609c9-8f33-43e9-b07d-3f338b4599f6.mp4


app/constants/navigation/Routes.ts
```diff
@@ -6,6 +6,7 @@ const Routes = {
     ID: 'FiatOnRampAggregator',
     GET_STARTED: 'GetStarted',
     PAYMENT_METHOD: 'PaymentMethod',
+    PAYMENT_METHOD_HAS_STARTED: 'PaymentMethodHasStarted',
     AMOUNT_TO_BUY: 'AmountToBuy',
     AMOUNT_TO_BUY_HAS_STARTED: 'AmountToBuyHasStarted',
```
alaahd, line 11: Since we are updating the entry point for returning users, do we need to drop `AMOUNT_TO_BUY_HAS_STARTED` if it not being used any where?

alaahd (changes requested):
LGTM, left a small comment!

app/constants/navigation/Routes.ts
wachunei: Yes, I'll remove it

wachunei: Addressed in e91039fb7

alaahd (approved):

gantunesr (approved):

Issue #4811, opened 2022-08-05
[Force upgrade] Fetch app config from github pages endpoint and marshal it accordingly


owencraston:
# Description
- fetch the AppConfig from the Github pages endpoint and marshal it to a useable typescript interface
- this involves
    - defining api fetching best practices for typescript
    - deciding if we need a library for this, if so what library
### Builds off of 
- https://github.com/MetaMask/metamask-mobile/pull/4483/files
- https://github.com/MetaMask/metamask-mobile/issues/4810

#Situation
- not user facing

**Acceptance Criteria**
- when the app loads, it fetches the object and converts it into a useable interface within the app

**Process** (adapt per story)
- [ ] User research / data
- [ ] Ideation / Kickoff in cross-functional team
- [ ] Pattern research in mobile (crypto / non-crypto space)
- [ ] Proof of concept to check technical feasibility (if applicable)
- [ ] Design reviews with PM, Engineer(s)
- [ ] User testing
- [ ] Design - Dev handoff
- [x] Implementation
- [x] Code review
- [ ] Stakeholder review on latest build on iOS and Android
- [ ] QA - unit test, e2e tests, regression testing

PR #4862, opened 2022-08-15
[Force upgrade] Fetch and parse app config

owencraston:
**Description**

- fetches minimum versions from https://metamask.github.io/metamask-mobile/AppConfig/v1/AppConfig.json
- parses into useable typescript object
- this will allow us to query the minimum versions from the main app and trigger the upgrade warning
- adding the FetchState interfaces means we can reuse this same interface for any view that is waiting for a response from an api
- this follows a similar pattern as [MVVM](https://www.techtarget.com/whatis/definition/Model-View-ViewModel#:~:text=Model%2DView%2DViewModel%20(MVVM)%20is%20a%20software%20design,Ken%20Cooper%20and%20John%20Gossman.)

#### Testing
- to test paste this code into a component
```jsx
  const appConfig = useAppConfig();
  useEffect(() => {
    if (appConfig.type === 'Success') {
      console.log('app config response', appConfig.data);
    } else {
      console.log({ appConfig });
    }
  });
```

**Screenshots/Recordings**
This is what is hosted on the site
Screen Shot 2022-08-15 at 2 04 19 PM

temporarily logged the response in the `app/components/Nav/App/index.js` file and got sane values back
Screen Shot 2022-08-15 at 1 53 11 PM

**Issue**

Progresses https://github.com/MetaMask/metamask-mobile/issues/4811

**Checklist**

### DIFF
--- a/app/components/hooks/AppConfig/AppConfig.ts
+++ b/app/components/hooks/AppConfig/AppConfig.ts
@@ -1,0 +1,13 @@
+interface MinimumVersions {
+  appMinimumBuild: number;
+  appleMinimumOS: number;
+  androidMinimumAPIVersion: number;
+}
+
+interface Security {
+  minimumVersions: MinimumVersions;
+}
+
+export default interface AppConfig {
+  security: Security;
+}
--- a/app/components/hooks/AppConfig/AppConfig.ts
+++ b/app/components/hooks/AppConfig/AppConfig.ts
@@ -1,3 +1,4 @@
+// These are the minimum app, android, and iOS versions that we support for security
 interface MinimumVersions {
   appMinimumBuild: number;
   appleMinimumOS: number;
@@ -7,7 +8,10 @@ interface MinimumVersions {
 interface Security {
   minimumVersions: MinimumVersions;
 }
-
+/*
+  More information on this interface can be found here: https://github.com/MetaMask/metamask-mobile/tree/gh-pages
+  - this interface should match the this api: https://github.com/MetaMask/metamask-mobile/tree/gh-pages#app-config-api
+*/
 export default interface AppConfig {
   security: Security;
 }
--- a/app/components/hooks/AppConfig/useAppConfig.test.tsx
+++ b/app/components/hooks/AppConfig/useAppConfig.test.tsx
@@ -1,0 +1,32 @@
+import { renderHook } from '@testing-library/react-hooks';
+import AppConfig from './AppConfig';
+import useAppConfig from './useAppConfig';
+
+describe('useAppConfig', () => {
+  test('it should start with a state of "Loading"', () => {
+    const { result } = renderHook(() => useAppConfig());
+    expect(result.current.type).toEqual('Loading');
+  });
+
+  test('it should return a state of "Success"', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useAppConfig());
+    expect(result.current.type).toBe('Loading');
+    await waitForNextUpdate();
+    expect(result.all[1].type).toEqual('Success');
+  });
+  test('after a successful fetch, the AppConfig data should be available', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useAppConfig());
+    const expected: AppConfig = {
+      security: {
+        minimumVersions: {
+          appMinimumBuild: 700,
+          appleMinimumOS: 6,
+          androidMinimumAPIVersion: 21,
+        },
+      },
+    };
+    await waitForNextUpdate();
+    expect(result.all[1].type).toEqual('Success');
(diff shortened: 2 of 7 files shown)


app/components/hooks/AppConfig/AppConfig.ts
```diff
@@ -0,0 +1,13 @@
+interface MinimumVersions {
+  appMinimumBuild: number;
```
sethkfman, line 2: This is looking good. Would you be able to add comments just incase we have new people who need a bit more explanation.

app/components/hooks/AppConfig/useAppConfig.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { FetchState } from '../../../util/fetch/FetchState';
+
+type State = FetchState<AppConfig>;
+const initialState: Readonly<State> = {
+  type: 'Loading',
+};
+
+const useAppConfig = (): State => {
+  const appConfigURL =
+    'https://metamask.github.io/metamask-mobile/AppConfig/v1/AppConfig.json';
```
sethkfman, line 12: Move this to the url constants file.

app/components/hooks/AppConfig/useAppConfig.tsx
```diff
@@ -0,0 +1,56 @@
...
+        } catch (e: any) {
+          setState({
+            type: 'Error',
+            error: e,
+            message: `error parsing AppConfig ${e.message}`,
+          });
+        }
+      })
+      .catch((e: any) => {
+        setState({
```
sethkfman, line 41: When we get an error what is the plan at the UI level? Will it fail silent?

app/components/hooks/AppConfig/AppConfig.ts
```diff
@@ -0,0 +1,13 @@
...
+  androidMinimumAPIVersion: number;
+}
+
+interface Security {
+  minimumVersions: MinimumVersions;
+}
+
+export default interface AppConfig {
+  security: Security;
+}
```
gantunesr, line 13: The idea behind this is to add more interfaces other than `Security` to the `AppConfig`?

app/components/hooks/AppConfig/useAppConfig.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }, []);
+
+  useEffect(() => {
+    fetchAppConfig();
+  }, [fetchAppConfig]);
+
+  return state;
+};
+
+export default useAppConfig;
```
gantunesr, line 56: You can add unit test to this hooks using https://react-hooks-testing-library.com/

app/components/hooks/AppConfig/useAppConfig.tsx
owencraston: I am not sure, as you said I think the handling of errors is case by case as theoretically this could be used for a lot more data. For force upgrade I would imagine we would fail silently.

owencraston: @sethkfman do you think I should add extra logging here or will errors be covered by sentry automatically?

app/components/hooks/AppConfig/AppConfig.ts
owencraston: yeah, as this goes (maybe feature flags) it will be easy to add to without changing the existing interfaces.

app/components/hooks/AppConfig/useAppConfig.tsx
gantunesr: https://react-hooks-testing-library.com/ is already part of the project, it should be ok to use it in this case

owencraston: I added tests in this [file](https://github.com/MetaMask/metamask-mobile/pull/4862/files#diff-3c3eaa49c400725fa138752c59039b8e0973dbe1574b64df31a859e9d7671459R5)

app/components/hooks/AppConfig/useAppConfig.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    expect(result.current.type).toBe('Loading');
+    await waitForNextUpdate();
+    expect(result.all[1].type).toEqual('Success');
+  });
+  test('after a successful fetch, the AppConfig data should be available', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useAppConfig());
+    const expected: AppConfig = {
+      security: {
+        minimumVersions: {
+          appMinimumBuild: 700,
```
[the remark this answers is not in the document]
sethkfman: Love that idea

sethkfman (approved):
LGTM

gantunesr (approved):

PR #4924, opened 2022-08-26
[FIX] isTestNetwork not callable

tommasini:
**Description**
When the warning with not enough funds and we are on the Polygon Mainnet, will appear _Go To Faucet_ instead of _Buy Now_ to redirect to swaps
We also are always showing the Ticket instead of the native currency on the Transaction Review Information Modal

**Proposed Solution**
The function now is being called.

**Code Impact**
Low -> It only affect the transaction Review Information Modal UI

**Screenshots/Recordings**
error:

solution:

**Issue**

Progresses #???

**Checklist**


chrisleewilcox (approved):
LGTM

PR #4995, opened 2022-09-15
[FIX] Revert upgrades to correct testing step

gantunesr:
The testing step in the CI workflow is failing, this PR reverts the code that causes the issue meanwhile it's investigated the cause of it.


tommasini (approved):
LGTM!

PR #5024, opened 2022-09-23
[FIX] updated app name for Bitrise release-to-store step

sethkfman:
**Description**

Update the release to store Bitrise step to look for the correct apk filename.


chrisleewilcox (approved):
LGTM

PR #5033, opened 2022-09-27
Update default iOS simulator

rickycodes:
> latest Xcode upgrade does not ship with iPhone 11 Pro which is our default iOS simulator, maybe it is time to upgrade it?


wachunei (approved):
Nice

PR #5061, opened 2022-09-28
Run tests in parallel

rickycodes:
Our test job has gotten too big and is taking too long!

recently we started using `macos-12` for the test job. this was based on an assumption that the container would have more resources and that indeed did work. however, there's a limitation on the amount of concurrent jobs we can run using that runner (5 as apposed to 20). see: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

instead what I've done now is split the tests into 4 so that they can be run in parallel.

The breakdown is as follows:

```
previously: 857 total
---
non components: 372 total
ui components: 136 total
View components: 56 total
Remaining components: 293 total
```
Now the longest running test is ~8m. This means that we should be saving ~7m on each `ci` run after this is merged. hooray!

There's also no concurrent limit of 5 now because I've removed the `macos-12` runner entirely.

behold:
image


blackdevelopa (approved):

sethkfman (approved):
LGTM

PR #5088, opened 2022-10-04
Add no payment methods screen

wachunei:
**Description**

This PR adds a UI for the edge case a region has no available payment methods so the user can reset the flow.

**Screenshots/Recordings**
simulator_screenshot_A8AC2678-27D7-423D-9605-CB9CE3D10F38


app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -119,6 +120,26 @@ const PaymentMethod = () => {
...
+    if (needsReset) {
+      navigation.reset({
+        routes: [{ name: Routes.FIAT_ON_RAMP_AGGREGATOR.REGION }],
+      });
+    } else {
+      navigation.goBack();
+    }
+  }, [
+    // @ts-expect-error navigation params error
+    params?.showBack,
```
owencraston, line 137: if you leveraged the `useParams` hooks you could get strong typing on params and avoid the null check. Check this [out as an example](https://github.com/MetaMask/metamask-mobile/blob/20946ebfa724768798d7f94bbbb18589a25bb76e/app/components/Views/BrowserUrlModal/BrowserUrlModal.tsx#L32)

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -119,6 +120,26 @@ const PaymentMethod = () => {
     [setSelectedPaymentMethodId, trackEvent],
   );
 
+  const handleResetState = useCallback(() => {
+    setSelectedRegion(null);
+    setSelectedPaymentMethodId(null);
+    // @ts-expect-error navigation params error
```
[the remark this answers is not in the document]
wachunei: Yeah I'm getting this Screen Shot 2022-10-27 at 16 19 27

Just saw your other comment

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
wachunei: Nice, I'm expecting this error in a few places. I'm definitely adopting that in a subsequent PR.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -185,61 +206,83 @@ const PaymentMethod = () => {
...
-                        : () => handlePaymentMethodPress(id)
-                    }
-                    amountTier={amountTier}
-                    paymentTypeIcon={getPaymentMethodIcon(paymentType)}
-                    logo={logo}
-                  />
-                </View>
-              ),
+        {filteredPaymentMethods?.length === 0 ? (
+          <ErrorView
```
owencraston, line 210: It would be nice if we could components these views, it would make the switch a lot cleaner to read.

wachunei: this is already:
```tsx
{filteredPaymentMethods?.length === 0 ?
  <ErrorView ... />
  : 
  // the other case
}
```
What do you think could be better?

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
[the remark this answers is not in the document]
owencraston: I suspect that this error could be mitigated by using the useParams as well but I am not certain

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
owencraston: all I am advocating for here is that we move these subviews into their own files and name them accordingly. The your switch would look something like this....

```
{filteredPaymentMethods?.length === 0 ? 
    <NoPaymentsAvailable region={selectedRegion?.name} /> 
    : <PaymentMethods /> }
```

Then it would be much easier to test these cases...you could simply mock the `filteredPaymentMethods` and verify that `NoPaymentsAvailable` is/is not rendered.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -185,61 +206,83 @@ const PaymentMethod = () => {
...
+        )}
       </ScreenLayout.Body>
-      <ScreenLayout.Footer>
-        <ScreenLayout.Content>
-          {(currentPaymentMethod?.paymentType === PaymentType.ApplePay ||
-            currentPaymentMethod?.paymentType ===
-              PaymentType.DebitCreditCard) && (
+
+      {filteredPaymentMethods && filteredPaymentMethods.length > 0 && (
+        <ScreenLayout.Footer>
```
owencraston, line 255: same comment as above, these lines should be brought into their own sub component files to separate the concerns. This is clearly the parent screen where the branching logic lives so it would be best to make that aspect stand out as much as possible and only if a dev cares about a specific view will they need to dive deeper into the views.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -185,61 +206,83 @@ const PaymentMethod = () => {
...
+          </ScrollView>
+        )}
       </ScreenLayout.Body>
-      <ScreenLayout.Footer>
-        <ScreenLayout.Content>
-          {(currentPaymentMethod?.paymentType === PaymentType.ApplePay ||
-            currentPaymentMethod?.paymentType ===
-              PaymentType.DebitCreditCard) && (
+
+      {filteredPaymentMethods && filteredPaymentMethods.length > 0 && (
```
owencraston, line 254: we should add tests for this conditional.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -185,61 +206,83 @@ const PaymentMethod = () => {
...
-                        ? undefined
-                        : () => handlePaymentMethodPress(id)
-                    }
-                    amountTier={amountTier}
-                    paymentTypeIcon={getPaymentMethodIcon(paymentType)}
-                    logo={logo}
-                  />
-                </View>
-              ),
+        {filteredPaymentMethods?.length === 0 ? (
```
owencraston, line 209: can we test this conditional?

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
owencraston, line 254: Nit pick: we should use a ternary here that explicitly returns null vs using the logic AND operator. Read more [here on why](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx). Like I said though, its not major/blocking.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
```diff
@@ -185,61 +206,83 @@ const PaymentMethod = () => {
...
-              <Text small grey centered>
-                {currentPaymentMethod?.paymentType === PaymentType.ApplePay &&
-                  strings(
-                    'fiat_on_ramp_aggregator.payment_method.apple_cash_not_supported',
-                  )}
-                {currentPaymentMethod?.paymentType ===
-                  PaymentType.DebitCreditCard &&
-                  strings('fiat_on_ramp_aggregator.payment_method.card_fees')}
-              </Text>
+              <StyledButton
```
owencraston, line 273: you should try using the latest and greatest shared button components from the [component library](https://github.com/MetaMask/metamask-mobile/tree/main/app/component-library/components/Buttons/Button). They are better maintained and use the latest design system styles which will safe us time updating this button in the future. The API is also pretty clean and they have several [variants](https://github.com/MetaMask/metamask-mobile/tree/main/app/component-library/components/Buttons/Button/variants) that should fit your use case.

app/components/UI/FiatOnRampAggregator/Views/PaymentMethod.tsx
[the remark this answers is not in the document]
wachunei: I will adopt useParams later.

owencraston (approved):
I am approving this under the assumption that the requested changes will occur in future changes. It would be great to get some tests on these!

PR #5129, opened 2022-10-19
FIX: Use search on an URL object to replace an undefined query

buynao:
**Description**

A syntax error change, there is currently no `query` field on the URL object, so I think this should refer to the `search` field.

**Code Impact**

Low

**Screenshots/Recordings**

This URL API doc:
[https://developer.mozilla.org/en-US/docs/Web/API/URL](https://developer.mozilla.org/en-US/docs/Web/API/URL)


buynao:

PR #5138, opened 2022-10-22
Import wallet srp password validation

chrisleewilcox:
**Development & PR Process**

**Description**

**Screenshots/Recordings**

**Issue**

Progresses #???

**Checklist**

### DIFF
--- a/app/components/Views/ImportFromSeed/index.js
+++ b/app/components/Views/ImportFromSeed/index.js
@@ -62,6 +62,8 @@ import {
   IMPORT_FROM_SEED_SCREEN_SEED_PHRASE_INPUT_ID,
   IMPORT_FROM_SEED_SCREEN_SUBMIT_BUTTON_ID,
   IMPORT_FROM_SEED_SCREEN_TITLE_ID,
+  IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
+  IMPORT_FROM_SEED_SCREEN_PASSWORD_STRENGTH_ID,
 } from '../../../../wdio/features/testIDs/Screens/ImportFromSeedScreen.testIds';
 import { IMPORT_PASSWORD_CONTAINER_ID } from '../../../constants/test-ids';
 
@@ -636,11 +638,11 @@ class ImportFromSeed extends PureComponent {
               <OutlinedTextField
                 style={styles.input}
                 containerStyle={inputWidth}
-                // {...generateTestId(
-                //   Platform,
-                //   IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
-                // )}
-                testID={'create-password-first-input-field'}
+                {...generateTestId(
+                  Platform,
+                  IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
+                )}
+                //testID={'create-password-first-input-field'}
                 placeholder={strings('import_from_seed.new_password')}
                 placeholderTextColor={colors.text.muted}
                 returnKeyType={'next'}
@@ -655,7 +657,13 @@ class ImportFromSeed extends PureComponent {
               />
 
               {(password !== '' && (
-                <Text style={styles.passwordStrengthLabel}>
+                <Text
+                  style={styles.passwordStrengthLabel}
+                  {...generateTestId(
+                    Platform,
+                    IMPORT_FROM_SEED_SCREEN_PASSWORD_STRENGTH_ID,
+                  )}
+                >
                   {strings('choose_password.password_strength')}
                   <Text style={styles[`strength_${passwordStrengthWord}`]}>
                     {' '}
--- a/app/components/Views/ImportFromSeed/index.js
+++ b/app/components/Views/ImportFromSeed/index.js
@@ -642,7 +642,7 @@ class ImportFromSeed extends PureComponent {
                   Platform,
                   IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
                 )}
-                //testID={'create-password-first-input-field'}
+                testID={'create-password-first-input-field'}
                 placeholder={strings('import_from_seed.new_password')}
                 placeholderTextColor={colors.text.muted}
                 returnKeyType={'next'}
--- a/wdio/features/screen-objects/ImportFromSeedScreen.js
+++ b/wdio/features/screen-objects/ImportFromSeedScreen.js
@@ -59,9 +59,14 @@ class ImportFromSeed {
 
   async assertPasswordStrength(text){
     const elem = await this.passwordStrength;
-    const strength = await elem.getText();
-    console.log(`## Password strength is: ${strength}`);
-    assert.strictEqual(strength, text);      
+    const passwordStrengthText = await elem.getText();
+    assert.strictEqual(passwordStrengthText, text);      
+  }
+
+  async assertDeviceAlert(text){
+    const msg = await driver.getAlertText();
+    assert(msg.includes(text));
+    await driver.acceptAlert();  
   }
 }
 
--- a/wdio/features/screen-objects/ImportFromSeedScreen.js
+++ b/wdio/features/screen-objects/ImportFromSeedScreen.js
@@ -4,9 +4,11 @@ import {
   IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
   IMPORT_FROM_SEED_SCREEN_CONFIRM_PASSWORD_INPUT_ID,
   IMPORT_FROM_SEED_SCREEN_SUBMIT_BUTTON_ID,
+  IMPORT_FROM_SEED_SCREEN_PASSWORD_STRENGTH_ID
 } from '../testIDs/Screens/ImportFromSeedScreen.testIds';
 import Selectors from '../helpers/Selectors';
 import Gestures from '../helpers/Gestures';
+import assert from 'assert';
 
 class ImportFromSeed {
 
@@ -30,6 +32,10 @@ class ImportFromSeed {
     return Selectors.getElementByPlatform(IMPORT_FROM_SEED_SCREEN_SUBMIT_BUTTON_ID);
   }
 
+  get passwordStrength(){
+    return Selectors.getElementByPlatform(IMPORT_FROM_SEED_SCREEN_PASSWORD_STRENGTH_ID);
+  }
+
   async verifyScreenTitle() {
     await expect(this.screenTitle).toBeDisplayed();
   }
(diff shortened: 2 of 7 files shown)


wdio/features/step-definitions/onboarding.js
```diff
@@ -100,22 +102,20 @@ When(/^I type (.*) in confirm password field/, async (text) => {
...
-      break;
-    case 'a valid Confirm password':
-      await ImportFromSeedScreen.typeConfirmPassword(validAccount.password);
-      break;
-    default:
-      throw new Error('Condition not found');
-  }
-})
+Then(/^device alert (.*) is displayed/, async (text) => {
+  const msg = await driver.getAlertText();
```
cortisiko, line 106: line 106 to line 109 should be an assertion method that lives in the ImportWallet page object.

wdio/features/step-definitions/onboarding.js
```diff
@@ -41,6 +42,7 @@ Then(/^"([^"]*)?" is displayed/, async (text) => {
       await ImportFromSeedScreen.verifyScreenTitle();
       break;
     case 'Welcome to your new wallet!':
+      // await driver.pause(10000);
```
cortisiko, line 45: lets not leave comments in the code unless absolutely necessary.

app/components/Views/ImportFromSeed/index.js
```diff
@@ -636,11 +638,11 @@ class ImportFromSeed extends PureComponent {
...
-                // {...generateTestId(
-                //   Platform,
-                //   IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
-                // )}
-                testID={'create-password-first-input-field'}
+                {...generateTestId(
+                  Platform,
+                  IMPORT_FROM_SEED_SCREEN_NEW_PASSWORD_INPUT_ID,
+                )}
+                //testID={'create-password-first-input-field'}
```
cortisiko, line 645: let's not leave comments in the code

wdio/features/screen-objects/WalletMainScreen.js
```diff
@@ -3,7 +3,7 @@ import { ONBOARDING_WIZARD_STEP_1_CONTAINER_ID } from '../testIDs/Components/Onb
 class WalletMainScreen {
   async validateOnboardingWizard() {
     const elem = await $(`~${ONBOARDING_WIZARD_STEP_1_CONTAINER_ID}`);
-    expect(elem).toBeDisplayed();
+    await expect(elem).toBeDisplayed();
```
cortisiko, line 6: Create a getter for the object then assert that the object is visible. Something like:


```suggestion

  get onBoardingWizard() {
    return Selectors.getElementByPlatform(ONBOARDING_WIZARD_STEP_1_CONTAINER_ID);
  }

  async isOnboardingWizardVisible() {
    await expect(this.onBoardingWizard).toBeDisplayed();
  }
```

wdio/features/screen-objects/ImportFromSeedScreen.js
```diff
@@ -50,6 +56,13 @@ class ImportFromSeed {
     await Gestures.waitAndTap(this.screenTitle);
     await Gestures.waitAndTap(this.importButton);
   }
+
+  async assertPasswordStrength(text){
+    const elem = await this.passwordStrength;
+    const strength = await elem.getText();
+    console.log(`## Password strength is: ${strength}`);
```
cortisiko, line 63: Was this for debugging purposes? Do not keep console log messages in the code.

wdio/features/screen-objects/ImportFromSeedScreen.js
```diff
@@ -50,6 +56,13 @@ class ImportFromSeed {
     await Gestures.waitAndTap(this.screenTitle);
     await Gestures.waitAndTap(this.importButton);
   }
+
+  async assertPasswordStrength(text){
+    const elem = await this.passwordStrength;
```
cortisiko, line 61: this method seems like it is doing a lot. You are creating a variable for the password element then creating another variable to get the string of the password element. Why not do:

`const passwordStrengthText = await this.passwordStrength.getText();`

wdio/features/screen-objects/ImportFromSeedScreen.js
```diff
@@ -50,6 +56,13 @@ class ImportFromSeed {
...
+    const strength = await elem.getText();
+    console.log(`## Password strength is: ${strength}`);
+    assert.strictEqual(strength, text);
```
cortisiko, line 64: any reason why we are not using webdriverio built-in assertions?

cortisiko (changes requested):
see comments

app/components/Views/ImportFromSeed/index.js
chrisleewilcox: This is a detox id that is commented out.  Are you sure you want this removed?

wdio/features/screen-objects/ImportFromSeedScreen.js
chrisleewilcox: Done.

wdio/features/screen-objects/ImportFromSeedScreen.js
chrisleewilcox: Removed.

wdio/features/screen-objects/ImportFromSeedScreen.js
chrisleewilcox: Any reason why we want to only use webdriverio assertions?  No need for us to be limited with assertion capabilities.

wdio/features/screen-objects/WalletMainScreen.js
chrisleewilcox: This file was implemented by either Yande or Fatia.  I just made a simple fix by adding await in the return.  I would actually like this file to be renamed to "TakeATourComponent.js" or something like that and probably best to do maintenance work on another PR.

wdio/features/step-definitions/onboarding.js
chrisleewilcox: Removed.

wdio/features/step-definitions/onboarding.js
chrisleewilcox: Done.

app/components/Views/ImportFromSeed/index.js
cortisiko: Is there a reason why you commented it out?

wdio/features/screen-objects/WalletMainScreen.js
cortisiko: That is fair.

wdio/features/screen-objects/ImportFromSeedScreen.js
cortisiko: mmmm okay! Your function still does a lot for an assertion. 

how about splitting that function into two? One that gets the password strength text and the other for the assertion.

cortisiko:
```
async PasswordStrengthText(){     
    const passwordStrengthTextElement = await this.passwordStrength;     
    const strength = await passwordStrengthTextElement.getText();      
    return strength;   
  }  
```

```
  async assertPasswordStrength(text){     
    assert.strictEqual(await this.PasswordStrengthText(), text);       
    }
```

The assertion function does what it is set to do.

wdio/features/screen-objects/ImportFromSeedScreen.js
```diff
@@ -50,6 +56,18 @@ class ImportFromSeed {
...
     await Gestures.waitAndTap(this.importButton);
   }
+
+  async assertPasswordStrength(text){
+    const elem = await this.passwordStrength;
+    const passwordStrengthText = await elem.getText();
+    assert.strictEqual(passwordStrengthText, text);      
+  }
+
+  async assertDeviceAlert(text){
```
cortisiko, line 66: This function should be split up into two. The action "tapping(accepting)" the alert should be in its function.

wdio/features/screen-objects/ImportFromSeedScreen.js
cortisiko: I recommend using wdio assertions. In this case [toHaveText](https://webdriver.io/docs/api/expect-webdriverio/#tohavetext). I do agree with not being limited by assertions but I also believe it is important to have consistency. The question would be: when to use assert vs wdio assertions.

app/components/Views/ImportFromSeed/index.js
chrisleewilcox: If I recall I think it was commented out when I created the branch.  I uncommented and doesn't look like that causes any issue.  Done.

wdio/features/screen-objects/ImportFromSeedScreen.js
chrisleewilcox: This ended up being just a oneliner change that inherently getText() and compares so I didn't have to break that out into multiple functions.  

`await expect(this.passwordStrengthLabel).toHaveText(text);`

I like this approach.  Done.

wdio/features/screen-objects/ImportFromSeedScreen.js
chrisleewilcox: Do we want to have a separate class like `/helpers/Device.js` and put device control functions there?  Can we discuss and address in another PR?

wdio/features/screen-objects/ImportFromSeedScreen.js
chrisleewilcox: Done.  Related to below change.

wdio/features/screen-objects/ImportFromSeedScreen.js
cortisiko: Yeah but you do so in this PR. This way, we lay a good foundation to deal with alert boxes when writing test cases in the future. It is as simple as creating a new class that has a function that taps "Ok" in the alert message.

chrisleewilcox: So I went down the route of trying to click "OK" on an alert and it didn't work out.  Doesn't look like there is a way to get an element for a device alert.  I then resorted to using webdriverio support for device alerts which is limited to...
[dismissAlert](https://webdriver.io/docs/api/jsonwp/#dismissalert)
[acceptAlert](https://webdriver.io/docs/api/jsonwp/#acceptalert)
[getAlertText](https://webdriver.io/docs/api/jsonwp/#getalerttext)
[sendAlertText](https://webdriver.io/docs/api/jsonwp/#sendalerttext)

Sounds like we need a class for device alert support for all our test.  There will also need to be some thought put into how we bundle up buttons on a device for acceptAlert(Yes, OK, Agree, Allow, etc...) and dismissAlert(No, Cancel, Close, etc...).  With all these considerations and others, I think this is separate work.  I feel more comfortable if there are no further changes necessary at this point related to SRP and password strength to merge this PR in before adding device support.  We should be mindful of not letting code reviews force PR's to drag on and take on other work that may be better suited in a separate PR.  Adding device support to this PR is essentially code creep for automation.  I agree we need it but think it's best suited in another PR so we can move on from this one.
Let me know how you want to proceed.

cortisiko: Thank you for doing the necessary research. The purpose of the code review is to ensure we are making the code as clean and readable as possible not necessarily dragging on the PR.

cortisiko: I will ask Fatia or Yande to handle. Thanks!

cortisiko (approved):
Much better. Feel free to merge whenever @chrisleewilcox

PR #5145, opened 2022-10-25
Restructuring Methods and variables

cortisiko:
**Development & PR Process**

**Description**

I created folders for specific files, restructured files and renamed variables so it is easier to follow. 

**Screenshots/Recordings**

**Issue**

Progresses #???

**Checklist**


wdio/features/screen-objects/Modals/OnboardingWizardModal.js.js
```diff
@@ -0,0 +1,17 @@
...
+    return Selectors.getElementByPlatform(ONBOARDING_WIZARD_STEP_1_CONTAINER_ID);
+  }
+
+
+  async isVisible(){
+    await expect(this.onBoardingWizardContainer).toBeDisplayed();
+  }
+
+}
+export default new OnboardingWizardModal();
```
chrisleewilcox, line 17: Is this the take a tour modal?  I think this should be named TakeATourModal.js.

wdio/features/screen-objects/Onboarding/ImportFromSeedScreen.js
```diff
@@ -36,7 +36,7 @@ class ImportFromSeed {
     return Selectors.getElementByPlatform(IMPORT_FROM_SEED_SCREEN_PASSWORD_STRENGTH_ID);
   }
 
-  async verifyScreenTitle() {
+  async isScreenTitleVisible() {
```
chrisleewilcox, line 39: So I think this should be verifyScreenTitle() and we can add more to check to verify the title like text of the title.  You also have to think about if we are saying isScreenTitleVisisble() then I would assuming we should return a boolean but I think this is not the approach we want.  I would vote to change this back to verifyScreenTitle().

wdio/features/screen-objects/Onboarding/ImportFromSeedScreen.js
```diff
@@ -61,9 +61,12 @@ class ImportFromSeed {
     await expect(this.passwordStrengthLabel).toHaveText(text);  
   }
 
-  async assertDeviceAlertText(text){
+  async isAlertTextCorrect(text){
```
chrisleewilcox, line 64: We have no need to return a boolean so I think this should be maybe verifyDeviceAlertText().

wdio/features/screen-objects/Onboarding/ImportFromSeedScreen.js
```diff
@@ -61,9 +61,12 @@ class ImportFromSeed {
...
     const msg = await driver.getAlertText();
     assert(msg.includes(text));
+  }
+
+  async tapOkInAlertMessage(){
```
chrisleewilcox, line 69: So it may not be Ok that is in the alert to tap.  We may be getting ahead of ourselves by assuming too much.  I would leave as is for now.

wdio/features/screen-objects/Onboarding/OnboardingCarousel.js
```diff
@@ -22,7 +22,7 @@ class WelcomeScreen {
     return Selectors.getElementByPlatform(WELCOME_SCREEN_GET_STARTED_BUTTON_ID);
   }
 
-  async verifySplashScreen() {
+  async isSplashScreenVisible() {
```
chrisleewilcox, line 25: As mentioned functions that start with is should return a boolean and we have no need for that here.  I would keep as is.

wdio/features/screen-objects/WalletMainScreen.js
```diff
@@ -1,9 +1,14 @@
...
 class WalletMainScreen {
-  async validateOnboardingWizard() {
-    const elem = await $(`~${ONBOARDING_WIZARD_STEP_1_CONTAINER_ID}`);
-    await expect(elem).toBeDisplayed();
+
+  get WalletScreenContainer(){
+    return Selectors.getElementByPlatform(WALLET_CONTAINER_ID)
+    
+  }
+  async isVisible() {
```
chrisleewilcox, line 10: Let's discuss about how we want to verify an element.  I feel like we are reinventing the wheel unnecessarily by us checking if an element isVisibile().

wdio/features/step-definitions/onboarding.js
```diff
@@ -18,31 +23,32 @@ When(/^I launch MetaMask mobile app/, async () => {
...
-      await OptinMetricsScreen.verifyScreenTitle();
+      await MetaMetricsScreen.isScreenTitleVisible();
       break;
     case 'Import from seed':
-      await ImportFromSeedScreen.verifyScreenTitle();
+      await ImportFromSeedScreen.isScreenTitleVisible();
       break;
     case 'Welcome to your new wallet!':
-      await WalletMainScreen.validateOnboardingWizard();
+      await OnboardingWizardModal.isVisible();
```
chrisleewilcox, line 50: Having us check if something is visible is not the way.

wdio/features/screen-objects/Onboarding/ImportFromSeedScreen.js
cortisiko: If you think about it, Aren't assertions booleans? We are checking if something exists or doesn't exist. I am open to having a discussion.

wdio/features/screen-objects/Modals/OnboardingWizardModal.js.js
cortisiko: I am following the naming convention we have in the application. https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/OnboardingWizard/index.js

wdio/features/screen-objects/Onboarding/ImportFromSeedScreen.js
cortisiko: As you mentioned in a previous review, this would require a bit more work. For clarity's sake, I removed the test action you had in the assertion into its own method.

Andepande (approved):

chrisleewilcox (approved):
LGTM

PR #5158, opened 2022-10-27
On-Ramp: Provider payment method custom action and custom order ids

wachunei:
**Development & PR Process**

**Description**

This PRs includes the custom action for provider's payment method and the use of custom order ids.

~~**Screenshots/Recordings**~~

~~_If applicable, add screenshots and/or recordings to visualize the before and after of your change_~~

**Issue**

Progresses #???

**Checklist**


app/components/UI/FiatOnRampAggregator/containers/CustomActionButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+      const customIdData = createCustomOrderIdData(
+        customOrderId,
+        selectedChainId,
+        selectedAddress,
+      );
+
+      dispatch(addFiatCustomIdData(customIdData));
+
+      if (await InAppBrowser.isAvailable()) {
+        dispatch(setLockTime(-1));
```
alaahd, line 95: what is the main purpose of this line?

app/components/UI/FiatOnRampAggregator/containers/CustomActionButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   * * Handle custom action
+   */
+
+  const handleCustomAction = useCallback(async () => {
+    if (!sdk || !customAction) {
+      return;
+    }
+    const prevLockTime = lockTime;
```
alaahd, line 65: what is the main functionality of storing lockTime?

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * If the custom order had errors, we don't fetch it unless
+   * POLLING_FRECUENCY ^ (errorCount + 1) seconds has passed
+   */
+  if (
+    customOrderIdData.errorCount > 0 &&
+    customOrderIdData.lastTimeFetched +
+      Math.pow(POLLING_FREQUENCY / 1000, customOrderIdData.errorCount + 1) *
+        1000 >
+      now
```
alaahd, line 37: I find it hard to understand the logic here, I guess `processCustomOrderIdData` is the callback method for interval polling. the client will call this method every X amount of seconds and maybe we need to simplify it and make the logic leaner

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
```diff
@@ -0,0 +1,74 @@
...
+          lastTimeFetched: Date.now(),
+          errorCount: customOrderIdData.errorCount + 1,
+        },
+        null,
+      ];
+    } else if (updatedCustomOrderIdData.status === OrderStatusEnum.IdExpired) {
+      /** In this case the order is expired and will be removed from the list */
+      return [{ ...customOrderIdData, expired: true }, null];
+    } else if (updatedCustomOrderIdData.status !== OrderStatusEnum.Precreated) {
+      /** In this case the order is not error, expired or precreated, so it is an actual order */
```
alaahd, line 63: don't we have an explicit final state for actual order? something like `OrderStatusEnum.valid`

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
```diff
@@ -0,0 +1,74 @@
...
+      customOrderIdData.id,
+      customOrderIdData.account,
+    );
+
+    if (updatedCustomOrderIdData.status === OrderStatusEnum.Unknown) {
+      /** This represents an error, we update the error count and the last time fetched */
+      return [
+        {
+          ...customOrderIdData,
+          lastTimeFetched: Date.now(),
```
alaahd, line 54: I am just challenging the idea if we can offload the tracking logic of `lastTimeFetched` and `errorCount` on the favor of simplifying the logic. I feel this is adding some complexity. so imagine if we keep polling and fetching the updated state of all custom orders that are not yet finalized which is very similar to the approach we have in `processAggregatorOrder` one addition we can have is to create a standalone garbage collector responsible to settle orphan or stale orders.

app/components/UI/FiatOnRampAggregator/containers/CustomActionButton.tsx
wachunei: We store and then restore the `lockTime` while displaying system UI, this is the case for Apple Pay UI and InApp Browser.

This is done at the start and end of the system UI interaction to avoid locking the app while it is "behind" the system UI.

app/components/UI/FiatOnRampAggregator/containers/CustomActionButton.tsx
wachunei: check https://github.com/MetaMask/metamask-mobile/pull/5158#discussion_r1007962985

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
wachunei: This is the exponential backoff case.

When the custom orders have been facing errors, we update the `errorCount` property, because we want to start delaying the next request and only make it after a "delayed time".

In this case we basically want to the detect the case the delayed time has not passed yet, i.e. it is still in the "future".
In a timeline this would mean: `delay > now`.

The `delay` increases as the error count increases. Using `10` seconds as frecuency (`POLLING_FRECUENCY / 1000`):

- error count: 1
    - 10 ^ 2 s = 100 s = 100000 ms 
- error count: 2
    - 10 ^ 3 s = 1000 s = 1000000 ms
- errorCount: 3
    - `Math.pow(10, errorCount + 1) * 1000`
- In general: `Math.pow(POLLING_FRECUENCY / 1000, errorCount + 1) * 1000`

The `delay` time is counted from the last time this was fetched. 

So `delay > now` translates to `lastTimeFetched + Math.pow(POLLING_FRECUENCY / 1000, errorCount + 1) * 1000 > now`

In this case we don't perform any request and we just return the same object.

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
wachunei: An order at this point can be `PENDING`, `COMPLETED`, `CANCELLED` or `FAILED`. It will be added to the orders in any case.

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
wachunei: I did not get this very well. Can you clarify the case you are proposing and why?

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
alaahd: Ok I understand the exponential correlation with error count. but If I may ask why we need to delay fetching order status exponentially every time we encounter error?

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
alaahd: In that case I prefer we create a helper function to explicitly check for one of the valid 4 different states: 
```suggestion
      /** In this case the order is not error, expired or precreated, so it is an actual order */
      const isValid = (orderState) => 
           orderState === OrderStatusEnum.PENDING ||
           orderState === OrderStatusEnum.COMPLETED ||
           orderState === OrderStatusEnum.CANCELLED ||
           orderState === OrderStatusEnum.FAILED 
      else if (isValid(updatedCustomOrderIdData.status)) {
           return [customOrderIdData, updatedCustomOrderIdData];
      }
```

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
wachunei: So we don't saturate a server that is already failing.

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
alaahd: my question is simply what are the consequences if we remove tracking `errorCount` and `lastTimeFetched` and keep interval polling until we reach a valid final state! are we throttling fetching the updated state of the order to avoid a resource intensive operation on our backend and consequently on the provider side?

wachunei: Yes, we only do this for 5xx errors which mean the provider server is down and we should avoid saturating it, until is back online.

app/components/UI/FiatOnRampAggregator/orderProcessor/customOrderId.ts
wachunei: I can leave a comment saying that given the value of `OrderStatusEnum` these are the only cases possible, but makes no sense to check it programmatically, it is going to be always true.

app/components/UI/FiatOnRampAggregator/sdk/index.tsx
```diff
@@ -70,12 +70,15 @@ interface IProviderProps<T> {
 const isDevelopment = process.env.NODE_ENV !== 'production';
 const VERBOSE_SDK = isDevelopment;
 
+const ACTIVATION_KEYS = process.env.ONRAMP_ACTIVATION_KEYS?.split(',') || [];
+
 export const SDK = OnRampSdk.create(
-  isDevelopment ? Environment.Staging : Environment.Production,
+  isDevelopment ? Environment.Staging : Environment.Staging,
```
wachunei, line 76: Reminder to revert this.

bitrise.yml
```diff
@@ -426,10 +426,10 @@ app:
       PROJECT_LOCATION_IOS: ios
     - opts:
         is_expand: false
-      VERSION_NAME: 5.9.0
+      VERSION_NAME: 5.10.0
     - opts:
         is_expand: false
-      VERSION_NUMBER: 990
+      VERSION_NUMBER: 996
```
wachunei, line 432: Reminder to revert this as well

sethkfman (approved):
Minor comments

wachunei:
@SocketSecurity ignore fsevents@1.2.13 react-native@0.56.0 react-native@0.56.1 react-native@0.57.8 wordwrap@0.0.2

Issue #5128, opened 2022-10-18
ENS Resolved Incorrectly - Lost Funds on Transfer


pavlovcik:
# **Describe the bug**

I attempted to transfer funds from my gf wallet to mine (pavlovcik.eth) and when I entered pavlovcik.eth and clicked send right away it resolved to [0x000000000000000000000000000017f34c43bddb](https://etherscan.io/address/0x000000000000000000000000000017f34c43bddb) ?? 

Either it shouldn't resolve (so to 0x0) and ERC20 transfer would fail attempting to send to 0x0, or it should resolve to my wallet, as expected ([0x4007CE2083c7F3E18097aeB3A39bb8eC149a341d](https://etherscan.io/address/pavlovcik.eth))

I only realized after several minutes that I didn't receive the funds. I checked Etherscan of her wallet and tracked down the funds. Now I'm missing $672.61! 

Transactions:

- https://etherscan.io/tx/0x733a04e355b82d58d8e8bc6d1e8bd278c90c10de40567da70d82219b45846ed6
- https://etherscan.io/tx/0x7f10a69d10c35044a934f0f12aa941ecd172ca5e07e095738a872706f0bdd705

I'm livid because I've been using MetaMask since 2016 and ENS since 2017 never had anything like this happen to me. I need to be reimbursed for these. 

# **Screenshots**
_If applicable, add screenshots or links to help explain your problem_

Update: I tried again but this time from my phone. I didn't hit send, but in the recents, it shows this invalid `OxPAVLOVcIK.eth` address. There's some issue with the parser for the textbox input perhaps?

 ![IMG_2114](https://user-images.githubusercontent.com/4975670/196528455-9a354315-40b2-4b77-bf71-92de855b8540.PNG)

![IMG_2949326E1B86-1](https://user-images.githubusercontent.com/4975670/196529060-8f2d1234-7276-4efe-a9dc-c8b8dabc6e79.jpeg)

![IMG_2115](https://user-images.githubusercontent.com/4975670/196528447-9ee967b7-6b5a-4b32-a90b-ea8c21ee8965.PNG)

I'm able to constantly reproduce this. If you go very slowly, it resolves correctly. If you go at a normal speed, it resolves to an invalid address. I cropped the video because my keyboard doxxed a ton of people with their home addresses in my contacts.

https://user-images.githubusercontent.com/4975670/196533115-0d0311bb-722b-4564-88ae-6bce9baf2925.MP4

# **To Reproduce**

[...]

PR #5159, opened 2022-10-27
[HOTFIX] ENS name being resolved correctly

tommasini:
**Description**
When the user had a slower internet connection and it was really quick to press the confirm button, the button would be enabled but the address would not be ready yet. Because of that we only proceeded to the next screen with the ENS.

**Proposed Solution**
Now we ensure that we have on that screen the address of the ens to go to the amount screen.

**Technical**
This was due to a delay in the component state on the validation function that happens on pressing the confirm button. This hotfix's a solution but it's better that we refactor some of the validation logic of it.

**Test cases**
If you can slow down your internet, this could be done on browser stack or with the `Network Link Conditioner` on macos 
Case 1:
(Repeat this many times until it happens)
* Try ti send some USDC
* Write an ens on the address input of Send screen and try to hit Confirm as fast as you can (For me I needed to use the enter of the Keyboard)
* Then write an amount on the amount screen
* On confirm screen check if the address is 0x0000000 (if it is this address is incorrect)

**Issue**

Progresses #5128 

**Checklist**


sethkfman (approved):
LGTM

PR #5182, opened 2022-10-31
(initial setup): include react-native-webrtc setup

jpcloureiro:
**Description**

react-native-webrtc on its current version requires bitcode to be downloaded.
This PR includes it in the initial setup so the developer doesn't need to run it separately.


andreahaku (approved):
There's another `setup:react-native-webrtc` in `bitrise.yml`. You can remove it there as well.

andreahaku (approved):
LGTM

Issue #5141, opened 2022-10-23
Allow to receive any data from scanned QR code via wallet_scanQRCode API


clzola:
I think it would be really useful to enable `wallet_scanQRCode` API to return URL. Some QR codes that could be found out in the world contains URLs and if someone wants to develop application to scan QR codes and use data stored in them, it cannot, as currently when `wallet_scanQRCode` is called, it opens camera, however if content of QR code is URL it automatically redirects user to that page, instead of returning data of the QR code. These QR codes may be from 3rd party vendors to which dev team does not have access or cannot issue request for changing content of QR code to be something else

I looked trough the source code and it looks like you are using already implemented component for scanning which has its own behavior which is ok, but from the DApps point of view, I think that is not right behavior and if DApp requested to scan QR code it should get that content no matter what. I think some flag should be added when navigating to QRScanner component to tell it to bypass all the checks and just return content of QR code.

I think that this RPC is really great feature that DApps can use, and grants access to camera. Otherwise developers needs to resort to javascript and html libraries for camera which often have issues with focus and cannot read codes if they are too small. I do not know if this poses any risks but would like to here if it does. Also if it is great risk, QRScanner component should act on its own if it scans URL that is related to MetaMask (like if it reads deep link)...

PR #5189, opened 2022-11-02
Allows DApps to read URLs from QR codes

clzola:
**Description**

I have added followURL parameter to QRScanner component. This parameter controls behavior when scanning QR code. Default behavior is that when scanner detects URL it asks users for redirection. However this is not practical when DApp requests for scanning of QR code, because DApp may expect URLs to be scanned but unfortunately current QRScanner component will never return URL. DApp can bypass this by using HTML5 and Javascript libraries for accessing Camera but these libraries have their own issues.

Requires: `needs-dev-review`, `needs-qa`

**Issue**

Progresses  [#5141]

**Checklist**

### DIFF
--- a/app/components/Views/QRScanner/index.tsx
+++ b/app/components/Views/QRScanner/index.tsx
@@ -44,6 +44,7 @@ export interface QRScannerParams {
   onScanError?: (error: string) => void;
   onStartScan?: (data: any) => Promise<void>;
   origin?: string;
+  followURL?: boolean = true;
 }
 
 export const createQRScannerNavDetails =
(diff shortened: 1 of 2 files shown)


clzola:

sethkfman:
@clzola Can you please clarify this statement below, is this because the user rejects? `DApp may expect URLs to be scanned but unfortunately current QRScanner component will never return URL`

If so, this is the expected behavior if the user rejects the URL the DApp should not receive the URL. This is a security feature to protect our users.

clzola:
@sethkfman If I call `wallet_scanQRCode` from my DApp it will open camera to scan QR code. If that QR code contains URL, a prompt will be presented to user with information about URL. The user has two options to reject or continue. If he chooses to reject camera closes and nothing happens. My code haven't received URL, which is fine. However if user click to continue, MetaMask redirects user to that URL and again DApp hasn't received QR code.

I don't see why is security risk for DApps to receive URL from QR codes when  `wallet_scanQRCode` API is called?

Issue #5176, opened 2022-10-31
ENS not being parsed correctly with deeplinks


tommasini:
**Describe the bug**
When we try to do a transaction of a ERC20 token with deeplinks, sending to an ENS name instead of an address, the ENS is not resolved to the correct address.

**Screenshots**
<img width=250 src="https://user-images.githubusercontent.com/46944231/199011498-4a6dbd53-01df-4875-ab3f-1539d299ea0c.jpeg" />

**To Reproduce**
* Go to https://metamask.github.io/metamask-deeplinks/
* Click on Payment Request
* Click on ERC20 token transfer
* Fill contract address input with 0x6B175474E89094C44Da98b954EedeAC495271d0F (DAI stable coin contract address)
* Fill the address receiver input with an ens of your choice (example: securitae.eth)
* Fill the chain Id input with 1
* Fill the amount of tokens with an amount that you want
* Open or scan the qr code 
* You can see that the address it's wrongly resolved

**Expected behavior**
The ens being resolved or an alert saying that we do not support ens deep link transactions.

tommasini:
Related with https://github.com/MetaMask/metamask-mobile/issues/1631

Issue #4853, opened 2022-08-12
Deeplinks/QR transfer scans no longer estimate the needed gas limit: "intrinsic gas too low"


PaulAtFlexa:
**Describe the bug**
Previously, when opening an ERC20 transfer deeplink the gas limit would be estimated, allowing users to "just click" send.  Now, when a user opens an ERC20 transfer, the gas limit is set to the default of 21000, which is far too low.  Thus, when clicking send, they get an "intrinsic gas too low" error, as metamask knows the tx would fail so does not send it (Thanks there :D ).  What should be happening is the gas is set to something like 75k depending on the complexity of a contract's transfer function.

I'm pretty familiar with the EIP-681 standard, so I took a stab at using the query params "gasLimit" and "gas" in the deeplink to at least fix this for my users temporarily.  Luckily for me, "gas" is accepted in the deeplink and QR scanning formats (though gasLimit is not, and per the 681 standard, should be).  But with this solution, users have to pay more gas than they need to up front since I had to set it to a high default.

The desired UX is that gas estimation is always done if gas or gasLimit is not specified in the link.  This currently works when you click send on a token, so the functionality exists already within the app.

**Screenshots**
![image](https://user-images.githubusercontent.com/93164809/184445352-bc2c286b-627d-48d7-90e3-3c1b5e5aaa04.png)

**To Reproduce**
_Steps to reproduce the behavior_
1. Create a deeplink for an ERC20 transfer https://metamask.github.io/metamask-deeplinks/#
2. Open it using a mobile device
3. Try to send the transaction without editing gas parameters
4. See it fail

**Expected behavior**
1. Open a deeplink for an ERC20 transfer
2. MetaMask looks up the token contract and estimates gas needed
3. User clicks send, it works as the gas was estimated correctly

**Smartphone (please complete the following information):**
 - Device: iPhone 13 Pro,
 - OS: 15.6
 - App Version v5.5.0 (939)

Same behavior reproduced on some of my colleagues phones as well, will update with their info if needed.

PS:

I work at https://flexa.network/ , we're a digital assets payments firm.  We'd love to establish closer relationships with wallet organizations like MetaMask.  You can reach me at paul@flexa.co !  We're building a way for wallet developers to embed a top notch payment experience within their app that's usable at tens of thousands of stores, as well as some of the biggest crypto conferences like Consensus.

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

[...]

PR #5191, opened 2022-11-02
Support for ens on deeplink transactions

tommasini:
**Description**
When reading a deep link with an ENS instead of an address, when it was an ERC20 token it was being resolved to the wrong address on the confirm screen. When it was eth, the transaction would not have a destination address to send the eth.

**Proposed Solution**
It is implementing support for deep link ERC20 and ETH transactions with ENS. When parsing the recipient from the deep link, we resolve it to an address if it is an ENS.

**Screenshots/Recordings**
* transactions with wrong ens and address (*Alert message approved 🚀*)
https://recordit.co/rcLMaD4ek0

* transactions ERC20 with ENS and address
 https://recordit.co/wTglGApwEz
 
* transaction ETH with ENS and address
https://recordit.co/2y3piAzfTe

* Small fix on UI of confirm screen

| Bug | Solution |
|-----|-----|
| Wrong UI | Right UI |

**Test Cases**
Use https://metamask.github.io/metamask-deeplinks/# to generate the deep links

Case1 (ERC20 tx):
* The ens resolve in the correct way
* Couldn't submit the transaction due to this [issue](https://github.com/MetaMask/metamask-mobile/issues/4853)

Case2: (ETH tx):
 * The ens resolve in the correct way
 * Transaction submitted with success 

Case3: (with the wrong ens or wrong address):
* Should appear one alert saying the recipient is wrong (*Alert message approved 🚀*)

**Code impact**
Low - Will impact the Send component and the deep link (QR Code) transactions

**Issue**

Progresses #5176 

**Checklist**


app/components/Views/Send/index.js
```diff
@@ -266,10 +267,14 @@ class Send extends PureComponent {
...
    */
-  handleNewTxMetaRecipient = (recipient) => {
+  handleNewTxMetaRecipient = async (recipient) => {
     let ensRecipient, to;
+
     if (isENS(recipient)) {
       ensRecipient = recipient;
+      if (!to) {
+        to = await doENSLookup(ensRecipient, this.props.network);
+      }
```
gantunesr, line 277: Is there a case where `to` will be defined before the `if` condition?

app/components/Views/Send/index.js
```diff
@@ -266,10 +267,14 @@ class Send extends PureComponent {
...
     } else if (recipient && recipient.toLowerCase().substr(0, 2) === '0x') {
       to = toChecksumAddress(recipient);
     }
```
gantunesr, line 280: We probably shouldn't convert an address to the checksum and throw an error

app/components/Views/Send/index.js
tommasini: No, you are totally right, completely escaped, thank you!!!

app/components/Views/Send/index.js
tommasini: Not understanding very well why we should throw an error here. When we have an address instead of an ens on the deep link we do this checksum 

Should not be converted?

gantunesr: We use the checksum as a way of verification that the address is correct, by doing an auto-checksum it bypasses this safety check. Please check https://github.com/MetaMask/mobile-planning/issues/288 for further reference.

tommasini: Done it

gantunesr (commented):
LGTM. Just left a last question

gantunesr (approved):
LGTM. Sending it to QA

plasmacorral:
Observed on Android 11 device (Samsung a515f) that if the deeplink is to an invalid ENS name (borangefox.eth), that the user is presented with a (seemingly) infinite spinning wheel.  

deeplink tested:
https://metamask.app.link/send/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48@1/transfer?address=borangefox.eth&uint256=1e15

https://recordit.co/pzAAfjaMdS

plasmacorral:
RE: Small fix on UI of confirm screen

I believe that we should leave the line spacing as-is rather than centering the address in the field.  There is an issue in production that if the TO: address is saved as a contact that neither the saved name nor the ENS are presented above the resolved address, like it would if the ENS name was not stored as a contact and was instead entered via the keyboard into the TO: field.  

I will open the production issue in another ticket.

plasmacorral:
Validated patch and invalid address with ENS names, as well as dapp and payment deeplinks.

PR #5198, opened 2022-11-04
Improve loading experience

Cal-L:
**Development & PR Process**

**Description**

[#459](https://github.com/MetaMask/mobile-planning/issues/459)

**Screenshots/Recordings**

**Issue**

Progresses #???

**Checklist**

### DIFF
--- a/app/components/Views/BrowserTab/index.js
+++ b/app/components/Views/BrowserTab/index.js
@@ -228,7 +228,6 @@ export const BrowserTab = (props) => {
   const url = useRef('');
   const title = useRef('');
   const icon = useRef(null);
-  const webviewUrlPostMessagePromiseResolve = useRef(null);
   const backgroundBridges = useRef([]);
   const fromHomepage = useRef(false);
   const wizardScrollAdjusted = useRef(false);
@@ -813,6 +812,10 @@ export const BrowserTab = (props) => {
    * When website finished loading
    */
   const onLoadEnd = ({ nativeEvent }) => {
+    // Do not update URL unless website has successfully completed loading.
+    if (nativeEvent.loading) {
+      return;
+    }
     // Use URL to produce real url. This should be the actual website that the user is viewing.
     const urlObj = new URL(nativeEvent.url);
     const { origin, pathname = '', query = '' } = urlObj;
@@ -953,7 +956,7 @@ export const BrowserTab = (props) => {
     if (!isAllowedUrl(hostname)) {
       return handleNotAllowedUrl(nativeEvent.url);
     }
-    webviewUrlPostMessagePromiseResolve.current = null;
+
     setError(false);
 
     changeUrl(nativeEvent);
--- a/app/components/Views/BrowserTab/index.js
+++ b/app/components/Views/BrowserTab/index.js
@@ -27,7 +27,6 @@ import onUrlSubmit, { getHost, getUrlObj, isTLD } from '../../../util/browser';
 import {
   SPA_urlChangeListener,
   JS_DESELECT_TEXT,
-  JS_WEBVIEW_URL,
 } from '../../../util/browserScripts';
 import resolveEnsToIpfsContentId from '../../../lib/ens-ipfs/resolver';
 import Button from '../../UI/Button';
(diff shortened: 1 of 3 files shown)


app/components/Views/BrowserTab/index.js
```diff
@@ -814,26 +813,15 @@ export const BrowserTab = (props) => {
    * When website finished loading
    */
   const onLoadEnd = ({ nativeEvent }) => {
-    if (nativeEvent.loading) return;
-    const { current } = webviewRef;
-
-    current && current.injectJavaScript(JS_WEBVIEW_URL);
-
-    const promiseResolver = (resolve) => {
-      webviewUrlPostMessagePromiseResolve.current = resolve;
```
NicholasEllul, line 823: Are we able to remove `webviewUrlPostMessagePromiseResolve` completely from this file if we are removing it from here and the spot further down below?

app/components/Views/BrowserTab/index.js
```diff
@@ -814,26 +813,15 @@ export const BrowserTab = (props) => {
...
-    });
+    // Use URL to produce real url. This should be the actual website that the user is viewing.
+    const urlObj = new URL(nativeEvent.url);
+    const { origin, pathname = '', query = '' } = urlObj;
+    const realUrl = `${origin}${pathname}${query}`;
+    // Generate favicon.
+    const favicon = `https://api.faviconkit.com/${getHost(realUrl)}/32`;
+    // Update navigation bar address with title of loaded url.
+    changeUrl({ ...nativeEvent, url: realUrl, icon: favicon });
+    changeAddressBar({ ...nativeEvent, url: realUrl, icon: favicon });
```
NicholasEllul, line 824: Love how much cleaner this is now. 

How does setting this here change rather than during `onMessage` below change the behaviour of the browser? Is there any notable changes we should note?

app/components/Views/BrowserTab/index.js
```diff
@@ -814,26 +813,15 @@ export const BrowserTab = (props) => {
...
-      webviewUrlPostMessagePromiseResolve.current = resolve;
-    };
-    const promise = current
-      ? new Promise(promiseResolver)
-      : Promise.resolve(url.current);
-
-    promise.then((info) => {
-      const { hostname: currentHostname } = new URL(url.current);
-      const { hostname } = new URL(nativeEvent.url);
-      if (info.url === nativeEvent.url && currentHostname === hostname) {
```
NicholasEllul, line 832: `&& currentHostname === hostname`. Hmm this must explain why on redirects we experience weirdness. As during redirects `currentHostname` is still the previous domain, whereas hostname is now our newly redirected to one

app/components/Views/BrowserTab/index.js
```diff
@@ -857,22 +845,6 @@ export const BrowserTab = (props) => {
...
-				case 'FRAME_READY': {
-					const { url } = data.payload;
-					onFrameLoadStarted(url);
-					break;
-				}*/
-        case 'GET_WEBVIEW_URL': {
-          const { url } = data.payload;
-          if (url === nativeEvent.url)
-            webviewUrlPostMessagePromiseResolve.current &&
-              webviewUrlPostMessagePromiseResolve.current(data.payload);
```
NicholasEllul, line 873: A bit of an echo from my earlier comment, do we have context on why we opted for setting this information from `onMessage` in the first place?

NEllusion (commented):
Can we add a regression test that will fail if the old behaviour ever comes back again?

app/components/Views/BrowserTab/index.js
Cal-L: Prior to these changes, we relied on a promise to be resolved by the `onMessage` callback before we updated the address bar. This change that we made removed the promise all together and simplified the logic so that we only rely on the `onLoadEnd` event to update the address bar.

app/components/Views/BrowserTab/index.js
Cal-L: Following the bread crumbs - I believe the reason for this approach was to rely on the webpage to resolve + return the favicon. With our fix, we just construct the favicon URL ourselves.

sethkfman (approved):
Thanks for the adding the comment for the WebView patch. LGTM

Cal-L:
> Can we add a regression test that will fail if the old behaviour ever comes back again?

Tagging @chrisleewilcox @cortisiko since they'll be adding automation for regression tests in the near future

PR #5200, opened 2022-11-04
Update Controllers to version 33.0.0

tommasini:
**Description**
Updating the controllers to version 33.0.0, will update the way Network Controller works
It was updated the PhishingControler
The CollectiblesController was renamed to NftController

**Test Cases**
Case1:
* Tested swaps on mainnet (uniswap) and custom network (Binance Smart Chain) (pancakeswap)
Case2:
* Tested transfer of collectible to myself on Mainnet
Case3:
* Tested token detection and nft Detection
Case4:
* Tested transfers between accounts, and the gas fee seems to be working and well defined (confirmed with etherscan and polyscan)
Case 5:
* Tested deep link with network changing
Case 6:
* Tested the activity view in different networks
Case 7:
* Tested switching accounts and verifying the balance on them (Also switching networks)
Case 8:
* Tested transfer collectables to other account on goerly
Case 9:
* Tested to create an new account it should change to it 

Would be good to do an overall test of the app.

**Issue**

Progresses #5092 

**Checklist**

* [ ] There is a related GitHub issue
* [ ] Tests are included if applicable
* [ ] Any added code is fully documented

PR #5215, opened 2022-11-08
Handle PhishingController breaking changes

Gudahtt:
The PhishingController was changed in v31 and v32 to no longer update the phishing configuration regularly in the background. Instead it has to be updated manually.

We now check for phishing configuration updates each time we're using it to test whether a URL is suspicious or not. This matches how the extension works. It should be roughly equivalent to how the Phishing Controller worked in past versions, except that now it will no longer be polling for updates when the browser is not in-use.

The phishing configuration is also explicitly updated once upon construction.

Note that this PR is pointing at #5200, so I have skipped most of the steps in the PR template because I am unsure how or whether they apply.

### DIFF
--- a/app/components/Views/BrowserTab/index.js
+++ b/app/components/Views/BrowserTab/index.js
@@ -410,11 +410,11 @@ export const BrowserTab = (props) => {
     const { PhishingController } = Engine.context;
 
     // Update phishing configuration if it is out-of-date
-    const phishingListsAreOutOfDate = this.phishingController.isOutOfDate();
+    const phishingListsAreOutOfDate = PhishingController.isOutOfDate();
     if (phishingListsAreOutOfDate) {
       // This is async but we are not `await`-ing it here intentionally, so that we don't slow
       // down network requests. The configuration is updated for the next request.
-      this.phishingController.updatePhishingLists();
+      PhishingController.updatePhishingLists();
     }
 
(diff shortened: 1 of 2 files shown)


app/components/Views/BrowserTab/index.js
```diff
@@ -408,6 +408,15 @@ export const BrowserTab = (props) => {
...
     const { PhishingController } = Engine.context;
+
+    // Update phishing configuration if it is out-of-date
+    const phishingListsAreOutOfDate = this.phishingController.isOutOfDate();
+    if (phishingListsAreOutOfDate) {
+      // This is async but we are not `await`-ing it here intentionally, so that we don't slow
+      // down network requests. The configuration is updated for the next request.
+      this.phishingController.updatePhishingLists();
+    }
+
```
tommasini, line 419: Should this be this way?
```Suggestion 

    // Update phishing configuration if it is out-of-date
    const phishingListsAreOutOfDate = PhishingController.isOutOfDate();
    if (phishingListsAreOutOfDate) {
      // This is async but we are not `await`-ing it here intentionally, so that we don't slow
      // down network requests. The configuration is updated for the next request.
      PhishingController.updatePhishingLists();
    }

```

tommasini (changes requested):

tommasini (approved):

PR #5216, opened 2022-11-08
Update controllers to v33

Gudahtt:
The controllers package has been updated to v33. This release included various breaking changes relating to NFTs. All controller APIs and state have been updated to use the term "NFT" instead of "collectible".

Note that this PR is pointing at https://github.com/MetaMask/metamask-mobile/pull/5200, so I have skipped most of the steps in the PR template because I am unsure how or whether they apply.


app/components/UI/Transactions/index.test.tsx
```diff
@@ -29,8 +29,8 @@ const initialState = {
       TokenRatesController: {
         contractExchangeRates: {},
       },
-      CollectiblesController: {
-        allCollectibleContracts: { '0x0': { 1: [] } },
+      NftController: {
+        allNfts: { '0x0': { 1: [] } },
```
tommasini, line 33: Should this be `allNftsContracts`?

tommasini (changes requested):

app/components/UI/Transactions/index.test.tsx
tommasini: I think this is the line that are doing the CI tests failing

app/components/UI/CollectibleContracts/index.js
```diff
@@ -253,16 +253,14 @@ const CollectibleContracts = ({
...
-          <View style={styles.emptyView}>
-            <CollectibleDetectionModal
-              onDismiss={dismissNftInfo}
-              navigation={navigation}
-            />
-          </View>
-        )}
+      {isMainNet(chainId) && !nftDetectionDismissed && !useNftDetection && (
+        <View style={styles.emptyView}>
+          <CollectibleDetectionModal
```
sethkfman, line 258: Will there be plans to update `CollectibleDetectionModal`?

tommasini (approved):
LGTM, awesome, thanks!

Gudahtt:
I considered going down the route of renaming all of the variables/components/etc. in mobile, but there are A LOT. It would have blown up the scope of this update enormously. I think we should do this as a follow-up refactor task instead.

PR #5217, opened 2022-11-09
[FIX] Send to the wrong address

tommasini:
All the information on https://github.com/MetaMask/mobile-planning/issues/472#issuecomment-1308689200

### DIFF
--- a/app/components/Views/SendFlow/Amount/index.js
+++ b/app/components/Views/SendFlow/Amount/index.js
@@ -662,7 +662,7 @@ class Amount extends PureComponent {
       transactionObject.readableValue = value;
     }
 
-    if (selectedAsset.isETH) {
+    if (selectedAsset.isETH) transactionObject.to = transactionTo;
       transactionObject.to = transactionTo;
     }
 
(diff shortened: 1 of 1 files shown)


app/components/Views/SendFlow/Amount/index.js
```diff
@@ -662,6 +662,10 @@ class Amount extends PureComponent {
       transactionObject.readableValue = value;
     }
 
+    if (selectedAsset.isETH) {
```
sethkfman, line 665:
```suggestion
    if (selectedAsset.isETH) transactionObject.to = transactionTo;
```

sethkfman: Not a blocker but I think this would accomplish the same thing with a couple less lines.

sethkfman (approved):
LGTM

app/components/Views/SendFlow/Amount/index.js
tommasini: Pushed suggestion

PR #5218, opened 2022-11-09
Improve deeplinks experience

jpcloureiro:
**Development & PR Process**

**Description**

**Screenshots/Recordings**

**Issue**

Progresses [#460](https://github.com/MetaMask/mobile-planning/issues/460)

**Checklist**


app/components/Views/BrowserTab/index.js
```diff
@@ -1339,6 +1410,15 @@ export const BrowserTab = (props) => {
   /**
    * Main render
    */
+
+  const originWhitelist = Object.keys(knownProtocols).map((key) => {
+    const protocol = key.startsWith('http')
```
NicholasEllul, line 1415: Do we know if theres a case that `http` could have any upper cases (ex: `hTTp://google.com`, or does that get normalized somewhere?

app/components/Views/BrowserTab/index.js
```diff
@@ -785,15 +785,86 @@ export const BrowserTab = (props) => {
...
+        {
+          text: 'Allow',
+          onPress: () => allowLinkOpen(url),
+          style: 'allow',
+        },
+      ]);
+      return false;
+    }
+
+    return true;
```
NicholasEllul, line 853: In the event a protocol comes in such as `foobar://` this if-statement will evaluate false as there is no alert message, and as a result the `verifyKnownProtocols` function and the `onShouldLoadWithRequest` will return true. This will lead to us proceeding with the execution of unknown protocols. Can we also have a generic alert to protect users in that case?

NEllusion (commented):
I love this approach of intercepting the request! Are we able to add any tests to catch regressions in this behaviour?

app/components/Views/BrowserTab/index.js
```diff
@@ -785,15 +785,86 @@ export const BrowserTab = (props) => {
...
+    const { url, isTopFrame } = event;
+
+    const [protocol] = url.split(':');
+
+    const { alertMsg } = knownProtocols[protocol];
+
+    if (!isTopFrame && Boolean(alertMsg)) {
+      Alert.alert(alertMsg, '', [
+        {
+          text: 'Ignore',
```
sethkfman, line 840: Check to see if there is something already available in our locales for translations.

andreahaku:
@jpcloureiro did you check how this change behaves with the app specific schemas like `metamask://`, `dapp://`, 'wc://' or 'ethereum://'?

jpcloureiro:
> @jpcloureiro did you check how this change behaves with the app specific schemas like `metamask://`, `dapp://`, `wc://` or `ethereum://`?

@andreahaku opening any of these schemas inside the webview will open the alert dialog. If the user presses allow, it will execute as before.

app/components/Views/BrowserTab/index.js
jpcloureiro: I've modified the code so any non whitelisted protocol will trigger an alert dialog with a generic message

app/components/Views/BrowserTab/index.js
```diff
@@ -785,6 +785,56 @@ export const BrowserTab = (props) => {
     });
   }, []);
 
+  const protocolWhitelist = ['about:', 'http:', 'https:'];
```
sethkfman, line 788: If any these are not impacting state, could they be moved out to a `ults` class that is TSX and can be tested?

app/util/browser/index.test.ts
```diff
@@ -1,4 +1,29 @@
...
+  isTLD,
+  getAlertMessage,
+  protocolWhitelist,
+  sanitizeUrl,
+  getUrlObj,
+  getHost,
+} from '.';
+import { strings } from '../../../locales/i18n';
+
+describe('Browser utils :: sanitizeUrl', () => {
```
sethkfman, line 11: Nice work on these test cases

app/util/browser/index.ts
```diff
@@ -0,0 +1,141 @@
...
+import isUrl from 'is-url';
+
+/**
+ * Returns a sanitized url
+ *
+ * @param url - String corresponding to url
+ * @param defaultProtocol - Protocol string to append to URLs that have none
+ * @returns - String corresponding to sanitized input depending if it's a search or url
+ */
+export const sanitizeUrl = (url: string, defaultProtocol = 'https://') => {
```
sethkfman, line 11: I like the effort to refactor to type script.

app/components/Views/BrowserTab/index.js
```diff
@@ -784,14 +791,38 @@ export const BrowserTab = (props) => {
...
+    // Stops normal loading when it's ens, instead call go to be properly set up
     if (isENSUrl(url)) {
       go(url.replace(/^http:\/\//, 'https://'));
       return false;
     }
-    return true;
+
+    // Continue request loading it the protocol is whitelisted
+    const { protocol } = new URL(url);
+    if (protocolWhitelist.includes(protocol)) return true;
```
sethkfman, line 806: Could we call change this to `protocolAllowlist()`?

sethkfman (approved):
One minor comment for a method rename but other than that it looks good.

app/components/Views/BrowserTab/index.js
jpcloureiro: Yes, changing `protocolWhitelist` to `protocolAllowList`

sethkfman (approved):
LGTM

app/util/browser/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (!isUrl(input) && !regEx.test(input)) {
+    // Add exception for localhost
+    if (
+      !input.startsWith('http://localhost') &&
+      !input.startsWith('localhost')
+    ) {
+      // In case of keywords we default to google search
+      let searchUrl = 'https://www.google.com/search?q=' + escape(input);
+      if (searchEngine === 'DuckDuckGo') {
+        searchUrl = 'https://duckduckgo.com/?q=' + escape(input);
```
NicholasEllul, line 44: Escape is now [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) and instead its recommended that we use [encodeUriComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

app/util/browser/index.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * @param url - String corresponding to url
+ * @param defaultProtocol - Protocol string to append to URLs that have none
+ * @returns - String corresponding to sanitized input depending if it's a search or url
+ */
+export const sanitizeUrl = (url: string, defaultProtocol = 'https://') => {
+  const hasProtocol = /^[a-z]*:\/\//.test(url);
+  const sanitizedURL = hasProtocol ? url : `${defaultProtocol}${url}`;
+  return sanitizedURL;
+};
```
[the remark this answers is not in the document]
jpcloureiro: Agree, `prefixUrlWithProtocol` is better suited for what the function does.

sethkfman:
@jpcloureiro Can you add the scenarios you tested and update the lint errors?

cortisiko (changes requested):
@jpcloureiro sending this back your way!

**Issue** 

On the webview error screen, when I tap the "return to home page" button I get:`[TypeError: Invalid URL: home.metamask.io]`. Furthermore, I am not taken back to the mm home page.  
see [recording](http://recordit.co/RncotziS6t)

To reproduce:
- open the browser
- enter an invalid URL. For example https://asfasfo.aox
- When the error screen appears
- tap on "return to home page". Notice the error.

jpcloureiro:
> @jpcloureiro sending this back your way!
> 
> **Issue**
> 
> On the webview error screen, when I tap the "return to home page" button I get:`[TypeError: Invalid URL: home.metamask.io]`. Furthermore, I am not taken back to the mm home page. see [recording](http://recordit.co/RncotziS6t)
> 
> To reproduce:
> 
> * open the browser
> * enter an invalid URL. For example https://asfasfo.aox
> * When the error screen appears
> * tap on "return to home page". Notice the error.

@cortisiko pushed a fix. It should go back to mm home page now.

cortisiko (approved):
Fix looks good @jpcloureiro, This is QA passed! 🌮 🌮

PR #5318, opened 2022-11-29
Fix Estimated gas fee calculation on the tx detail page 

dan437:
## Description
- Fix Estimated gas fee calculation on the tx detail page for Swaps on EIP-1559 networks
- Fix NAN ETH in Total amount if estimated gas fee is < 0.00001 ETH

## Screenshots
Before:
image

After:
image

Before:
image

After: 
image


gantunesr (approved):
LGTM

PR #5324, opened 2022-11-30
Use npm for dependencies

rickycodes:
previously we were using git for some deps... this was blocking #5147 

it's also a security concern in the cases where we weren't pinning to a SHA

I worked with @Gudahtt to fork all the deps and publish them under the `@metamask/` namespace.

This PR makes use of the newly published modules.

I'm opening this as a separate PR so #5147 isn't so heavy handed.

- [x] update git deps to use npm
- [x] update imports
- [x] update mocks
- [x] update patches


package.json
```diff
@@ -9,13 +9,13 @@
...
     "clean:android": "rm -rf android/app/build",
-    "clean:node": "rm -rf node_modules && yarn --frozen-lockfile",
+    "clean:node": "rm -rf node_modules && yarn",
     "clean": "yarn clean:ios && yarn clean:android && yarn clean:node",
     "clean-android": "yarn clean:android && yarn clean:node",
     "lint": "eslint '**/*.{js,ts,tsx}' --ignore-path=.prettierignore",
     "lint:fix": "eslint '**/*.{js,ts,tsx}' --fix --ignore-path=.prettierignore",
     "format": "prettier '**/*.{js,ts,tsx,json}' --write",
-    "setup": "yarn clean && yarn pod:install && yarn allow-scripts && ./scripts/postinstall.sh",
+    "setup": "yarn pod:install && yarn allow-scripts && ./scripts/postinstall.sh",
```
Gudahtt, line 18: It looks like node dependencies aren't being installed here anymore. That doesn't seem right.

rickycodes: I removed it from `package.json`, but added it in ci: https://github.com/MetaMask/metamask-mobile/pull/5324/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR29

Gudahtt: A pattern I used elsewhere was to create a `postinstall` script, and use `yarn && yarn postinstall` for `yarn setup`, and `yarn --frozen-lockfile && yarn postinstall` in CI.

Or `yarn clean` in this case rather than `yarn` maybe. Not sure on the background there; presumably we're deleting `node_modules` each time for good reason.

rickycodes: I'm just going to take this change out for now. I'll make the changes you suggested in a follow-up

package.json
```diff
@@ -294,7 +294,8 @@
     "uuid": "^8.3.2",
     "valid-url": "1.0.9",
     "vm-browserify": "1.1.2",
-    "zxcvbn": "4.4.2"
+    "zxcvbn": "4.4.2",
+    "react-native-tcp": "^3.2.1"
```
Gudahtt, line 298: Would like to understand how this is getting re-added before we merge this. We don't want two copies of this

rickycodes: after taking a closer look at this one the PR (`aprock/react-native-tcp#11/head`) had actually been merged. So I think we're fine to _not_ fork here and simply use latest `4.0.0`

Gudahtt: Actually wait a minute, it doesn't look like that change is a part of v4.0.0. The changes in that PR diff aren't in the commit tagged as v4, it's not in the commit history either.

rickycodes: ya. closing this for now since it's state (the work) and stale (the context in my brain)

also, apparently the mobile team is in the process of upgrading react native and that will affect all of this... so probably better to wait and do the work interactively as you mentioned

cc @Gudahtt

PR #4917, opened 2022-08-25
[Force upgrade]Trigger UpdateNeeded screen

owencraston:
**Description**
- [designs](https://www.figma.com/file/BVcxQCC87m22iqI2lCFDj8/Forced-Upgrades?node-id=83%3A15)
_Write a short description of the changes included in this pull request, also include relevant motivation and context. Have in mind the following questions,_
_1. What is the reason for the change?_
- in order to make users more secure we need a way to warn them if their app has a security vulnerability 
- this change prompts users to enable automatic checks (since we do not want to enable it by default as it will expose their IP address to github)
- after this we can verify if a users app is up to date enough to remain secure
_2. What is the improvement/solution?_
- Prompts the user to enable the automatic security checks feature
    - if the user says no thanks then the we will not show the modal to update their app
- if the user enables this feature then we will check if their app version is over the minimum version hosted here: https://github.com/MetaMask/metamask-mobile/tree/gh-pages
- if the minimum version is higher than the current users app version we will prompt them to update
- if the user clicks the the "Update to latest version" button it will take them to the app store/play store
- if they click "Remind me later" then the modal will go away until they kill the app and open it again
**Screenshots/Recordings**

https://user-images.githubusercontent.com/22918444/192904894-d2d55220-4927-4ab6-85cc-9df19ed00cfc.MP4

## QA
- in order to QA you will need three builds, one that is 1 versions below the minimum version, one that is equal to the minimum version and one that is above the minimum version.
#### Below minimum version
- [v5.10.1 #998](https://app.bitrise.io/build/8a5274e3-7e60-4c94-a1a2-008194b28da3?tab=log)

#### Equal to minimum version
- [v5.10.1 #999](https://app.bitrise.io/build/002996ba-ac30-480c-acdb-f193c99e18e2)

#### Above the minimum version
- [v5.10.1 #1001](https://app.bitrise.io/build/d0b3e65a-1a6e-4cea-af25-21e5d80fef3e)

### QA steps
1.  download the version below the minimum version (987)
2. setup wallet

#### Path 1: Enabling automatic security checks
1. a prompt will appear to enable the `automatic security check`, click `enable`
2. on build `897` the force upgrade modal should appear
3. dismiss this modal by either swiping down or or clicking remind me later
4. continue to use the wallet and the modal `should not appear`
5. force quit the app and reopen it
6. the `modal should appear again`
7. clicking `Update to latest version` should deep link to the app store/play store
8. Update the app via test flight or by reinstalling the apk (989)
[...]

PR #5333, opened 2022-12-01
[Force upgrade] only fetch minimum versions if permissions enabled

owencraston:
**Development & PR Process**

**Description**

- I noticed that due to an oversight I made, we were making fetch calls to Github before the user granted persmission. Luckily this was not in production. 
- this is a small change that checks if this permission is given before making the fetch. If not we will not fetch and return a state of Error for the useAppConfig.
- This fix patches a bug made in this [original PR](https://github.com/MetaMask/metamask-mobile/pull/4917)

**Screenshots/Recordings**

https://user-images.githubusercontent.com/22918444/205169106-7abbc9e4-5be9-409e-beeb-b13bc3d22d2c.mov

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/505

**Checklist**

### DIFF
--- a/app/components/hooks/AppConfig/useAppConfig.tsx
+++ b/app/components/hooks/AppConfig/useAppConfig.tsx
@@ -8,46 +8,53 @@ const initialState: Readonly<State> = {
   type: 'Loading',
 };
 
-const useAppConfig = (): State => {
+const useAppConfig = (hasGithubPermissions: boolean): State => {
   const [state, setState] = useState<State>(initialState);
 
   useEffect(() => {
     const fetchAppConfig = () => {
-      fetch(MM_APP_CONFIG_URL)
-        .then((response) => response.json())
-        .then((data) => {
-          try {
-            const minimumVersions = data.security.minimumVersions;
-            const appConfig: AppConfig = {
-              security: {
-                minimumVersions: {
-                  appMinimumBuild: minimumVersions.appMinimumBuild,
-                  appleMinimumOS: minimumVersions.appleMinimumOS,
-                  androidMinimumAPIVersion:
-                    minimumVersions.androidMinimumAPIVersion,
+      if (hasGithubPermissions) {
+        fetch(MM_APP_CONFIG_URL)
+          .then((response) => response.json())
+          .then((data) => {
+            try {
+              const minimumVersions = data.security.minimumVersions;
+              const appConfig: AppConfig = {
+                security: {
+                  minimumVersions: {
+                    appMinimumBuild: minimumVersions.appMinimumBuild,
+                    appleMinimumOS: minimumVersions.appleMinimumOS,
+                    androidMinimumAPIVersion:
+                      minimumVersions.androidMinimumAPIVersion,
+                  },
                 },
-              },
-            };
-            setState({ type: 'Success', data: appConfig });
-          } catch (e: any) {
+              };
+              setState({ type: 'Success', data: appConfig });
+            } catch (e: any) {
+              setState({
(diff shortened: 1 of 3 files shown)


app/components/hooks/AppConfig/useAppConfig.tsx
```diff
@@ -8,46 +8,53 @@ const initialState: Readonly<State> = {
...
-            const minimumVersions = data.security.minimumVersions;
-            const appConfig: AppConfig = {
-              security: {
-                minimumVersions: {
-                  appMinimumBuild: minimumVersions.appMinimumBuild,
-                  appleMinimumOS: minimumVersions.appleMinimumOS,
-                  androidMinimumAPIVersion:
-                    minimumVersions.androidMinimumAPIVersion,
+      if (hasGithubPermissions) {
+        fetch(MM_APP_CONFIG_URL)
```
owencraston, line 17: if you want to test this you will need to change the endpoint to the test endpoint `MM_APP_CONFIG_TEST_URL` which points to https://metamask.github.io/metamask-mobile/AppConfig/test/MockAppConfig.json

app/components/hooks/MinimumVersions/useMinimumVersions.tsx
```diff
@@ -7,12 +7,12 @@ import { useNavigation } from '@react-navigation/native';
 import { InteractionManager } from 'react-native';
 
 const useMinimumVersions = () => {
-  const minimumValues = useAppConfig();
-  const currentBuildNumber = Number(getBuildNumber());
-  const navigation = useNavigation();
   const allowAutomaticSecurityChecks = useSelector(
     (state: any) => state.security.automaticSecurityChecksEnabled,
   );
+  const minimumValues = useAppConfig(allowAutomaticSecurityChecks);
```
owencraston, line 13: this is a very small change. Now we are getting the permissions first and passing that to the useAppConfig hook. If the permissions are granted then everything will perform the same, if they are not we will not perform the fetch.

owencraston: The key issue here was, we would not trigger the force update screen before because `allowAutomaticSecurityChecks` would be false but we still performed the call to github with the `useAppConfig`. Now we won't trigger the screen because `allowAutomaticSecurityChecks` is false `AND` because we won't get a response from github.

gantunesr (approved):
LGTM

app/components/hooks/AppConfig/useAppConfig.tsx
```diff
@@ -8,46 +8,53 @@ const initialState: Readonly<State> = {
...
-        .catch((e: any) => {
-          setState({
-            type: 'Error',
-            error: e,
-            message: `error fetching AppConfig ${e.message}`,
           });
+      } else {
+        setState({
+          type: 'Error',
+          message: `GitHub request permissions not granted by user. See hasUserSelectedAutomaticSecurityCheckOption global state`,
```
sethkfman, line 51: Are these errors bubbled up to the user?

sethkfman (approved):
LGTM

app/components/hooks/AppConfig/useAppConfig.tsx
owencraston: These errors will never bubble up into the UI. If you look in the `useMinimumVersions` hook we check for data being available on the response which we do not set here. Therefore we can safely say that the `shouldTriggerUpdateFlow` will be false

chrisleewilcox (approved):
LGTM

PR #5459, opened 2023-01-06
remove extra zero balance account potentially created from seeking ahead

rickycodes:
This is to match functionality with: https://github.com/MetaMask/metamask-extension/pull/12074

Explanation:
 - A previous PR added 'seeking' to make sure that when users import their wallets with their seed, consecutive accounts with non-zero balances are automatically generated. However, the PR missed part of the logic: removing the final, zero balance extra account, if created. More motivation behind this PR described in issue https://github.com/MetaMask/metamask-extension/issue/12073.
 

| Before: | After |
| --- | ----------- |
| image | image
| (account 3 is imported with zero eth) | (account 3 is removed immediately after import (matching extension functionality)) |

### DIFF
--- a/app/util/importAdditionalAccounts.js
+++ b/app/util/importAdditionalAccounts.js
@@ -72,5 +72,11 @@ export default async () => {
     i++;
   }
 
+  // remove extra zero balance account potentially created from seeking ahead
+  if (accounts.length > 1 && lastBalance === '0x0') {
+    await KeyringController.removeAccount(accounts[accounts.length - 1]);
+    accounts = await KeyringController.getAccounts();
+  }
+
   updateIdentities(accounts);
 };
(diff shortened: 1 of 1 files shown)


owencraston (commented):
Can we edit/add tests for this functionality similar to the ones [here](https://github.com/MetaMask/metamask-extension/pull/12074/files)?

owencraston (approved):

Cal-L (approved):
Is there a way to prevent adding the zero balance account in the first place rather than removing after it's added? In any case, since this is just a quick patch and doesn't mess with existing logic, LGTM.

rickycodes:
> Is there a way to prevent adding the zero balance account in the first place rather than removing after it's added?

unfortunately no. you have to add the account with the zero balance before you can check it's balance

PR #5475, opened 2023-01-11
Fix import of `isEIP1559Transaction`

Gudahtt:
**Development & PR Process**

**Description**

The import of `isEIP1559Transaction` was not working; that function was accidentally removed from the package exports in a recent update. It has been updated to import directly from the util files in the package instead.

This is a temporary solution. We will also update the `@metamask/transaction-controller` package to export this method, so we don't need to reference a file internal to the package. This is being fixed like this for now so that we don't have to wait for the package update, because the next package update will be a lot of work.

**Checklist**


Gudahtt:
This function is exported in this PR: https://github.com/MetaMask/core/pull/1058

wachunei (approved):
LGTM

PR #5556, opened 2023-01-20
Swaps disabled on fresh install on some networks

tommasini:
**Description**
Swaps were disabled on networks after a fresh install.
This is an old bug since I went back until 5.6.0 and it was happening https://recordit.co/EZ2P5eyFeT

**Proposed Solution**
Swaps are enabled on fresh installation for all networks

**Technical Solution**
When the network changed, the SwapsLiveness component was not mounted again, which makes the `hasMountChecked` state true and not false like it supposes to be. Since there isn't any reason on my sight for not running the `checkLiveness` function if the `hasMountChecked` is true, I removed the condition.

**Test Cases**
* Fresh install
* Changed to avalanche network (swaps enabled)
* Changed to Binance Smart Chain (swaps enabled)
* Performed one swap

**Screenshots/Recordings**
https://recordit.co/UR0XppwkNQ

**Code Impact**
Small

**Issue**

Progresses #[589](https://github.com/MetaMask/mobile-planning/issues/589)

**Checklist**

### DIFF
--- a/app/components/UI/Swaps/SwapsLiveness.js
+++ b/app/components/UI/Swaps/SwapsLiveness.js
@@ -41,7 +41,7 @@ function SwapLiveness({ isLive, chainId, setLiveness }) {
 
   // Check on mount
   useEffect(() => {
-    if (isSwapsAllowed(chainId) && !isLive && !hasMountChecked) {
+    if (isSwapsAllowed(chainId) && !isLive) {
       setHasMountChecked(true);
       checkLiveness();
     }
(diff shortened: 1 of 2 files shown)


wachunei:
@tommasini  I think we need to have a different approach here and is to reset the value of `hasMountChecked` to `false` when the `chainId` changes, instead of skipping this check which is logically correct ("if it has not been checked, mark it as checked")

wachunei (approved):
LGTM

Issue #5513, opened 2023-01-13
Migrate provider configuration to  RpcMethodMiddleware


Gudahtt:
### **Description**

Migrate the functions setup in the network controller `providerConfig` to the RpcMethodMiddleware.

This provider configuration will be removed in a future release of the network controller. It is used today to support certain RPC methods. It can be migrated to be alongside the rest of mobile's RPC method handlers, so that mobile is unaffected by this future change to the network controller.

### **Technical Details**

- Migrate the `eth_sendTransaction` handler from `Engine.js` to `RpcMethodMiddleware.ts`.
- Remove the `getAccounts` function, instead using the `getAccounts` function in `RpcMethodMiddleware.ts` to meet this need for any RPC method currently relying on this.

Note that once both of these entries are removed from the `providerConfig`, it will still need to be set to an empty object. The provider config being set triggers the network controller to initialize, so it should not be removed entirely.

### **Acceptance Criteria**

-   The provider configuration is empty.

### **References**

This helps to unblock https://github.com/MetaMask/core/issues/970

### Definition of Done
- Whenever applies, any change unit tested, reviewed(approved) and documented(JSDOC at least)
- [ ] We have changelog entries for any related changes
- [ ] Any changed APIs have comprehensive inline documentation
- [ ] Any changed public APIs are well covered by unit tests
- Other items
- [ ] If a planning/research ticket, then the plan has been reviewed and approved by at least 1 team members

xinnanyemm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Core-621e46b4d7103800171d1b02/issues/MetaMask/metamask-mobile/5513?planning-poker) with Zenhub @mcmire @Gudahtt

Gudahtt:
This has been completed in these PRs:

https://github.com/MetaMask/metamask-mobile/pull/5619
https://github.com/MetaMask/metamask-mobile/pull/5620
https://github.com/MetaMask/metamask-mobile/pull/5627

PR #5619, opened 2023-01-26
Refactor `eth_sendTransaction` handler

Gudahtt:
**Development & PR Process**
1. Follow MetaMask [Mobile Coding Standards](https://docs.google.com/document/d/1VJLwTRsUw_5EDq_o8d6sSbXUAYENLurkRitYO45eY5o/edit?usp=sharing)
2. Add `release-xx` label to identify the PR slated for a upcoming release (will be used in release discussion)
3. Add `needs-dev-review` label when work is completed
4. Add `needs-qa` label when dev review is completed
5. Add `QA Passed` label when QA has signed off

**Description**

The handler for `eth_sendTransaction` was previously spread between `RPCMethodMiddleware.ts` and the static middleware of `web3-provider-engine`. Instead it is all handled in `RPCMethodMiddleware.ts` now.

This should have no functional impact. It is difficult to trace through `web3-provider-engine`, but this case is one of the easier ones because the static middleware is run first, and in this case it will always end the request.

**Issue**

This relates to #5513

**Checklist**

* [x] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

PR #5620, opened 2023-01-26
Refactor RPC `getAccounts` usage

Gudahtt:
**Development & PR Process**

**Description**

The `getAccounts` function is defined and used in two different places in the RPC pipeline: `RPCMethodMiddleware.ts` and
`web3-provider-engine`. The second one has been removed, so now all usage is in `RPCMethodMiddleware.ts`.

`getAccounts` is passed into `web3-provider-engine` and used in the `HookedWalletSubprovider`: https://github.com/MetaMask/web3-provider-engine/blob/cf612f898002833c36730d23972fe4c4dd483c76/subproviders/hooked-wallet.js

It is used in these methods:
* `eth_coinbase`
* `eth_accounts`
* `parity_defaultAccount`

It's also called to validate the sender, for the following methods:
* `eth_signTypedData`
* `eth_signTypedData_v3`
* `eth_signTypedData_v4`
* `encryption_public_key`
* `eth_decryptMessage`
* `personal_sign`
* `eth_sign`
* `eth_signTransaction`
* `eth_sendTransaction`

Of these methods, most of them are intercepted in
`RPCMethodMiddleware.ts`, so the requests never make it to `web3-provider-engine`.

These three methods will make their way there: `parity_defaultAccount`, `encryption_public_key`, and `eth_decryptMessage`. The decryption- related messages rely on constructor parameters that mobile does not pass in, so those always throw an error. The only method this hook is used for in practice is `parity_defaultAccount`.

The `parity_defaultAccount` method was added to
`RPCMethodMiddleware.ts`, so now that method won't make it to `web3-provider-engine` either. Functionally it is the same as `eth_coinbase`, so the same implementation has been used.

This should have no functional changes.

**Issue**

This relates to https://github.com/MetaMask/metamask-mobile/issues/5513

**Checklist**

### DIFF
--- a/app/core/RPCMethods/RPCMethodMiddleware.ts
+++ b/app/core/RPCMethods/RPCMethodMiddleware.ts
@@ -324,6 +324,7 @@ export const getRpcMethodMiddleware = ({
       },
       eth_accounts: getEthAccounts,
       eth_coinbase: getEthAccounts,
+      parity_defaultAccount: getEthAccounts,
       eth_sendTransaction: async () => {
         checkTabActive();
         const { TransactionController } = Engine.context;
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -111,23 +111,8 @@ class Engine {
       };
 
       const networkController = new NetworkController(networkControllerOpts);
-      networkController.providerConfig = {
-        getAccounts: (
-          end: (arg0: null, arg1: any[]) => void,
-          payload: { hostname: string | number },
-        ) => {
-          const { approvedHosts } = store.getState();
-          const isEnabled = approvedHosts[payload.hostname];
-          const { KeyringController } = this.context;
-          const isUnlocked = KeyringController.isUnlocked();
-          const selectedAddress =
-            this.context.PreferencesController.state.selectedAddress;
-          end(
-            null,
-            isUnlocked && isEnabled && selectedAddress ? [selectedAddress] : [],
-          );
-        },
-      };
+      // This still needs to be set because it has the side-effect of initializing the provider
+      networkController.providerConfig = {};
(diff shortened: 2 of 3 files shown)


app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -285,6 +285,12 @@ export const getRpcMethodMiddleware = ({
         const accounts = await getAccounts();
         res.result = accounts.length > 0 ? accounts[0] : null;
       },
+
+      parity_defaultAccount: async () => {
```
sethkfman, line 289: Does extension also have a `parity_defaultAccount` I am trying to understand where this name is derived?

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -322,6 +322,7 @@ export const getRpcMethodMiddleware = ({
       },
       eth_accounts: getEthAccounts,
       eth_coinbase: getEthAccounts,
+      parity_defaultAccount: getEthAccounts,
```
Gudahtt, line 325: I can add a unit test for this, but it would be easiest to wait until after https://github.com/MetaMask/metamask-mobile/pull/5619 is merged first. That PR establishes baseline RPC middleware test helpers/patterns that I could use here as well.

Gudahtt: Unit tests have been added in 2029c06d0f89415611a22dc27bfde3a7b695b32b

Cal-L (approved):
Nice, it looks like like this change will enable Permission system for parity_defaultAccount as well. LGTM

Cal-L:
Also applied spot check on release since no functional changes

PR #5627, opened 2023-01-27
Refactor remaining `web3-provider-engine` methods

Gudahtt:
**Development & PR Process**

**Description**

The remaining non-static methods from `web3-provider-engine` have been migrated to `RPCMethodMiddleware.ts` with the rest of the method handlers. The two methods left were `personal_ecRecover` and `parity_checkRequest`.

`personal_ecRecover` will recover the signing account from a signature. This operation requires no private keys. Commonly dapps will do this themselves. Persumably it was added to the dapp API at some point for convenience. It has been preserved just to avoid making breaking changes to the dapp API.

`parity_checkRequest` would return the result of a parity signature or transaction. It returned `null` if none were found. The parity signature and request methods have been throwing an error for some time now, so there is never any result to check. As a result, this method has been returning `null` for all requests in practice. It has been preserved just to avoid making breaking changes to the dapp API.

**Issue**

This relates to https://github.com/MetaMask/metamask-mobile/issues/5513

**Checklist**

### DIFF
--- a/app/core/RPCMethods/RPCMethodMiddleware.ts
+++ b/app/core/RPCMethods/RPCMethodMiddleware.ts
@@ -2,6 +2,7 @@ import { Alert } from 'react-native';
 import { getVersion } from 'react-native-device-info';
 import { createAsyncMiddleware } from 'json-rpc-engine';
 import { ethErrors } from 'eth-json-rpc-errors';
+import { recoverPersonalSignature } from '@metamask/eth-sig-util';
 import RPCMethods from './index.js';
 import { RPC } from '../../constants/network';
 import { NetworksChainId, NetworkType } from '@metamask/controller-utils';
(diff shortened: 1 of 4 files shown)


app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -375,6 +376,21 @@ export const getRpcMethodMiddleware = ({
...
+        const address = recoverPersonalSignature({ data, signature });
+
+        res.result = address;
+      },
+
+      parity_checkRequest: () => {
+        // This method is retained for legacy reasons
+        // It doesn't serve it's intended purpose anymore of checking parity requests,
+        // because our API doesn't support parity requests.
+        res.result = null;
```
BelfordZ, line 391: Should we return an error here instead? That way we can give a useful message about this methods removal / deprecation (if anyone is even using it)

Gudahtt: My intent here was just to preserve the pre-existing behavior, and leave the mobile team and/or your team to decide what to do with it from here on out.

Though, with the new permission controller update, this method may have been broken already :thinking: I don't see it listed as an unrestricted method, which would mean it would no longer work. So maybe this decision has already been made. I'll follow up on this.

BelfordZ (approved):
nice catch!

app/core/RPCMethods/RPCMethodMiddleware.ts
Gudahtt: It looks like the permission system is not yet used for WalletConnect or SDK, so I'll assume we still want to preserve these for now.

Gudahtt:
I can add unit tests for these method handlers after this PR is merged: https://github.com/MetaMask/metamask-mobile/pull/5619

Cal-L (approved):
Currently discussing next steps on removing parity rpc methods. https://consensys.slack.com/archives/CBW7S9FSN/p1680805775457939 
Since there is no change in functionality here, LGTM

Issue #5648, opened 2023-01-31
 [Bug]: ENS resolving to a wrong address when using a Deeplink for an ERC20 Approve tx 


seaona:
**Describe the bug**
 When I am performing an ERC20 Approve using a Deeplink, the ENS resolves to a wrong value

**Screenshots**

https://user-images.githubusercontent.com/54408225/215779167-e0389c09-177b-46c8-b937-c71e59dc9a1a.mp4

![image](https://user-images.githubusercontent.com/54408225/215779416-fd88b9d9-31d9-4455-ad6c-ef86ff626410.png)

**To Reproduce**
_Steps to reproduce the behavior_
1. Select Mainnet
2. Open MM Scanner
3. Scan this QR code. It represents this deeplink: https://metamask.app.link/approve/0xFd09911130e6930Bf87F2B0554c44F400bD80D3e@1/approve?address=seaona.eth&uint256=3e18

![image](https://user-images.githubusercontent.com/54408225/215779797-9c13749f-9afa-417c-89db-579a27ba1a78.png)

4. See MM popup
5. Check tx details -- see how the ENS address is resolved to a wrong value (not an address)

**Expected behavior**
ENS Should resolve to the correct address. The address should be seen as the first param of the hex data, after the signature.
See example:

![image](https://user-images.githubusercontent.com/54408225/215780230-77280b05-2aeb-4b7e-b769-e94aae2128d5.png)

**Smartphone (please complete the following information):**

- Device: Samsung SM-A217F
- OS: Android 11.0
- App Version: v5.12.3 (1033)

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

bschorchit:
Closing as https://github.com/MetaMask/metamask-mobile/pull/5656 was merged

PR #5656, opened 2023-02-01
[FIX] ENS Resolves to Wrong Address when DeepLink

blackdevelopa:
**Development & PR Process**

**Description**
Approving an ERC20 using deep link resolves incorrect ENS. See [issue](https://github.com/MetaMask/metamask-mobile/issues/5648)

**Screenshots/Recordings**
https://recordit.co/NEnSKMPdrB

**Issue**

Progresses #5648

**Checklist**

### DIFF
--- a/app/components/Views/Send/index.js
+++ b/app/components/Views/Send/index.js
@@ -37,7 +37,7 @@ import {
   generateTransferData,
 } from '../../../util/transactions';
 import Logger from '../../../util/Logger';
-import { isENS, isValidHexAddress } from '../../../util/address';
+import { getAddress } from '../../../util/address';
 import TransactionTypes from '../../../core/TransactionTypes';
 import { MAINNET } from '../../../constants/network';
 import BigNumber from 'bignumber.js';
@@ -47,7 +47,6 @@ import AnalyticsV2 from '../../../util/analyticsV2';
 
 import { KEYSTONE_TX_CANCELED } from '../../../constants/error';
 import { ThemeContext, mockTheme } from '../../../util/theme';
-import { doENSLookup } from '../../../util/ENSUtils';
 
 const REVIEW = 'review';
 const EDIT = 'edit';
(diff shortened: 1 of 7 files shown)


app/core/DeeplinkManager.js
```diff
@@ -82,12 +86,29 @@ class DeeplinkManager {
       throw new Error('The parameter uint256 should be an integer');
 
     const value = uint256Number.toString(16);
+    let spenderAddress;
+
```
tommasini, line 90: We have this logic already [here](https://github.com/MetaMask/metamask-mobile/blob/01680ea4df25febd930b2d52289e36f0e33fa19e/app/components/Views/Send/index.js#L270) and we could re-use it here, we should also validate if the address is valid correct

It would be nice to refactor and use the same function in both places

tommasini (changes requested):
It would be nice to have the validation logic covering both scenarios on one util folder, wdyt?

tommasini (approved):
LGTM!

Just good to have:
Could we have test cases when:
* Address is not valid
* ENS is not valid

jpuri (approved):
LGTM

app/components/Views/Send/index.js
```diff
@@ -269,17 +268,10 @@ class Send extends PureComponent {
...
-      ensRecipient = recipient;
-      to = await doENSLookup(ensRecipient, this.props.network);
-    }
-    if (
-      recipient &&
-      isValidHexAddress(recipient, { mixedCaseUseChecksum: true })
-    ) {
-      to = recipient;
-    }
+    let ensRecipient;
```
jpcloureiro, line 271: `ensRecipient` is not being used anymore. Is it safe to remove it?

app/util/address/index.js
```diff
@@ -437,3 +437,19 @@ export const stripHexPrefix = (str) => {
...
+ */
+export async function getAddress(toAccount, network) {
+  if (isENS(toAccount)) {
+    return await doENSLookup(toAccount, network);
+  }
+  if (isValidHexAddress(toAccount, { mixedCaseUseChecksum: true })) {
+    return toAccount;
+  }
+  return null;
+}
```
jpcloureiro, line 455: Clean solution 👍

jpcloureiro (approved):
Code looks good to me ✅ 
Left one suggestion

app/components/Views/Send/index.js
blackdevelopa: Ah. It's used [here](https://github.com/MetaMask/metamask-mobile/blob/d2250bcc3be237c725f06b19dd51957a113ad90e/app/components/Views/Send/index.js#L284).

blackdevelopa: I was wrong. Good catch. https://github.com/MetaMask/metamask-mobile/pull/5656/commits/6d64e74a9db950274c5b5780cc090cf48b1463ce

jpcloureiro (approved):

PR #5711, opened 2023-02-06
Improve signature request message

jpuri:
Sanitising string on signature request pages

Ref: https://app.zenhub.com/workspaces/confirmations-6245e6e2348677001213b8d2/issues/gh/metamask/metamask-planning/306


jpuri:

blackdevelopa (commented):
Great PR. Do you think we should also add this to `MessageSign` as well?

jpuri:
> Great PR. Do you think we should also add this to `MessageSign` as well?

Message sign is encoded data, its is not needed there.

blackdevelopa (approved):
Lgtm

digiwand (approved):
Nice 🙌

seaona:
I've QAd the branch and when I click on Personal Sign I am seeing a crashing error saying `str.replaceAll` is not a function. See below:


https://user-images.githubusercontent.com/54408225/219611988-55975dd3-a0d9-4de5-b655-0dd5702dea64.mp4

jpuri:
@seaona : I found that it was crashing in android, PR is updated with fix.

seaona:
The error is fixed now. Thank you @jpuri  :star: 

I've realized that when we try to Verify the Signature, the verification fails and the address is not recovered. This happens only for that specific case that the original issue was reporting. For the rest of the cases the Verify Signature works fine. 

I didn't realize this is also happening with the Improve signature on Extension. It is a rare edge case but we might want to resolve it correctly too. In that case I can open a new issue for Extension.

The error message is:

```
Failed to verify signer when comparing 0x9c90317969c3a97587acb0fee81e88778fe6ee95 to 0x1c53dc20d1e36ed8359250de626acae36bd28a29
```

https://user-images.githubusercontent.com/54408225/220589230-88256dd9-f029-4b5e-a62c-6a49eafe7d18.mp4

image

jpuri:
@seaona : the issue above is not related this this PR. I would suggest that we log it separately.

PR #5724, opened 2023-02-07
[FIX] Migrate dapps access into permission controller state

jpcloureiro:
**Description**

Create a migration from all allowed hosts & selectedAddress into PermissionController state.

**Issue**
Progresses: https://github.com/MetaMask/mobile-planning/issues/631

### DIFF
--- a/app/store/migrations.js
+++ b/app/store/migrations.js
@@ -328,13 +328,19 @@ export const migrations = {
     return state;
   },
   13: (state) => {
+    // If for some reason we already have PermissionController state, bail out.
+    const hasPermissionControllerState = Boolean(
+      state.engine.backgroundState.PermissionController?.subjects,
+    );
+    if (hasPermissionControllerState) return state;
+
     const { approvedHosts } = state.privacy;
     const { selectedAddress } =
       state.engine.backgroundState.PreferencesController;
 
     const hosts = Object.keys(approvedHosts);
-
-    if (!hosts) return state;
+    // If no dapps connected, bail out.
+    if (hosts.length < 1) return state;
 
     const { subjects } = hosts.reduce(
       (accumulator, host, index) => ({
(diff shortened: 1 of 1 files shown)


app/store/migrations.js
```diff
@@ -326,6 +327,54 @@ export const migrations = {
...
+                id: random(),
+                parentCapability: 'eth_accounts',
+                invoker: host,
+                caveats: [
+                  {
+                    type: 'restrictReturnedAccounts',
+                    value: [
+                      {
+                        address: selectedAddress,
+                        lastUsed: Date.now() - index,
```
jpcloureiro, line 356: Not sure about this `Date.now() - index`. It looks hacky and might not have any benefit at all. 

We use it on the Permission Controller requests approvals when multiple accounts are added so the `lastUsed` doesn't repeat across all the accounts

gantunesr: I think its ok given the context or you could use a random `Date` since it will not reflect the real value of the last time it was used. For example,

```js
function randomDate(start, end) {
    return new Date(start.getTime() + Math.random() * (end.getTime() - start.getTime()));
}
```

[source](https://stackoverflow.com/questions/9035627/elegant-method-to-generate-array-of-random-dates-within-two-dates)

app/store/migrations.js
```diff
@@ -326,6 +327,54 @@ export const migrations = {
 
     return state;
   },
+  13: (state) => {
+    const { approvedHosts } = state.privacy;
+    const { selectedAddress } =
+      state.engine.backgroundState.PreferencesController;
```
gantunesr, line 333: Maybe I don't fully follow the logic here, but it looks like you're only doing this for one address. Is that correct?

jpcloureiro: That's right, I am maintaining dapps connection with `selectedAddress` only.
I thought migrating all wallet accounts would not be desired, but I would love to get more input on this approach.

gantunesr: Now I understand what you're trying to achieve, it makes sense to me

gantunesr (approved):

cortisiko (approved):
🌮 🌮 🌮 🌮

PR #5725, opened 2023-02-08
Show transfer view to user for approve with value

jpuri:
Improve approve transaction flow

https://app.zenhub.com/workspaces/confirmations-6245e6e2348677001213b8d2/issues/gh/metamask/metamask-planning/317


app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -355,7 +355,11 @@ const RootRPCMethodsUI = (props) => {
           });
         }
 
-        if (data && data.substr(0, 10) === APPROVE_FUNCTION_SIGNATURE) {
+        if (
+          data &&
+          data.substr(0, 10) === APPROVE_FUNCTION_SIGNATURE &&
+          (!value || value === '0x0')
```
blackdevelopa, line 361: This change looks good to me. Just a random thought, if we check for value === 0, shouldn't `isZero` work too?

jpuri: here value is hex string so I am checking this

but totally valid point, can this be BigNumber? Are you suggesting to use `isZero` from some lib.

jpuri: PR is updated

blackdevelopa (approved):
Lgtm. Audit is failing. You may need to rebase when the fix is updated to main

digiwand (approved):

seaona:
Looks good! I land into the correct screen.

https://user-images.githubusercontent.com/54408225/220321115-9e700f9e-4e32-48e8-81c4-4dd99bebed93.mp4

seaona:
@bschorchit pending to discuss RC target

PR #5796, opened 2023-02-20
Update `@metamask/phishing-controller` to v2

Gudahtt:
**Development & PR Process**

**Description**

This update should dramatically reduce network traffic and double the frequency of phishing config updates.

This was accomplished by combining both of our phishing configurations into one list (the "stalelist"), then creating a separate list of the changes just the past few days (the "hotlist"). Now users will download a smaller list more frequently (every 30 minutes rather than every hour), whereas the full list is only updated every 4 days.

**Expected Behavior**

The phishing warning functionality should behave exactly as before, except that it will update slightly more frequently and use less network bandwidth. Previously updates would be downloaded between 0-90 minutes after the change was made, and now it's between 0-60 minutes instead.

We should do performance testing to ensure it doesn't negatively impact cold start as well. I don't anticipate it will, but it's touching the same area of the codebase as the previous cold start issue.

**Checklist**


patches/@metamask+phishing-controller+1.1.0.patch
```diff
@@ -1,20 +0,0 @@
-diff --git a/node_modules/@metamask/phishing-controller/dist/PhishingController.js b/node_modules/@metamask/phishing-controller/dist/PhishingController.js
-index be85782..3756186 100644
---- a/node_modules/@metamask/phishing-controller/dist/PhishingController.js
-+++ b/node_modules/@metamask/phishing-controller/dist/PhishingController.js
-@@ -202,10 +202,14 @@ _PhishingController_inProgressUpdate = new WeakMap(), _PhishingController_instan
-         if (metamaskConfigLegacy) {
-             configs.push(metamaskConfig);
-         }
-+
-+        // Create a Set for more efficient checking when filtering phishfortHotlist.
```
Gudahtt, line 10: The patched code here was replaced in the latest version. The equivalent to this code is now here: https://github.com/MetaMask/core/blob/main/packages/phishing-controller/src/utils.ts#L28-L40

It already uses `Set`s, and we did some quick benchmarking to ensure it was the fastest method of those we thought of.

Due to the way we've restructured the phishing updates, the list sizes being processed should be much smaller now as well.

Cal-L (approved):
LGTM! Before we merge - Could you add an expected behavior section? This would help QA know exactly what to look for when spot checking.

Gudahtt:
Sure, done!

Gudahtt:
Rebased to resolve a conflict.

I added the "needs-qa" label due to the performance concern, but feel free to substitute that for the "spot check" label if you think it's more appriopriate; I don't really know the criteria for when to use which.

app/core/Engine.js
```diff
@@ -253,6 +253,7 @@ class Engine {
       });
 
       const phishingController = new PhishingController();
+      phishingController.maybeUpdateState();
```
Gudahtt, line 256: A call to `updatePhishingLists` was recently removed from this spot in #5797 to improve start time. I am unsure of the impact of that change; it might put the user at-risk of not being warned if the _very first page_ they visit is a phishing page, but I'm not sure and the risk is probably minimal.

An update call has been restored here for a few reasons:
* This `maybeUpdateState` call is conditional, and does nothing if the configuration is already sufficiently up-to-date
* Phishing config updates should be far less CPU-intensive with the new controller update
* This eliminates any risk to users for the first browser page load of that session.

Issue #5794, opened 2023-02-20
App crashes when deep-linking to external app from in-app-browser


Cal-L:
### **Description**

Discovered in 6.0.0 production release - There is a crash that exists when a user tries to deep-link to an app from within the app-browser. Ex. In certain instances - opening CoinGecko from the in-app browser will show a prompt asking to open CoinGecko's app. Once tapped, the app will crash and display and error similar to this `View: Root TypeError: Invalid URL: null`. Wallet reset is needed at this stage since app will continue to crash after re-opening. There has been reports on both iOS and Android. Note - This was reproducible by @jpcloureiro on Android but is yet to be reproduced on iOS.

### **Technical Details**

- TBD @jpcloureiro is looking into this

### **Acceptance Criteria**

- TBD - What is the expected behavior. Should the Dapp be prompting open external app in the first place? Or is this expected and the deeplink should work?

### **References**

-  Slack thread with the report - https://consensys.slack.com/archives/C029JG63136/p1676913300770899?thread_ts=1676576106.582359&cid=C029JG63136

jpcloureiro:
We should maintain the same behavior as 5.14.0 and present this screen to the user whenever they try to open these special URLs (deeplinks) from within In-App browser

<img width="387" alt="image" src="https://user-images.githubusercontent.com/1649425/220201130-537e458f-e3d0-4ef2-9751-dc1fa3112713.png">

samlogic-ship:
I have also been locked out of the mobile app 

![IMG_5547](https://user-images.githubusercontent.com/113727328/220411905-daa42202-5efd-47fa-b39a-f4ea04dc7741.png)

PR #5799, opened 2023-02-20
[HOTFIX] Browser: handle unsupported URLs

jpcloureiro:
**Description**

In-App browser do not handle special URLs (deeplinks like market://details?...)

Change from native URL API to `url-parse` package. 
This allows the app to handle unsupported URLs instead of throwing an error.

**Screenshots/Recordings**

Screen that will appear when the user tries to use a special URL / deeplink inside In-App browser
image

**Issue**

Progresses #5794

### DIFF
--- a/app/components/Nav/Main/MainNavigator.js
+++ b/app/components/Nav/Main/MainNavigator.js
@@ -69,6 +69,7 @@ import { getPermittedAccountsByHostname } from '../../../core/Permissions';
 import { TabBarIconKey } from '../../../component-library/components/Navigation/TabBar/TabBar.types';
 import { isEqual } from 'lodash';
 import { strings } from '../../../../locales/i18n';
+import isUrl from 'is-url';
 
 const Stack = createStackNavigator();
 const Tab = createBottomTabNavigator();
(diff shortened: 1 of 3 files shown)


Gudahtt:
We are already using this polyfill: https://github.com/charpeni/react-native-url-polyfill which apparently follows the whatwg standard for `URL`, which does support custom protocols. I'm curious to know which specific scenario is failing that `url-parse` fixes.

jpcloureiro:
> We are already using this polyfill: https://github.com/charpeni/react-native-url-polyfill which apparently follows the whatwg standard for `URL`, which does support custom protocols. I'm curious to know which specific scenario is failing that `url-parse` fixes.

Invalid URL error handling needs to be implemented before replacing `parse-url` with the native URL API.

In this case (market://details?src=io.metamask. ...) the deeplink gets transformed somewhere into an invalid URL in the form of `null?src=io.metamask`.

`url-parse` being designed to not throw errors (like legacy nodejs url API) keeps the app from crashing

Gudahtt:
>deeplinks are transformed into an invalid URL

Ahhh I see. I thought you meant that `new URL` was throwing directly. That makes more sense.

andreahaku (approved):
LGTM

chrisleewilcox:
@jpcloureiro what is expected if user has coingecko app installed on device?  Will coingecko app open?  Can you confirm this?  Will need this confirmed in development before handing this over to QA.

chrisleewilcox (changes requested):
Please review my comments about behavior for coingecko app on device.

jpcloureiro:
@chrisleewilcox 
In 5.14.0, the browser presents the "Something went wrong" screen regardless of having the coingecko app installed or not.

I would say this is the expected behaviour for this PR as well.

chrisleewilcox (approved):

PR #5802, opened 2023-02-21
Consolidate Bitrise setup

Gudahtt:
**Development & PR Process**
1. Follow MetaMask [Mobile Coding Standards](https://docs.google.com/document/d/1VJLwTRsUw_5EDq_o8d6sSbXUAYENLurkRitYO45eY5o/edit?usp=sharing)
2. Add `release-xx` label to identify the PR slated for a upcoming release (will be used in release discussion)
3. Add `needs-dev-review` label when work is completed
4. Add `needs-qa` label when dev review is completed
5. Add `QA Passed` label when QA has signed off

**Description**

The Bitrise config previously had two "code setup" steps, where Yarn dependencies were installed. They were identical except that one had caching and the other didn't (though the caching was broken, so they were in effect identical).

They have been consolidated into a single step, and the caching has been removed for now.

**Issue**

This was done to make it easier to add [preview build](https://github.com/MetaMask/core/blob/main/docs/contributing.md#using-packages-in-other-projects-during-developmenttesting) support to Bitrise, allowing us to generate QA builds and run e2e tests using preview builds. I don't think there is an issue for this yet, and I'm not sure which work it most directly relates to.

**Checklist**

* [ ] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

PR #5811, opened 2023-02-21
Add Bitrise caching

### LINKED ISSUE #5810: Cache Yarn dependencies in Bitrise
### **Description**

We should cache dependencies in Bitrise to improve speed and reliability of CI workflows.

### **Technical Details**

We should ensure this cache does not make the build difficult to reproduce, which would make auditing difficult. The build must be reproducible, at least to the degree that the lockfile allows.

Caching `node_modules` could result in a difficult-to-reproduce build because it could lead to a different directory structure when the build runs, and it could cache the output of `postinstall` steps that aren't intended to be run multiple times.

However it should be save to cache the Yarn cache because it doesn't affect hoisting, and doesn't include any content that gets modified during the install process.

### **Acceptance Criteria**

- Dependencies are cached between CI runs
- Builds are no more difficult to reproduce locally than they were before
  - e.g. we're caching the Yarn cache directory, not `node_modules`

Gudahtt:
**Development & PR Process**

**Description**

Caching has been added to our Bitrise CI config. It uses the [legacy branch caching API](https://devcenter.bitrise.io/en/builds/caching/branch-based-caching.html) because our current plan does not support the newer [key-based caching API](https://devcenter.bitrise.io/en/builds/caching/key-based-caching.html).

Typically the way caching works on Bitrise is that each step will have internal steps to "stage" data for caching, and the explicit cache step saves all of this data in the end. However the `yarn` step doesn't stage things correctly for us for two reasons:
* It only supports caching the `install` command, but we use `setup`
* It caches the `node_modules` directory, but we want to cache the Yarn cache directory.

Caching `node_modules` could result in a difficult-to-reproduce build because it could lead to a different directory structure when the build runs, and it could cache the output of `postinstall` steps that aren't intended to be run multiple times. However it should be save to cache the Yarn cache because it doesn't affect hoisting, and doesn't include any content that gets modified during the install process.

**Issue**

Fixes #5810

**Checklist**

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -32,7 +32,7 @@ workflows:
           inputs:
             # The Yarn cache is easier and safer to cache than `node_modules`, which could include
             # postinstall script modifications and could affect hoisting on subsequent installs.
-            - cache_paths: "$YARN_CACHE_DIR"
+            - cache_paths: "$YARN_CACHE_DIR -> ./yarn.lock"
       - yarn@0:
           title: Lint
           inputs:
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -27,12 +27,6 @@ workflows:
           inputs:
             - command: setup
           title: Yarn Setup
-      - cache-push@2:
-          # Manual cache path needed here because the `yarn` step only caches for the `install` command
-          inputs:
-            # The Yarn cache is easier and safer to cache than `node_modules`, which could include
-            # postinstall script modifications and could affect hoisting on subsequent installs.
-            - cache_paths: "$YARN_CACHE_DIR -> ./yarn.lock"
       - yarn@0:
           title: Lint
           inputs:
@@ -41,6 +35,12 @@ workflows:
           inputs:
             - command: audit:ci
           title: Audit Dependencies
+      - cache-push@2:
+          # Manual cache path needed here because the `yarn` step only caches for the `install` command
+          inputs:
+            # The Yarn cache is easier and safer to cache than `node_modules`, which could include
+            # postinstall script modifications and could affect hoisting on subsequent installs.
+            - cache_paths: "$YARN_CACHE_DIR -> ./yarn.lock"
   code_setup_dev:
     before_run:
       - setup
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -16,10 +16,23 @@ workflows:
     before_run:
       - setup
     steps:
+      - script@1:
+          inputs:
+            - content: |-
+                #!/usr/bin/env bash
+                envman add --key YARN_CACHE_DIR --value "$(yarn cache dir)"
+          title: Get Yarn cache directory
+      - cache-pull@2: {}
       - yarn@0:
           inputs:
             - command: setup
           title: Yarn Setup
+      - cache-push@2:
+          # Manual cache path needed here because the `yarn` step only caches for the `install` command
+          inputs:
+            # The Yarn cache is easier and safer to cache than `node_modules`, which could include
+            # postinstall script modifications and could affect hoisting on subsequent installs.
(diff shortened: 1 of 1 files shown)


Gudahtt:
This depends upon #5802

bitrise.yml
```diff
@@ -16,10 +16,23 @@ workflows:
...
+            - content: |-
+                #!/usr/bin/env bash
+                envman add --key YARN_CACHE_DIR --value "$(yarn cache dir)"
+          title: Get Yarn cache directory
+      - cache-pull@2: {}
       - yarn@0:
           inputs:
             - command: setup
           title: Yarn Setup
+      - cache-push@2:
```
NicolasMassart, line 30: How does ["The **Bitrise.io Cache:Push** Step should be the very last Step in your Workflow."](https://github.com/bitrise-steplib/steps-cache-push/blob/f1aa51a651f8b4076f61a1f099ee0cc954711634/step.yml#L8) apply to our script here? Any risk not to have it in a "run_after" step?

bitrise.yml
```diff
@@ -16,10 +16,23 @@ workflows:
...
+          # Manual cache path needed here because the `yarn` step only caches for the `install` command
+          inputs:
+            # The Yarn cache is easier and safer to cache than `node_modules`, which could include
+            # postinstall script modifications and could affect hoisting on subsequent installs.
+            - cache_paths: "$YARN_CACHE_DIR"
```
NicolasMassart, line 35: Do we have to use the `yarn.lock` as an "update indicator file" here?

bitrise.yml
Gudahtt: I had assumed that was meant to ensure nothing was missing from the cache. But the steps after this one do not need to be cached.

I guess we could move this until last anyway. I doubt the dependency audit or linter would modify the Yarn cache.

bitrise.yml
Gudahtt: Ah, good catch, that does seem like the right thing to do here.

```suggestion
            - cache_paths: "$YARN_CACHE_DIR -> ./yarn.lock"
```

DigitalsupportTeam:

bitrise.yml
Gudahtt: Moved to the last step here: https://github.com/MetaMask/metamask-mobile/pull/5811/commits/932278d70f972def42c0216289edee7c9d25e121

Gudahtt:
Tested in Bitrise. Here is where the cache was created: https://app.bitrise.io/build/11996e8f-44ec-4a49-b975-ae3aece4b411

And here in a later build, the cache was used: https://app.bitrise.io/build/e971851a-526c-45fc-8d7e-f166e34aec2c

sethkfman (approved):
LGTM

NicolasMassart:
> Tested in Bitrise. Here is where the cache was created: https://app.bitrise.io/build/11996e8f-44ec-4a49-b975-ae3aece4b411
> 
> And here in a later build, the cache was used: https://app.bitrise.io/build/e971851a-526c-45fc-8d7e-f166e34aec2c

Have you tried running this since [the aborted build](https://app.bitrise.io/app/be69d4368ee7e86d?branch=add-bitrise-caching)?

Gudahtt:
Nope, that was the most recent test I've done. I don't recall why that was aborted.... I will kick off another test

Gudahtt:
I rebased this branch and triggered a new Bitrise build here: https://app.bitrise.io/build/efc66d9a-a37c-4e93-8298-d7fd264fc1e7

It passed, but it was not substantially faster than it was without caching when you account for the time it took to download the cache itself (the jobs with a cache hit took basically the same amount of time as the install job on `main` today, with the cache miss taking an extra ~2 minutes to populate the cache).

PR #5793, opened 2023-02-20
E2e appium/send native and erc tokens

Andepande:
**Description**
- SendToken Flow originating from this [ticket](https://github.com/MetaMask/mobile-planning/issues/546)
- Increased Appium test coverage 
- Minor implementations to improve test runs

**ScreenShots**
image

Failing tests
/wdio/features/ExploringWizard.feature - Passed locally
/wdio/features/BrowserFlow/ENSWebsite.feature - Last step Browser tests currently on pause
/wdio/features/AppLaunchTime.feature - Fails due to biometric screen( Passes when run in Isolation
/wdio/features/AddressFlow.feature - Fails due to no Eth on test account
/wdio/features/ImportCustomToken.feature Fails on my machine due to lack of importing Goerli token, passes on other machines

**Issue**

Progresses #???

**Checklist**

* [x] There is a related GitHub issue
* [x] Tests are included if applicable
* [ ] Any added code is fully documented

PR #5825, opened 2023-02-23
[Improvement] Send Token E2E test

Andepande:
**Description**
- SendToken Flow originating from this [ticket](https://github.com/MetaMask/mobile-planning/issues/546)
- Increased Appium test coverage 
- Minor implementations to improve test runs
- PR based on deprecated PR [5793](https://github.com/MetaMask/metamask-mobile/pull/5793) with resolved comments
**ScreenShots**
Local test run:
image

Browserstack test successful runs X3:
1 - https://app-automate.browserstack.com/dashboard/v2/builds/be5310c82607f6abeeb859c3db73cfc6112c2f55/sessions/5c13b2805691db2580f319962941791e4ad89fde
2 - https://app-automate.browserstack.com/dashboard/v2/builds/be5310c82607f6abeeb859c3db73cfc6112c2f55/sessions/e2dae7067d83f8031c2ddad4ac7479105997448a
3- https://app-automate.browserstack.com/dashboard/v2/builds/be5310c82607f6abeeb859c3db73cfc6112c2f55/sessions/a94b38101f748c07620e7777e4ecc2f21ea3a007

Failing tests

**Issue**

Progresses #???

**Checklist**


wdio.conf.js
```diff
@@ -115,7 +115,7 @@ export const config = {
   baseUrl: 'http://localhost',
   //
   // Default timeout for all waitFor* commands.
-  waitforTimeout: 220000,
+  waitforTimeout: 45000,
```
NicolasMassart, line 118: It's great if we can reduce timeouts this much. what made this possible?

wdio/features/SendToken.feature
```diff
@@ -0,0 +1,115 @@
...
+            | AVAX Fuji | https://api.avax-test.network/ext/C/rpc | 43113   | AVAX   |
+
+    Scenario Outline: Import Custom Token
+        When I tap on the navbar network title button
+        And I tap on <NETWORK> on Networks list to switch
+        And I tap on Got it in the network education modal
+
+        When I tap Import Tokens
+        And I type <TOKENADDRESS> into token Address field
+        Then The Token Symbol is displayed
```
NicolasMassart, line 41: We only tap to focus on token symbol field here, but we don't check the symbol is actually displayed, right?
Doing so would prevent the pause in the next "I tap on the Import button" step.
Am I missing something here?

wdio/features/SendToken.feature
```diff
@@ -0,0 +1,115 @@
...
+        When On the Main Wallet view I tap "Send"
+        Then I enter address "<Address>" in the sender's input box
+
+        When I tap button "Next" on Send To view
+        Then I proceed to the amount view
+
+        When I type amount "<AMOUNT>" into amount input field
+        And I tap button "Next" on the Amount view
+        Then I should be taken to the transaction confirmation view
+        And the token <TOKEN> being sent is visible
```
NicolasMassart, line 63: The amount will change for each test if we send tokens from the same address, right?

wdio/features/SendToken.feature
```diff
@@ -0,0 +1,115 @@
...
+        And the token amount <AMOUNT> to be sent is visible
+
+        When I tap button "Send" on Confirm Amount view
+        Then Sending token takes me to main wallet view
+        And the transaction is submitted with Transaction Complete! toast appearing
+
+        Examples:
+            | NETWORK   | TOKEN | AMOUNT | Address                                    |
+            # | Goerli            | ETH   | 0.0004 | 0xebe6CcB6B55e1d094d9c58980Bc10Fed69932cAb |
```
NicolasMassart, line 72: Why commenting this case?

wdio/features/SendToken.feature
```diff
@@ -0,0 +1,115 @@
...
+    Scenario Outline: A user can send ERC-20 tokens to an Address via token overview screen
+        Given I am on the wallet view
+        And I see "<NETWORK>" visible in the top navigation bar
+        When I tap Token "<TOKEN>" on Main Wallet view
+        Then I am taken to the token overview screen
+
+        When I tap button "Send" on Token screen view
+        And I enter address "<Address>" in the sender's input box
+        And I tap button "Next" on Send To view
+        And I type amount "<AMOUNT>" into amount input field
```
NicolasMassart, line 84: this case requires token balance to be >, how do we prevent it to fail?

wdio/features/SendToken.feature
```diff
@@ -0,0 +1,115 @@
...
+
+        When I tap button "Next" on Send To view
+        And I type amount "<AMOUNT>" into amount input field
+        And I tap button "Next" on the Amount view
+        Then "Insufficient funds" is visible
+
+        Examples:
+            | NETWORK   | TOKEN  | AMOUNT                | Address                                    |
+            # | Goerli            | LINK  | 121212121212121212121  | 0xA7E129A38860432492708568465AE1A5ca50b373 |
+            | AVAX Fuji | 1 USDC | 121212121212121212121 | 0xebe6CcB6B55e1d094d9c58980Bc10Fed69932cAb |
```
NicolasMassart, line 115: Maybe we could try a more realistic amount? I know it depends on the wallet balance but if we check first, we can find a value that's over the balance and not a hardcoded one?

wdio/screen-objects/AmountScreen.js
```diff
@@ -0,0 +1,37 @@
...
+        expect(await element.getText()).toContain(token);
+      }
+    
+      async isCorrectTokenAmountDisplayed(amount) {
+        const element = await this.confirmAmount;
+        expect(await element.getText()).toContain(amount);
+      }
+
+      async enterAmount(text) {
+        await Gestures.tap(this.amountInputField);
```
NicolasMassart, line 33: We should check the visibility of the field before tapping in.

wdio/step-definitions/import-tokens.steps.js
```diff
@@ -1,8 +1,9 @@
+/* global driver */
 import { When, Then } from '@wdio/cucumber-framework';
 import AddCustomImportTokensScreen from '../screen-objects/AddCustomImportTokensScreen.js';
 import WalletMainScreen from '../screen-objects/WalletMainScreen.js';
 
-let setTimeout = 1500; //added to run on physical device
+const setTimeout = 1500; //added to run on physical device
```
NicolasMassart, line 6: I know it's not the change you made here, it's just the const, but any way we could prevent the pause (for instance on line 9) with wait for custom token address field to be visible instead?

wdio/features/SendToken.feature
Andepande: Makes sense, we could check the current value and add 1 maybe

wdio/step-definitions/import-tokens.steps.js
Andepande: So there are some instances where an element would be available in the dom before it can be tapped, but appium would just tap that and move on to the next statement where it would fail.

Andepande: We can try asserting for specific attributes like whether an element is clickable before it taps but I believe when we tried that before it wasn't consistent

wdio/screen-objects/AmountScreen.js
Andepande: Implemented in Tap function within the Gestures class

cortisiko: You should probably use `Gestures.waitAndTap`

wdio/helpers/Gestures.js
```diff
@@ -98,6 +99,25 @@ class Gestures {
...
+        break;
+      case 'RELEASE':
+        await elem.touchAction(Actions.RELEASE);
+        break;
+      default:
+        throw new Error('Tap type not found');
+    }
+  }
+
+  static async tapTextContains(text, tapType = 'TAP') {
```
cortisiko, line 118: What do you think about renaming to either `tapByTextContaining` or `tapByStringContainingText`

wdio/screen-objects/AmountScreen.js
```diff
@@ -0,0 +1,32 @@
...
+class AmountScreen {
+  get amountInputField() {
+    return Selectors.getElementByPlatform(TRANSACTION_AMOUNT_INPUT);
+  }
+
+  get amountScreen() {
+    return Selectors.getElementByPlatform(AMOUNT_SCREEN);
+  }
+
+  async isCorrectTokenConfirm(token) {
```
cortisiko, line 17: perhaps rename to `isTokenCorrect`

wdio/screen-objects/AmountScreen.js
```diff
@@ -0,0 +1,32 @@
...
+    const element = await this.confirmAmount;
+    expect(await element.getText()).toContain(token);
+  }
+
+  async isCorrectTokenAmountDisplayed(amount) {
+    const element = await this.confirmAmount;
+    expect(await element.getText()).toContain(amount);
+  }
+
+  async enterAmount(text) {
```
cortisiko, line 27: put the test action above the assertions

wdio/screen-objects/TokenOverviewScreen.js
```diff
@@ -0,0 +1,28 @@
...
+    get backButtonTokenOverview() {
+      return Selectors.getElementByPlatform(ASSET_BACK_BUTTON);
+    }
+  
+    async isTokenOverviewVisible() {
+      const element = await this.tokenAssetOverview;
+      await element.waitForDisplayed();
+    }
+  
+    async tapBackButton() {
```
cortisiko, line 23: put the test action before the assertion

wdio/step-definitions/common-steps.js
```diff
@@ -89,13 +90,19 @@ Then(/^"([^"]*)?" is visible/, async (text) => {
...
+  await CommonScreen.isTextDisplayed(text);
+});
+
 Then(/^"([^"]*)?" is not displayed/, async (text) => {
   const timeout = 1000;
   await driver.pause(timeout);
   await CommonScreen.isTextElementNotDisplayed(text);
 });
 
-Then(/^I am on the main wallet view/, async () => {
```
cortisiko, line 98: changing this step might cause other tests to fail. Can we confirm this is not the case?

cortisiko (changes requested):
left a few comments. We will need to put our wallet keys behind an env variable to prevent loss of funds. We can discuss it tomorrow.

cortisiko:
A couple more things: 

- Rename this PR. `[Improvement] Send Token E2E test`
- Fix the listing issues.

DigitalsupportTeam:

wdio/features/SendToken.feature
Andepande: Yeah, the amount will change each time we send from that account, an idea I thought of was to have a hook that would transfer the amount back after the test has completed.

wdio/features/SendToken.feature
Andepande: Commented for now as there are no funds on the goerli account therefore the test will fail

wdio/features/SendToken.feature
Andepande: Maybe a close loop test that would send the funds back?

wdio/helpers/Gestures.js
Andepande: yeah sure, thought maybe it was too long but I'll change it to that

wdio.conf.js
Andepande: Browserstack had a delay in starting the app, therefore tests would begin before the app, Found some solutions with @SamuelSalas that would dynamically wait for pages to exist/Disappear which enabled the reduction of this

wdio/features/SendToken.feature
Andepande: true, it was missing the assertion which is now added, I think someone was added the extra pause while running on a physical device as it was failing without that extra pause, but never removed the pause in the PR.

wdio/step-definitions/common-steps.js
Andepande: That is not the case, it was a duplicate step

wdio/helpers/Gestures.js
```diff
@@ -98,6 +99,25 @@ class Gestures {
...
+        throw new Error('Tap type not found');
+    }
+  }
+
+  static async tapByTextContaining(text, tapType = 'TAP') {
+    const elem = await Selectors.getXpathElementByTextContains(text);
+    await elem.isDisplayed();
     switch (tapType) {
       case 'TAP':
         await elem.touchAction(Actions.TAP);
```
SamuelSalas, line 123: Mayber outside of scope, but i see that there are a lot of tap methods that practically do the same. Just with small changes.

wdio/screen-objects/ChangePasswordScreens.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-undef */
```
SamuelSalas, line 1: any reason to add this exception?

wdio/helpers/Gestures.js
Andepande: Yeah there, so getXpathElementByText() taps by text with the exact text, which is good for static text but for some buttons that have dynamic text like on the main wallet view if you want to tap on a specific token  e.g (1 USDC) without knowing the amount, using getXpathElementByTextContains() was more usefull.

wdio/screen-objects/ChangePasswordScreens.js
Andepande: ES-Lint expection for driver.HideKeyboard on line 29

wdio/helpers/Gestures.js
SamuelSalas: So, if the selectors methods are outside the tap method, the method is going to fail? Like, instead of having the text param, to have a  element param in the taptext methods?

Andepande: E.g On the main wallet view, the test has to tap on the USDC Token button which contains text "USDC" however there is always a prefix which is dynamic, like "1" or "2" USDC, but therefore if I use tapTextByXpath() to tap on USDC it won't work because you need to account for the prefix. But using getXpathElementByTextContains() would enable you to select the button by declaring the text it contains 'USDC'

SamuelSalas (approved):

cortisiko (approved):
🌮 🌮 🌮 🌮

PR #5833, opened 2023-02-24
Adding e2e for simpe send flow using default gas

### LINKED ISSUE #5695: [e2e] Send ETH from inside MetaMask using default gas

jpuri:
Fixes: #5695

Adding e2e for simple send flow using default gas.

The e2e succeeded only after I added some GoreliETH to the account. I am not sure what is right way to get an account with balance.

cc @seaona , @cortisiko


e2e/viewHelper.js
```diff
@@ -0,0 +1,60 @@
...
+    //
+  }
+};
+
+export const switchToGoreliNetwork = async () => {
+  await WalletView.tapNetworksButtonOnNavBar();
+  await NetworkListModal.changeNetwork(GOERLI);
+  await WalletView.isNetworkNameVisible(GOERLI);
+  await NetworkEducationModal.tapGotItButton();
+};
```
jpuri, line 60: These helpers will be useful as we plan to write more e2e for send flow.

e2e/specs/send-eth-flow.spec.js
```diff
@@ -0,0 +1,56 @@
...
+import TransactionConfirmationView from '../pages/TransactionConfirmView';
+import WalletView from '../pages/WalletView';
+import {
+  importWalletWithRecoveryPhrase,
+  switchToGoreliNetwork,
+} from '../viewHelper';
+
+const MYTH_ADDRESS = '0x1FDb169Ef12954F20A15852980e1F0C122BfC1D6';
+
+describe('Addressbook Tests', () => {
```
cortisiko, line 15: rename the test description

e2e/viewHelper.js
cortisiko: Yup yup, I like this solution.

e2e/specs/send-eth-flow.spec.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
```
cortisiko, line 1: can you create a subfolder called "Confirmations" in the specs folder, this way we put all the confirmations e2e tests

cortisiko (changes requested):

e2e/specs/send-eth-flow.spec.js
jpuri: Good Idea I implemented this.

DigitalsupportTeam:

cortisiko (approved):
looks good to me. We are going to hide the SRP used in the tests behind an env variable. You can track the ticket [here](https://github.com/MetaMask/mobile-planning/issues/672).

Issue #5759, opened 2023-02-14
[Bug] Hex data is not cleared when chaging a Token Transfer to Send ETH tx


seaona:
**Describe the bug**
When I am on the Confirmation screen for a Send ERC20 transaction, if I go back to Edit it and change the Asset to ETH (or Goerli ETH) and proceed, I see that the Hex Data from the previous Token Transfer is not cleared.

**Screenshots**

https://user-images.githubusercontent.com/54408225/218755988-3644d02a-b7aa-48c8-962c-399bddc148ec.mp4

**To Reproduce**
_Steps to reproduce the behavior_
0. Enable Hex Data from Advanced Settings
1. Import any ERC20 Token (you can use LINK faucet if you don't have any https://faucets.chain.link/)
2. Click on Send ERC20 Token
3. Select recipient
4. Select Amount
5. Proceed to Confirmation screen
6. Click Edit
7. Change Asset to ETH/GoerliETH
8. Proceed to Confirmation screen
9. Check how Hex Data is not emptied

**Expected behavior**
_A clear and concise description of what you expected to happen_
The Hex Data field should be cleared as we have updated the tx to be a Send ETH tx, where hex should be empty by default.

**Smartphone (please complete the following information):**
Device: Samsung SM-A217F
OS: Android 11.0
App Version: v5.14.0 (1055)

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

bschorchit:
question that came up during planning: what happens if you change back to a ERC-20 transaction?

bschorchit:
> question that came up during planning: what happens if you change back to a ERC-20 transaction?

Tested now and it gets updated correctly. So the only fix needed is clearing up hex data for native token sends.

DigitalsupportTeam:
We'd be happy to look into this for you
kindly file a support ticket below 
https://forms.gle/Wmpm6NLzYm9uXYuU9

PR #5839, opened 2023-02-27
Clear Hex data when Token Transfer reverts ETH

blackdevelopa:
**Development & PR Process**

**Description**
When you send ETH the hex data should be empty by default (0x). For token transfers, we are interacting with contacts, so the Hex Data is then filled with the long values (representing the method, and the arguments passed). If we switch back to ETH, it should clear the hex data to default (0x)

**Screenshots/Recordings**
http://recordit.co/Alt9uBAOz0

**Issue**

Progresses #5759 

**Checklist**

### DIFF
--- a/app/components/Views/SendFlow/Amount/index.js
+++ b/app/components/Views/SendFlow/Amount/index.js
@@ -89,6 +89,7 @@ import {
   selectProviderType,
   selectTicker,
 } from '../../../../selectors/networkController';
+import { PREFIX_HEX_STRING } from '../../../../constants/transaction';
 
 const KEYBOARD_OFFSET = Device.isSmallDevice() ? 80 : 120;
 
@@ -704,7 +705,7 @@ class Amount extends PureComponent {
     }
 
     if (selectedAsset.isETH) {
-      transactionObject.data = '0x';
+      transactionObject.data = PREFIX_HEX_STRING;
       transactionObject.to = transactionTo;
     }
 
--- a/app/components/Views/SendFlow/Amount/index.js
+++ b/app/components/Views/SendFlow/Amount/index.js
@@ -703,7 +703,10 @@ class Amount extends PureComponent {
       transactionObject.readableValue = value;
     }
 
-    if (selectedAsset.isETH) transactionObject.to = transactionTo;
+    if (selectedAsset.isETH) {
+      transactionObject.data = undefined;
+      transactionObject.to = transactionTo;
+    }
 
     setTransactionObject(transactionObject);
   };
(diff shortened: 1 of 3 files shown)


jpuri (approved):

seaona:
QAd, comments below:

- I've noticed that when we try to Deploy a Contract, now we get the error `to address:undefined` and cannot proceed

https://user-images.githubusercontent.com/54408225/222171961-4efc78b0-e194-45c4-b681-2f636655678e.mp4

- I think there is a problem with gas calculation when we switch from one type of Send to the other. It looks like gas is not re-calculated resulting in non-desired behaviors for both cases. These issues are in production so I think we can treat them separately. I have opened [a dedicated ticket for that](https://github.com/MetaMask/metamask-mobile/issues/5865), so we don't have to fix this in this PR @blackdevelopa @bschorchit .

blackdevelopa:
Thanks @seaona. I made an update and here's a screen recording.
http://recordit.co/3iHCiaZLVg

jpuri (approved):

jpuri:
Changes look good 👍

bschorchit:
Following our new best practices, this needs a second review before going back to QA and being merged.

app/components/Views/SendFlow/Amount/index.js
```diff
@@ -699,7 +699,10 @@ class Amount extends PureComponent {
       transactionObject.readableValue = value;
     }
 
-    if (selectedAsset.isETH) transactionObject.to = transactionTo;
+    if (selectedAsset.isETH) {
+      transactionObject.data = '0x';
```
tommasini, line 703: Can we turn '0x' in a constant with the name "EMPTY_ADDRESS" or other for we know what represents?
Also, do you think it's better to comment on the code that if it's an ETH transaction there is no need for transaction data? (For me at least that is not obvious 🙏)

tommasini (approved):
LGTM!

app/components/Views/SendFlow/Amount/index.js
blackdevelopa: Thanks. `0x` is now a constant. https://github.com/MetaMask/metamask-mobile/pull/5839/commits/715a77f7116ea7d287840a2f768aafb86203df2c

seaona:
QA'd again and I can see that the issue is fixed. 
The Hex Data is cleared successfully after updating transactions.

https://user-images.githubusercontent.com/54408225/226586586-237c37ec-3132-4981-80e5-ae08c656ffc9.mp4

PR #5886, opened 2023-03-02
fix: bundle video in app to prevent crash when not available

### LINKED ISSUE #5497: Bundle the onboarding SRP explaining video in the app build
@wachunei noticed that the specific Security & Privacy section takes a while to load because of the seedphrasevideo. He suggested "to defer that to after the navigation animation has finished so the UI does no feel laggy or add a screenshot as a thumbnail instead of autoplaying, maybe autoplay the first time ever. Or at least avoid doing it over cellular data perhaps."

We can optimize this critical video in the build to 1. avoid external calls and 2. reduce the loading time in that critical screen.

Extension is already bundling the video with the build. 

I don't know how it affects `l10n`. A developer should scope it before working on it.

l10n tracks are loaded separately and played in a overlay, here’s the video source 
https://github.com/MetaMask/metamask-mobile/blob/main/app/videos/recovery-phrase.mp4?raw=true

subs 
https://github.com/MetaMask/metamask-mobile/tree/main/app/videos/subtitles/secretPhrase

NicolasMassart:
## Development & PR Process

## Description

Make the SRP video bundled in the app instead of retrieved from Github repos.

App crashed when video player tried to load the remote video and remote was not available (or network down or anything)

- Bundling the video in the app to prevent crash when remote not available
- Keep remote subtitles as react-native-video only allow loading with remote url
- Reduced video size from 5.7Mb to 1.9Mb with ffmpeg:
  ```bash
  #!/usr/bin/env sh

  VIDEO_BITRATE="75k"
  AUDIO_BITRATE="80k"

  ffmpeg -y -i recovery-phrase-source.mp4 -pass 1 -c:v libx264 -b:v $VIDEO_BITRATE -b:a $AUDIO_BITRATE -an -f null /dev/null && \
  ffmpeg -i recovery-phrase-source.mp4 -pass 2 -c:v libx264 -b:v $VIDEO_BITRATE -b:a $AUDIO_BITRATE recovery-phrase-mobile-2pass-optimized.mp4

  ```
image
The video is displayed by default on its original format on portrait mobile, meaning this video is most of the time around 5x3cm, very hard to see the quality diff when not really looking for it. Fullscreen makes it a bit more visible though.
The source video is still in the repos (but not bundled in the app) and the above script allows to compress with higher bitrates if requested.

### Source video

https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase-source.mp4?raw=true

### Compressed video

https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase-mobile-2pass-optimized.mp4?raw=true

## App size impact
The app size is only 1.9 additional Mb for the video. Makes it arround 27.7Mb for iOS and 274.6Mb for Android.

## Issue

Relates to MetaMask/mobile-planning#471
fixes #5497

## Testing

- The parts that were unit testable are updated.
- Video testing has been donne manually as no e2e exists to test the video and previous attempts to handle the player in e2e failed.

Here are the tests done:
- Using a local http server I controlled access to the subtitles and was able to return errors or make it not respond: lack of access to subtitles only prevents them to show but doesn't crash and video plays fine.
- No network at all (plane mode) as required in issue scenario `GIVEN: User has NO access to internet (airplane mode)`: the app has a system preventing it to be used and requires the user to turn network on, so not relevant.
- Slow network: subtitles appear with no noticeable delay as they are starting only at 00:00:00.780 so the loading time is before the display.
- Tested video from the multiple places it plays in the app: onboarding, settings.

### Test artifacts
- QA builds available on Bitrise public install pages:
  - [Android QA build 1072](https://app.bitrise.io/app/be69d4368ee7e86d/build/97676cfa-af33-46a6-bba7-8c155046bf99/artifact/1dc43aba4f90890c/p/9a60e41200453600caeb8a3cb033f2cb)
  - [iOS QA build 1072](https://app.bitrise.io/app/be69d4368ee7e86d/build/f91e3a5e-dc69-4a9e-8307-e62c3e31b17e/artifact/78bfcb8fbfa147bd/p/cfea43fb50293e51f4427a9bfa7da477)

## TODO

test that the video player is not broken for other usecases (NFTs) -> requires a video NFT in the test wallet, will create and drop one for testing.

## Checklist


app/util/video/index.ts
```diff
@@ -1,11 +1,22 @@
...
-  'https://github.com/MetaMask/metamask-mobile/blob/main/app/videos/';
-const subtitlePath = 'subtitles/secretPhrase/subtitles-';
-const ext = '.vtt?raw=true';
-const path = `${baseUrl}${subtitlePath}`;
+import _ from 'lodash';
+import videoSourceUri from '../../videos/recovery-phrase-mobile-2pass-optimized.mp4';
 
+// Remote path for subtitles that can't be bundled with the app
+// because only remote subtitles are supported by react-native-video
+const baseRemoteSubtitlesPath =
```
sethkfman, line 6: You may want to move these to the constants/url.js file

NicolasMassart: moved as `SRP_SUBTITLES_URL_TEMPLATE`

sethkfman (approved):
LGTM

holantonela:
Closes https://github.com/MetaMask/metamask-mobile/issues/5497

Andepande (commented):
- Tested on physical Pixel 6 
- Tested on Browserstack devices (IOS/Android) while throttling network
- QA PASSED

PR #5887, opened 2023-03-02
To address should be visible on contract interaction pages

### LINKED ISSUE #4622: Transaction confirmation screen doesn't show recipient address!!!!
This is the confirmation screen for a transaction.

This can be a transaction initiated by a webpage that I have opened in the in-app browser and that is using the MetaMask javascript API, or it can be a transaction initiated from another device connected via WalletConnect.

Regardless, the recipient address, i.e. the address I'm sending the freaking money to, should be shown. You are only showing the sender address.

image

jpuri:
Fixes: #4622

To address should be visible on transactions.

QA Steps:
------------------------------
- for different types of transactions where `to` address exists it should be visible on confirmation screen.


jpuri:
Hey @holantonela : please see pages below:

Only from address (there is a recent change in this not related to this PR):
Screenshot 2023-03-03 at 12 57 07 AM

From and to addresses:
Screenshot 2023-03-03 at 3 40 43 AM

app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.tsx
```diff
@@ -0,0 +1,208 @@
+import React, { useEffect, useState } from 'react';
```
jpuri, line 1: I have extracted out from / to address header from send page so that it can be re-used.

app/components/UI/AddressInputs/index.js
```diff
@@ -6,23 +6,23 @@ import {
   TouchableOpacity,
```
jpuri, line 6: This file is only moved out of send view.

app/components/UI/ApproveTransactionHeader/ApproveTransactionHeader.tsx
```diff
@@ -1,13 +1,10 @@
 import React, { useState, useEffect, useMemo } from 'react';
```
jpuri, line 1: Here I have extracted code to show site origin into a different component so that it can be re-used.

holantonela:
I'm confused. Why we do have these 2 different headers in the same interaction? Which one is the one that is rendering this PR?

Screen Shot 2023-03-05 at 10 23 40

bschorchit:
> I'm confused. Why we do have these 2 different headers in the same interaction? Which one is the one that is rendering this PR?

Another PR was merged while this PR was in the works. The other PR updated the origin pill and added the new header to this screen (screenshot). 
Screenshot 2023-03-06 at 00 13 10


The picture you attached on the left is Jyoti's original solution before the other PR was merged and the picture on the right is the solution after the new PR. My opinion is that we should update the new solution to not hide the new header so we have consistency with other screens and also because this is the place we're showing network information. How should we solve this?

bschorchit:
For reviewers: further conversation on the comments from Antonela and I above was done in slack and in the original issue as comments. Antonela gave her 👍 to the new changes here https://github.com/MetaMask/metamask-mobile/issues/4622#issuecomment-1462308896

blackdevelopa (approved):
lgtm

segun (approved):

seaona:
I'll post the QA findings here (will extend the comment).
### 1. The application breaks with the error `undefined is not an object (evaluating address.toLowerCase)` whenever trying to Send an ERC20 token. 
**Repro steps:**
1. Deploy an ERC20 Token from test dapp
2. Add token to Wallet
3. Go to MetaMask
4. Send TST Token to another address
5. Click Confirm -- you'll land on the error page

https://user-images.githubusercontent.com/54408225/226389483-5f1a8c3b-e02b-4034-bed9-643f4f35cdd6.mp4

### 2. Sending an ERC20 token from a dapp throws Warning  `unhandled promise rejection` whenever trying to Send an ERC20 token. 
**Repro steps:**
1. Deploy an ERC20 Token from test dapp
2. Add token to Wallet
3. Click Transfer -- see warning appearing

https://user-images.githubusercontent.com/54408225/226389466-223f62ea-91b8-45d9-b0ad-1421723bd883.mp4

jpuri:
Hey @seaona : thanks for the detailed testing. I fixed the errors and warnings, can you plz re-test.

About point-3 **Address saved on Contact book is not displayed with its contact name but as an address**

I re-used the address from / to component from send page which always show to address as raw address even if it is saved by a name. But this can be a good improvement, can you plz suggest @bschorchit

bschorchit:
I agree this would be a really nice improvement overall! On the first screen of the send flow (the one where you pick the address you're sending to), we display both nickname and and address. Could we use that same pattern?

Should this be a new issue or part of this same PR?

jpuri:
Thanks @bschorchit , @seaona : this can be done in this PR itself.

jpuri:
@bschorchit , @seaona : there was a condition in current code to show name for `to address` only if it was `ens` name. I am not sure what was reason for this check. Thus not to account name should be visible if it is saved along with the address also.

Screenshot 2023-03-21 at 11 21 46 PM

seaona:
I've reviewed the issues and I all of them fixed! Great job @jpuri !! Thank you @bschorchit for your suggestions :100: I merge the PR.

seaona:
yes @jpuri ! I've just added the comment on that PR too

PR #5902, opened 2023-03-06
improv: add workflow triggers for tags on release and QA

NicolasMassart:
## Development & PR Process

## Description

Add workflow triggers for tags on release and QA and update build notifications to also run on success

We have to trigger manually right now and notifications are not optimal.
However, if it triggers automatically (people may not be aware of the Bitrise trigger when creating a tag), it's important to have notification for both success and failure to raise awareness on the workflow run.

  - Add triggers on tags
  - Modify notifications to send Slack messages on failure on individual platflorm builds and on success of global build workflow

Slack notification utils are now success or failure instead of build or e2e. The messages have been changed to adapt to both.
It simplifies the bitrise file.
Now we have two workflows to allow choosing when to notify: on success, on failure or both, no mater what workflow it is about.

A workflow for instance `create_qa_builds` will notify on failure for each `build_android_qa` or `build_ios_qa`
but will only notify success if both success and `create_qa_builds` eventually succeeds.

The goal is to only provide relevant notifications and prevent notification flood that leads to ignored notifications.

### Example notifications process

Running `create_qa_builds` runs `build_android_qa` and `build_ios_qa` in parallele.
- `build_android_qa` ✅ and `build_ios_qa` 🛑 -> only `build_ios_qa` **failure** notification sent
- `build_android_qa` 🛑 and `build_ios_qa` ✅ -> only `build_android_qa` **failure** notification sent
- `build_android_qa` ✅ and `build_ios_qa` ✅ -> only `create_qa_builds` **success** notification sent

## Issue

fixes MetaMask/mobile-planning#443

## Checklist


bitrise.yml
```diff
@@ -610,3 +614,7 @@ meta:
 trigger_map:
   - push_branch: release/*
     workflow: start_e2e_tests
+  - tag: "v*.*.*-RC-*"
```
Gudahtt, line 617: Is there documentation somewhere on when these tags are expected to be created? I don't see many `-RC-` tags in the repo today.

Gudahtt: Also: thoughts on splitting this into two PRs? The tag triggers and the Slack notification refactor seem pretty distinct from each other

sethkfman: The tag documentation will be added into the release procedures. Right now they are only defined with in the ticket that @NicolasMassart is working on. We want to try this workflow for a couple releases and see how it goes. 

The idea is that the tagging will create a more explicit build process and contributors will not have to go into Bitrise to do this and we can leverage GH actions after it is manually validated.

Gudahtt: So the intent is that tags would get created manually? What are the expected tag formats?

Having some basic description of what the intended workflow would be helpful, otherwise it's hard to say whether this config is correct.

Gudahtt: Thanks, makes sense, I didn't fully grasp the connection at first

NicolasMassart: For the tags naming patterns, the only enforceable restriction is to use [tag rules](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) to prevent specific tags to be created by anyone but admins and maintainers. Other than that, I think it has to be dealt with by having clear contribution guidelines

sethkfman: @Gudahtt Here is the ticket associated with the PR: https://github.com/MetaMask/mobile-planning/issues/443

sethkfman (approved):
LGTM

PR #5924, opened 2023-03-08
Update the README setup steps

Gudahtt:
**Description**

A few project setup steps in the README were obsolete or missing important details. The following changes have been made:
- The `pod install` step was removed from the "Building locally" instructions; this is now a part of `yarn setup` and does not need to be run separately.
- Update the CocoaPods install command to specify the version, ensuring that the lockfile does not change upon install.
- Remove the note about CocoaPods not being required.
- Update the Node.js and Yarn install instructions to be separate, and link the Yarn install instructions.


Cal-L (approved):
LGTM

PR #5926, opened 2023-03-08
[FEATURE] Create detectSnapLocation method to install a Snap From NPM iOS

owencraston:
**Development & PR Process**

**Description**

- Allow snaps to be installed from NPM
- enable snaps on mobile since snaps are stored on NPM

**Screenshots/Recordings**

https://user-images.githubusercontent.com/22918444/226066773-549e5eec-410f-422d-b41e-36921aaac5c3.mov

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/504

**Checklist**


ios/RNTar.swift
```diff
@@ -0,0 +1,61 @@
...
+}
+
+@available(iOS 13.0, *)
+@objc(RNTar)
+class RNTar: NSObject {
+  let SNAPS_NPM_LOG_TAG = "snaps/ NPM";
+
+  func uncompressTgzFile(atPath path: String, toDirectory directory: String) throws -> String {
+    print(SNAPS_NPM_LOG_TAG, "uncompressTgzFile", path, directory)
+    let atURL = URL(fileURLWithPath: path)
```
elefantel, line 24: Could name `atURL` and `toURL` as `sourceUrl` and `destinationUrl`. I think it reads better

ios/RNTar.swift
```diff
@@ -0,0 +1,61 @@
...
+    let toURL = URL(fileURLWithPath: directory, isDirectory: true)
+    // Read the compressed data from the file
+    let data = try! Data(contentsOf: atURL)
```
elefantel, line 27: Force unwrap is strongly discouraged as it's unsafe. Perhaps we can say:
```
guard 
    let data = try? Data(contentsOf: atURL),
    let decompressedData = data.isGzipped ? try? data.gunzipped() : data
else {
   throw UnTarError.unableToDecompressFile
}
```

elefantel, line 27: Force unwrap is strongly discouraged as it's unsafe. Perhaps we can say:
```
guard 
    let data = try? Data(contentsOf: atURL),
    let decompressedData = data.isGzipped ? try? data.gunzipped() : data
else {
   throw UnTarError.unableToDecompressFile
}
```

ios/RNTar.swift
elefantel: Before we do anything we also need to check that the file at the source url exists, otherwise the write will fail.

```
guard fileManager.fileExists(atPath: path) else {
    throw error here
}
```

ios/Light-Swift-Untar.swift
```diff
@@ -0,0 +1,177 @@
...
+  case corruptFile(type: UnicodeScalar)
+  public var errorDescription: String? {
+    switch self {
+    case let .notFound(file: file): return "Source file \(file) not found"
+    case let .corruptFile(type: type): return "Invalid block type \(type) found"
+    }
+  }
+}
+
+public extension FileManager {
```
elefantel, line 26: This is OK since you took this file from somewhere as is otherwise extending modules is generally not recommended because it makes testing difficult. It's best to create a class that will handle the persistence at a high level, e.g `FileStore`. Then we can create an Untar class that will use `FileStore` for persistence. This class will then handle the untar specific logic.

owencraston:
For the future, we should investigate the performance gains that we would get from [using this package](https://github.com/mrousavy/react-native-blob-jsi-helper).

ios/RNTar.swift
owencraston: I love the chaos. I implemented this suggestion in my latest commit!

ios/RNTar.swift
```diff
@@ -0,0 +1,61 @@
...
+    if data.isGzipped {
+        decompressedData = try! data.gunzipped()
+    } else {
+        decompressedData = data
+    }
+    let fileManager = FileManager.default
+    
+    if !fileManager.fileExists(atPath: toURL.path) {
+      print(SNAPS_NPM_LOG_TAG, "uncompressTgzFile creating directory at", toURL.path)
+      try! fileManager.createDirectory(atPath: toURL.path, withIntermediateDirectories: true)
```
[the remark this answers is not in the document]
owencraston: How would this work? `createFilesAndDirectories` is the extension function, I am not sure if it will create the target directory for us.

ios/Light-Swift-Untar.swift
owencraston: I agree. We should review this before we release to production but for the sake of speed I will keep it as is for now.

PR #5940, opened 2023-03-09
Revert "fix: bundle video in app to prevent crash when not available"

NicolasMassart:
Reverts MetaMask/metamask-mobile#5886
Because current apps are relying on the video hosted on main branch and this PR renamed the video to add source in name making it fail to load and displays a black square instead of the video.
We have to apply the PR again with taking this retro compatibilty into account.


Cal-L (approved):
LGTM

sethkfman (approved):
LGTM

PR #5941, opened 2023-03-09
fix: bundle video in app to prevent crash when not available (2nd attempt)

### LINKED ISSUE #5497: Bundle the onboarding SRP explaining video in the app build

NicolasMassart:
## Development & PR Process

## Important

This PR is the second attempt to push this change as the previous one #5886 failed because of an unexpected side effect and required revert, see #5940
This new one takes the side effect into account, see description bellow.

## Description

Make the SRP video bundled in the app instead of retrieved from Github repos.
Have the subtitles embedded in the video, removing the need to load them from an external source. 

App crashed when video player tried to load the remote video and remote was not available (or network down or anything)

  - Bundling the video in the app to prevent crash when remote not available while still **keeping it in the source repo with same name as the previous one for compatibility with currently in production apps**
  - Added readme in video folder to explain and warn others.
  - Keep current **subtitles files in the same location for compatibility with currently in production apps**
  - Embed subtitles in the video directly and select track using the language 2 letters code,
  - Reduced video size from 5.7Mb to 2Mb with ffmpeg 2 passes encoding (including subtitles tracks).
  - Created an ffmpeg encoding and subtitles embedding script in scripts folder if we need to regenerate the video (in case of subtitle change, video update, ...) 

The video is displayed by default on its original format on portrait mobile, meaning this video is most of the time around 5x3cm, very hard to see the quality diff when not really looking for it. Fullscreen makes it a bit more visible though.
The source video is still in the repos (but not bundled in the app) and the ffmpeg script allows to compress with higher bitrates if requested (change variables in the script).

### Source video

https://github.com/MetaMask/metamask-mobile/raw/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase-source.mp4

### Compressed video with embedded subs

https://github.com/MetaMask/metamask-mobile/raw/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase.mp4

## App size impact
The app size is only 2 additional Mb for the video. Makes it around 27.8Mb for iOS and 274.7Mb for Android.

## Issue

Relates to MetaMask/mobile-planning#471
fixes #5497

## Testing

https://user-images.githubusercontent.com/4677568/225321865-3883943b-e8f1-4554-a89f-d9c12d34e0bd.mov

- Video testing has been done manually as no e2e exists to test the video and previous attempts to handle the player in e2e failed.
- retro compatibility tests: due to the revert, we add a step to check on currently in prod apps that the video is not broken.

Here are the tests done:
- No network at all (plane mode) as required in issue scenario `GIVEN: User has NO access to internet (airplane mode)`: the app has a system preventing it to be used and requires the user to turn network on, so not relevant.
- Slow network: because video and subtitles are now bundled in the app, none is impacted.
- Tested video from the multiple places it plays in the app: onboarding, settings.
- Tested with the previous version of app (before change in code to have bundled video) that the video and subs still work when path points to this PR branch 😓 making sure the name and path of the video and subtitles are ok (using console logs) and that it will still work for legacy SRP video code.
  ```text
   LOG  video_source_uri https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase.mp4?raw=true
   LOG  getSubtitleUri https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/subtitles/secretPhrase/subtitles-en.vtt?raw=true
  ...
   LOG  video_source_uri https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase.mp4?raw=true
   LOG  getSubtitleUri https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/subtitles/secretPhrase/subtitles-de.vtt?raw=true
  ...
   LOG  video_source_uri https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/recovery-phrase.mp4?raw=true
   LOG  getSubtitleUri https://github.com/MetaMask/metamask-mobile/blob/471_make_SRP_video_embeded_in_app/app/videos/subtitles/secretPhrase/subtitles-hi.vtt?raw=true
  ```

### Test artifacts
- QA builds available on Bitrise public install pages:
  - [Android QA build 6389 (apk)](https://app.bitrise.io/build/d3b22c9c-8ffb-4922-92a5-dd8c63385329?tab=artifacts)
  - [iOS QA build 6390 (ipa)](https://app.bitrise.io/build/39f3e427-1b3c-4ab3-b995-4f11f78a15d1?tab=artifacts)

## Checklist


app/constants/urls.ts
```diff
@@ -9,6 +9,10 @@ export const KEEP_SRP_SAFE_URL =
   'https://metamask.zendesk.com/hc/en-us/articles/4407169552667-Scammers-and-Phishers-Rugpulls-and-airdrop-scams';
 export const LEARN_MORE_URL =
   'https://metamask.zendesk.com/hc/en-us/articles/360015489591-Basic-Safety-and-Security-Tips-for-MetaMask';
+// Remote path for subtitles that can't be bundled with the app
+// because only remote subtitles are supported by react-native-video
+export const SRP_SUBTITLES_URL_TEMPLATE =
+  'https://metamask.github.io/metamask-mobile/Subtitles/v1/secretPhrase/subtitles-${language}.vtt';
```
NicolasMassart, line 15: This changed compared to #5886 and now targets the Github Pages hosted subs.

app/util/video/index.test.ts
NicolasMassart, line 1: file removed as it was only useful for the SRP video and subs remote path

app/videos/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+This folder contains video and subtitles that are currently served as raw content from GitHub repos directly to the mobile app.
+
+:warning: Video is now bundled in the app but until 100% of apps in production are using the new bundled version,
+there's still chances that the raw video served from GitHub is used.
+Do not rename or remove the `app/videos/recovery-phrase.mp4` file until you are totally sure of what you do.
+
+:warning: Subtitles are moved to `gh-pages` branch but until 100% of apps in production are using the new Github Pages hosted
+subtitles path, they are till served raw from GitHub repos main.
+Do not rename or remove the `app/videos/Subtitles` folder until you are totally sure of what you do.
```
NicolasMassart, line 11:
```suggestion
:warning: Subtitles are moved in the bundled video branch but until 100% of apps in production are using the new bundled video, they are till served raw from GitHub repos main.
Do not rename or remove the `app/videos/Subtitles` folder until you are totally sure of what you do.
```

app/videos/README.md
```diff
@@ -0,0 +1,14 @@
+# Warning about video and subtitles
+
+This folder contains video and subtitles that are currently served as raw content from GitHub repos directly to the mobile app.
+
+:warning: Video is now bundled in the app but until 100% of apps in production are using the new bundled version,
+there's still chances that the raw video served from GitHub is used.
+Do not rename or remove the `app/videos/recovery-phrase.mp4` file until you are totally sure of what you do.
+
+:warning: Subtitles are moved in the bundled video branch but until 100% of apps in production are using the new bundled video, they are till served raw from GitHub repos main.
```
NicolasMassart, line 9:
```suggestion
:warning: Subtitles are moved in the bundled video branch but until 100% of apps in production are using the new bundled video, they are still served raw from GitHub repos main.
```

scripts/compress_and_add_sub_in_video.sh
```diff
@@ -0,0 +1,109 @@
...
+ffmpeg_subtitle_map_inputs=""
+ffmpeg_subtitle_metadata=""
+ffmpeg_subtitle_map_inputs_start=1
+ffmpeg_subtitle_default_language="en"
+ffmpeg_subtitle_default_track=0
+
+# map all language to language code
+# as languages 2 letter codes are not the same as language codes used by ffmpeg
+# https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes
+language_code_map=(
```
NicolasMassart, line 31: Yes, hardcoded list. I wanted to prevent over engineering here (picking an iso code lib and stuff) and given list will probably not change, it's fine.
If it does, we just have to add the new code.

scripts/compress_and_add_sub_in_video.sh
```diff
@@ -0,0 +1,109 @@
...
+# check if ffmpeg is installed
+if ! command -v ffmpeg &> /dev/null
+then
+    echo "ffmpeg could not be found."
+    echo "Please install ffmpeg before running this script."
+    echo "https://ffmpeg.org/download.html"
+    exit 1
+fi
+
+#chech source video exists
```
Cal-L, line 58: Check

scripts/compress_and_add_sub_in_video.sh
Cal-L: That's reasonable

app/declarations.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*.mp4';
```
sethkfman, line 1: What is this doing?

app/components/Views/MediaPlayer/index.js
```diff
@@ -70,11 +76,11 @@ MediaPlayer.propTypes = {
    */
   onClose: PropTypes.func,
   /**
-   * Array of possible text tracks to display
+   * Array of remote possible text tracks to display
```
sethkfman, line 79: Is "remote" still accurate?

scripts/compress_and_add_sub_in_video.sh
sethkfman: Can we add a comment in the SRPVideo file that if a subtitle language is added it needs to be added to this hardcoded list?

sethkfman (commented):
Minor comments

app/declarations.d.ts
NicolasMassart: Mp4 is not an actual TS/JS module. So import doesn't work and requires that the node environment realises that it can import this like it does for images or jsons. So this globally tells that mp4 can be imported. Having it at the app level also helps IDEs to syntax check this and not complain.

app/components/Views/MediaPlayer/index.js
NicolasMassart: Yes. Because this is at the player level, not just for the SRP video. So if someone whants to use the mediaPlayer elsewhere in the app, we updated the prop types to allow either bundled video or remotely hosted video.
At the same time we kept the text tracks (subtitles) feature to use non embedded tracks alongside the video, if needed. Who knows. It's not used for SRP video, but can be used elsewhere.
So you can either input a video with embedded subs or a video and remote subs.
But you can not use subs from files bundled in the app but separated from the video.
This warning is to remind that only remote subs files work with react-native-video textTrack prop and not bundled files.

app/videos/README.md
```diff
@@ -0,0 +1,14 @@
...
+:warning: Video is now bundled in the app but until 100% of apps in production are using the new bundled version,
+there's still chances that the raw video served from GitHub is used.
+Do not rename or remove the `app/videos/recovery-phrase.mp4` file until you are totally sure of what you do.
+
+:warning: Subtitles are moved in the bundled video branch but until 100% of apps in production are using the new bundled video, they are still served raw from GitHub repos main.
+Do not rename or remove the `app/videos/Subtitles` folder until you are totally sure of what you do.
+
+## Video encoding
+
+`recovery-phrase.mp4` is encoded from source `recovery-phrase-source.mp4` video using the script `scripts/compress_and_add_sub_in_video.sh`.
```
NicolasMassart, line 14:
```suggestion
`recovery-phrase.mp4` is encoded from source `recovery-phrase-source.mp4` video using the script `scripts/compress_and_add_sub_in_video.sh`.

## Subtitles languages

Encoding the subtitles in the video uses a static map of languages codes. If you add a new language subtitle file you also have to add the mapping in the `language_code_map` variable in  `scripts/compress_and_add_sub_in_video.sh`
```

scripts/compress_and_add_sub_in_video.sh
NicolasMassart: done in app/videos/README.md

app/declarations.d.ts
Cal-L: Also just FYI, in case you wanted to give it an explicit return type you can do something like this
```
declare module '*. mp4' {
  import { ImageSourcePropType } from 'react-native';
  const content: ImageSourcePropType;
  export default content;
}
```
something to reference for future declarations

Cal-L (approved):
LGTM

Andepande (commented):
- Tested on bitrise with IOS & Android devices
- Video together with subtitles load fine even when network is throttled
- Verified video loads correctly with appropriate subtitles in different Languages [en, es, hi]
- Verified with @NicolasMassart production issue that occurred on previous merge has been rectified.

Issue #5911, opened 2023-03-07
The connected accounts action sheet resets after importing account via private key


cortisiko:
We have a bug when importing an account via the connect modal. After importing the account, I am taken back to step1 in the connect accounts modal.

See main: http://recordit.co/oSBxIzZdJm 
 See prod v6.1.1 here: http://recordit.co/nvdv53X83X 

To reproduce:
Go to a dapp
Trigger the connect modal
Tap on “connect more accounts”
Import an account via private key
Notice you are no longer in the connect multiple accounts screen

cortisiko:
This is still an issue on main even after 6.1.2 got merged in. See [here](http://recordit.co/wFkVwCsl8E)


See prod [v6.1.2](http://recordit.co/0tSn9QZ8KP)

cortisiko:
fixed in https://github.com/MetaMask/metamask-mobile/pull/5962

PR #5962, opened 2023-03-14
[FIX] Remove select address as a prop on App index.js

Cal-L:
**Description**

This PR fixes the bug where the account sheet re-renders after the mutation of an account. This includes creating, importing, or removing. The solution was to remove `selectedAddress` from the props to prevent re-render triggers.

**Screenshots/Recordings**

Reach out to @Cal-L on slack for video

**Issue**
#5911 

**Testing**
- Scenario: Account sheet should be on multi-select after the creation of an account
   - Given I am on the Browser view
   - When I tap on the right button on the Nav bar
   - And I tap on `Connect more accounts`
   - And I tap on `Create a new account`
   - Then a new account should be created
   - And we should still be on the account multi-select screen
 
- Scenario: Repeat the above scenario but import an account or remove instead

Progresses #???

**Checklist**

### DIFF
--- a/app/components/Nav/App/__snapshots__/index.test.tsx.snap
+++ b/app/components/Nav/App/__snapshots__/index.test.tsx.snap
@@ -3,13 +3,11 @@
 exports[`App should render correctly when logged in 1`] = `
 <App
   dispatch={[Function]}
-  selectedAddress="0x1"
 />
 `;
 
 exports[`App should render correctly when logged in 2`] = `
 <App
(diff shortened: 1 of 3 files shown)


owencraston (approved):
Much better. I did not verify the fix but the code looks good.

cortisiko (approved):
I can confirm that this PR fixes https://github.com/MetaMask/metamask-mobile/issues/5911

PR #5968, opened 2023-03-15
Move `eth_sendTransaction` method handler to a separate module

Gudahtt:
**Development & PR Process**

**Description**

The method handler for `eth_sendTransaction` has been moved to a separate module to simplify unit testing. Unit tests have been added as well.

This module was written in TypeScript, which asked for some additional validation of the method parameters. We now throw a more explicit message when the params are missing, and when the transaction parameters are not an object. Previously the method would still throw an error in these scenarios, just with a less helpful message. The change in error message should be the only functional change here.

**Issue**

This refactor is intended to make testing PR #5619 easier.

**Checklist**


app/core/RPCMethods/eth_sendTransaction.test.ts
```diff
@@ -0,0 +1,110 @@
+// eslint-disable-next-line import/no-nodejs-modules
```
Gudahtt, line 1: I don't think we need this rule enabled in unit tests. I'll look into disabling this for tests in a follow-up PR.

package.json
```diff
@@ -152,11 +152,12 @@
     "@metamask/message-manager": "^1.0.1",
     "@metamask/network-controller": "^2.0.0",
     "@metamask/permission-controller": "^1.0.2",
-    "@metamask/preferences-controller": "^2.1.0",
     "@metamask/phishing-controller": "^2.0.0",
+    "@metamask/preferences-controller": "^2.1.0",
     "@metamask/sdk-communication-layer": "^0.1.0",
     "@metamask/swaps-controller": "^6.8.0",
     "@metamask/transaction-controller": "^2.0.0",
+    "@metamask/utils": "^5.0.0",
```
Gudahtt, line 160: This was causing problems on the extension because an indirect dependency (that we don't actually invoke) uses WASM. I'll need to verify that this doesn't impact the mobile build.

Gudahtt: Seems to have no impact! I was able to build successfully.

gantunesr (approved):
LGTM

package.json
```diff
@@ -324,6 +324,7 @@
     "@metamask/eslint-config": "^7.0.0",
     "@metamask/eslint-config-typescript": "^7.0.0",
     "@metamask/mobile-provider": "^2.1.0",
+    "@metamask/utils": "^5.0.0",
```
Gudahtt, line 327: Oops, my bad, this is a production dependency. I added it here first because initially it was only used in a test

Cal-L (approved):
Really clean implementation. LGTM

PR #5970, opened 2023-03-15
[FIX] Fix displaying balance on watch asset screen

Cal-L:
**Description**

This PR fixes showing the correct balance on the watch asset screen
Issue - https://app.zenhub.com/workspaces/mobile-release-regression-6249e5242464b70013315a98/issues/gh/metamask/mobile-planning/712

**Screenshots/Recordings**

https://user-images.githubusercontent.com/10508597/225441560-3ef83170-18dd-4654-8d2b-da0d533729aa.mov

**Testing**
- Scenario: Watch asset screen should show correct token balance for the interacting account
  - Given I go the the Browser view and go to Uniswap
  - And I connect an account to Uniswap
  - When I swap ETH for UNI
  - And I tap `Add UNI` to trigger watch asset
  - Then I should see the correct balance for UNI

**Checklist**

### DIFF
--- a/app/components/UI/WatchAssetRequest/index.js
+++ b/app/components/UI/WatchAssetRequest/index.js
@@ -1,7 +1,6 @@
 import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { StyleSheet, View, Text, InteractionManager } from 'react-native';
-import { connect } from 'react-redux';
 import URL from 'url-parse';
 import ActionView from '../ActionView';
 import { MetaMetricsEvents } from '../../../core/Analytics';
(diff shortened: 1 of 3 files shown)


tommasini (approved):
LGTM!

app/components/UI/WatchAssetRequest/index.js
```diff
@@ -90,15 +90,20 @@ const createStyles = (colors) =>
...
   suggestedAssetMeta,
   currentPageInformation,
-  selectedAddress,
   onCancel,
   onConfirm,
 }) => {
   const { asset, interactingAddress } = suggestedAssetMeta;
-  let [balance] = useTokenBalance(asset.address, selectedAddress);
-  balance = renderFromTokenMinimalUnit(balance, asset.decimals);
+  // TODO - Once TokensController is updated, interactingAddress should always be defined
```
tommasini, line 97: Don't understand quite well the TODO here

There is a patch made with the interactingAddress already to the TokensController right? 
So it should be always defined right?

Cal-L: Yep. But this is going to be released before the controller update so we have to handle the undefined case.

tommasini: Thanks Cal!

cortisiko:
When I try to add assets via an asset watcher such as coin gecko I see: “failed” as the token balance. 
https://www.screencast.com/t/vaXvzaULbaYt


This is not the case in prod: https://www.screencast.com/t/egdY2rOqveGp

cortisiko (approved):
✅ ✅ ✅ ✅ ✅

PR #5997, opened 2023-03-20
[MC 0.5][FEATURE] Account selector on swaps screen

tommasini:
**Description**
This PR aims to add account selector to the swaps screen, the reason for this it's when the user uses the new wallet action button (button on the center of the bottom tab menu), they can be on a dapp with a different selected address, that way if they try to swap, they will understand which account are they using

**Test Steps**
* Go to swaps
* Change the account on account selector
* Notice every time you change the account, the swap state reset to the initial state

**Screenshots/Recordings**
Ability to switch accounts on Swaps screen: https://recordit.co/SnpmupJKI3
Account shown on Swaps screen is not synced with the active Dapp accounts: https://recordit.co/BzWPZtujiA
Account change on Swaps screen reset the swap screen: https://recordit.co/RrwuwcJUcT

Removing imported account via the swaps screen: https://recordit.co/Va4ZTyDTmQ
creating an account with a long name then going to the swaps view: https://recordit.co/z0WBeOZTmg
Import account via the account selector from dapp connect modal via the browser then remove the imported account via the swaps account selector: https://recordit.co/OK7YaukMnU

**Issue**

Progresses #???

**Checklist**

### DIFF
--- a/app/components/UI/Swaps/index.js
+++ b/app/components/UI/Swaps/index.js
@@ -76,6 +76,8 @@ import {
   selectChainId,
   selectProviderConfig,
 } from '../../../selectors/networkController';
+import AccountSelector from '../FiatOnRampAggregator/components/AccountSelector';
+import { set } from 'lodash';
 
 const createStyles = (colors) =>
   StyleSheet.create({
@@ -88,6 +90,12 @@ const createStyles = (colors) =>
     content: {
       flexGrow: 1,
       justifyContent: 'center',
+    },
+    accountSelector: {
+      width: '100%',
+      justifyContent: 'center',
+      alignItems: 'center',
+      marginBottom: 16,
     },
     keypad: {
       flexGrow: 1,
@@ -380,6 +388,17 @@ function SwapsAmountView({
     })();
   }, [isTokenInBalances, selectedAddress, sourceToken]);
 
+  useEffect(() => {
+    setAmount('0');
+    setSourceToken(
+      swapsTokens?.find((token) =>
+        toLowerCaseEquals(token.address, initialSource),
+      ),
+    );
+    setDestinationToken(null);
+    setSlippage(AppConstants.SWAPS.DEFAULT_SLIPPAGE);
+  }, [selectedAddress, swapsTokens, initialSource]);
+
   const hasInvalidDecimals = useMemo(() => {
     if (sourceToken) {
       return amount?.split('.')[1]?.length > sourceToken.decimals;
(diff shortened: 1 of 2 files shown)


app/components/UI/Swaps/index.js
```diff
@@ -89,6 +90,12 @@ const createStyles = (colors) =>
       flexGrow: 1,
       justifyContent: 'center',
     },
+    accountSelector: {
+      width: '100%',
+      justifyContent: 'center',
```
Cal-L, line 95: Probably do not need justifyContent

app/components/UI/Swaps/index.js
```diff
@@ -380,6 +387,17 @@ function SwapsAmountView({
     })();
   }, [isTokenInBalances, selectedAddress, sourceToken]);
 
+  useEffect(() => {
```
Cal-L, line 390: Can you comment here that this is meant to reset the states from an account change?

app/components/UI/Swaps/index.js
```diff
@@ -380,6 +387,17 @@ function SwapsAmountView({
...
+    setAmount('0');
+    setSourceToken(
+      swapsTokens?.find((token) =>
+        toLowerCaseEquals(token.address, initialSource),
+      ),
+    );
+    setDestinationToken(null);
+    setSlippage(AppConstants.SWAPS.DEFAULT_SLIPPAGE);
+  }, [selectedAddress, swapsTokens, initialSource]);
```
Cal-L, line 399: To ensure that the other dependencies don't unexpectedly reset your states, you may need to track something like `previousSelectedAddress` and do a comparison `selectedAddress !== previousSelectedAddress` before resetting. This guarantees that the only thing that will trigger a reset is a change in account.

app/components/UI/Swaps/index.js
tommasini: True! Good catch!

app/components/UI/Swaps/index.js
tommasini: Done!

app/components/UI/Swaps/index.js
tommasini: It seems for doing that I would need to change the `AccountSelector` to receive from `route.params` a function with callback to save the previousSelectedAddress on the `Swaps` component state

I think a better implementation would be on the PreferencesController we have that `previousSelectedAddress` on the state, because this is something that probably we would need across multiple places as we saw on the `TokenRatesController`, what do you think?

Meanwhile on this PR I can update the useEffect for this
```suggestion
// eslint-disable-next-line react-hooks/exhaustive-deps
  }, [selectedAddress);
```

Cal-L (approved):
LGTM

Cal-L:
@kevinghim No events triggered yet. Let's add one and reuse if possible.

Issue #5946, opened 2023-03-10
Documentation - metamask mobile confirmation architecture


jpuri:
Creating a detailed documentation for metamask mobile confirmations frontend (mobile-app / react-native side) architecture.

bschorchit:
All 3 PRs have been merged, should this issue be closed, @jpuri ?

PR #6003, opened 2023-03-21
Adding document to refactor send flow

jpuri:
Ref: https://github.com/MetaMask/metamask-mobile/issues/5946

Architectural document for send page. It also include code cleanup we should do in these components.

The document can be read here: https://github.com/MetaMask/metamask-mobile/tree/send_architectural_doc/docs/confirmation-refactoring/send#readme


docs/confirmation-refactoring/send/README.md
```diff
@@ -0,0 +1,125 @@
...
+        - SendTo.test.tsx
+```
+There will be more components added to the list above as we do refactoring.
+
+### SendTo Component
+[/Views/SendFlow/SendTo](https://github.com/MetaMask/metamask-mobile/tree/main/app/components/Views/SendFlow/SendTo) is responsible to render following send page:
+
+<img src="https://github.com/MetaMask/metamask-mobile/blob/send_architectural_doc/docs/confirmation-refactoring/send/send_to.png?raw=true" width="150"/>
+
+The component is taking care to do following:
```
blackdevelopa, line 38: These seems to be UI components and maybe should be in the UI directory instead of Views?

docs/confirmation-refactoring/send/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Amount Page
+Amount page is rendered by [/Views/SendFlow/Amount](https://github.com/MetaMask/metamask-mobile/tree/main/app/components/Views/SendFlow/Amount) component. The page is again quite huge and taking care of 2 different inputs:
+
+- token amount input
+- asset input
+
+These can be broken down into different components to reduce complexity of this component.
+
+Additionally the components has a lot of different transaction / gas related methods defined in it, these should be converted(fully or partially) into re-uable utility methods or hooks. Here is a list of the methods I could find but there will be more to it:
```
blackdevelopa, line 70: It'll seem that some of these methods are only used once in the code, moving them will help reduce code length if this is what we are going for? Wdyt?

jpuri: Yep that is reason I mentioned `fully or partially` some are fully re-usable and some can be broken down to extract out re-usable logic.

docs/confirmation-refactoring/send/README.md
jpuri: Yes 👍

jpuri: Let me add this point to doc also.

blackdevelopa (approved):

docs/confirmation-refactoring/send/README.md
```diff
@@ -0,0 +1,143 @@
...
+#### Amount Page
+
+Amount page is rendered by [/Views/SendFlow/Amount](https://github.com/MetaMask/metamask-mobile/tree/main/app/components/Views/SendFlow/Amount) component. The page is again quite huge and taking care of 2 different categories inputs:
+
+- token amount input
+- asset input
+
+These can be broken down into 2 different components to reduce complexity of this component.
+
+Additionally the components has a lot of different transaction / gas related methods defined in it, these should be converted(fully or partially) into re-uable utility methods or hooks. A function not requiring state changes can be made a utility method, function more closely tied to state change should be hook. Here is a list of the methods I could find but there will be more to it:
```
NiranjanaBinoy, line 82:
```suggestion
Additionally the components has a lot of different transaction / gas related methods defined in it, these should be converted(fully or partially) into reusable utility methods or hooks. A function not requiring state changes can be made a utility method, function more closely tied to state change should be hook. Here is a list of the methods I could find but there will be more to it:
```

docs/confirmation-refactoring/send/README.md
```diff
@@ -0,0 +1,143 @@
...
+<img src="https://github.com/MetaMask/metamask-mobile/blob/send_architectural_doc/docs/confirmation-refactoring/send/send_confirm_gas.png?raw=true" width="150"/>
+
+3. Legacy Gas modal
+4. Custom Nonce modal
+5. Hex Data modal
+6. Transaction Info:
+
+<img src="https://github.com/MetaMask/metamask-mobile/blob/send_architectural_doc/docs/confirmation-refactoring/send/send_confirm_info.png?raw=true" width="150"/>
+
+The component also has a lot of different transaction / gas related methods defined in it, these should be converted (fully or partially) into re-uable utility methods or hooks:
```
NiranjanaBinoy, line 117:
```suggestion
The component also has a lot of different transaction / gas related methods defined in it; these should be converted (fully or partially) into reusable utility methods or hooks:
```

docs/confirmation-refactoring/send/README.md
```diff
@@ -0,0 +1,143 @@
...
+#### Other refactoring areas:
+
+1. A lot of components have logic to update NarBar defined as:
+   ```
+   componentDidUpdate = () => {
+       this.updateNavBar();
+   };
+   ```
+   As we make components functional this can be converted to a tested re-usable hook used across all the components.
+2. Code to start / stop gas polling is used at few places, it can be converted to re-usable hoo.
```
NiranjanaBinoy, line 137:
```suggestion
2. Code to start / stop gas polling is used at a few places; it can be converted to a re-usable hook.
```

NiranjanaBinoy (approved):
Looks good to me; I added some NIT to fix typos.

PR #6037, opened 2023-03-24
Feature/update sdk persistence

abretonc7s:
**Description**

- Complement previously merged session persistence.
-  Fix issue on android that could cause double re-connection to a given session (from a RN dapp).
- Improve layout of the SDK session management UI to handle long name/url.
- begin refactor to split logic of SDKConnect file into separate components.


andreahaku (approved):
LGTM

PR #6021, opened 2023-03-22
[REVERT] Move `eth_sendTransaction` method handler to a separate module (#5968)

Cal-L:
**Description**

This reverts commit 49a63050ae5e1d6889706c025eb222543b2ff9a1. It was reported from @BelfordZ & QA that the changes resulted in a `BigInt` issue being thrown on Android.

**Issue**
Reverts #5968 

**Testing**
Try sending legacy transaction. It should be working and should not throw any errors.
Reference convos from slack: https://consensys.slack.com/archives/CBW7S9FSN/p1679427170122229, https://consensys.slack.com/archives/C02U025CVU4/p1679445188858539

**Checklist**

* [ ] There is a related GitHub issue
* [ ] Tests are included if applicable
* [ ] Any added code is fully documented

PR #6060, opened 2023-03-28
Refactor send transaction v2

Gudahtt:
**Development & PR Process**

**Description**

This is a repeat of #5968, which was reverted in #6021 to fix a bug on Android. Investigation into that bug is ongoing, but it related to the `@metamask/utils` import. That dependency has been replaced in this PR in [the most recent commit](https://github.com/MetaMask/metamask-mobile/commit/3d6b6a0e4d60a3b8450c2b455d0be46cc540b86f); all previous commits are identical to the previous PR (except that they have been rebased onto `main`).

Original description:

The method handler for eth_sendTransaction has been moved to a separate module to simplify unit testing. Unit tests have been added as well.

This module was written in TypeScript, which asked for some additional validation of the method parameters. We now throw a more explicit message when the params are missing, and when the transaction parameters are not an object. Previously the method would still throw an error in these scenarios, just with a less helpful message. The change in error message should be the only functional change here.

**Issue**

This refactor is intended to make testing PR https://github.com/MetaMask/metamask-mobile/pull/5619 easier.

**Checklist**


Cal-L (approved):
LGTM

PR #6071, opened 2023-03-29
Add documentation for creating new confirmations

matthewwalsh0:
**Description**

Add documentation for creating new confirmations.

See [here](https://github.com/MetaMask/metamask-mobile/blob/chore/728-confirmations-guidelines/docs/confirmations.md) for markdown.

**Checklist**


matthewwalsh0:

bergarces (approved):
LGTM.

Particularly glad to see that the backend part shares a lot of similarities already with extension thanks to `ApprovalController`.

docs/confirmations.md
```diff
@@ -0,0 +1,189 @@
...
+    setShowPendingApproval(false);
+    setExampleRequest(undefined);
+};
+
+const onExampleReject = () => {
+    rejectPendingApproval(exampleRequest.id, exampleRequest.requestData);
+    setShowPendingApproval(false);
+    setExampleRequest(undefined);
+};
+```
```
jpuri, line 140: Above code can be in places other than `RootRPCMethodsUI`. `RootRPCMethodsUI` has increased a lot in size as it is doing many different things.

We plan to extract away confirmation code from here. Here is a recent refactoring PR: https://github.com/MetaMask/metamask-mobile/pull/6097

docs/confirmations.md
```diff
@@ -0,0 +1,189 @@
...
+
+### 7. Render Modal
+
+Update the [RootRPCMethodsUI](../app/components/Nav/Main/RootRPCMethodsUI.js) component to render a modal for the new approval type.
+
+Use a new method to encapsulate the code for the approval type.
+
+Include the new approval component as a child of the `Modal` component.
+
+Reference the new handlers in the `Modal` properties `onSwipeComplete` and `onBackdropPress`, and in the appropriate properties of the new approval component.
```
jpuri, line 150: Modal rendering is not needed for requests like `Send` not coming from DAPP

docs/confirmations.md
matthewwalsh0: I agree in regard to the approval components themselves and any related domain logic, and there is definitely lots to move out of the component.

But I believe the ideal future is that all modals are triggered from approval requests meaning it would be cleanest to have a single listener for `pendingApprovals` with cases for each type which would just set the local pending approval state, which would in turn render the right modal.

This means the above handlers wouldn't have to include any domain logic, but we'd at least need a callback so we can clear the pending approval state.

docs/confirmations.md
matthewwalsh0: Are you referring to the `renderDappTransactionModal`, `renderApproveModal`, and `renderQRSigningModal`?

I can see they're not rendered inside a `Modal`, but they do all use `Modal` internally. I thought best practice would be to create these approval components without coupling them to a modal to support any future usage also.

So there could even be an argument to remove the `Modal` from inside the above components, and instead provide the `Modal` here, but there is lots of duplicated properties in the current usages so perhaps a `ConfirmationModal` with suitable defaults to reduce the amount of code?

jpuri (approved):

PR #6079, opened 2023-03-30
Refactor sanitization middleware

### LINKED ISSUE #5846: Refactor: migrate sanitization middleware from `web3-provider-engine` to mobile
### **Description**

The package `web3-provider-engine` is used internally by the network controller (`@metamask/network-controller`). We are preparing to replace it. In preparation for this change, we are migrating some middleware functionality from `web3-provider-engine` into mobile to prevent this replacement from having any functional impact on the dapp API.

This issue is for the santization middleware, which formats the parameters of any request that looks similar to `eth_sendTransaction`.

### **Technical Details**

- Migrate the sanitization middleware from `web3-provider-engine` to mobile's `RPCMethodMiddleware` module.
- Ensure this will result in no functional changes
  - Ensure that the order of execution is identical, or that any differences will not affect the results
  - Ensure that by introducing a second copy of this santization step, we will not be affecting the results (if we find this to be non-idempotent, we can remove the `web3-provider-engine` santitization step using a patch)

### **Acceptance Criteria**

-  We are confident there are no functional changes
- We have ensured that the removal of `web3-provider-engine` will not change how RPC methods are sanitized

### **References**

This helps to unblock https://github.com/MetaMask/core/issues/970

BelfordZ:
**Development & PR Process**

**Description**

Add sanitiziation RPC middleware to process requests just before they get sent into our internal network provider. This is meant to be a refactor with no functional changes; this middleware is identical to [the middleware already used today within `web3-provider-engine`](https://github.com/MetaMask/web3-provider-engine/blob/9ac79a46ddbe37612027b08e22f395fd7dfeef04/subproviders/sanitizer.js).

This is being added so that we can remove `web3-provider-engine` with fewer functional changes.

**Test Instructions**

This should have no functional impact. This middleware is expected to process any RPC request not intercepted earlier in the pipeline however (essentially any request that gets sent to the RPC endpoint). It should ignore most requests (as the unit tests demonstrate), but it will process any that have an object as the first parameter. Known examples are `eth_call`, `eth_estimateGas`, and `eth_getLogs`.

To manually test this, you could try invoking these methods with parameters that the middleware should be expected to filter out. Or invoke them with parameters that are allowed, to confirm they remain unaffected. You could also try testing requests that we anticipate are unaffected. In all cases, we should see the same behavior before and after.

The test dapp has a new "Request" page that might be useful in testing this. See the README Usage section for details: https://github.com/MetaMask/test-dapp

**Issue**

Closes #5846

**Checklist**

### DIFF
--- a/app/core/SanitizationMiddleware.test.ts
+++ b/app/core/SanitizationMiddleware.test.ts
@@ -1,6 +1,9 @@
 'use strict';
 
-import { createSanitizationMiddleware } from './SanitizationMiddleware';
+import {
+  createSanitizationMiddleware,
+  permittedKeys,
+} from './SanitizationMiddleware';
 
 const noop = () => {
   // does nothing
@@ -89,52 +92,8 @@ describe('createSanitizationMiddleware', () => {
     expect(testTxLike.params[0]).toStrictEqual({});
   });
 
-  it('does nothing when already prefixed and lowercased', () => {
-    const sanitizationMiddleware = createSanitizationMiddleware();
-    const testTxLike = {
-      jsonrpc: '2.0' as const,
-      id: '1',
-      method: 'eth_call',
-      params: [
-        {
-          to: '0x123',
-        },
-      ],
-    };
-
-    sanitizationMiddleware(
-      testTxLike,
-      { jsonrpc: '2.0', id: 'any' },
-      noop,
-      noop,
-    );
-    expect(testTxLike.params[0]).toStrictEqual({ to: '0x123' });
-  });
-
-  it('does nothing to non-string permitted properties', () => {
-    const sanitizationMiddleware = createSanitizationMiddleware();
-    const testTxLike = {
-      jsonrpc: '2.0' as const,
-      id: '1',
-      method: 'eth_call',
-      params: [
-        {
-          to: 100,
-        },
-      ],
-    };
-
-    sanitizationMiddleware(
-      testTxLike,
-      { jsonrpc: '2.0', id: 'any' },
-      noop,
-      noop,
-    );
-    expect(testTxLike.params[0]).toStrictEqual({ to: 100 });
-  });
-
-  for (const blockRef of ['latest', 'pending', 'earliest']) {
-    it(`does nothing to permitted properties set to ${blockRef}`, () => {
+  for (const permittedProperty of permittedKeys) {
+    it(`leaves '${permittedProperty}' unaltered when already prefixed and lowercased`, () => {
       const sanitizationMiddleware = createSanitizationMiddleware();
       const testTxLike = {
         jsonrpc: '2.0' as const,
@@ -142,7 +101,77 @@ describe('createSanitizationMiddleware', () => {
         method: 'eth_call',
         params: [
           {
-            from: blockRef,
+            [permittedProperty]: '0x123',
+          },
+        ],
+      };
+
+      sanitizationMiddleware(
+        testTxLike,
+        { jsonrpc: '2.0', id: 'any' },
+        noop,
+        noop,
+      );
+      expect(testTxLike.params[0]).toStrictEqual({
+        [permittedProperty]: '0x123',
+      });
+    });
+
+    it(`leaves '${permittedProperty}' unaltered when it has a non-string value`, () => {
+      const sanitizationMiddleware = createSanitizationMiddleware();
+      const testTxLike = {
+        jsonrpc: '2.0' as const,
+        id: '1',
+        method: 'eth_call',
+        params: [
+          {
+            [permittedProperty]: 100,
+          },
+        ],
+      };
+
+      sanitizationMiddleware(
+        testTxLike,
+        { jsonrpc: '2.0', id: 'any' },
+        noop,
+        noop,
+      );
+      expect(testTxLike.params[0]).toStrictEqual({ [permittedProperty]: 100 });
+    });
+
+    for (const blockRef of ['latest', 'pending', 'earliest']) {
+      it(`leaves '${permittedProperty}' unaltered when set to ${blockRef}`, () => {
+        const sanitizationMiddleware = createSanitizationMiddleware();
+        const testTxLike = {
+          jsonrpc: '2.0' as const,
+          id: '1',
+          method: 'eth_call',
+          params: [
+            {
+              [permittedProperty]: blockRef,
+            },
+          ],
+        } as any;
+
+        sanitizationMiddleware(
+          testTxLike,
+          { jsonrpc: '2.0', id: 'any' },
+          noop,
+          noop,
+        );
+        expect(testTxLike.params[0][permittedProperty]).toBe(blockRef);
+      });
+    }
+
+    it(`hex prefixes and lowercases '${permittedProperty}'`, () => {
+      const sanitizationMiddleware = createSanitizationMiddleware();
+      const testTxLike = {
+        jsonrpc: '2.0' as const,
+        id: '1',
+        method: 'eth_call',
+        params: [
+          {
+            [permittedProperty]: '123ABC',
           },
         ],
       } as any;
@@ -153,29 +182,7 @@ describe('createSanitizationMiddleware', () => {
         noop,
         noop,
       );
-      expect(testTxLike.params[0].from).toBe(blockRef);
+      expect(testTxLike.params[0][permittedProperty]).toBe('0x123abc');
     });
   }
-
-  it('hex prefixes and lowercases any tx param that is missing one', () => {
-    const sanitizationMiddleware = createSanitizationMiddleware();
-    const testTxLike = {
-      jsonrpc: '2.0' as const,
-      id: '1',
-      method: 'eth_call',
-      params: [
-        {
-          from: '123ABC',
-        },
-      ],
-    } as any;
-
-    sanitizationMiddleware(
-      testTxLike,
-      { jsonrpc: '2.0', id: 'any' },
-      noop,
-      noop,
-    );
-    expect(testTxLike.params[0].from).toBe('0x123abc');
-  });
 });
--- a/app/core/SanitizationMiddleware.test.ts
+++ b/app/core/SanitizationMiddleware.test.ts
@@ -7,7 +7,47 @@ const noop = () => {
 };
 
 describe('createSanitizationMiddleware', () => {
-  it('does nothing when the first param is not an object', () => {
+  it('does nothing if the request is something other than eth_call', () => {
+    const sanitizationMiddleware = createSanitizationMiddleware();
+    const req = {
+      jsonrpc: '2.0' as const,
+      id: '1',
+      method: 'something_else',
+      params: [
+        {
+          foo: '123',
+        },
+      ],
+    };
+
+    sanitizationMiddleware(req, { jsonrpc: '2.0', id: 'any' }, noop, noop);
+
+    expect(req.params).toStrictEqual([
+      {
+        foo: '123',
+      },
+    ]);
+  });
+
+  it('does nothing if the params are an object', () => {
+    const sanitizationMiddleware = createSanitizationMiddleware();
+    const req = {
+      jsonrpc: '2.0' as const,
+      id: '1',
+      method: 'something_else',
+      params: {
+        foo: '123',
+      },
+    };
+
+    sanitizationMiddleware(req, { jsonrpc: '2.0', id: 'any' }, noop, noop);
+
+    expect(req.params).toStrictEqual({
+      foo: '123',
+    });
+  });
+
+  it('does nothing if the first param is not an object', () => {
     const sanitizationMiddleware = createSanitizationMiddleware();
     const req = {
       jsonrpc: '2.0' as const,
@@ -18,10 +58,38 @@ describe('createSanitizationMiddleware', () => {
 
     sanitizationMiddleware(req, { jsonrpc: '2.0', id: 'any' }, noop, noop);
 
-    expect(req.params[0]).toBe('FOO');
+    expect(req.params).toStrictEqual(['FOO']);
   });
 
-  it('removes any property that is not transaction-like or topic-like', () => {
+  it('does nothing if the first param is an array', () => {
+    const sanitizationMiddleware = createSanitizationMiddleware();
+    const req = {
+      jsonrpc: '2.0' as const,
+      id: '1',
+      method: 'eth_call',
+      params: [['FOO']],
+    };
+
+    sanitizationMiddleware(req, { jsonrpc: '2.0', id: 'any' }, noop, noop);
+
+    expect(req.params).toStrictEqual([['FOO']]);
+  });
+
+  it('does nothing if the first param is null', () => {
+    const sanitizationMiddleware = createSanitizationMiddleware();
+    const req = {
+      jsonrpc: '2.0' as const,
+      id: '1',
+      method: 'eth_call',
+      params: [null],
+    };
+
+    sanitizationMiddleware(req, { jsonrpc: '2.0', id: 'any' }, noop, noop);
+
+    expect(req.params).toStrictEqual([null]);
+  });
+
+  it('removes any properties that are not transaction-like or topic-like', () => {
     const sanitizationMiddleware = createSanitizationMiddleware();
     const testTxLike = {
       jsonrpc: '2.0' as const,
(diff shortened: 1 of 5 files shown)


.node-version
```diff
@@ -0,0 +1 @@
+14.21.3
```
BelfordZ, line 1: remove this

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,94 @@
...
+export function createSanitizationMiddleware(): JsonRpcMiddleware<
+  unknown,
+  unknown
+> {
+  return (req: JsonRpcRequest<unknown>, _: any, next: () => any) => {
+    if (req.method !== 'eth_call') {
+      next();
+      return;
+    }
+    const params = req.params as unknown[];
```
Gudahtt, line 80: Nit: We can avoid casting by ensuring the params are an array ahead of time:

```suggestion
    if (req.method !== 'eth_call' || !Array.isArray(req.params)) {
      next();
      return;
    }
    const params = req.params;
```

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,94 @@
...
+    const txParams = params[0];
+
+    if (
+      typeof txParams === 'object' &&
+      !Array.isArray(txParams) &&
+      txParams !== null
+    ) {
+      const sanitized = cloneTxParams(txParams);
+      (req as JsonRpcRequest<any>).params[0] = sanitized;
```
Gudahtt, line 89: This case doesn't appear to be necessary:
```suggestion
      req.params[0] = sanitized;
```

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,94 @@
...
+  'gas',
+  'gasPrice',
+  'nonce',
+  'fromBlock',
+  'toBlock',
+  'address',
+  'topics',
+];
+
+function cloneTxParams(txParams: TransactionParamsLike | any) {
```
Gudahtt, line 36: Nit: Any type union including `any` can be reduced to `any`.

Instead maybe we can use the type `Record<PropertyKey, unknown>` here. By the time this function is called, we have confirmed that `txParams` is a non-null object, which is basically what `Record<PropertyKey, unknown>` would be.

Type checking doesn't really apply here because the `JsonRpcRequest` type we're using here has `any` in it (I wouldn't recommend bothering to try and improve that in this PR; I tried once by introducing `@metamask/utils` but it caused build failures on Android). But at least that type reflects what we ought to know by this point.

```suggestion
function cloneTxParams(txParams: Record<PropertyKey, unknown>) {
```

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,94 @@
...
+            sanitize(item),
+          );
+        } else {
+          copy[permitted] = sanitize(txParams[permitted]);
+        }
+      }
+      return copy;
+    },
+    {} as TransactionParamsLike,
+  );
```
Gudahtt, line 51: Thoughts on dropping the `TransactionParamsLike` type altogether?  It doesn't really line up with how this middleware functions in practice because any non-string values are returned unaltered. Applying that type would seem misleading and inaccurate.

If we use `Record<string, unknown>` here we don't need to do any casting:

```suggestion
  const sanitized = permittedKeys.reduce<Record<string, unknown>>(
    (copy, permitted) => {
      if (permitted in txParams) {
        const value = txParams[permitted];
        if (Array.isArray(value)) {
          copy[permitted] = value.map((item: unknown) => sanitize(item));
        } else {
          copy[permitted] = sanitize(value);
        }
      }
      return copy;
    },
    {},
  );
```

I introduced the `value` variable because TypeScript's type inference doesn't work well with properties. Pulling that out into a variable ensured TypeScript recognized it as an array after the `isArray` check.

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  return sanitized;
+}
+
+function sanitize(value: any) {
```
Gudahtt, line 56: No need to disable type checking here:
```suggestion
function sanitize(value: unknown) {
```

Gudahtt (commented):
This generally looks correct to me! Both the functionality and the way the middleware was integrated. I left a bunch of suggestions on improving the types.

The main question I have left is about whether we should restrict this just to `eth_call`. From reading the code, it still seems like any requests bound for the network (e.g. anything not intercepted by the `RPCMethodMiddleware` module) would be processed by this middleware.

I'll keep looking into that, but if you could share your method of testing that, that might be helpful. I recall you saying that you did some testing already.

Gudahtt:
I am investigating this:

>The main question I have left is about whether we should restrict this just to `eth_call`. From reading the code, it still seems like any requests bound for the network (e.g. anything not intercepted by the `RPCMethodMiddleware` module) would be processed by this middleware.

Once that question is answered, this will be ready for review

Gudahtt:
I have reviewed all supported RPC methods to see which might be affected by the santization middleware today in `web3-provider-engine`. I can't anticipate all hypothetical RPC methods supported by custom networks, so I restricted my search just to Infura-supported methods.

This middleware only affects requests with an object as the first parameter. The Infura methods with an object as the first parameter are:
- `eth_call`
- `eth_estimateGas`
- `eth_getLogs`
- `eth_newFilter`


`eth_newFilter` gets intercepted by mobile's RPC middleware, so the first three are all that we need to consider.

Gudahtt:
The only first-parameter property for `eth_getLogs` getting filtered out is `blockhash`. I have confirmed that `eth_getLogs` works on both extension and mobile, but on mobile the `blockhash` parameter is filtered out.

For `eth_call` and `eth_estimateGas`, the only properties getting filtered out are `maxPriorityFeePerGas` and `maxFeePerGas`.

It seems that the santization middleware was written to account for all known Infura/Ethereum RPC methods at the time, but doesn't count for parameters added in later EIPs.

This seems to confirm that this affects all methods that reach the network middleware, not just `eth_call`

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+// We use this to clean any custom params from the txParams
+// TODO: Update this to allow `blockhash` (used by `eth_getLogs`) and EIP-1559
+// transaction parameters
+// TODO: Consider making this middleware method-specific so that we can more
+// easily understand what it does
+// TODO: Consider rejecting invalid inputs, rather than silently dropping them
+// TODO: Consider deleting this middleware; we can let the RPC handle
+// validation
+const permittedKeys = [
+  'from',
```
mcmire, line 15: It seems that only `from` and `to` are being tested, but none of these other properties are being tested.

app/core/SanitizationMiddleware.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    sanitizationMiddleware(
+      testTxLike,
+      { jsonrpc: '2.0', id: 'any' },
+      noop,
+      noop,
+    );
+    expect(testTxLike.params[0]).toStrictEqual({});
+  });
+
+  it('does nothing when already prefixed and lowercased', () => {
```
mcmire, line 92: It seems that only `from` and `to` are being tested, but none of the other properties are being tested. It seems that all of the tests from here on out could be put into a loop.

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+    case 'pending':
+      return value;
+    case 'earliest':
+      return value;
+    default:
+      if (typeof value === 'string') {
+        return addHexPrefix(value.toLowerCase());
+      }
+      return value;
+  }
```
mcmire, line 77: Nit: This could be simplified to:

``` suggestion
  if (
    typeof value === 'string' &&
    !['latest', 'pending', 'earliest'].includes(value)
  ) {
    return addHexPrefix(value.toLowerCase());
  }
  return value;
```

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+  unknown,
+  unknown
+> {
+  return (req: JsonRpcRequest<unknown>, _: any, next: () => any) => {
+    if (!Array.isArray(req.params)) {
+      next();
+      return;
+    }
+    const params = req.params;
+    const txParams = params[0];
```
mcmire, line 98: Nit: This could be

``` suggestion
    const txParams = req.params[0];
```

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    if (
+      typeof txParams === 'object' &&
+      !Array.isArray(txParams) &&
+      txParams !== null
+    ) {
+      const sanitized = sanitizeRpcParameter(txParams);
+      req.params[0] = sanitized;
```
mcmire, line 106: Nit: This could be

``` suggestion
      req.params[0] = sanitizeRpcParameter(txParams);
```

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,111 @@
...
+ * @param parameter - The parameter to sanitize.
+ * @returns The given parameter containing just permitted keys.
+ */
+function sanitizeRpcParameter(parameter: Record<PropertyKey, unknown>) {
+  const sanitized = permittedKeys.reduce<Record<string, unknown>>(
+    (copy, permitted) => {
+      if (permitted in parameter) {
+        const value = parameter[permitted];
+        if (Array.isArray(value)) {
+          copy[permitted] = value.map((item: unknown) => sanitize(item));
```
mcmire, line 42: Nit: This could be

``` suggestion
          copy[permitted] = value.map(sanitize);
```

app/core/SanitizationMiddleware.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    const testTxLike = {
+      jsonrpc: '2.0' as const,
+      id: '1',
+      method: 'eth_call',
+      params: [
+        {
+          foo: '123',
+        },
+      ],
+    } as any;
```
mcmire, line 81: Nit: This typecasting is extraneous

``` suggestion
    };
```

app/core/SanitizationMiddleware.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    const testTxLike = {
+      jsonrpc: '2.0' as const,
+      id: '1',
+      method: 'eth_call',
+      params: [
+        {
+          from: '123ABC',
+        },
+      ],
+    } as any;
```
mcmire, line 171: Nit: This typecasting is extraneous

``` suggestion
    };
```

app/core/SanitizationMiddleware.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      const testTxLike = {
+        jsonrpc: '2.0' as const,
+        id: '1',
+        method: 'eth_call',
+        params: [
+          {
+            from: blockRef,
+          },
+        ],
+      } as any;
```
mcmire, line 148: Nit: This typecasting is extraneous

``` suggestion
      };
```

mcmire:
@Gudahtt I've double-checked your work against the Infura docs, and you are correct about `eth_call`, `eth_estimateGas`, `eth_getLogs`, and `eth_getFilterLogs`.

I also looked at the [Ethereum execution API docs](https://ethereum.github.io/execution-apis/api-documentation/), and it appears that some new RPC methods have been added since I last looked. These are the ones affected, with the parameters that would be excluded (skipping `eth_sendTransaction` since I think we override that one?):

- eth_createAccessList
  - type
  - input
  - maxPriorityFeePerGas
  - maxFeePerGas
  - accessList
  - chainId
- engine_forkchoiceUpdatedV1
  - headBlockHash
  - safeBlockHash
  - finalizedBlockHash
- engine_forkchoiceUpdatedV2
  - headBlockHash
  - safeBlockHash
  - finalizedBlockHash
- engine_newPayloadV1
  - parentHash
  - feeRecipient
  - stateRoot
  - receiptsRoot
  - logsBloom
  - prevRandao
  - blockNumber
  - gasLimit
  - gasUsed
  - timestamp
  - extraData
  - baseFeePerGas
  - blockHash
  - transactions
- engine_newPayloadV2
  - parentHash
  - feeRecipient
  - stateRoot
  - receiptsRoot
  - logsBloom
  - prevRandao
  - blockNumber
  - gasLimit
  - gasUsed
  - timestamp
  - extraData
  - baseFeePerGas
  - blockHash
  - transactions
  - withdrawals
- engine_exchangeTransitionConfigurationV1
  - terminalTotalDifficulty
  - terminalBlockHash
  - terminalBlockNumber

Not sure it is worth considering these, but I've noted them for completeness.

app/core/SanitizationMiddleware.test.ts
Gudahtt: Good idea! Done in https://github.com/MetaMask/metamask-mobile/pull/6079/commits/b3a17fbbd3397e0eb66f9d8c865dc382a4bdf056

app/core/SanitizationMiddleware.ts
Gudahtt: Done in 79b64b8de

app/core/SanitizationMiddleware.ts
Gudahtt: Done in 813103ba5

app/core/SanitizationMiddleware.test.ts
Gudahtt: Done in a1f46ddaa

app/core/SanitizationMiddleware.test.ts
Gudahtt: Done in a1f46ddaa

app/core/SanitizationMiddleware.test.ts
Gudahtt: Done in a1f46ddaa

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,110 @@
+'use strict';
```
mcmire, line 1: Nit: Is this line necessary?

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,110 @@
...
+        } else {
+          copy[permitted] = sanitize(value);
+        }
+      }
+      return copy;
+    },
+    {},
+  );
+
+  return sanitized;
```
mcmire, line 60: Nit: This can also just be

``` suggestion
  return permittedKeys.reduce<Record<string, unknown>>(
    (copy, permitted) => {
      if (permitted in parameter) {
        const value = parameter[permitted];
        if (Array.isArray(value)) {
          copy[permitted] = value.map(sanitize);
        } else {
          copy[permitted] = sanitize(value);
        }
      }
      return copy;
    },
    {},
  );
```

app/core/SanitizationMiddleware.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    sanitizationMiddleware(
+      testTxLike,
+      { jsonrpc: '2.0', id: 'any' },
+      noop,
+      noop,
+    );
+    expect(testTxLike.params[0]).toStrictEqual({});
+  });
+
+  for (const permittedProperty of permittedKeys) {
```
mcmire, line 95: It seems that the sanitization middleware also covers the case in which the property value is an array instead of a singular value. Should we also test for that?

app/core/SanitizationMiddleware.ts
Gudahtt: Nope! I missed that. Removed in https://github.com/MetaMask/metamask-mobile/pull/6079/commits/2e553d50f2b4d092f0a65d20c98af8771effd071

app/core/SanitizationMiddleware.ts
Gudahtt: Simplified here: https://github.com/MetaMask/metamask-mobile/pull/6079/commits/dc844cc9f8f3401d08b5cd382e6e110a860d7db5

app/core/SanitizationMiddleware.test.ts
Gudahtt: Good catch; tests added here: https://github.com/MetaMask/metamask-mobile/pull/6079/commits/9f5f38ceb12ea9c972df36ecc077bc4664b3aa05

mcmire (approved):
LGTM!

Gudahtt:
Bitrise runs
* [iOS e2e tests](https://app.bitrise.io/build/caffee41-3424-4c85-957b-479e3d7b6f1a) (passing)
* [Android e2e tests](https://app.bitrise.io/build/728bd334-4658-4bad-97fb-a5050a7b8135) (2 test failures that are [also failing on `main`](https://app.bitrise.io/build/f059119f-c8c1-4c71-b44e-4a4ad01a7b97))

app/core/SanitizationMiddleware.ts
```diff
@@ -0,0 +1,103 @@
+import { JsonRpcMiddleware, JsonRpcRequest } from 'json-rpc-engine';
+import { addHexPrefix } from 'ethereumjs-util';
+
+// We use this to clean any custom params from the txParams
+// TODO: Update this to allow `blockhash` (used by `eth_getLogs`) and EIP-1559
```
sethkfman, line 5: How urgent are these TODOs?

sethkfman (approved):
LGTM. Just one clarifying question.

app/core/SanitizationMiddleware.ts
Gudahtt: Up to you; these are pre-existing problems. This line for example refers to request parameter filtering that may be impacting dapps on mobile.

This has functioned this way since the initial release of mobile though, so none of these is "urgent" exactly.

Gudahtt: These topics are related to the broader discussion about how we should maintain the mobile dapp API and make it more consistent with extension.

cortisiko (approved):
I tested the test dapp as well as other dapp interactions. Nothing to report here. ✅

PR #6088, opened 2023-03-31
Use Node environment for e2e tests

Gudahtt:
**Development & PR Process**

**Description**

The Jest configuration used for Detox e2e tests has been updated to use a Node.js environment rather than React Native presets. This was a requirement for using Ganache as part of e2e tests (it was incompatible with some of the global changes introduced by the React Native environment).

The Jest process doesn't need to execute any React Native code directly as part of the e2e test suite. The only direct reference today was via the `i18n` module, which was accessed to retrieve localized messages. All such references were updated to use the JSON file directly, to avoid importing React Native libraries.

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/765

**Checklist**

### DIFF
--- a/app/component-library/components/Cells/Cell/variants/CellSelect/CellSelect.tsx
+++ b/app/component-library/components/Cells/Cell/variants/CellSelect/CellSelect.tsx
@@ -11,7 +11,7 @@ import SelectItem from '../../../../Select/Select/SelectItem';
 import CellBase from '../../foundation/CellBase';
 
 // Internal dependencies.
-import { CELL_SELECT_TEST_ID } from './CellSelect.constants';
+import { CELL_SELECT_TEST_ID } from '../../../../../../constants/test-ids';
 import styleSheet from './CellSelect.styles';
 import { CellSelectProps } from './CellSelect.types';
 import generateTestId from '../../../../../../../wdio/utils/generateTestId';
--- a/app/component-library/components/Avatars/Avatar/variants/AvatarNetwork/AvatarNetwork.constants.ts
+++ b/app/component-library/components/Avatars/Avatar/variants/AvatarNetwork/AvatarNetwork.constants.ts
@@ -11,5 +11,3 @@ export const TEST_REMOTE_IMAGE_SOURCE: ImageSourcePropType = {
 
 /* eslint-disable-next-line */
 export const TEST_LOCAL_IMAGE_SOURCE: ImageSourcePropType = require('../../../../../../images/ethereum.png');
-
-export const NETWORK_AVATAR_IMAGE_ID = 'network-avatar-image';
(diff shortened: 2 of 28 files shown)


Gudahtt:
It seems that the [static image imports](https://reactnative.dev/docs/images#static-image-resources) might be causing some trouble; I'll investigate.

Gudahtt:
The e2e test errors we encountered should now be resolved. They were caused by imports of test IDs in e2e tests that also brought in unrelated lines of code that would only work correctly in a React Native context. The test IDs have been migrated to an alternative constants file instead that just includes string literal constants, no imports.

app/component-library/components/Cells/Cell/variants/CellSelect/CellSelect.tsx
```diff
@@ -11,7 +11,7 @@ import SelectItem from '../../../../Select/Select/SelectItem';
 import CellBase from '../../foundation/CellBase';
 
 // Internal dependencies.
-import { CELL_SELECT_TEST_ID } from './CellSelect.constants';
+import { CELL_SELECT_TEST_ID } from '../../../../../../constants/test-ids';
```
NicolasMassart, line 14: REQ: 
- this test id constant is used twice in the component, only one testID prop should be added. [Line 36](https://github.com/MetaMask/metamask-mobile/pull/6088/files#diff-0fd6dbdb0f9e3ae036c54f237f82817f617ee7a57f4d8a33e0c9e34f1432e899R36) `testID={CELL_SELECT_TEST_ID}` could be removed.
- test id is not unique. This item is a select item that will appear in a list in the app. It works here in the integration test isolated context, but will fail in e2e as we won't be able to target a specific item if ids are not unique for each item of the list. A suffix should be used to make this unique like the cell base title or something from the avatar props.

NicolasMassart (changes requested):

app/component-library/components/Cells/Cell/variants/CellSelect/CellSelect.tsx
Gudahtt: This is a pre-existing problem though right? I'm not introducing this test ID in this PR. I just had to move it in order to avoid React Native style png imports in the Jest test runner process.

NicolasMassart (approved):
Looks good to me.
The testID change will have to be handled in another PR, we have MetaMask/mobile-planning#856 tracking this

Gudahtt:
I ran the iOS e2e tests on Bitrise here: https://app.bitrise.io/build/c4dfd2cd-7316-4d2d-b37a-7c8cd8a2a15a

Like on `main`, it timed out and the deeplink tests failed. Presumably that is expected for now, until it's fixed on main.

But there was one additional failure, for the test `Browser Tests › should test phishing sites`. I'm investigating that now.

Gudahtt:
It looks like `main` has the same failure at the moment. This branch passes all tests that are passing on `main`.

PR #6089, opened 2023-03-31
Add support for using Ganache in Detox iOS e2e tests

Gudahtt:
**Development & PR Process**

**Description**

We can now use Ganache in iOS Detox e2e tests. This lets us write tests that use network currency without relying upon test ETH in the account used for testing, and it reduces reliance upon the network.

The `advanced-gas-fees` test suite has been updated to use ganache, to demonstrate that this integration works.

**Issue**

Fixes https://github.com/MetaMask/mobile-planning/issues/765

**Checklist**

### DIFF
--- a/app/util/test/ganache.js
+++ b/app/util/test/ganache.js
@@ -3,8 +3,6 @@ import ganache from 'ganache';
 const defaultOptions = {
   blockTime: 2,
   network_id: 1337,
-  mnemonic:
-    'phrase upgrade clock rough situate wedding elder clever doctor stamp excess tent',
   port: 8545,
   vmErrorsOnRPCResponse: false,
   hardfork: 'muirGlacier',
@@ -13,6 +11,9 @@ const defaultOptions = {
 
 export default class Ganache {
   async start(opts) {
+    if (!opts.mnemonic) {
+      throw new Error('Missing required mnemonic');
+    }
     const options = { ...defaultOptions, ...opts };
     const { port } = options;
     this._server = ganache.server(options);
--- a/app/util/test/ganache.js
+++ b/app/util/test/ganache.js
@@ -1,0 +1,53 @@
+import ganache from 'ganache';
+
+const defaultOptions = {
+  blockTime: 2,
+  network_id: 1337,
+  mnemonic:
+    'phrase upgrade clock rough situate wedding elder clever doctor stamp excess tent',
+  port: 8545,
+  vmErrorsOnRPCResponse: false,
+  hardfork: 'muirGlacier',
+  quiet: true,
+};
+
+export default class Ganache {
+  async start(opts) {
+    const options = { ...defaultOptions, ...opts };
+    const { port } = options;
+    this._server = ganache.server(options);
+    await this._server.listen(port);
+  }
+
+  getProvider() {
+    return this._server.provider;
+  }
+
+  async getAccounts() {
+    return await this.getProvider().request({
+      method: 'eth_accounts',
+      params: [],
+    });
+  }
+
+  async getBalance() {
+    const accounts = await this.getAccounts();
+    const balanceHex = await this.getProvider().request({
+      method: 'eth_getBalance',
+      params: [accounts[0], 'latest'],
+    });
+    const balanceInt = parseInt(balanceHex, 16) / 10 ** 18;
+
+    const balanceFormatted =
+      balanceInt % 1 === 0 ? balanceInt : balanceInt.toFixed(4);
+
+    return balanceFormatted;
+  }
+
+  async quit() {
+    if (!this._server) {
+      throw new Error('Server not running yet');
+    }
+    await this._server.close();
+  }
+}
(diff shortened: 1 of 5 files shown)


Gudahtt:
This depends upon #6088

app/util/test/ganache.js
```diff
@@ -0,0 +1,53 @@
...
+    const { port } = options;
+    this._server = ganache.server(options);
+    await this._server.listen(port);
+  }
+
+  getProvider() {
+    return this._server.provider;
+  }
+
+  async getAccounts() {
```
Gudahtt, line 26: This file was taken from the extension repository. We don't use these functions anywhere at the moment, but I left them here because they might be useful in tests.

Gudahtt:
@SocketSecurity ignore ganache@7.7.7
@SocketSecurity ignore leveldown@6.1.0

`leveldown` is used by `ganache`. Both libraries have legitimate needs for using native code and are non-malicious.

e2e/viewHelper.js
```diff
@@ -71,6 +76,33 @@ export const importWalletWithRecoveryPhrase = async () => {
   }
 };
 
+export const addLocalhostNetwork = async () => {
```
Gudahtt, line 79: It would be great to include this as a built-in network instead, in builds intended for e2e tests. We don't have a build type for e2e tests yet though, so that can be a future improvement.

e2e/specs/confirmations/send-eth-flow.spec.js
```diff
@@ -36,15 +47,14 @@ describe('Send ETH Tests', () => {
...
-  it('should switch currency from crypto to fiat and back to crypto', async () => {
-    await AmountView.typeInTransactionAmount('0.004');
-    await AmountView.tapCurrencySwitch();
-    await AmountView.isTransactionAmountConversionValueCorrect(
-      '0.004 GoerliETH',
-    );
-    await AmountView.tapCurrencySwitch();
-    await AmountView.isTransactionAmountCorrect('0.004');
-  });
+  // TODO: Add support for conversion rate on localhost during e2e tests
```
Gudahtt, line 50: We solve this on the extension by making some modifications to the application on builds intended for e2e tests. In this case, we would use Ethereum Mainnet conversion rates for ganache in an e2e test build. That can be done in a future PR

e2e/viewHelper.js
```diff
@@ -71,6 +76,33 @@ export const importWalletWithRecoveryPhrase = async () => {
...
+  await NetworkView.isNetworkViewVisible();
+
+  await TestHelpers.delay(3000);
+  await NetworkView.tapAddNetworkButton();
+  await NetworkView.switchToCustomNetworks();
+
+  await NetworkView.typeInNetworkName('Localhost');
+  await NetworkView.typeInRpcUrl(LOCALHOST_URL);
+  await NetworkView.typeInChainId('1337');
+  await NetworkView.typeInNetworkSymbol('TST\n');
```
seaona, line 95: could we set the network symbol as `ETH`?

e2e/specs/confirmations/send-eth-flow.spec.js
```diff
@@ -57,7 +67,7 @@ describe('Send ETH Tests', () => {
 
   it('should send ETH to Account 2', async () => {
     // Check that the amount is correct
-    await TransactionConfirmationView.isTransactionTotalCorrect('0 GoerliETH');
+    await TransactionConfirmationView.isTransactionTotalCorrect('0 TST');
```
seaona, line 70: If we set network symbol as `ETH` here we could replace `TST` by `ETH`

seaona (approved):
Overall LGTM! I've added one small comment

e2e/viewHelper.js
seaona: I suggest this just to avoid confusion.
I would reserve the `ETH`symbol for the native token from the network as in https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/fixture-builder.js#L204. 
So whenever we sent ETH, it's displayed as ETH, and whenever we sent an ERC20 we display `TST`, since it's the token symbol from the ERC20 token from the test dapp.
Let me know if this makes sense @Gudahtt

Gudahtt: Ah right, good point, I will make this change

e2e/specs/confirmations/send-eth-flow.spec.js
Gudahtt: Done here: https://github.com/MetaMask/metamask-mobile/pull/6089/commits/4bdf25b069b1e5459d775c7af248b0e4f2884d30

cortisiko (approved):
Looks good 🌮 🌮

Gudahtt:
Tested on Bitrise here: https://app.bitrise.io/build/9a44fa75-29e5-4f6e-80f0-956bfe25d32f

Only iOS tests run, due to known errors on Android e2e tests. iOS tests have timed out and 2 have failed, but the failures seem unrelated to these changes. The only affected workflow in this PR passes.

Gudahtt:
Switched to using `beforeAll`/`afterAll` because these tests all run in the same application context, unlike extension which spins up a new wallet/state for each `it` block. This seems to have resolved the error seen upon running this locally repeatedly as well.

PR #6148, opened 2023-04-07
Added Security and Privacy WebDriver.IO e2e test scripts

SamuelSalas:
**Description**
Added the Added Security and Privacy WebDriver.IO e2e feature file with the respective step definitions. Also, solve a couple of flakiness of the wdio test scripts. Added driver and $ variable to .eslintrc.js file.

**Checklist**


wdio/screen-objects/Modals/SkipAccountSecurityModal.js
```diff
@@ -16,8 +15,8 @@ class SkipAccountSecurityModal {
   }
 
   async proceedWithoutWalletSecure() {
-    const settimeout = 2000;
-    await driver.pause(settimeout);
+    const setTimeout = 2000;
+    await driver.pause(setTimeout);
```
NicolasMassart, line 19: Q: could this timeout be removed by adding a wait for the element you want to tap to be actually visible?

wdio/step-definitions/import-wallet-via-private-key.steps.js
```diff
@@ -32,11 +32,11 @@ Then(/^The account is imported/, async () => {
 
 Then(/^I am on the imported account/, async () => {
   await driver.pause(2500);
-  WalletAccountModal.isAccountNameLabelEqualTo('Account 3'); // this can be better
+  await WalletAccountModal.isAccountNameLabelEqualTo('Account 2'); // this can be better
```
NicolasMassart, line 35: C: interesting comment `// this can be better` 😅  Coud we explain more about how this could be better?

app/components/UI/SettingsButtonSection/index.tsx
```diff
@@ -53,6 +55,7 @@ const SettingsButtonSection = ({
           onPress={onButtonPress}
           containerStyle={styles.confirmButton}
           disabled={buttonDisabled}
+          testID={testID}
```
NicolasMassart, line 58: Q: why not using the generateTestId() function here?

wdio/step-definitions/import-wallet-via-private-key.steps.js
Andepande: we could parameterise the Account Name label assertion from cucumber rather than make call it here

Andepande (approved):

Andepande (approved):
LGTM, resolve the comments raised by Nicholas, you may also need to add test results links for browserStack

app/components/UI/SettingsButtonSection/index.tsx
SamuelSalas: Because this is testID is a prop defined for the StyleButton component. The generateTestId method is executed inside that component.

wdio/screen-objects/Modals/SkipAccountSecurityModal.js
SamuelSalas: Probably, i just change the lower case t to upper case.

NicolasMassart (approved):
Looks good to me.

PR #6166, opened 2023-04-11
feat: trigger walletconnect modal using approval controller

OGPoyraz:
**Description**

This PR aims to make WalletConnect use `ApprovalController` and its own accept/reject mechanism instead of `walletconnectSessionRequest` `walletconnectSessionRequest::approved` `walletconnectSessionRequest::rejected`.

UI flow of WalletConnect is not affected, only functional changes.

**Screenshots/Recordings**

https://user-images.githubusercontent.com/7644512/231426627-0e90b20c-54a9-49c9-bbb3-6f0bd6bc87ce.mp4

**Issue**
https://github.com/MetaMask/mobile-planning/issues/768

**Checklist**

### DIFF
--- a/app/components/Nav/Main/RootRPCMethodsUI.js
+++ b/app/components/Nav/Main/RootRPCMethodsUI.js
@@ -477,6 +477,7 @@ const RootRPCMethodsUI = (props) => {
       walletConnectRequestInfo.id,
       walletConnectRequestInfo.data,
     );
+    setWalletConnectRequestInfo(undefined);
   };
 
   const onWalletConnectSessionRejected = () => {
@@ -485,9 +486,11 @@ const RootRPCMethodsUI = (props) => {
       walletConnectRequestInfo.id,
       ethErrors.provider.userRejectedRequest(),
     );
+    setWalletConnectRequestInfo(undefined);
   };
 
   const renderWalletConnectSessionRequestModal = () => {
+    const meta = walletConnectRequestInfo?.data?.peerMeta || null;
     return (
       <Modal
         isVisible={showPendingApproval?.type === ApprovalTypes.WALLET_CONNECT}
--- a/app/components/Nav/Main/RootRPCMethodsUI.js
+++ b/app/components/Nav/Main/RootRPCMethodsUI.js
@@ -488,10 +488,6 @@ const RootRPCMethodsUI = (props) => {
   };
 
   const renderWalletConnectSessionRequestModal = () => {
-    const meta = walletConnectRequestInfo?.data?.peerMeta || null;
-    // walletConnectRequestInfo is reset to undefined in onWalletConnectSessionApproval as soon as the approval happens
-    if (!meta) return;
-
     return (
       <Modal
         isVisible={showPendingApproval?.type === ApprovalTypes.WALLET_CONNECT}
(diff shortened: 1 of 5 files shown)


app/core/WalletConnect.js
```diff
@@ -339,21 +340,19 @@ class WalletConnect {
...
-        }
-      });
-      hub.on('walletconnectSessionRequest::rejected', (peerId) => {
-        if (peerInfo.peerId === peerId) {
-          reject(new Error('walletconnectSessionRequest::rejected'));
-        }
+  sessionRequest = async (peerInfo) => {
+    const { ApprovalController } = Engine.context;
+    try {
+      return await ApprovalController.addAndShowApprovalRequest({
```
matthewwalsh0, line 346: We can just use the `add` method here for clarity as there is no show callback provided to the `ApprovalController` in mobile as it's handled by React and local state.

OGPoyraz: Yes that sounds much better, fixed.

app/core/WalletConnect.js
```diff
@@ -339,21 +340,19 @@ class WalletConnect {
...
+        id: peerInfo.peerId,
```
matthewwalsh0, line 347: Is the `peerId` unique or is it fixed per dApp? Wondering if it's safer to generate a random value rather than potentially reuse old IDs, especially if we persist in future.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -458,25 +472,29 @@ const RootRPCMethodsUI = (props) => {
   };
 
   const onWalletConnectSessionApproval = () => {
-    const { peerId } = walletConnectRequestInfo;
-    setWalletConnectRequestInfo(undefined);
-    WalletConnect.hub.emit('walletconnectSessionRequest::approved', peerId);
+    setShowPendingApproval(false);
```
matthewwalsh0, line 475: Could we also use `setWalletConnectRequestInfo` here and on reject to clear the domain state also, just for tidiness sake?

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -458,25 +472,29 @@ const RootRPCMethodsUI = (props) => {
...
+      walletConnectRequestInfo.id,
+      ethErrors.provider.userRejectedRequest(),
+    );
   };
 
   const renderWalletConnectSessionRequestModal = () => {
-    const meta = walletConnectRequestInfo?.peerMeta || null;
+    const meta = walletConnectRequestInfo?.data?.peerMeta || null;
     // walletConnectRequestInfo is reset to undefined in onWalletConnectSessionApproval as soon as the approval happens
     if (!meta) return;
```
matthewwalsh0, line 493: Could we remove this early return so it's consistent with the other `renderX` methods and just uses the `isVisible` to hide?

app/core/WalletConnect.js
matthewwalsh0, line 1: I think this is the perfect time to create some unit tests for this file, at least to verify the creation o the approval requests.

OGPoyraz: Added basic approved and rejected test cases.

matthewwalsh0 (approved):

PR #6213, opened 2023-04-19
Cleanup approve transaction review

### LINKED ISSUE #6159: ApproveTransactionReview - extract function to render tooltip into a separate component.
Extract the function to render tooltip [here](https://github.com/MetaMask/metamask-mobile/blob/f5d3bb82924bce231fee76ef29d7ba077886bc17/app/components/UI/ApproveTransactionReview/index.js#L564) into a separate component.

- the new component can be placed [here](https://github.com/MetaMask/metamask-mobile/tree/f5d3bb82924bce231fee76ef29d7ba077886bc17/app/components/UI/ApproveTransactionReview)
- all tooltip related functions / callbacks should be moved to new component
- ensure that new component has test coverage

jpuri:
Fixes: #6159

The task was about extracting out tooltip component from ApproveTransactionReview component but I found that these components were actually not being used. Thus I delete them from code.

QA Steps:
- The PR only removes un-used code, a check can be done to ensure that gas tooltip on confirmation pages work as before.

Screenshot 2023-04-19 at 2 18 32 PM


app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -1011,40 +957,10 @@ class ApproveTransactionReview extends PureComponent {
     });
   };
 
-  renderQRDetails() {
```
blackdevelopa, line 1014: Hey @jpuri, just checking why this was removed. I can't seem to find where it was moved to. TY

blackdevelopa (approved):
It seem `isSigningQRObject` isn't passed to the ApproveTransactionReview component and so has no value. So yeah, lgtm. Great find, @jpuri

app/components/UI/ApproveTransactionReview/index.js
segun: +1

gantunesr: Hi team, I have the same question as @blackdevelopa. As far as I know, that component is used with QR based hardware wallets like keystone

jpuri: This is dead code, it is not reachable and its safe to get rid of.

segun (approved):

plasmacorral (changes requested):
In a spot check I am I encountering an issue with spending approvals when using Keystone. After tapping approve, the button becomes disabled but the QR modal fails to open and present the composed transaction (to be scanned with Keystone).

iPhone 13 mini
iOS 15.6.1

https://recordit.co/FDiMZnGtz5

Steps to reproduce:
1. Import SRP
2. bind with Keystone
3. Be on testnet or change to it (Sepolia)
4. Open test dapp
5. tap `Create Token` under Send Token (patience required here)
6. `Add token to wallet` and `Transfer tokens` (may not be required steps)
7. tap `Approve tokens`
8. Tap `Approve`
9. Note absence of QR

jpuri:
Hey @gantunesr , @plasmacorral : thanks a lot for helping with this. @plasmacorral : reported that this PR indeed has issue with QR scanning flow. I have updated the PR to remove QR code related cleanup. I will create a separate PR for cleaning that up.

seaona:
From QA changes look good @jpuri . The tooltip is working as before in the different flows (Send ETH / Send token / Contract Interaction ..) and in different networks


https://user-images.githubusercontent.com/54408225/235204056-74e254b6-2f07-4527-b7ad-29c961270763.mp4

PR can be merged. Whenever I get the Keystone, I can QA the other changes. 

PS Thank you @plasmacorral for doing an initial QA on the QR code part !

plasmacorral:
Just want to confirm no issues observed with Keystone approval in commit 99fa02b56b931b92e9ab9b5aa2e4f6054f65d1de on the same iphone I previously tested with.

https://recordit.co/wrCPIUufPa

plasmacorral (approved):

PR #6214, opened 2023-04-19
Use core signature controller

matthewwalsh0:
**Description**

Replace direct usage of the message managers with the `SignatureController` provided by `@metamask/signature-controller` to improve consistency with the extension and to simplify the related code. 

**Checklist**


temp/signature-controller/dist/SignatureController.js
matthewwalsh0, line 1: The dependency is loaded via this directory temporarily until `@metamask/signature-controller` is released.

matthewwalsh0:
@SocketSecurity ignore-all

ios/Podfile.lock
```diff
@@ -874,4 +874,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 74e503a275b4ed097cb01c171027882e9cdc699d
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.11.3
```
vinistevam, line 877: Just out of curiosity, do you know what this file is about?

vinistevam (approved):

ios/Podfile.lock
OGPoyraz: It's a lock file for pods, like yarn.lock
I think this is getting updated with your cocoapods version, so we better leave it `1.11.2`

app/core/Engine.ts
```diff
@@ -410,6 +406,34 @@ class Engine {
...
+              `${approvalController.name}:addRequest`,
+              `${approvalController.name}:acceptRequest`,
+              `${approvalController.name}:rejectRequest`,
+            ],
+          }),
+          isEthSignEnabled: () =>
+            Boolean(
+              preferencesController.state?.disabledRpcMethodPreferences
+                ?.eth_sign,
+            ),
```
bergarces, line 422: I'm not even sure if regular signature is supported at all in mobile, but this should set it to false anyway.

bergarces (approved):

app/core/Engine.ts
matthewwalsh0: It's the same as the extension so there is a toggle in the advanced settings to enable it, have tested manually via the test dApp.

Issue #6158, opened 2023-04-11
First time build issue


dswilson4:
### **Description**

Was building locally for the first time today but am running into issues that I'm looking to see if others have also encountered recently.

### **Technical Details**

Environment

- Mac OS: Ventura 13.3.1
- Xcode version 14.3 & Command line tools 14.3
- IOS Device: Iphone 12 pro, IOS 16.4 (latest)

1. The first error with the build I’m seeing is :

```
ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It seems the newest version of Ventura has this issue where the arc folder is no longer present where expected, here’s a relevant thread with the resolution I’ve followed (copy previous libarclite_iphonesimulator.a file into new `arc` folder): [https://developer.apple.com/forums/thread/727527](https://developer.apple.com/forums/thread/727527)

2. After resolving the first issue, I’m seeing this build error error with `yoga` in React Native 0.66.5:

```
node_modules/react-native/ReactCommon/yoga/yoga/Yoga.cpp:2232:9: error: use of bitwise '|' with boolean operands [-Werror,-Wbitwise-instead-of-logical]
node->getLayout().hadOverflow() |
```

This issue was raised in early February tracked here in FB’s RN repo: [https://github.com/facebook/yoga/issues/1224](https://github.com/facebook/yoga/issues/1224)

The issue was fixed in April 2022, but versions of RN > 0.66.   Should developers instead use an older version of XCode assuming the intended Metamask RN version is 0.66.5?

### **Alternative**
I also tried uninstalling Xcode 14 and installing Xcode 13 instead, but it is incompatible with Ventura OS.  Does anybody else see any of the above issues when trying to build on a clean machine and test on an IOS simulator?

YDisa:
I am on the same issue.

After the xcode update,
My client's old app doesn't build
(Builded before update)

---
xcode : Version 14.3 (14E222b)
react-native : 0.66.2

dswilson4:
@YDisa I haven't made any progress so far

wantaekchoi:
same problem with me

roti-c137:
Temporary fix is to click on the error to auto navigate to the file and line of code, and make the fix as suggested

wantaekchoi:
[...]

PR #6220, opened 2023-04-19
Upgrade React Native to 0.71.6

jpcloureiro:
# Description

React Native upgrade to 0.71.6 requires

## Upgrade Helper

By using React-native upgrade helper (https://react-native-community.github.io/upgrade-helper/) we were able to bump one major version at a time. This approach allows us to handle specific version bump issues separately.

Most of the native (ios, android) changes were grabbed from this helper

## React native Reanimated

Using Reanimated 2 we have the option to use Reanimated API v1 or V2. https://docs.swmansion.com/react-native-reanimated/docs/2.x/fundamentals/migration

Reanimated team added support for RN 0.71 on 2.14.4 but it is crashing our app on iOS native side.

We are forced to use Reanimated 3. This version only ships API v2, forcing us to do the V1 -> V2 migration as well as the `gesture-handler` that depends on it.

## Hermes engine on Android

JSC does not fully support BigInt operations & that is a requirement for our app.

hermes version that comes with 0.71.6 crashes when some specific errors are thrown.

 We can't use the latest version of hermes on 0.71.6 because its API is not compatible.

commit that fixes the crash https://github.com/facebook/hermes/commit/f2f1a2e38e86632f24176d9e607c06507f900496
We checked out https://github.com/facebook/hermes/releases/tag/hermes-2023-03-07-RNv0.71.4-31fdcf738940875c9bacf251e149006cf515d763, and cherry-picked the fix.

here's the tarball we are using https://github.com/MetaMask/hermes/tree/ae2ac688c3faf868d2360a10a7791226ed5702fb

`hermes-engine/build.gradle` had to be patched to allow the build process to fetch the tarball from our fork.

## Why not Hermes on iOS as well?
We use JSC on iOS because it already has the necessary BigInt support. We tested hermes on iOS & it was unstable on our project (crashes)

**Issue**

Progresses https://github.com/metamask/mobile-planning/issues/690

**Checklist**


leotm:
`eth-rpc-errors+4.0.3.patch` now removed https://github.com/MetaMask/metamask-mobile/pull/6220/commits/8da441569ddbb283dca35f3e6b9b2997d9ee5c0f
favouring our custom Hermes build from before https://github.com/MetaMask/metamask-mobile/pull/6220/commits/8a5dce5433b0ca5fc5075e047b5d3e21c2ff2b56
no longer able to repro Android crash after connecting

Screenshot_1683659894

michaelmccallam:
Release labels added by mistake - I think this was an integration issue - will remove manually and investigate the root cause.

app/components/Views/Wallet/index.test.tsx
```diff
@@ -145,8 +144,8 @@ describe('Wallet', () => {
     expect(wrapper).toMatchSnapshot();
   });
   it('should render scan qr icon', () => {
-    render(Wallet);
-    const scanButton = screen.getByTestId('wallet-scan-button');
```
wachunei, line 149: This change should not be done, I recently learned we are following this pattern https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

scripts/build.sh
```diff
@@ -134,8 +134,8 @@ prebuild_ios(){
 }
 
 prebuild_android(){
-	adb kill-server
-	adb start-server
+	# adb kill-server
```
jpcloureiro, line 137: undo this

jpcloureiro:
@SocketSecurity ignore shell-quote@1.6.1

app/util/transactions/index.js
```diff
@@ -1210,7 +1210,7 @@ export const parseTransactionLegacy = (
   const parsedTicker = getTicker(ticker);
   const transactionFee = `${renderFromWei(weiTransactionFee)} ${parsedTicker}`;
 
-  const totalHex = valueBN.add(hexToBN(weiTransactionFee));
+  const totalHex = valueBN.add(weiTransactionFee);
```
Cal-L, line 1213: I'm assuming this was changed because `weiTransactionFee` is already in the form of BN? If so, do we know why the underlying value changed and where it originated from?

app/component-library/components/Accordions/Accordion/Accordion.tsx
```diff
@@ -32,23 +28,23 @@ const Accordion: React.FC<AccordionProps> = ({
 }) => {
   const { styles } = useStyles(styleSheet, { style });
   const [expanded, setExpanded] = useState(isExpanded);
-  const ref = useRef<TransitioningView>(null);
```
Cal-L, line 35: TODO - Add this back in later PR

Cal-L (commented):
Left some comments

app/util/transactions/index.js
jpcloureiro: `stripHexPrefix` changed its behavior. it now throws an error when the argument is not a string.

```
const stripHexPrefix = (str) => {
    if (typeof str !== 'string')
        throw new Error(`[stripHexPrefix] input must be type 'string', received ${typeof str}`);
    return isHexPrefixed(str) ? str.slice(2) : str;
};
```

This function comes from `ethereumjs-util` package that had a minor version bump, from `7.0.10` to `7.1.5`
This version bump happened when we had to regenerate `yarn.lock`

This is how its implemented on `7.0.10` 

```
const stripHexPrefix = (str) => {
  if (typeof str !== 'string') {
    return str;
  }

  return isHexPrefixed(str) ? str.slice(2) : str;
}
```

This change affects `hexToBN` because the function is called inside.

```
export function hexToBN(inputHex: string) {
  return new BN(stripHexPrefix(inputHex), 16);
}
```

So now every time this function is called without an argument of type string, it will throw an error.
A util function was created as a catchall function while we don't fix all the calls that pass a non-string argument.

https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-a6357366a5d16e5c4784e82b73bae918a7eb165fc7e63146d759760fa1b6cd59

```
/**
 * Converts a hex string to a BN object.
 * Adapt function with non string argument handler
 *
 * @param inputHex - Number represented as a hex string.
 * @returns A BN instance.
 */
export const hexToBN = (inputHex) =>
  typeof inputHex !== 'string'
    ? new BN(inputHex, 16)
    : controllerHexToBN(inputHex);
```

`weiTransactionFee` value was already a BN.

This was the first occurrence we detected. Reverting this line would go against our goal of fixing all `hextToBN` calls & eventually remove the catchall function safely.

app/component-library/components/Accordions/Accordion/Accordion.tsx
jpcloureiro: Should we create a PR already or a GH issue?

app/components/Views/Wallet/index.test.tsx
jpcloureiro: Good catch

addressed in [99d469e](https://github.com/MetaMask/metamask-mobile/pull/6220/commits/99d469ece69de637448aa8c9dc49ee035aed5c0a)

app/component-library/components/Accordions/Accordion/Accordion.tsx
Cal-L: Created a ticket here to track - https://github.com/MetaMask/metamask-mobile/issues/6525

android/app/build.gradle
```diff
@@ -1,3 +1,66 @@
...
+    //   The hermes compiler command to run. By default it is 'hermesc'
+    // hermesCommand = "$rootDir/my-custom-hermesc/bin/hermesc"
+    //
+    //   The list of flags to pass to the Hermes compiler. By default is "-O", "-output-source-map"
+    // hermesFlags = ["-O", "-output-source-map"]
+}
+
+project.ext.react = [
+	entryFile: "index.js",
+	enableHermes: true //System.getenv('METAMASK_ENVIRONMENT') == 'production'
```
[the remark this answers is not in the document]
jpcloureiro: That's great. Thanks for checking.

addressed in [36cba55](https://github.com/MetaMask/metamask-mobile/pull/6220/commits/36cba5580773f9432c1865a9e87ef9a3f70c1d24)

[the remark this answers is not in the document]
leotm: np ^ i think the commit go [deleted](https://github.com/MetaMask/metamask-mobile/assets/1881059/97492405-11b8-4456-ac05-6e9cf2d11b27) so we need to reapply the change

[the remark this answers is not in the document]
leotm: just checking further looks like our [`project.ext.react = [`](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R52) array been replaced with [`react {`](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R10) obj
so we can still remove `enableHermes: true` ^
but move our [`entryFile: "index.js"`](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R53) logic to [within](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R39) `react {`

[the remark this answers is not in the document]
leotm: (as an aside, i wonder if our `System.getenv('METAMASK_ENVIRONMENT') == 'production'` comment is still relevant?)

android/app/build.gradle
```diff
@@ -1,3 +1,66 @@
...
+]
+
+ project.ext.sentryCli = [
```
leotm, line 57: > just checking further looks like our [`project.ext.react = [`](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R52) array been replaced with [`react {`](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R10) obj so we can still remove `enableHermes: true` ^ but move our [`entryFile: "index.js"`](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R53) logic to [within](https://github.com/MetaMask/metamask-mobile/pull/6220/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R39) `react {`

noting the above ^ i'm going to check if we need to refactor our Sentry (Groovy) config here as well

leotm: yep all good ^ false alarm
https://github.com/getsentry/sentry-react-native/blob/main/sample/android/app/build.gradle#L70C26-L77
_https://github.com/getsentry/sentry-react-native/blob/main/sample-new-architecture/android/app/build.gradle#L63-L70_

android/app/build.gradle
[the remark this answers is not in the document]
Cal-L: Good catch. Removed project.ext.react

Cal-L:
https://github.com/SocketSecurity ignore-all

leotm (dismissed):
thanks for hopping on call with me @Cal-L, lgtm

patches/react-native+0.71.6.patch
```diff
@@ -0,0 +1,130 @@
...
+diff --git a/node_modules/react-native/ReactAndroid/hermes-engine/build.gradle b/node_modules/react-native/ReactAndroid/hermes-engine/build.gradle
+index c714f87..dca2750 100644
+--- a/node_modules/react-native/ReactAndroid/hermes-engine/build.gradle
++++ b/node_modules/react-native/ReactAndroid/hermes-engine/build.gradle
+@@ -43,11 +43,11 @@ def overrideHermesDir = System.getenv("REACT_NATIVE_OVERRIDE_HERMES_DIR") != nul
+ def hermesDir = System.getenv("REACT_NATIVE_OVERRIDE_HERMES_DIR") ?: new File(reactNativeRootDir, "sdks/hermes")
+ def hermesBuildDir = new File("$buildDir/hermes")
+ 
+-def hermesVersion = "main"
+-def hermesVersionFile = new File(reactNativeRootDir, "sdks/.hermesversion")
```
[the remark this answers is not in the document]
leotm: we want to keep this part to avoid using the default RN 0.71.6 [`sdks/.hermesversion`](https://github.com/facebook/react-native/blob/v0.71.6/sdks/.hermesversion)
so [`hermes-2023-03-07-RNv0.71.4-31fdcf738940875c9bacf251e149006cf515d763`](https://github.com/facebook/hermes/releases/tag/hermes-2023-03-07-RNv0.71.4-31fdcf738940875c9bacf251e149006cf515d763)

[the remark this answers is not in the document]
leotm: since the tag's too old ^ so doesn't include the `eth-rpc-errors` crashfix commit we want/cherry-picked https://github.com/facebook/hermes/commit/f2f1a2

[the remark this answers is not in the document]
leotm: we did [patch](https://github.com/MetaMask/metamask-mobile/pull/6220/commits/794e7c05e41e9ac52da67391b1ca5374579fc1d5) `eth-rpc-errors` earlier, but better instead side w official solution closest to core via our [custom Hermes build](https://hermesengine.dev/docs/react-native-integration/#using-a-custom-hermes-build-in-a-react-native-app)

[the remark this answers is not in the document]
leotm: so below we're just downloading the Hermes tarball from our fork instead
https://github.com/MetaMask/hermes/tree/ae2ac688c3faf868d2360a10a7791226ed5702fb

(which is [`hermes-2023-03-07-RNv0.71.4-31fdcf738940875c9bacf251e149006cf515d763`](https://github.com/facebook/hermes/releases/tag/hermes-2023-03-07-RNv0.71.4-31fdcf738940875c9bacf251e149006cf515d763) + https://github.com/facebook/hermes/commit/f2f1a2)

rather than fetching from the official fb tarball (old tag w/o crashfix commit)

leotm (dismissed):

Cal-L (dismissed):
LGTM

Gudahtt:
>This version bump happened when we had to regenerate yarn.lock

Hmm, why was this required? Regenerating the lockfile is a big change. If we could avoid this, we could reduce scope of this PR substantially and delay resolving the `hexToBN` issue described

app/components/UI/ReusableModal/index.tsx
```diff
@@ -1,257 +1,212 @@
-/* eslint-disable */
+/* eslint-disable react/prop-types */
+
+// Third party dependencies.
```
Gudahtt, line 3: Could you share additional details on how these changes relate to the react native upgrade?

jest.config.js
```diff
@@ -3,6 +3,7 @@ process.env.TZ = 'America/Toronto';
 const config = {
   preset: 'react-native',
   setupFilesAfterEnv: ['<rootDir>/app/util/test/testSetup.js'],
+  testEnvironment: 'jest-environment-node',
```
Gudahtt, line 6: Hmm, I'm curious to know why this change was needed. The `react-native` preset is all that was required before to setup the environment, and the docs suggest this isn't needed as well.

app/components/Views/Wallet/index.tsx
```diff
@@ -65,8 +65,7 @@ const createStyles = ({ colors, typography }: Theme) =>
       marginTop: 16,
     },
     textStyle: {
-      ...(typography.sBodyMD as TextStyle),
-      fontWeight: '500',
+      ...(typography.sHeadingSM as TextStyle),
```
Gudahtt, line 68: Could you share more context on the purpose of this change?

app/component-library/components/Sheet/SheetBottom/SheetBottom.tsx
```diff
@@ -88,7 +88,7 @@ const SheetBottom = forwardRef<SheetBottomRef, SheetBottomProps>(
     const onHidden = useCallback(() => {
       // Sheet is automatically unmounted from the navigation stack.
       navigation.goBack();
-      onDismissed?.();
+      onDismissed?.(!!postCallback.current);
```
Gudahtt, line 91: Was this change necessary? It seems that this `onDismissed` prop is only passed in in one place (`AccountConnnect.tsx`), and this parameter is not used there.

app/components/UI/ReusableModal/index.tsx
leotm: altho reanimated v2.14.0 we were on supports vanilla RN 0.71
remembering we ran into [issues](https://consensys.slack.com/archives/C02U025CVU4/p1679497625411209?thread_ts=1677607289.365609&cid=C02U025CVU4) fixed by bumping to v3 cc @jpcloureiro
notably added debugging support too (Chrome DevTools and Flipper)

leotm: so this code's upgrading our components from using reanimated v1 API (supported in v2) to v2 API (supported in v3) https://github.com/MetaMask/metamask-mobile/commit/f943b6df5a8d9e63604434911aaa533c4a5ea182 cc @Cal-L

app/components/Views/DetectedTokens/index.tsx
```diff
@@ -270,18 +269,15 @@ const DetectedTokens = () => {
   };
 
   return (
-    <ReusableModal
```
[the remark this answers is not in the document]
leotm: indeed https://github.com/MetaMask/metamask-mobile/pull/6220/commits/169ac93d9ab1754825ceb44f72401b221366ecf8 as follow-up to https://github.com/MetaMask/metamask-mobile/pull/6220#discussion_r1223250368 cc @Cal-L

ios/Podfile.lock
```diff
@@ -776,102 +823,105 @@ EXTERNAL SOURCES:
...
   sovran-react-native: fd3dc8f1a4b14acdc4ad25fc6b4ac4f52a2a2a15
   TcpSockets: a8eb6b5867fe643e6cfed5db2e4de62f4d1e8fd0
-  Yoga: b316a990bb6d115afa1b436b5626ac7c61717d17
+  Yoga: ba09b6b11e6139e3df8229238aa794205ca6a02a
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
-PODFILE CHECKSUM: 74e503a275b4ed097cb01c171027882e9cdc699d
+PODFILE CHECKSUM: 7ce98d5a1951b5810edb5cdb520381ad6ccbdf6c
 
-COCOAPODS: 1.12.0
```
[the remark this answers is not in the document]
leotm: good spot ^ updated/bumped in https://github.com/MetaMask/metamask-mobile/pull/6220/commits/cc031e3f971172ff5ab0b196f25e3d6d5790469a re-generating would be a good follow-up

app/components/Views/Wallet/index.tsx
leotm: good spot again, looks accidental https://github.com/MetaMask/metamask-mobile/commit/7251926bbff4f2ae45b33388cdd077758a1a0f03 cc @sethkfman
since unrelated to lint/audit blockers from earlier bitrise testing
reverted in https://github.com/MetaMask/metamask-mobile/pull/6220/commits/49dc21a10a0644f149af7cca739aa1611db585e9 linting still passing locally

ios/Podfile
```diff
@@ -1,7 +1,25 @@
...
+# If you are using a `react-native-flipper` your iOS build will fail when `NO_FLIPPER=1` is set.
+# because `react-native-flipper` depends on (FlipperKit,...) that will be excluded
+#
+# To fix this you can also exclude `react-native-flipper` using a `react-native.config.js`
+# ```js
+# module.exports = {
+#   dependencies: {
+#     ...(process.env.NO_FLIPPER ? { 'react-native-flipper': { platforms: { ios: null } } } : {}),
+# ```
+flipper_config = ENV['NO_FLIPPER'] == "1" ? FlipperConfiguration.disabled : FlipperConfiguration.enabled
```
leotm, line 16: just spotted this `flipper_config` isn't being used

leotm: pos related to Xcode 14.3.1 build error RE YogaKit

Screenshot 2023-06-14 at 14 36 55

ios/Podfile
```diff
@@ -38,17 +56,30 @@ def common_target_logic
...
+    :hermes_enabled => false,
+    :fabric_enabled => flags[:fabric_enabled],
+    # Enables Flipper.
+    #
+    # Note that if you have use_frameworks! enabled, Flipper will not work and
+    # you should disable the next line.
+    #:flipper_configuration => flipper_config,
+    # An absolute path to your application root.
+    :app_path => "#{Pod::Config.instance.installation_root}/..",
+    :flipper_configuration => FlipperConfiguration.enabled(["Debug"], {"Flipper" => "0.93.0", "Flipper-DoubleConversion" => "1.1.7"}),
```
leotm, line 78:
```suggestion
    :flipper_configuration => flipper_config,
```

leotm: ref: https://github.com/facebook/react-native/blob/v0.71.6/template/ios/Podfile#L41

leotm: this change led to a build error due to our custom Podfile, fixed/resolved in: https://github.com/MetaMask/metamask-mobile/pull/6220/commits/0a733513c4188b7595e5b0b16420ce9368791dcf

ios/Podfile
leotm: both projects already appear configured for Swift 5 :suspect: 

Screenshot 2023-06-14 at 16 24 31

leotm: project format updated from `12.0-compat` to `14.0-compat` in https://github.com/MetaMask/metamask-mobile/pull/6220/commits/c8313a47f099afc2513a7e384d56ba2eca38941d

Screenshot 2023-06-14 at 16 31 44

but YogaKit build error persists :suspect:

leotm: small update: spotted also happening to (1.5k) others past couple months https://developer.apple.com/forums/thread/728161 continuing to monitor for a solution

gutek2 (approved):

ios/Podfile
leotm: small update: tested on [Xcode 15.0 Beta](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes) too

again required mixpanel and branch bumps from `IPHONEOS_DEPLOYMENT_TARGET = 8.0` to `12.0` (`11.0` in Xcode 14)

Screenshot 2023-06-21 at 21 21 01

_nb: verify `git submodule update --init` not error'ing before commiting these changes, for `yarn setup`_

then required fixing below issue via workaround
- https://github.com/facebook/react-native/issues/37748
  - [landing officially in RN 0.71.11](https://github.com/reactwg/react-native-releases/discussions/70#discussioncomment-6120063)
  - _nb: also fixable via Podfile_

Screenshot 2023-06-21 at 20 52 49

but `YogaKit` not found still persists, despite bumping `YogaKit` pod `Swift Compiler - Language` from 4 to 5

Screenshot 2023-06-21 at 21 27 43

agreed with @Cal-L Xcode 14.3+ support better handled outside this branch as follow-up
continue to monitor https://developer.apple.com/forums/thread/728161 mentioned above

jpcloureiro:
Our unit tests are failing on the components `AdvancedSettings` & `EthSignFriction`

## EthSighFriction
Screenshot 2023-06-26 at 16 00 15
Screenshot 2023-06-26 at 16 00 41


It can't find the element with the role "button"

## AdvancedSettings
Screenshot 2023-06-26 at 16 01 06

We seem to have issues with the `navigation` mock.

There are also two snapshots that need to be updated.

cortisiko (dismissed):
This PR is officially QA passed.

android/build.gradle
```diff
@@ -3,68 +3,26 @@
...
-		buildToolsVersion = "30.0.3"
+		buildToolsVersion = "33.0.0"
 		minSdkVersion = project.hasProperty('minSdkVersion') ? project.getProperty('minSdkVersion') : 23
-        compileSdkVersion = 31
-         targetSdkVersion = 31
-		kotlin_version = "1.5.31"
+		compileSdkVersion = 33
+		targetSdkVersion = 33
+		// We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
+		ndkVersion = "24.0.8215888"
```
leotm, line 11:
```suggestion
		ndkVersion = "23.1.7779620"
```

ref: https://github.com/MetaMask/metamask-mobile/pull/6384/files#diff-197b190e4a3512994d2cebed8aff5479ff88e136b8cc7a4b148ec9c3945bd65aL11-R11 cc @cortisiko

app/component-library/components/Sheet/SheetBottom/SheetBottom.tsx
jpcloureiro: It is being used on `DetectTokens` View, where the `onDismissed` prop is being passed in with an arrow function called `trackCancelWithoutAction` that has 1 boolean parameter.

jpcloureiro:
> > This version bump happened when we had to regenerate yarn.lock
> 
> Hmm, why was this required? Regenerating the lockfile is a big change. If we could avoid this, we could reduce scope of this PR substantially and delay resolving the `hexToBN` issue described

Hey @Gudahtt we've revised `yarn.lock` to keep its changes to a minimum by running `yarn` on top of a `yarn.lock` file from main.

We also removed the `hexToBN` auxiliary function, leaving it as it was before.

Thanks for pointing that out!

jest.config.js
jpcloureiro: For some reason it does not assume the default environment.

If we remove that property, jest complains about a missing config.

android/build.gradle
jpcloureiro: Agreed with @leotm leave this change in https://github.com/MetaMask/metamask-mobile/pull/6384 

We should pay attention to performance degradation, it might be related to using an unsupported ndk version.

ios/Podfile
jpcloureiro: This is now being tracked on this separate issue: https://github.com/MetaMask/metamask-mobile/issues/6158

leotm (approved):
great pairing and reviewing this pr today, lgtm :shipit:

PR #6246, opened 2023-04-21
Fix for wrong to account information on confirmation page

jpuri:
Fixes issue with to account details not correctly displayed on confirmation page.


segun (approved):

digiwand (approved):

seaona:
@jpuri the fix looks good and it's working. I can see how the recipient displays the correct address. We can merge the PR

Screenshot from 2023-04-26 10-35-00

Issue #917, opened 2019-07-24
URI format for clicking links to make payments


jjs2484:
Hi,

On android the following URI format successfully opens the wallet and fills out the address field but fails to fill out the amount field... Is this something that will be fixed in the future? Or is there another URI format being utilized?

`<a href="ethereum:0x4A7f51F4c9AD183bdfD1A43cd6CeF4ef8c73788c?amount=0.00720891">Pay ETH</a>`

brunobar79:
Hey there! We’re supporting the standard ERC-681 ( https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md) where the value needs to be expressed in wei.

Here’s the relevant part:

```

The amount is specified in value parameter, in the atomic unit (i.e. Wei). The use of scientific notation is strongly encouraged. For example, requesting 2.014 ETH to address 0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359 would look as follows: ethereum:0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359?value=2.014e18

```

By looking at your example you’re missing `e18’ at the end of the value

PR #6258, opened 2023-04-24
[FIX] Prevent adding non supported tokens on networks

Cal-L:
**Development & PR Process**

**Description**

This PR addresses [#917](https://github.com/MetaMask/mobile-planning/issues/917). The correct behavior is that you should not be able to add a non-supported token while on certain networks through the watch asset flow (clicking on the MM icon from Dapps). Non-supported here means that the token address does not exist on the network that you are currently on. Follow steps listed in original issue to test for fix.

**Screenshots/Recordings**

Video of fix in thread - https://consensys.slack.com/archives/C029JG63136/p1682348127958249?thread_ts=1682111629.402559&cid=C029JG63136

**Issue**
[#917](https://github.com/MetaMask/mobile-planning/issues/917)

Progresses #???

**Checklist**

### DIFF
--- a/app/core/RPCMethods/RPCMethodMiddleware.ts
+++ b/app/core/RPCMethods/RPCMethodMiddleware.ts
@@ -20,6 +20,8 @@ import setOnboardingWizardStep from '../../actions/wizard';
 import { v1 as random } from 'uuid';
 import { getPermittedAccounts } from '../Permissions';
 import AppConstants from '../AppConstants.js';
+import { isSmartContractAddress } from '../../util/transactions';
+import { TOKEN_NOT_SUPPORTED_FOR_NETWORK } from '../../constants/error';
 const Engine = ImportedEngine as any;
 
 let appVersion = '';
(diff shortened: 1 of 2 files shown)


app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -617,10 +619,19 @@ export const getRpcMethodMiddleware = ({
             type,
           },
         } = req;
-        const { TokensController } = Engine.context;
+        const { TokensController, NetworkController } = Engine.context;
+        const { chainId } = NetworkController.state?.providerConfig || {};
```
tommasini, line 623: Why do we fall back to an empty object?

A good solution to the problem!

LGTM

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -617,10 +619,19 @@ export const getRpcMethodMiddleware = ({
...
 
         checkTabActive();
         try {
+          // Check if token exists on wallet's active network.
+          const isTokenOnNetwork = await isSmartContractAddress(
```
tommasini, line 628: Is there any chance of being a smart contract and not a token?

tommasini (approved):
LGTM

app/core/RPCMethods/RPCMethodMiddleware.ts
Cal-L: Just making sure it doesn't crash when `chainId` is undefined. We could also do something like
`const chainId = NetworkController.state?.providerConfig?.chainId`, which would be the same thing

app/core/RPCMethods/RPCMethodMiddleware.ts
Cal-L: Yep that is possible. For now, we are just reusing the same check logic as adding a custom token. Since this check is only meant to see if a contract exists on the current network, we can track checking for token standards in another ticket.

Issue #5760, opened 2023-02-14
[Bug] Gas Fee modal for Legacy tx is displaying the EIP1559 fee modal


seaona:
**Describe the bug**
Gas Fee modal for Legacy tx is displaying the EIP1159 fee modal

**Screenshots**

https://user-images.githubusercontent.com/54408225/218763488-000a2c00-a600-41c0-9b8b-b29a3e094eb8.mp4

**To Reproduce**
_Steps to reproduce the behavior_
1. Go to the test-dapp https://metamask.github.io/test-dapp/
2. Connect MM
3. Click Send Legacy Transaction
4. Click Fees
5. Click Edit Fees
6. See the gas modal for EIP1159 tx

**Expected behavior**
_A clear and concise description of what you expected to happen_
On legacy transactions we don't have priority fees. We should display the Gas limit and the Gas price. Take Extension as a reference:

![image](https://user-images.githubusercontent.com/54408225/218763903-dc58ea38-a385-4494-8d1e-629469d60ebe.png)

**Smartphone (please complete the following information):**
Device: Samsung SM-A217F
OS: Android 11.0
App Version: v5.14.0 (1055)

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

DigitalsupportTeam:
We'd be happy to look into this for you
kindly file a support ticket below 
https://forms.gle/Wmpm6NLzYm9uXYuU9

seaona:
I've added this for the next Sprint if possible, as it is an inconvenience for adding tests as I cannot rely on the correct gas info displayed

bschorchit:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/-Confirmations-Secure-UX-6245e6e2348677001213b8d2/issues/MetaMask/metamask-mobile/5760?planning-poker) with Zenhub @blackdevelopa @jpuri @seaona @segun

bschorchit:
Q: Do we have legacy modals on mobile or will this require design too? Not clear if we do have legacy modal

SayaGT:
Hello, you can find the design here
https://www.figma.com/file/aHdSNvg9ahXZMtThzz3bUo/Bugs?type=design&node-id=1%3A13803&mode=design&t=tobOzS87YOReqDgm-1

PR #6287, opened 2023-04-27
chore: Custom Gas Modal Component

blackdevelopa:
**Development & PR Process**

**Description**
This PR will introduce a custom gas modal component that should handle the different transaction type for editing gas fees. It's a code improvement PR and so there should be no visible changes to the app UX or UI. This PR will cover the confirm screen for send flow.

Test Case:

**Screenshots/Recordings**
http://recordit.co/1yUAJhqhOy

E2E: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/4b890180-5a54-4361-bc1d-955a03d41f5d

**Issue**

Progresses #???

**Checklist**


bschorchit:
Hey @gauthierpetetin, did you remove the `needs-dev-review` and added a `in-progress` for this PR by mistake? We currently use the `needs-dev-review` label for internal processes in the team. cc: @hilvmason

gauthierpetetin:
Hi @bschorchit , we setup an automation yesterday to update the label of the Github issue every time the corresponding Zenhub issue is moved to a new column (more detail [here](https://github.com/MetaMask/metamask-zaps/tree/main/zaps/github-labels-automation)). In this case the ticket was moved from "needs-dev-review" to "in-progress" column.

However, I don't understand why it impacts your board, because the automation is only meant to work for the MetaMask Mobile board. I'll investigate and come back to you, sorry for that.

gauthierpetetin:
@bschorchit we've just added an additional filter to make sure the automation only applies to issues with the label "team-mobile-client". It shall not happen again on your board. In case it does, don't hesitate to reach out.

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,184 @@
...
+  isAnimating,
+  onlyGas,
+  validateAmount,
+  updateParent,
+  legacy,
+  legacyGasData,
+  EIP1559GasData,
+  EIP1559GasTxn,
+}: CustomGasModalProps) => {
+  const { colors } = useAppThemeFromContext() || mockTheme;
```
jpuri, line 25: `useAppThemeFromContext() || mockTheme` does not looks good. We should rather change `useAppThemeFromContext` to return mockTheme by default or
`const { colors = mockTheme.colors } = useAppThemeFromContext();`

app/components/UI/CustomGasModal/CustomGasModal.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+            totalMaxHex: '',
+            error: undefined,
+          }}
+          selectedAsset={{ address: '0xABC', symbol: 'ABC', decimals: 0 }}
+        />
+      </Provider>,
+    );
+    expect(wrapper).toMatchSnapshot();
+  });
+});
```
jpuri, line 92: Above is shallow snapshot matching, is not very useful.

app/components/Views/SendFlow/Confirm/index.js
```diff
@@ -1263,9 +1061,7 @@ class Confirm extends PureComponent {
             )}
           </StyledButton>
         </View>
-        {mode === EDIT && this.renderCustomGasModalLegacy()}
         {mode === EDIT_NONCE && this.renderCustomNonceModal()}
-        {mode === EDIT_EIP1559 && this.renderCustomGasModalEIP1559()}
         {this.renderHexDataModal()}
       </SafeAreaView>
     );
```
jpuri, line 1067: Great cleanup 👏

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,184 @@
...
+      setEIP1559GasObj(gasObj);
+      updateParent({
+        EIP1559GasTransaction: gasTxn,
+        EIP1559GasObject: gasObj,
+        gasSelectedTemp: selectedGas,
+        gasSelected: selectedGas,
+        closeModal: true,
+      });
+    }
+  };
```
jpuri, line 95: Above function can also be broken down into 2 functions for legacy and 1559 transactions.

jpuri:
@blackdevelopa : while testing with legacy transaction from test DAPP I found that it is showing advance options for legacy transaction:

Uploading Screen Recording 2023-05-01 at 2.28.30 PM.mov…

blackdevelopa:
@jpuri thanks for flagging this. It exists in production and a new issue should be created for it.

> @blackdevelopa : while testing with legacy transaction from test DAPP I found that it is showing advance options for legacy transaction:
> 
>  Screen.Recording.2023-05-01.at.2.28.30.PM.mov

app/components/UI/CustomGasModal/CustomGasModal.test.tsx
blackdevelopa: Added some [here](https://github.com/MetaMask/metamask-mobile/pull/6287/commits/90da478ffa41abb258184d2b450bdd2a24e60b0d)

app/components/UI/CustomGasModal/CustomGasModal.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+const legacyGasData = {
+  legacyGasLimit: '',
+  suggestedGasPrice: '',
+};
+
+describe('CustomGasModal', () => {
+  it('should render correctly', () => {
+    const wrapper = render(
```
segun, line 103: Consider using `renderWithProvider` from `app/util/test/renderWithProvider.tsx`

segun (commented):
Just one comment.

app/components/UI/CustomGasModal/CustomGasModal.test.tsx
blackdevelopa: Updated. Thank you. https://github.com/MetaMask/metamask-mobile/pull/6287/commits/84c225da3f201dc770b5a03258d498b249b3087e

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+      setEIP1559GasObj(gasObj);
+      updateParent({
+        EIP1559GasTransaction: gasTxn,
+        EIP1559GasObject: gasObj,
+        gasSelectedTemp: selectedGas,
+        gasSelected: selectedGas,
+        closeModal: true,
+      });
+    };
+
```
jpuri, line 159: Nested components declaration is not good idea.

jpuri (changes requested):
Hey, can you plz get rid of nested component declaration. PR looks good otherwise.

segun (dismissed):

segun (dismissed):

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+  const getGasAnalyticsParams = () => {
+    try {
+      return {
+        active_currency: { value: selectedAsset.symbol, anonymous: true },
+        gas_estimate_type: gasEstimateType,
+      };
+    } catch (error) {
+      return {};
+    }
+  };
```
jpuri, line 58: Above function can just be constant, its returning static values and really does not needs try/catch block.

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+    updateParent({ gasSelected: gas });
+  };
+
+  const onCancelGas = () => {
+    updateParent({
+      stopUpdateGas: false,
+      gasSelectedTemp: selectedGas,
+      closeModal: true,
+    });
+  };
```
jpuri, line 71: Variable names `updateParent`, `stopUpdateGas` are not very intuitive.

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+          legacyGasObject: gasObj,
+          gasSelected: gasSelect,
+          closeModal: true,
+          stopUpdateGas: false,
+          advancedGasInserted: !gasSelect,
+          gasSelectedTemp: gasSelect,
+        });
+      },
+    [transaction, validateAmount, updateParent],
+  );
```
jpuri, line 100: is it possible that we keep `closeModal` in this component itself and not in parent ?
Null can be returned if `closeModal` is true.

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+            onCancel={onCancelGas}
+            onSave={onSaveEIP1559GasOption}
+            animateOnChange={animateOnChange}
+            isAnimating={isAnimating}
+            analyticsParams={getGasAnalyticsParams()}
+            view={'SendTo (Confirm)'}
+            selectedGasObject={eip1559GasObject}
+            onlyGas={onlyGas}
+          />
+        )}
```
jpuri, line 198: We can avoid passing here values which are fetched from selector. Those can be obtained in child components.

blackdevelopa: Yeah. Just checked again and don't think there's any value here that can be gotten from selector that is passed.

app/components/UI/CustomGasModal/CustomGasModal.tsx
blackdevelopa: Yeah. I updated `updateParent` to `updateParentState`. Let me know if this works. I am not sure what a better name for `stopUpdateGas` should be

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+        });
+
+        setEIP1559Txn(gasTxn);
+        setEIP1559GasObj(gasObj);
+        updateParentState({
+          EIP1559GasTransaction: gasTxn,
+          EIP1559GasObject: gasObj,
+          gasSelectedTemp: selectedGas,
+          gasSelected: selectedGas,
+          closeModal: true,
```
jpuri, line 118: @blackdevelopa : parent state params may be different for different components. How will that get handled ?

app/components/UI/CustomGasModal/CustomGasModal.tsx
```diff
@@ -0,0 +1,198 @@
...
+          gasSelected: gasSelect,
+          closeModal: true,
+          stopUpdateGas: false,
+          advancedGasInserted: !gasSelect,
+          gasSelectedTemp: gasSelect,
+        });
+      },
+    [transaction, validateAmount, updateParentState],
+  );
+
```
jpuri, line 95: I am bit confused, below looks like callback of callback.

jpuri (commented):
Changes look good, just some small questions.

app/components/UI/CustomGasModal/CustomGasModal.tsx
blackdevelopa: Done. [345c539](https://github.com/MetaMask/metamask-mobile/pull/6287/commits/345c539f781c5139988e8e40405e2e137b226071)

app/components/UI/CustomGasModal/CustomGasModal.tsx
blackdevelopa: Updated. [fd5e14d](https://github.com/MetaMask/metamask-mobile/pull/6287/commits/fd5e14dbc3fe1f289c142bad77619e5a25e39273)

jpuri (dismissed):
Changes look good 👍

segun (dismissed):

seaona:
do you think we could prioritize [this issue](https://github.com/MetaMask/metamask-mobile/issues/5760) soon? It is interfering whenever we test anything related with gas.

cc. @bschorchit @blackdevelopa

seaona:
From QA everything looks good except a small issue I am seeing with Send ETH using a Deeplink.
Basically I can see how the gas does not have a correct value, after re-calculated, making the transaction fail.
Notice: this only happens with Send ETH and not with Send ERC20 token using a deeplink.


https://github.com/MetaMask/metamask-mobile/assets/54408225/68bcf786-1e13-43c0-91a6-dd710a9bbe57



I've tried on `main` and apparently it's working fine. Do you think something from this PR might have affected that function @blackdevelopa ?

1. Select Sepolia
2. Scan this QR code
3. See how gas is set to >0.00000....1
4. Try to Send the tx
5. Tx fails with `Intrinsic gas too low` error

image

seaona:
hi @blackdevelopa ! The issue above seems intermittent and probably not caused by this PR, as we have other reports related to Deeplink gas calculations. For this reason it can be investigated separately.

However I encounter another issue described below.

### Wallet breaks when changing gas priority
Whenever I am sending an ERC20 token, if I edit the gas priority the wallet breaks with the error `Cannot read property 'subst' of undefined`


https://github.com/MetaMask/metamask-mobile/assets/54408225/9f67fad5-4226-4cf3-be39-492aa1e543d8

**Steps**:
1. Import an ERC20 token
2. Click Send
3. Add any amount and recipient
4. Click next
5. Click Edit Gas
6. Change priority Gas
7. Click Save
8. Wallet breaks

seaona:
I can see the issue fixed @blackdevelopa . This looks good from QA. We need x2 re-approvals

jpuri (dismissed):

app/components/UI/CustomGasModal/CustomGasModal.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+jest.mock('react-native-keyboard-aware-scroll-view', () => {
+  const KeyboardAwareScrollView = jest.requireActual('react-native').ScrollView;
+  return { KeyboardAwareScrollView };
+});
+
+const gasSelected = 'high';
+const dummyAddress = '0x0';
+
+const initialState = {
```
cryptotavares, line 25: There is a [common background initial state](https://github.com/MetaMask/metamask-mobile/blob/main/app/util/test/initial-background-state.json) for unit tests. If possible we should use that one, as it reduces the unit tests maintenance effort.

blackdevelopa: Great. Updated [here](https://github.com/MetaMask/metamask-mobile/pull/6287/commits/1217799a6b685ffbc659fd9b4f936c03563c39b4)

jpuri (approved):

segun (approved):

blackdevelopa:
Merging this PR as it is now QA passed

PR #6291, opened 2023-04-27
refactor: trigger transaction modals using approval requests

vinistevam:
**Development & PR Process**

**Description**

This PR applies a change to TransactionController that has already been merged in the core repo. Unfortunately, there are breaking changes blocking the ability to upgrade the package directly.

The changes from the core [PR](https://github.com/MetaMask/core/pull/1241) for visibility. It implies that confirmations for transaction requests will no longer be handled by an event, but sent directly to ApprovalController. This is part of the work to standardise and simplify how confirmations work across extension and mobile.
1. Unifying the trigger modal approach
2. Included in the transaction-controller (core) the capability to create an ApprovalRequest whenever the unapprovedTransaction is fired and leverage it to control the transaction modals in mobile.

**Screenshots/Recordings**
[Transaction tests](https://github.com/MetaMask/metamask-mobile/assets/45455812/30ccf6d3-496b-40ae-9895-06e20f9fdbf5)

**Issue**
resolves: https://github.com/MetaMask/mobile-planning/issues/770

**Checklist**


app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -446,19 +439,35 @@ const RootRPCMethodsUI = (props) => {
...
   };
 
-  const renderDappTransactionModal = () =>
-    props.dappTransactionModalVisible && (
-      <Approval
-        navigation={props.navigation}
-        dappTransactionModalVisible
-        toggleDappTransactionModal={props.toggleDappTransactionModal}
-      />
+  const isTransactionApprovalVisible = () =>
```
bergarces, line 442: I think this name can be a bit confusing, as it's not telling us if it's visible or not, but if we want to render it or not.

It's one thing to have the `isVisible` property in a modal, but here I think something such as `renderTransactionApproval` or `renderTransactionApprovalModal` would be better.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -446,19 +439,35 @@ const RootRPCMethodsUI = (props) => {
...
-  const renderApproveModal = () =>
-    props.approveModalVisible && (
-      <Approve modalVisible toggleApproveModal={props.toggleApproveModal} />
+  const renderApproveModal = () => {
+    const transactionApprovalVisible = isTransactionApprovalVisible();
+    return (
+      transactionApprovalVisible &&
+      transactionType === 'transaction' && (
+        <Approve
+          modalVisible={transactionApprovalVisible}
```
bergarces, line 465: Don't need to assign `transactionApprovalVisible`, as it wouldn't have reached this far if it was false.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -446,19 +439,35 @@ const RootRPCMethodsUI = (props) => {
...
+    showPendingApproval?.type === ApprovalTypes.TRANSACTION;
+
+  const renderDappTransactionModal = () => {
+    const transactionApprovalVisible = isTransactionApprovalVisible();
+    return (
+      transactionApprovalVisible &&
+      transactionType === 'dapp' && (
+        <Approval
+          navigation={props.navigation}
+          dappTransactionModalVisible={transactionApprovalVisible}
```
bergarces, line 452: Don't need to assign transactionApprovalVisible, as it wouldn't have reached this far if it was false.

app/components/Views/Approval/index.js
```diff
@@ -81,9 +81,9 @@ class Approval extends PureComponent {
      */
     networkType: PropTypes.string,
     /**
-     * Hides or shows the dApp transaction modal
+     * Set approval dApp transaction modal visible or not
      */
-    toggleDappTransactionModal: PropTypes.func,
+    setShowTransactionApproval: PropTypes.func,
```
bergarces, line 86: I would avoid changing the original name here (`setShowPendingApproval`), it makes it more difficult to find references.

app/components/Nav/Main/RootRPCMethodsUI.js
vinistevam: I loved `showTransactionApproval`, pushing it now.

app/components/Nav/Main/RootRPCMethodsUI.js
vinistevam: I was assigning it here because, in the QR signing PR, we can get rid of those props (`modalVisible`, `dappTransactionModalVisible`). Currently, those props are also being considered to not display the QR modal.
Though I could explore if I can remove it now in this PR.

app/components/Nav/Main/RootRPCMethodsUI.js
vinistevam: Same as above, I will double-check the implication of removing those props.

app/components/Views/Approval/index.js
vinistevam: Pushed the change, thanks!

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -446,19 +439,35 @@ const RootRPCMethodsUI = (props) => {
...
-        toggleDappTransactionModal={props.toggleDappTransactionModal}
-      />
+  const showTransactionApproval = () =>
+    showPendingApproval?.type === ApprovalTypes.TRANSACTION;
+
+  const renderDappTransactionModal = () => {
+    const transactionApprovalVisible = showTransactionApproval();
+    return (
+      transactionApprovalVisible &&
+      transactionType === 'dapp' && (
```
matthewwalsh0, line 449: As we're using these strings as constants, can we put them in a local enum like `TransactionModalType`?

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -36,10 +36,6 @@ import AccountApproval from '../../UI/AccountApproval';
 import TransactionTypes from '../../../core/TransactionTypes';
 import AddCustomNetwork from '../../UI/AddCustomNetwork';
 import SwitchCustomNetwork from '../../UI/SwitchCustomNetwork';
-import {
-  toggleDappTransactionModal,
```
matthewwalsh0, line 40: Are these the only usages? If so can we delete the definitions also.

app/components/Views/Approval/index.js
```diff
@@ -81,9 +81,9 @@ class Approval extends PureComponent {
      */
     networkType: PropTypes.string,
     /**
-     * Hides or shows the dApp transaction modal
+     * Set approval dApp transaction modal visible or not
      */
-    toggleDappTransactionModal: PropTypes.func,
+    setShowPendingApproval: PropTypes.func,
```
matthewwalsh0, line 86: Rather than coupling these to pending approvals, can we simplify them even more and just call them `hideModal` and drop the argument? I can't seem to find any instances where we set it to `true`.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -361,9 +356,9 @@ const RootRPCMethodsUI = (props) => {
           data.substr(0, 10) === APPROVE_FUNCTION_SIGNATURE &&
           (!value || isZeroValue(value))
         ) {
-          toggleApproveModal();
+          setTransactionType('transaction');
```
matthewwalsh0, line 359: This logic is inside `onUnapprovedTransaction` which is still using the event directly from the `TransactionController`.

Ideally, we'd trigger all this logic solely from the pending approval in order to decouple it from the controller and background.

We could get the same `transactionMeta` by asking the controller for it, or perhaps cleaner if we include it all in the approval request.

There's also a good argument to move all this transaction logic and modal rendering into a separate componentbut I appreciate this is already a huge improvement though so maybe cleaner to do this move in a separate ticket? What do you think?

vinistevam: Let me check if I'm following along, remove the listeners and trigger this logic in the `handlePendingApprovals` whenever we receive a `transaction`, get the `transactionMeta` from the controller and perform the logic.
I really sympathize with the idea of moving all to a separate component, I will add a draft ticket so we can refine it during DSU.

app/components/Views/Approval/index.js
vinistevam: Renaming to `hideModal` and removing the argument, as you pointed out there are no instances where sets it to `true`

app/components/Nav/Main/RootRPCMethodsUI.js
vinistevam: `toggleDappTransactionModal` is also used [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/Send/index.js#L217), but `toggleApproveModal` can be removed.

app/components/Nav/Main/RootRPCMethodsUI.js
vinistevam: Yes, perfect. I'm pushing it now.

app/components/Views/ApproveView/Approve/index.js
```diff
@@ -93,9 +93,9 @@ class Approve extends PureComponent {
      */
     modalVisible: PropTypes.bool,
     /**
-    /* Token approve modal visible or not
+    /* Set approve modal visible or not
     */
-    toggleApproveModal: PropTypes.func,
+    setShowPendingApproval: PropTypes.func,
```
matthewwalsh0, line 98: Can we do the same here and rename to `hideModal` without the argument?

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -731,6 +732,13 @@ const RootRPCMethodsUI = (props) => {
             origin: request.origin,
           });
           break;
+        case ApprovalTypes.TRANSACTION:
+          await onUnapprovedTransaction(requestData.txId);
```
matthewwalsh0, line 736: If this is all working, that was much simpler than I feared!

We'll have to be very thorough with testing and Bitrise etc.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -276,7 +276,11 @@ const RootRPCMethodsUI = (props) => {
   );
 
   const onUnapprovedTransaction = useCallback(
-    async (transactionMeta) => {
+    async (transactionMetaId) => {
```
matthewwalsh0, line 279: `transactionId`?

matthewwalsh0, line 279: My only concern of doing this in the same ticket is we may have to be more targeted about what is done in response to the approval request, and what should really be a backend flow still using the event.

As an example, the `autoSign` function and usage below is worrying me as that very much feels like backend logic, that should be done during the `Engine` initialisation for example, as it's doing fundamental non-UI related things like automatically signing swaps transactions.

If it's easy to revert, it may be a lot safer to merge this with the event listener still in place and then we can split this in the other ticket, move the swaps signing to Engine, still in an transaction controller event listener, but keep all the state setting logic here as a response to the approval request.

app/components/Views/ApproveView/Approve/index.js
vinistevam: I just renamed and pushed the changes.

app/components/Nav/Main/RootRPCMethodsUI.js
vinistevam: Sorry for the misunderstanding. I reverted the event listener and now it can be handled by the ticket https://github.com/MetaMask/mobile-planning/issues/1006.

matthewwalsh0 (approved):

bergarces (approved):

PR #6311, opened 2023-05-02
Fix token balance displayed in approval pages

### LINKED ISSUE #5955: [Bug] ERC20 Approve is displaying ETH balance instead of token balance
**Describe the bug**
The ERC20 Approve screen is displaying ETH balance instead of token balance.

**Screenshots**

image

**To Reproduce**
1. Go to https://metamask.github.io/test-dapp/
2. Connect MM
3. Click Create Token and Accept MM prompt
4. Click Approve Tokens -- see we are displaying ETH balance instead of the token balance

**Expected behavior**
In the ERC20 Approve screen, we should display the token balance, in the same way we do on Extension. Example

image

- Device: Pixel 6
- OS: Android 12
- App Version: v6.0.1 (1078

**Severity**

- Is this tech debt?

jpuri:
Fixes: #5955

For ERC20 Approve confirmation show right balance (of token being approved) on confirmation page. The PR fixes balance displayed in different scenario like Approve ERC20, transfer ERC20, etc

QA Steps:
1. ERC20 approve show correct token balance of from address
2. Transfer ERC20 should also show correct balance
For confirmation except collectibles the header should show balance of token being transacted.


blackdevelopa (dismissed):

app/components/hooks/useAddressBalance/useAddressBalance.ts
```diff
@@ -0,0 +1,113 @@
...
+            const { AssetsContractController } = Engine.context;
+            fromAccBalance = await AssetsContractController.getERC20BalanceOf(
+              contractAddress,
+              address,
+            );
+            fromAccBalance = `${renderFromTokenMinimalUnit(
+              fromAccBalance || '0',
+              decimals,
+            )} ${symbol}`;
+            setAddressBalance(fromAccBalance);
```
digiwand, line 102: [suggestion] I think we can add a return statement in the catch block below and call `setAddressBalance(fromAccBalance);` one time at the bottom of this useEffect hook

digiwand (dismissed):
no blockers, only 1 minor suggestion above. Nice cleanup!

app/components/hooks/useAddressBalance/useAddressBalance.ts
jpuri: Hey, this actually looks better to me.

jpuri:
@seaona : the issue does not seems related to code changes in this branch. I spent time debugging this yesterday and could even replicate in `main` branch after few builds.

I tried again today updating the branch with latest from `main` and issue seems to be fixed.

seaona:
After testing again, I can see how the above issue is solved. I've found the following issues:

## Transfer ERC20 when token is not in the wallet
When the token is not added to the wallet we see the following issues:

- Value passed from a dapp is `1.5` but now I see `6.462`
- Token Balance is displayed as `0`


https://github.com/MetaMask/metamask-mobile/assets/54408225/d1b5859e-2703-4c02-adb5-ad9405866f41

**Expected**: the same as we see when token is added to the wallet? Is this possible in Mobile?

image


## Approve ERC20
Whenever the token is not added to the Wallet, we can see how the token symbol is not displayed.


image


**Expected**: the same as we see when token is added to the wallet? Is this possible in Mobile? On Extension we see the token symbol even if the token is not added to the wallet cc @bschorchit . See

image

seaona:
@jpuri PR looks good!! It can be merged after fixing conflicts with test file. Thank you v much, great work :100:

blackdevelopa (dismissed):

blackdevelopa (approved):

PR #6354, opened 2023-05-09
Trigger signing modals from approval requests

matthewwalsh0:
**Description**

Trigger the signature request modals using the generic `ApprovalController` requests rather than events specific to the `SignatureController`.

**Checklist**


app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -695,6 +709,27 @@ const RootRPCMethodsUI = (props) => {
...
+            origin: request.origin,
+          });
+          break;
+        case ApprovalTypes.ETH_SIGN_TYPED_DATA:
+          setSignMessageParams(requestData);
+          showPendingApprovalModal({
+            type: ApprovalTypes.ETH_SIGN_TYPED_DATA,
+            origin: request.origin,
+          });
+          break;
```
bergarces, line 732:
```suggestion
        case ApprovalTypes.ETH_SIGN:
        case ApprovalTypes.PERSONAL_SIGN:
        case ApprovalTypes.ETH_SIGN_TYPED_DATA:
          setSignMessageParams(requestData);
          showPendingApprovalModal({
            type: request.type,
            origin: request.origin,
          });
          break;
```

matthewwalsh0: Done.

bergarces (dismissed):

ios/Podfile.lock
```diff
@@ -874,4 +874,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 74e503a275b4ed097cb01c171027882e9cdc699d
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.12.1
```
OGPoyraz, line 877: I guess this was not intentional

matthewwalsh0: A previous commit bumped it so I updated my gem to match, but then another commit set it back to theirs so not sure which was intentional 😄 

Will remove mine and leave as is.

vinistevam (approved):

PR #6355, opened 2023-05-09
[FEAT] wallet connect v2

abretonc7s:
**Description**

- WalletConnect V2 Integration.
- Compatibility with walletconnect v1.

WALLET_CONNECT_PROJECT_ID needs to be set in `js.env`


app/core/WalletConnect.js
sethkfman, line 1: I would move this file into the WalletConnect directory.

abretonc7s: updated.

andreahaku (approved):
LGTM

PR #6202, opened 2023-04-17
Adopt ApprovalType from core

vinistevam:
**Development & PR Process**
1. Follow MetaMask [Mobile Coding Standards](https://docs.google.com/document/d/1VJLwTRsUw_5EDq_o8d6sSbXUAYENLurkRitYO45eY5o/edit?usp=sharing)
2. Add `release-xx` label to identify the PR slated for a upcoming release (will be used in release discussion)
3. Add `needs-dev-review` label when work is completed
4. Add `needs-qa` label when dev review is completed
5. Add `QA Passed` label when QA has signed off

**Description**

Adopt `ApprovalType` to approval requests/confirmations leaving room for a more cohesive usage between extension and mobile.
_1. Align approval requests and confirmations between mobile and extension.

**Screenshots/Recordings**

_If applicable, add screenshots and/or recordings to visualize the before and after of your change_

**Issue**

resolves: https://github.com/MetaMask/mobile-planning/issues/795

**Checklist**

* [x] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

PR #6357, opened 2023-05-09
refactor: use approval controller for watch asset confirmation

bergarces:
**Development & PR Process**

**Description**

This PR applies a change to `TokensController` that has already been merged in the core repo. Unfortunately, there are breaking changes blocking the ability to upgrade the package directly.

The change implies that confirmations for `wallet_watchAsset` requests will no longer be handled by an event, but sent directly to `ApprovalController`. This is part of the work to standardise and simplify how confirmations work across extension and mobile.

**Screenshots/Recordings**

[watchAsset.webm](https://github.com/MetaMask/metamask-mobile/assets/1970725/fa5f1095-c7d1-4c08-88d5-6962f54e2d78)

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/767

**Checklist**


app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -33,6 +33,7 @@ export enum ApprovalTypes {
   SWITCH_ETHEREUM_CHAIN = 'SWITCH_ETHEREUM_CHAIN',
   REQUEST_PERMISSIONS = 'wallet_requestPermissions',
   WALLET_CONNECT = 'WALLET_CONNECT',
+  WATCH_ASSET = 'wallet_watchAsset',
```
bergarces, line 36: We should create a PR to replace all these approval types with the enum from `@metamask/controller-utils`. Not all of them have matching names.

app/components/Nav/Main/RootRPCMethodsUI.js
bergarces, line 1: In case you are wondering where are the calls to `acceptWatchAsset` and `rejectWatchAsset`. They are all within the `WatchAssetRequest` component.

https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/WatchAssetRequest/index.js

patches/@metamask+assets-controllers+5.0.0.patch
bergarces, line 1: Changes applied here are from this PR:
https://github.com/MetaMask/core/pull/1166/files#diff-4de3568278666a6d7a1c5084b38802cc730ba3ce94b7fd90de875d758e462d99

Note that the constant from controller-utils (`ORIGIN_METAMASK`) has been replaced by its equivalent string to avoid having to patch that package as well. When packages are upgraded and this patch is removed, it won't make any difference.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -587,34 +586,34 @@ const RootRPCMethodsUI = (props) => {
   );
 
   /**
-   * On rejection addinga an asset
+   * On closing adding an asset modal
    */
-  const onCancelWatchAsset = () => {
-    setWatchAsset(false);
+  const onWatchAssetClosed = () => {
+    setShowPendingApproval(false);
```
matthewwalsh0, line 592: Should we be cautious and also reset the `watchAsset` to `undefined`?

app/core/RPCMethods/RPCMethodMiddleware.ts
matthewwalsh0: https://github.com/MetaMask/metamask-mobile/pull/6202

app/components/Nav/Main/RootRPCMethodsUI.js
bergarces: Definitely, it's also done for other modals, but somehow I missed it.

bergarces: It's actually not working if I do that.

The component `WatchAssetRequest` expects `suggestedAssetMeta` to be an object. It'll fail if undefined is passed (which occurs even though we first set `showPendingApproval` to false).

bergarces: It was worse than expected.

I had to add a few optional chainings to `WatchAssetRequest` so that it wouldn't crash. The component already had a few optional chainings with `asset`, so it's not completely unexpected.

app/components/UI/WatchAssetRequest/index.js
```diff
@@ -93,22 +93,31 @@ const WatchAssetRequest = ({
   onCancel,
   onConfirm,
 }) => {
-  const { asset, interactingAddress } = suggestedAssetMeta;
+  const {
+    asset,
+    interactingAddress,
+    id: suggestedAssetMetaId,
+  } = suggestedAssetMeta ?? {};
```
bergarces, line 100: Had to do this, as otherwise `setWatchAsset(undefined);` when closing the dialog immediately crashes the app when trying to extract `asset` from `suggestedAssetMeta`.

matthewwalsh0, line 100: My only concern with all the optional chaining is if we never want to show the component to the user without this state provided, then is it cleaner for maintainability to insist that state be present and therefore remove the burden of assuming each value could be missing.

If this were TypeScript for example, it's the difference between choosing an optional property versus a required one and we wouldn't make it optional if we had no use cases without the property.

bergarces: I've added a guard to the `renderWatchAssetModal` to not render the modal at all when `watchAsset` is `undefined`.

That how the `SignatureRequest` root component handles it.

app/core/RPCMethods/RPCMethodMiddleware.ts
vinistevam: I'm dreaming of this day.

vinistevam (approved):

matthewwalsh0 (approved):

PR #6432, opened 2023-05-18
feat(ci): Configure SonarCloud Analysis on CI

sethkfman:
**Description**

This PR updates the SonarCloud analysis to be run after the unit tests complete for PR analysis. You can see the successful SonarCloud run and results in the body of this PR. 

Further configuration will be needed because we haven't had a successful run on `main`.

**Screenshots/Recordings**

**Issue**

Progresses #???

**Checklist**


sonar-project.properties
```diff
@@ -2,11 +2,17 @@ sonar.projectKey=metamask-mobile
...
 
 # This is the name and version displayed in the SonarCloud UI.
-#sonar.projectName=MetaMask mobile project
+sonar.projectName=MetaMask Mobile Project
 #sonar.projectVersion=1.0
 
-# Path is relative to the sonar-project.properties file. Replace "\" by "/" on Windows.
-#sonar.sources=.
+# Root for sonar analysis.
+sonar.sources=app/
```
rickycodes, line 9: i would think that if we're adding `app/` here then we wouldn't need to add exclusions for things outside sources? (like android, ios, locales etc.)

sonar-project.properties
```diff
@@ -2,11 +2,17 @@ sonar.projectKey=metamask-mobile
...
+
+# Excluded project files from analysis.
+sonar.exclusions=android/**, ios/**, locales/**, patches/**, storybook/**, sourcemaps/**, **.java, **.h, **.m, scripts/**, *.js, app/__mocks__/**, app/lib/ens-ipfs/**
```
rickycodes, line 12: i would also think if we were to have a sploit of some sort they would probably come from code in the platform directories? how does this protect us from native code exploits?

rickycodes (commented):
just some initial feedback based on my assumptions so forgive my ignorance if i am not following things correctly

sonar-project.properties
sethkfman: At this point in time, we are focusing on JS code smells. It will be another effort to add capabilities to detect native code smell.

sonar-project.properties
sethkfman: Good catch. I have removed and tested with a successful run.

rickycodes (approved):

PR #6440, opened 2023-05-19
ci(GH Action): validate PR title validator against conventional commit format

sethkfman:
**Description**

Added a GH action step to validate the PR title against conventional commit standard. https://www.conventionalcommits.org/en/v1.0.0/

**Screenshots/Recordings**

**Issue**

Progresses #???

**Checklist**


owencraston (dismissed):
Same comment as https://github.com/MetaMask/metamask-extension/pull/19229#pullrequestreview-1434784687. Is there a way that we can link to the conventional commits doc in the case that it fails? LGTM otherwise!

sethkfman:
@owencraston if there is a failure the logs will print out your options here is an example error output. Let me know if that addresses you comment. 

```
No release type found in pull request title "docs(readme) :added readme for new action". Add a prefix to indicate what kind of release this pull request corresponds to. For reference, see https://www.conventionalcommits.org/

Available types:
 - feat: A new feature
 - fix: A bug fix
 - docs: Documentation only changes
 - style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
 - refactor: A code change that neither fixes a bug nor adds a feature
 - perf: A code change that improves performance
 - test: Adding missing tests or correcting existing tests
 - build: Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm)
 - ci: Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs)
 - chore: Other changes that don't modify src or test files
 - revert: Reverts a previous commit
```

owencraston:
@sethkfman this is even better

.github/workflows/ci.yml
```diff
@@ -32,9 +32,18 @@ jobs:
...
+    steps:
+      # v5.2.0
+      - name: Convetional Commit Validation PR Title
+        uses: amannn/action-semantic-pull-request@c3cd5d1ea3580753008872425915e343e351ab54
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   scripts:
     runs-on: ubuntu-20.04
-    needs: setup
+    needs: pr-linter
```
Gudahtt, line 46: It might be better to leave this as-is, and add `pr-linter` as a requirement of the `all-jobs-pass` step. That way it runs in parallel rather than serially

.github/workflows/ci.yml
```diff
@@ -32,9 +32,18 @@ jobs:
             echo "Working tree dirty at end of job"
             exit 1
           fi
+  pr-linter:
+    runs-on: ubuntu-latest
+    needs: setup
+    steps:
+      # v5.2.0
+      - name: Convetional Commit Validation PR Title
```
Gudahtt, line 40: Typo:
```suggestion
      - name: Conventional Commit Validation PR Title
```

.github/workflows/ci.yml
```diff
@@ -32,9 +32,18 @@ jobs:
             echo "Working tree dirty at end of job"
             exit 1
           fi
+  pr-linter:
+    runs-on: ubuntu-latest
+    needs: setup
```
Gudahtt, line 37: Does this need to depend upon setup?

```suggestion
```

Gudahtt (dismissed):
LGTM! Still two pending suggestions but they don't need to block

.github/workflows/ci.yml
Gudahtt: This is now even more urgent since setup was removed from pr-linter. Switching this back is required to prevent this job from being broken


```suggestion
    needs: setup
```

sethkfman: Should be addressed and thanks for the feedback.

Gudahtt: I'm not sure what you mean. This is not addressed, and as a result this job is no longer using the cached dependencies from the setup job.

sethkfman: In a better spot now. Let me know what you think.

Gudahtt (approved):

owencraston (approved):
👏

PR #6442, opened 2023-05-19
fix(ci): Remove `restore-build` steps

rickycodes:
**Description**

Previously the cache behaviour we were using here didn't make much sense. This removes the caching and restoring of the whole project path and instead only caches yarn as it should (this is the same strategy we use in `core`  and other projects).

<strike>I've also added a new output to the setup job to cut down on some of the duplication in the later jobs.</strike>

we're using `node-version-file` now for `actions/setup-node` so we don't have to read from the file ourselves.

There are some improvements that can be made (like caching some of the `postinstall` related steps). I believe some jobs might not even strictly require some of the `postinstall` steps at all? I'll be exploring that stuff and making improvements in a follow-up PR


.github/workflows/ci.yml
```diff
@@ -6,6 +6,8 @@ on:
 jobs:
   setup:
     runs-on: ubuntu-20.04
+    outputs:
+      node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
```
Gudahtt, line 10: Nit: We could have `setup-node` read the `nvmrc` file for us instead, so that we wouldn't need to keep track of it explicitly

Gudahtt (dismissed):
LGTM!

.github/workflows/ci.yml
rickycodes: good point! I've adjusted to use `node-version-file` instead

Gudahtt (approved):

sethkfman (approved):
LGTM

PR #6454, opened 2023-05-24
[e2e] Send ERC20 token from test-dapp (Android)

### LINKED ISSUE #6455: [e2e] Send ERC20 token from test-dapp (Android)

seaona:
## Description
This PR adds a test for sending an ERC20 token from a dapp. It is done the following:
- The HST contract from the `test-dapp` package is deployed on the background using Ganache Seeder
- The test uses the URL `https://metamask.github.io/test-dapp/?contract=${TOKEN_CONTRACT_ADDRESS}` this way we can skip the deploy step from the UI and go directly to Transfer Tokens

Additional changes that this PR brings:
- The Ganache steps are moved from the `common-steps` file to its dedicate file. This is done in order to keep code organized and in the forecast that more steps will be added regarding Ganache (different kind of contracts will be needed to be deployed in future tests)
- Using the `world` state from Cucumber allows us to access variables across the same scenario. This is needed for accessing the contract addresses deployed on Ganache. In the cases we access the world state, we cannot use arrow functions as they do not bind `this`, that's why it's changed in those cases
- Multisig address is now grabbed from the world state instead of beign hardcoded

## Screenshots/Recordings**

https://github.com/MetaMask/metamask-mobile/assets/54408225/b7d564c6-d630-4e72-aabf-fd16cf176259

image

Successful run on Bitrise https://app.bitrise.io/build/e2a5bf92-260c-4220-bc01-fcaa75fd6bcb

image

**Issue**

- Closes https://github.com/MetaMask/metamask-mobile/issues/6455

## Checklist


wdio/features/Confirmations/SendEthMultisig.feature
```diff
@@ -18,7 +18,7 @@ Feature: Sending ETH to a Multisig
     And Multisig contract is deployed
     When On the Main Wallet view I tap "ETHER"
     And On the Main Wallet view I tap "Send"
-    And I enter address "<MultisigAddress>" in the sender's input box
+    And I enter address "MultisigAddress" in the sender's input box
```
seaona, line 21: Improvement: we'll access the MultisigAddress from the `world` state instead of hardcoding it

wdio/step-definitions/ganache-steps.js
```diff
@@ -0,0 +1,30 @@
...
+  const contractRegistry = ganacheSeeder.getContractRegistry();
+  this.multisig = await contractRegistry.getContractAddress(SMART_CONTRACTS.MULTISIG);
+});
+
+Given(/^ERC20 token contract is deployed$/, async function() {
+  const ganacheSeeder = await new GanacheSeeder(ganacheServer.getProvider());
+  await ganacheSeeder.deploySmartContract(SMART_CONTRACTS.HST);
+  const contractRegistry = ganacheSeeder.getContractRegistry();
+  this.erc20 = await contractRegistry.getContractAddress(SMART_CONTRACTS.HST);
+});
```
seaona, line 30: Moved all ganache steps in a specific file and removed them from the `common-steps` file. We should expect to have more steps related to deploying new contracts, so it's a good reason to have a separate file just for that.

wdio/step-definitions/ganache-steps.js
```diff
@@ -0,0 +1,30 @@
...
+
+Given(/^Ganache server is started$/, async () => {
+  await ganacheServer.start({ mnemonic: validAccount.seedPhrase });
+});
+
+Then(/^Ganache server is stopped$/, async () => {
+  await ganacheServer.quit();
+});
+
+Given(/^Multisig contract is deployed$/, async function() {
```
seaona, line 18: Notice here we cannot use arrow functions, as we are passing the `this.multisig` to the world state from Cucumber, for being able to access it from another step

wdio/step-definitions/ganache-steps.js
```diff
@@ -0,0 +1,30 @@
...
+  const ganacheSeeder = await new GanacheSeeder(ganacheServer.getProvider());
+  await ganacheSeeder.deploySmartContract(SMART_CONTRACTS.MULTISIG);
+  const contractRegistry = ganacheSeeder.getContractRegistry();
+  this.multisig = await contractRegistry.getContractAddress(SMART_CONTRACTS.MULTISIG);
+});
+
+Given(/^ERC20 token contract is deployed$/, async function() {
```
seaona, line 25: Same as above, cannot use arrow functions as we need to bind `this`

wdio/step-definitions/send-flow.steps.js
```diff
@@ -26,9 +26,15 @@ Then(/^I tap the Save button/, async () => {
 
 Given(
   /^I enter address "([^"]*)?" in the sender's input box/,
-  async (address) => {
+  async function (address) {
```
seaona, line 29: Same as above, we use a world property `this.multisig` so arrow functions are not suitable here

chrisleewilcox: We have existing tests `SendToken.feature` and `AddressFlow.feature` that use this step and want to make sure that this change works for those tests as well.

seaona: thanks for pointing this @chrisleewilcox . Changes should not affect the mentioned tests:
- changing the arrow function only adds `this` binding, which is needed for this current test, but non-related to the other tests
- the switch below will only affect the `multisig` test and leaves unchanged the other cases (see `default`)

That being said, I've locally run the `AddressFlow.feature` and there is no change in behavior with or without these changes. However, the test is failing to click Save on the adress book - this is a non related error, as it also fails without the changes.
I've tried to proceed with the test, skipping this step and it works fine.

This test should be fixed in a different PR. Could not try the `SendToken` as requires ETH.

image

image

chrisleewilcox (approved):
LGTM

chrisleewilcox (approved):
LGTM

wdio/features/Confirmations/SendEthMultisig.feature
blackdevelopa: This isn't a feedback, just trying to understand what you mean by `world`?

blackdevelopa (approved):
PR looks good to me.

wdio/features/Confirmations/SendEthMultisig.feature
seaona: thank you for asking @blackdevelopa 
```
World, is an isolated scope for each scenario, exposed to the steps and most hooks as this. 
It allows you to set variables in one step and recall them in a later step. 
All variables set this way are discarded when the scenario concludes. 
```
You can see more here info: https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/world.md

PR #6486, opened 2023-05-29
feat: Add disabled prop on base button

### LINKED ISSUE #6475: Allow components library buttons to be disabled
Allow buttons in `app/component-library/components/Buttons` to handle a disabled state.
- apply a 50% transparency on a layer over the button
- add an isDisabled prop to all buttons (so maybe to `app/component-library/components/Buttons/Button/foundation/ButtonBase`)
- prevent onPress event to fire when disabled

NicolasMassart:
**Development & PR Process**

**Description**

- Add a `isDisabled` prop that changes the opacity of a button: default value when enabled (for backward compat) and 0.5 when disabled.
- map `isDisabled` prop to existing `disabled` prop from `TouchableWithoutFeedbackProps`
- update snapshot tests
- update readme

**Issue**

fixes #6475 

**Checklist**

### DIFF
--- a/app/component-library/components/Buttons/Button/foundation/ButtonBase/ButtonBase.styles.ts
+++ b/app/component-library/components/Buttons/Button/foundation/ButtonBase/ButtonBase.styles.ts
@@ -20,7 +20,7 @@ const styleSheet = (params: {
   vars: ButtonBaseStyleSheetVars;
 }) => {
   const { vars, theme } = params;
-  const { style, size, labelColor, width } = vars;
+  const { style, size, labelColor, width, disabled } = vars;
   const isAutoSize: boolean = size === ButtonSize.Auto;
   let widthObject;
   switch (width) {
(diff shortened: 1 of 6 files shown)


tommasini (commented):
I think the readme, can be updated as well 

The button component of component-library it's being used on a lot of places, I wonder if would be cool if instead of `isDisabled`, we use `disabled`, and destructure it from the props on ButtonBase.

. I think if we use `isDisabled` we need to change it also on the `ButtonPrimary`, `ButtonSecondary` and `ButtonLink` component properties because they receive the disabled property like on `UI/OptinMetrics` line 401

NicolasMassart:
> I think the readme, can be updated as well

Yes, thanks for the reminder, will do

> The button component of component-library it's being used on a lot of places, I wonder if would be cool if instead of `isDisabled`, we use `disabled`, and destructure it from the props on ButtonBase.
> 
> . I think if we use `isDisabled` we need to change it also on the `ButtonPrimary`, `ButtonSecondary` and `ButtonLink` component properties because they receive the disabled property like on `UI/OptinMetrics` line 401

That's what I've done at first, see commit 30de63a41 but then I realised that the CheckBox component was using `isDisabled` and as it was what I was first told to do, I decided to stick to this for consistency and to prevent unexpected issues that may be the reason why we do it this way. More history would be needed to confirm. Also, the ButtonIcon is using `disabled` directly, so maybe we could. That's something to improve but if we do, we could also improve CheckBox at the same time. I would recommend another PR for this.

brianacnguyen (approved):

PR #6489, opened 2023-05-29
feat: [MC 0.75] Show test network toggle

tommasini:
**Description**
Implementation of test network toggle on Network Selector component and persist the value on redux 

**Screenshots/Recordings**
scenario: Open the app, activate the test network, close it, open it, it should show the test networks: https://recordit.co/yH4xwUKVPv
scenario: Open the app, deactivate the test network, close it, open it, it should not show the test networks: https://recordit.co/OyrEYWCPgO

scenario: Toggle should be on and disabled whenever a testnet is selected:
https://recordit.co/JMNzBaV75M

**Core pull request**
https://github.com/MetaMask/core/pull/1418

**Issue**

Progresses #???

**Checklist**


app/components/Views/NetworkSelector/NetworkSelector.test.tsx
```diff
@@ -167,4 +174,12 @@ describe('Network Selector', () => {
 
     expect(setRpcTargetSpy).toBeCalled();
   });
+  it('toggles the test networks switch correctly', () => {
```
Cal-L, line 177: Can we also add a test where the toggle is on and disabled when testnet is selected?

tommasini: Addeeeedd!

patches/@metamask+preferences-controller+2.1.0.patch
```diff
@@ -1,16 +1,20 @@
...
-index 4ebc51c..74825d5 100644
+index 4ebc51c..45a0787 100644
 --- a/node_modules/@metamask/preferences-controller/dist/PreferencesController.js
 +++ b/node_modules/@metamask/preferences-controller/dist/PreferencesController.js
-@@ -29,6 +29,7 @@ class PreferencesController extends base_controller_1.BaseController {
+@@ -29,9 +29,11 @@ class PreferencesController extends base_controller_1.BaseController {
              useTokenDetection: true,
              useNftDetection: false,
              openSeaEnabled: false,
 +            isMultiAccountBalancesEnabled: true,
```
Cal-L, line 9: Is this change already in the controllers repo or do we need to follow up on creating a PR for it?

tommasini: Already there https://github.com/MetaMask/core/blob/43dfaf348c1af150a9e2b688ac23c2ab7c318fee/packages/preferences-controller/src/PreferencesController.ts#L39

tommasini: This core [PR](https://github.com/MetaMask/core/pull/1418) didn't go, we could create an issue to update the preferences controller since we are on version 2 and the last version it's version 4, since it has two major versions it would be nice to take time to see what's the breaking changes, wdyt?

Cal-L: Ah ok, looks like we'll have a follow up. LGTM

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -218,14 +232,38 @@ const NetworkSelector = () => {
...
   };
 
+  const renderTestNetworksSwitch = () => (
+    <View style={styles.switchContainer}>
+      <Text variant={TextVariant.BodyLGMedium} color={TextColor.Alternative}>
+        {strings('networks.show_test_networks')}
+      </Text>
+      <Switch
+        onValueChange={(value: boolean) => {
+          const { PreferencesController } = Engine.context;
```
NicolasMassart, line 242: C: [we have discussed internaly](https://consensys.slack.com/archives/C02U025CVU4/p1687356081509889) that it would be better not to access state with the Engine:
> it would be nice to eliminate the need for directly accessing Engine.context all together. I believe the reason why we used Engine.context is for areas where Redux wasn't easily accessible. However, even those areas, we should be able to access redux data via importing the store.

NicolasMassart (approved):
Looks good to me.

app/components/Views/NetworkSelector/NetworkSelector.tsx
tommasini: I would say no, because you don't have the method on the redux store, you only store the data. Here we are accessing the Engine to use a PreferencesController method

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -176,10 +190,10 @@ const NetworkSelector = () => {
               name,
               imageSource: image,
             }}
-            isSelected={
+            isSelected={Boolean(
```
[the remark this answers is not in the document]
tommasini: In this case, the expression doesn't return a boolean
`
'1' === '1' && 'test'
--> returns 'test'

Boolean('1' === '1' && 'test')
--> returns true
`
We need to check if the rpcTarget it's defined because it's the way we know it is a custom network, since Mainnet, goerli and sepolia don't have it on their data structure on the app, we check if there isn't a rpcTarget on rendering mainnet, to make sure if we have two ethereum rpc nodes on the wallet we only have one selected

Issue #6434, opened 2023-05-19
[Bug] Account name incorrectly displayed on another address


seaona:
**Describe the bug**
When I perform a transaction from a Dapp I can see how the `To` field displays `Account 2` to a recipient that it's not Account 2. It seems that this is caching the value from a previous transaction performed from inside MetaMask to Account 2.

**Screenshots**

https://github.com/MetaMask/metamask-mobile/assets/54408225/4a18af33-8816-4053-aad4-f05830cca273

**To Reproduce**
_Steps to reproduce the behavior_
1. Install MetaMask
2. Initiate a Send ETH to Account 2
3. Cancel Tx
4. Go to dapp
5. Connect MM
6. Click Send EIP1559
7. See how Account 2 name is displayed

**Expected behavior**
We should not cache the Account name whenever initiating a new transaction from a Dapp

**Smartphone (please complete the following information):**
 - Device: Pixel 6
 - OS: Android
 - App Version: `v6.5.0` or even earlier

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

bschorchit:
Seems possibly related with https://github.com/MetaMask/metamask-mobile/issues/6435

PR #6502, opened 2023-05-31
fix: [7.1.0] Account incorrectly displayed on transaction review modal

tommasini:
**Description**
The transaction on send flow was not being reset when we left the send flow.

**Proposed Solution**
Reset the transaction state when you hit cancel on the send flow.

**Screenshots/Recordings**
https://recordit.co/3qy43LC5KN

**Issue**

Progresses #https://github.com/MetaMask/metamask-mobile/issues/6434

**Checklist**

### DIFF
--- a/app/components/Views/SendFlow/Confirm/index.js
+++ b/app/components/Views/SendFlow/Confirm/index.js
@@ -276,10 +276,16 @@ class Confirm extends PureComponent {
   };
 
   updateNavBar = () => {
-    const { navigation, route } = this.props;
+    const { navigation, route, resetTransaction } = this.props;
     const colors = this.context.colors || mockTheme.colors;
     navigation.setOptions(
-      getSendFlowTitle('send.confirm', navigation, route, colors),
+      getSendFlowTitle(
+        'send.confirm',
+        navigation,
+        route,
+        colors,
+        resetTransaction,
+      ),
     );
   };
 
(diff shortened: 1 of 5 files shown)


app/components/Views/SendFlow/Confirm/index.js
```diff
@@ -276,10 +276,16 @@ class Confirm extends PureComponent {
   };
 
   updateNavBar = () => {
-    const { navigation, route } = this.props;
+    const { navigation, route, resetTransaction } = this.props;
```
Cal-L, line 279: Seems like it might be cleaner to place the resetTransaction in the unmount life cycle instead?

tommasini: If we use it in the unmount life cycle when we go back to the amount screen from the confirm screen, it will reset the transaction, and we don't want that, am I missing something?

Cal-L: I see, it's needed in previous flows too. Can you check if it's possible to swipe to dismiss though on the last screen? And in that scenario, does the transaction get reset successfully?

tommasini: Checking it!

Cal-L (approved):
Lgtm

chrisleewilcox (approved):
Tested on....

- Samsung Galaxy A42 Android 13: PASS
- iPhone 12 iOS 14.4 simulator: PASS

LGTM

PR #6510, opened 2023-06-01
chore: pending review feedback for token details related changes

nikoferro:
Changes included for asset overview view

- Avoid trying to convert the price to selected currency if there is no fiat price reference available
- Add matching gradient to the area chart
- Make chart size bigger in relation to the rest of the asset overview
- Better gesture handling when interacting with the chart
- Subtle shadow on asset overview footer to match other parts of the app
- Change name from ETHER to Ethereum for mainnet's native currency

**Screenshots/Recordings**

screen


app/components/Views/Asset/index.js
```diff
@@ -497,7 +505,7 @@ class Asset extends PureComponent {
           />
         )}
         {!asset.balanceError && (
-          <View style={styles.footer}>
+          <View style={{ ...styles.footer, ...styles.footerBorder }}>
```
NicolasMassart, line 508: Q: I'm not sure what this this spread `...` does here with styles. Can you explain please?

leotm: ^ should be equivalent to the style array format

```tsx
<View style={[styles.footer, styles.footerBorder }]>
```

so just applying `footer` style first
then applying (overriding) with `footerBorder` style

> last style in the array has precedence, so you can use this to inherit styles

nikoferro: ☝️  @NicolasMassart pretty much what @leotm said. on the app it seems we use both approaches (the array and the spread)

jpcloureiro (approved):
code lgtm

leotm (approved):

PR #6511, opened 2023-06-01
test: Send ETH with Gas API Down - Introducing Mock Responses

seaona:
## Description
This PR adds a testcase for Sending ETH when the Gas API is down. We will validate that the correct gas fallback is displayed.
For achieving this, we introduce the ability to Mock responses on the background, using `mockttp` library.

**Sidenote**: it also fixes broken tests due to update on the Home/Token View

## Screenshots/Recordings

https://github.com/MetaMask/metamask-mobile/assets/54408225/e9d8a9cc-83fe-43c6-9d0b-3cbf07f7066c

Screenshot from 2023-06-01 13-21-41

**Bitrise successful build**: https://app.bitrise.io/build/305d41a6-02c2-4f18-9640-028a74df22ad

image

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/1007

**Checklist**


wdio/step-definitions/common-steps.js
```diff
@@ -237,22 +237,6 @@ Given(/^I close the Whats New modal$/, async () => {
   await WhatsNewModal.waitForDisappear();
 });
 
-Given(/^Ganache server is started$/, async () => {
```
seaona, line 240: For some reason this section was added back in another PR - this was moved to the `ganache-steps` file, so it should be removed from here

wdio/step-definitions/mock-steps.js
```diff
@@ -0,0 +1,25 @@
+import { Given, Then } from '@wdio/cucumber-framework';
+const mockttp = require('mockttp');
+
+Given(/^Mock server is started with mock Gas API Down$/, async () => {
```
seaona, line 4: Here we mock the API response to return a 500, so we then check that we display the Fallback Gas API screen

wdio/features/Confirmations/SendETHGasApiDown.feature
```diff
@@ -0,0 +1,36 @@
...
+    When I tap Edit Gas link
+    Then EIP1559 gas options should not be visible
+    When I tap Save Gas Values
+    And I tap button "Send" on Confirm Amount view
+    Then I am on the main wallet view
+    And the transaction is submitted with Transaction Complete! toast appearing
+    Then Ganache server is stopped
+    Examples:
+      | Address                                    | Amount |
+      | 0x1FDb169Ef12954F20A15852980e1F0C122BfC1D6 | 1      |
```
chrisleewilcox, line 36: I feel like this can be better understood by breaking this scenario down into more scenarios....

Scenario: Setup Genache Server
Scenario Outline: Enter Send Amount
Scenario: Edit Gas Fees
Scenario: Send Transaction

Can we break it down like that?  No changes in steps just grouping them in scenarios that make the tests and steps more understood.  This will also make it easier to expand and add to this test for each scenarios such as negative test for inputing invalid values for send amount and other things.
@seaona

chrisleewilcox (changes requested):

wdio/features/Confirmations/SendETHGasApiDown.feature
seaona: Thank you @chrisleewilcox I've updated the test breaking it down in more Scenarios

chrisleewilcox (dismissed):

blackdevelopa (dismissed):

naugtur:
@SocketSecurity ignore-all

Checked the ones where a maintainer changed, but these are very old releases with decent download counts, nothing to worry about.
FS and Net access for dependencies of a http mocking tool makes sense :) 
Some of the stuff is a bit dated, but fine for a dev dependency.

segun (dismissed):

NiranjanaBinoy (dismissed):

SamuelSalas (dismissed):

chrisleewilcox (dismissed):

wdio/features/Confirmations/SendEthGasApiDown.feature
```diff
@@ -0,0 +1,34 @@
...
+
+  Scenario: Import wallet
+    Given the app displayed the splash animation
+    And I have imported my wallet
+    And I tap No Thanks on the Enable security check screen
+    And I tap No thanks on the onboarding welcome tutorial
+
+  @mock
+  @gasApiDown
+  Scenario: Display fallback on the Gas Edit screen
```
seaona, line 16: Each test must be under the same scenario as we use hooks for starting/stopping the needed servers and pass them on the World object (which is reset on each scenario)

segun (approved):

NiranjanaBinoy (approved):

PR #6519, opened 2023-06-02
test: introducing basic infrastructure to support fixtures on e2e tests

vinistevam:
**Description**

The aim of this PR is to have a stable POC to include the possibility to use fixtures in e2e tests on Appium and Detox (ios/android) based on what already exists in the extension.

```
# add export IS_TEST="true" in the file .js.env and e2e.env

# Run on Appium

yarn watch
# On another terminal
yarn start:android:qa 
# Once the build finishes
yarn test:wdio:android --spec ./wdio/features/Wallet/LockResetWallet.feature


## Run on Detox (Tested with branch add-android-detox-coverage)
yarn watch
# On another terminal
yarn test:e2e:ios:debug / detox build -c ios.sim.debug
# Once the build finishes
yarn test:e2e:ios:debug:single e2e/specs/delete-wallet.spec.js

#  Detox android
yarn test:e2e:android:debug / detox build -c android.emu.debug
yarn test:e2e:android:debug:single e2e/specs/delete-wallet.spec.js
```

I've tested the detox e2e test on Android and the fixture POC works as in Appium. The branch `android-detox-fixture` is based on `add-android-detox-coverage`.

**Screenshots/Recordings**
E2E pipeline with fixtures: 
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/2ddc3470-dbe5-42fe-a0ea-5c8586b00d4f

[fixture-poc-detox-android.webm](https://github.com/MetaMask/metamask-mobile/assets/45455812/d3f682d5-b18b-4694-a09d-7d7517775dd1)

Smoke tests with fixtures
image

### Overview
```mermaid
sequenceDiagram
    box grey Test Context
    participant TC as Test Case
    participant FB as Fixture Builder
    participant FH as Fixture Helper
    participant FS as Fixture Server
    end
    box MMM App
    participant NS as Network Store
    participant RS as Redux Store
    participant AS as Async Storage (wrapped)
    end
    Note over TC: Before hook
    TC->>+FS: Start fixture server
    FS->>-TC: Server started
    TC->>+FB: Build fixtures 
    FB->>-TC: JSON object 
    TC->>FS: Load JSON State 
    Note over FS: Fixture server ready
    Note over TC: Starting test case
    Note over NS: Initializing
    NS->>FS: Get /state.json
    FS->>NS: 
    Note over NS: Initialized
    NS->>RS: Provide state to
    NS->>AS: Provide state to
```

### Components

[fixture-server](https://github.com/MetaMask/metamask-mobile/blob/1dc3dd40e468371ae22d7d7ac6c60cb6549a6b29/wdio/fixtures/fixture-server.js): sets up a `Koa` server to handle requests. It includes functions for starting and stopping the server, as well as loading JSON state into the server. The class also includes methods for performing substitutions on the state fixture, allowing for dynamic values. The server responds to `GET` requests to `/state.json` by returning the current state stored.

[fixture-builder](https://github.com/MetaMask/metamask-mobile/blob/1dc3dd40e468371ae22d7d7ac6c60cb6549a6b29/wdio/fixtures/fixture-builder.js): allows constructing fixture data for tests. The class provides methods for setting up various properties and configurations of the fixture. It is used to create consistent and reusable fixtures for testing purposes.

[fixture-helper](https://github.com/MetaMask/metamask-mobile/blob/1dc3dd40e468371ae22d7d7ac6c60cb6549a6b29/wdio/fixtures/fixture-helper.js): sets up a fixture server and includes functions for loading, starting, and stopping the fixture server. The `loadFixture` function loads a fixture or creates a default one using `FixtureBuilder` and loads it into the fixture server. It also checks if the state is properly loaded by making a request to the server. The `startFixtureServer` function starts the fixture server, while the `stopFixtureServer` function stops it. The `withFixtures` function wraps a test suite, starting the fixture server, loading fixtures, running the test suite, and stopping the server afterwards.

[network-store](https://github.com/MetaMask/metamask-mobile/blob/1dc3dd40e468371ae22d7d7ac6c60cb6549a6b29/app/util/test/network-store.js): initializes and manages network state. It fetches the network state from a fixture server, initializes the state asynchronously, and provides methods to access and modify it.

[store/index.js](https://github.com/MetaMask/metamask-mobile/blob/poc-add-fixures-e2e/app/store/index.js): creates the redux store and whenever `isTest` is true via the pre-determined state provided by network-store injects the state into the redux store.

[async-storage-wrapper](https://github.com/MetaMask/metamask-mobile/blob/1dc3dd40e468371ae22d7d7ac6c60cb6549a6b29/app/store/async-storage-wrapper.js): acts as a wrapper around the `AsyncStorage` module from the `@react-native-async-storage/async-storage` package. The purpose of this class is to provide a unified interface for storing, retrieving, and removing data using the `AsyncStorage` module.

[isTest utility function](https://github.com/MetaMask/metamask-mobile/blob/1dc3dd40e468371ae22d7d7ac6c60cb6549a6b29/app/util/test/utils.js#L4): provides a simple way to determine if the environment is set to a testing mode. It allows conditional logic to be applied based on the value of the `IS_TEST` environment variable, enabling different behaviour or configurations for testing scenarios.

### Example
**How to use fixtures on cucumber tests?**
```javascript
//wdio.conf.js
    if (tags.filter((e) => e.name === FIXTURES_SKIP_ONBOARDING).length > 0) {
      const state = new FixtureBuilder().build();
      await loadFixture({fixture: state});
    }

//step-definition/ (It's also possible to use step definition for specific use cases)
  Given(/^State is loaded with fixtures to skip onboarding steps$/, async () => {
    const state = new FixtureBuilder().build();
    await loadFixture({fixture: state});
  });

//test.feature
@androidApp
@smoke
@onboarding
@fixturesSkipOnboarding
Feature: Skip onboarding test fixture

  Scenario: Skip onboarding and unlock wallet 
    User opens the app and unlock wallet
    Given the app displayed the splash animation
    And the splash animation disappears
    And I unlock wallet with <password>
    Then Wallet view is displayed
    And "Account 1" is visible
    Examples:
      | password   |
      | 123123123 |
```

**How to use fixtures on detox and jest tests?**
```javascript
// 1- Loading fixtures using BeforeAll hook
    beforeAll(async () => {
      const fixture = new FixtureBuilder().build();
      await loadFixture({ fixture });
      await device.launchApp({ newInstance: true });
    });

// 2- Using the utility function `withFixtures`, using this function you can wrap the test suit, load fixtures and get the device ready
  it('should launch the app and log in', async () => {
    const fixture = new FixtureBuilder().build();
    await withFixtures({ fixture, restartDevice: true }, async () => {
      await LoginView.isVisible();
      await LoginView.enterPassword(PASSWORD);

      await WalletView.isVisible();
    ... // more actions
    });
  });
```

**How to create new fixtures?**
```javascript
// fuxture-builder.js
  withPermissionController(data) {
    merge(this.fixture.data.PermissionController, data);
    return this;
  }

  withPermissionControllerConnectedToTestDapp() {
    return this.withPermissionController({
      subjects: {
        [DAPP_URL]: {
          origin: DAPP_URL,
          permissions: {
            eth_accounts: {
              id: 'ZaqPEWxyhNCJYACFw93jE',
              parentCapability: 'eth_accounts',
              invoker: DAPP_URL,
              caveats: [
                {
                  type: 'restrictReturnedAccounts',
                  value: ['0x5cfe73b6021e818b776b421b1c4db2474086a7e1'],
                },
              ],
              date: 1664388714636,
            },
          },
        },
      },
    });
  }
```

**Issue**

https://github.com/MetaMask/mobile-planning/issues/843
https://github.com/MetaMask/mobile-planning/issues/974

**Checklist**


app/components/Nav/App/index.js
```diff
@@ -237,7 +238,7 @@ const App = ({ userLoggedIn }) => {
     const appTriggeredAuth = async () => {
       const { PreferencesController } = Engine.context;
       const selectedAddress = PreferencesController.state.selectedAddress;
-      const existingUser = await AsyncStorage.getItem(EXISTING_USER);
+      const existingUser = await asyncStorageWrapper.getItem(EXISTING_USER);
```
matthewwalsh0, line 241: Rather than replacing every import, could we instead update the library directly by assigning functions, or using a third party library, or even just patching it?

app/components/Views/Root/index.js
```diff
@@ -34,16 +34,46 @@ export default class Root extends PureComponent {
...
-      <PersistGate persistor={persistor}>
-        <ConnectedRoot />
-      </PersistGate>
-    </Provider>
-  );
+  async componentDidMount() {
+    const { isTest } = this.state;
+    if (isTest) {
+      // Wait until store is initialized
+      await new Promise((resolve) => {
```
matthewwalsh0, line 48: Is it worth encapsulating this in a private `waitForStore` method?

app/images/static-logos.js
```diff
@@ -2,13 +2,17 @@
 /* eslint-disable */
 
 export default {
+  'ftn.svg': require('metamask/node_modules/@metamask/contract-metadata/images/ftn.svg'),
```
matthewwalsh0, line 5: Bad merge?

app/store/async-storage-wrapper.js
```diff
@@ -0,0 +1,24 @@
+import ReadOnlyNetworkStore from '../util/test/network-store';
+import { AsyncStorage } from '@react-native-async-storage/async-storage';
+
+const isTest = process.env.IS_TEST === 'true';
```
matthewwalsh0, line 4: Is it worth putting this in a method somewhere as we do it in multiple places?

app/store/index.js
```diff
@@ -15,8 +15,30 @@ import Logger from '../util/Logger';
 import EngineService from '../core/EngineService';
 import { Authentication } from '../core';
 import Device from '../util/device';
+import ReadOnlyNetworkStore from '../util/test/network-store';
 
 const TIMEOUT = 40000;
+const isTest = process.env.IS_TEST === 'true';
+
+const ReadOnlyNetworkStorage = {
```
matthewwalsh0, line 23: Did we need this in the end? If I understand right, we are overriding three different flows by updating:

- The initial state of the root store.
- The references to `AsyncStorage`.
- The `storage` property in the `persistConfig`.

Are all of these in use and unrelated to each other? Do we have a ballpark figure of which is most prevalent?

app/store/index.js
```diff
@@ -109,29 +131,42 @@ const persistUserTransform = createTransform(
...
-/**
- * Initialize services after persist is completed
- */
-const onPersistComplete = () => {
-  EngineService.initalizeEngine(store);
-  Authentication.init(store);
-};
+// eslint-disable-next-line import/no-mutable-exports
+let store, persistor;
+if (isTest) {
```
matthewwalsh0, line 159: Would it be cleaner to put this logic inside a method that we call inline?

The `else` condition is almost identical also so we could remove some duplication.

app/util/test/network-store.js
```diff
@@ -0,0 +1,91 @@
+import axios from 'axios';
```
matthewwalsh0, line 1: Is it worth commenting that this is identical to what is in the extension?

We could eventually move it to core to avoid the duplication.

wdio/fixtures/fixture-server.js
matthewwalsh0, line 1: Would we want both this server and the builder to exist above the Android tests specifically?

Am I right this has been tested on Android but not yet Detox?

app/store/index.js
matthewwalsh0: As discussed, in an ideal world, we'd have a single asynchronous intersection point where all flows ultimately read state from, that we could then override or configure to inject our own logic which would use the state from the server.

In the extension for example, all state ultimately comes from the local store initialised in the background script.

If this doesn't exist in mobile, it'd likely be worth the investment to align any existing flows to use a single mechanism.

app/images/static-logos.js
vinistevam: removed, my bad.

app/store/index.js
vinistevam: After changing the approach to only overwrite the `state.getState()` I run some tests using the original `MigratedStorage` with fixtures and they all passed. I removed `ReadOnlyNetworkStorage` which will simplify the changes to create the persistor and store.

app/util/test/network-store.js
vinistevam: It's not 100% identical, but 100% compatible I just had to include some extra functions to cover the Async Storage.

leotm (commented):
thanks again for walking thru this today ^ just a reminder that could be worth adding notes on this to our ReadMe.md (for internal/external contributors)

wdio/config/android.config.debug.js
```diff
@@ -8,7 +8,7 @@ config.capabilities = [
     noReset: false,
     fullReset: false,
     maxInstances: 1,
-    deviceName: 'Pixel 5 API 32',
+    deviceName: 'Pixel 3 API 33',
```
seaona, line 11: I think we can leave this as it was previously, as it depends on each setup

wdio/features/Onboarding/SkipOnboarding.feature
```diff
@@ -0,0 +1,16 @@
+@androidApp
+@smoke
+@onboarding
+Feature: Skip onboarding test fixture
+
+  Scenario: Skip onboarding and unlock wallet 
+    User opens the app and unlock wallet
+    Given State is loaded with fixtures to skip onboarding steps
```
seaona, line 8: I am wondering if instead of using fixtures as a "step" in the feature file, we could trigger them on the `hooks` level, like we do with Ganache server, deployment of smarts contracts, mocks, etc. and then use the cucumber tag in the feature file, for indicating that the test uses that particular fixture.

See ganache, mocks, etc implemented in the `hooks` : https://github.com/MetaMask/metamask-mobile/blob/main/wdio.conf.js#L316

See how this is then triggered on a particular test, by using the `tags`: i.e. in this case `@ganache` starts and tears down ganache server, and `@erc721` deploys an NFT conctract, before the test begins.

https://github.com/MetaMask/metamask-mobile/blob/main/wdio/features/Confirmations/SendERC721.feature#L8

vinistevam: That is a great suggestion, thanks :1st_place_medal: ! I pushed the changes to introduce the tag to support loading the state before any step of the scenario.  [here](https://github.com/MetaMask/metamask-mobile/pull/6519/files#diff-f297ad33212595210fbfc7c32d334f7025fe8eee14cc4179d2527c3f4543035fR347)

vinistevam:
@SocketSecurity ignore koa@2.14.2

app/store/index.js
```diff
@@ -124,14 +126,27 @@ const persistConfig = {
...
 const pReducer = persistReducer(persistConfig, rootReducer);
 
-export const store = createStore(pReducer, undefined, applyMiddleware(thunk));
+// eslint-disable-next-line import/no-mutable-exports
+let store, persistor;
+const createStoreAndPersistor = async () => {
+  const state = isTest ? await ReadOnlyNetworkStore.getState() : undefined;
+  store = createStore(pReducer, undefined, applyMiddleware(thunk));
+  // Use preloaded state from fixture
+  if (state) store.getState = () => state;
```
cryptotavares, line 135: Why can't we just pass the fixture state as preloaded state when creating the redux store? Something like:
```js
  const state = isTest ? await ReadOnlyNetworkStore.getState() : undefined;
  store = createStore(pReducer, state, applyMiddleware(thunk));
```

e2e/init.js
```diff
@@ -1,3 +1,20 @@
...
+} from '../wdio/fixtures/fixture-helper';
+
 beforeAll(async () => {
+  // Avoid port forwarding in Bitrise
+  const isRunningBitrise = process.env.BITRISE_APP_TITLE;
+  if (!isRunningBitrise) {
+    await device.reverseTcpPort(12345);
+  }
+  await startFixtureServer();
+  await loadFixture();
```
cryptotavares, line 14: Should we load the fixtures in a global before hook? or should we follow the extension approach of starting the server, loading the fixtures and then tearingdown the server per test? (in this way allowing tests to setup their own fixtures)

e2e/init.js
```diff
@@ -1,3 +1,20 @@
...
+  loadFixture,
+  startFixtureServer,
+  stopFixtureServer,
+} from '../wdio/fixtures/fixture-helper';
+
 beforeAll(async () => {
+  // Avoid port forwarding in Bitrise
+  const isRunningBitrise = process.env.BITRISE_APP_TITLE;
+  if (!isRunningBitrise) {
+    await device.reverseTcpPort(12345);
```
cryptotavares, line 11: I know that it is not in this PR scope, but should we add the reverse port for ganache as well? (just as we have for appium)

wdio/fixtures/fixture-builder.js
```diff
@@ -0,0 +1,528 @@
+/**
```
cryptotavares, line 1: Maybe we could move this files into the e2e directory (where all the detox files are).. as we intend to migrate everyting into detox

vinistevam:
@SocketSecurity ignore koa@2.14.2

e2e/init.js
vinistevam: I migrate one of the existing jest tests to use fixtures through the `withFixtures` function. Unfortunately, it's still necessary to load a general fixture because once the flag `IS_TEST` is `true` all the tests are using the fixtures' infrastructure and need it to provide an empty initial state to be backwards compatible with the existent tests.

app/store/index.js
vinistevam: Depending on the fixtures it takes a long time to create a store with an initial state and then override the `getState` after creating the store. All my attempts to workaround that ended up with the [initializeEngine](https://github.com/MetaMask/metamask-mobile/blob/5a8723529f81b243b414b76abc061f7112a9dcc6/app/core/EngineService/EngineService.ts#L27)  setting an empty state to the controllers.

e2e/init.js
cryptotavares: Ok it makes sense.. I am just wondering then if we shouldn't reset the state before each test execution? (the idea is to make tests atomic.. no app state leakeage between tests).. Although I am not sure on the impact that that might have on already existing tests.
Let's keep it as is and create a subsequent task to look into it.

e2e/fixtures/fixture-builder.js
```diff
@@ -0,0 +1,530 @@
...
+class FixtureBuilder {
+  /**
+   * Create a new instance of FixtureBuilder.
+   * @param {Object} options - Options for the fixture builder.
+   * @param {boolean} options.onboarding - Flag indicating if onboarding fixture should be used.
+   */
+  constructor({ onboarding = false } = {}) {
+    // Initialize the fixture based on the onboarding flag
+    onboarding === true
+      ? this.withOnboardingFixture()
```
cryptotavares, line 13: The naming seems confusing to me... `withOnboardingFixture` seems to suggest that onboarding will be fixture, and thus the e2e test will bypass it. But, again, the same is true for the extension fixtures code base... So, for consistency reasons let's keep it like this

e2e/fixtures/fixture-helper.js
```diff
@@ -0,0 +1,78 @@
...
+  let failed;
+  try {
+    // Start the fixture server
+    await startFixtureServer();
+    await loadFixture({ fixture });
+    console.log(
+      'The fixture server is started, and the initial state is successfully loaded.',
+    );
+
+    if (restartDevice) {
```
cryptotavares, line 64: Can we add a small explanation or comment to when one would expect to pass the `restartDevice=true`? In which cases do we want to set it to true on e2e tests?

vinistevam: Sure I will include a comment here.
Currently, this is required because the app is launched on `init.js` at the beginning of the process, so once in the test file we load the fixtures a restart of the app is necessary. Ideally, as progressively we start to use fixtures we move the `await device.launchApp();` to the file and use `beforeAll` or `beforeEach` hooks to load the fixtures and launch the app or not using the hooks and rely on `withFixtures` function, depending on the use case.

Andepande (dismissed):
LGTM some great work, looking forward to using it

Andepande (dismissed):

e2e/viewHelper.js
```diff
@@ -181,3 +182,11 @@ export const switchToGoreliNetwork = async () => {
   await WalletView.isNetworkNameVisible(GOERLI);
   await NetworkEducationModal.tapGotItButton();
 };
+
+export const skipOnboarding = async () => {
+  const PASSWORD = '123123123';
+  await LoginView.isVisible();
+  await LoginView.enterPassword(PASSWORD);
```
cortisiko, line 189: You can import the password from the validAccount object. 

```suggestion
  await LoginView.enterPassword(validAccount.password);
```

app/components/Views/RevealPrivateCredential/RevealPrivateCredential.tsx
```diff
@@ -10,7 +10,7 @@ import {
   View,
 } from 'react-native';
 import { useDispatch, useSelector } from 'react-redux';
-import AsyncStorage from '@react-native-async-storage/async-storage';
+import AsyncStorage from '../../../store/async-storage-wrapper';
```
cortisiko, line 13: Whenever we intend to use fixtures in our tests, is it necessary to include 

`import AsyncStorage from '../../../store/async-storage-wrapper';`

 in all areas of the app code that the test interacts with?

package.json
```diff
@@ -34,12 +34,12 @@
...
     "build:android:pre-release": "./scripts/build.sh android release --pre",
     "build:android:pre-release:bundle": "GENERATE_BUNDLE=true ./scripts/build.sh android release --pre",
-    "build:android:pre-release:bundle:qa": "GENERATE_BUNDLE=true ./scripts/build.sh android QA --pre",
+    "build:android:pre-release:bundle:qa": "IS_TEST='true' GENERATE_BUNDLE=true ./scripts/build.sh android QA --pre",
     "build:ios:release": "./scripts/build.sh ios release",
     "build:ios:release:e2e": "./scripts/build.sh ios releaseE2E",
     "build:ios:pre-release": "./scripts/build.sh ios release --pre",
     "build:ios:qa": "./scripts/build.sh ios QA",
-    "build:ios:pre-qa": "./scripts/build.sh ios QA --pre",
+    "build:ios:pre-qa": "IS_TEST='true' ./scripts/build.sh ios QA --pre",
```
cortisiko, line 42: any reason why we add this flag to the pre-qa build for both iOS and android?

e2e/init.js
```diff
@@ -1,3 +1,20 @@
+import {
+  loadFixture,
+  startFixtureServer,
+  stopFixtureServer,
+} from './fixtures/fixture-helper';
+
 beforeAll(async () => {
```
cortisiko, line 7: We won't use fixtures for all tests. Instead, we'll enable fixtures at the test level, allowing us to use them selectively. Some tests will remain unchanged, while for others, such as the confirmation tests, we'll utilize fixtures.

cortisiko (changes requested):

e2e/fixtures/fixture-builder.js
vinistevam: I was thinking of an option `withoutFixtures` / `noFixtures` but as you mentioned I left it like this for consistency at least for the infrastructure PR.

package.json
```diff
@@ -34,7 +34,7 @@
     "build:android:checksum:verify": "shasum -a 512 -c sha512sums.txt",
     "build:android:pre-release": "./scripts/build.sh android release --pre",
     "build:android:pre-release:bundle": "GENERATE_BUNDLE=true ./scripts/build.sh android release --pre",
-    "build:android:pre-release:bundle:qa": "GENERATE_BUNDLE=true ./scripts/build.sh android QA --pre",
+    "build:android:pre-release:bundle:qa": "IS_TEST='true' GENERATE_BUNDLE=true ./scripts/build.sh android QA --pre",
```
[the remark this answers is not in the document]
vinistevam: Included in the `.js.env.example`.

e2e/specs/delete-wallet.spec.js
```diff
@@ -13,89 +13,67 @@ import SecurityAndPrivacyView from '../pages/Drawer/Settings/SecurityAndPrivacy/
...
 
-    it('should delete wallet', async () => {
-      await DeleteWalletModal.tapIUnderstandButton();
-      await DeleteWalletModal.typeDeleteInInputBox();
-      await DeleteWalletModal.tapDeleteMyWalletButton();
-      await TestHelpers.delay(2000);
-      await OnboardingView.isVisible();
+    it('should log in into the app, change password then delete wallet flow', async () => {
+      const fixture = new FixtureBuilder().build();
+      await withFixtures({ fixture, restartDevice: true }, async () => {
```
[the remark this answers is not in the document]
vinistevam: Yes, because the app is started at `init.js` using the `beforeAll` hook so every time we want to provide fixture to support our test the app needs to reload in order to obtain the new fixture provided specifically to support the test.
One way to avoid it is to start the app using `beforeAll` hook in the test file or using `withFixtures` function with the option `restartDevice` set to `true`. It starts the app if the app is not running.

e2e/viewHelper.js
vinistevam: I was trying to avoid disruption here because the password on the fixtures account is different from the one in `Accounts.js`.  One thing to use `validAccount.password` for this case again is if I change the password in the fixtures `KeyringController` to match the one in `validAccount.password`.

app/components/Views/RevealPrivateCredential/RevealPrivateCredential.tsx
vinistevam: No, It acts as a wrapper around the AsyncStorage module from the `@react-native-async-storage/async-storage` package. The purpose of this class is to provide a unified interface for storing, retrieving, and removing data using the AsyncStorage module. 

Using the wrapper I can provide the fixtures to async storage in a central place.

package.json
vinistevam: Sorry for my confusing here. I changed it to `test:e2e:ios:debug` and `test:e2e:android:bitrise:debug`.
I'm not sure if I also should include `IS_TEST='true'` to  `test:e2e:android:debug`. 
The goal here is no need to add environment variables through the Bitrise UI.

e2e/init.js
vinistevam: To clarify here it's just loading an "empty fixture" as the infrastructure of the fixtures relies on the env `IS_TEST` once it's `true` it will try to load the fixtures and by providing this "empty fixture" which to simplify is the same as it was before going through all the onboarding process.

I thought the fixtures were meant to be used on almost all tests, in the example I've created using the fixtures to skip the onboarding process it can be applied to all the tests besides the flows to test specifically the onboarding flows.

cortisiko (changes requested):

cortisiko (dismissed):
PR looks good!

cortisiko (approved):

matthewwalsh0 (approved):

vinistevam:
@SocketSecurity ignore-all

OGPoyraz (approved):

Andepande (approved):
LGTM

PR #6557, opened 2023-06-07
test: add signing e2e tests

matthewwalsh0:
**Description**

Create Detox E2E tests to verify signing messages via the test dApp.

**Checklist**


OGPoyraz (dismissed):
LGTM

e2e/specs/confirmations/sign-messages.spec.js
bergarces, line 1: I'm guessing we are waiting for better tools before we do extra assertions to check that they are really returning the signature and not erroring?

bergarces (dismissed):

e2e/specs/confirmations/sign-messages.spec.js
matthewwalsh0: Exactly, as soon as Detox can interact with the browser elements, we can verify the text itself from the output.

e2e/specs/confirmations/sign-messages.spec.js
```diff
@@ -7,8 +7,15 @@ import { TestDApp } from '../../pages/TestDApp';
 import SettingsView from '../../pages/Drawer/Settings/SettingsView';
 import AdvancedSettingsView from '../../pages/Drawer/Settings/AdvancedView';
 import { Smoke } from '../../tags';
+import TestHelpers from '../../helpers';
+
+const MAX_ATTEMPTS = 3;
 
 describe(Smoke('Sign Messages'), () => {
+  beforeAll(async () => {
+    jest.setTimeout(150000);
```
bergarces, line 16: Do we need so much time?

bergarces (approved):

e2e/specs/confirmations/sign-messages.spec.js
matthewwalsh0: This was standard on every other smoke test I believe so I just added for consistency.

PR #6560, opened 2023-06-07
fix: Fix header artifact when transitioning to Import screens

Cal-L:
**Description**

This fix addresses an artifact that shows when navigating to both Import tokens and Import NFT screens. This same header is used when navigating to the Asset screen as well, which is responsible for the header gradient animation.

**Screenshots/Recordings**

This video shows a smooth transition to both the Import screen and Asset screen. It also shows that the gradient animation still works while scrolling on the Asset screen.

https://github.com/MetaMask/metamask-mobile/assets/10508597/6dc4acbb-e3d5-46ce-ba47-7a199a937d52


tommasini (approved):
LGTM!

PR #6562, opened 2023-06-08
test: adding hooks on confirmations e2e wdio and fix failing test

seaona:
## Description
In this PR we introduce the use of wdio hooks for running some background actions/servers (ganache and mocks) before and after Scenarios and removing them from the steps feature files.

This will help as reduce the steps on feature files as well as leverage the conditional tagging for enable/disable background servers/actions in an easier way.

**Note:** Approve ERC20 test is failing due to a change in the Approve flow. We (confirmations) still miss another change in the UI requested by product, so test will be updated after the final flow is working on the app.

## Considerations
1. `beforeScenario` runs before each scenario, and holds during that specific scenario. This has implications in the organization of feature files, as we need to write all the needed steps for that test under the same scenario, otherwise background processes will be stopped after the scenario is over.

2. Cucumber `world` is accessible and set for each scenario and reset after the scenario terminates. This has the same implications as the above point. We need to write all the test steps under the same scenario if we want to access world variables.
On the `beforeScenario` hook from wdio, we can access cucumber world using `context`.

3. We cannot use hooks with `Scenario Outlines` meaning we should always use `Scenario` for tests that use hooks.

## Screenshots

More info about cucumber world state [here](https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/world.md)

image

More info about wdio including world (context) in their hooks, which is then passed to cucumber

https://github.com/webdriverio/webdriverio/issues/7651

**Issue**

Progresses #6574  

**Checklist**

### DIFF
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -4,6 +4,12 @@ dotenv.config({ path: '.e2e.env' });
 import generateTestReports from './wdio/utils/generateTestReports';
 import ADB from 'appium-adb';
 import { gasApiDown, cleanAllMocks } from './wdio/utils/mocks';
+import {
+  startGanache,
+  stopGanache,
+  deployMultisig,
+  deployErc20,
+ } from './wdio/utils/ganache';
 const { removeSync } = require('fs-extra');
 
 export const config = {
@@ -272,7 +278,6 @@ export const config = {
       return capabilities.platformName;
     };
     const adb = await ADB.createADB();
-    await adb.reversePort(8000, 8000);
     await adb.reversePort(8545, 8545);
   },
   /**
@@ -296,9 +301,26 @@ export const config = {
    * @param {ITestCaseHookParameter} world    world object containing information on pickle and test step
    * @param {Object}                 context  Cucumber World object
    */
-  beforeScenario: ({tags: '@gasApiDown'}, async function (world, context) {
-    context.mock = gasApiDown();
-  }),
+  beforeScenario: async function (world, context) {
+    const tags = world.pickle.tags;
+
+    if(tags.filter(e => e.name === '@ganache').length > 0){
+      await startGanache();
+    }
+
+    if(tags.filter(e => e.name === '@multisig').length > 0){
+      const multisig = await deployMultisig();
+      context.multisig = multisig
+    }
+
+    if(tags.filter(e => e.name === '@erc20').length > 0){
+      context.erc20 = await deployErc20();
+    }
+
+    if(tags.filter(e => e.name === '@gasApiDown').length > 0){
+      context.mock = gasApiDown();
+    }
+  },
   /**
    *
    * Runs before a Cucumber Step.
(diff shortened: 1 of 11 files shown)


wdio.conf.js
```diff
@@ -296,9 +301,26 @@ export const config = {
...
+  beforeScenario: async function (world, context) {
+    const tags = world.pickle.tags;
+
+    if(tags.filter(e => e.name === '@ganache').length > 0){
+      await startGanache();
+    }
+
+    if(tags.filter(e => e.name === '@multisig').length > 0){
+      const multisig = await deployMultisig();
+      context.multisig = multisig;
```
seaona, line 313: we pass multisig address we just deployed to the cucumber `world` so we can access it later in the scenario steps

wdio.conf.js
```diff
@@ -296,9 +301,26 @@ export const config = {
    * @param {ITestCaseHookParameter} world    world object containing information on pickle and test step
    * @param {Object}                 context  Cucumber World object
    */
-  beforeScenario: ({tags: '@gasApiDown'}, async function (world, context) {
-    context.mock = gasApiDown();
-  }),
+  beforeScenario: async function (world, context) {
```
seaona, line 304: we use conditional hooks depending on the tags for each scenario

wdio.conf.js
```diff
@@ -331,9 +353,17 @@ export const config = {
    * @param {number}                 result.duration  duration of scenario in milliseconds
    * @param {Object}                 context          Cucumber World object
    */
-  afterScenario: ({tags: '@mock'}, async function (world, result, context) {
-    cleanAllMocks();
-  }),
+  afterScenario: async function (world, context) {
```
seaona, line 356: tear down

wdio/step-definitions/send-flow.steps.js
```diff
@@ -27,7 +27,7 @@ Then(/^I tap the Save button/, async () => {
 
 Given(
   /^I enter address "([^"]*)?" in the sender's input box/,
-  async (address) => {
+  async function (address) {
```
seaona, line 30: we need to bind `this` for accessing cucumber `world` , so we cannot use arrow functions

wdio/features/Confirmations/ApproveERC20.feature
```diff
@@ -2,25 +2,21 @@
 @confirmations
 @regression
 
-Feature: Approving an ERC20 token from a dapp
+Feature: Approve an ERC20 token from a dapp
   A user should be able to approve an ERC20 token from a dapp.
-
-  Scenario: Import wallet
+  @ganache
+  @erc20
```
seaona, line 8: we use tags which will trigger the desired hooks for that particular scenario

wdio/features/Confirmations/ApproveERC20.feature
```diff
@@ -2,25 +2,21 @@
...
     Given the app displayed the splash animation
     And I have imported my wallet
     And I tap No Thanks on the Enable security check screen
     And I tap No thanks on the onboarding welcome tutorial
-  
-  Scenario: Approve an ERC20 token from a dapp with default amount
-    Given Ganache server is started
     And I close the Whats New modal
     And Ganache network is selected
-    And ERC20 token contract is deployed
```
seaona, line 18: we can remove background steps as they are now executed with the hooks

wdio/features/Confirmations/ApproveERC20.feature
```diff
@@ -2,25 +2,21 @@
...
     When I navigate to the browser
     And I am on Home MetaMask website
     And I navigate to "test-dapp-erc20"
     And I connect my active wallet to the test dapp
     And I scroll to the ERC20 section
     And I approve ERC20 tokens
-    Then the transaction is submitted with Transaction Complete! toast appearing
-    Then Ganache server is stopped
```
seaona, line 26: tear down also happens in the hooks level now

SamuelSalas (approved):

segun (approved):

PR #6596, opened 2023-06-13
feat: Memoise token balance controler hook

NicolasMassart:
## Development & PR Process

## Description

Create a hook that memoises the state request for token balances from `TokenBalancesController`.
This hook uses a deep object equality check to replace the default shallow comparison of `useSelector`.
Shallow comparison is not fitting the need for the complex data type we use here and it ends in some unwanted re-rendering or render misses due to data not being properly evaluated. 

Need for performance improvement that prevents re-rendering when data hasn't deeply changed.

- Use deep data comparison with `isEqual` form Lodash lib on `useSelector` dependencies to trigger a new state retrieval.
- Replace the `useSelector` call with this new hook in `app/components/UI/Tokens/index.tsx`
- Implement the hook in order to start using the `{data,loading,retry,error}` return object pattern. Currently only `{data}` is implemented as the others require some deep controller changes out of the scope of this PR. This is a first step toward generalised controller hooks.
- `{data,loading,retry,error}` return value type interface created with optional loading, retry and error as they are not yet implemented on controlers. But leads the way for future consistent use.
- Implement tests that ensure the data is only returned when changed to prevent triggering of unnecessary render.

## Issue

fixes https://github.com/MetaMask/mobile-planning/issues/808

## QA test builds

iOS and Android builds available at https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/d0e62919-1ca8-4518-887c-0cf96b2afea3

## Recording

https://github.com/MetaMask/metamask-mobile/assets/4677568/107a0bb4-523c-42a6-89b0-51ecda5a8695

## Checklist


app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    // check that the initial balances are returned
+    expect(result.current.data).toStrictEqual(initialBalances);
+
+    // update token balances with identical object but different reference
+    mockState.engine.backgroundState.TokenBalancesController.contractBalances =
+      lodash.cloneDeep(initialBalances);
+
+    // rerender the hook to get the new balances
+    rerender({ wrapper: Wrapper });
```
Cal-L, line 98: You might be able to wrap this in `act` or `waitFor` then place the assertion inside. One or the other waits for a re-render

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  afterEach(() => {
+    jest.clearAllMocks();
+    renderResult.unmount();
+  });
+
+  it('should return stored data when state changed', async () => {
+    const { result, rerender } = renderResult;
+
+    // useSelector hook equality check returns true as the storage state haven't changed yet
+    expect(spyOnIsEqual).toHaveNthReturnedWith(1, true);
```
Cal-L, line 56: Could you elaborate why the spy on lodash is needed?

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    // check that the initial balances are returned
+    expect(result.current.data).toStrictEqual(initialBalances);
+
+    // update token balances with identical object but different reference
+    mockState.engine.backgroundState.TokenBalancesController.contractBalances =
```
Cal-L, line 94: lodash cloneDeep might not simulate that a new reference is being created with the same data set. You can just provide it with a new object and spread the old data.

Cal-L (commented):
left some comments

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
NicolasMassart: I removed this completely

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
NicolasMassart: Done, and indeed it looks far better. thanks!

app/components/hooks/useTokenBalancesController/useTokenBalancesController.ts
```diff
@@ -0,0 +1,15 @@
+import { useSelector } from 'react-redux';
+import { EngineState } from '../../../selectors/types';
+import { isEqual } from 'lodash';
+
+const useTokenBalancesController = () => {
+  const tokenBalances = useSelector(
+    (state: EngineState) =>
+      state.engine.backgroundState?.TokenBalancesController?.contractBalances,
```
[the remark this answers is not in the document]
NicolasMassart: Will other data have to be returned by this specific controller hook or do you mean that we will have other data returned by other controller hooks? I thought this specific hook was expected to only deal with the token balances.

app/components/hooks/useTokenBalancesController/useTokenBalancesController.ts
```diff
@@ -0,0 +1,15 @@
+import { useSelector } from 'react-redux';
+import { EngineState } from '../../../selectors/types';
+import { isEqual } from 'lodash';
+
+const useTokenBalancesController = () => {
```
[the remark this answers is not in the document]
NicolasMassart: added an interface that can be common for other controller hooks.
Only data is typed as required for now so we can still use it while we don't have retry, errors and loading returns implemented.

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+describe('useTokenBalancesController()', () => {
+  it('should only rerender when state changed', async () => {
+    // render the component with the test store
+    const { toJSON } = render(<DummyTestComponent />, { wrapper: Wrapper });
+
+    // check the useEffect is called once and initial balances are returned
+    await waitFor(async () => {
+      // check that the initial balances are returned
+      expect(spyOnDummyUseEffect).toHaveBeenCalledTimes(1);
+      expect(toJSON()).toMatchSnapshot();
```
Cal-L, line 69: No need for checking snapshots within these tests since it's just a dummy component. Let's split the re-render and non-rerender test into separate tests though to make it a bit cleaner.

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+// test reducer for the test store
+const testBalancesReducer = (state: any, action: any) => {
+  if (action.type === 'add-balances') {
+    const newBalances = {
+      ...state.engine.backgroundState.TokenBalancesController.contractBalances,
+      ...action.value,
+    };
+    state.engine.backgroundState.TokenBalancesController.contractBalances =
+      newBalances;
+    return {
```
Cal-L, line 31: I don't think `value` resembles the shape of the actual shape of what TokenBalancesController contains. We can use remove line 33

Cal-L (commented):
left some comments

app/components/hooks/useTokenBalancesController/useTokenBalancesController.test.tsx
NicolasMassart: I'm checking snapshot to check the result is the expected token balances and the hook reliably returned the right data. Not to test the component itself as it's indeed a dummy. It looked easier to just check against the whole rendered result with a snapshot than querying an element in the rendered component. That way we test the effect triggered and the content of the returned balances.

Cal-L (approved):
LGTM

cortisiko:
hey @NicolasMassart ! Great work on getting this done. Can you add recordings of what you tested?

NicolasMassart:
@cortisiko I added recording and a link to Bitrise QA builds

cortisiko (approved):
This is QA passed! Feel free to merge whenever @NicolasMassart

PR #6600, opened 2023-06-14
refactor: accept SignController approval request from frontend

OGPoyraz:
**Development & PR Process**

**Description**

This PR aims to remove the `sign` and `cancel` methods usage from `SignatureController` and use promisified middlewares named `newUnsignedXMessage` methods.

Patched PR detail could be seen here: https://github.com/MetaMask/core/pull/1377 (It's merged now)

**Manual Testing Steps**

- Go to [test-dapp](https://metamask.github.io/test-dapp/) 
- Use any sign action, or cancel sign action without any issues. 

**Recordings**
[sign_msgs_tests.webm](https://github.com/MetaMask/metamask-mobile/assets/45455812/bffd7617-4edc-46b7-b853-903d6a13651a)

**Bitrise E2E pipeline**
https://app.bitrise.io/build/30379484-ceba-4d17-b1a1-97552d5fac68
(pr_smoke_e2e_pipeline) https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/16a00d66-5006-4204-b682-1337eb051e49 
July 20th - https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/8969daf3-08ee-4e71-8990-2dc5f5ce2eeb

**Issue**
https://github.com/MetaMask/mobile-planning/issues/885

**Checklist**

* [X] Tests are included if applicable (e2e tests are already covering this signatures)


app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -401,7 +383,7 @@ const RootRPCMethodsUI = (props) => {
     setShowPendingApproval(false);
     rejectPendingApproval(
       walletConnectRequestInfo.id,
-      ethErrors.provider.userRejectedRequest(),
+      serializeError(ethErrors.provider.userRejectedRequest()),
```
matthewwalsh0, line 386: What is the main driver here to serialize? Is it just to reduce the complexity of the error object that gets passed around?

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -654,6 +639,15 @@ const RootRPCMethodsUI = (props) => {
   };
 
   const onSign = () => {
+    resolvePendingApproval(signMessageParams.metamaskId);
+    setSignMessageParams(undefined);
+  };
+
+  const onCancel = () => {
```
matthewwalsh0, line 646: Is it best we have a more unique name such as `onSignReject` to match the format of the other modals?

We could even rename `onSign` to `onSignConfirm` to be extra explicit.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -654,6 +639,15 @@ const RootRPCMethodsUI = (props) => {
   };
 
   const onSign = () => {
+    resolvePendingApproval(signMessageParams.metamaskId);
```
matthewwalsh0, line 642: I appreciate the approval ID and the `metamaskId` are currently the same value, but is it safer and more explicit to do what the other modals do and store the actual approval request ID in the state also?

It would avoid any future errors if we stop using the message ID as the approval ID for example.

app/components/UI/MessageSign/index.js
```diff
@@ -124,16 +125,18 @@ class MessageSign extends PureComponent {
 
   signMessage = async () => {
     const { messageParams } = this.props;
-    const { SignatureController } = Engine.context;
-    await SignatureController.signMessage(messageParams);
+    const { resolvePendingApproval } = Engine;
+    resolvePendingApproval(messageParams.metamaskId);
```
matthewwalsh0, line 129: As the previous logic used an `await` to wait for the signing to finish, this behaviour is now subtly different as `resolvePendingApproval` will return instantly and not throw any errors from the signing for example.

One fix is to use version `3.1.0` of the approval controller, along with the `waitForResult` option to wait until the `SignatureController` has indicated a result via the result callbacks. Unfortunately this would require updating the `SignatureController` to create the approval request with the `expectsResult` option and to call `success` or `error` accordingly on the returned `resultCallbacks`. This is what is done for both of the transaction controllers.

matthewwalsh0: Another side effect to look out for is that when we accept the pending approval, it is deleted meaning the root UI component will detect the approval controller state change and close all modals by setting `showPendingApproval` to false as there is no pending approvals.

We likely don't want this as the current behaviour was to keep the modal open while the signing takes place, so we can conditionally close based on the current pending approval type and let the signing modal close itself when finished.

This is also addressed in the `TransactionController` PR [here](https://github.com/MetaMask/metamask-mobile/pull/6641/files#diff-a00fa937658e508f6da02d5b6e07f28e1601bd2317e40a5be549ba9ad20aa17dR775).

app/core/Engine.ts
```diff
@@ -785,6 +782,27 @@ class Engine {
     this.controllerMessenger.clearSubscriptions();
   }
 
+  resolvePendingApproval = (id, data) => {
```
matthewwalsh0, line 785: Is it more explicit to keep this as `acceptPendingApproval` as it's more clear what it's doing with the pending approval?

matthewwalsh0: As mentioned in the other comment, this may have to become `async` and have an optional argument to support `waitForResult`.

app/components/Nav/Main/RootRPCMethodsUI.js
OGPoyraz: Originally it's been used like that in extension so I keep it. Basically serializes the given error into JSON rpc-compatible object. But the provided error here is already an RPC error so no big deal. 

I still find it useful to have it in code, so in future if someone adds an incompatible error, it tells them that they need to serialize.

app/components/Nav/Main/RootRPCMethodsUI.js
OGPoyraz: Changed all to `onSignReject` and `onSignConfirm` until it hit the signature components 👍

app/components/Nav/Main/RootRPCMethodsUI.js
OGPoyraz: I've didn't notice the convention up there, totally make sense to keep it the same. 
Fixed.

app/components/UI/MessageSign/index.js
```diff
@@ -124,16 +125,18 @@ class MessageSign extends PureComponent {
...
 
   rejectMessage = async () => {
     const { messageParams } = this.props;
-    const { SignatureController } = Engine.context;
-    const messageId = messageParams.metamaskId;
-    await SignatureController.cancelMessage(messageId);
+    const { rejectPendingApproval } = Engine;
+    rejectPendingApproval(
+      messageParams.metamaskId,
+      serializeError(ethErrors.provider.userRejectedRequest()),
```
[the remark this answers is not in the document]
OGPoyraz: Removed 👍

app/core/Engine.ts
```diff
@@ -785,6 +782,27 @@ class Engine {
...
+    const { ApprovalController } = this.context;
+    try {
+      ApprovalController.accept(id, data);
+    } catch (err) {
+      // Ignore err if request already approved or doesn't exists.
```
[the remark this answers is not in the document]
OGPoyraz: I thought this is added to not block users if unexpected thing happens while approving or rejecting (as you said bad approval ids). Because otherwise, we won't be able to close popup via `setSignMessageParams(undefined);` in `RootRPCUIMethods.js`, it will never get executed. 

I agree this is not entirely correct for `accept` case, because it will directly close modal without notifying user. But it's not harmful for `rejecting`. Could we create a future task to show some notification in this case?

app/core/Engine.ts
```diff
@@ -785,6 +782,27 @@ class Engine {
...
+
+  rejectPendingApproval = (id, error) => {
+    const { ApprovalController } = this.context;
+    try {
+      ApprovalController.reject(
+        id,
+        new EthereumRpcError(error.code, error.message, error.data),
+      );
+    } catch (error) {
+      Logger.error(error, 'Reject while rejecting pending connection request');
```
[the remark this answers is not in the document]
OGPoyraz: Sorry if it's my misunderstanding, error was included at the beginning of the logger, are you asking that part? Also 
will fix message here.

app/components/UI/PersonalSign/PersonalSign.tsx
```diff
@@ -88,44 +88,22 @@ const PersonalSign = ({
...
   const confirmSignature = async () => {
-    try {
-      await signMessage();
-      AnalyticsV2.trackEvent(
-        MetaMetricsEvents.SIGN_REQUEST_COMPLETED,
-        getAnalyticsParams(),
-      );
-      onConfirm();
-    } catch (e: any) {
-      if (e?.message.startsWith(KEYSTONE_TX_CANCELED)) {
```
OGPoyraz, line 121: This case should be catched and triggered from middleware function

app/components/UI/MessageSign/index.js
OGPoyraz: First one fixed, second c/p'ed from your PR.

app/core/Engine.ts
OGPoyraz: Renamed all to `acceptPendingApproval`

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -658,15 +633,22 @@ const RootRPCMethodsUI = (props) => {
     );
   };
 
-  const onSign = () => {
+  const onSignConfirm = async () => {
+    await acceptPendingApproval(signMessageParams.id);
```
matthewwalsh0, line 637: In addition to the `await`, we need to pass the `waitForResult` argument to `ApprovalController.accept` to ensure it won't resolve until the approval request creator has indicated if it worked or not. This is only needed for the signing accept so can be an argument in the new `acceptPendingApproval` in `Engine`. See [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/Transactions/index.js#L505) for an example.

Unfortunately this means we'll have to re-release `SignatureController` with changes to call `success` or `error` on the `resultCallbacks`. See [here](https://github.com/MetaMask/core/blob/main/packages/transaction-controller/src/TransactionController.ts#L1186) for an example.

app/core/Engine.ts
```diff
@@ -785,6 +782,27 @@ class Engine {
     this.controllerMessenger.clearSubscriptions();
   }
 
+  acceptPendingApproval = async (id, data) => {
+    const { ApprovalController } = this.context;
+    try {
+      await ApprovalController.accept(id, data);
+    } catch (error) {
+      Logger.error('Error while approving approval request', error);
```
matthewwalsh0, line 790: I'd drop this to a `debug` so we don't scare anyone or accidentally create a full error alert which may be the default.

OGPoyraz: Fixed

vinistevam:
Bitrise E2E pipeline
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/16a00d66-5006-4204-b682-1337eb051e49

matthewwalsh0 (dismissed):

matthewwalsh0 (dismissed):

app/components/UI/MessageSign/MessageSign.tsx
matthewwalsh0, line 1: Just for audit sake, what was the driver to change this to TypeScript?

Am I right there is no code changes in here?

app/core/Engine.ts
```diff
@@ -799,10 +798,14 @@ class Engine {
     }
   }
 
-  acceptPendingApproval(id: string, requestData?: Record<string, Json>) {
+  acceptPendingApproval(
+    id: string,
+    requestData?: Record<string, Json>,
+    opts: AcceptOptions = { waitForResult: false },
```
matthewwalsh0, line 804: Very minor, but I don't think the default is needed here as this is also the default in the `ApprovalController` itself.

Could we instead make the `opts` argument optional?

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -145,6 +145,59 @@ export const checkActiveAccountAndChainId = async ({
   }
 };
 
+const generateRawSignature = async ({
```
matthewwalsh0, line 148: Would it be more explicit to call this `newUnsignedTypedMessage` as it's only for typed messages?

app/util/confirmation/signatureUtils.js
matthewwalsh0, line 1: If this is a new file, should it be TypeScript?

matthewwalsh0, line 1: Should we instead name this file `signature` as util is implied from the path?

matthewwalsh0, line 1: As this is new, do we require unit tests for these methods or are they already tested in the components for now?

package.json
```diff
@@ -360,9 +360,9 @@
     "@testing-library/react-native": "12.1.2",
     "@types/enzyme": "^3.10.9",
     "@types/is-url": "^1.2.30",
+    "@types/qs": "^6.9.7",
```
matthewwalsh0, line 363: Should we revert this to keep the list alphabetical?

matthewwalsh0 (dismissed):
Added some minor suggestions but none of them functional so perhaps best to do in a separate PR given the size and age of this PR.

app/components/UI/MessageSign/MessageSign.tsx
vinistevam: Sonar was blocking this PR to be merged due to duplicate code. I've tried to work around creating the `signatreUtils` but I was not able to reduce until the percentage that satisfies sonar. I think those components could be refactored using Higher-Order Component but I tried to avoid those sorts of big changes.

app/util/confirmation/signatureUtils.js
vinistevam: They are already tested in the component, I think it's >85% coverage but I created a ticket with the remaining actions and one is to extract the tests so we have the same level.

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

vinistevam (approved):

PR #6619, opened 2023-06-15
fix: Token allowance flow update

blackdevelopa:
**Development & PR Process**

**Description**
* [x] Approve screen should be pre-filled with the dapp suggested spending cap, instead of having a blank input field.
* [x] Users should be able to click on the value to change the spending cap
* [x] Users should be able to click on "Next" to go to the second screen with the gas information
* [x] Learn More link should be added below the copy. Link should direct to https://support.metamask.io/hc/en-us/articles/6055177143579-How-to-customize-token-approvals-with-a-spending-cap.
* [x] Custom spending cap  should be updated to `Spending cap`
* [x] Set a spending cap for your and Review the spending cap for your should be updated to Spending cap request for your

**Screenshots/Recordings**

https://github.com/MetaMask/metamask-mobile/assets/29962968/25a82d01-0676-404e-8b51-386f0020d156

**Issue**

Progresses https://github.com/MetaMask/MetaMask-planning/issues/685

**Checklist**


app/component-library/components-temp/CustomSpendCap/CustomSpendCap.tsx
```diff
@@ -223,7 +236,12 @@ const CustomSpendCap = ({
       {!isEditDisabled && (
         <View style={styles.descriptionContainer}>
           <Text variant={TextVariant.BodyMD} style={styles.description}>
-            {message}
```
jpuri, line 226: `{''}` can be removed I think

jpuri (dismissed):
🚢

app/constants/urls.ts
```diff
@@ -55,3 +55,5 @@ export const MM_PLAY_STORE_LINK = `market://details?id=${AppConstants.BUNDLE_IDS
 
 // SDK
 export const MM_SDK_DEEPLINK = `https://${AppConstants.MM_UNIVERSAL_LINK_HOST}/connect?`;
+
+export const TOKEN_APPROVAL_SPENDING_CAP = `https://support.metamask.io/hc/en-us/articles/6055177143579-How-to-customize-token-approvals-with-a-spending-cap`;
```
digiwand, line 59: [nit] can be moved under `// SUPPORT` section

app/component-library/components-temp/CustomSpendCap/CustomSpendCap.tsx
digiwand: ++ hm still seems to be here even though this was marked as resolved. what is this used for?

digiwand (dismissed):
no blockers. minor comments above

app/constants/urls.ts
blackdevelopa: Updated. Thank you. https://github.com/MetaMask/metamask-mobile/pull/6619/commits/e245dfec9eb5275a2cdffda4d97f8647af7f7592

app/component-library/components-temp/CustomSpendCap/CustomSpendCap.tsx
blackdevelopa: Oh, this was discussed on a call. The reason is because we are wrapping a button link component in a text.

digiwand: oh. I still don't understand what {''} is used for in the context of wrapping the button link component in a text.  

Interesting. Actually, I think internationalization may be broken or would be hard to set up with the current use of translations in this file because the sentence ordering may change between languages.

digiwand: I see now. `{message} {''}` seems to be equavalent to `{message}{' '}`. Thanks for the explanation @blackdevelopa! 

I have a slight concern about how this renders with right-to-left languages. Has this been verified as well?

bschorchit:
Updated copy changes:
1. Copy for top right link should be`Use site suggestion` instead of `Use third party suggestion`
Screenshot 2023-06-20 at 13 26 29

3. Copy for when amount is higher than balance should be: `This allows the third party to spend all your token balance until it reaches the cap or you revoke the spending cap. If this is not intended, consider setting a lower spending cap. Learn more`
Screenshot 2023-06-20 at 13 27 23

2. Copy for when there's an input error or for when field is empty should be: `Only enter a number that you're comfortable with the third party spending now or in the future. You can always increase the spending cap later. Learn more` as suggested by Mariona here https://github.com/MetaMask/metamask-mobile/pull/6634#issuecomment-1598371089
Screenshot 2023-06-20 at 13 32 47

app/component-library/components-temp/CustomSpendCap/CustomSpendCap.tsx
blackdevelopa: Oh great. I have now updated to `{message}{' '}` so it's more aligned. I think for RTL, we can always verify this in the [locales](https://github.com/MetaMask/metamask-mobile/tree/main/locales/languages)

app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -1040,13 +1044,14 @@ class ApproveTransactionReview extends PureComponent {
...
-    const { showBlockExplorerModal, address } = this.state;
+    const { showBlockExplorerModal, address, learnMoreURL } = this.state;
 
     const styles = this.getStyles();
     const closeModal = () => {
-      showVerifyContractDetails();
+      !learnMoreURL && showVerifyContractDetails();
       this.setState({
         showBlockExplorerModal: !showBlockExplorerModal,
+        learnMoreURL: undefined,
```
digiwand, line 1054: [nit] I think null would be preferred here since we are setting the value 
```suggestion
        learnMoreURL: null,
```

blackdevelopa: Done. https://github.com/MetaMask/metamask-mobile/pull/6619/commits/082a6f9ade63b0530ac308b728db7074fcd167e7

digiwand (approved):
🙌

jpuri (approved):

seaona:
The PR looks good @blackdevelopa . I can see how the new Approve flow has been updated with the specs above.
I have encountered one issue: whenever the user sets a custom value, if clicks `View transaction details` the value is changed back to the default dapp value. This same behaviour also happens if you click `Edit` - you will see that the value is again the default by the dapp instead of the one the user filled.

https://github.com/MetaMask/metamask-mobile/assets/54408225/add4de94-55ab-400f-9103-0f35548072d2

Since this is a pretty big PR, maybe the PR could be merged and we could fix that in a separate issue.
What do you think @blackdevelopa @bschorchit ?

bschorchit:
I agree with fixing in a different PR ✅

seaona:
@blackdevelopa from QA rebase looks good. I think we need x2 confirmations from @jpuri and @digiwand that the last changes also look good from dev side before merging :pray:  thank you!!

jpuri (approved):

digiwand (approved):

PR #6621, opened 2023-06-15
E2e/655 remove drawer

SamuelSalas:
**Description**
Update Appium and Detox test scripts and related files to run with the 655-remove-drawer changes.

**Issue**

Progresses #???

**Checklist**


e2e/specs/delete-wallet.spec.js
```diff
@@ -123,11 +120,9 @@ describe.skip(
...
-      await WalletView.tapDrawerButton();
-
-      await DrawerView.isVisible();
-      await DrawerView.tapLockAccount();
-      await DrawerView.tapYesAlertButton();
+      await TabBarComponent.tapActions();
+      await SettingsView.tapLock();
+      await SettingsView.tapYesAlertButton();
       await LoginView.isVisible();
     });
```
NicolasMassart, line 127: Q: can you explain what this _'should open drawer and log out'_ test actually does? It's not clear to me as I tried to follow the steps to reproduce manually but get lost at await SettingsView.tapCloseButton(); as there's no close button on the settings view for me and we have to scroll the settings to reach the Lock entry.
Also I feel that it should be renamed as 'should open drawer and log out' is not right as there's no drawer anymore. (and by the way the fox icon does nothing for what I noticed, code says it's popping back in the nav stack but I can't see where it can actually take effect)

wdio/screen-objects/SettingsScreen.js
```diff
@@ -0,0 +1,20 @@
...
+import Gestures from '../helpers/Gestures';
+
+class SettingsScreen {
+  get lockOption() {
+    return Selectors.getElementByPlatform(LOCK_SETTINGS);
+  }
+
+  async tapLockOption() {
+    const lockOption = await this.lockOption;
+    while (!(await lockOption.isDisplayed())) {
```
NicolasMassart, line 12: Nice! Better than scrolling % !

wdio/screen-objects/testIDs/Screens/EditGasFeeScreen.testids.js
```diff
@@ -1,2 +1,2 @@
-export const EDIT_PRIOTIRY_SCREEN_TEST_ID = 'edit-priority-screen';
-export const MAX_PRIORITY_FEE_INPUT_TEST_ID = 'max-priority-fee-range-input';
\ No newline at end of file
+export const EDIT_PRIORITY_SCREEN_TEST_ID = 'edit-priority-screen';
```
NicolasMassart, line 1: Thanks for noticing these typos!

e2e/specs/delete-wallet.spec.js
SamuelSalas: Hi Nico, I was not able to test that specific spec file as at this moment right now is getting skip. That might be the reason that specific file at this moment is not following the latest changes. I'll update it, thanks.

wdio/screen-objects/testIDs/Screens/EditGasFeeScreen.testids.js
SamuelSalas: Thanks for noticing, always trying to fix those kind of issues in every pr if possible.

wdio/screen-objects/SettingsScreen.js
SamuelSalas: Thanks, this solution is making the scrolling more consistent in the setting screen.

e2e/pages/Drawer/Settings/SettingsView.js
```diff
@@ -1,21 +1,37 @@
...
+  LOCK_SETTINGS,
+  NETWORKS_SETTINGS,
+  SECURITY_SETTINGS,
+} from '../../../../wdio/screen-objects/testIDs/Screens/Settings.testIds';
 
 const ANDROID_BACK_BUTTON_ON_SETTINGS_PAGE_ID = 'nav-android-back';
 export default class SettingsView {
   static async tapGeneral() {
-    await TestHelpers.tapByText('General');
+    await TestHelpers.waitAndTap(GENERAL_SETTINGS);
```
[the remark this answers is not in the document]
SamuelSalas: I tend to prefer test id mostly because of personal tradition. And from the appium side, relaying a lot on those text selectors create a lot of inconsistencies on the test runs. And to provide consistency with the new methods that i wrote.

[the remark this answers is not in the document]
NicolasMassart: Well there's two schools on this obviously. The test id should be a last resort to me only in case there's no way to get another selector. Using testIds ties your test to implementation and doesn't reflect the real user experience.
We favour not using test ids on the component integration test side. That would be interesting to join forces to define a common policy for all kind of tests.

e2e/specs/delete-wallet.spec.js
NicolasMassart: :Oh, right, haven't noticed the skip on the describe as not shown in the diff.

Andepande (approved):
LGTM

cortisiko (approved):
looks good @SamuelSalas

PR #6645, opened 2023-06-20
[Flask] bitrise pipelines

owencraston:
**Development & PR Process**

**Description**

- This allows us to run flask builds/releases on bitrise
- it will enable the smooth rollout of flask and should handle all the environment setup. 
- modify the build.sh to build flask on android/ios in release mode. For this I followed the exact patterns that were already in place for the qa/release versions of the app. 
- created yarn scripts to call our build.sh with the correct arguments
- created bitrise workflows and pipelines to create the builds for us
- created a release pipeline that takes the builds and releases it to the stores.
- updated the signing for MetaMask-Flask in xcode which modified the .xcodeproject file
- Modified the sentry environment variable to handle the different types of flask releases
- I was getting the error `"Invalid App Store Icon. The App Store Icon in the asset catalog in 'MetaMask-Flask.app' can't be transparent nor contain an alpha channel. (ID: 9e93a511-109f-43f3-b3a5-a80ee40734cd)";` when I tried to upload to testflight so I needed to modify the ios app icons to remove the alpha. 

Screenshot 2023-06-27 at 11 55 26 AM

## Test Builds
- https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/207891f7-53a6-4746-931e-dd2acb8f1fd1?tab=workflows
- latest one with all the correct artifacts: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/14942ea9-e976-43e5-93a2-3fd0fe87e12d

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/1044
Progresses https://app.zenhub.com/workspaces/metamask-accounts-team-62505028c3853700162f65e0/issues/gh/metamask/mobile-planning/1043

**Checklist**


scripts/checksum.sh
```diff
@@ -8,7 +8,12 @@ FILE=./android/app/build/outputs/apk/qa/release/app-qa-release.apk
   if test -f "$FILE"; then
     shasum -a 512 "$FILE" > ./android/app/build/outputs/apk/qa/release/sha512sums.txt
   fi
+elif [ "$MODE" == "Flask" ]; then
+FILE=./android/app/build/outputs/apk/flask/release/app-flask-release.apk
```
owencraston, line 12: Note: Is this the correct file?

scripts/build.sh
```diff
@@ -346,6 +367,32 @@ buildAndroidRelease(){
...
+	cd android && ./gradlew assembleFlaskRelease --no-daemon --max-workers 2
+
+	# GENERATE BUNDLE
+	if [ "$GENERATE_BUNDLE" = true ] ; then
+		./gradlew bundleFlaskRelease
+	fi
+
+	if [ "$PRE_RELEASE" = true ] ; then
+		# Generate sourcemaps
+		yarn sourcemaps:android
```
owencraston, line 386: do we need a different sourcemap for flask?

scripts/build.sh
```diff
@@ -106,6 +106,46 @@ checkParameters(){
...
+    echo "Successfully remapped $old_var_name to $new_var_name."
+}
+
+remapFlaskEnvVariables() {
+	# remap flask env variables to match what the app expects
+
+	echo "Remapping flask env variable names to match production"
+
+	# js.env variables
+	remapEnvVariable "MM_FLASK_PUBNUB_SUB_KEY" "MM_PUBNUB_SUB_KEY"
```
sethkfman, line 134: Remove PubNub keys

scripts/checksum.sh
sethkfman: should be

scripts/build.sh
```diff
@@ -106,6 +106,46 @@ checkParameters(){
...
+	remapEnvVariable "MM_FLASK_PUBNUB_PUB_KEY" "MM_PUBNUB_PUB_KEY"
+	remapEnvVariable "MM_FLASK_OPENSEA_KEY" "MM_OPENSEA_KEY"
+	remapEnvVariable "FLASK_MOONPAY_API_KEY_STAGING" "MOONPAY_API_KEY_STAGING"
+    remapEnvVariable "SEGMENT_FLASK_DEV_KEY" "SEGMENT_DEV_KEY"
+	remapEnvVariable "WALLET_CONNECT_FLASK_PROJECT_ID" "WALLET_CONNECT_PROJECT_ID"
+	remapEnvVariable "MM_FLASK_SENTRY_DSN" "MM_SENTRY_DSN"
+
+	# ios.env/ios.env variables
+	remapEnvVariable "MM_FLASK_BRANCH_KEY_TEST" "MM_BRANCH_KEY_TEST"
```
sethkfman, line 141: Branch Test shouldn't be needed

bitrise.yml
```diff
@@ -655,14 +664,55 @@ workflows:
           is_always_run: false
           is_skippable: true
           inputs:
-            - pipeline_intermediate_files: ios/build/output/MetaMask-QA.ipa:BITRISE_APP_STORE_IPA_PATH
-            - deploy_path: ios/build/output/MetaMask-QA.ipa
+            - pipeline_intermediate_files: ios/build/output/MetaMask.ipa:BITRISE_APP_STORE_IPA_PATH
+            - deploy_path: ios/build/output/MetaMask.ipa
```
sethkfman, line 668: This will be MetaMask-Flask.ipa

bitrise.yml
```diff
@@ -655,14 +664,55 @@ workflows:
...
           title: Deploy iOS IPA
+      - deploy-to-bitrise-io@1.6.1:
+          is_always_run: false
+          is_skippable: true
+          inputs:
+            - deploy_path: ios/build/MetaMask.xcarchive
```
sethkfman, line 674: Same name update here

bitrise.yml
```diff
@@ -655,14 +664,55 @@ workflows:
...
+              METAMASK_ENVIRONMENT='production' yarn build:ios:pre-flask
+          title: iOS Sourcemaps & Build
+          is_always_run: false
       - deploy-to-bitrise-io@2.2.3:
           is_always_run: false
           is_skippable: true
-          inputs:     
-            - deploy_path: ios/build/MetaMask-QA.xcarchive
+          inputs:
+            - pipeline_intermediate_files: ios/build/output/MetaMask.ipa:BITRISE_APP_STORE_IPA_PATH
```
sethkfman, line 708: same here

bitrise.yml
```diff
@@ -655,14 +664,55 @@ workflows:
...
+            - deploy_path: ios/build/output/MetaMask.ipa
+          title: Deploy iOS IPA
+      - deploy-to-bitrise-io@1.6.1:
+          is_always_run: false
+          is_skippable: true
+          inputs:
+            - deploy_path: ios/build/MetaMask.xcarchive
```
sethkfman, line 715: same here

bitrise.yml
```diff
@@ -671,6 +721,74 @@ workflows:
...
+                node -v
+                METAMASK_ENVIRONMENT='production' yarn build:android:pre-release:bundle:flask
+          title: Build Android Pre-Release Bundle
+          is_always_run: false
+      - save-gradle-cache@1: {}
+      - deploy-to-bitrise-io@2.2.3:
+          is_always_run: false
+          is_skippable: true
+          inputs:
+            - pipeline_intermediate_files: $PROJECT_LOCATION/app/build/outputs/apk/flask/release/app-flask-release.apk:BITRISE_PLAY_STORE_APK_PATH
```
sethkfman, line 757: This looks correct. I would just run to confirm

bitrise.yml
owencraston: I accidentally edited the QA steps here. I will remove these.

bitrise.yml
```diff
@@ -671,6 +680,115 @@ workflows:
...
+          title: iOS Sourcemaps & Build
+          is_always_run: false
+      - deploy-to-bitrise-io@2.2.3:
+          is_always_run: false
+          is_skippable: true
+          inputs:
+            - pipeline_intermediate_files: ios/build/output/MetaMask-Flask.ipa:BITRISE_APP_STORE_IPA_PATH
+            - deploy_path: ios/build/output/MetaMask-Flask.ipa
+          title: Deploy iOS IPA
+      - deploy-to-bitrise-io@1.6.1:
```
owencraston, line 711: @sethkfman I noticed that in the `build_ios_release` steps we use both `deploy-to-bitrise-io@2.2.3` and `deploy-to-bitrise-io@1.6.1`. What is the reason for this?

NicolasMassart: That's a very good question!
I think it's because we needed `pipeline_intermediate_files` that is not present on 1.6.1 and we only updated the steps where we needed it. All can very probably be upgraded to 2.2.3

bitrise.yml
```diff
@@ -202,7 +221,7 @@ workflows:
       - yarn@0:
           inputs:
             - command: build:announce
-          title: Accouncing pre-release
+          title: Announcing pre-release
```
NicolasMassart, line 224: thanks!

scripts/build.sh
```diff
@@ -106,6 +106,43 @@ checkParameters(){
 	fi
 }
 
+remapEnvVariable() {
```
NicolasMassart, line 109: Q: not sure to get why you do this remapping. Do we care about some Flask build script backward compatibility here?

scripts/build.sh
NicolasMassart: I understood the app source code will be different for Flask as it would include flask only features, right?

bitrise.yml
```diff
@@ -671,6 +690,115 @@ workflows:
             - pipeline_intermediate_files: sourcemaps/ios/index.js.map:BITRISE_APP_STORE_SOURCEMAP_PATH
             - deploy_path: sourcemaps/ios/index.js.map
           title: Deploy Source Map
+  build_ios_flask_release:
```
NicolasMassart, line 693: C: I wonder if we have a way to prevent duplication of these big yaml blocks and just have one called with params. they really mostly look the same.

scripts/build.sh
owencraston: the reason for this is so that for certain env variables, we can store a separate flask version in bitrise and remap them to the variable names that the code is looking for. This is much easier than changing all the code to check the build type and then change the variable name conjointly. This also makes it easier to not disturb the stable release. Not all of the env variables I have listed here will need remapping. I believe the ones we need are mixpanel, sentry and segment.

bitrise.yml
owencraston: @NicolasMassart regarding the sourcemap comment, yes they will be different. I am wondering if I need to change the sourcemap logic since it is failing [on bitrise](https://app.bitrise.io/build/06e5a589-468d-49c5-97a6-79a1fda0e021)

scripts/build.sh
NicolasMassart: Maybe adding some comment in the code then ;)

bitrise.yml
```diff
@@ -671,6 +690,115 @@ workflows:
...
+            - build_short_version_string: $FLASK_VERSION_NAME
+            - build_version: $FLASK_VERSION_NUMBER
+            - plist_path: $PROJECT_LOCATION_IOS/MetaMask/MetaMask-Flask-Info.plist
+      - cocoapods-install@2: {}
+      - script@1:
+          inputs:
+          - content: |-
+              #!/usr/bin/env bash
+              node -v
+              METAMASK_ENVIRONMENT='flask' yarn build:ios:pre-flask
```
owencraston, line 711: @sethkfman @Gudahtt I have set the `METAMASK_ENVIRONMENT` to `flask` here so that we can set the [sentry environment property](https://github.com/MetaMask/metamask-mobile/blob/1b283a2d6100749d8ca24ae77ad93b6e76e6499a/app/util/sentryUtils.js#L167) accordingly. Is this sufficient?

bitrise.yml
```diff
@@ -671,6 +690,115 @@ workflows:
...
+      - install-missing-android-tools@2:
+          inputs:
+            - ndk_revision: $NDK_VERSION
+            - gradlew_path: $PROJECT_LOCATION/gradlew
+      - script@1:
+          inputs:
+            - content: |-
+                #!/usr/bin/env bash
+                node -v
+                METAMASK_ENVIRONMENT='flask' yarn build:android:pre-release:bundle:flask
```
owencraston, line 759: Same here, `METAMASK_ENVIRONMENT` is set to flask here.

bitrise.yml
```diff
@@ -671,6 +690,115 @@ workflows:
...
+          is_skippable: true
+          inputs:
+            - pipeline_intermediate_files: $PROJECT_LOCATION/app/build/outputs/bundle/flaskRelease/app-flask-release.aab:BITRISE_PLAY_STORE_ABB_PATH
+            - deploy_path: $PROJECT_LOCATION/app/build/outputs/bundle/flaskRelease/app-flask-release.aab
+          title: Bitrise Deploy AAB
+      - deploy-to-bitrise-io@2.2.3:
+          is_always_run: false
+          is_skippable: true
+          inputs:
+            - pipeline_intermediate_files: /bitrise/src/sourcemaps/android/index.js.map:BITRISE_PLAY_STORE_SOURCEMAP_PATH
```
owencraston, line 795: I am not sure why bit I was only able to get this to work by hardcoding the fullpath.

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -442,9 +442,9 @@
 				15ACC9FC22655C3A0063978B /* Lottie.framework in Frameworks */,
 				15F7795E22A1B7B500B1DF8C /* Mixpanel.framework in Frameworks */,
 				153F84CA2319B8FD00C19B63 /* Branch.framework in Frameworks */,
-				0FD509E0336BF221F6527B24 /* BuildFile in Frameworks */,
-				D45BF85DECACCB74EDCBE88A /* BuildFile in Frameworks */,
-				B638844E306CAE9147B52C85 /* BuildFile in Frameworks */,
+				0FD509E0336BF221F6527B24 /* (null) in Frameworks */,
```
owencraston, line 445: This was auto generated but should we be concerned by these null values?

scripts/build.sh
```diff
@@ -346,6 +407,37 @@ buildAndroidRelease(){
 	fi
 }
 
+buildAndroidFlaskRelease(){
+	# remap flask env variables to match what the app expects
+	remapFlaskEnvVariables
+
+	echo $PRE_RELEASE
```
owencraston, line 414: Remove

scripts/build.sh
owencraston: Correct

bitrise.yml
Gudahtt: Hmm. This is an option, but it's a bit tricky.

We use `METAMASK_ENVIRONMENT` to distinguish between dev builds, test builds, production-like builds, and the real production builds that get submitted to stores. We have a separate "build type" variable on extension to distinguish between each distribution (flask, beta, main, mmi, etc.).

If we conflate the two here, it means we'll have no way to differentiate between dev, test, production-like, and production builds of Flask. The error logs will all get merged together.

owencraston:
🥳 the release pipeline is working on iOS!
Screenshot 2023-06-28 at 2 10 19 PM

bitrise.yml
owencraston: I discussed this with @Gudahtt on slack and we decided it might be best to introduce a new env variable to track the built type called METAMASK_BUILD_TYPE. For the flask builds we set `METAMASK_BUILD_TYPE=flask` and then set combine that for non-main build types when setting the Sentry environment variable, like on [extension](https://github.com/MetaMask/metamask-extension/blob/34375a57e558853aab95fe35d5f278aa52b66636/app/scripts/lib/setupSentry.js#L91).

app/util/sentryUtils.js
```diff
@@ -154,8 +155,15 @@ export function setupSentry() {
   const init = async () => {
     const dsn = process.env.MM_SENTRY_DSN;
 
+    /* Similar to the environment logic we use on extension:
+     - https://github.com/MetaMask/metamask-extension/blob/34375a57e558853aab95fe35d5f278aa52b66636/app/scripts/lib/setupSentry.js#L91
+    */
     const environment =
-      __DEV__ || !METAMASK_ENVIRONMENT ? 'development' : METAMASK_ENVIRONMENT;
+      __DEV__ || !METAMASK_ENVIRONMENT
```
sethkfman, line 162: Can you add tests with explicit output and comments? This thing is getting out of hand.

bitrise.yml
sethkfman: @NicolasMassart is correct.

app/util/sentryUtils.js
Gudahtt: On extension we found it easier to disable Sentry entirely in development builds (see https://github.com/MetaMask/metamask-extension/blob/230c0c6fa1626bc1afd68c22321f5c166425dd12/app/scripts/lib/setupSentry.js#L88). Most of our Sentry testing was done with QA or production-like builds as part of regression testing. On the rare occasion that we wanted to test Sentry with a development build, it's easy to comment out that line.

That would reduce this condition to a single ternary:

```
const environment = METAMASK_BUILD_TYPE === 'main' ?
  METAMASK_ENVIRONMENT :
  `${METAMASK_ENVIRONMENT}-${METAMASK_BUILD_TYPE}`;
```

Gudahtt: If we wanted to go that route here, and disable Sentry for development builds, that would be simpler to do in a separate PR. Better to keep this one focused if we can.

A more direct solution to cleaning this up for now would be to split this into two steps: determine the environment, then the sentry environment. e.g.

```
const metamaskEnvironment = __DEV__ || !METAMASK_ENVIRONMENT ? 'development' : METAMASK_ENVIRONMENT;

const sentryEnvironment = METAMASK_BUILD_TYPE === 'main' ?
  METAMASK_ENVIRONMENT :
  `${METAMASK_ENVIRONMENT}-${METAMASK_BUILD_TYPE}`;
```

owencraston: I have went ahead and modified this logic to live inside a function I am calling `deriveSentryEnvironment`. 
```js
/**
 * Derives the Sentry environment based on input parameters.
 * This function is similar to the environment logic used in MetaMask extension.
 * - https://github.com/MetaMask/metamask-extension/blob/34375a57e558853aab95fe35d5f278aa52b66636/app/scripts/lib/setupSentry.js#L91
 *
 * @param {boolean} isDev - Represents if the current environment is development (__DEV__ global variable).
 * @param {string} [metamaskEnvironment='local'] - The environment MetaMask is running in
 *                                                  (process.env.METAMASK_ENVIRONMENT).
 *                                                  It defaults to 'local' if not provided.
 * @param {string} [metamaskBuildType='main'] - The build type of MetaMask
 *                                              (process.env.METAMASK_BUILD_TYPE).
 *                                              It defaults to 'main' if not provided.
 *
 * @returns {string} - The Sentry environment. Possible values are 'development', 'local',
 *                     'production', or a string in the format `${metamaskEnvironment}-${metamaskBuildType}`.
 *                     'development' is returned if 'isDev' is true or 'metamaskEnvironment' is not provided.
 *                     'metamaskEnvironment' is returned if 'metamaskBuildType' is 'main' or undefined.
 *                     `${metamaskEnvironment}-${metamaskBuildType}` is returned for other cases,
 *                     for example 'production-flask' or 'debug-flask'.
 */
export function deriveSentryEnvironment(
  isDev,
  metamaskEnvironment = 'local',
  metamaskBuildType = 'main',
) {
  const environment =
    isDev || !metamaskEnvironment
      ? 'development'
      : metamaskBuildType === 'main'
      ? metamaskEnvironment
      : `${metamaskEnvironment}-${metamaskBuildType}`;

  return environment;
}
```

This function is well tested in the new `sentryUtils.test.ts` file. I opted to not split up the meta mask environment and sentry environment to keep continuity with what we have already published in sentry. what would be the use case for having a separate `metamaskEnvironment` if we don't log it in sentry?

app/util/sentry/sentryUtils.js
```diff
@@ -150,24 +150,54 @@ function sanitizeAddressesFromErrorMessages(report) {
...
+ *                     'metamaskEnvironment' is returned if 'metamaskBuildType' is 'main' or undefined.
+ *                     `${metamaskEnvironment}-${metamaskBuildType}` is returned for other cases,
+ *                     for example 'production-flask' or 'debug-flask'.
+ */
+export function deriveSentryEnvironment(
+  isDev,
+  metamaskEnvironment = 'local',
+  metamaskBuildType = 'main',
+) {
+  const environment =
```
Gudahtt, line 178: Nit: Thoughts on splitting this up onto multiple lines? The standard metamask ESLint config (not yet used here) disallows nested ternaries. They're a bit hard to read.

app/util/sentry/sentryUtils.js
```diff
@@ -150,24 +150,54 @@ function sanitizeAddressesFromErrorMessages(report) {
...
+ *
+ * @param {boolean} isDev - Represents if the current environment is development (__DEV__ global variable).
+ * @param {string} [metamaskEnvironment='local'] - The environment MetaMask is running in
+ *                                                  (process.env.METAMASK_ENVIRONMENT).
+ *                                                  It defaults to 'local' if not provided.
+ * @param {string} [metamaskBuildType='main'] - The build type of MetaMask
+ *                                              (process.env.METAMASK_BUILD_TYPE).
+ *                                              It defaults to 'main' if not provided.
+ *
+ * @returns {string} - The Sentry environment. Possible values are 'development', 'local',
```
Gudahtt, line 166: Nit: it might be better to include less specifics here, so we don't have to update it each time we add a new environment

app/util/sentry/sentryUtils.js
```diff
@@ -150,24 +150,54 @@ function sanitizeAddressesFromErrorMessages(report) {
...
+ *                     for example 'production-flask' or 'debug-flask'.
+ */
+export function deriveSentryEnvironment(
+  isDev,
+  metamaskEnvironment = 'local',
+  metamaskBuildType = 'main',
+) {
+  const environment =
+    isDev || !metamaskEnvironment
+      ? 'development'
```
Gudahtt, line 180: Should we include the build type even for development builds? Seems like it could be useful

Gudahtt:
It's not yet clear to me whether the metric events are being flagged as being from Flask rather than from the main release. This is quite important. We rely on those metrics a lot, so we need to ensure we know precisely what they mean.

I'm not familiar with how metrics are setup on mobile at the moment, so I'm not sure how to approach this.

owencraston:
The environment `production-flask` is working as expected

Screenshot 2023-07-12 at 5 33 26 PM

owencraston:
I created a production testflight build of this branch (v 7.2.1 1145) and it appears that there are no "extra" environments in sentry with the new logic.
Screenshot 2023-07-13 at 11 13 04 AM

app/util/sentry/sentryUtils.js
owencraston: I have broken it up into multiple return statements in my latest commit.

android/app/build.gradle
```diff
@@ -255,6 +254,8 @@ android {
         }
         flask {
           dimension "version"
+          versionName "0.0.2"
```
owencraston, line 257: with this we can have separate versions and build numbers

bitrise.yml
sethkfman: We should update all the version but it is not necessary on this PR.

ios/MetaMask/MetaMask-Flask-Info.plist
```diff
@@ -128,14 +126,16 @@
 		<key>test</key>
 		<string>$(MM_BRANCH_KEY_TEST)</string>
 	</dict>
+	<key>branch_universal_link_domains</key>
```
sethkfman, line 129: we should be adding the iOS bundle identifier and android apk hash to branch or deep linking won't work

scripts/build.sh
```diff
@@ -106,6 +106,43 @@ checkParameters(){
...
+}
+
+remapFlaskEnvVariables() {
+	# remap flask env variables to match what the app expects
+
+	echo "Remapping flask env variable names to match production"
+
+	# js.env variables``
+	remapEnvVariable "FLASK_MOONPAY_API_KEY_STAGING" "MOONPAY_API_KEY_STAGING"
+    remapEnvVariable "SEGMENT_FLASK_DEV_KEY" "SEGMENT_DEV_KEY"
```
sethkfman, line 135: format error

ios/MetaMask/MetaMask-Flask-Info.plist
owencraston: I will work with Andrea to get this working after we submit the builds.

Issue #6654, opened 2023-06-21
[Bug] Approve button is always disabled for ERC721 tokens


seaona:
**Describe the bug**
The Approve button for ERC721 tokens is always disabled

**Screenshots**

https://github.com/MetaMask/metamask-mobile/assets/54408225/ffd0d1f6-fef1-44c6-beb2-eb13b9beb6eb

**To Reproduce**
1. Go to test-dapp
2. Deploy an NFT contract
3. Mint a token
4. Click Approve
5. See how Approve button is always disabled

**Expected behavior**
Approve button should be enabled once gas estimation is ready

**Smartphone (please complete the following information):**
 - Device: Pixel 6
 - OS: Android
 - App Version: 7.2.0 / main

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

marctatham:
I'm still seeing this issue on Android 🙄 
**Device:** pixel 6 
**OS:** Android 15 

**Recording:** 
https://github.com/user-attachments/assets/7d2cb0f7-b346-4111-a59c-8572d6047b04

PR #6660, opened 2023-06-21
fix: Approve button is always disabled for ERC721 tokens

blackdevelopa:
**Development & PR Process**

**Description**

**Screenshots/Recordings**

https://github.com/MetaMask/metamask-mobile/assets/29962968/6cc0028e-8d45-474e-a565-20624511db07

**Issue**

Progresses #6654

**Checklist**

### DIFF
--- a/app/components/UI/ApproveTransactionReview/index.js
+++ b/app/components/UI/ApproveTransactionReview/index.js
@@ -653,10 +653,8 @@ class ApproveTransactionReview extends PureComponent {
 
   goToSpendCap = () => this.setState({ isReadyToApprove: false });
 
-  customSpendInputValid = (value) => {
-    if (this.state.token?.tokenStandard === ERC20) {
-      this.setState({ isCustomSpendInputValid: value });
-    }
+  handleSetIsCustomSpendInputValid = (value) => {
+    this.setState({ isCustomSpendInputValid: value });
   };
 
   toggleLearnMoreWebPage = (url) => {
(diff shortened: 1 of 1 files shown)


app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -653,7 +653,9 @@ class ApproveTransactionReview extends PureComponent {
   goToSpendCap = () => this.setState({ isReadyToApprove: false });
 
   customSpendInputValid = (value) => {
```
digiwand, line 655: [out-of-scope] we should prepend functions and methods with verbs. it would be nice to update this method name e.g. maybe 
`customSpendInputValid` -> `handleSetIsCustomSpendInputValid`

app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -653,7 +653,9 @@ class ApproveTransactionReview extends PureComponent {
...
-    this.setState({ isCustomSpendInputValid: value });
+    if (this.state.token?.tokenStandard === ERC20) {
+      this.setState({ isCustomSpendInputValid: value });
+    }
```
digiwand, line 658: wrapping this with the ERC20 check does not seem necessary since it seems the UI with access to this method is only shown when the token is an ERC20 token. 

If we are adding a check here, I think we should include a warning rather than having a case like this fail silently

digiwand, line 658: wrapping this with the ERC20 check does not seem necessary since the UI with access to this method is only shown when the token is an ERC20 token.

If we are adding a check here, I think we should include a warning rather than having a case like this fail silently

digiwand (approved):
🙌🏼

jpuri (approved):

digiwand (approved):
Lgtm!

seaona:
From QA looks good too! I can see how the issue has been fixed @blackdevelopa great work :fire: 


https://github.com/MetaMask/metamask-mobile/assets/54408225/bf4e1137-f958-4af2-9287-aea90fce475c

Issue #6505, opened 2023-05-31
[BUG] Deeplink: Request payment is not calculating gas, intrinsic gas too low error


chrisleewilcox:
Related to https://github.com/MetaMask/metamask-mobile/issues/6490

seaona:
This PR also fixes this issue https://github.com/MetaMask/metamask-mobile/pull/6677 cc @segun @bschorchit

Issue #6490, opened 2023-05-30
[Bug] Request token payment does not calculate gas accordingly, making transaction always fail with `Intrinsic gas too low`


seaona:
**Describe the bug**
Whenever we are scanning a QR code for Request an ERC20 token payment, we can see how the gas is not correctly calculated making the transaction always fail with the error `Intrinsic gas too low`.

**Screenshots**

https://github.com/MetaMask/metamask-mobile/assets/54408225/4c7b520b-0c38-460b-9bfc-03a98e531a0d

**To Reproduce**
_Steps to reproduce the behavior_
1. Import an ERC20 token
2. Click token details
3. Click Receive
4. Click Request Payment
5. Add an amount
6. Copy the QR code
7. Change accounts
8. Open the QR code scanner
9. Scan the QR code from step 6
10. See how gas is not correctly calculated, making the transaction always fail

**Expected behavior**
Gas should be calculated accordingly to the ERC20 token transfer.

https://github.com/MetaMask/metamask-mobile/assets/54408225/b6acc92c-a0d0-4186-b570-d897103342c1

**Smartphone (please complete the following information):**
 - Device: Pixel 6
 - OS: Android
 - App Version: `6.6.0` or `6.7.1` - in current production (`6.5.0`) is working fine

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

PR #6677, opened 2023-06-22
Fix: deeplink intrinsic gas issue

### LINKED ISSUE #4853: Deeplinks/QR transfer scans no longer estimate the needed gas limit: "intrinsic gas too low"

segun:
**Description**

**Describe the bug**
Previously, when opening an ERC20 transfer deeplink the gas limit would be estimated, allowing users to "just click" send.  Now, when a user opens an ERC20 transfer, the gas limit is set to the default of 21000, which is far too low.  Thus, when clicking send, they get an "intrinsic gas too low" error, as metamask knows the tx would fail so does not send it (Thanks there :D ).  What should be happening is the gas is set to something like 75k depending on the complexity of a contract's transfer function.

I'm pretty familiar with the EIP-681 standard, so I took a stab at using the query params "gasLimit" and "gas" in the deeplink to at least fix this for my users temporarily.  Luckily for me, "gas" is accepted in the deeplink and QR scanning formats (though gasLimit is not, and per the 681 standard, should be).  But with this solution, users have to pay more gas than they need to up front since I had to set it to a high default.

The desired UX is that gas estimation is always done if gas or gasLimit is not specified in the link.  This currently works when you click send on a token, so the functionality exists already within the app.

**Screenshots**
Before

image

After

image

**Issue**
Fixes #4853
Progresses/Fixes

- #6505 
- #6490 

**NOTE**
The issue is basically that the gas fees are not being calculated properly. This solution makes sure the gas fees are properly calculated. 

When the suggested gas fee is 21000, the transaction will fail with intrinsic gas too low. But when the gas is properly calculated, it's usually above 21000.

But even after this solution, on the confirm screen there's usually a brief moment (during my testing like 1 to 3 seconds) where the Send button is active while the gas fee is 21000. 

I'm not sure how to fix this short of hardcoding suggestedGas * 2 when suggestedGas <= 21000.

**To Reproduce**
1. Create a deeplink for an ERC20 transfer https://metamask.github.io/metamask-deeplinks/#
2. Open it using a mobile device
3. Try to send the transaction without editing gas parameters
4. See it fail

**Checklist**

* [x ] There is a related GitHub issue

### DIFF
--- a/app/components/UI/TransactionReview/TransactionReviewEIP1559Update/index.tsx
+++ b/app/components/UI/TransactionReview/TransactionReviewEIP1559Update/index.tsx
@@ -84,7 +84,11 @@ const TransactionReviewEIP1559Update = ({
       updateTransactionState(gasTransaction);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [gasEstimationReady, updateTransactionState]);
+  }, [
+    gasEstimationReady,
+    updateTransactionState,
+    gasTransaction.suggestedGasLimit,
+  ]);
 
   const openLinkAboutGas = useCallback(
     () => Linking.openURL(AppConstants.URLS.WHY_TRANSACTION_TAKE_TIME),
--- a/app/components/UI/TransactionReview/TransactionReviewEIP1559Update/index.tsx
+++ b/app/components/UI/TransactionReview/TransactionReviewEIP1559Update/index.tsx
@@ -77,6 +77,7 @@ const TransactionReviewEIP1559Update = ({
     transactionFeeFiat,
     transactionTotalAmount,
     transactionTotalAmountFiat,
+    suggestedGasLimit,
   } = gasTransaction;
 
   useEffect(() => {
@@ -84,11 +85,7 @@ const TransactionReviewEIP1559Update = ({
       updateTransactionState(gasTransaction);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [
-    gasEstimationReady,
-    updateTransactionState,
-    gasTransaction.suggestedGasLimit,
-  ]);
+  }, [gasEstimationReady, updateTransactionState, suggestedGasLimit]);
 
   const openLinkAboutGas = useCallback(
     () => Linking.openURL(AppConstants.URLS.WHY_TRANSACTION_TAKE_TIME),
--- a/app/components/Views/Send/index.js
+++ b/app/components/Views/Send/index.js
@@ -379,7 +379,7 @@ class Send extends PureComponent {
         newTxMeta.gasPrice = toBN(gas);
       }
 
-      // if gas or gasPrice is not defined in the deeplink, we should define them
+      // if gas and gasPrice is not defined in the deeplink, we should define them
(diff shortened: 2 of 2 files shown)


app/components/UI/TransactionReview/TransactionReviewEIP1559Update/index.tsx
```diff
@@ -84,7 +84,11 @@ const TransactionReviewEIP1559Update = ({
       updateTransactionState(gasTransaction);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [gasEstimationReady, updateTransactionState]);
+  }, [
+    gasEstimationReady,
+    updateTransactionState,
+    gasTransaction.suggestedGasLimit,
```
blackdevelopa, line 90: We should be able to destructure `suggestedGasLimit` out of gasTransaction after line 79.

blackdevelopa (approved):

digiwand (approved):

seaona:
@segun great job! I can see the 2 issues fixed. The 2 issues can be closed now


https://github.com/MetaMask/metamask-mobile/assets/54408225/77b8d6e6-7955-4145-8535-f0ac03b2c05b

PR #6717, opened 2023-06-28
fix: patch zero balance when getBalance query times out

### LINKED ISSUE #6590: [Bug] Balance amount displays `0` for several seconds and back the correct balance in a repeated sequence
**Describe the bug**
On the Home screen, I can see how balance is set to `0` for some seconds and then got back to my current balance. This happens very often (minimum once every 1 or 2 minutes).
I discard an API issue as I can see how it's working fine in production, with the same API and RPC node.

**Screenshots**

https://github.com/MetaMask/metamask-mobile/assets/54408225/f95fc28a-c480-4e07-8911-fbd238e5b0ac

**To Reproduce**
1. Go to Home screen
2. Select Mainnet
3. Wait
4. See balance is set to `0` 
5. See balance is changed back to your current balance
6. See balance is set to `0`
(this sequence keeps repeating)

**Expected behavior**
Balance should not be `0` if I have funds

 - Device: Pixel 6
 - OS: Android
 - App Version: happening on current `main` (last commit https://github.com/MetaMask/metamask-mobile/commit/1b79574407323fc7bac30d1ac06190d8f0017644)

**Severity**

- Is this tech debt?

NicolasMassart:
## Development & PR Process

## Description

Main token (eth) balance shows zero balance when the getBalance query times out as the store is updated to the 0x0 default value instead of not being updated at all. 

This randomly shows a zero balance making the wallet display frightening zero ETH and show the "Add crypto to get started" button. Button only disapear and balance is ok when a new successful call is done. So you have the wrong state for 10s between two polls.

Return an error instead of 0x0 as a balance when an error actually happen. This requires an update in `patches/@metamask+assets-controllers+5.0.0.patch`

## Testing

You can trigger timeouts by reducing timeout value in `node_modules/@metamask/controller-utils/dist/util.js` at Line 194:

`function safelyExecuteWithTimeout(operation, logError = false, timeout = 500) {`
--> `function safelyExecuteWithTimeout(operation, logError = false, timeout = 1) {`

## Issue

fixes MetaMask/metamask-mobile#6590
fixes MetaMask/mobile-planning#1079
fixes MetaMask/mobile-planning#1071

## Checklist


patches/@metamask+assets-controllers+5.0.0.patch
```diff
@@ -65,8 +67,8 @@ index 130e3dc..912c84c 100644
 +     */
 +    getBalanceFromChain(address) {
 +        return __awaiter(this, void 0, void 0, function* () {
-+            let balance = '0x0';
-+            yield (0, controller_utils_1.safelyExecuteWithTimeout)(() => __awaiter(this, void 0, void 0, function* () {
++            let balance;
++            const result = yield (0, controller_utils_1.safelyExecuteWithTimeout)(() => __awaiter(this, void 0, void 0, function* () {
```
tommasini, line 71: If we are not changing the util function `safelyExecuteWithTimeout` and only initiating the balance without a default '0x0' value, the result variable that has the return of the `safelyExecuteWithTimeout`  is not being used.

tommasini (changes requested):
Left a comment! 
I just see a possible edge case, if the get balance fails and we don't have a default value, I can see the app crashing creating a new wallet for example, unless it's handled on other place that I'm not seeing

patches/@metamask+assets-controllers+5.0.0.patch
NicolasMassart: Sorry I forgot to remove it...

NicolasMassart:
> I just see a possible edge case, if the get balance fails and we don't have a default value, I can see the app crashing creating a new wallet for example, unless it's handled on other place that I'm not seeing

[Each new account balance state is initialised with 0x0](https://github.com/MetaMask/core/blob/dc9eb67eed23375bdcaa9db2ddfadd995d81f397/packages/assets-controllers/src/AccountTrackerController.ts#L69-L71), so it can not be undefined in state unless the state is broken, but then we have other issues...

tommasini (dismissed):
Amazing! LGTM!

NicolasMassart:
Fix time tracking: 3d

tommasini (approved):
LGTM!!!!

tommasini (dismissed):
LGTM!!!!

chrisleewilcox:
@NicolasMassart I changed the base branch back to release/7.2.0 but there were merge conflicts with a few files.  I then changed the base branch back to main for now.  At least for this release we'll continue working out of the release branch.  Can you take a look at the merge conflicts against release branch?  I'll retest once the conflicts are resolved.

sethkfman (approved):
LGTM

PR #6718, opened 2023-06-28
feat: added missing requestPermissions and getPermissions API methods

shanejonas:
**Description**
Added missing methods `wallet_requestPermissions` and `wallet_getPermissions` to the mobile API.

This brings it in line with the extension API so that dapps that build against requestPermissions don't have to have extra logic to do `requestAccounts` instead.

fixes https://github.com/MetaMask/MetaMask-planning/issues/861

**Screenshots/Recordings**

some screenshots of it working in the new api playground:

### `wallet_requestPermissions`
image
image
#### error case:
image

### `wallet_getPermissions`

image

### DIFF
--- a/app/core/RPCMethods/RPCMethodMiddleware.ts
+++ b/app/core/RPCMethods/RPCMethodMiddleware.ts
@@ -1,11 +1,15 @@
 import { Alert } from 'react-native';
 import { getVersion } from 'react-native-device-info';
-import { createAsyncMiddleware } from 'json-rpc-engine';
+import {
+  createAsyncMiddleware,
+  JsonRpcEngineCallbackError,
+} from 'json-rpc-engine';
 import { ethErrors } from 'eth-json-rpc-errors';
 import { recoverPersonalSignature } from '@metamask/eth-sig-util';
 import RPCMethods from './index.js';
 import { RPC } from '../../constants/network';
 import { NetworksChainId, NetworkType } from '@metamask/controller-utils';
+import { permissionRpcMethods } from '@metamask/permission-controller';
 import Networks, {
   blockTagParamIndex,
   getAllNetworks,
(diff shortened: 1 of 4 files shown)


sethkfman:
@shanejonas I would update this PR as a feature these are new methods implemented.

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -223,7 +227,61 @@ export const getRpcMethodMiddleware = ({
...
+        new Promise<any>((resolve) => {
+          getPermissionsHandler.implementation(
+            req,
+            res,
+            next,
+            () => {
+              resolve(undefined);
+            },
+            {
+              requestPermissionsForOrigin:
```
Cal-L, line 243: Doesn't look like this is used in the `getPermissionsHandler` logic in the controller

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -223,7 +227,61 @@ export const getRpcMethodMiddleware = ({
...
+                }
+                resolve(undefined);
+              },
+              {
+                requestPermissionsForOrigin:
+                  Engine.context.PermissionController.requestPermissions.bind(
+                    Engine.context.PermissionController,
+                    { origin: hostname },
+                  ),
+                getPermissionsForOrigin:
```
Cal-L, line 275: Doesn't look like this is used in the `requestPermissionsHandler` logic in the controller

app/core/RPCMethods/RPCMethodMiddleware.ts
Cal-L: But the input type expects it - Looks like the input type on the controller level needs to be updated?

Cal-L (commented):
Left some comments

Cal-L (dismissed):
LGTM, just need to fix a lint issue

PR #6742, opened 2023-07-05
fix: Browser external application alert on trusted deeplink protocols

jpcloureiro:
**Development & PR Process**

**Description**

With the introduction of [Improve deeplinks experience](https://github.com/MetaMask/metamask-mobile/pull/5218) feature, all URI are intercepted & matched with an allowlist.
URIs with protocols on this allowlist are opened by in-app browser with no restriction.

All protocols not present in this allowlist will trigger an alert box that prompts the user to allow or ignore the action.
When the user allows, this URI is forwarded to the OS deep-link handler.

We've seen this alert box might disrupt the user experience when specific deeplink protocols are used, such as `wc://...` or `metamask://...`

In order to improve such user experience, we have created a list of `trusted deeplink protocols` where these URI won't be opened by in-app browser but won't prompt the user to allow or reject, being automatically redirected to the OS deeplink handler.

This changes the old behavior described [here](https://github.com/MetaMask/metamask-mobile/pull/5218#issuecomment-1315506062)

**Issue**

Progresses # https://github.com/MetaMask/mobile-planning/issues/779

**Checklist**


jpcloureiro:
Hey @sethkfman @NicholasEllul would love your thoughts on this approach

app/util/browser/index.ts
```diff
@@ -100,6 +100,17 @@ export const isTLD = (hostname: string, error: any) =>
  */
 export const protocolAllowList = ['about:', 'http:', 'https:'];
 
+/**
+ *
+ * List of all trusted protocols for OS Linker to handle
+ */
+export const trustedProtocolToDeeplink = [
```
sethkfman, line 107: Do we need to do this on dapp and metamask?

NEllusion (commented):
Overall no concerns with this. However, do we have documentation anywhere that highlights the use-cases for each deeplink protocol? Is there a reason these sites deeplink into MetaMask rather than interact with the in-page provider?

sethkfman:
@jpcloureiro The purpose of the alert box is to make sure users are aware that the navigation is coming from an external source.

app/util/browser/index.ts
sethkfman: After discussing with @jpcloureiro the these protocols should be allow listed since they are default behavior but we want to restrict other activities.

sethkfman (approved):
LGTM

jpcloureiro:
> Overall no concerns with this. However, do we have documentation anywhere that highlights the use-cases for each deeplink protocol? Is there a reason these sites deeplink into MetaMask rather than interact with the in-page provider?

Aside from `metamask:` URI protocol, they are Ethereum Improvement Proposals

`wc:` eip-1328
`ethereum:` eip-831
`dapp:` ERC-1710

Dapps should always use the in-page provider when available. In the event of dapps not detecting / using the in-page provider, we want to support other forms of communication so we don't leave the dapp user in the dark. (for example, Uniswap does not detect in-page provider, using in-app browser on Android)

NEllusion (approved):

cortisiko (approved):
This is QA passed 👍

PR #6749, opened 2023-07-06
feat: PPOM integration with MetaMask mobile app

jpuri:
PPOM Mobile Integration


jpuri:
@SocketSecurity ignore-all

app/core/Engine.ts
```diff
@@ -237,6 +241,33 @@ class Engine {
       const phishingController = new PhishingController();
       phishingController.maybeUpdateState();
 
+      let ppomController;
+      if (process.env.MM_BLOCKAID_UI_ENABLED) {
```
digiwand, line 245: I'm not seeing any checks to make sure both `cdnBaseUrl` and `blockaidPublic` exist here or PPOMController. should we add these checks in somewhere? I think this could especially be handy running the build in multiple environments, testing, etc.

app/core/Engine.ts
```diff
@@ -237,6 +241,33 @@ class Engine {
...
+        try {
+          ppomController = new PPOMController({
+            messenger: this.controllerMessenger.getRestricted({
+              name: 'PPOMController',
```
digiwand, line 249: [out-of-scope]  I see writing these arbitrary strings directly has been how we are setting the value. would be nice to create const for these controllerMessenger names.

app/lib/ppom/PPOMView.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+describe('PPOMView', () => {
+  it('should render correctly shallow', () => {
+    const wrapper = shallow(<PPOMView />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render correctly deeply', () => {
+    const wrapper = render(<PPOMView />);
+    expect(wrapper).toMatchSnapshot();
```
digiwand, line 15: Q: why have both shallow snapshop test and deep snapshot test rather than only deep snapshot test?

app/lib/ppom/PPOMView.tsx
```diff
@@ -0,0 +1,98 @@
...
+  _validateJsonRpc = invoke.bindAsync('PPOM.validateJsonRpc');
+  initPromise: PPOM | undefined = undefined;
+
+  constructor(jsoRpc: any, files: any[][]) {
+    invoke.defineAsync('PPOM.jsonRpc', jsoRpc);
+    files = convertFilesToBase64(files);
+    this.initPromise = this._new(files);
+  }
+
+  async free() {
```
digiwand, line 41: I searched `github.com/MetaMask/ppom-validator` and couldn't find what this method does. can we add comments to provide more insights and possible links to sources that can explain the code more please?

app/components/Nav/App/index.js
```diff
@@ -581,6 +582,7 @@ const App = ({ userLoggedIn }) => {
     // do not render unless a route is defined
     (route && (
       <>
+        {process.env.MM_BLOCKAID_UI_ENABLED && <PPOMView />}
```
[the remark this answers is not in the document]
jpuri: done 👍

app/core/Engine.ts
```diff
@@ -237,6 +241,33 @@ class Engine {
...
+            onNetworkChange: (listener) =>
+              this.controllerMessenger.subscribe(
+                AppConstants.NETWORK_STATE_CHANGE_EVENT,
+                listener,
+              ),
+            ppomProvider: { PPOM, ppomInit },
+            securityAlertsEnabled: true,
+            onPreferencesChange: () => undefined,
+            cdnBaseUrl: process.env.BLOCKAID_FILE_CDN as string,
+            blockaidPublicKey: process.env.BLOCKAID_PUBLIC_KEY as string,
```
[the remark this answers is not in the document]
jpuri: Done

app/lib/ppom/PPOMView.test.tsx
jpuri: Good point, I removed the shallow one.

app/lib/ppom/PPOMView.tsx
jpuri: Its used here: https://github.com/MetaMask/ppom-validator/blob/main/src/ppom-controller.ts#L409

app/core/Engine.ts
jpuri: Hey, do you have suggestion about how should we check this ? I think it's not great idea to throw error if these are absent.

Blockaid validations will fail if these are not provided.

app/core/Engine.ts
```diff
@@ -237,6 +242,38 @@ class Engine {
...
+      let ppomController;
+      if (process.env.MM_BLOCKAID_UI_ENABLED) {
+        try {
+          ppomController = new PPOMController({
+            chainId: addHexPrefix(
+              networkController.state.providerConfig.chainId,
+            ),
+            blockaidPublicKey: process.env.BLOCKAID_PUBLIC_KEY as string,
+            cdnBaseUrl: process.env.BLOCKAID_FILE_CDN as string,
+            messenger: this.controllerMessenger.getRestricted({
```
blackdevelopa, line 254: It seems there are some TS related error here. Can we handle this? [Context](https://github.com/MetaMask/metamask-mobile/issues/6960)

jpuri: PR is updated.

blackdevelopa (dismissed):

app/core/Engine.ts
digiwand: I was initially thinking something like console.error could help inform developers they are missing ENV values, but on second thought it might be superfluous if the current errors provide enough information.

Blockaid validations failing if these are not provided might be tricky to debug

app/core/Engine.ts
```diff
@@ -243,6 +248,39 @@ class Engine {
...
+              networkController.getProviderAndBlockTracker().provider,
+            ppomProvider: {
+              PPOM,
+              ppomInit,
+            },
+            storageBackend: new RNFSStorageBackend('PPOMDB'),
+            securityAlertsEnabled: true,
+          });
+        } catch (e) {
+          Logger.log(`Error initializinf PPOMController: ${e}`);
```
digiwand, line 279:
```suggestion
          Logger.log(`Error initializing PPOMController: ${e}`);
```

app/lib/ppom/PPOMView.tsx
digiwand: I see it's used there. I was wondering what `this._free()` does

digiwand: I can see where it's being used. I was wondering what `this._free()` does

segun (commented):
looks good. few comments

app/core/Engine.ts
jpuri: It is not something we usually do - warning devs about ENV values.

app/core/Engine.ts
```diff
@@ -471,6 +509,10 @@ class Engine {
         }),
       ];
 
+      if (ppomController) {
+        controllers.push(ppomController as any);
```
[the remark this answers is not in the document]
jpuri: I could not do that as type of `controllers` is derived from initializing array.

package.json
```diff
@@ -203,12 +203,14 @@
     "@walletconnect/utils": "^2.9.0",
     "appium-adb": "^9.11.4",
     "asyncstorage-down": "4.2.0",
+    "await-semaphore": "^0.1.3",
```
blackdevelopa, line 206: I can't seem to find where this package, `await-semaphore` is used

blackdevelopa (commented):
Changes looks good.

package.json
jpuri: Good catch @blackdevelopa, I removed it.

blackdevelopa (dismissed):

blackdevelopa (approved):

segun (approved):

PR #6762, opened 2023-07-07
refactor: Use selectors for account tracker controller state access

cryptodev-2s:
**Description**

Account Tracker controller Redux state is now accessed excusively through selectors. This makes future state changes easier to manage.

Most of these changes consist of replacing the direct access to AccountTrackerController State, by selectors.

**Issue**

Relates to https://github.com/MetaMask/mobile-planning/issues/1059

**Checklist**

### DIFF
--- a/app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.types.tsx
+++ b/app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.types.tsx
@@ -1,5 +1,5 @@
 interface Account {
-  balance: number;
+  balance: string;
 }
 
 type Accounts = Record<string, Account>;
--- a/app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.types.tsx
+++ b/app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.types.tsx
@@ -1,9 +1,3 @@
-interface Account {
-  balance: string;
-}
-
-type Accounts = Record<string, Account>;
-
 interface Identity {
   address: string;
   name: string;
@@ -32,7 +26,6 @@ export interface Transaction {
 }
 
 export interface AccountFromToInfoCardProps {
-  accounts: Accounts;
   contractBalances: Record<string, number>;
   identities: Identities;
   network: string;
--- a/app/selectors/accountTrackerController.ts
+++ b/app/selectors/accountTrackerController.ts
@@ -10,8 +10,8 @@ const selectAccountTrackerControllerState = (state: EngineState) =>
 
 export const selectAccounts = createSelector(
   selectAccountTrackerControllerState,
-  (currencyRateControllerState: AccountTrackerState) =>
-    currencyRateControllerState?.accounts,
+  (accountTrackerControllerState: AccountTrackerState) =>
+    accountTrackerControllerState?.accounts,
 );
 
 export const selectAccountsLength = createSelector(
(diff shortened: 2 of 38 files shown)


app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.types.tsx
```diff
@@ -1,5 +1,5 @@
 interface Account {
-  balance: number;
+  balance: string;
```
Gudahtt, line 2: I guess this was a pre-existing error?

Gudahtt: Interestingly, it seems that the `accounts` prop is never used by `AccountFromToInfoCard`, so we could delete it completely instead. It seems to be leftover from when it was used inside a hook. It's still referenced in the hook's dependencies, so wasn't flagged as unused.

app/selectors/accountTrackerController.ts
```diff
@@ -0,0 +1,21 @@
...
+} from '@metamask/assets-controllers';
+import { EngineState } from './types';
+
+const selectAccountTrackerControllerState = (state: EngineState) =>
+  state?.engine?.backgroundState?.AccountTrackerController;
+
+export const selectAccounts = createSelector(
+  selectAccountTrackerControllerState,
+  (currencyRateControllerState: AccountTrackerState) =>
+    currencyRateControllerState?.accounts,
```
Gudahtt, line 14: Wrong variable name referenced here:
```suggestion
  (accountTrackerControllerState: AccountTrackerState) =>
    accountTrackerControllerState?.accounts,
```

app/components/UI/AccountFromToInfoCard/AccountFromToInfoCard.types.tsx
Gudahtt: We can delete this code in a follow-up PR instead if you want. Not a blocker.

cryptodev-2s: Indeed it's not used I will remove it

Gudahtt (dismissed):
LGTM!

Gudahtt:
Passing e2e test run here: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/8cd2a343-e352-45e1-ba61-b9dbd5a79664

sethkfman (dismissed):
LGTM

app/selectors/accountTrackerController.ts
```diff
@@ -0,0 +1,21 @@
+import { createSelector } from 'reselect';
+import {
+  AccountTrackerState,
+  AccountInformation,
+} from '@metamask/assets-controllers';
+import { EngineState } from './types';
+
+const selectAccountTrackerControllerState = (state: EngineState) =>
+  state?.engine?.backgroundState?.AccountTrackerController;
```
mcmire, line 9: Do we want to use `?.` here? I looked through and didn't see `?.` being used in of any places where we were accessing the state before. Adding `?.` introduces the possibility that this function could return null or undefined, but if we can avoid that, then we should:

``` suggestion
  state.engine.backgroundState.AccountTrackerController;
```

app/selectors/accountTrackerController.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const selectAccounts = createSelector(
+  selectAccountTrackerControllerState,
+  (accountTrackerControllerState: AccountTrackerState) =>
+    accountTrackerControllerState?.accounts,
```
mcmire, line 14: Same as above — seems like we could drop the `?.` here to guarantee that we always return something?

``` suggestion
    accountTrackerControllerState.accounts,
```

mcmire (dismissed):
Looks good!

mcmire (approved):

PR #1441, opened 2020-03-20
Detox: Address Book Tests 

ibrahimtaveras00:
This PR contains e2e tests for our address book feature

Flow seen here = https://recordit.co/RVjwGoIXT4

Screen Shot 2020-03-22 at 4 43 25 PM

PR #6770, opened 2023-07-10
feat: Bump `@metamask/keyring-controller` to v6.0.0

gantunesr:
## Description

Bump `@metamask/keyring-controller` from `v1.0.1` to `v6.0.0`.

- Patches
  - `micro-ftch+0.3.1`: Added to comment out a specific node logic that was causing a compatibility issue with another library.
  - `@metamask+message-manager+7.0.1`: Updated since we're using the same version for both `SignatureController` and `KeyringController`.
  - `@metamask+message-manager+1.0.1`: Removed
  - `@metamask+keyring-controller++@metamask+message-manager+7.3.1`: Added to update the data type `PersonalMessageParams`.
- UI/UX changes
  - To reveal the SRP or Private Key, the CTA to verify the password will be disabled until a string of 8 characters or more is provided. After it, the CTA will be enabled and will call to verify the password against the KeyringController and provide a result.

## Test cases

- [ ] Onboarding
  - [ ] Create new wallet
  - [ ] Import SRP
- [ ] Authentication - Lock and unlock wallet using different authentication mechanisms (password, fingerprint, face ID, device pin)
- [ ] Reveal SRP
- [ ] Reveal private key
- [ ] Import private key - Creation of Simple Keyring
  - [ ] Upgrade the app from previous version and import private key
- [ ] Import QR wallet accounts - Creation of QR Keyring
  - [ ] Upgrade the app from previous version and import QR wallet accounts 
- [ ] Send a simple transaction using a HD account
- [ ] Signature methods (Including QR wallet)
  - [ ] Eth Sign
  - [ ] Personal Sign
  - [ ] Sign Typed Data
  - [ ] Sign Typed Data V3
  - [ ] Sign Typed Data V4 
- [ ] Delete imported account (Simple Keyring)
- [ ] Disconnect Keystone (QR Keyring) 

## Screenshots/Recordings

https://github.com/MetaMask/metamask-mobile/assets/17601467/3050db0a-8951-4c27-aaf1-85ff6d281ef5

## Issue

Fixes https://github.com/MetaMask/mobile-planning/issues/755

## Checklist


gantunesr:
@SocketSecurity ignore @metamask/browser-passworder@4.1.0

gantunesr:
Asking confirmation to @AlexJupiter if this behaviour is acceptable,

_To reveal the SRP or Private Key, the CTA to verify the password will be disabled until a string of 8 characters or more is provided. After it, the CTA will be enabled and will call to verify the password against the `KeyringController` and provide a result._

https://github.com/MetaMask/metamask-mobile/assets/17601467/3c17f564-6835-4bf3-ab76-9a07f9662f64

app/core/Engine.test.js
```diff
@@ -40,6 +40,10 @@ describe('Engine', () => {
   // Use this to keep the unit test initial background state fixture up-to-date
   it('matches initial state fixture', () => {
     const engine = Engine.init({});
+    // Replace [Function MetaMaskKeyring] in KeyringController since JSON can't support methods
+    engine.datamodel.state.KeyringController.keyringTypes = [
+      ['MetaMaskKeyring'],
+    ];
```
gantunesr, line 46: @Gudahtt not sure if this is the best approach here since I can't declare functions inside a json file. Any suggestions?

patches/@metamask+keyring-controller+1.0.1.patch
```diff
@@ -1,80 +0,0 @@
...
-     }
-     /**
-@@ -313,8 +315,8 @@ class KeyringController extends base_controller_1.BaseController {
-             const accounts = yield newKeyring.getAccounts();
-             const allAccounts = yield __classPrivateFieldGet(this, _KeyringController_keyring, "f").getAccounts();
-             this.updateIdentities(allAccounts);
--            this.setSelectedAddress(accounts[0]);
--            return this.fullUpdate();
-+            const importedAccountAddress = accounts[0];
-+            return { importedAccountAddress, keyringState: this.fullUpdate() };
```
Cal-L, line 36: Are the areas in the app where it uses these methods still exposing the correct fields? For example this call in `AddAccountsActions` component? 
image

gantunesr: Yes, I think all these cases are correct now and the patch is no longer necessary

app/core/Engine.test.js
Gudahtt: Sorry I'm not sure I follow. `keyringTypes` should be an array of strings, right? Where would a function come in?

AlexJupiter:
[This experience](https://github.com/MetaMask/metamask-mobile/pull/6770#issuecomment-1647743642) looks good to me.

app/core/Engine.test.js
gantunesr: Yeah, the constructor was outdated with those types. Disregard this message

package.json
```diff
@@ -160,14 +161,16 @@
     "@metamask/contract-metadata": "^2.1.0",
     "@metamask/controller-utils": "^3.4.0",
     "@metamask/design-tokens": "^1.11.1",
+    "@metamask/eth-keyring-controller": "10.0.1",
     "@metamask/eth-sig-util": "^4.0.1",
     "@metamask/etherscan-link": "^2.0.0",
     "@metamask/gas-fee-controller": "4.0.0",
-    "@metamask/keyring-controller": "^1.0.1",
+    "@metamask/keyring-controller": "^6.0.0",
```
Gudahtt, line 168: This introduces a new peer dependency warning, which is not ideal. However, it seems safe to ignore it temporarily because the keyring controller only relies on these methods:

```
  removeIdentity: PreferencesController['removeIdentity'];
  syncIdentities: PreferencesController['syncIdentities'];
  updateIdentities: PreferencesController['updateIdentities'];
  setSelectedAddress: PreferencesController['setSelectedAddress'];
  setAccountLabel?: PreferencesController['setAccountLabel'];
  ```
  
  which have not changed since the current version used in mobile

File: app/core/BackgroundBridge/BackgroundBridge.js [hunk 4]
```diff
@@ -100,8 +100,11 @@ export class BackgroundBridge extends EventEmitter {
     );
     Engine.context.PreferencesController.subscribe(this.sendStateUpdate);
 
-    Engine.context.KeyringController.onLock(this.onLock.bind(this));
-    Engine.context.KeyringController.onUnlock(this.onUnlock.bind(this));
+    Engine.controllerMessenger.subscribe('KeyringController:lock', this.onLock);
```
Gudahtt, line 103: We probably want to still bind this:

```suggestion
    Engine.controllerMessenger.subscribe('KeyringController:lock', this.onLock.bind(this));
```

Or wrap it in an arrow function:

```suggestion
    Engine.controllerMessenger.subscribe('KeyringController:lock', () => this.onLock());
```

Otherwise `this` will not be set in the `onLock` function. The same goes for the other handlers as well.

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -153,10 +156,14 @@ export class BackgroundBridge extends EventEmitter {
       return;
     }
 
-    this.sendNotification({
-      method: NOTIFICATION_NAMES.unlockStateChanged,
-      params: true,
-    });
+    try {
```
Gudahtt, line 159: Is this related to the package update in some way? The connection isn't clear to me.

app/core/Engine.ts
```diff
@@ -248,30 +248,36 @@ class Engine {
...
+        setAccountLabel: preferencesController.setAccountLabel.bind(
+          preferencesController,
+        ),
+        encryptor,
+        // @ts-expect-error Error expected.
+        messenger: this.controllerMessenger.getRestricted({
+          name: 'KeyringController',
+        }),
+        state: keyringState,
+        keyringBuilders: [keyringBuilderFactory(QRHardwareKeyring)],
```
Gudahtt, line 279: Nit: It's easy to create a builder function without using this helper, e.g.

```typescript
const qrKeyringBuilder = () => new QRHardwareKeyring();
qrKeyringBuilder.type = QRHardwareKeyring.type;
```

This would avoid the need to import the `@metamask/eth-keyring-controller` package

app/core/Engine.ts
```diff
@@ -248,30 +248,36 @@ class Engine {
         return newIdentities;
       };
 
-      const keyringState =
-        initialKeyringState || initialState.KeyringController;
+      const keyringState = {
+        keyringTypes: [KeyringTypes.qr],
```
Gudahtt, line 252: Is this needed? It's not totally clear to me what this does.

app/core/Engine.ts
```diff
@@ -248,30 +248,36 @@ class Engine {
...
+        ...initialState.KeyringController,
+        ...initialKeyringState,
```
Gudahtt, line 254: Previously we had always used one of these or the other, not both. Perhaps we should continue with that pattern rather than merging them together?

app/core/Engine.ts
```diff
@@ -248,30 +248,36 @@ class Engine {
...
+          preferencesController,
+        ),
+        setSelectedAddress: preferencesController.setSelectedAddress.bind(
+          preferencesController,
+        ),
+        setAccountLabel: preferencesController.setAccountLabel.bind(
+          preferencesController,
+        ),
+        encryptor,
+        // @ts-expect-error Error expected.
```
Gudahtt, line 274: This error appears to be caused by `controllerMessenger` not being declared as a property of this class. It's a pre-existing problem but certainly not intentional/expected. Maybe it would be best to omit this comment, and we can fix this type error in another PR? The same error appears to affect many lines in this file.

package.json
```diff
@@ -140,6 +140,7 @@
     "oss-attribution-generator/**/debug": "^2.6.9",
     "d3-color": "3.1.0",
     "**/fast-xml-parser": "4.2.4",
+    "@ethereumjs/util": "8.0.2",
```
Gudahtt, line 143: What was this resolution for?

package.json
```diff
@@ -160,14 +161,16 @@
     "@metamask/contract-metadata": "^2.1.0",
     "@metamask/controller-utils": "^3.4.0",
     "@metamask/design-tokens": "^1.11.1",
+    "@metamask/eth-keyring-controller": "10.0.1",
```
Gudahtt, line 164: Probably better to use a `^` range if we're keeping this, though maybe we can get rid of it if we manually create the keyring builder

app/components/Views/ResetPassword/index.js
```diff
@@ -474,10 +474,7 @@ class ResetPassword extends PureComponent {
     const { KeyringController } = Engine.context;
     const { originalPassword } = this.state;
     const keychainPassword = originalPassword;
-    const mnemonic = await KeyringController.exportSeedPhrase(
```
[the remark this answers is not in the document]
gantunesr: True. Removed in [bf4c22c](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/bf4c22cbf9b6b4540222347cce7dc6070578ddf8)

app/util/mnemonic/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      'The method uint8ArrayToMnemonic expects a non-empty array',
+    );
+  }
+
+  const recoveredIndices = Array.from(
+    new Uint16Array(new Uint8Array(uint8Array).buffer),
+  );
+  return JSON.stringify(recoveredIndices.map((i) => wordlist[i]).join(' '))
+    .replace(/"/g, '')
+    .trimEnd();
```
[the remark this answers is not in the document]
gantunesr: I used the same approach as the HD keyring to be more consistent. See commits [6ea9407](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/6ea94076d04c82e886ccab5772896707220f16c3) and [223f956](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/223f956a1a0f89a3da6c74fc328175a9886fd784)

app/components/Views/ChoosePassword/index.js
```diff
@@ -551,10 +551,7 @@ class ChoosePassword extends PureComponent {
     const { KeyringController } = Engine.context;
     const { password } = this.state;
     const keychainPassword = this.keyringControllerPasswordSet ? password : '';
-    const mnemonic = await KeyringController.exportSeedPhrase(
-      keychainPassword,
-    ).toString();
-    return JSON.stringify(mnemonic).replace(/"/g, '');
+    return await KeyringController.exportSeedPhrase(keychainPassword);
```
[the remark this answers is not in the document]
gantunesr: @Gudahtt I'm not sure if I follow this comment, which parameters are you referring to?

package.json
gantunesr: Where are you seeing the peer dependency warning?

package.json
gantunesr: There's a dependency problem with this package after `8.0.2`,

```
error: Error: Unable to resolve module zlib from /Users/guto/MetaMask/metamask-mobile/node_modules/micro-ftch/index.js: zlib could not be found within the project or in these directories:
  node_modules
  54 |     const http = require('http');
  55 |     const https = require('https');
> 56 |     const zlib = require('zlib');
     |                           ^
  57 |     const { promisify } = require('util');
  58 |     const { resolve: urlResolve } = require('url');
  59 |     const isSecure = !!/^https/.test(url);
```

package.json
gantunesr: Removed in [d6f4a2f](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/d6f4a2f93643da747534a2bb84b171f98945fb7c)

app/core/Engine.ts
gantunesr: Sounds good, I have removed the comment

app/core/Engine.ts
gantunesr: It's part of the `KeyringController` state as defined in this type, I'm aware that previously it was used to instantiate new keyrings but seems like it's not necessary anymore with the builders,

```ts
type KeyringControllerState = {
  vault?: string;
  isUnlocked: boolean;
  keyringTypes: string[];
  keyrings: Keyring[];
  encryptionKey?: string;
  encryptionSalt?: string;
};
```

app/core/Engine.ts
gantunesr: Good point. See [0e8418a](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/0e8418a1163c912bd8839913903b991eb750a545)

app/core/Engine.ts
gantunesr: Thanks for the suggestion. This way seems better.

app/core/BackgroundBridge/BackgroundBridge.js
gantunesr: This was a temporal fix I forgot to remove.

app/components/Views/ChoosePassword/index.js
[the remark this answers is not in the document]
mikesposito: The return value from `KeyringController.exportSeedPhrase` is now a `Uint8Array`.
This is functionally still ok because the return value from this `getSeedPhrase` function is then passed to `AuthenticationService.newWalletVaultAndRestore`, which then passes the seed value to `KeyringController.createNewVaultAndRestore`, which accepts `Uint8Array` correctly - but all these functions still have `string` as seed type

app/core/Engine.ts
mikesposito: `keyringTypes` was indeed part of the `KeyringController` state, but was unused.
It has been removed in version `7.0.0`

app/components/Views/ChoosePassword/index.js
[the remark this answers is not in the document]
gantunesr: Thanks. Updated in [084b20d](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/084b20dfe2411718ffa9bc7d9bca37c6f0fd5033)

app/core/Engine.ts
gantunesr: Addressed in [f116a95](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/f116a9517dc74532f2dbaffb5148b59e93be2f53) and [3e2dc1c](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/3e2dc1c2acabab9a3a0a12c7a73a7fe8c60a75e1)

app/declarations.d.ts
```diff
@@ -19,3 +19,5 @@ declare module '*.png' {
   const content: ImageSourcePropType;
   export default content;
 }
+
+declare module '@metamask/eth-keyring-controller';
```
mikesposito, line 23: Why do we need to declare EthKeyringController's module?

package.json
Gudahtt: You can see it when running a fresh install (e.g. after deleting `node_modules`. Peer dependency warnings are printed to the console

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -252,17 +252,12 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
     setBlockingModalVisible(true);
-    const importedAccountAddresses = [];
     try {
       for (const account of checkedAccounts) {
-        const accountAddress =
-          await KeyringController.unlockQRHardwareWalletAccount(account);
-        importedAccountAddresses.push(accountAddress);
+        await KeyringController.unlockQRHardwareWalletAccount(account);
       }
-      PreferencesController.setSelectedAddress(importedAccountAddresses[0]);
```
Gudahtt, line 265: It seems that we used to update the selected address beforehand when connecting a QR hardware wallet, but now we don't. What was the reason for this change?

app/util/mnemonic/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  const recoveredIndices = Array.from(
+    new Uint16Array(new Uint8Array(uint8Array).buffer),
+  );
+
+  return recoveredIndices
+    .map((i) => wordlist[i])
+    .join(' ')
+    .trimEnd();
```
Gudahtt, line 56: Nit: Do we need this `trimEnd` step? It should not be possible for there to be trailing whitespace, unless perhaps one of these indexes is larger than the wordlist set. If that happens, we would want to throw an error instead because it wouldn't be a valid SRP.


```suggestion
    .join(' ');
```

patches/@metamask+keyring-controller+1.0.1.patch
```diff
@@ -1,80 +0,0 @@
...
-         });
-     }
-     /**
-@@ -615,16 +617,19 @@ class KeyringController extends base_controller_1.BaseController {
-             yield __classPrivateFieldGet(this, _KeyringController_keyring, "f").addNewAccount(keyring);
-             const newAccounts = yield __classPrivateFieldGet(this, _KeyringController_keyring, "f").getAccounts();
-             this.updateIdentities(newAccounts);
-+            let newHardwareWalletAddress = '';
```
Gudahtt, line 44: It looks like this set of changes (to `unlockQRHardwareWalletAccount`) is still needed in  this patch

package.json
Gudahtt: `zlib` is a native Node.js API, so perhaps we could use a React Native polyfill for it here

Gudahtt: Actually.... it looks like that package is meant to work on both Node.js and Browser, and this function is the Node.js version. Maybe we can patch it to detect the React Native environment as a browser environment. I think we already have the browser polyfills it needs, just not the Node.js polyfills.

Gudahtt:
From the `@metamask/keyring-controller@7.0.0` changelog:

>- **BREAKING**: The `address` type accepted by the `removeAccount` method is now `Hex` ([#1441](https://github.com/MetaMask/core/pull/1441))

I see two calls to `removeAccount` in mobile where the address is of type `string`. We should convert it to Hex first, or update the type used for the address to `Hex` if it's already guaranteed to be `Hex` today.

app/components/Views/ConnectQRHardware/index.tsx
gantunesr: It's to match the new behaviour when adding a new account, see this example when you create a new account using the HD keyring,

https://github.com/MetaMask/metamask-mobile/assets/17601467/4460c697-dc03-44ba-95b7-a4f83d84e477

Now the QR hardware wallet keyring provide the same experience with this change,

https://github.com/MetaMask/metamask-mobile/assets/17601467/038667f4-85d4-4e1c-8b99-a7a9e5e0088e

app/declarations.d.ts
gantunesr: We don't, I forgot to remove it. [ba7e726](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/ba7e726c401164cb3e2b016dd553e48a9427f4e6)

app/core/Engine.test.js
```diff
@@ -45,6 +45,8 @@ describe('Engine', () => {
     backgroundState.PhishingController.listState.allowlist = [];
     backgroundState.PhishingController.listState.blocklist = [];
     backgroundState.PhishingController.listState.fuzzylist = [];
+    // Add mock value to vault
+    backgroundState.KeyringController.vault = 'mock-vault';
```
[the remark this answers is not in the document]
gantunesr: This was undone by going back to version 6.0.0

app/core/Engine.ts
```diff
@@ -248,30 +246,35 @@ class Engine {
...
+          preferencesController,
+        ),
+        encryptor,
+        keyringBuilders: [qrKeyringBuilder],
+        messenger: this.controllerMessenger.getRestricted({
+          name: 'KeyringController',
+        }),
+        state: {
+          vault:
+            initialKeyringState?.vault || initialState.KeyringController?.vault,
```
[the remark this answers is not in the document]
gantunesr: This was undone by going back to version 6.0.0

app/core/Engine.ts
```diff
@@ -516,19 +519,21 @@ class Engine {
...
-            Logger.log('Engine', 'Vault backup failed', result.error);
-          }
-        })
-        .catch((error) => {
-          Logger.error(error, 'Engine Vault backup failed');
-        }),
+    // @ts-expect-error Expect type error
+    this.controllerMessenger.subscribe(
+      'KeyringController:stateChange',
+      (state: any) =>
```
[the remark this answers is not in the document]
gantunesr: Updated in [65409dc](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/65409dc3a6fccd1c2f2d5adc3256c67e3afb9eca)

app/util/mnemonic/index.ts
gantunesr: Correct, this was a mistake on my part. Fixed in [e5148b2](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/e5148b220293377fabd97b80e883b16e4a0cdb1b)

patches/@metamask+keyring-controller+1.0.1.patch
gantunesr: Currently we don't use the return value anymore. In the comment https://github.com/MetaMask/metamask-mobile/pull/6770/files#r1295269704 you can see the current behaviour.

app/core/Engine.ts
```diff
@@ -516,19 +519,21 @@ class Engine {
...
-          if (result.success) {
-            Logger.log('Engine', 'Vault back up successful');
-          } else {
-            Logger.log('Engine', 'Vault backup failed', result.error);
-          }
-        })
-        .catch((error) => {
-          Logger.error(error, 'Engine Vault backup failed');
-        }),
+    // @ts-expect-error Expect type error
```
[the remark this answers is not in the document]
gantunesr: Removed in [ebbeec2](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/ebbeec2be818a31619b6783d5a713ad8c6601218)

app/components/Views/ConnectQRHardware/index.tsx
Gudahtt: Ah I see, so it's not related to the keyring update?

package.json
gantunesr: @Gudahtt I'll reach out to you to discuss this further

app/components/Views/ConnectQRHardware/index.tsx
gantunesr: It had an impact and this seemed the reasonable approach to take in this case

package.json
gantunesr:
```
warning " > @metamask/keyring-controller@6.0.0" has incorrect peer dependency "@metamask/preferences-controller@^4.1.0"
```

Ah I got it, it should be safe to ignore it

Gudahtt:
Looks like CI is still failing

package.json
gantunesr: The `micro-ftch` was patched as discussed to avoid import imcompatible node dependencies, in this case `zlib`. Changes in,

- [4335f0f](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/4335f0fb2be0261969455441c8265f8cb57c5370)
- [ca30637](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/ca30637d7b320d2f6b8e2321719534b6e91bd8bd)
- [3bab33b](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/3bab33bd3b7d6ec06e592fd2a5f25c62fc65b506)

gantunesr:
@Gudahtt I'm trying to figure out what's the issue here, looks like the `signature-controller` does not have `message-manager` inside the node_modules and the patch `@metamask+signature-controller++@metamask+message-manager+7.0.1.patch` fails to apply. But when installing in the main branch, the dependency is present in the node_modules of the `signature-controller`. I guess something is off with the `keyring-controller @ v6.0.0`.

gantunesr:
E2E Result Report,

- [iOS - Test Report ](https://addons-testing.bitrise.io/builds/cdc129d6-55e8-4859-9d23-caa1b76465a6/summary?status=failed)
- [Android - Test Report](https://addons-testing.bitrise.io/builds/a2963d2d-7a25-4d85-8832-627b8f63ba5a/summary?status=failed)

gantunesr:
[E2E Result Report](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/af681567-4819-4891-be37-20ca823cdb53)

jpuri:
@gantunesr : I recently added a patch to message-managers here: https://github.com/MetaMask/metamask-mobile/pull/7068/files#diff-f34e4a6b1f0eb3560ee3c99746e50dfe5778ae15bacf0294b640c2e5cde92973

You may need to update the patch in this PR.

OGPoyraz:
@gantunesr patches are looking good. But before we merge this we must let this PR to be manually tested by the QA team including signatures & transactions with Keystone QR hardware wallet. 
We don't have any e2e tests around that right now and last time broke it unintentionally.

NicolasMassart (commented):
Q: what is the word we should use consistently? SRP, Mnemonic or Seed phrase? I can see all of them in this code and not sure it's expected to be the same thing or not.

gantunesr:
@jpuri I removed that patch because bumping the `keyring-controller` removed the `@metamask/message-manager v1.0.1`, I have included the patch in the `@metamask/message-manager v7.0.1`.

@OGPoyraz I have added the test cases in the PR description, let me know if there's one missing.

@NicolasMassart The BIP39 uses the word `mnemonic` is the spec. I think we should use it over `seedphrase` or `srp` in the code.

app/core/Engine.ts
```diff
@@ -728,19 +729,21 @@ class Engine {
...
-          } else {
-            Logger.log('Engine', 'Vault backup failed', result.error);
-          }
-        })
-        .catch((error) => {
-          Logger.error(error, 'Engine Vault backup failed');
-        }),
+    this.controllerMessenger.subscribe(
+      // @ts-expect-error To Do: Include 'KeyringController:stateChange' to type.
+      AppConstants.KEYRING_STATE_CHANGE_EVENT,
```
gantunesr, line 734: Where is this type defined? I can't find it in the `messenger-controller`

```
Argument of type '"KeyringController:stateChange"' is not assignable to parameter of type '"NetworkController:stateChange" | "ApprovalController:stateChange" | "CurrencyRateController:stateChange" | "GasFeeController:stateChange" | "TokenListController:stateChange" | "NetworkController:providerConfigChange" | "PermissionController:stateChange" | "SignatureController:stateChange"'.ts(2345)
```

app/core/Engine.ts
```diff
@@ -401,29 +399,33 @@ class Engine {
...
+        preferencesController,
+      ),
+      setAccountLabel: preferencesController.setAccountLabel.bind(
+        preferencesController,
+      ),
+      encryptor,
+      // @ts-expect-error Error might be caused by base controller version mismatch
+      messenger: this.controllerMessenger.getRestricted({
+        name: 'KeyringController',
+      }),
```
mikesposito, line 425: Events and Actions needed by the controller should be added here:

```suggestion
      messenger: this.controllerMessenger.getRestricted({
        name: 'KeyringController',
        allowedEvents: [
          'KeyringController:lock',
          'KeyringController:unlock',
          'KeyringController:stateChange',
          'KeyringController:accountRemoved',
        ],
        allowedActions: ['KeyringController:getState'],
      }),
```

app/core/Engine.ts
gantunesr: Resolved in [d8f8565](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/d8f856562787e0d67d693b7e10fcf9b78efd8323). Shout out to @mikesposito for his help!

owencraston (dismissed):
Looks good to me!


https://github.com/MetaMask/metamask-mobile/assets/22918444/57af14a2-66ea-4499-9145-fefbaf4a6f8c

gantunesr:
E2E tests executing in [#9753](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/8f3b977c-3117-449d-bf58-79bae5316a99)

app/components/Views/RevealPrivateCredential/RevealPrivateCredential.tsx
```diff
@@ -533,7 +531,7 @@ const RevealPrivateCredential = ({
         onCancelPress={unlocked ? done : cancelReveal}
         onConfirmPress={() => tryUnlock()}
         showConfirmButton={!unlocked}
-        confirmDisabled={!enableNextButton()}
+        confirmDisabled={password.length < 8}
```
Gudahtt, line 534: There is a `passwordRequirementsMet` function in `app/util/password` that we should use for this instead. If we change these requirements, having them inconsistent across the codebase would be unfortunate!

patches/micro-ftch+0.3.1.patch
```diff
@@ -0,0 +1,245 @@
...
+-            });
+-        }
+-        if (options.keepAlive)
+-            req.setNoDelay(true);
+-        if (opts.body)
+-            req.write(opts.body);
+-        req.end();
+-    });
+-}
++// function fetchNode(url, _options) {
```
Gudahtt, line 126: Could you add a comment to this patch to explain what its purpose is? We should try to ensure all patches have comments moving forward so that they area easier to maintain

patches/@metamask+keyring-controller+1.0.1.patch
```diff
@@ -1,80 +0,0 @@
...
-             const keyring = yield this.getOrAddQRKeyring();
-+            const accountsIncludingHardware = (yield __classPrivateFieldGet(this, _KeyringController_keyring, "f").getAccounts());
-             keyring.forgetDevice();
--            const accounts = (yield __classPrivateFieldGet(this, _KeyringController_keyring, "f").getAccounts());
--            accounts.forEach((account) => {
--                this.setSelectedAddress(account);
--            });
-+            const remainingAccounts = (yield __classPrivateFieldGet(this, _KeyringController_keyring, "f").getAccounts());
-+            const removedAccounts = accountsIncludingHardware.filter((address) => !remainingAccounts.includes(address));
-+            this.updateIdentities(remainingAccounts);
```
Gudahtt, line 73: Looks like this step is missing now. After forgetting the device, the preferences controller identities state might still list the accounts from the now-removed QR keyring unless we call this

patches/@metamask+keyring-controller++@metamask+message-manager+7.3.1.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/node_modules/@metamask/keyring-controller/node_modules/@metamask/message-manager/dist/PersonalMessageManager.d.ts b/node_modules/@metamask/keyring-controller/node_modules/@metamask/message-manager/dist/PersonalMessageManager.d.ts
+index 5bf0005..f372485 100644
+--- a/node_modules/@metamask/keyring-controller/node_modules/@metamask/message-manager/dist/PersonalMessageManager.d.ts
++++ b/node_modules/@metamask/keyring-controller/node_modules/@metamask/message-manager/dist/PersonalMessageManager.d.ts
+@@ -25,7 +25,10 @@ export interface PersonalMessage extends AbstractMessage {
+  */
+ export interface PersonalMessageParams extends AbstractMessageParams {
+     data: string;
+-    siwe?: SIWEMessage;
++    siwe?: {
```
Gudahtt, line 10: Could you add a comment to explain the purpose of this patch? This one seems new to this PR (I can't find it in the removed/moved patches, unless I'm missing something). Was this meant to fix a type error you encountered?

Gudahtt:
Looks great! A few minor suggestions, but functionally everything looks good except maybe for the missing `updateIdentities` call after calling `forgetQRKeyring`

app/components/Views/RevealPrivateCredential/RevealPrivateCredential.tsx
gantunesr: Refactored in [38bacfd](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/38bacfdd233ccfb635bd2ca8da966975b73182cc)

patches/micro-ftch+0.3.1.patch
gantunesr: Good idea! Done in [63965dc](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/63965dc158417f366edd582392cf5c138dee9660)

patches/@metamask+keyring-controller+1.0.1.patch
gantunesr: Done in [279c270](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/279c27046db8c0ffe79dec35299d17962b82eaf3) and also added patch information in [6c4b74d](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/6c4b74d0887acbd359abdee6e261fb6d2a15bd12)

patches/@metamask+keyring-controller++@metamask+message-manager+7.3.1.patch
gantunesr: A type error was introduced in the patch `@metamask+message-manager+7.0.1`. The error in question is displayed below. This new patch resolves the issue by updating to the expected data type.

```
$ tsc --project ./tsconfig.lint.json
app/core/Engine.ts:629:11 - error TS2322: Type '(messageParams: PersonalMessageParams) => any' is not assignable to type '(messsageParams: PersonalMessageParams) => Promise<string>'.
  Types of parameters 'messageParams' and 'messsageParams' are incompatible.
    Type 'import("/Users/guto/MetaMask/metamask-mobile/node_modules/@metamask/message-manager/dist/PersonalMessageManager").PersonalMessageParams' is not assignable to type 'import("/Users/guto/MetaMask/metamask-mobile/node_modules/@metamask/keyring-controller/node_modules/@metamask/message-manager/dist/PersonalMessageManager").PersonalMessageParams'.
      Types of property 'siwe' are incompatible.
        Type '{ isSIWEMessage: boolean; parsedMessage: null; } | undefined' is not assignable to type 'SIWEMessage | undefined'.
          Type '{ isSIWEMessage: boolean; parsedMessage: null; }' is not assignable to type 'SIWEMessage | undefined'.
            Type '{ isSIWEMessage: boolean; parsedMessage: null; }' is not assignable to type '{ isSIWEMessage: false; parsedMessage: null; }'.
              Types of property 'isSIWEMessage' are incompatible.
                Type 'boolean' is not assignable to type 'false'.

629           signPersonalMessage:
              ~~~~~~~~~~~~~~~~~~~

  node_modules/@metamask/signature-controller/dist/SignatureController.d.ts:37:5
    37     signPersonalMessage: (messsageParams: PersonalMessageParams) => Promise<string>;
           ~~~~~~~~~~~~~~~~~~~
    The expected type comes from property 'signPersonalMessage' which is declared here on type 'KeyringController'


Found 1 error in app/core/Engine.ts:629
```

I have added this information in the patch in [61781ff](https://github.com/MetaMask/metamask-mobile/pull/6770/commits/61781ff25a5e3942686c312a1e04a079e14ebdc3).

gantunesr:
Smoke E2E executing in [#9767](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/2eb49d9f-46b1-4afa-9912-b0ec343b32a9)

patches/@metamask+keyring-controller++@metamask+message-manager+7.3.1.patch
Gudahtt: Actually... just realized something. There are two versions of the messenger controller here but v7.3.1 is within the requested range of both the keyring controller and the signature controller. We can deduplicate these and have only one copy.

This actually might be important for not breaking the blockaid integration because the patch that was applied to v7.3.1 for that _will not actually get used_. v7.3.1 of the message manager is being used by the keyring controller, which only references its types. No code run. The message manager used at runtime is the one constructed by the signature controller, v7.0.1 at the moment. We can update both to use v7.3.1

PR #6775, opened 2023-07-10
refactor: Use selectors for preferences controller state access

cryptodev-2s:
**Description**

Preferences controller Redux state is now accessed excusively through selectors. This makes future state changes easier to manage.

Most of these changes consist of replacing the direct access to PreferencesController State, by selectors.

**Issue**

Relates to https://github.com/MetaMask/mobile-planning/issues/1017

**Checklist**

### DIFF
--- a/app/components/Nav/App/index.js
+++ b/app/components/Nav/App/index.js
@@ -86,6 +86,7 @@ import EditAccountName from '../../Views/EditAccountName/EditAccountName';
 import WC2Manager, {
   isWC2Enabled,
 } from '../../../../app/core/WalletConnect/WalletConnectV2';
+import { selectFrequentRpcList } from '../../../selectors/preferencesController';
 
 const clearStackNavigatorOptions = {
   headerShown: false,
@@ -231,10 +232,7 @@ const App = ({ userLoggedIn }) => {
       dispatch(setCurrentBottomNavRoute(route));
     }
   };
-  const frequentRpcList = useSelector(
-    (state) =>
-      state?.engine?.backgroundState?.PreferencesController?.frequentRpcList,
-  );
+  const frequentRpcList = useSelector(selectFrequentRpcList);
 
   useEffect(() => {
     if (prevNavigator.current || !navigator) return;
--- a/app/selectors/preferencesController.ts
+++ b/app/selectors/preferencesController.ts
@@ -1,0 +1,76 @@
+import { createSelector } from 'reselect';
+import { PreferencesState } from '@metamask/preferences-controller';
+import { EngineState } from './types';
+
+const selectPreferencesControllerState = (state: EngineState) =>
+  state.engine.backgroundState.PreferencesController;
+
+export const selectIdentities = createSelector(
+  selectPreferencesControllerState,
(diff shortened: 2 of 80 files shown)


Gudahtt (dismissed):
LGTM! I'll start the e2e tests

Gudahtt:
Looks like there are new conflicts on this one

cryptodev-2s:
> Looks like there are new conflicts on this one

I just did a rebase

Gudahtt (dismissed):

app/components/Nav/App/index.js
```diff
@@ -231,16 +235,12 @@ const App = ({ userLoggedIn }) => {
...
-      state?.engine?.backgroundState?.PreferencesController?.frequentRpcList,
-  );
+  const frequentRpcList = useSelector(selectFrequentRpcList);
+  const selectedAddress = useSelector(selectSelectedAddress);
 
   useEffect(() => {
     if (prevNavigator.current || !navigator) return;
     const appTriggeredAuth = async () => {
-      const { PreferencesController } = Engine.context;
-      const selectedAddress = PreferencesController.state.selectedAddress;
```
Gudahtt, line 243: This change might be responsible for the test failures.

The authentication steps here are now being triggered upon account switch whereas they were not previously.

app/selectors/preferencesController.ts
```diff
@@ -0,0 +1,76 @@
+import { createSelector } from 'reselect';
+import { PreferencesState } from '@metamask/preferences-controller';
+import { EngineState } from './types';
+
+const selectPreferencesControllerState = (state: EngineState) =>
+  state?.engine?.backgroundState?.PreferencesController;
```
mcmire, line 6: Is there a reason why we want to use `?.` here? It seems like it would be better to fix the upstream code rather than allow `state` to be undefined or null here.

cryptodev-2s:
Successful e2e https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/3c7193e0-94d6-440e-a910-9a5551cce290

app/selectors/preferencesController.ts
```diff
@@ -0,0 +1,76 @@
...
+  (preferencesControllerState: PreferencesState) =>
+    preferencesControllerState.disabledRpcMethodPreferences,
+);
+
+// isMultiAccountBalancesEnabled is a patched property - ref patches/@metamask+preferences-controller+2.1.0.patch
+export const selectIsMultiAccountBalancesEnabled = createSelector(
+  selectPreferencesControllerState,
+  (preferencesControllerState: PreferencesState) =>
+    (
+      preferencesControllerState as PreferencesState & {
```
mcmire, line 61: If PreferencesController is patched, why do we need to typecast here? Is there any way we could remove this typecast?

Gudahtt: It's not ideal, but since we're in the process of updating the packages anyway, maybe we don't need to block. It'll be resolved in subsequent PRs

Gudahtt (approved):
LGTM!

mcmire (approved):

PR #6789, opened 2023-07-11
ci(code-cov): adding code-cov to project

sethkfman:
**Development & PR Process**

**Description**

Adding GH workflow code to add code-cov. This is the initial configuration and will require a bit more fine tuning once main has been updated with the workflow.

This is to support a more flexible coverage tool and complement SonarCloud static analysis. 

**Screenshots/Recordings**

NA

**Checklist**

* [NA] There is a related GitHub issue
* [NA] Tests are included if applicable


codecov.yml
```diff
@@ -0,0 +1,21 @@
+codecov:
+  bot: 'codecov-io'
+  require_ci_to_pass: yes
+coverage:
+  status:
+    project:
+      default: # This can be anything, but it needs to exist as the name
+        # basic settings
+        target: auto
+        threshold: 32%
```
Gudahtt, line 10: Have you considered setting thresholds in Jest instead, and using that as the quality gate? That's what the extension (and basically all of our other repos) do. The extension uses codecov just for better coverage reports (and used to use it for better status checks as well, though that has been temporarily disabled due to an extension-specific bug in how that integration is setup)

sethkfman: I have enabled the global code coverage in Jest. I see extension has a lot more code to parallel test runs and evaluate coverage. I will exclude that work on this PR. 

The current workflow would require devs to look at a failing result of test cases update. You can see an example here: https://github.com/MetaMask/metamask-mobile/actions?query=branch%3Aci%2Fadd-code-cov

.github/workflows/ci.yml
```diff
@@ -69,6 +69,10 @@ jobs:
       - run: yarn test:unit --forceExit --silent
         env:
           NODE_OPTIONS: --max_old_space_size=20480
+      - name: Upload coverage reports to Codecov
```
Gudahtt, line 72: It looks like this step wouldn't run in the case where coverage was too low. The report would be helpful then in showing where to increase coverage. You could ensure that this always runs by doing something like this:

```suggestion
      - name: Upload coverage reports to Codecov
        if: ${{ always() }}
```

Not sure how that would act if it failed during setup though, or if no coverage results exist. It might handle it gracefully.

Gudahtt (dismissed):
LGTM! One pending suggestion but this still seems like a step forward regardless

jest.config.js
```diff
@@ -15,7 +15,22 @@ const config = {
   snapshotSerializers: ['enzyme-to-json/serializer'],
   // This is an environment variable that can be used to execute logic only in development
   collectCoverage: process.env.NODE_ENV !== 'production',
-  coveragePathIgnorePatterns: ['/node_modules/', '__mocks__', '<rootDir>/e2e/'],
+  coveragePathIgnorePatterns: [
```
NicolasMassart, line 18: Q: Do we also have to exclude `.github` folder? Not sure if scripts in CI are included in coverage.

Gudahtt: Ah, thanks for noticing this! We don't need to include `.github` because Jest by default only collects coverage from modules that are run by Jest. But that itself is a problem because we're missing files not touched yet by Jest at all.

We should be setting `collectCoverageFrom`: https://jestjs.io/docs/configuration#collectcoveragefrom-array
Then we only have to worry about excluding modules in the range of files we're collecting coverage from (and those executed by Jest in the course of running tests, such as `node_modules`)

sethkfman: Good improvement! I have updated and confirmed values are still accurate. @Gudahtt

jest.config.js
```diff
@@ -15,7 +15,16 @@ const config = {
   snapshotSerializers: ['enzyme-to-json/serializer'],
   // This is an environment variable that can be used to execute logic only in development
   collectCoverage: process.env.NODE_ENV !== 'production',
-  coveragePathIgnorePatterns: ['/node_modules/', '__mocks__', '<rootDir>/e2e/'],
```
Gudahtt, line 18: I think we'll still want to ignore `__mocks__`. There are some examples of that present in the `app` directory`.

We might want to ignore `__snapshots__` too, I see that around

Gudahtt: Oh I see, the snapshots would be excluded due to their file extensions, disregard that idea.

The mocks are JavaScript/TypeScript though, so that still seems valid to include.

Gudahtt (approved):
LGTM!

PR #6799, opened 2023-07-13
feat: prioritize resume over deeplink to accelerate reconnection flow

abretonc7s:
**Description**

Previously deeplink were prioritized to re-established connection which would interrupt the resume() call and re-create a full key exchange.
Now that the connection flow has been stabilized, re-enabled the resume call which greatly improve performance on android.


andreahaku (dismissed):
LGTM

andreahaku (approved):
LGTM

Issue #6818, opened 2023-07-17
Create a test file for migrations


tommasini:
### **Description**
The migrations file has been raising flags that need tests.

Needs to be tested:
* If the migration latest version it's the same as the last migration
* Migrations are returning state

PR #6819, opened 2023-07-17
test: migrations file

tommasini:
**Description**
This is an implementation to test and avoid future migration issues.

This test covers two cases
* The latest version of migration is updated with the number of migrations
* The latest migration implemented it's returning the state

**Issue**

Progresses #6818 

**Checklist**

### DIFF
--- a/app/store/migrations.test.js
+++ b/app/store/migrations.test.js
@@ -1,18 +1,6 @@
 import { migrations, version } from './migrations';
 
 describe('Redux Persist Migrations', () => {
-  it('should apply migration number 19 and return state', () => {
-    let currentState = {
-      recents: '0x1',
-    };
-
-    const migration = migrations[19];
-
-    currentState = migration(currentState);
-
-    expect(currentState).toBeDefined();
-    expect(currentState.recents).toBeUndefined();
-  });
   it('should apply last migration version and return state', () => {
     let currentState = {
       recents: '0x1',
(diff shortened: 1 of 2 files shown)


Gudahtt (dismissed):
LGTM!

tommasini:
Yeah, you are right! If we change the state we are actually breaking the test!

It's actually a good idea to have each member write a test if a migration it's implemented!!

Will push that code asap

cc: @Gudahtt What do you think about this approach?

MarioAslau (dismissed):
Looking good!

Gudahtt (dismissed):
Makes sense!

app/store/migrations.test.js
```diff
@@ -0,0 +1,20 @@
+import { migrations, version } from './migrations';
+
+describe('Redux Persist Migrations', () => {
+  it('should apply last migration version and return state', () => {
+    let currentState = {
```
Cal-L, line 5: rename to oldState

app/store/migrations.test.js
```diff
@@ -0,0 +1,20 @@
...
+      recents: '0x1',
+    };
+
+    const migration = migrations[version];
+
+    currentState = migration(currentState);
```
Cal-L, line 11: rename to newState

Cal-L (approved):
LGTM

PR #6832, opened 2023-07-19
feat: add support for template and header result pages

vinistevam:
**Development & PR Process**

**Description**

This PR aims to include support for header and components on result pages (success & error). Additionally, it also includes support for templates identical to what already exists in the extension.
- Add `MetamaskTemplateRenderer` with a `safeComponentList` to ensure only allow components can be rendered.
- Add support for header and components on result pages (success and error).

**Screenshots/Recordings**

Note: the recordings below are for testing the components no change was done on the add & switch approval.

[header_template_support_error.webm](https://github.com/MetaMask/metamask-mobile/assets/45455812/c53128b7-2de2-46ed-96b9-5d53e3b13321)

[header_template_support_success.webm](https://github.com/MetaMask/metamask-mobile/assets/45455812/2200346b-8e58-4862-9c5a-312e59aeb172)

**Issue**

Progresses: https://github.com/MetaMask/mobile-planning/issues/1028

**Pipeline E2E**
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/6836897f-b4ef-4564-8707-68419ad75ca3

**Checklist**


app/components/UI/MetamaskTemplateRenderer/MetamaskTemplateRenderer.js
matthewwalsh0, line 1: Shall we just call this `TemplateRenderer` as `Metamask` is implied?

matthewwalsh0, line 1: I don't want to delay this if it proves complex, but what our obligations concerning creating new components that aren't in TypeScript? Are we contributing technical debt or is this allowed as we're matching an exiting extension component?

matthewwalsh0, line 1: Also, not required now, but I wonder if there is a future where this is moved to a package to avoid the duplication? Alternatively that may not be possible due to client specific differences.

app/components/UI/MetamaskTemplateRenderer/SafeComponentList.js
```diff
@@ -0,0 +1,46 @@
...
+  IconSize,
+} from '../../../component-library/components/Icons/Icon';
+import BottomSheetFooter, {
+  ButtonsAlignment,
+} from '../../../component-library/components/BottomSheets/BottomSheetFooter';
+import { View } from 'react-native';
+import Box from '../../../components/UI/FiatOnRampAggregator/components/Box';
+
+// eslint-disable-next-line import/prefer-default-export
+export const safeComponentList = {
```
matthewwalsh0, line 23: Just to clarify, is this the minimum necessary to support our action result template?

app/components/UI/MetamaskTemplateRenderer/MetamaskTemplateRenderer.js
matthewwalsh0: It's comforting to go in identical to the extension for now so it's pre-tested in a sense.

app/components/UI/Approval/Confirmation.tsx
matthewwalsh0, line 1: Rather than putting all these files in the root of `UI/Approval`, can we create a new directory such as `Confirmation` to isolate all this from other approval components such as the `ApprovalFlowLoader`?

That would include the `Confirmation` component itself, plus the templates and utils.

app/components/UI/Approval/Confirmation.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onConfirm: () => void;
+  onCancel?: () => void;
+}
+
+type Action = () => void;
+export interface Actions {
+  [key: string]: Action | undefined;
+}
+
+const Confirmation = ({
```
matthewwalsh0, line 29: Is it worth renaming this to `TemplateConfirmation` to be explicit that this confirmation is generic and using templates?

app/components/UI/Approval/Confirmation.tsx
```diff
@@ -0,0 +1,83 @@
...
+            colors,
+          )
+        : {},
+    [approvalRequest, onConfirm, onCancel, colors],
+  );
+
+  const buttons = [
+    {
+      variant: ButtonVariants.Primary,
+      label: templatedValues.submitText ?? strings('approval_result.ok'),
```
matthewwalsh0, line 55: Shall we rename these translation strings now their usage is more generic?

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -442,14 +441,9 @@ const RootRPCMethodsUI = (props) => {
         swipeDirection={'down'}
         propagateSwipe
       >
-        <ApprovalResult
-          requestData={approvalResultRequest?.data}
+        <Confirmation
+          approvalRequest={approvalResultRequest?.request}
           onConfirm={onApprovalResultConfirm}
```
matthewwalsh0, line 446: Do we also need to rename the accept callback and the approval result state as both are now more generic?

app/components/UI/Approval/Confirmation.tsx
```diff
@@ -0,0 +1,83 @@
...
+      size: ButtonSize.Lg,
+      onPress: onConfirm,
+    },
+  ];
+
+  if (onCancel) {
+    buttons.push({
+      variant: ButtonVariants.Secondary,
+      label: templatedValues.cancelText ?? strings('approval_result.cancel'),
```
matthewwalsh0, line 64: Another translation string here that is more generic.

app/components/UI/Approval/Templates/index.ts
```diff
@@ -0,0 +1,85 @@
...
+export const TEMPLATED_CONFIRMATION_APPROVAL_TYPES =
+  Object.keys(APPROVAL_TEMPLATES);
+
+const ALLOWED_TEMPLATE_KEYS: string[] = [
+  'cancelText',
+  'content',
+  'onCancel',
+  'onSubmit',
+  'onConfirm',
+  'networkDisplay',
```
matthewwalsh0, line 22: Should we remove the values that are not yet supported?

app/components/UI/Approval/Templates/index.ts
```diff
@@ -0,0 +1,85 @@
...
+  'submitText',
+  'loadingText',
+];
+
+export async function getTemplateAlerts(
```
matthewwalsh0, line 27: Can we remove both this method and `getTemplateState` if they're not yet being used, for the sake of simplicity?

app/components/UI/Approval/Templates/index.ts
```diff
@@ -0,0 +1,85 @@
...
+export function getTemplateValues(
+  pendingApproval: ApprovalRequest<any>,
+  stringFn: (key: string) => string,
+  actions: Actions,
+  colors: Colors,
+) {
+  const fn = APPROVAL_TEMPLATES[pendingApproval.type]?.getValues;
+  if (!fn) {
+    throw new Error(
+      `MESSAGE_TYPE: '${pendingApproval.type}' is not specified in approval templates`,
```
matthewwalsh0, line 68: `APPROVAL_TYPE:` ?

app/components/UI/Approval/Templates/ApprovalResult.ts
```diff
@@ -0,0 +1,123 @@
...
+  return {
+    content: [
+      {
+        key: 'header',
+        element: 'View',
+        props: {
+          style: styles.accountCardWrapper,
+        },
+        children: [
+          ...(pendingApproval.requestData.header || []),
```
matthewwalsh0, line 53: Is this definitely working?

I worry this is also incorrect in the extension, but do we not have to put this through `convertResultComponents` to ensure it's in the template renderer format rather than the `ApprovalController.ResultComponent` format?

vinistevam: You are right, I tested with the wrong type. I've included a `processComponent` to ensure `ResultComponent` and `ResultComponent[]` are properly converted to `TemplateRendererComponent`.

app/components/UI/Approval/Templates/index.ts
vinistevam: yes, it was changed. It might be good to change the extension as well. I will open a ticket.

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -442,14 +441,9 @@ const RootRPCMethodsUI = (props) => {
         swipeDirection={'down'}
         propagateSwipe
       >
-        <ApprovalResult
-          requestData={approvalResultRequest?.data}
+        <Confirmation
```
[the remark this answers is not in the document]
vinistevam: Make sense, I've included this check on `TemplateConfirmationModal`.

app/components/UI/Approval/Confirmation.tsx
vinistevam: Done

app/components/UI/MetamaskTemplateRenderer/SafeComponentList.js
vinistevam: Nope, only leaving the necessary elements to support result pages.

app/components/UI/MetamaskTemplateRenderer/MetamaskTemplateRenderer.js
vinistevam: There are some differences that I had to address such as how react native needs to be rendered in <Text> component, but nothing that can't be configurable, so I think in the future moving this to a package is possible.

vinistevam: Done

vinistevam: Pushed the migration of those files to typescript.

app/components/Approvals/TemplateConfirmationModal/TemplateConfirmationModal.tsx
```diff
@@ -18,18 +18,18 @@ const ApprovalResultModal = () => {
   }
 
   return (
-    <ApprovalModal isVisible onCancel={onConfirm}>
-      <ApprovalResult
-        requestData={approvalRequest?.requestData}
+    <ApprovalModal
+      isVisible={TEMPLATED_CONFIRMATION_APPROVAL_TYPES.includes(
```
matthewwalsh0, line 22: Rather than relying on `isVisible` which I've found to be a little unstable, shall we just rely on the early return above which needs updating also?

app/components/UI/Approval/TemplateConfirmation/TemplateConfirmation.tsx
```diff
@@ -0,0 +1,84 @@
...
+        : {},
+    [approvalRequest, onConfirm, onCancel, colors],
+  );
+
+  const buttons = [
+    {
+      variant: ButtonVariants.Primary,
+      label: templatedValues.submitText ?? strings('template_confirmation.ok'),
+      size: ButtonSize.Lg,
+      onPress: onConfirm,
```
matthewwalsh0, line 57: If I'm following this right, we pass default action handlers from the props to the templates, but then ignore the actions returned by the templates.

So we're still using the `onConfirm` and `onCancel` from the props, when we should be using `onSubmit` and `onCancel` from the `templatedValues`?

app/components/UI/Approval/TemplateConfirmation/Templates/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  'onSubmit',
+  'onConfirm',
+  'submitText',
+  'loadingText',
+];
+
+export function getTemplateValues(
+  pendingApproval: ApprovalRequest<any>,
+  stringFn: (key: string) => string,
+  actions: Actions,
```
matthewwalsh0, line 29: Should we name this `defaultActions` so it's clear the template can choose to incorporate or ignore them?

app/components/UI/Approval/TemplateConfirmation/Templates/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  'cancelText',
+  'content',
+  'onCancel',
+  'onSubmit',
+  'onConfirm',
+  'submitText',
+  'loadingText',
+];
+
+export function getTemplateValues(
```
matthewwalsh0, line 26: Can we define a type for `TemplateValues` which this method returns and that each template uses on their `getValues` method to keep them consistent?

app/components/UI/Approval/TemplateConfirmation/util.ts
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * Processes a message or ResultComponent and returns
+ * an array of strings | TemplateRendererComponents.
+ *
+ * @param input - The component to process.
+ * @returns The processed component.
+ */
+export function processComponent(
+  input: undefined | string | ResultComponent | ResultComponent[],
```
matthewwalsh0, line 69: The `header` argument in the `ApprovalController` insists this be an array of strings or `ResultComponent`.

Can we minimise the logic and input types here?

app/components/UI/Approval/TemplateConfirmation/util.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/**
+ * Processes a message or ResultComponent and returns
+ * an array of strings | TemplateRendererComponents.
+ *
+ * @param input - The component to process.
+ * @returns The processed component.
+ */
+export function processComponent(
```
matthewwalsh0, line 68: Would calling this `processHeader` be more explicit?

app/components/UI/Approval/TemplateConfirmation/util.ts
```diff
@@ -0,0 +1,211 @@
+import { ResultComponent } from '@metamask/approval-controller';
+
+interface TemplateRendererComponent {
```
matthewwalsh0, line 3: Can we remove this type now the `TemplateRenderer` is in TypeScript and so can export its own types?

app/components/UI/TemplateRenderer/types.ts
```diff
@@ -0,0 +1,12 @@
+export type Sections = string | SectionShape | (string | SectionShape)[];
+
+export interface SectionShape {
+  element: string;
+  key: string;
+  props?: { [key: string]: any };
+  children?: Sections;
+}
+
+export interface SafeComponentList {
```
matthewwalsh0, line 10: Rather than define a type separately for the `SafeComponentList`, can we export some types generated from the actual values of the list and therefore only include the valid keys?

Could we use it on the above `SectionShape` type on the `element` key to filter it to safe elements?

Something like `element: keyof SafeComponentList`?

app/components/UI/TemplateRenderer/TemplateRenderer.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const component = section?.element;
+  const Element = safeComponentList[component];
+  if (!Element) {
+    throw new Error(
+      `${component} is not in the safe component list for MetaMask template renderer`,
+    );
+  }
+  return Element;
+}
+interface TemplateRendererProps {
```
matthewwalsh0, line 18: Should we keep the types at the top of the file?

app/components/Approvals/TemplateConfirmationModal/TemplateConfirmationModal.tsx
```diff
@@ -18,18 +18,18 @@ const ApprovalResultModal = () => {
...
+        approvalRequest.type as ApprovalTypes,
+      )}
+      onCancel={onConfirm}
+    >
+      <TemplateConfirmation
+        approvalRequest={approvalRequest}
         onConfirm={onConfirm}
```
matthewwalsh0, line 29: Do we need the `onCancel` providing here too so we can provide the templates with a default action?

app/components/Approvals/TemplateConfirmationModal/TemplateConfirmationModal.tsx
```diff
@@ -18,18 +18,18 @@ const ApprovalResultModal = () => {
...
 
   return (
-    <ApprovalModal isVisible onCancel={onConfirm}>
-      <ApprovalResult
-        requestData={approvalRequest?.requestData}
+    <ApprovalModal
+      isVisible={TEMPLATED_CONFIRMATION_APPROVAL_TYPES.includes(
+        approvalRequest.type as ApprovalTypes,
+      )}
+      onCancel={onConfirm}
```
matthewwalsh0, line 25: Rather than assuming this is positive or doing the default action, do we need to notify the child component so it can do run the template `onCancel` logic?

There's a few different ways to do this but the on unmount handler might be the cleanest and most decoupled?

app/components/UI/Approval/TemplateConfirmation/Templates/index.ts
```diff
@@ -0,0 +1,52 @@
+import { omit, pick } from 'lodash';
+import approvalResult from './ApprovalResult';
+import { ApprovalTypes } from '../../../../../core/RPCMethods/RPCMethodMiddleware';
+import { Actions } from '../TemplateConfirmation';
+import { Colors } from 'app/util/theme/models';
+import { ApprovalRequest } from '@metamask/approval-controller';
+
+const APPROVAL_TEMPLATES: { [key: string]: any } = {
```
matthewwalsh0, line 8: Can we define a `ConfirmationTemplate` type that insists each of these have a `getValues` method returning the `TemplateValues` type that I mentioned below?

app/components/UI/Approval/TemplateConfirmation/TemplateConfirmation.tsx
vinistevam: Added the handles returned by the template and as a default the `onConfirm` and `onCancel` from the props same as the labels.

app/components/UI/Approval/TemplateConfirmation/Templates/index.ts
vinistevam: Perfect, including the `ConfirmationTemplateValues`  type for `getTemplateValues`.

app/components/UI/Approval/TemplateConfirmation/util.ts
vinistevam: I've reduced the types of input and the logic of the function as you mentioned that `header` is already enforced in the ApprovalController.

app/components/UI/TemplateRenderer/types.ts
vinistevam: Pushed a separate type created from the list of safe elements.

app/components/UI/Approval/TemplateConfirmation/Templates/index.ts
vinistevam: Defined a `ConfirmationTemplate` type to lock that down and ensure all our templates are valid and have the right methods.

app/components/UI/Approval/TemplateConfirmation/util.ts
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @param input - The header to process.
+ * @returns The processed header.
+ */
+export function processHeader(
+  input: (string | ResultComponent)[],
+): (string | SectionShape)[] {
+  const currentInput = convertResultComponents(input);
+
+  return Array.isArray(currentInput)
```
matthewwalsh0, line 68: Are we still not breaking the typing here?

If the type says it's an array, then we should't need any fallback if it's a string?

app/components/UI/TemplateRenderer/types.ts
```diff
@@ -0,0 +1,10 @@
+import { SafeComponentListValues } from './SafeComponentList';
+
+export type Sections = string | SectionShape | (string | SectionShape)[];
```
matthewwalsh0, line 3: These are beautiful types, but are they explicit enough when used externally?

Perhaps `TemplateRendererInput` and `TemplateRendererComponent`?

app/components/UI/Approval/TemplateConfirmation/util.ts
```diff
@@ -0,0 +1,211 @@
...
+ * or an array of strings | SectionShape.
+ *
+ * @param input - The message or component to process.
+ * @param fallback - The fallback message to use when the input is not valid.
+ * @returns The processed error component.
+ */
+export function processError(
+  input: undefined | string | ResultComponent | ResultComponent[],
+  fallback: string,
+): string | SectionShape | (string | SectionShape)[] {
```
matthewwalsh0, line 19: Can we use the `Section` type for these return types?

app/components/UI/TemplateRenderer/types.ts
vinistevam: Makes sense outside the TemplateRender is complicated to understand, renaming to your suggestions.

app/components/UI/Approval/TemplateConfirmation/util.ts
vinistevam: If I remove the fallback the compiler complains that `currentInput` can be a `string`.

matthewwalsh0: That's because `convertResultComponents` supports string inputs but we know we're not passing any in so we could cast its output to `(string | SectionShape)[]` and just return that?

matthewwalsh0 (approved):

OGPoyraz (approved):

PR #6865, opened 2023-07-22
feat: Create a performance E2E test for warm starts

SamuelSalas:
**Description**
Feature test case created to test the warm start of the app from multiple devices. It sends the app to the background for 30 seconds so the Login screen appears.

Progresses #???

**Checklist**


Andepande (dismissed):
LGTM

wdio/features/Performance/WarmStartPerformance.feature
```diff
@@ -0,0 +1,20 @@
+@smoke
+@androidApp
+@performance
+Feature: Measure the app launch times for warm starts
+
+  Scenario: Measure cold start launch time after importing a wallet
```
cortisiko, line 6: change from "cold" to "warm"

wdio/features/Performance/WarmStartPerformance.feature
```diff
@@ -0,0 +1,20 @@
...
+    Given the app displayed the splash animation
+    And I have imported my wallet
+    And I tap No Thanks on the Enable security check screen
+    And I tap No thanks on the onboarding welcome tutorial
+    And I am on the wallet view
+    When I navigate to the browser
+    And I navigate to "https://app.sushi.com/swap"
+    Then the connect modal should be displayed
+    And I connect my active wallet to the dapp
+    When I background the app for 30 seconds
```
cortisiko, line 16:
```suggestion
    When I background the app for 30 seconds
    And I relaunch the app
    And the timer starts running
    When I fill my password in the Login screen
    And I log into my wallet
    Then the app should launch within "5" seconds

File: wdio/features/Performance/WarmStartPerformance.feature [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+    Then the timer starts running
+    And I fill my password in the Login screen
+    And I log into my wallet
+    And the app should launch within "15000" seconds
```
cortisiko, line 20: This seems a bit excessive. We do need to establish a baseline for the warm starts. My idea was to have the test run on main for a week or so to establish a baseline then we create a ticket to set that baseline.

cortisiko (changes requested):
code looks good, just left some feedback on your feature file.

wdio/features/Performance/WarmStartPerformance.feature
SamuelSalas: Adding the "And I relaunch the app" pretty much would be empty. The Background functionality from Appium already relaunch the app the set time pass

wdio/features/Performance/WarmStartPerformance.feature
SamuelSalas: Of course, just set that time to pass the test cases and to use existing steps. I can update or create a new step for this one.

wdio/features/Performance/WarmStartPerformance.feature
cortisiko: Good to know. Without having that context it is hard to tell whether the app has been foregrounded. Following the steps above, backgrounding the app and not specifying the app was foregrounded is a bit confusing. Adding a step to indicate that the app was brought to the foreground would enhance clarity and ease of understanding.

wdio/features/Performance/WarmStartPerformance.feature
cortisiko: yup yup. So in this case, do not specify the expected time in milliseconds. Follow what we already did: https://github.com/MetaMask/metamask-mobile/blob/7a7653652ef2f0629b7b284613986a9e9478508f/wdio/features/Performance/AppLaunchTime.feature#L11 

Also let's set the launch time to be 6 seconds as our baseline. We can adjust as needed.

wdio/step-definitions/warm-start-performance.steps.js
```diff
@@ -0,0 +1,5 @@
+import { Then, When } from '@wdio/cucumber-framework';
+
+When(/^I background the app for (\d+) seconds$/, async (time) => {
```
cortisiko, line 3: We shouldn't even create a new step file for this. Let's move this step to the app launch times step file.

cortisiko:
Lets also go ahead and change the name of https://github.com/MetaMask/metamask-mobile/blob/feature/700-performance-test-warm-starts/wdio/features/Performance/AppLaunchTime.feature to `ColdStartLaunchTimes.feature`

wdio/step-definitions/warm-start-performance.steps.js
cortisiko, line 3: This step can live within `app-launch-times.steps.js`

wdio/features/Performance/WarmStartPerformance.feature
SamuelSalas: Done

wdio/step-definitions/warm-start-performance.steps.js
SamuelSalas: Move to app lunch times step file

SamuelSalas: Move it to app launch times steps file

cortisiko (approved):
PR looks good to me sam! great work!

PR #6877, opened 2023-07-24
ci: Run unit tests in parallel

Gudahtt:
**Development & PR Process**

**Description**

Run Jest unit tests in parallel. The unit tests are now split into "shards" by Jest, and run on separate machines by GitHub Actions. The coverage reports are uploaded as artifacts and merged in the next job, then uploaded to CodeCov as usual.

Since Jest is no longer running all tests in the same process, the global coverage thresholds could no longer be effectively enforced. Instead they have been extracted from the Jest config, and are being enforced by the `nyc` package instead. `nyc` is also what is used to merge the coverage reports.

The tests should now be roughly twice as fast to run in CI.

**Issue**

No associated issue.

**Checklist**


Gudahtt:
@SocketSecurity ignore-all

These warnings are all from nyc, a widely-used package that hasn't been updated in quite a long time. I recognize all of the new authors as well known open source contributors with good reputations, and the packages with filesystem access all seem appropriate to have that access

NicolasMassart (approved):
The changes look good to me but I just wonder how much time we saved with 10 parallel test thread?
I looked at the summary and compared it to another PR test run where tests are still in one thread. the sharded one takes 21 min when the normal one takes 30 min. Do we only save 9 min? It looks strange as I always feel like the test phase is the big part of the CI on PRs.
But anyway, even 9 min saved is better than none.

Gudahtt:
I was expecting more of an increase as well. I think the underwhelming results are due to two factors:

* Installing dependencies takes a long time. This is mitigated with caching, but even when a cache is restored the postinstall scripts still need to be run in their entirety for each job. This step usually takes 2-3 minutes but can be much slower at times (I've seen it at ~7 minutes).
* Some unit tests are much slower than others, and they can end up grouped together in the same shard. Usually most shards complete in ~2-3 minutes but one or two take over double that time

Not sure we can do much about unit test runtime inconsistencies, but there are a lot of opportunities to improve performance of unit tests across the board that we could pursue to help reduce the impact of that.

For the dependencies step, this is why I started looking into depcheck (here: https://github.com/MetaMask/metamask-mobile/pull/7124) and deduplicating dependencies (here: https://github.com/MetaMask/metamask-mobile/pull/7123 though check out the smaller PRs from lego as well). Currently our dependencies are about 3.2 GB, and a full 1GB of that is unnecessary and removed by deduplication. Of the remaining 2.2 GB, a lot of them are unused and can be removed (I don't yet know how much, or what the postinstall impact would be). That would likely speed things up a great deal

PR #6872, opened 2023-07-24
refactor: Update controller packages to v44

Gudahtt:
**Development & PR Process**
1. Follow MetaMask [Mobile Coding Standards](https://docs.google.com/document/d/1VJLwTRsUw_5EDq_o8d6sSbXUAYENLurkRitYO45eY5o/edit?usp=sharing)
2. Add `release-xx` label to identify the PR slated for a upcoming release (will be used in release discussion)
3. Add `needs-dev-review` label when work is completed
4. Add `needs-qa` label when dev review is completed
5. Add `QA Passed` label when QA has signed off

**Description**

Controller packages have been updated to roughly match the versions used in v44 of the core monorepo (this doesn't include the keyring controller, and a few packages are ahead of v44 already).

This update includes a significant breaking change: the migration of `frequentRpcList` in the `PreferencesController` to
`networkConfigurations` in the `NetworkController`. Various other packages have also been updated at the same time because they have a peer dependency upon the `NetworkController`.

**Issue**

Closes https://github.com/MetaMask/mobile-planning/issues/798

**Checklist**

* [x] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

PR #6878, opened 2023-07-24
refactor: Simplify `handleNetworkSwitch` helper

Gudahtt:
**Development & PR Process**

**Description**
The `handleNetworkSwitch` helper function no longer requires the `frequentRpcList` state as a parameter. Instead it takes a reference to the `PreferencesController`, and gets the RPC list state from that.

Typically it would be preferrable to reference controller state from Redux rather than accesing it directly from the Engine global. However, this function will be called from non-UI contexts, so it has to get controller state from global context regardless.

**Issue**

This change was done to simplify PR #6872, which is part of https://github.com/MetaMask/mobile-planning/issues/798

**Checklist**


Gudahtt:
Successful e2e test run: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/fcb12164-9033-4a54-a04b-f0adc815d09f

Cal-L (approved):
LGTM

PR #6881, opened 2023-07-24
test: Improve migration unit tests

Gudahtt:
**Development & PR Process**

**Description**

The migration unit tests have been updated to group tests by the migration number. One additional test case has been added as well, to cover the case where the affected state doesn't exist.

**Issue**

This change was done to simplify PR #6872, which is part of https://github.com/MetaMask/mobile-planning/issues/798

**Checklist**

### DIFF
--- a/app/store/migrations.test.js
+++ b/app/store/migrations.test.js
@@ -1,7 +1,14 @@
+import { cloneDeep } from 'lodash';
 import { migrations, version } from './migrations';
 
 describe('Redux Persist Migrations', () => {
+  it('should have all migrations up to the latest version', () => {
+    // Assert that the latest migration index matches the version constant
+    expect(Object.keys(migrations).length - 1).toBe(version);
+  });
+
   it('should apply last migration version and return state', () => {
+    // update this state to be compatible with the most recent migration
     const oldState = {
       recents: '0x1',
     };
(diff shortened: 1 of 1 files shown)


app/store/migrations.test.js
```diff
@@ -11,10 +18,31 @@ describe('Redux Persist Migrations', () => {
...
+
+  describe('#19', () => {
+    it('should not change state if recents are missing', () => {
+      const oldState = {
+        foo: 'bar',
+      };
+
+      const migration = migrations[19];
+
+      const newState = migration(cloneDeep(oldState));
```
leotm, line 31: do we need lodash? think we're ok w/o it

Screenshot 2023-07-26 at 15 48 23

leotm: and spotted we rarely use it

Screenshot 2023-07-26 at 15 50 29

leotm (approved):
lgtm!

app/store/migrations.test.js
Gudahtt: I used it in this case because these tests are comparing old to new state, so if we don't clone it, it might have been modified and the test would still pass. We want to ensure the state was not modified at all.

This `cloneDeep` function seemed to be how we cloned objects recursively elsewhere.

app/store/migrations.test.js
```diff
@@ -1,7 +1,14 @@
...
+  it('should have all migrations up to the latest version', () => {
+    // Assert that the latest migration index matches the version constant
+    expect(Object.keys(migrations).length - 1).toBe(version);
+  });
+
   it('should apply last migration version and return state', () => {
+    // update this state to be compatible with the most recent migration
     const oldState = {
       recents: '0x1',
     };
```
leotm, line 14: an aside thought, i wonder if in near-future it's worth replacing `oldState` with `new FixtureBuilder().build()`
 https://github.com/MetaMask/metamask-mobile/pull/6519/files#diff-1f76f131ddd71f14064d8cfefa51361f5d0004112d23724fff116daf3631d6d6

Gudahtt: Good idea, I will look into that

PR #6899, opened 2023-07-25
fix: Ensure controller-utils patch is applied by deduplicating dependencies

Gudahtt:
**Development & PR Process**

**Description**

A `deduplicate` script has been added to `package.json`, which allows deduplicating dependencies in the lockfile. This feature is built-in as part of Yarn v3, but on Yarn v1 it's provided by the `yarn-deduplicate` package.

It has been run, so the lockfile has been purged of unnecessary copies of dependencies. This means many dependencies have been updated, but it should result in no breakages as long as all affected packages are SemVer-compliant or handled by one of our resolutions.

This fixes a bug introduced in #6874 where the app would crash when switching to Linea. The bug was caused by the `@metamask/controller-utils` patch not being applied to the version of the package used by the `NetworkController` (because it wasn't deduplicated). Now that dependencies are deduplicated, the `NetworkController` is using the same patched version of the `@metamask/controller-utils` package as this repository.

**Issue**

No associated issue

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "stacktrace:ios": "stack-beautifier sourcemaps/ios/index.js.map -t sourcemaps/trace.txt",
     "update-changelog": "./scripts/auto-changelog.sh",
     "prestorybook": "rnstl",
+    "deduplicate": "yarn yarn-deduplicate",
     "create-release": "./scripts/set-versions.sh && yarn update-changelog",
     "add-release-label-to-pr-and-linked-issues": "ts-node ./.github/scripts/add-release-label-to-pr-and-linked-issues.ts"
   },
@@ -436,7 +437,8 @@
     "ts-node": "^10.5.0",
     "typescript": "^4.4.2",
     "wdio-cucumberjs-json-reporter": "^4.4.3",
-    "xml2js": "^0.5.0"
+    "xml2js": "^0.5.0",
+    "yarn-deduplicate": "^6.0.2"
   },
   "config": {
     "react-native-storybook-loader": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -72,6 +72,7 @@
     "stacktrace:ios": "stack-beautifier sourcemaps/ios/index.js.map -t sourcemaps/trace.txt",
     "update-changelog": "./scripts/auto-changelog.sh",
     "prestorybook": "rnstl",
+    "deduplicate": "yarn yarn-deduplicate",
```
Gudahtt, line 75: We could consider adding a CI check to require deduplicated dependencies. We have this elsewhere, e.g. on extension and some libraries.

Gudahtt: Added in a later PR: https://github.com/MetaMask/metamask-mobile/pull/6900

patches/@metamask+network-controller+5.0.0.patch
```diff
@@ -1,17 +1,34 @@
...
 --- a/node_modules/@metamask/network-controller/dist/NetworkController.js
 +++ b/node_modules/@metamask/network-controller/dist/NetworkController.js
-@@ -81,6 +81,8 @@ class NetworkController extends base_controller_1.BaseControllerV2 {
-             case controller_utils_1.MAINNET:
+@@ -31,7 +31,9 @@ const name = 'NetworkController';
+ exports.defaultState = {
+     network: 'loading',
+     isCustomNetwork: false,
+-    providerConfig: { type: controller_utils_1.MAINNET, chainId: controller_utils_1.NetworksChainId.mainnet },
++    // PATCH: Replace `MAINNET` constant with `NetworkType.mainnet`.
```
Gudahtt, line 10: Just noticed an unintended breaking change in `@metamask/controller-utils@3.1.0`

`MAINNET` was moved to `NetworkType.mainnet`

Gudahtt:
Lots of e2e test failures, likely caused by deduplication unrelated to controller-utils. I'll put this back in draft for now.

I have prepared a more targeted PR to fix the Linea network switching bug: #6903

Gudahtt:
The controller-utils patch was fixed in #6903. I'll pursue deduplicating dependencies later.

PR #6901, opened 2023-07-25
feat: blockaid what's new popup

segun:
## Explanation

Let's communicate to users this amazing new feature that we have!! 🎉

Acceptance Criterias

A new message with image + text should be added to our existing What's new modal
Clicking on Enable security alerts should take the user to this feature within Settings 
This should be behind the same feature/build flag that we have for the rest of Blockaid integration feature

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/927

## Screenshots/Screencaps
Screenshot 2023-07-26 at 10 12 26

## Manual Testing Steps
- install MM afresh
- Register/Import wallet
- What's new popup should be shown.
**Checklist**


app/components/UI/WhatsNewModal/whatsNewList.ts
```diff
@@ -1,17 +1,50 @@
...
+import { strings } from '../../../../locales/i18n';
 import { WhatsNew } from './types';
 
 export const whatsNew: WhatsNew = {
   // All users that have <6.4.0 and are updating to >=6.4.0 should see
-  onlyUpdates: true, // Users who updated the app and new installs will see this.
-  maxLastAppVersion: '6.4.0', // Only users who had a previous version <6.4.0 version will see this
-  minAppVersion: '6.4.0', // Only users who updated to a version >= 6.4.0 will see this
+  onlyUpdates: false, // false: Users who updated the app and new installs will see this. true: only users who update will see it
+  maxLastAppVersion: '7.3.0', // Only users who had a previous version <7.3.0 version will see this
```
blackdevelopa, line 8: Any idea why we are going with this version (7.3.0)?

app/components/UI/WhatsNewModal/whatsNewList.ts
```diff
@@ -1,17 +1,50 @@
...
+        type: 'description',
+        description: strings('whats_new.blockaid.description_3'),
+      },
+      {
+        type: 'button',
+        buttonText: strings('whats_new.blockaid.action_text'),
+        buttonType: 'blue',
+        onPress: (props) =>
+          props.navigation.navigate('SettingsView', {
+            screen: 'ExperimentalSettings',
```
blackdevelopa, line 45: Can we add this to the Routes [file](https://github.com/MetaMask/metamask-mobile/blob/65bc18fb31fdb209f3645f9329f16cdf60b491e6/app/constants/navigation/Routes.ts#L1) instead?

app/components/UI/WhatsNewModal/whatsNewList.ts
```diff
@@ -1,17 +1,51 @@
...
+        type: 'description',
+        description: strings('whats_new.blockaid.description_3'),
+      },
+      {
+        type: 'button',
+        buttonText: strings('whats_new.blockaid.action_text'),
+        buttonType: 'blue',
+        onPress: (props) =>
+          props.navigation.navigate(Routes.SETTINGS_VIEW, {
+            screen: Routes.SETTINGS.EXPERIMENTAL_SETTINGS,
```
blackdevelopa, line 46: Nice

app/components/UI/WhatsNewModal/whatsNewList.ts
jpuri: Also, reason for this `onlyUpdates: false` ?

segun: onlyUpdates = true: Means only users who updated the app (not new installs) will see the popup
onlyUpdates = false: Means all users (new installs and updates) will see the popup.
The version is currently not used (a bug). @bschorchit will talk to the core team about fixing.

blackdevelopa: Thank you for clarifying this. Do you think that we can maybe improve the name so it's a bit more intuitive and understood better? Or introduce another boolean to handle this difference Say, 
```
showForAppUpdates: true, // this should show for users that updates the app
showForNewInstalls: true, // should show for new installs, but if it should only show for app updates, then this becomes false.
```

app/util/blockaid/index.ts
```diff
@@ -0,0 +1,3 @@
+// eslint-disable-next-line import/prefer-default-export
+export const showBlockaidUI = () =>
+  process.env.MM_BLOCKAID_UI_ENABLED === 'true';
```
jpuri, line 3: May be `BLOCKAID_ENABLED`

jpuri (approved):
LGTM, some small feedbacks.

blackdevelopa (approved):

app/util/blockaid/index.ts
segun: 1. Environment variables in mobile starts with MM_
2. Since this is only enabling or disabling the UI and not the whole blockaid, I decided on adding it to the name for clarity.

jpuri: Something like `MM_BLOCKAID_ENABLED` not use `UI` so this name will be meaningful for backend changes also

PR #6903, opened 2023-07-26
fix: Fix crash when switching to Linea

Gudahtt:
**Development & PR Process**

**Description**

A recent `@metamask/controller-utils` update in https://github.com/MetaMask/metamask-mobile/pull/6874 caused the app to crash when switching to Linea because it removed the resolution for this package. Without that resolution, multiple copies of the package were introduced, but the Linea patch was only applied to one of them. The network controller was using a copy without the patch applied, resulting in the crash.

This was resolved by restoring the resolution. We can remove it again once we have deleted the patch.

Additionally, the patch for `@metamask/controller-utils` has been updated to restore constants (`MAINNET and `TESTNET_TICKER_SYMBOLS`) that were removed in v3.1.0 in a change that we missed when preparing the release (which is why it was omitted from the changelog and not recognized as breaking).

They have been restored until we no longer rely on packages that use `@metamask/controller-utils` at versions under v3.1.0.

**Issue**

No associated issue

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "react-native-level-fs/**/bl": "^1.2.3",
     "react-native-level-fs/**/semver": "^4.3.2",
     "@metamask/contract-metadata": "^2.1.0",
+    "@metamask/controller-utils": "^3.4.0",
     "@metamask/approval-controller": "3.4.0",
     "@exodus/react-native-payments/validator": "^13.7.0",
     "react-devtools-core": "4.22.1",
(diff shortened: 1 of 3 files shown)


Gudahtt:
I noticed the missing constants earlier and tried patching the network controller to no longer use them, but still encountered e2e test failures from usages in other packages. So I've scrapped that idea, and instead have restored the missing constants to controller-utils. See the updated PR description and [this commit](https://github.com/MetaMask/metamask-mobile/pull/6903/commits/7a8039501f7750aa9c65ed48892897776127514d) and [this changelog update](https://github.com/MetaMask/core/pull/1553) for details.

Gudahtt:
Passing e2e tests: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/f82d2734-77a9-4a59-ae32-af0196a8c088

sethkfman (approved):
LGTM

PR #6969, opened 2023-08-04
fix (#17): Unknown labels on transaction screen for MetaMask and impo…

mlallai:
**Description**

fix unknown label on transaction screen by removing any label on normal accounts


mlallai:

PR #6980, opened 2023-08-07
feat: swaps on send flow when amount is insufficient 

tommasini:
**Description**
PROPOSAL:
During the send flow on the amount screen:
* Possibility of swaps when ERC-20 is selected and the amount is insufficient.
* Possibility of Buy when the main Token amount is insufficient.

**UX to improve**
After navigating to the swaps 
* Reason why we do not navigate back to the send flow on the end, it's because the transaction object on redux it's being replaced by the swaps transaction

**Future development proposal**
The transaction object should have a type (swaps, send_flow, dapp) or split the transaction objects, this way we can have more than one transaction in redux state, which could open opportunities to improve the UX, such as this example. 

**Screenshots/Recordings**
Recording on buy: https://recordit.co/NBdij78xev
Recording on swap: https://recordit.co/4kUMifjSk6

**E2E test**
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/55932a50-d34a-4599-9825-cca5aa681230

**Issue**

fixes #???

**Checklist**


bschorchit:
Love this! Could the "Swap" tokens follow the same design pattern as the "Buy more" and be a link within the warning instead of new button?

bschorchit:
This might be outside of scope, but it would be nice to add metrics on this so we can measure its impact and effectiveness. It could be triggering a `Link Clicked` event with properties:
- `location` = `insufficient_funds_warning`
- `text` = `buy_more` or `swap_tokens`

tommasini:
@bschorchit addressed, Here it goes the recording: https://recordit.co/4kUMifjSk6

app/components/Views/SendFlow/Amount/index.js
```diff
@@ -1192,10 +1234,48 @@ class Amount extends PureComponent {
...
+    };
+
+    const navigateToBuyOrSwaps = () => {
+      if (
+        !selectedAsset.isETH &&
+        AppConstants.SWAPS.ACTIVE &&
+        swapsIsLive &&
+        isSwapsAllowed(chainId) &&
+        amountError === strings('transaction.insufficient')
+      ) {
```
NicolasMassart, line 1265: Q: same as for the app/components/UI/Swaps/index.js useEffect question. Find a way to clarify the test conditions. For instance use a variable `isToken` and `isSwapPossible` and `isFundingNeeded` . Then you have this conditions:
```suggestion
      if (
        isToken
        && isSwapPossible
        && isFundingNeeded
      ) {
```
That may be more readable.

app/components/UI/Swaps/index.js
```diff
@@ -368,6 +375,29 @@ function SwapsAmountView({
...
   ]);
 
+  useEffect(() => {
+    if (
+      !isDestinationSet &&
+      initialDestination &&
+      swapsControllerTokens &&
+      swapsTokens?.length > 0 &&
+      !destinationToken
+    ) {
```
NicolasMassart, line 385: Q: could you explain what this composed expression really checks? What is the global meaning of this test? Maybe use a variable with an explicit naming that will hold this value and then use it in the if.

app/components/Views/SendFlow/Amount/index.js
```diff
@@ -1259,7 +1339,27 @@ class Amount extends PureComponent {
...
+                <Text style={[styles.error, styles.underline]}>
+                  {strings('transaction.buy_more')}
+                </Text>
+              )}
+
+              {!selectedAsset.isETH &&
+                AppConstants.SWAPS.ACTIVE &&
+                swapsIsLive &&
+                isSwapsAllowed(chainId) &&
+                amountError === strings('transaction.insufficient') && (
```
NicolasMassart, line 1357: Suggestion: as this is the exact same condition as above, you could have a function that returns this info for both.

app/components/Views/SendFlow/Amount/index.js
tommasini: Thanks for the review @NicolasMassart definitely addressed it!

app/components/UI/Swaps/index.js
tommasini: Thanks for the review @NicolasMassart definitely addressed it!

app/components/Views/SendFlow/Amount/index.js
tommasini: Thanks for the review @NicolasMassart definitely addressed it!

NicolasMassart (dismissed):
Looks good to me.

NicolasMassart (dismissed):

tommasini:
Regression Pipeline running that have 2 swaps test flows https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/8b0d7b14-8b16-4df9-a816-c50d08de6bf1

Important to notice that the swaps e2e are 4 test files, 2 of them are on the smoke e2e and 2 of them are on the regression pipeline

tommasini:
Last recording: https://recordit.co/O7a8J82Xg2

NicolasMassart (dismissed):
Looks good to me.

cortisiko (approved):
this looks good to me. Scenarios I also considered:

- attempting to send funds on a network that does not support buying. Ensuring the alert which takes you to the buy flow is not displayed 
- attempting to send funds on a network that does not support swapping. Ensuring the alert which takes you to the swap flow is not displayed

PR #6998, opened 2023-08-16
refactor: use block tracker to poll incoming transactions

matthewwalsh0:
**Description**

Currently the mobile client uses a version of the core `TransactionController` where incoming transactions are polled on request via a `fetchAll` method called from the frontend in a polling loop and explicitly on events such as the network changing.

In addition, the client is currently responsible for specifying what block number to search from and therefore is also responsible for maintaining it's own object in the persisted state to track the last fetched block number per account per network.

This PR primarily updates the `TransactionController` to a version which both polls internally for incoming transactions using a block tracker referencing the current network, and maintains its own last fetched block number object.

Note that this update is via a patch due to additional breaking changes blocking a direct version update.

As a result, this PR also includes the following changes:

- Remove all usage of the `fetchAll` controller method.
- Remove the `Engine.refreshTransactionHistory` method.
- Use the controller `startIncomingTransactionPolling` and `stopIncomingTransactionPolling` methods to control internal polling.
- Use the `updateIncomingTransactions` controller method when immediate updates are required such as when changing networks.
- Add the `eth-block-tracker` and `@metamask/swappable-obj-proxy` packages to support the patched `TransactionController`.

**Issue**

Fixes [#1195](https://github.com/MetaMask/mobile-planning/issues/1195)

**Checklist**


matthewwalsh0:
@SocketSecurity ignore-all

OGPoyraz (approved):

app/core/Engine.ts
```diff
@@ -570,73 +585,6 @@ class Engine {
...
 
-  refreshTransactionHistory = async (forceCheck: any) => {
-    const { TransactionController, PreferencesController, NetworkController } =
-      this.context;
-    const { selectedAddress } = PreferencesController.state;
-    const { type: networkType } = NetworkController.state.providerConfig;
-    const { networkId } = Networks[networkType];
-    try {
-      const lastIncomingTxBlockInfoStr = await AsyncStorage.getItem(
-        LAST_INCOMING_TX_BLOCK_INFO,
```
vinistevam, line 581: We could also remove the constant `LAST_INCOMING_TX_BLOCK_INFO` from `app/constants/storage.ts`.

vinistevam (approved):
:rocket:  :rocket:   Awesome work! Great to see this on mobile.

PR #7005, opened 2023-08-17
fix: Unhandled exception on uncomplete ledger transaction

stanleyyconsensys:
**Description**
when cancel /speed up transaction with default value,  general ledger disconnect error was outputted, instead it should output correct error message

**Issue**
1. when cancel /speed up transaction, default value for fee suggestion is incorrect  (too low)
2. when cancel /speed up transaction with low value, no specific error handled, instead general ledger disconnect error

**step reproduce**
1. Pair ledger with imported SRP on MMM
2. Access [test dapp](https://metamask.github.io/test-dapp/)
3. Connect the dapp
4. use goerli testnet
5. create a token and approve on Ledger
6. while the contract has not complete
7. cancel / speed up the txn
8. it will say your device got disconnected

***Expected result:*** 
1. correct the fee suggestion -> (not to done in this ticket)
2. add validation error dialog when fee not correct  
3. 
3. even validation pass, the transaction may still have error, should not show disconnected message, instead other  error message

***Actual result:***
error message"your device got disconnected" is showing

### DIFF
--- a/app/components/UI/ActionView/index.js
+++ b/app/components/UI/ActionView/index.js
@@ -52,6 +52,8 @@ export default function ActionView({
   style = undefined,
 }) {
   const { colors } = useTheme();
+  confirmText = confirmText || strings('action_view.confirm');
+  cancelText = cancelText || strings('action_view.cancel');
 
   return (
     <View style={baseStyles.flexGrow}>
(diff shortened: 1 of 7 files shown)


stanleyyconsensys:

PR #7008, opened 2023-08-17
fix: [7.5.0]  nft not importing/hiding

tommasini:
**Description**
NFT was not importing or not being able to hide because we were loosing the chain id value on the nft controller. In order to solve that we will give to the nft controller the chain id value on the engine initialization.

**Screenshots/Recordings**

**E2E test**
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/435a79a0-9d93-453c-9bc0-990cca23cdf0

**Issue**

fixes #???

**Checklist**

### DIFF
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -138,6 +138,11 @@ class Engine {
             listener,
           ),
       });
+      console.log(
+        '   networkController.state.providerConfig.chainId,',
+        networkController.state.providerConfig.chainId,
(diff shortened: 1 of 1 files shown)


chrisleewilcox:
The test dapp NFTs are imported but do not show on the NFTs view.  Still an improvement and will QA Pass this.

chrisleewilcox (approved):

PR #7011, opened 2023-08-17
fix: support importing account while connecting

matthewwalsh0:
**Description**

Resolve issue causing the modal to prematurely close after importing an account while connecting to a dApp.

The issue was that the navigation in the `PermissionApproval` component was running on every render, meaning duplicate navigations were being performed that ultimately meant closing the import success screen closed the entire navigation stack rather than returning to the original connect modal.

See issue for steps to reproduce.

**Screenshots / Recordings**

https://github.com/MetaMask/metamask-mobile/assets/19745142/bb39ce5e-95de-400e-b650-659dd85ea797

**Issue**

Fixes [#1192](https://github.com/MetaMask/mobile-planning/issues/1192)

**Checklist**

### DIFF
--- a/app/components/Approvals/PermissionApproval/PermissionApproval.tsx
+++ b/app/components/Approvals/PermissionApproval/PermissionApproval.tsx
@@ -1,5 +1,5 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import React from 'react';
+import { useEffect, useRef } from 'react';
 import useApprovalRequest from '../../hooks/useApprovalRequest';
 import { ApprovalTypes } from '../../../core/RPCMethods/RPCMethodMiddleware';
 import AnalyticsV2 from '../../../util/analyticsV2';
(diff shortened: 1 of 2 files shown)


vinistevam (approved):

Cal-L:
@matthewwalsh0 Could you provide a video of the fix in action?

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
```diff
@@ -15,28 +15,38 @@ export interface PermissionApprovalProps {
...
-  } = requestData;
+    const {
+      metadata: { id },
+    } = requestData;
 
-  AnalyticsV2.trackEvent(MetaMetricsEvents.CONNECT_REQUEST_STARTED, {
-    number_of_accounts: totalAccounts,
-    source: 'PERMISSION SYSTEM',
-  });
+    if (isProcessing.current) return;
```
Cal-L, line 34: This check implies that this useEffect is still being triggered multiple times. Does it actually trigger more than once or at unexpected times?

matthewwalsh0: It is triggering more than once but not necessarily unexpectedly.

We use the `useEffect` now so we can limit the checks at least to when the dependencies change, but that means the effect still triggers when the `totalAccounts` increases after import, plus when the `props.navigation` instance is changed during the navigation of the import flow.

Hence the `isProcessing` reference acts as a final guard to ensure we don't navigate again until the approval request is gone or there is a new one with a different type.

Cal-L:
I'm a bit worried of using component life cycles to control the permissions approval since leads to unexpected fragile conditions such as the one in the video 

https://github.com/MetaMask/metamask-mobile/assets/10508597/bcb1a2e6-acc2-4d74-9852-7e831250e870

This happens when my lock timer is set to immediate. In the video, the navigation logic is triggered again once a user re-auths, resulting in a frozen/broken state.

Cal-L (approved):
LGTM

PR #7018, opened 2023-08-18
fix: Fix blank screen after locking + unlocking app

Cal-L:
**Description**

This PR fixes a bug from 7.5 regression related to a blank screen state.
[#1193](https://github.com/MetaMask/mobile-planning/issues/1193)

**Screenshots/Recordings**

Video of fix
https://github.com/MetaMask/metamask-mobile/assets/10508597/b9a99b6d-2fe9-4b70-95d5-7c9490e9400b

**Issue**

fixes #???

**Checklist**

### DIFF
--- a/app/core/Authentication/Authentication.ts
+++ b/app/core/Authentication/Authentication.ts
@@ -469,14 +469,13 @@ class AuthenticationService {
    * Logout and lock keyring contoller. Will require user to enter password. Wipes biometric/pin-code/remember me
    */
   lockApp = async (reset = true): Promise<void> => {
-    const { KeyringController, ApprovalController }: any = Engine.context;
+    const { KeyringController }: any = Engine.context;
     if (reset) await this.resetPassword();
     if (KeyringController.isUnlocked()) {
       await KeyringController.setLocked();
     }
     this.authData = { currentAuthType: AUTHENTICATION_TYPE.UNKNOWN };
     this.dispatchLogout();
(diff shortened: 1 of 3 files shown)


matthewwalsh0 (approved):

app/core/Authentication/Authentication.ts
```diff
@@ -469,13 +469,14 @@ class AuthenticationService {
...
   lockApp = async (reset = true): Promise<void> => {
-    const { KeyringController }: any = Engine.context;
+    const { KeyringController, ApprovalController }: any = Engine.context;
     if (reset) await this.resetPassword();
     if (KeyringController.isUnlocked()) {
       await KeyringController.setLocked();
     }
     this.authData = { currentAuthType: AUTHENTICATION_TYPE.UNKNOWN };
     this.dispatchLogout();
+    ApprovalController.clear();
```
matthewwalsh0, line 479: I don't think it warrants addressing right now given these are release blockers, but just for future reference, the `clear` method expects a `EthereumRpcError` error as an argument so when it rejects any pending approval requests promises it can provide a useful error.

In practice though, this is still functional and just rejects with `undefined` which I don't believe warrants another release blocker given the changes of this scenario are slim.

It's also already used in mobile without an argument when handling `eth_requestAccounts` in the middleware.

cortisiko (approved):
fix looks good

PR #6967, opened 2023-08-03
feat: github actions to automatically create and close bug report issue

gauthierpetetin:
**Description**

The purpose of these new Github actions is to automatically:
- create a bug report issue when a new release is cut, i.e. the upon the creation of release branch `release/x.y.z`.
- close the bug report issue when the release branch `release/x.y.z` gets merged

**Screenshots/Recordings**

Recording for bug report issue creation: https://recordit.co/zcmmMXjJPC
Recording for bug report issue closed: https://recordit.co/lkKNtcmXgb

**Issue**

Progresses https://github.com/MetaMask/mobile-planning/issues/1184

## Manual Testing Steps

- Create a public test Github repo
- Add these Github actions to the test repo
- Create a personal access token with "repo" permissions
- In your test repo's settings, add the personal access token as BUG_REPORT_TOKEN in section "Secrets and variables">"Actions">"Secrets">"New repository secret".
- In your test repo, create a new branch with the following format: `release/x.y.z`
- This shall create a new bug report issue in the MM planning repo, called "vx.y.z Bug Report"
- In your test repo, merge the `release/x.y.z` branch into master
- This shall close the bug report issue called "vx.y.z Bug Report" in the MM planning repo

**Checklist**

* [x] There is a related GitHub issue
* [ ] Tests are included if applicable
* [x] Any added code is fully documented

**Other**

- [Same PR for Extension](https://github.com/MetaMask/metamask-extension/pull/20391)

PR #7024, opened 2023-08-21
fix(action): octokit not supported on MetaMask repos

gauthierpetetin:
**Description**

The github action was constantly failing with the following error: octokit/request-action@v2.x is not allowed to be used in MetaMask/metamask-mobile.

The fix is to use a curl instead of using octokit.

**Screenshots/Recordings**

- See this PR: https://github.com/MetaMask/metamask-mobile/pull/6967

**Issue**

- None, it's just a fix to the PR mentioned above.

**Checklist**


tommasini (approved):
LGTM!

PR #7025, opened 2023-08-21
feat: NFT media logic

tommasini:
**Description**
More info [here](https://github.com/MetaMask/mobile-planning/issues/1166)

**Screenshots/Recordings**

**Issue**

fixes #???

**Checklist**


app/store/migrations.js
```diff
@@ -487,6 +487,19 @@ export const migrations = {
     }
     return state;
   },
+
```
NicolasMassart, line 490: C: You could make this state migration even stronger by considering that the state you rely one could be corrupted.
This would prevent crashes and provide default values.
For instance, if `openSeaEnabled` is not defined in the state, `displayNftMedia` will not be set. If you want `displayNftMedia` to have a default value (`true`?), you should do somethings like:
```
PreferencesController.displayNftMedia = PreferencesController.openSeaEnabled || true;
```

Also even it there's really no reason in our state to miss `state.engine.backgroundState.PreferencesController` or `state.user`, a corruption could lead to issues as this migrations takes the existance of them for granted. If we add strong tests and values in each migration, even if it looks very unlikely to happen, we will improve the strength of all our migrations and state and prevent the few migration issues and state corruption that we actually had in the past.
It's clearly not this piece of code specific issue, but a sane habit to take for all our state handling IMO.
You can simply add tests like:
```
if (state.engine?.backgroundState?.PreferencesController)
```
and
```
if (state.user)
```

NicolasMassart (commented):
Nice PR, I added some comments, but nothing blocking you.

app/components/Views/AddAsset/AddAsset.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+});
+
+const initialState = {
+  engine: {
+    backgroundState: initialBackgroundState,
+  },
+};
+
+describe('AddAsset component', () => {
+  it('renders correctly', () => {
```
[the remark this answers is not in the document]
tommasini: What name do you suggest? Just following the pattern of all other components

app/components/Views/AddAsset/AddAsset.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const navigation = useNavigation();
+  const { assetType, collectibleContract } = useParams<AddAssetParams>();
+
+  const {
+    styles,
+    theme: { colors },
+  } = useStyles(styleSheet, {});
+
+  const chainId = useSelector(selectChainId);
+  const displatNftMedia = useSelector(selectDisplayNftMedia);
```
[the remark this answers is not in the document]
tommasini: Good catch!

app/store/migrations.js
tommasini: > PreferencesController.displayNftMedia = PreferencesController.openSeaEnabled || true;

I think this would always fallback to true, because if the openSeaEnabled was false, we will need something as '??' that would only give the right value if the left is null or undefined 

yeah definetely I will add to the migration the conditional operator, amazing tip! Thanks!!!

NicolasMassart (approved):
Looks good to me.

NicolasMassart (approved):
Looks good to me.

cortisiko (approved):
🚀 🚀 🌑  PR looks good from a QA POV

MarioAslau (approved):
Looks good!

app/components/UI/CollectibleOverview/index.js
```diff
@@ -294,8 +305,14 @@ const CollectibleOverview = ({
       nativeEvent: {
         layout: { height },
       },
-    }) => setWrapperHeight(height),
-    [],
+    }) => {
+      //This condition is needed to prevent bouncing when the component is rendered
```
MarioAslau, line 309: Great work on the PR Tomás! 

My only question is in regards to this, while it helps mitigate the effects of the bug, we still don't know why we keep getting different heigh values and the extra re-renders. Would it be worth adding these findings somewhere to keep track of re-rendering issues in the app? Maybe we could find a common thread later

tommasini: Totally!!! Will let @hesterbruikman know, I believe there is a ticket that needs to create to implement a new component for the nft sheet

hesterbruikman: The design for this sheet is owned by a new designer. I would consider it `draft` rather than `ready for dev` because of this. That said, indeed a new design will come at some point

hesterbruikman:
@gauthierpetetin can we remove the `blocked` and `team-mobile-client` label from this one?

gauthierpetetin:
I removed the `blocked` label but it's important to keep the team labels on our PRs and that's what allows us to put together this [metrics diagram](https://lookerstudio.google.com/u/1/reporting/e7e8f90e-72aa-4128-ae01-6305bf3393f4/page/p_ckf409yn7c).

PR #7033, opened 2023-08-22
test: Send ETH to Multisig refactor

seaona:
## Description
This PR does some refactor on the Send ETH Multisig, and can be a reference PR for refactoring the rest of Confirmations testcases. It does the following;:
- Deploy Multisig contract using Ganache Seeder and use the address for the Send
- Remove a repeated Multisig test
- Use fixtures to skip the onboarding process
- Use fixtures to skip the add network for Ganache

https://github.com/MetaMask/metamask-mobile/assets/54408225/e41b4f6c-3b54-48d2-80c4-66dd5e87bdb9

## Screenshots/Recordings
On the right, you can see the deployment transaction and the Multisig contract address.
On the left, you can see the emulator Sending ETH to the newly deployed contract.

image

Screenshot from 2023-08-22 11-10-51

Bitrise run: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/8c137c80-7055-45fa-aede-f98d7a67d1a5

Screenshot from 2023-08-30 18-35-53

**Issue**

fixes https://github.com/MetaMask/MetaMask-planning/issues/1205

## Checklist


e2e/specs/confirmations/send-eth-multisig.spec.js
```diff
@@ -31,30 +35,37 @@ describe(Regression('Send tests'), () => {
     }
     ganacheServer = new Ganache();
     await ganacheServer.start({ mnemonic: validAccount.seedPhrase });
+    ganacheSeeder = new GanacheSeeder(ganacheServer.getProvider());
+    await ganacheSeeder.deploySmartContract(MULTISIG);
+    contractRegistry = ganacheSeeder.getContractRegistry();
+    multisig = contractRegistry.getContractAddress(MULTISIG);
```
seaona, line 41: In the next refactor, we should move these methods into a helpers file and implement the ganache seeder using fixtures, instead of `beforeAll` hooks, in the same way it's done in Extension. Since this will imply several code changes, it's left outside the scope of this PR.

e2e/specs/confirmations/send-eth-multisig.spec.js
```diff
@@ -31,30 +35,36 @@ describe(Regression('Send tests'), () => {
...
   });
 
   afterAll(async () => {
     await ganacheServer.quit();
   });
 
   it('Send ETH to a Multisig address from inside MetaMask wallet', async () => {
-    await importWalletWithRecoveryPhrase();
-    await addLocalhostNetwork();
+    const fixture = new FixtureBuilder().withGanacheNetwork().build();
```
seaona, line 49: initiate the test  skipping Onboarding and with Ganache localhost network in the state

e2e/fixtures/fixture-builder.js
```diff
@@ -517,6 +531,33 @@ class FixtureBuilder {
...
+        chainId: '1337',
+        rpcTarget: 'http://localhost:8545',
+        nickname: 'Localhost',
+        ticker: 'ETH',
+      },
+    };
+
+    fixtures.AccountTrackerController = {
+      accounts: {
+        '0x76cf1CdD1fcC252442b50D6e97207228aA4aefC3': {
```
seaona, line 550: this account corresponds to the one imported in the wallet, set on the vault key

e2e/fixtures/fixture-builder.js
```diff
@@ -517,6 +531,33 @@ class FixtureBuilder {
...
+          balance: '0x3635C9ADC5DEA00000',
```
seaona, line 551: in Ganache we seed the account with 1000ETH, so we change the balance to this value instead of `0x0`.
Was hoping that the balance value was updated automatically but it seems it's not the case at the moment. We should see if this can be changed in the future, so we can rely on balance being updated automatically, instead of injecting the value from fixtures

e2e/specs/confirmations/send-eth-multisig.spec.js
```diff
@@ -1,28 +1,32 @@
...
+import { withFixtures } from '../../fixtures/fixture-helper';
 
 const validAccount = Accounts.getValidAccount();
-const MULTISIG_ADDRESS = '0x0C1DD822d1Ddf78b0b702df7BF9fD0991D6255A1';
+const MULTISIG = 'multisig';
 const AMOUNT_TO_SEND = '0.12345';
 const TOKEN_NAME = root.unit.eth;
 
-describe(Regression('Send tests'), () => {
+describe(Smoke('Send ETH to Multisig'), () => {
```
seaona, line 24: Changing this to `Smoke` temporarily, so we can run the test on bitrise.

e2e/specs/confirmations/send-eth-multisig.spec.js
```diff
@@ -31,30 +35,36 @@ describe(Regression('Send tests'), () => {
...
+    multisig = contractRegistry.getContractAddress(MULTISIG);
   });
 
   afterAll(async () => {
     await ganacheServer.quit();
   });
 
   it('Send ETH to a Multisig address from inside MetaMask wallet', async () => {
-    await importWalletWithRecoveryPhrase();
-    await addLocalhostNetwork();
```
seaona, line 42: we can remove this step now, as network is already in the state and selected

e2e/specs/confirmations/send-eth-multisig.spec.js
```diff
@@ -6,55 +6,67 @@ import TestHelpers from '../../helpers';
...
+import GanacheSeeder from '../../../app/util/test/ganache-seeder';
 import root from '../../../locales/languages/en.json';
+import FixtureBuilder from '../../fixtures/fixture-builder';
+import { withFixtures } from '../../fixtures/fixture-helper';
 
-const validAccount = Accounts.getValidAccount();
-const MULTISIG_ADDRESS = '0x0C1DD822d1Ddf78b0b702df7BF9fD0991D6255A1';
+// SRP corresponding to the vault set in the default fixtures
+const seedPhrase =
+  'drive manage close raven tape average sausage pledge riot furnace august tip';
```
seaona, line 20: this is a test SRP which corresponds to the vault set on default fixtures, so the account imported is seeded with ETH

blackdevelopa (dismissed):

e2e/specs/confirmations/send-eth-multisig.spec.js
NicolasMassart: Maybe add a note about the fact that this SRP is absolutely not a secret because if we search for leaked secrets later, without the context, we may be confused.

NicolasMassart (dismissed):
That's a clean and easy to understand set of changes. Thanks!

e2e/specs/confirmations/send-eth-multisig.spec.js
seaona: thank you for your suggestion @NicolasMassart , that's a very good point. I've just added a clarifying comment

blackdevelopa (approved):
Looks good to me

cortisiko (approved):
Looks good!

jpuri (approved):

PR #7056, opened 2023-08-24
fix: safeguard utill/address functions for undefined arguments

### LINKED ISSUE #6771: [Android BUG]Two minor versions of 7.x that crashed every time I upgraded. View: ActivitvView TypeError: undefined is not an object(evaluating e.toLowerCase)
**Describe the bug**
 **[Android 13]Two minor versions of 7.x that crashed every time I upgraded.** 
 ### `I've had it happen to me 2 times now, and I'm going to crash each time I manually import dozens of wallets. You guys are so disappointing.`
### `Metamask earns so many transaction fees, I hope you do more things that are beneficial to users, please do a complete upgrade test, testing is important, testing is important, testing is important !!!!`

**Screenshots**
Metamask Bug

**To Reproduce**
just upgrade then crashed.
**Expected behavior**

 - Device: Redmi K40Pro
 - OS: Android
 - App Version：
 **1. from 7.0 -> 7.1 crashed. uninstall it and install 7.1**
 **2. from 7.1 -> 7.2 crashed. uninstall it and install 7.2**

#  There's definitely something wrong with your upgrade compatibility, it crashes every time I upgrade!

**Severity**

- Is this tech debt?

jpcloureiro:
**Development & PR Process**

**Description**

This PR aims at preventing the appearance of the ErrorBoundary screen on a specific user flow that seems to be the most affected by [this bug](https://github.com/MetaMask/metamask-mobile/issues/7203)

The user flow we're touching is the navigation to the Transaction screen while the wallet has a non-empty list of transactions to be displayed.

```
`TypeError: undefined is not an object(evaluating e.toLowerCase)`  (Under the error boundary screen "ActivityView"). 
```
is the error we've been seeing the most.

This fix only implements an access check on a couple of util functions inside `app/util/address` that are being used in the Transaction screen components

**Issue**

fixes #6771 

**Checklist**


app/util/address/index.js
```diff
@@ -225,7 +229,7 @@ export function isENS(name) {
  * @param {address} string
  */
 export function resemblesAddress(address) {
-  return address.length === 2 + 20 * 2;
+  return address && address.length === 2 + 20 * 2;
```
tommasini, line 232: Question a little outside of the scope: 
Why 2 +20 *2 and not 42?

tommasini:
The `getAddressAccountType` is used only for analytics purposes, just want to do the reference that the property account_type will be false when the address is undefined on the following events:
* Sign Request Started
* Sign Request Completed
* Sign Request Cancelled
* Send Transaction Started
* Account Renamed
* Contract Address Copied
* Connect Request Started

tommasini (dismissed):
LGTM!

app/util/address/index.js
jpcloureiro: That is great question which I don't have answer for 😁

jpcloureiro:
> The `getAddressAccountType` is used only for analytics purposes, just want to do the reference that the property account_type will be false when the address is undefined on the following events:
> * Sign Request Started
> * Sign Request Completed
> * Sign Request Cancelled
> * Send Transaction Started
> * Account Renamed
> * Contract Address Copied
> * Connect Request Started
> 
> Do you confirm @jpcloureiro ?

Good catch.

That function shouldn't return a boolean.

I'll make the function throw an error if all calls happen to be wrapped inside a try / catch block

app/util/address/index.js
NicolasMassart: I believe it's a habit to decompose numbers into their meaningful components to show where the final number comes from. For instance, instead of 1500000ms (for 25min), devs will write 25 * 60 * 1000 so you can immediately guess this human understandable value as 1000 and 60 are well nown constants in time computation, so the only one you have to care about is 25.

Here 42 is the length of the [ethereum address](https://info.etherscan.com/what-is-an-ethereum-address/) which is composed of a 2 chars header (0x) to show that it's hexadecimal format and 20 bytes (the last 20 bytes of the corresponding public key) but that are represeted by 2 chars for each byte.
So we have an address as a text string of 42 chars composed of 2 (prefix) + 20 (bytes) * 2 (because of each byte being represented as 2 chars in ascii)

Note that 42 is also "Answer to the Ultimate Question of Life, the Universe, and Everything" but that's something else 😉

jpcloureiro:
getAddressAccountType throws an error when `address` is undefined in [24308d6](https://github.com/MetaMask/metamask-mobile/pull/7056/commits/24308d6b49ac439c9e1d428f3b0acf0c79cc37b6)

NicolasMassart (commented):
Q: not sure why you don't cascade the checks in the functions.

Right now, each address check function has to do similar checks (defined, length, valid hex, regex,...).

Why not just one that checks the definition, then the format?
In the end, all of the functions that use an address seems to require a fully valid hex ethereum address. So I feel that one "isValidEthAddress" that does a definition check and a regex match would be enough to be called from all other places.
It would also reduce the number of places where you do checks and make code stronger and easier to understand to me.

jpcloureiro:
> Q: not sure why you don't cascade the checks in the functions.
> 
> Right now, each address check function has to do similar checks (defined, length, valid hex, regex,...).
> 
> Why not just one that checks the definition, then the format? In the end, all of the functions that use an address seems to require a fully valid hex ethereum address. So I feel that one "isValidEthAddress" that does a definition check and a regex match would be enough to be called from all other places. It would also reduce the number of places where you do checks and make code stronger and easier to understand to me.


We should be using [isValidAddress](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/docs/README.md#isvalidaddress) from [@ethereumjs/util](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util) to validate the `selectedAddress` value.

This validation should occur at a root level rather than in the multiple places where it is being used. 
We should also be controlling more user flows based on the validity of `selectedAddress` value.

With that in mind, I prefer to leave those changes outside this PR for the sake of keeping the scope close to the [issue](https://github.com/MetaMask/metamask-mobile/issues/6771)

NicolasMassart (dismissed):
Looks good to me.

tommasini (dismissed):
LGTM!

app/util/address/index.js
```diff
@@ -225,7 +229,7 @@ export function isENS(name) {
  * @param {address} string
```
leotm, line 231:
```suggestion
 * @param {string} address The 42 character address composed of:
 * 2 ('0x': 2 char hex prefix) + 20 (last 20 bytes of public key) * 2 (as each byte is 2 chars in ascii)
```

Screenshot 2023-08-30 at 13 31 17

app/util/address/index.js
leotm: today i learnt something new (;
tidied up our doc comment w @NicolasMassart's explanation
https://github.com/MetaMask/metamask-mobile/pull/7056/commits/0116bc64bcb62dad3a52388c765e66b09eec7644#r1310204725

MarioAslau (approved):
Looks good!

app/util/address/index.js
```diff
@@ -170,6 +172,10 @@ export function isQRHardwareAccount(address) {
  * @returns {String} - Returns address's account type
  */
 export function getAddressAccountType(address) {
+  if (!isValidHexAddress(address)) {
+    throw new Error(`Invalid address: ${address}`);
```
Cal-L, line 176: This throw would still result in an error boundary if `getAddressAccountType` isn't wrapped in a try catch.

app/util/address/index.js
```diff
@@ -149,6 +149,8 @@ export async function importAccountFromPrivateKey(private_key) {
  * @returns {Boolean} - Returns a boolean
  */
 export function isQRHardwareAccount(address) {
+  if (!isValidHexAddress(address)) return false;
```
Cal-L, line 152: We would probably want to throw here since an invalid address is neither true or false WRT being a hardware account.

Cal-L (commented):
Left some comments - Do we know the root of why selectedAddress would be undefined?

hesterbruikman:
Test [recording](https://recordit.co/Z8EF7xSBoL) as shared on Slack Sep 4

app/util/address/index.js
jpcloureiro: Correct. All `getAddressAccountType` calls must bee wrapped in a try / catch

jpcloureiro:
> Left some comments - Do we know the root of why selectedAddress would be undefined?

One of the causes is bad redux hydration.

Cal-L:
> > Left some comments - Do we know the root of why selectedAddress would be undefined?
> 
> Bad redux hydration is one of them

Is there a solution for fixing the root cause (fixing the bad rehydration condition)? Currently, the patch still results in the app showing potentially incorrect states since selectedAddress can still be undefined.

jpcloureiro:
Hey, I've updated the PR description to provide more context around this approach.

The motivation for this PR relies on preventing a specific error that  has been reported multiple times. All of them appear to occur on the transaction screen.

The approach is to allow specific functions to not throw errors when their arguments are undefined rather than preventing the `selectedAddress` from becoming undefined.

Efforts on `selectedAddress` could be tracked on [this issue](https://github.com/MetaMask/metamask-mobile/issues/7203)

cortisiko (approved):
I do not see anything out of the ordinary to report here. ✅

PR #7095, opened 2023-08-31
fix: update blockaid copies

segun:
Fixes [https://github.com/MetaMask/MetaMask-planning/issues/1192](https://github.com/MetaMask/MetaMask-planning/issues/1192)

**Checklist**


ios/Gemfile.lock
```diff
@@ -0,0 +1,11 @@
...
+  remote: https://rubygems.org/
+  specs:
+
+PLATFORMS
+  arm64-darwin-22
+
+DEPENDENCIES
+
+BUNDLED WITH
+   2.4.18
```
jpuri, line 11: @segun : above looks like not a required change.

jpuri (commented):
Looks good except this change:

https://github.com/MetaMask/metamask-mobile/pull/7095/files#diff-2083a642fe77c874e90d9873c92b0f47ddf014d4a57bd0c5ee1368900c3bb9de

jpuri (approved):

locales/languages/en.json
```diff
@@ -2195,9 +2195,8 @@
     "title": "See what's new",
     "blockaid": {
       "title": "Stay safe with security alerts",
-      "description_1": "Steer clear of known scams while still preserving your privacy with security alerts powered by Blockaid.",
+      "description_1": "Steer clear of known scams while still preserving your privacy with security alerts powered by Blockaid on Ethereum Mainnet.",
```
blackdevelopa, line 2198: It will seem that the original text is already translated to other languages. [Here's one](https://github.com/MetaMask/metamask-mobile/blob/4fd873aa5db81017d8cd15709c16970937be94a7/locales/languages/de.json#L2167). If I remember correctly, translations uses the key to check for new additions which can lead to only `en` been correct and others not getting translated/updated. We might have to delete the other translations so it's regenerated or have a sync with the core team for a more efficient approach.

bschorchit:
Given that Infura has added support on other networks to the method we needed, we might not need this anymore.

Issue #6994, opened 2023-08-16
Signing feedback not showing


christopherferreira9:
**Describe the bug**
After signing on a Dapp there should be a modal confirming or denying the success of the sign.

**Screenshots**

**To Reproduce**
1. Build main
2. Go to the test dapp - https://c0f4f41c-2f55-4863-921b-sdk-docs.github.io/test-dapp/
3. Sign

**Expected behavior**
Once you successfully approve the signing you should get a bottom popup feedback

**Smartphone (please complete the following information):**
All

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt?

bschorchit:
@christopherferreira9 could you share more detailed repro steps? I was not able to reproduce this and could see the success feedback message showing up.

bschorchit:
Assigning this to sdk team as they might have more context

christopherferreira9:
Adding a video recording of the issue for reference:
https://github.com/MetaMask/metamask-mobile/assets/104831203/556451a4-b789-429d-9678-f2aa07b17392

christopherferreira9:
This is fixed by https://github.com/MetaMask/metamask-mobile/pull/7102

PR #7102, opened 2023-09-01
fix: sdk connection issues

abretonc7s:
**Description**

- Fixes session error on when reconnecting on existing session (ios).
- Also fixed indirect issue linked to connection stability.

### DIFF
--- a/app/components/Views/SDKSessionsManager/SDKSessionItem.tsx
+++ b/app/components/Views/SDKSessionsManager/SDKSessionItem.tsx
@@ -69,23 +69,24 @@ export const SDKSessionItem = ({
   useEffect(() => {
     let _sessionName = connection.id;
 
-    if (
-      connection.originatorInfo?.title &&
-      connection.originatorInfo?.title?.length > 0
-    ) {
-      _sessionName = connection.originatorInfo.title;
-    } else if (
-      connection.originatorInfo?.url &&
-      connection.originatorInfo?.url?.length > 0
-    ) {
-      _sessionName = connection.originatorInfo.url;
+    const {
+      title,
+      url,
+      icon: _icon,
+    } = connection.originatorInfo || {
+      title: '',
+      url: '',
+      icon: '',
+    };
+
+    if (title?.length > 0) {
+      _sessionName = title;
+    } else if (url?.length > 0) {
+      _sessionName = url;
     }
 
-    if (
-      connection.originatorInfo?.icon &&
-      connection.originatorInfo?.icon?.length > 0
-    ) {
-      setIcon(connection.originatorInfo?.icon);
+    if (_icon && _icon?.length > 0) {
+      setIcon(_icon);
     }
 
     setSessionName(_sessionName);
--- a/app/components/Views/SDKSessionsManager/SDKSessionItem.tsx
+++ b/app/components/Views/SDKSessionsManager/SDKSessionItem.tsx
@@ -67,11 +67,27 @@ export const SDKSessionItem = ({
   const [icon, setIcon] = useState<string>();
 
   useEffect(() => {
-    const _sessionName =
-      connection.originatorInfo?.url ??
-      connection.originatorInfo?.title ??
-      connection.id;
-    setIcon(connection.originatorInfo?.icon);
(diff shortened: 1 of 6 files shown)


app/components/Views/SDKSessionsManager/SDKSessionItem.tsx
```diff
@@ -67,11 +67,27 @@ export const SDKSessionItem = ({
...
+      _sessionName = connection.originatorInfo.url;
+    }
+
+    if (
+      connection.originatorInfo?.icon &&
+      connection.originatorInfo?.icon?.length > 0
+    ) {
+      setIcon(connection.originatorInfo?.icon);
+    }
+
```
omridan159, line 90:
```suggestion
    let _sessionName = connection.id;

    const { title, url, icon } = connection.originatorInfo || {};

    if (title?.length > 0) {
      _sessionName = title;
    } else if (url?.length > 0) {
      _sessionName = url;
    }

    // Update the icon state if icon information is available
    if (icon?.length > 0) {
      setIcon(icon);
    }
```

omridan159: More clean  way

app/core/SDKConnect/SDKConnect.ts
```diff
@@ -413,10 +424,16 @@ export class Connection extends EventEmitter2 {
...
-            await wait(1000);
+          if (!this.receivedDisconnect) {
+            while (!this.isReady) {
+              await wait(1000);
+            }
+            this.sendAuthorized();
+          } else {
+            // Reset state to continue communication after reconnection.
+            this.isReady = true;
+            this.receivedDisconnect = false;
```
omridan159, line 435:
```suggestion
  if (!this.receivedDisconnect) {
    let attempts = 0;

    while (!this.isReady && attempts < 10) {
      await wait(1000);
      attempts++;
    }

    this.sendAuthorized();
  } else {
    // Reset state to continue communication after reconnection.
    this.isReady = true;
    this.receivedDisconnect = false;
  }
```

Are we sure that we don't want to create a max attempts logic?

LMk WDYT? :)

app/components/Views/SDKSessionsManager/SDKSessionItem.tsx
abretonc7s: this would also gives linting ts errors:
eg:'icon.length' is possibly 'undefined'.

omridan159 (dismissed):

elefantel (dismissed):
Looks good.

elefantel (approved):
Looks good

christopherferreira9:
**Context:**
in the flow Connect Dapp -> Approve connection in MMM -> Kill MMM -> Tap sign, the user was being prompt with a connect request instead of the sining prompt. This made that the user would need to approve the connection he did before making it be misleading into making the user think it is a new connection. 
The second issue with this is that if the user taps cancel instead of connect in this resuming prompt the connection would be deleted and the Dapp would never know that so the user gets into a stale state.
The third issue occurs with the Signing feedback mentioned [here](https://github.com/MetaMask/metamask-mobile/issues/6994) that was indirectly caused by this reconnection flow making the appearance of this toaster forever hidden.

**Error Video:**

https://github.com/MetaMask/metamask-mobile/assets/104831203/6c296f77-f9a2-42e4-ab91-b03ca9ef9b84



**Acceptance Criteria:**
- After killing the MetaMask Wallet you're able to resume with sign (meaning that the user is able to tap sign on Dapp and get deeplinked to MetaMask)
- After the MetaMask Wallet resumes the user is prompt with the sign approval
- Approving the sign request presents the success message in MetaMask
- Approving the sign request presents the rejection message in MetaMask


**SDKs and Dapps tested:**
- Android SDK
- Unity SDK
- iOS SDK
- Javascript SDK: Create-React-App; NextJS; ReactNative; ElectronJS; NodeJS

**Video Fix for sign on resume:**
Uploading ios_fix.mov…

**Video fix for the signing feedback:**

https://github.com/MetaMask/metamask-mobile/assets/104831203/0fb2f804-b96a-4b94-bc54-304cd7d05b2d

PR #7068, opened 2023-08-28
Adding blockaid banner to confirmation pages

jpuri:
Adding bloakaid banner to confirmation pages.

Signature request pages:

Screenshot 2023-08-29 at 3 49 31 PM

Confirmation pages:

Screenshot 2023-08-29 at 3 49 16 PM

Followup task to style confirmation pages: https://app.zenhub.com/workspaces/-confirmations-secure-ux-6245e6e2348677001213b8d2/issues/gh/metamask/metamask-planning/1249

PR #7104, opened 2023-09-01
feat: Blockaid dapp metrics

segun:
**Description**

Adding analytics to Blockaid will help us understand and evaluate the performance of this feature.

For users that have the Blockaid feature enabled, on performing a dapp transaction, the following should happen

1. Add a security_alert_response property to Transactions and Signature events, the value of this property should be the [result_type parameter that is returned by Blockaid](https://wobbly-nutmeg-8a5.notion.site/PPOM-Validation-Schema-Attack-Types-555e7f6a5b6c4b6caa9ced59561c8667) (e.g. set_approval_for_all_famring).
2. Add a security_alert_reason property to Transactions and Signature events, the value of this property should be the [reason parameter that is returned by Blockaid](https://wobbly-nutmeg-8a5.notion.site/PPOM-Validation-Schema-Attack-Types-555e7f6a5b6c4b6caa9ced59561c8667). When there's no value for this property (because user has the feature disabled or Blockaid response does not contain a reason value), we should pass it the value not_applicable.
Create ui_customizations list property for transactions and signatures events on mobile - this property should be similar to the one created here on extension [MetaMask/metamask-extension#17688](https://github.com/MetaMask/metamask-extension/pull/17688). This property should be a type list property (more info [here](https://help.mixpanel.com/hc/en-us/articles/115004547063-Properties-Supported-Data-Types#list).
3. Whenever we display a warning to the user as a result of Blockaid flagging a request as malicious, we should add the flagged_as_malicious value to the ui_customizations property in the respective Transactions or Signature events

**Issue**

https://github.com/MetaMask/MetaMask-planning/issues/929

Bitrise: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/12a125d8-2d06-4184-8f0d-b5a0977a954c

**Checklist**

* [x] Bitrise smoke tests passed


app/util/blockaid/index.ts
```diff
@@ -1,3 +1,37 @@
...
+  Reason,
+  ResultType,
+  SecurityAlertResponse,
+} from '../../components/UI/BlockaidBanner/BlockaidBanner.types';
+
 // eslint-disable-next-line import/prefer-default-export
 export const showBlockaidUI = () =>
   process.env.MM_BLOCKAID_UI_ENABLED === 'true';
+
+export const getAdditionalMetricsParams = (
```
jpuri, line 11: The function name is very generic.

app/components/Views/Send/index.js
```diff
@@ -640,9 +641,11 @@ class Send extends PureComponent {
    * Call Analytics to track confirm started event for send screen
    */
   trackConfirmScreen = () => {
+    const additionalParams = getAdditionalMetricsParams(this.props.transaction);
+
     Analytics.trackEventWithParameters(
       MetaMetricsEvents.TRANSACTIONS_CONFIRM_STARTED,
-      this.getTrackingParams(),
+      { ...this.getTrackingParams(), ...additionalParams },
```
jpuri, line 648: Similarly here change can go into `this.getTrackingParams(),`.

jpuri (dismissed):
LGTM

segun:
Once [this](https://github.com/MetaMask/metamask-mobile/pull/7068) is merged, rebase to main

app/lib/ppom/ppom.html.js
```diff
@@ -1,15 +1,15 @@
 var A = {
```
segun, line 1: Not entirely sure why this file is included in this change. I think yarn lint:fix is reformatting it.

blackdevelopa: If we revert the change in this file locally and pushed without running `yarn lint`, would that overwrite this in your PR?

app/components/UI/PersonalSign/PersonalSign.tsx
```diff
@@ -67,7 +80,7 @@ const PersonalSign = ({
       MetaMetricsEvents.SIGN_REQUEST_STARTED,
       getAnalyticsParams(),
```
blackdevelopa, line 81: Should we be passing the expected [props](https://github.com/MetaMask/metamask-mobile/pull/7104/files#diff-9eb780c5c84ee4e372468c43a809dfd448ac21398bcc485ad9a1ac85a68b3d60R17) to `getAnalyticsParams`?

segun: This one doesn't take any props. I'm not entirely sure why PersonalSign has a different separate method to do this separate from the rest....but this is analytics forte and that question is beyond the scope of this PR.

app/components/UI/SignatureRequest/index.js
```diff
@@ -165,6 +167,20 @@ class SignatureRequest extends PureComponent {
...
+      getAnalyticsParams(
+        {
+          from: fromAddress,
+          securityAlertResponse,
+        },
+        'signature_request',
+      ),
+    );
+  };
+
```
jpuri, line 183: Just curious: was `SIGN_REQUEST_STARTED` event not captured before ?

app/components/Views/Send/index.js
```diff
@@ -703,6 +713,7 @@ class Send extends PureComponent {
           (selectedAsset.symbol || selectedAsset.contractName)) ||
         'ETH',
       assetType,
+      ...blockaidParams,
```
jpuri, line 716: This may break if `blockaidParams` is `undefined`

app/lib/ppom/ppom.html.js
jpuri: Plz revert changes to this file, this is file provided by blockaid team and audited by diligence team and should be preserved exactly as audited.

app/util/confirmation/signatureUtils.js
```diff
@@ -9,19 +9,32 @@ import { InteractionManager } from 'react-native';
...
     const { currentPageInformation } = messageParams;
     const chainId = selectChainId(store.getState());
     const url = new URL(currentPageInformation?.url);
+
+    let blockaidParams = {};
+    if (isBlockaidFeatureEnabled()) {
+      blockaidParams = getBlockaidMetricsParams(
+        messageParams.securityAlertResponse,
+      );
+    }
```
jpuri, line 28: `if` condition above can be omitted as it is taken care of in `getBlockaidMetricsParams` function.

jpuri (changes requested):
Please revert changes in `app/lib/ppom/ppom.html.js`

app/components/UI/SignatureRequest/index.js
segun: I was curious too, this method exists with the parameters, but all the place the method was called, it was called without the parameters. Honestly not sure why they call with without the parameters.

I need to use the method in this PR and there's [another PR](https://github.com/MetaMask/metamask-mobile/pull/7120) that specified the event values, so I followed the same value naming in that PR to call this method with the proper parameters.

Alternatively, I could leave the method as is and use something else for blockaid stuffs. But I don't think that's optimal.

jpuri (approved):

app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -535,6 +539,17 @@ class ApproveTransactionReview extends PureComponent {
...
+      if (showBlockaidUI()) {
+        const blockaidParams = getBlockaidMetricsParams(
+          transaction.securityAlertResponse,
+        );
+
+        params = {
+          ...params,
+          ...blockaidParams,
+        };
+      }
```
jpuri, line 552: Nit: a way to simplify using may `if` conditions could be `getBlockaidMetricsParams` return empty object if blockaid is not enabled or `transaction.securityAlertResponse` is not defined.

Thus we do not need to check `if (showBlockaidUI())` at many places.

blackdevelopa (approved):

PR #7105, opened 2023-09-01
fix: Fix all Engine type errors

Gudahtt:
**Development & PR Process**

**Description**

All errors in Engine.ts have been resolved. This file is frequently imported throughout the application, so fixing type errors here is essential for making progress with eliminating type errors.

A few errors were caused by dependency issues that will be resolved soon after additional dependency updates, but I made note of some cases that require further investigation and fixes.

As part of this effort, a `EngineState` type has been added as well. This will be useful in later refactors related to controller upgrades.

**Issue**

This relates to https://github.com/MetaMask/mobile-planning/issues/1226

**Checklist**

### DIFF
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -920,6 +920,11 @@ class Engine {
   }
 }
 
+/**
+ * Assert that the given Engine instance has been initialized
+ *
+ * @param instance Either an Engine instance, or null
+ */
 function assertEngineExists(
   instance: Engine | null,
 ): asserts instance is Engine {
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -920,25 +920,27 @@ class Engine {
   }
 }
 
+function assertEngineExists(
+  instance: Engine | null,
+): asserts instance is Engine {
+  if (!instance) {
+    throw new Error('Engine does not exist');
+  }
+}
+
 let instance: Engine | null;
 
 export default {
   get context() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     return instance.context;
   },
   get controllerMessenger() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     return instance.controllerMessenger;
   },
   get state() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     const {
       AccountTrackerController,
       AddressBookController,
@@ -996,27 +998,19 @@ export default {
     };
   },
   get datamodel() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     return instance.datamodel;
   },
   getTotalFiatAccountBalance() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     return instance.getTotalFiatAccountBalance();
   },
   hasFunds() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     return instance.hasFunds();
   },
   resetState() {
-    if (!instance) {
-      throw new Error('Engine does not exist');
-    }
+    assertEngineExists(instance);
     return instance.resetState();
   },
   destroyEngine() {
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -1,37 +1,59 @@
 /* eslint-disable @typescript-eslint/no-shadow */
 import {
   AccountTrackerController,
+  AccountTrackerState,
   AssetsContractController,
   CurrencyRateController,
+  CurrencyRateState,
   CurrencyRateStateChange,
   GetCurrencyRateState,
   GetTokenListState,
   NftController,
   NftDetectionController,
+  NftState,
   TokenBalancesController,
+  TokenBalancesState,
   TokenDetectionController,
   TokenListController,
+  TokenListState,
   TokenListStateChange,
   TokenRatesController,
+  TokenRatesState,
   TokensController,
+  TokensState,
 } from '@metamask/assets-controllers';
-import { AddressBookController } from '@metamask/address-book-controller';
-import { ControllerMessenger } from '@metamask/base-controller';
+import {
+  AddressBookController,
+  AddressBookState,
+} from '@metamask/address-book-controller';
+import { BaseState, ControllerMessenger } from '@metamask/base-controller';
 import { ComposableController } from '@metamask/composable-controller';
 import {
   KeyringController,
+  KeyringState,
   SignTypedDataVersion,
 } from '@metamask/keyring-controller';
 import {
   NetworkController,
   NetworkControllerActions,
   NetworkControllerEvents,
+  NetworkState,
 } from '@metamask/network-controller';
-import { PhishingController } from '@metamask/phishing-controller';
-import { PreferencesController } from '@metamask/preferences-controller';
-import { TransactionController } from '@metamask/transaction-controller';
+import {
+  PhishingController,
+  PhishingState,
+} from '@metamask/phishing-controller';
+import {
+  PreferencesController,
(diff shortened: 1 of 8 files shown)


app/util/number/index.js
```diff
@@ -524,7 +524,7 @@ export function addCurrencySymbol(
 /**
  * Converts wei expressed as a BN instance into a human-readable fiat string
  *
- * @param {number} wei - BN corresponding to an amount of wei
+ * @param {number|string} wei - BN corresponding to an amount of wei
```
Gudahtt, line 527: We were passing a string in for this parameter in at least one case. This parameter gets passed to `fromWei` which already accepts strings or numbers, so this was fine

app/core/Permissions/specifications.js
```diff
@@ -73,7 +73,6 @@ export const getCaveatSpecifications = ({ getIdentities }) => ({
  *
  * @param {{
  *   getAllAccounts: () => Promise<string[]>,
- *   getIdentities: () => Record<string, Identity>,
```
Gudahtt, line 76: This parameter did not exist in practice, this must be leftover

app/core/Engine.ts
```diff
@@ -78,10 +103,14 @@ import {
   SignatureControllerActions,
   SignatureControllerEvents,
 } from '@metamask/signature-controller';
-import { Json } from '@metamask/controller-utils';
+import { hasProperty, Json } from '@metamask/controller-utils';
+import { SwapsState } from '@metamask/swaps-controller/dist/SwapsController';
```
Gudahtt, line 107: This package should be updated to export this type. We need the state type to work with the controller effectively.

app/core/Engine.ts
```diff
@@ -109,15 +164,42 @@ class Engine {
...
+    AddressBookController: AddressBookController;
+    ApprovalController: ApprovalController;
+    AssetsContractController: AssetsContractController;
+    CurrencyRateController: CurrencyRateController;
+    GasFeeController: GasFeeController;
+    KeyringController: KeyringController;
+    NetworkController: NetworkController;
+    NftController: NftController;
+    NftDetectionController: NftDetectionController;
+    // TODO: Fix permission types
```
Gudahtt, line 182: The warnings are suppressed here because of the use of `any`. It wasn't possible to fix them without converting more files from JavaScript to TypeScript

app/core/Engine.ts
```diff
@@ -217,6 +303,7 @@ class Engine {
         ),
       },
       {
+        // @ts-expect-error NftController constructor config type is wrong
```
Gudahtt, line 306: The constructor uses `BaseConfig` rather than `NftConfig`, even though the class uses `NftConfig`.

app/core/Engine.ts
```diff
@@ -268,6 +363,7 @@ class Engine {
         ),
       getCurrentNetworkEIP1559Compatibility: async () =>
         await networkController.getEIP1559Compatibility(),
+      // @ts-expect-error Incompatible string types, fixed in upcoming version
```
Gudahtt, line 366: Should be fixed in the core v53 upgrade

app/core/Engine.ts
```diff
@@ -442,7 +542,9 @@ class Engine {
       }),
       new SwapsController(
         {
+          // @ts-expect-error TODO: Resolve mismatch between gas fee and swaps controller types
```
Gudahtt, line 545: These two packages both use the same type for this function, but they're out-of-sync with each other. Requires further investigation

app/core/Engine.ts
```diff
@@ -520,27 +626,38 @@ class Engine {
...
 
     const {
       NftController: nfts,
       KeyringController: keyring,
       TransactionController: transaction,
     } = this.context;
 
-    nfts.setApiKey(process.env.MM_OPENSEA_KEY);
+    if (process.env.MM_OPENSEA_KEY) {
+      nfts.setApiKey(process.env.MM_OPENSEA_KEY);
```
Gudahtt, line 659: Calling `setApiKey` with a falsy value should be equivalent to never calling it at all, so I don't expect this to have any impact

app/core/Engine.ts
```diff
@@ -623,7 +740,6 @@ class Engine {
       chainId: NetworkController.state?.providerConfig?.chainId,
       pollCountLimit: AppConstants.SWAPS.POLL_COUNT_LIMIT,
     });
-    TransactionController.configure({ provider });
```
Gudahtt, line 626: This controller doesn't accept a `provider` configuration property anymore, so this call wasn't necessary

app/core/Engine.ts
```diff
@@ -680,6 +792,8 @@ class Engine {
                 )
               : undefined);
           const tokenBalanceFiat = balanceToFiatNumber(
+            // TODO: Fix this by handling or eliminating the undefined case
+            // @ts-expect-error This variable can be `undefined`, which would break here.
             tokenBalance,
```
Gudahtt, line 797: This seems like a legitimate bug. Looks like it'd crash for tokens that have no balance.

app/core/Engine.ts
```diff
@@ -702,6 +816,7 @@ class Engine {
       const {
         engine: { backgroundState },
       } = store.getState();
+      // TODO: Check `allNfts[currentChainId]` property instead
```
Gudahtt, line 819: This was a type error highlighted in a later PR that improves type safety further. It's a bug, there is no `nfts` property

app/core/Engine.ts
```diff
@@ -764,8 +879,6 @@ class Engine {
     TokenRatesController.update({ contractExchangeRates: {} });
 
     TransactionController.update({
-      internalTransactions: [],
-      swapsTransactions: {},
```
Gudahtt, line 768: Don't think `swapsTransactions` ever existed here? And `internalTransactions` was removed recently

app/core/Engine.ts
```diff
@@ -806,16 +919,25 @@ class Engine {
   }
 }
 
-let instance: Engine;
+let instance: Engine | null;
 
 export default {
   get context() {
-    return instance?.context;
```
Gudahtt, line 813: Returning `instance.context` OR `undefined` was forcing us to handle the `undefined` case throughout the application. But there was no way to handle it in most places. The rest of the codebase assumes that it won't run until after initialization.

It seemed easiest to check and throw an error here explicitly, so that TypeScript lets us assume that it's defined. And if it's ever called in a situation where it wasn't defined, it already would have thrown an error when you tried accessing properties on it or deconstructing it.

It's possible this could break something if we were calling `Engine.context?.[property]` somewhere (e.g. using optional chaining or conditions to ensure it was defined first). But I have searched every single file that uses `Engine` and found no examples of this. If we do find an example though, we can add a `get maybeContext()` getter or method or something like that, to handle that case without inconveniencing us everywhere else.

app/core/Engine.ts
```diff
@@ -873,15 +996,27 @@ export default {
...
     return instance.getTotalFiatAccountBalance();
   },
   hasFunds() {
+    if (!instance) {
+      throw new Error('Engine does not exist');
+    }
     return instance.hasFunds();
   },
   resetState() {
+    if (!instance) {
```
sethkfman, line 1017: Consider creating a method for doesEngineExist() to reduce code duplication.

Cal-L (dismissed):
LGTM

app/core/Engine.ts
Gudahtt: Good idea! Done in https://github.com/MetaMask/metamask-mobile/pull/7105/commits/01df7eefbfefe6423efe329b049da1782919a77f

app/core/Engine.ts
```diff
@@ -806,16 +920,32 @@ class Engine {
   }
 }
 
-let instance: Engine;
+/**
+ * Assert that the given Engine instance has been initialized
+ *
+ * @param instance Either an Engine instance, or null
```
Gudahtt, line 926: Typo in JSDoc comment:
```suggestion
 * @param instance - Either an Engine instance, or null
```

sethkfman (approved):
LGTM

Issue #6950, opened 2023-08-01
[TS] Fix TS errors in `util` directory


Cal-L:
### **Description**

Fix Typescript errors in the `util` directory

### **Technical Details**

Number of errors by file
```
     1  app/util/address/index.test.ts:14
    23  app/util/conversion/index.test.ts:11
    22  app/util/custom-gas/index.test.ts:19
     1  app/util/date/index.test.ts:42
     3  app/util/general/index.test.ts:24
     4  app/util/networks/handleNetworkSwitch.ts:32
     2  app/util/networks/index.test.ts:108
    30  app/util/number/index.test.ts:108
     1  app/util/password/index.test.ts:6
     8  app/util/sync/index.test.ts:109
     1  app/util/tokens/tokens.test.ts:4
     7  app/util/transactions/index.test.ts
```

### **Acceptance Criteria**

-  running `yarn tsc` should not show any issues in the `util` directory

Issue #5993, opened 2023-03-18
☂️ Fix remaining TS/TSX static type errors


leotm:
**Describe the issue**
Increase ability to catch bugs at compile time. I.e. identify and resolve faster.

This is a follow-up to
- https://github.com/MetaMask/metamask-mobile/issues/5878
- https://github.com/MetaMask/metamask-mobile/issues/5881

_which enables TSC to emit our remaining TS/TSX type errors to fix_
_~570ish errors in ~90ish files, tracked in our internal google sheet_

+ https://github.com/MetaMask/metamask-mobile/pull/6080

once complete at 0 errors (and 0 warnings ideally)
we can re-enable TSC in our pre-commit hooks and CI
- https://github.com/MetaMask/metamask-mobile/issues/4605
  - already partially complete in https://github.com/MetaMask/metamask-mobile/pull/7100

then later upgrade to TypeScript v5 once our MM shared ESLint config is on v5
- https://github.com/MetaMask/metamask-mobile/issues/5984

a good long-term follow-up would be
- converting remaining `.js` files to `ts` remains, a huge task worthy of it's own umbrella issue / epic
  - e.g. _app/util/transactions/index.js_ -> _app/util/transactions/index.ts_ (no changes) exposes
  - _Found 234 errors in 48 files_ -> _Found 394 errors in 50 files_

**Screenshots**

In issues above ^

**To Reproduce**

`yarn tsc`

**Expected behavior**

0 errors
0 warnings (ideally)

**Smartphone (please complete the following information):**

N/A

-------------------------------------------------------------
_to be added after bug submission by internal support / PM_
**Severity**

- How critical is the impact of this bug on a user?
- Add stats if available on % of customers impacted
- Is this visible to all users?
- Is this tech debt? yes

hesterbruikman:
@leotm @Cal-L 

Added a note to give context to this work and make it easier to discuss relevance:

> "Increase ability to catch bugs at compile time. I.e. identify and resolve faster."


Any other reasons for why this work should be prioritised? Dependencies with libraries?

leotm:
> Any other reasons for why this work should be prioritised?

we could add correct types make it easier (and more welcoming) working on our codebase
as well as getting VS Code clean/free of type errors (linted in red @sethkfman raised originally)

leotm:
> Dependencies with libraries?

[...]

PR #7109, opened 2023-09-01
fix: Fix Typescript errors in utils dir

Cal-L:
**Description**

This PR fixes all TS errors in the `utils` directory

**Screenshots/Recordings**

TS errors after running `yarn tsc`
image

**Issue**

#6950 

**Checklist**


Cal-L:
E2E Smoke tests - https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/be6ab37a-5f25-4c55-8e28-be0964fed36c

MarioAslau (dismissed):
LGTM!

app/util/password/index.ts
```diff
@@ -2,9 +2,8 @@ import SecureKeychain from '../../core/SecureKeychain';
 import Engine from '../../core/Engine';
 
 export const MIN_PASSWORD_LENGTH = 8;
-type PasswordStrength = 0 | 1 | 2 | 3 | 4;
 
-export const getPasswordStrengthWord = (strength: PasswordStrength) => {
+export const getPasswordStrengthWord = (strength: number) => {
```
Gudahtt, line 6: This would provide no type warning if a number other than 0-4 was passed in - it would fail silently. Perhaps it would be better to add a runtime error if an unrecognized number is provided? Or handle the edge cases explicitly (e.g. error for below zero, but consider anything above 4 "strong")

app/util/number/index.test.ts
```diff
@@ -377,7 +378,8 @@ describe('Number utils :: hexToBN', () => {
 
 describe('Number utils :: isBN', () => {
   it('isBN', () => {
-    expect(isBN('0x539')).toEqual(false);
+    const notABN = '0x539' as any;
```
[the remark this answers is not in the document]
Gudahtt: The `isBN` function seemingly isn't meant to accept a string, so this test is covering invalid input. Using `as any` seems like a good strategy in general for "invalid input" tests, since any of these cases would be flagged and prevented by TypeScript anyway in theory.

If string input is expected, we should definitely update `isBN` accordingly though

NicolasMassart (dismissed):
I like this cleanup! Thanks!

app/util/transactions/index.js
```diff
@@ -127,7 +127,7 @@ const actionKeys = {
  * @param {Object} opts - Optional asset parameters
  * @returns {String} - String containing the generated transfer data
  */
-export function generateTransferData(type, opts) {
+export function generateTransferData(type = undefined, opts = {}) {
```
leotm, line 130: was going to suggest renaming to `.ts` (like the test file), then

```suggestion
export function generateTransferData(
  type = undefined,
  opts = {},
): asserts type {
```

but the rename exposes _a lot_ more type errors 😅

so noted it here as a worthy long-term follow-up
- https://github.com/MetaMask/metamask-mobile/issues/5993

app/util/transactions/index.js
```diff
@@ -127,7 +127,7 @@ const actionKeys = {
...
   if (!type) {
     throw new Error('[transactions] type must be defined');
```
leotm, line 132: nit

```suggestion
    throw new TypeError(
      `[transactions] type must be defined, but received ${type}`,
    );
```

leotm (dismissed):

app/util/transactions/index.js
Cal-L: Updated

app/util/password/index.ts
Cal-L: Good catch. Updated to handle unrecognized strength (less than 0) and anything above 4

app/util/date/index.test.ts
```diff
@@ -37,9 +37,9 @@ describe('Date util :: msBetweenDates', () => {
     const mockDate = new Date();
     const spy = jest
       .spyOn(global, 'Date')
-      .mockImplementation(function (...args: any[]) {
-        if (args.length) {
-          return new DateReal(...args);
+      .mockImplementation(function (value: string | number | Date) {
```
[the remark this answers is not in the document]
Cal-L: Clean suggestion. Updated!

app/util/networks/index.test.ts
```diff
@@ -103,17 +103,21 @@ describe('NetworkUtils::getNetworkTypeById', () => {
     try {
       getNetworkTypeById();
     } catch (error) {
-      expect(error.message).toEqual(NetworkSwitchErrorType.missingNetworkId);
+      if (error instanceof Error) {
```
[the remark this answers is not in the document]
Cal-L: ah, you're right. Updated!

app/util/number/index.test.ts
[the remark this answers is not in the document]
Cal-L: Removed cast and updated input type to include string type

app/util/number/index.js
```diff
@@ -94,7 +94,7 @@ export function fromWei(value = 0, unit = 'ether') {
  * Converts token minimal unit to readable string value
  *
  * @param {number|string|Object} minimalInput - Token minimal unit to convert
- * @param {string} decimals - Token decimals to convert
+ * @param {numberstring} decimals - Token decimals to convert
```
Gudahtt, line 97: Typo:
```suggestion
 * @param {number|string} decimals - Token decimals to convert
```

Cal-L: Fixed ✌️

Gudahtt (approved):
LGTM!

PR #7118, opened 2023-09-04
chore: dedupe babel packages

legobeat:
Reduced `node_modules` by ~400MB.

Cuts package install time into almost half:
- CI yarn setup [before](https://github.com/MetaMask/metamask-mobile/actions/runs/6102573379/job/16561390690): `3m 57s` 
- CI yarn setup [after](https://github.com/MetaMask/metamask-mobile/actions/runs/6106203624/job/16570974868?pr=7118): `2m 2s`

PR #7127, opened 2023-09-04
chore: Dedupe ethereumjs packages

legobeat:
- This dedupes packages with the `@ethereumjs/` and `ethereumjs-` prefixes. It does not upgrade them at this point, though an upgrade of `@ethereumjs/` packages can and should be made. 

- Also bumps `ethereumjs-wallet` from `1.0.1` to `1.0.2`.
  - _(This is a legacy package which is deprecated in favor of `@ethereumjs/wallet`. Replacing it properly is out of scope here)_

Based on:
  - [x] #7118

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1705,23 +1705,7 @@
     bufio "^1.0.7"
     chai "^4.3.4"
 
-"@ethereumjs/common@^2.0.0":
-  version "2.6.0"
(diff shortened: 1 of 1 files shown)


leotm (approved):
lgtm, same results locally
https://github.com/MetaMask/metamask-mobile/commits/deps-dedupe-ethereumjs-test
https://www.diffchecker.com/xxg5GG56/

PR #7128, opened 2023-09-04
deps: Dedupe `ethers`@5.*

legobeat:
Based on:

- [x] #7127 
  - [x] #7118  

Dedupes all instances of `ethers`@`^5`

PR #7129, opened 2023-09-04
fix(deps): Dedupe ethereum-cryptography

legobeat:
Dedupes old `@noble/` versions

#### Based on:
- [x] #7128 
  - [x] #7127 
    - [x] #7118  

#### Blocking
- #7131


leotm (changes requested):
think we want to preserve all `keccak` and `secp256k1` scripts on `true` right?

noting this earlier in:
- https://github.com/MetaMask/metamask-mobile/pull/7032

if so, spotted couple prev wrong values in https://github.com/MetaMask/metamask-mobile/pull/7267#discussion_r1356501314
we could also update those to `true` here (or separate pr)

edit: some more info i found
- https://github.com/cryptocoinjs/secp256k1-node/blob/master/package.json#L31-L33 (spotted you're a contributor @kumavis)
- https://github.com/cryptocoinjs/keccak/blob/master/package.json#L26-L28
> `"install": "node-gyp-build || exit 0"`

nb: we also use https://github.com/paulmillr/noble-secp256k1/blob/main/package.json#L15-L23 but unrelated here, since not containing any relevant npm life cycle scripts lavamoat is concerned with

leotm (approved):
⚡ speed, LGTM

nb: potential lavamoat/allow-scripts improvement - one rule to rule them all
e.g. 1 rule instead of tons to set `keccak` to `true` recursively/transitively

leotm:
i noticed these take a while to run

Screenshot 2023-10-17 at 1 22 47 pm

so noting some thoughts to discuss as follow-up
- why/how does deduping (ethereum-cryptography) cause lavamoat to detect lots more recursive/transitive keccak/secp256k1 scripts?
- do we rly need to run all of them? or are some of them enough (like before)
  - if they are all needed, can we improve the perf?

PR #7130, opened 2023-09-04
chore(deps): dedupe semver

legobeat:

PR #7123, opened 2023-09-04
chore: Deduplicate dependencies

Gudahtt:
**Development & PR Process**
1. Follow MetaMask [Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/coding_guidelines/CODING_GUIDELINES.md)
2. Add `release-xx` label to identify the PR slated for a upcoming release (will be used in release discussion)
3. Add `needs-dev-review` label when work is completed
4. Add the appropiate QA label when dev review is completed
    - `needs-qa`: PR requires manual QA.
    - `No QA/E2E only`: PR does not require any manual QA effort. Prior to merging, ensure that you have successful end-to-end test runs in Bitrise. 
    - `Spot check on release build`: PR does not require feature QA but needs non-automated verification. In the description section, provide test scenarios. Add screenshots, and or recordings of what was tested.
5. Add `QA Passed` label when QA has signed off (Only required if the PR was labeled with `needs-qa`)
6. Add your team's label, i.e. label starting with `team-` (or `external-contributor` label if your not a MetaMask employee)

**Description**

A `deduplicate` script has been added to `package.json`, which allows deduplicating dependencies in the lockfile. This feature is built-in as part of Yarn v3, but on Yarn v1 it's provided by the `yarn-deduplicate` package.

It has been run, so the lockfile has been purged of unnecessary copies of dependencies. This means many dependencies have been updated, but it should result in no breakages as long as all affected packages are SemVer-compliant or handled by one of our resolutions.

A CI check has been added to ensure dependencies are always deduplicated. If duplicates are detected, the error from CI will explain how to remove them.

### Before

```
$ du -hs ./node_modules/
3.2G    ./node_modules/
```

### After

```
$ du -hs ./node_modules/                                                                                                  
2.2G    ./node_modules/
```

**Issue**

No related issue

**Checklist**

* [ ] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

PR #7131, opened 2023-09-04
chore: Deduplicate remaining dependencies; add deduplicate script

legobeat:
#7123 but split up dedupes and rebased on:
- [x] #7130 
- [x] #7129 
  - [x] #7128 
    - [x] #7127 
      - [x] #7118  

**Description**

A `deduplicate` script has been added to `package.json`, which allows deduplicating dependencies in the lockfile. This feature is built-in as part of Yarn v3, but on Yarn v1 it's provided by the `yarn-deduplicate` package.

It has been run, so the lockfile has been purged of unnecessary copies of dependencies. This means many dependencies have been updated, but it should result in no breakages as long as all affected packages are SemVer-compliant or handled by one of our resolutions.

A CI check has been added to ensure dependencies are always deduplicated. If duplicates are detected, the error from CI will explain how to remove them.

**Issue**

No related issue

**Checklist**


leotm (approved):
🧹🤖

PR #7172, opened 2023-09-08
Feat: Fix the Message Sign issue for ledger devices

dawnseeker8:
**Description**

After rebase from `main` branch to `ledger-intergtation-main` branch, the whole ledger message sign including `eth_sign`, `personal_sign`, `eth_signTypedDataV1` to V4 all break, this PR include fix from above operation.

following tasks have been done:
1. add two missing patch files in current branch but exists in `main` branch (`patches/@metamask+signature-controller++@metamask+message-manager+7.0.1.patch`, `patches/@metamask+signature-controller+4.0.1.patch`)
2. add `rpcEvent` actions and reducer to track the difference stage of whilelisted RPC event stage, currently mainly for `signing` relative RPC usage..
3. modify `LedgerMessageSignModal.tsx` to subscribe the `signingEvents` in redux store. to perform dismissModal. 
4. refactory the RPCMethodMiddleware to inject the track RPC stage dispatch event in middleware code to listen to all whitelisted RPC methods.
5. fix the code errors in MessageSIgn.tsx and PersonalSign.tsx fils for navigation.

**Caution** unit tests for testing reducer and RPCMethodMiddleware.ts will be done in another unit test branch.


app/components/UI/LedgerModals/LedgerMessageSignModal.tsx
```diff
@@ -25,45 +29,46 @@ export const createLedgerMessageSignModalNavDetails =
...
-      rawSignature = await KeyringController.signTypedMessage(
-        messageParams,
-        version,
-      );
-    }
+  const executeOnLedger = useCallback(async () => {
+    // This requires the user to confirm on the ledger device
+    let rawSignature;
 
     onConfirmationComplete(true, rawSignature);
```
stanleyyconsensys, line 70: shall we update onConfirmationComplete, as rawSignature will be undefined anyway

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -151,6 +160,33 @@ export const checkActiveAccountAndChainId = async ({
...
+) => {
+  try {
+    store.dispatch(setSignMessageStage(SignMessageStageTypes.REQUEST_SEND));
+    const rawSig = await signMsgCallback;
+
+    store.dispatch(setSignMessageStage(SignMessageStageTypes.COMPLETE));
+
+    return rawSig;
+  } catch (e) {
+    console.error(e);
```
stanleyyconsensys, line 184: console.error may needed to remove?

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -526,17 +562,14 @@ export const getRpcMethodMiddleware = ({
...
+          const signMessageFunc = SignatureController.newUnsignedMessage({
             data: req.params[1],
             from: req.params[0],
             ...pageMeta,
             origin: hostname,
           });
 
-          res.result = rawSig;
-        } else {
-          res.result = AppConstants.ETH_SIGN_ERROR;
```
stanleyyconsensys, line 538: it look like ur generic method doesnt handle this

why not just have a try catch on the req method, hence to leave everything not change , doesnt it more easy ? as the logic will update in future too

app/components/UI/SignatureRequest/types.ts
```diff
@@ -16,7 +16,12 @@ export interface PageMeta {
 export interface MessageParams {
   data: string;
   from: string;
-  metamaskId: string;
+  metamaskId?: string;
```
stanleyyconsensys, line 19: what is the reason to change to optional?

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -151,6 +160,33 @@ export const checkActiveAccountAndChainId = async ({
...
 };
 
+/**
+ * abstract method to handle common sign message flow, flow will record when the request is sent and when it is completed,
+ * if error occur, error will be recorded in redux store with error stage.
+ * @param {Function} signMsgCallback - async callback function will trigger the actual sign message in SignatureController.
+ */
+const handleCommonSignFlow = async (
+  signMsgCallback: (
+    messageParams: MessageParams,
```
stanleyyconsensys, line 170: for the type , you can refer to sign controller

stanleyyconsensys (changes requested):

app/core/RPCMethods/RPCMethodMiddleware.ts
dawnseeker8: will remove later.

app/components/UI/LedgerModals/LedgerMessageSignModal.tsx
```diff
@@ -25,45 +29,46 @@ export const createLedgerMessageSignModalNavDetails =
...
+  }, [dispatch]);
 
-    if (type === 'signPersonalMessage') {
-      rawSignature = await KeyringController.signPersonalMessage(messageParams);
+  useEffect(() => {
+    //Close the modal when the signMessageStage is complete or error, error will return the error message to the user
+    if (
+      signMessageStage === SignMessageStageTypes.COMPLETE ||
+      signMessageStage === SignMessageStageTypes.ERROR
+    ) {
```
stanleyyconsensys, line 61: when we on unmount, shall we dismiss modal too?

app/components/UI/SignatureRequest/types.ts
dawnseeker8: The reason is metamaskId is mandatory defined in params in `SignatureController.newUnsign<>Message(messageParam)`, if you look at `RPCMethodMiddleware.ts` file for difference sign method, the params inside signMsgCallback may not include metamaskId, that is why i define it optional, however, the best way is getting the param types directly from @Metamask-SignatureController, but our library version didnt export the type, and i saw you already define a duplicate one in this file, i just directly use it to solve the tsc error.

app/core/RPCMethods/RPCMethodMiddleware.ts
dawnseeker8: I check the @Metamask-SignatureController library,  the current version didnt export the types including `MessageParams` and `OriginalRequest` , the only way is upgrade or downgrade library, i saw you already use the `MessageParams` in `SignatureRequest/types.ts`, that is why i directly modify there, and add a `OriginalRequest` type there.

app/core/RPCMethods/RPCMethodMiddleware.ts
dawnseeker8: Sorry, my mistake to remove the `} else {` probably from copy and parse the code.  anyway, my generic method will  only process the signing bit, so after i add the `} else {` the `eth_sign` will still work, the code will change as following
```Typescript
     if (req.params[1].length === 66 || req.params[1].length === 67) {
          await checkActiveAccountAndChainId({
            hostname,
            address: req.params[0].from,
            checkSelectedAddress: isMMSDK || isWalletConnect,
          });
          const signMessageFunc = SignatureController.newUnsignedMessage({
            data: req.params[1],
            from: req.params[0],
            ...pageMeta,
            origin: hostname,
          });

          res.result = await handleCommonSignFlow(signMessageFunc);
        } else {
          res.result = AppConstants.ETH_SIGN_ERROR;
          throw ethErrors.rpc.invalidParams(AppConstants.ETH_SIGN_ERROR);
        }
      ```

File: app/components/UI/SignatureRequest/types.ts (hunk 4 above)
Comment (stanleyyconsensys) [reply app/components/UI/SignatureRequest/types.ts:19]: i see, the type for MessageParams original was not used for RPCMethodMiddleware, but for approval request
until we move the sign function as a callback in RPCMethodMiddleware and then you have to define the type to bypass the type error

got it, but if we dont use that way (move the sign function as a callback) on RPCMethodMiddleware , just rmb to roll it back, thx

File: app/core/RPCMethods/RPCMethodMiddleware.ts (hunk 5 above)
Comment (stanleyyconsensys) [reply app/core/RPCMethods/RPCMethodMiddleware.ts:170]: they did export , as they import from metamask/message-manager

import type {
  MessageParams,
  MessageParamsMetamask,
  PersonalMessageParams,
  PersonalMessageParamsMetamask,
  TypedMessageParams,
  TypedMessageParamsMetamask,
  AbstractMessageManager,
  AbstractMessage,
  MessageManagerState,
  AbstractMessageParams,
  AbstractMessageParamsMetamask,
  OriginalRequest,
  TypedMessage,
  PersonalMessage,
  Message,
} from '@metamask/message-manager';

File: app/components/UI/LedgerModals/LedgerMessageSignModal.tsx (hunk 1 above)
Comment (dawnseeker8) [reply app/components/UI/LedgerModals/LedgerMessageSignModal.tsx:70]: i see the original code implemented by Zoltan, the rawSignature is optioinal in onConfirmationComplete(), let me test deviices when removing `rawSignature` in that method, see any unexpected behaviour or not.

File: app/components/UI/LedgerModals/LedgerMessageSignModal.tsx (hunk 6 above)
Comment (dawnseeker8) [reply app/components/UI/LedgerModals/LedgerMessageSignModal.tsx:61]: dismissModal wiill run first before unmounted,  so we dont need to handle dismissModal again here in unmounted callback

File: app/components/UI/LedgerModals/LedgerMessageSignModal.tsx (hunk 1 above)
Comment (stanleyyconsensys) [reply app/components/UI/LedgerModals/LedgerMessageSignModal.tsx:70]: onConfirmationComplete is passed by app/util/hardwareWallet/signatureUtils.ts, and since we no longer has rawSignature until keyring signed, hence, i think we dont need it anymore, you can double check

Comment (dawnseeker8) [reply app/components/UI/LedgerModals/LedgerMessageSignModal.tsx]: i have checked remove rawSignature and change to use `onComfirmationComplete(true)` and it still work as expected, so i have removed the rawSignature in above code.

File: app/core/RPCMethods/RPCMethodMiddleware.ts [hunk 7]
```diff
@@ -885,7 +899,20 @@ export const getRpcMethodMiddleware = ({
     if (!rpcMethods[req.method]) {
       return next();
     }
-    await rpcMethods[req.method]();
+
+    const isWhiteListedMethod = whiteListRPCs.includes(req.method);
+
+    try {
+      isWhiteListedMethod &&
+        store.dispatch(setSignMessageStage(SignMessageStageTypes.REQUEST_SEND));
```
stanleyyconsensys, line 907: do you wanna change the reducer name and action name to more general purpose as we discussed? or you prefer stay with explicitly name for sign message

dawnseeker8: last commit is to make a backup of workable version for further change, as i mention in the commit comment, further reducer change will be carried out later.

stanleyyconsensys (approved):

PR #7187, opened 2023-09-08
fix: sonar coverage path

sethkfman:
**Description**

This PR fixes the path for sonar coverage. 

**PR example of working coverage**

Test PR showing coverage being reported: https://github.com/MetaMask/metamask-mobile/pull/7214

**Checklist**


Gudahtt (dismissed):
Oops, good catch!

Gudahtt:
Actually... not sure if this will work. That entry is still for lcov, which is a different report format

sethkfman:
@Gudahtt I am updating to generate the lcov file

Gudahtt:
Found the docs: https://docs.sonarcloud.io/enriching/test-coverage/javascript-typescript-test-coverage/
Wasn't aware that SonarCloud only supported lcov, sorry about that! Good to know.

package.json
```diff
@@ -69,6 +69,7 @@
     "test:attribution-check": "./scripts/attributions-check.sh",
     "test:merge-coverage": "nyc merge ./tests/coverage ./tests/merged-coverage/coverage.json",
     "test:validate-coverage": "nyc check-coverage --nycrc-path ./coverage-thresholds.json -t ./tests/merged-coverage/",
+    "test:convert-coverage-json-lcov": "mkdir .nyc_output && cd ./tests/coverage && nyc report --reporter=text --reporter=lcov",
```
Gudahtt, line 72: Looks like this produces an empty report, which makes sense because it's looking in `.nyc_output` for the coverage data (which is empty). I'm also not sure why we'd want to move into the coverage directory for this step.

Rather than creating that directory and moving the report to it, we can use `-i` to tell nyc where the coverage data is (e.g. `-t ./tests/coverage/`)

Gudahtt: Also, rather than adding a second command to generate an lcov report from the JSON report, it looks like we could generate multiple reports and merge in the same step. The `nyc merge` command isn't needed to merge coverage data, `nyc report` does that too _and_ can generate multiple report types.

We can update `test:merge-coverage` to something like `nyc check-coverage -t ./coverage-thresholds.json --report-dir ./tests/merged-coverage/` --reporter json --reporter text --reporter lcov`

That gives us a JSON and lcov report in the same directory the JSON report is in now, plus provides a text report as CLI output

Gudahtt: Sorry I made a mistake in that comment, `check-coverage` should have been `report`. This was a suggested replacement for the merge command

package.json
```diff
@@ -68,7 +68,7 @@
     "test:tgz-check": "./scripts/tgz-check.sh",
     "test:attribution-check": "./scripts/attributions-check.sh",
     "test:merge-coverage": "nyc merge ./tests/coverage ./tests/merged-coverage/coverage.json",
```
Gudahtt, line 70: This is what I meant to suggest:

```suggestion
    "test:merge-coverage": "nyc report -t ./tests/coverage --report-dir ./tests/merged-coverage/ --reporter json --reporter text --reporter lcov",
```

This would generate the coverage report in three different formats using the intermediary coverage data from the shard runs. The check-coverage command does not require any changes

.github/workflows/ci.yml
```diff
@@ -116,6 +116,11 @@ jobs:
           name: coverage
           path: ./tests/merged-coverage/coverage.json
           if-no-files-found: error
+      - uses: actions/upload-artifact@v3
+        with:
+          name: coverage-lcov
+          path: ./tests/merged-coverage/lcov.info
+          if-no-files-found: error
```
Gudahtt, line 123: Nit: Perhaps it would be better to upload all reports at once?

```suggestion
          path: ./tests/merged-coverage
          if-no-files-found: error
```

sonar-project.properties
```diff
@@ -14,8 +14,8 @@ sonar.sources=app/
 # Inclusions for test files.
 sonar.test.inclusions=**.test.**
 
-# Test coverage path
-sonar.javascript.lcov.reportPaths=/tests/coverage/lcov.info
+# Test coverage path in GitHub action
+sonar.javascript.lcov.reportPaths=**/tests/coverage/lcov.info
```
Gudahtt, line 18: The docs use a relative path: https://docs.sonarcloud.io/enriching/test-coverage/javascript-typescript-test-coverage/

```suggestion
sonar.javascript.lcov.reportPaths=./tests/coverage/lcov.info
```

package.json
```diff
@@ -67,8 +67,8 @@
     "test:depcheck": "yarn depcheck",
     "test:tgz-check": "./scripts/tgz-check.sh",
     "test:attribution-check": "./scripts/attributions-check.sh",
-    "test:merge-coverage": "nyc merge ./tests/coverage ./tests/merged-coverage/coverage.json",
-    "test:validate-coverage": "nyc check-coverage --nycrc-path ./coverage-thresholds.json -t ./tests/merged-coverage/",
+    "test:merge-coverage": "nyc report --temp-dir ./tests/coverage --report-dir ./tests/merged-coverage/ --reporter json --reporter text --reporter lcovonly",
+    "test:validate-coverage": "nyc check-coverage --nycrc-path ./coverage-thresholds.json -t ./tests/merged-coverage/ --reporter json --reporter text --reporter lcov",
```
Gudahtt, line 71: Looks like the flags were accidentally left on this command as well:
```suggestion
    "test:validate-coverage": "nyc check-coverage --nycrc-path ./coverage-thresholds.json -t ./tests/merged-coverage/",
```

It doesn't seem to affect the command, so I guess they are ignored

sethkfman: CodeCov pulls them in but it isn't impacting the GH Action. The path issue should be sorted out and I will update the yarn command after that.

sethkfman:
@Gudahtt Okay, I have a linked test PR in the description showing a 0% coverage reporting on new code. That should be the evidence we need to move forward with this PR.

Gudahtt (approved):
LGTM!

PR #447, opened 2019-02-28
TX Notifications

brunobar79:
**Description**

- Adds transaction notifications for  pending, confirmed, rejected and also incoming TX
- Uses in-app notification if the app is in foreground otherwise it will use a local push notification

BONUS:
- Fixed all modals to be draggable #442 
- Did some initial refactor of the navigation structure #420  (required for this PR)
- Fix all vulns by bumping `metro-react-native-babel-preset` to 0.52.0

**Checklist**

* [x] There is a related GitHub issue
* [x] Tests are included if applicable
* [x] Any added code is fully documented

**Issue**

Resolves #176 
Resolves #442 

For incoming transactions to work properly on testnets we need to merge:  https://github.com/MetaMask/gaba/pull/62

DEMO:   http://recordit.co/8dys9QTkca

PR #7195, opened 2023-09-11
fix: Fix inconsistency in incoming transaction notifications

Gudahtt:
**Development & PR Process**

**Description**

Incoming transaction notifications were incorrectly being triggered for transactions on networks different than the current selected network. This situation only occurs when a custom network is selected, and when an incoming transaction is detected on a _different_ custom network. In all other circumstances, we only show incoming transaction notifications for the current selected network.
    
This problem date back to the original implementation of incoming transaction notifications (see https://github.com/MetaMask/metamask-mobile/pull/447/files#r261268602)

The filter has been updated to use chain ID rather than network ID, and to filter in the same way for built-in and custom networks.

**Screenshots/Recordings**

Testing this is not really feasible because the use-case isn't normally possible except in cases of a race condition. The buggy situation is where you get an incoming transaction on a custom network that is different from the current selected custom network, but in general incoming transaction polling is only active for the current network. So you'd need to switch at the _exact_ right time to demonstrate the problem, and even then it wouldn't be evident from the recording exactly what had happened.

I have prepared a recording that shows that incoming transaction notifications still work correctly on this branch though: https://recordit.co/ee4wSwZkm7

**Issue**

This relates to https://github.com/MetaMask/mobile-planning/issues/1226

**Checklist**

### DIFF
--- a/app/core/NotificationManager.js
+++ b/app/core/NotificationManager.js
@@ -2,7 +2,6 @@
 
 import PushNotification from 'react-native-push-notification';
 import Engine from './Engine';
-import Networks, { isKnownNetwork } from '../util/networks';
 import { hexToBN, renderFromWei } from '../util/number';
 import Device from '../util/device';
 import { strings } from '../../locales/i18n';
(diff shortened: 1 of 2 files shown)


Cal-L (approved):
LGTM

cortisiko (approved):
Transactions are working as expected ✅

cortisiko (approved):

PR #7202, opened 2023-09-12
test: build test dapp and run tests against localhost

seaona:
## Description
This PR adds logic for running (one/many) dapp server(s) locally whenever we run the e2e tests. This way we can run the tests in a controlled environment (against localhost) instead of on a live site.

Here some highlights on how it is done (based on existing Extension implementation):

- we leverage `test-dapp` node module and `serve-handler` for building a static server
- we add the corresponding options on the `withFixtures` method, in order to indicate if we want a dapp server and how many(if more than one)
- on the test, we just need to pass `dapp: true` to start the test dapp server locally

Test on ci continue to work:
- Bitrise run : https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/296f9b58-fb20-4e89-94da-0a475ec04caa

Screenshot from 2023-09-12 18-59-46

## Screenshots/Recordings
Notice how we are now going to `localhost` in the browser for accessing the test dapp

https://github.com/MetaMask/metamask-mobile/assets/54408225/ffbaa904-ae83-47d6-a423-e2210dfe26e5

## Issue

- fixes https://github.com/MetaMask/mobile-planning/issues/1253

## Checklist

### DIFF
--- a/e2e/create-static-server.js
+++ b/e2e/create-static-server.js
@@ -1,0 +1,22 @@
+/* eslint-disable import/no-nodejs-modules */
+import http from 'http';
+import path from 'path';
+import serveHandler from 'serve-handler';
+
+const createStaticServer = function (rootDirectory) {
+  return http.createServer((request, response) => {
+    if (request.url.startsWith('/node_modules/')) {
+      request.url = request.url.substr(14);
+      return serveHandler(request, response, {
+        directoryListing: false,
+        public: path.resolve('./node_modules'),
+      });
+    }
+    return serveHandler(request, response, {
+      directoryListing: false,
+      public: rootDirectory,
+    });
+  });
+};
+
+export default createStaticServer;
--- a/e2e/fixtures/fixture-builder.js
+++ b/e2e/fixtures/fixture-builder.js
@@ -1,6 +1,6 @@
 import { merge } from 'lodash';
 
-const DAPP_URL = 'metamask.github.io';
+const DAPP_URL = 'localhost';
 
 /**
  * FixtureBuilder class provides a fluent interface for building fixture data.
--- a/e2e/fixtures/fixture-helper.js
+++ b/e2e/fixtures/fixture-helper.js
@@ -1,9 +1,11 @@
-/* eslint-disable no-console */
+/* eslint-disable no-console, import/no-nodejs-modules */
 import FixtureServer from './fixture-server';
 import FixtureBuilder from './fixture-builder';
 import Ganache from '../../app/util/test/ganache';
 import GanacheSeeder from '../../app/util/test/ganache-seeder';
 import axios from 'axios';
+import path from 'path';
+import createStaticServer from '../create-static-server';
 
 const fixtureServer = new FixtureServer();
 
@@ -83,9 +85,16 @@ export async function withFixtures(options, testSuite) {
     restartDevice = false,
     ganacheOptions,
     smartContract,
+    dapp,
+    dappOptions,
+    dappPath = undefined,
+    dappPaths,
   } = options;
 
   const ganacheServer = new Ganache();
+  const dappBasePort = 8080;
+  let numberOfDapps = dapp ? 1 : 0;
+  const dappServer = [];
 
   try {
     let contractRegistry;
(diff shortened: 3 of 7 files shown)


seaona:
@SocketSecurity ignore serve-handler@6.1.5 reasons:
- it is a dev dependency for e2e
- it is also used in Extension e2e

e2e/fixtures/fixture-builder.js
```diff
@@ -1,6 +1,6 @@
 import { merge } from 'lodash';
 
-const DAPP_URL = 'metamask.github.io';
+const DAPP_URL = 'localhost';
```
seaona, line 3: this is used on `withPermissionsControllerConnectedToTestDapp` . Now we use localhost instead of the test-dapp link

e2e/fixtures/fixture-helper.js
```diff
@@ -83,9 +85,16 @@ export async function withFixtures(options, testSuite) {
     restartDevice = false,
     ganacheOptions,
     smartContract,
+    dapp,
+    dappOptions,
+    dappPath = undefined,
+    dappPaths,
```
seaona, line 91: we make available these new options for each test case using fixtures

e2e/fixtures/fixture-helper.js
```diff
@@ -98,6 +107,34 @@ export async function withFixtures(options, testSuite) {
...
+          );
+        }
+        dappServer.push(createStaticServer(dappDirectory));
+        dappServer[i].listen(`${dappBasePort + i}`);
+        await new Promise((resolve, reject) => {
+          dappServer[i].on('listening', resolve);
+          dappServer[i].on('error', reject);
+        });
+      }
+    }
```
seaona, line 137: here is the logic for creating the dapp server, same implementation as in Extension. See [here](https://github.com/MetaMask/metamask-extension/blob/develop/test/e2e/helpers.js#L82)

e2e/create-static-server.js
```diff
@@ -0,0 +1,22 @@
...
+      });
+    }
+    return serveHandler(request, response, {
+      directoryListing: false,
+      public: rootDirectory,
+    });
+  });
+};
+
+export default createStaticServer;
```
seaona, line 22: create server for running the dapp we want, in this case the test-dapp

e2e/pages/TestDApp.js
```diff
@@ -6,6 +6,7 @@ import Browser from './Drawer/Browser';
 import root from '../../locales/languages/en.json';
 
 export const TEST_DAPP_URL = 'https://metamask.github.io/test-dapp/';
+export const TEST_DAPP_LOCAL_URL = 'http://localhost:8080';
```
seaona, line 9: we leave the `TEST_DAPP_URL` var too because there is 1 remaining test in confirmations using it. This will be removed in the subsequent PR, where we can migrate the `signatures` test. (Not in this PR in order to not add too many changes at once)

e2e/specs/confirmations/send-erc721.spec.js
```diff
@@ -13,20 +13,23 @@ import {
...
-      await device.reverseTcpPort('8545');
+      await device.reverseTcpPort('8545'); // ganache
+      await device.reverseTcpPort('8080'); // test-dapp
     }
   });
 
   it('send an ERC721 token from a dapp', async () => {
     await withFixtures(
       {
+        dapp: true,
```
seaona, line 32: just by passing this option in fixtures, a test dapp server is started and we can now use it going to localhost:8080 (see above)

e2e/fixtures/fixture-helper.js
```diff
@@ -118,6 +155,20 @@ export async function withFixtures(options, testSuite) {
...
+            dappServer[i].close((error) => {
+              if (error) {
+                return reject(error);
+              }
+              return resolve();
+            });
+          });
+        }
+      }
+    }
```
seaona, line 171: teardown the server after finished

package.json
```diff
@@ -452,6 +452,7 @@
     "react-test-renderer": "18.2.0",
     "regenerator-runtime": "0.13.9",
     "rn-nodeify": "10.3.0",
+    "serve-handler": "^6.1.5",
```
seaona, line 455: same dev dependency used in Extension

e2e/fixtures/fixture-helper.js
```diff
@@ -83,9 +85,16 @@ export async function withFixtures(options, testSuite) {
...
     smartContract,
+    dapp,
+    dappOptions,
+    dappPath = undefined,
+    dappPaths,
   } = options;
 
   const ganacheServer = new Ganache();
+  const dappBasePort = 8080;
+  let numberOfDapps = dapp ? 1 : 0;
```
NicolasMassart, line 96: C: maybe you could directly use `dappOptions.numberOfDapps` here to initialise the `numberOfDapps` as a const?

NicolasMassart (approved):
Looks good to me, this is a very nice step in the direction where we have less constraints and dependencies with the "outside" when testing. Well done!

NOTE: Added a comment but it's not a blocker for merge. the code works and is consistent with extension, so maybe it's better to keep it like this. Commenting it perhaps a bit more could help though (the comments you added to this PR could be in the code).

jpuri (approved):
👍 great work

seaona:
thank you v much for your reviews @jpuri @NicolasMassart 
About this comment:

>  the code works and is consistent with extension, so maybe it's better to keep it like this. Commenting it perhaps a bit more could help though (the comments you added to this PR could be in the code).

This makes a lot of sense and I also see the need to keep things documented. I was thinking maybe to write some documentation around e2e in Mobile, on how to use of fixtures, ganache, contract deployments, and mocking (this is coming [very soon](https://github.com/MetaMask/mobile-planning/issues/1235)) and best practices and we could also have a meeting around that, once the mocking is done, so the infra will be ready. Let me know if that sounds good :pray:

cortisiko (approved):
Looks good!

NicolasMassart (approved):
Looks good to me.

PR #1251, opened 2019-12-17
Feature: approve screen

estebanmino:
**Description**

Updated ERC20 approve screen per https://www.figma.com/file/Emfl6CRPRtuQtzJEOJJ3hy/Approve-Screen?node-id=721%3A1

I reused some components from https://github.com/MetaMask/metamask-mobile/pull/1239 so it would be a good idea to merge that one first.

**Checklist**

* [ ] There is a related GitHub issue
* [ ] Tests are included if applicable
* [ ] Any added code is fully documented

**Issue**

Resolves #1218
Resolves #1243

Issue #1249, opened 2019-12-16
Second transaction is not shown to the user


PaulRBerg:
## Description
Here's my transaction flow:

1. Approve token
2. Submit dapp tx

And an approximation for my JavaScript code that creates these two transactions (with ethers.js).

```javascript
const txs = [];
const tokenApprovalTx = await createTokenApprovalTx();
txs.push(tokenApprovalTx);
const sablierTx = await createSablierTx();
txs.push(sablierTx);
```

The 2n transaction is not picked up by the mobile browser and the user doesn't see it at all.

## Expected Behaviour

The user should be prompted to sign the 2nd transaction, even if they just previously submitted another one.

## Steps

1. Go to [pay.sablier.finance](https://pay.sablier.finance)
2. Log in
3. Have some Eth and Dai
4. Tap "Stream Money"
5. Fill the form with "0.01", any recipient and any duration (caveat: if you don't pay yourself, the money is lost)
6. Submit form

## Environment

- iPhone XR and iOS 12.3.1
- MetaMask Mobile v0.2.8

PaulRBerg:
Any updates on this? this issue has caused trouble for a lot of our users. The dapp cannot be used with Metamask Mobile at all.

brunobar79:
@PaulRBerg  Can you add to the code snippet of what do you do with the tx array? That will help a lot.

PaulRBerg:
Hey @brunobar79 I created a gist for you:

https://gist.github.com/PaulRBerg/eadb8eeed13f0cb7d8576723948d532c

See the `submit` function.

dpazdan:
related support case 29299

dpazdan:
I was able to replicate this bug when trying to add OXT tokens to orchid account setup: http://account.orchid.technology/

PaulRBerg:
I haven't tested it, but the same bug might occur on [gitcoin.co/grants](https://gitcoin.co/grants/) too. They have a similar transaction flow.

brunobar79:
Thanks for the info. I'm gonna create a very simple repro app in ropsten to be able to debug better.

danjm:
@PaulRBerg Can you confirm that this is still a bug?

PaulRBerg:
@danjm it works well on version 0.2.14. I'll close this issue for now.

Thanks for the fix!

PR #7209, opened 2023-09-13
fix: handle etherscan rate limit errors

matthewwalsh0:
**Description**

The mobile client uses the core `TransactionController` which supports the use of an API key when querying Etherscan for incoming or updated transactions.

The extension, using its `TransactionController`, only queries Etherscan for normal transactions using a single callout.

The mobile client, using the core `TransactionController`, queries Etherscan for both normal transactions and token transfer transactions related to an account. This requires 2 callouts and therefore consistently results in a rate limit error on one of the callouts if no API key is provided.

Previously, if either request failed due to a rate limit error, the update would be skipped, and no transactions would be added. This PR instead processes whatever transactions are found, regardless of which specific callout returned a rate limit error.

Since both callouts are performed simultaneously, it is currently random chance which callout results in the rate limit error, meaning both normal and token transfer transactions are still supported, but either could experience sporadic delays.

Additional changes include:

- Add back the background timer to ensure incoming transactions continue to be polled, even when the app is in background mode.
- Disable the `startBlock` argument in Etherscan requests to match the previous behaviour where a bug meant this was never being utilised.

**Recordings**

#### Transaction History After Import

https://github.com/MetaMask/metamask-mobile/assets/19745142/9867b984-cfc7-4225-8bf3-dd9cf640e27d

#### New Incoming Transaction & Notification

https://github.com/MetaMask/metamask-mobile/assets/19745142/b9a97442-b39a-4618-98b5-dd84a85698b2

**Issue**

Fixes [#1249](https://github.com/MetaMask/mobile-planning/issues/1249), [#1250](https://github.com/MetaMask/mobile-planning/issues/1250), [#1251](https://github.com/MetaMask/mobile-planning/issues/1251)

**Checklist**


vinistevam (dismissed):

OGPoyraz (dismissed):

app/components/Nav/Main/index.js
```diff
@@ -167,6 +169,10 @@ const Main = (props) => {
       // the background timer, which is less intense
       if (backgroundMode.current) {
         removeNotVisibleNotifications();
+
+        BackgroundTimer.runBackgroundTimer(async () => {
+          await TransactionController.updateIncomingTransactions();
+        }, AppConstants.TX_CHECK_BACKGROUND_FREQUENCY);
```
sethkfman, line 175: This background activity may be flagged by the App Store reviewers and we may want to increase the timeout. Please annotate the background frequency on the description. Thx

matthewwalsh0: Apologies for the confusion, this was mistakenly removed previously but matches the same timer and interval used in 7.6.

https://github.com/MetaMask/metamask-mobile/blob/v7.6.0/app/components/Nav/Main/index.js#L171

sleepytanya:
All incoming transaction notifications work as expected, push notification are also seen when the app is on the background.
Will upload videos later.

sleepytanya (approved):
LGTM

sethkfman (approved):
LGTM

PR #7226, opened 2023-09-14
chore: add script to patch transaction controller

matthewwalsh0:
**Description**

Add a `patch:tx` package script to simplify and standardise the generation of the transaction controller patch using a core branch.

Note that the patch will only be updated going forward to resolve bugs or to facilitate controller version updates.

**Checklist**


scripts/patch-transaction-controller.sh
```diff
@@ -0,0 +1,48 @@
...
+CORE_DIR="$1"
+
+if [ -z "$CORE_DIR" ] ; then
+  echo "Usage: yarn patch:tx <core-directory>"
+  echo ""
+  echo "Example: yarn patch:tx ../core"
+  exit 1
+fi
+
+DIST_DIR_NEW="$1/packages/transaction-controller/dist"
```
Gudahtt, line 39: Nit: may as well use `CORE_DIR` here instead of `$1` for improved readability

```suggestion
DIST_DIR_NEW="$CORE_DIR/packages/transaction-controller/dist"
```

scripts/patch-transaction-controller.sh
```diff
@@ -0,0 +1,48 @@
...
+
+rm -rf $DIST_DIR_OLD
```
Gudahtt, line 41: These should all be quoted to prevent word splitting (see https://www.shellcheck.net/wiki/SC2086 )

scripts/patch-transaction-controller.sh
```diff
@@ -0,0 +1,48 @@
...
+
+set -e
+set -o pipefail
+
+PACKAGE_DIR_OLD="node_modules/@metamask/transaction-controller"
+DIST_DIR_OLD="$PACKAGE_DIR_OLD/dist"
+PATCH_FILE="patches/@metamask+transaction-controller+*.patch"
+COMMENT_LINE_COUNT=10
+
+COMMENT='+PATCH GENERATED FROM MetaMask/core branch refactor/transaction-controller-patch-mobile\
```
Gudahtt, line 11: Nit: ShellCheck gives a warning here, because it thinks this string is supposed to expand into something

It can be suppressed with a comment:
```suggestion
# ShellCheck is confused by the patch format; this is intentionally single-quoted and is not meant to expand
# shellcheck disable=SC2016
COMMENT='+PATCH GENERATED FROM MetaMask/core branch refactor/transaction-controller-patch-mobile\
```

Gudahtt:
Great idea! Thanks for putting this together

scripts/patch-transaction-controller.sh
```diff
@@ -0,0 +1,48 @@
...
+  echo ""
+  echo "Example: yarn patch:tx ../core"
+  exit 1
+fi
+
+DIST_DIR_NEW="$1/packages/transaction-controller/dist"
+
+rm -rf $DIST_DIR_OLD
+cp -r  $DIST_DIR_NEW $PACKAGE_DIR_OLD
+rm -f $DIST_DIR_NEW/*.map
```
Gudahtt, line 43: Should we delete the mocks too?

scripts/patch-transaction-controller.sh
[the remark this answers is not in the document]
matthewwalsh0: Done.

scripts/patch-transaction-controller.sh
```diff
@@ -0,0 +1,62 @@
...
+PACKAGE="@metamask/transaction-controller"
+PACKAGE_DIR_MOBILE="node_modules/$PACKAGE"
+DIST_DIR_MOBILE="$PACKAGE_DIR_MOBILE/dist"
+PATCH_FILE="patches/@metamask+transaction-controller+*.patch"
+COMMENT_LINE_COUNT=7
+
+COMMENT='+PATCH GENERATED FROM MetaMask/core branch: refactor/transaction-controller-patch-mobile\
++This patch backports various transaction controller features from the main branch of MetaMask/core\
++Steps to update patch:\
++* Create a new core branch from: refactor/transaction-controller-patch-mobile\
```
matthewwalsh0, line 26: I've simplified these instructions and tried to make it clear that we don't want to update the primary core branch until the new patch has been merged, so it always matches what is merged in mobile `main`.

scripts/patch-transaction-controller.sh
matthewwalsh0: Done.

scripts/patch-transaction-controller.sh
matthewwalsh0: Done.

scripts/patch-transaction-controller.sh
matthewwalsh0: It turned out it just didn't like the backticks for the yarn scripts so I've replaced with quotes.

scripts/patch-transaction-controller.sh
matthewwalsh0: In the current core branch its already been deleted, but I've added it in anyway just in case there is ever a need to run the script on an older core commit.

OGPoyraz (dismissed):

scripts/patch-transaction-controller.sh
Gudahtt: Nice, much better

Gudahtt (dismissed):
LGTM!

OGPoyraz (approved):

PR #7280, opened 2023-09-20
test: Enable fixtures smoke tests

SamuelSalas:
**Development & PR Process**

**Description**
The goal of this pr is to enable fixtures in the following smoke tests:

- request-token-flow.spec.js
- permission-system-revoking-multiple-accounts.spec.js
- browser-tests.spec.js
- addressbook-tests.spec.js

**Screenshots/Recordings**

**Issue**

fixes #???

**Checklist**


e2e/specs/permission-system-revoking-multiple-accounts.spec.js
```diff
@@ -20,12 +26,20 @@ describe(
     'Connecting to multiple dapps and revoking permission on one but staying connected to the other',
   ),
   () => {
+    beforeAll(async () => {
+      const fixture = new FixtureBuilder().build();
+      await startFixtureServer();
```
cortisiko, line 31: since the tests you are updating are technically new user tests. Shouldn't we have https://github.com/MetaMask/metamask-mobile/pull/7280/files#diff-ef1335a2f3476ec310770a42293fae99862b8d6a010e9e5bb91d7fee3799cb50R26 in the before hook as well?

SamuelSalas: We decided to add the `fixture.state.user.seedphraseBackedUp = false;` because the e2e/specs/request-token-flow.spec.js test was failing. We needed the wallet to be set as unprotected. Instead of having that step on every beforeAll hook, it would be better to update the change it false on the fixture-builder.js.

e2e/specs/addressbook-tests.spec.js
```diff
@@ -10,20 +10,34 @@ import TabBarComponent from '../pages/TabBarComponent';
...
+  });
+
   beforeEach(() => {
     jest.setTimeout(150000);
   });
 
-  it('should create new wallet', async () => {
-    await CreateNewWallet();
+  afterAll(async () => {
+    await stopFixtureServer();
```
seaona, line 40: As a suggestion, we could completely remove the `startFixtureServer` and `stopFixtureServer` if we use the `withFixtures` method. Example here:
https://github.com/MetaMask/metamask-mobile/blob/main/e2e/specs/confirmations/send-erc721.spec.js#L29

SamuelSalas: I decided using the start and stop fixtureServer because a couple of scripts test multiple behaviors from the feature. Merging all the it blocks into one block would make the script less readable.

e2e/specs/permission-system-revoking-multiple-accounts.spec.js
cortisiko: We should revisit this in a follow up PR. Thanks for explaining!

cortisiko (approved):
🌮 🌮 🌮 🌮

chrisleewilcox (approved):

PR #7517, opened 2023-10-17
feat: fix the looping problem of permission

dawnseeker8:
## **Description**
Fix the looping error in [`LedgerConfirmationModal.ts`](https://app.zenhub.com/workspaces/cet-metamask-mobile-ledger-integration-649aeabc8cdfe3878d13629f/issues/zh/87)

Following has been implemented:
* add a flag in `LedgerConfirmationModal.ts` to control whether to rerender the screen.
* Modify the logic inside `onRetry()` to set the errorDetail to undefined so that ledger screen can show up.
* Fix a edge case in `useBluetoothPermissions.ts` file when screen move away from metamask to android setting, `bluetoothPermissionError` haven't been set to undefined.


android/.project
```diff
@@ -1,17 +1,17 @@
...
-		<buildCommand>
-			<name>org.eclipse.buildship.core.gradleprojectbuilder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
-	</buildSpec>
-	<natures>
-		<nature>org.eclipse.buildship.core.gradleprojectnature</nature>
-	</natures>
+        <name>MetaMask</name>
```
dawnseeker8, line 3: Sorry, this is mistake for format on save. but content didnt change. so should be ok.

app/components/UI/LedgerModals/LedgerConfirmationModal.tsx
```diff
@@ -47,6 +47,7 @@ const LedgerConfirmationModal = ({
 }: LedgerConfirmationModalProps) => {
   const { colors } = useAppThemeFromContext() || mockTheme;
   const styles = useMemo(() => createStyles(colors), [colors]);
+  const [permissionErrorShown, setPermissionErrorShown] = useState(false);
```
dawnseeker8, line 50: flag to control whether rerender content of permission error below. default to false.

app/components/UI/LedgerModals/LedgerConfirmationModal.tsx
```diff
@@ -163,7 +166,7 @@ const LedgerConfirmationModal = ({
       }
     }
 
-    if (bluetoothPermissionError) {
+    if (bluetoothPermissionError && !permissionErrorShown) {
```
dawnseeker8, line 169: if permissionErrorShown is true, dont render the error screen again. 
this bluetoothPermissionError will be undefined and error repeatly set during `permissionCheck()` in `useBluetoothPermissions.ts` and cause the screen already rerenderiing.

app/components/UI/LedgerModals/LedgerConfirmationModal.tsx
```diff
@@ -95,7 +98,10 @@ const LedgerConfirmationModal = ({
   };
 
   useEffect(() => {
-    hasBluetoothPermissions && bluetoothOn && connectLedger();
+    hasBluetoothPermissions &&
+      bluetoothOn &&
+      !permissionErrorShown &&
```
dawnseeker8, line 103: this check here is to prevent, when user enable permission, screen will send connectLedger() immediately which cause pending transaction or ledgerDisconnect when user click `retry` which will resend `connectLedger()` again.

igorms-cons (approved):

PR #7537, opened 2023-10-18
feat: Fix the screen looping in scan ledger device

dawnseeker8:
## **Description**
Fix the similar loop issue for scan devices screen when user disable the `location` permisison in android.

### DIFF
--- a/app/components/Views/LedgerConnect/Scan.tsx
+++ b/app/components/Views/LedgerConnect/Scan.tsx
@@ -64,12 +64,14 @@ const Scan = ({
     hasBluetoothPermissions,
     bluetoothOn,
   );
+  const [permissionErrorShown, setPermissionErrorShown] = useState(false);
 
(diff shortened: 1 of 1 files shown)


igorms-cons (approved):

app/components/Views/LedgerConnect/Scan.tsx
```diff
@@ -148,7 +151,18 @@ const Scan = ({
...
-  }, [deviceScanError, bluetoothPermissionError, bluetoothConnectionError]);
+  }, [
+    deviceScanError,
+    bluetoothPermissionError,
+    bluetoothConnectionError,
+    permissionErrorShown,
+  ]);
+
+  useEffect(() => {
+    if (hasBluetoothPermissions) {
```
dawnseeker8, line 162: if permission has been set to on, set the flag off so that screen can move back automatically.

PR #7589, opened 2023-10-24
chore: adds the fixture env variable to debug android builds

christopherferreira9:
## **Description**

Adds `IS_TEST='true'` to android debug builds.


cortisiko (approved):
✅ ✅

PR #7760, opened 2023-11-10
chore(revert): revert "chore: Split the Swap test into two parts so that transaction activity is checked at the end"

wachunei:
Reverts MetaMask/metamask-mobile#7726

Consider reverting this PR because of the changes introduced to the `Base` folder components related to adding testID props. These testIDs are impacting `Ramp` snapshots instead of only `Swap` views.

I suggest these testIDs to be added to the Swaps views components instead of the Base ones. In case the Base components do not pass those props down, to just modify them by passing `testID` or all the `{...props}` down, so it does not impact the other Views that use them.


NicolasMassart (approved):
Thanks Pedro for raising this issue.
I approve the revert. Could you create a ticket to work again on this in a way that doesn't break things outside of the expected scope?

PR #7817, opened 2023-11-15
fix: blockaid banner appearance for approve transactions

jpuri:
## **Description**
Fix blockaid banner in approval pages. Currently the banner on approval pages was broken due to state issues.

## **Related issues**

Fixes: #7806, #7808

## **Manual testing steps**

1. Go to test DAPP and trigger malicious approval / approval for all transactions
2. Check that blockaid banner appears correctly

## **Screenshots/Recordings**

https://github.com/MetaMask/metamask-mobile/assets/2182307/c7b6b813-9602-4a6f-8d8f-4c79f6022585

## **Pre-merge author checklist**

- [X] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [X] I've clearly explained what problem this PR is solving and how it is solved.
- [X] I've linked related issues
- [X] I've included manual testing steps
- [X] I've included screenshots/recordings if applicable
- [X] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [X] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.
- [X] I’ve properly set the pull request status:
  - [X] In case it's not yet "ready for review", I've set it to "draft".
  - [ ] In case it's "ready for review", I've changed it from "draft" to "non-draft".

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #7822, opened 2023-11-15
fix: PPOM - Metrics information from ppom is not logged

### LINKED ISSUE #7795: [Bug]: PPOM - Metrics information from ppom is not logged
### Describe the bug

**Problem**: whenever we trigger a transaction from a dapp, we don't see any information about Blockaid validation

### Expected behavior

We should see ppom information in metric events.

From @segun : `so we need to wait until PPOM is done before sending metrics`

Question: if we follow that approach, then we need to think about this other edge case:

- if we wait for the ppom validation finishes, before sending metrics, in the case the validation is not finished but the user Cancels/Submits the tx  - should make sure that the metrics events are then sent (even if the validation hasn't finished)

to define what should be the expectation here cc @bschorchit should we have a new status for a tx saying `validation_in_process` or something similar? - so it's not benign, malicious, or invalid but just the validation didn't finish when tx was submitted? :thinking: 

From @bschorchit :

> We should add the value of security_alert_response to loading and ui_customizations to security_alert_loading for the scenario that Mariona described

### Screenshots/Recordings

https://github.com/MetaMask/metamask-mobile/assets/54408225/a918aa35-a881-46bb-a839-ac62e01e814d

### Steps to reproduce

1. Enable Blockaid
2. Trigger a tx
3. See metrics events - no blockaid info is logged

_No response_

### Version

develop (with blockaid feature flag on)

### Build type

None

### Device

Pixel 6

### Operating system

Android

### Additional context

_No response_

### Severity

_No response_

segun:
## **Description**

Whenever we trigger a transaction from a dapp, we don't see any information about Blockaid validation. This is because the blockaid banner now shows loading initially for some time before the actual alert is shown. We were sending metrics only on component load, but we should send metrics for blockaid alert

The solution used here is to delay metrics for blockaid until user clicks confirm or cancel. At that point we have a definitive blockaid status to send (if it's loading, we send loading, if it's loaded we send the normal metrics).

For the deleted test cases in MessageSign (because enzyme does not support useEffect). I had a call with @cryptotavares and we will convert the tests to RTL in another PR and add the required useEffect tests there. There's no need to delay this because of that.

## **Related issues**

Fixes: #7795 

## **Manual testing steps**

1. Checkout this branch and run MetaMask Mobile
2. Enable blockaid in settings
4. Open testdapp and perform a PPOM transaction
5. Check the console and notice how security alert metrics is not sent
6. Now click/confirm cancel while the banner is still showing loading and  check console to see how security alert metrics is sent with loading
7. Now perform another PPOM transaction, this time wait for the blockaid banner to fully load and then click confirm/cancel and see how the security alert response is sent in metrics.

### **Before**
[](https://github.com/MetaMask/metamask-mobile/assets/54408225/a918aa35-a881-46bb-a839-ac62e01e814d)

### **After**
https://github.com/MetaMask/metamask-mobile/assets/44811/3fb7f759-a1c7-4eaa-b42a-51243d6b340b


app/components/UI/SignatureRequest/index.js
```diff
@@ -22,8 +28,8 @@ import { selectProviderType } from '../../../selectors/networkController';
 import BlockaidBanner from '../BlockaidBanner/BlockaidBanner';
 import { getAnalyticsParams } from '../../../util/confirmation/signatureUtils';
 import { SigningModalSelectorsIDs } from '../../../../e2e/selectors/Modals/SigningModal.selectors';
-import setSignatureRequestSecurityAlertResponse from '../../../actions/signatureRequest';
-import { store } from '../../../store';
+// import setSignatureRequestSecurityAlertResponse from '../../../actions/signatureRequest';
+// import { store } from '../../../store';
```
cryptotavares, line 32: This needs to either be removed or uncommented

app/components/UI/TypedSign/index.js
```diff
@@ -85,19 +84,6 @@ class TypedSign extends PureComponent {
...
-    const {
-      messageParams: { metamaskId },
-      messageParams,
-    } = this.props;
-
-    AnalyticsV2.trackEvent(
-      MetaMetricsEvents.SIGNATURE_REQUESTED,
-      getAnalyticsParams(messageParams, 'typed_sign'),
-    );
-    addSignatureErrorListener(metamaskId, this.onSignatureError);
```
cryptotavares, line 98: Why are we removing registering the listner for signature errors?

app/util/blockaid/index.ts
```diff
@@ -12,22 +12,26 @@ export const getBlockaidMetricsParams = (
   securityAlertResponse?: SecurityAlertResponse,
 ) => {
   const additionalParams: Record<string, any> = {};
+  additionalParams.security_alert_reason = Reason.notApplicable;
```
cryptotavares, line 15: Do we want to record `security_alert_reason` as not applicable, even when blockaid feature is not enabled?

seaona:
>  In both cases we send metrics when the result type changes. It would initially be undefined (we don't send any metrics), then it goes to Loading (we send loading metrics) and then go to actual alert (we send metrics of the securityAlertReponse)

I think this might not be 100% what we expect here, as in the case the validation is complete, for one transaction we will have now x2 Confirm Started events instead of 1, so we will see suddenly more Transactions added to the wallet, which won't be the reality. We should either inform the data team about this change or look for an approach of only 1 transaction being counted  (Confirm Started) as added to the wallet - no matter if there was change of validation state

Ie. Transactions with Blockaid will log more "Confirm Started" events than transactions without Blockaid.

I would like to clarify the behaviour. What do you think? cc @segun  @cryptotavares @bschorchit

app/components/UI/SignatureRequest/index.js
```diff
@@ -168,6 +174,36 @@ class SignatureRequest extends PureComponent {
...
+      InteractionManager.runAfterInteractions(() => {
+        AnalyticsV2.trackEvent(
+          MetaMetricsEvents.SIGNATURE_REQUESTED,
+          getAnalyticsParams(this.props, type),
+        );
+      });
+    }
+  };
+
   /**
```
cryptotavares, line 207: I am not sure this events should be fired in this component.. Currently the structure is:
* SignatureRequest/Root/Root.tsx -> signature component (PersonalSign, TypedSign, etc..) -> SignatureRequest/index.js or SignatureRequest/ExpandedMessage/index.js

And depending on the value of `showExpandedMessage` either this component or the `ExpandedMessage` can be loaded in the dom. This means that if the user toggles the `showExpandedMessage` this component gets mounted into the dom and as a result will fire a `SIGNATURE_REQUESTED` event more than once for the same signature request.

This does not seem right. I believe that if you want to consolidate this logic in a single place, then you need to add this to the `SignatureRequest/Root/Root.tsx` component.

app/components/UI/TypedSign/index.js
segun: You're right, this shouldn't be removed.

app/util/confirmation/signatureUtils.js
```diff
@@ -31,7 +31,9 @@ export const getAnalyticsParams = (messageParams, signType) => {
     );
 
     return {
-      account_type: getAddressAccountType(messageParams.from),
+      account_type: getAddressAccountType(
+        messageParams.from || messageParams.fromAddress,
```
cryptotavares, line 35: Why are we adding the `fromAddress`? `AbstractMessageParams` type only has `from` prop.. :thinking_face:

cryptotavares:
> > In both cases we send metrics when the result type changes. It would initially be undefined (we don't send any metrics), then it goes to Loading (we send loading metrics) and then go to actual alert (we send metrics of the securityAlertReponse)
> 
> I think this might not be 100% what we expect here, as in the case the validation is complete, for one transaction we will have now x2 Confirm Started events instead of 1, so we will see suddenly more Transactions added to the wallet, which won't be the reality. We should either inform the data team about this change or look for an approach of only 1 transaction being counted (Confirm Started) as added to the wallet - no matter if there was change of validation state
> 
> Ie. Transactions with Blockaid will log more "Confirm Started" events than transactions without Blockaid.
> 
> I would like to clarify the behaviour. What do you think? cc @segun @cryptotavares @bschorchit

I agree with @seaona this does not seem the desired behaviour. Should we have a specific blockaid status event that we fire for a given transaction/signature whenever the blockaid status is updated? @bschorchit @jpuri

app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -462,10 +469,12 @@ class ApproveTransactionReview extends PureComponent {
         spendLimitCustomValue: minTokenAllowance,
       },
       () => {
-        AnalyticsV2.trackEvent(
-          MetaMetricsEvents.APPROVAL_STARTED,
-          this.getAnalyticsParams(),
-        );
+        if (currentTransactionSecurityAlertResponse?.response) {
```
cryptotavares, line 472: Why are we preventing the `APPROVAL_STARTED` event from being fired when blockaid response is falsy? This code is running on the componentDidMount.. so if at time of mount, blockaid response is not set on the state, we will never trigger the `APPROVAL_STARTED` event. Or am I missing something?

app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -501,6 +510,27 @@ class ApproveTransactionReview extends PureComponent {
...
+        .result_type;
+    const newBlockaidResult =
+      this.props.transaction?.currentTransactionSecurityAlertResponse?.response
+        .result_type;
+
+    if (previousBlockaidResult !== newBlockaidResult) {
+      const additionalParams = getBlockaidMetricsParams(
+        this.props.transaction?.currentTransactionSecurityAlertResponse
+          ?.response,
+      );
```
cryptotavares, line 525: We are missing checking if `currentTransactionSecurityAlertResponse.id` matches the `transaction.id`. See @jpuri comment here - https://github.com/MetaMask/metamask-mobile/pull/7817#discussion_r1394264059

Also the `currentTransactionSecurityAlertResponse` is only reset from the state when a `RESET_TRANSACTION` action is fired. But for example it is not reset when `NEW_ASSET_TRANSACTION` is fired. This means that we need to either work through the transaction redux state lifecycle (this should be the correct approach.. although the effort and risk is considerable) or we need to validate that the `currentTransactionSecurityAlertResponse.id` matches the `transaction.id`.

app/components/UI/ApproveTransactionReview/index.js
segun: The assumption is that we will always have this object in state, and this just prevents sending one APPROVAL_STARTED when blockaid response is undefined and another one when it's defined.  

But thinking about it, this event won't fire on other networks except mainnet because PPOM runs only on main net.

I could remove the condition but it means we will have 2 APPROVAL_STARTED events in cases where the component is mounted before PPOM sends response.

cryptotavares:
Also besides all the above, it would be good to add unit tests, where we check that intended metrics event fired (once and only once) and that it contained blockaid expected data

app/components/UI/SignatureRequest/Root/Root.tsx
```diff
@@ -70,7 +75,7 @@ const Root = ({
     >
       {approvalType === ApprovalTypes.PERSONAL_SIGN && (
         <PersonalSign
-          messageParams={messageParams}
+          messageParams={{ ...messageParams, securityAlertResponse }}
```
jpuri, line 78: We can use selector in child components and do not need to pass `securityAlertResponse`

app/components/Views/Approval/index.js
```diff
@@ -324,7 +336,7 @@ class Approval extends PureComponent {
     this.showWalletConnectNotification();
     AnalyticsV2.trackEvent(
       MetaMetricsEvents.DAPP_TRANSACTION_CANCELLED,
-      this.getAnalyticsParams(),
+      this.getAnalyticsParams({ withBlockaid: true }),
```
jpuri, line 339: May be we can do it without `withBlockaid` variable. We add `securityAlertresponse` if its is present.

cryptotavares (commented):
It is looking good! 🚀 
Besides @jpuri comments, I would just add some unit tests to ensure that the `AnalyticsV2.trackEvent` is being called properly with the blockaid data

app/components/UI/MessageSign/MessageSign.tsx
```diff
@@ -71,60 +73,82 @@ const createStyles = (colors: Colors) =>
...
+  const navigation = useNavigation();
+  const [truncateMessage, setTruncateMessage] = useState<boolean>(false);
+  const { securityAlertResponse } = useSelector(
+    (reduxState: any) => reduxState.signatureRequest,
+  );
+
+  const { colors } = useTheme();
+  const styles = createStyles(colors);
+
+  // static contextType = ThemeContext;
```
blackdevelopa, line 93: we removing this?

ios/MetaMask.xcodeproj/project.pbxproj
blackdevelopa, line 1: 🪓

app/components/UI/MessageSign/index.test.tsx
```diff
@@ -168,92 +182,55 @@ describe('MessageSign', () => {
     });
   });
 
-  describe('componentDidMount', () => {
```
segun, line 171: This test is testing methods called in componentDidMount, as we've moved to functional component, these methods are now called in useEffect.

app/components/UI/MessageSign/index.test.tsx
```diff
@@ -168,92 +182,55 @@ describe('MessageSign', () => {
...
-        expect.any(Function),
-      );
-    });
-  });
+  describe('trackEvent', () => {
+    it('tracks event for rejected requests', async () => {
+      const wrapper = createWrapper().dive();
+      await (wrapper.find(SignatureRequest).props() as any).onReject();
 
-  describe('unmount', () => {
```
segun, line 200: Same as above, we are no longer using class component and these methods are now called in useEffect.

app/components/UI/MessageSign/index.test.tsx
```diff
@@ -168,92 +182,55 @@ describe('MessageSign', () => {
...
-        'TestMessageId:signError',
-        expect.any(Function),
-      );
-    });
-  });
+      const rejectedMocks = (
+        analyticsV2.trackEvent as jest.Mock
+      ).mock.calls.filter((call) => call[0].category === 'Signature Rejected');
 
-  describe('shouldTruncateMessage', () => {
```
segun, line 230: These tests are wrong!!! the test is calling the method `shouldTruncateMessage` and checking that the state is updated, instead of checking the side effect of the method being called when component is loaded.

app/components/UI/MessageSign/index.test.tsx
```diff
@@ -168,92 +182,55 @@ describe('MessageSign', () => {
...
+        security_alert_response: 'Benign',
+        security_alert_reason: '',
+        ppom_eth_chainId_count: 1,
+        version: undefined,
       });
-      expect(instance.state.truncateMessage).toBe(false);
     });
-  });
 
-  describe('onSignatureError', () => {
```
segun, line 250: There are new event tracking tests.

segun (commented):
I've added some comments to MessageSign tests for the reason why some tests are removed. This should apply to the TypedSign too.

app/components/UI/ApproveTransactionReview/index.js
```diff
@@ -686,7 +692,8 @@ class ApproveTransactionReview extends PureComponent {
 
   onContactUsClicked = () => {
     const analyticsParams = {
-      ...this.getAnalyticsParams(),
+      ...this.getAnalyticsParams(true),
```
cryptotavares, line 695:
```suggestion
      ...this.getAnalyticsParams(),
```

app/components/UI/MessageSign/index.test.tsx
cryptotavares: I'll re-add this tests using react testing library in a subsequent PR.

cryptotavares (approved):

blackdevelopa (approved):

Issue #19, opened 2018-08-01
Investigate GABA module format


bitpshr:
If GABA isn't compiled to CommonJS, the Metro bundler blows up. We should look into this, as GABA should ideally be imported into this project as ES. This is a Metro-specific issue.

EtDu:
Does not seem to be an issue anymore

PR #7873, opened 2023-11-18
refactor: split migrations into multiple files

### LINKED ISSUE #7845: Split mobile migrations into separate files
**Background**

All state migrations functions in mobile are currently defined in a single file:

[app/store/migrations.js](https://github.com/MetaMask/metamask-mobile/blob/main/app/store/migrations.js)

All associated unit tests are also defined in a single file:

[app/store/migrations.test.js](https://github.com/MetaMask/metamask-mobile/blob/main/app/store/migrations.test.js)

This could limit the scalability of the migrations and hinder future maintenance.

**Goal**

Split each migration function and the associated unit tests into separate files.

**Notes**

Consider the extension approach which also exports the migration version from each individual migration file:

[app/scripts/migrations/102.ts](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/migrations/102.ts)

dbrans:
## **Description**

Split each migration function and the associated unit tests into separate files. 

See the extension approach which also exports the migration version from each individual migration file ([example](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/migrations/102.ts))

## **Related issues**

Fixes: #7845 

- [x] Run e2e test as smoke test
- [x] Ensure test for migration 25 is unaffected


app/store/migrations/019.test.js
```diff
@@ -0,0 +1,23 @@
+import migrate from './019';
+
+describe('#19', () => {
```
matthewwalsh0, line 3: Since these are now independent files, shall we update the describes to `Migration #19` so they have a little more context?

app/store/migrations/index.ts
```diff
@@ -0,0 +1,69 @@
+import { MigrationManifest } from 'redux-persist';
+
+import m0 from './000';
```
matthewwalsh0, line 3: This was recently changed on the extension so we can import directly into the list, is it worth doing the same here?

https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/migrations/index.js

app/store/migrations/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  m24,
+  m25,
+  m26,
+];
+
+export const version = allMigrations.length - 1;
+
+const migrations: MigrationManifest = {};
+
+allMigrations.forEach((m, i) => {
```
matthewwalsh0, line 65: ❤️  I really like this as it means we don't have to define the `version` in each file as the extension does.

The only think to remember is to always add to the bottom and to not remove any migrations as that will reduce the current version number.

app/store/persistConfig.ts
```diff
@@ -3,7 +3,7 @@ import AsyncStorage from './async-storage-wrapper';
 import FilesystemStorage from 'redux-persist-filesystem-storage';
 import autoMergeLevel2 from 'redux-persist/lib/stateReconciler/autoMergeLevel2';
 import { RootState } from '../reducers';
-import { migrations, version } from './migrations';
+import migrations, { version } from './migrations';
```
matthewwalsh0, line 6: Shall we use a standard export rather than a `default` one since we have multiple?

app/store/migrations/index.ts
matthewwalsh0: I wonder if it's worth a very basic unit test to assert the current version to catch this if it decreases?

app/store/migrations/019.test.js
dbrans: Done.

app/store/migrations/index.ts
dbrans: Done.

app/store/migrations/index.ts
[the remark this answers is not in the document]
dbrans: Thoughts on these changes?

dbrans: The new code ensures that no version number is skipped. Thoughts?

app/store/persistConfig.ts
dbrans: Done.

app/store/migrations/index.ts
```diff
@@ -0,0 +1,15 @@
...
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+import { MigrationManifest } from 'redux-persist';
+
+// The latest (i.e. highest) version number.
+export const version = 26;
+
+export const migrations: MigrationManifest = {};
+
+for (let i = 0; i < version; i++) {
```
matthewwalsh0, line 12: My only concern with this approach is we already have dynamic code in the import at the static level, plus now dynamic imports so I'm not sure what the error would look like if we miss a migration.

Would a safe compromise be the explicit array as before using `require` like in the extension, but assert the number of migrations in a `index.test.ts` unit test explicitly?

dbrans: I added a try/catch to provide a readable error. Does that satisfy this specific concern? I want to make sure I fully understand, so I'm happy to hop on a huddle tomorrow.

matthewwalsh0: Good idea, that definitely helps, but my only pending concern is how this works with the mobile build system and if it supports these dynamic imports. Happy with the logic itself though.

matthewwalsh0 (dismissed):

app/store/migrations/index.ts
dbrans: You raise a good point: in any environment, dynamically loading code is not ideal. Static analysis tools, like optimizers, need to know what's being imported. 

I've updated the code so that the filenames are static, which I believe is what you had suggested earlier when I had verbose imports.

app/store/migrations/027.test.js
```diff
@@ -0,0 +1,262 @@
+import migration from './027';
+
+describe('#27', () => {
```
matthewwalsh0, line 3: Super minor, but shall we make this `Migration #27`? I appreciate this was added after the PR via a merge.

matthewwalsh0 (dismissed):

matthewwalsh0 (approved):

vinistevam (approved):

PR #7897, opened 2023-11-21
test: Separate Detox build from test execution

vinistevam:
## **Description**
This PR aims to split the detox build and test execution in different steps on Bitrise.
It was necessary to split the commands for ios `test:e2e:ios:bitrise:build` / `test:e2e:ios:bitrise:run` and android `test:e2e:android:bitrise:build` / `test:e2e:android:bitrise:run`.
To reuse `node_modules`, `yarn.lock` and `build outputs` it was introduced [save-cache](https://bitrise.io/integrations/steps/save-cache) because it supports key-based caching in addition to [save-gradle-cache](https://bitrise.io/integrations/steps/save-gradle-cache).

The previous `android/ios_e2e_test` was split accordingly into `android/ios_e2e_build` where the dependencies and build are prepared and those outputs saved into the cache, and `android/ios_e2e_test` is where those caches are restored, the test is executed and saves the reports or screenshots in case of failure.

**Benefits**
- Cache dependencies and rebuild only when something changes in the `yarn.lock`.
- Cache cocoapods and only rebuild once it changes.
- Cache build to reuse across stages but also when re-running with the same commit.
-  Further parallelization of the test using shards or modules: https://devcenter.bitrise.io/en/builds/build-pipelines/currently-supported-use-cases-for-the-android-platform.html
- Run tests without build everything from scratch

**Limitations**
- Currently, it is necessary to wait for the completion of the first stage before initiating the next one.

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1311

## **Manual testing steps**

1. Go to Bitrise and run `pr_smoke_e2e_pipeline`

**SMOKE**
Pipeline without cache: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/0a971976-089b-4530-a10e-fe86ae1af1eb
Pipeline when build is cached: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/bb021668-0473-4461-bc98-f6b71f53cd33

**RECRESSION**
Pipeline without cache: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/5a5d101a-bfa8-4931-88d2-bf7f0b3a11f0

Pipeline: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/0a971976-089b-4530-a10e-fe86ae1af1eb 

### **After**

```mermaid
stateDiagram-v2
    direction LR
    [*] --> code_setup_dev
    state Build {
        direction LR
        code_setup_dev --> setup
        setup --> prep_environment
        prep_environment --> code_setup
        code_setup --> android_e2e_build
    }
     android_e2e_build --> [*]
```

```mermaid
stateDiagram-v2
    direction LR
    [*] --> setup
    state Test {
        direction LR
        setup --> prep_environment
        prep_environment --> e2e_setup
        e2e_setup --> android_e2e_test
    }
    android_e2e_test --> [*]
```


bitrise.yml
```diff
@@ -122,28 +142,10 @@ workflows:
...
-                asdf list all nodejs
-                asdf install nodejs "$NODE_VERSION"
-                asdf global nodejs "$NODE_VERSION"
-                asdf plugin add yarn
-                asdf list all yarn
-                asdf install yarn "$YARN_VERSION"
-                asdf global yarn "$YARN_VERSION"
-          title: Install node and yarn
       - yarn@0:
+          # run_if: '{{not (enveq "BITRISE_CACHE_HIT" "exact")}}'
```
vinistevam, line 148: It can save an extra 2-4 min, however, every time it skips `yarn setup` the pipeline fails on Android in the build process and ios on trying to run `detox test`. I believe `yarn setup` does more things behind the scenes such as in `postinstall.sh` so when it caches it is necessary to cache more than just `yarn.lock` and `node_modules`. I think we could handle it separately.

bitrise.yml
```diff
@@ -155,12 +157,13 @@ workflows:
           inputs:
             - command: audit:ci
           title: Audit Dependencies
-      - cache-push@2:
-          # Manual cache path needed here because the `yarn` step only caches for the `install` command
+      - save-cache@1:
```
vinistevam, line 160: Using save-cache instead of cache-push, because it supports key-based caching

bitrise.yml
```diff
@@ -171,6 +174,25 @@ workflows:
                 printf '%s\n\n%s' '@metamask:registry=https://npm.pkg.github.com' "//npm.pkg.github.com/:_authToken=${PACKAGE_READ_TOKEN}" > .npmrc
     after_run:
       - code_setup
+  prep_environment:
```
vinistevam, line 177: Created `prep_environment` step because as in the build step it already checked `lint`, `audit`, `yarn setup` is not necessary to do it in run tests step. So leaving here just what is necessary.

bitrise.yml
```diff
@@ -317,34 +353,161 @@ workflows:
...
+        title: Download cmake 3.22.1 with sdkmanager
+        is_always_run: false
+        inputs:
+        - content: |-
+            #!/usr/bin/env bash
+            "${ANDROID_HOME}/tools/bin/sdkmanager" "cmake;3.22.1"
+    - save-gradle-cache@1: {}
+    - script@1:
+        title: Run detox build
+        run_if: '{{not (enveq "BITRISE_CACHE_HIT" "exact")}}'
```
vinistevam, line 382: Skips build based on the `restore-cache` response, `BITRISE_CACHE_HIT` is the output of the last `restore-cache` based on the key. This cache can be improved using [checksum function provided by](https://bitrise.io/integrations/steps/save-cache#:~:text=checksum%20%3A%20This%20function%20takes%20one,%2D%7B%7B%20checksum%20%22package%2Dlock.)  `save-cache`.  Not sure if the checksum provided supports multiple folders but we could do something on our side to determine a unique cache key and skip it every time something changes in the determined files/folders.

bitrise.yml
```diff
@@ -31,32 +31,37 @@ pipelines:
...
   run_e2e_android_pipeline:
     stages:
-      - run_e2e_android_stage: {} #runs android detox E2E
+      - build_e2e_android_stage: {} #builds android detox E2E
+      - run_e2e_android_stage: {} #runs android detox test E2E
       - notify: {}
   #PR_e2e_verfication (build ios & android), run iOS (smoke), emulator Android
   release_e2e_pipeline:
     stages:
       - run_e2e_ios_android_stage: {}
```
matthewwalsh0, line 46: If this is using `ios_e2e_test` and `android_e2e_test`, do we not need to add the equivalent build stage here also to ensure it functions the same as before?

bitrise.yml
```diff
@@ -155,12 +157,13 @@ workflows:
...
           title: Audit Dependencies
-      - cache-push@2:
-          # Manual cache path needed here because the `yarn` step only caches for the `install` command
+      - save-cache@1:
+          title: Save cache dependencies 
           inputs:
-            # The Yarn cache is easier and safer to cache than `node_modules`, which could include
-            # postinstall script modifications and could affect hoisting on subsequent installs.
-            - cache_paths: '$YARN_CACHE_DIR -> ./yarn.lock'
+          - key: '{{ .OS }}-{{ .Arch }}-node-modules-{{ checksum "yarn.lock" }}'
```
matthewwalsh0, line 163: Can I clarify how this is different to the previous dependency caching?

Previously we were using the Yarn cache to cache the dependency archives but now we're explicitly caching the `node_modules` directory.

Does that not hit the issues mentioned in the old comment concerning post install scripts?

Will this mean the first `code_setup` is slower since it's no longer using the native Yarn cache?

bitrise.yml
```diff
@@ -317,34 +353,161 @@ workflows:
...
+              TEST_SUITE="Smoke"
+            fi
+            node -v
+            export METAMASK_ENVIRONMENT='local'
+            export METAMASK_BUILD_TYPE='main'
+            IGNORE_BOXLOGS_DEVELOPMENT="true" FORCE_BUNDLING=true yarn test:e2e:android:bitrise:build --testNamePattern="$TEST_SUITE"
+    - save-cache@1:
+        title: Save cache build
+        inputs:
+        - key: apk-cache-{{ .CommitHash }}
```
matthewwalsh0, line 399: If the build caches are keyed with the commit hash, does this mean the cache will be ignored on subsequent commits even if only the E2E tests have changed?

Could we use a persistent key such as a hash of a directory?

bitrise.yml
```diff
@@ -31,32 +31,37 @@ pipelines:
   #Run E2E test suite for iOS only
   run_e2e_ios_pipeline:
     stages:
+      - build_e2e_ios_stage: {}
```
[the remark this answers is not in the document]
vinistevam: That is a great question. 
From what I researched in the Bitrise documentation, stages run sequentially while every workflow within the stage runs simultaneously. So the workflows in a stage can be different devices, sharding tests by modules or VMs, while the stages are the next step dependent on the previous stage result in the pipeline. Generally from the examples I found the pipelines are different by OSs, but it's not clear what is the recommendation of Bitrise for that.

The major benefit will be in the next ticket where we will cache the native modules build, reducing the build time when not necessary a full rebuild and on top of that use test parallelization with different VMs. It can be done by dynamically creating multiple workflows in the run test stage https://devcenter.bitrise.io/en/builds/build-pipelines/currently-supported-use-cases-for-the-android-platform.html#-android--unit-test-sharding-by-module . 

For now, we gained, the ability to rerun the test without having to run all the build again and the infrastructure to leverage it and use multiple VMs to parallelize the tests. I would say to simplify this PR we could have just split the steps Build and another one for Tests as you mentioned but in the next ticket we would need to split into stages, cache artefacts between stages and find a solution to cache the native modules and just rebuild when something has changed.

bitrise.yml
vinistevam: Definitely, but I thought this was part of the next ticket. I tried to use the `checksum` function provided by bitrise but it only supports specific files, so we might need to create a script step to hash a directory/directories and use that as a key or even run in this script `git diff` to check the differences between commits.

bitrise.yml
vinistevam: I've done this way cause using the previous one has resulted in a bigger cache but concerning the comment and also an issue that I'm facing https://github.com/MetaMask/metamask-mobile/pull/7897#discussion_r1407433574 not being able to skip the yarn setup step properly and gain 2-4 min. I will cache the yarn cache using the `save-cache` instead of `cache-push`. Thanks for pointing out it might be the answer to save more time in the pipeline. :pray:

vinistevam: After some investigation, I don't think the cache was used before at all because there was no validation before running `yarn setup` so it always uses the `node_modules` with the `postinstall` script and several extra modifications done when running `setup`.
Unfortunately, `node_modules` is necessary to run `detox, It was not evident before cause everything was running in one stage and every time it was running `yarn setup`, alternatively, we could run on every stage but it would cost us an extra 2min.

bitrise.yml
vinistevam: Included `build_e2e_ios_android_stage` to cover `android_e2e_build` and `ios_e2e_build`

Andepande (approved):
LGTM I was able to branch off and understand the implementation better.

bitrise.yml
```diff
@@ -338,34 +384,163 @@ workflows:
...
+            #!/usr/bin/env bash
+            if [ "$TEST_SUITE" = "Regression" ]; then
+              TEST_SUITE="Regression"
+            else
+              TEST_SUITE="Smoke"
+            fi
+            node -v
+            export METAMASK_ENVIRONMENT='local'
+            export METAMASK_BUILD_TYPE='main'
+            IGNORE_BOXLOGS_DEVELOPMENT="true" FORCE_BUNDLING=true yarn test:e2e:android:bitrise:build --testNamePattern="$TEST_SUITE"
```
Andepande, line 426: May remove the build testNamePattern, as we only require yarn test:e2e:android:bitrise:build to build

bitrise.yml
```diff
@@ -377,49 +552,29 @@ workflows:
                 node -v
                 export METAMASK_ENVIRONMENT='local'
                 export METAMASK_BUILD_TYPE='main'
-                IGNORE_BOXLOGS_DEVELOPMENT="true" FORCE_BUNDLING=true yarn test:e2e:android:bitrise:debug --testNamePattern="$TEST_SUITE"
-          title: Detox Build & Test
+                IGNORE_BOXLOGS_DEVELOPMENT="true" FORCE_BUNDLING=true yarn test:e2e:ios:bitrise:build --testNamePattern="$TEST_SUITE"
```
Andepande, line 555: Remove the test pattern --testNamePattern="$TEST_SUITE" as we are only building with this step workflow

bitrise.yml
```diff
@@ -192,6 +205,25 @@ workflows:
...
+                #!/usr/bin/env bash
+                echo "Node and Yarn being installed"
+                asdf list all nodejs
+                asdf install nodejs "$NODE_VERSION"
+                asdf global nodejs "$NODE_VERSION"
+                asdf plugin add yarn
+                asdf list all yarn
+                asdf install yarn "$YARN_VERSION"
+                asdf global yarn "$YARN_VERSION"
+          title: Install node and yarn
```
NicolasMassart, line 222: Q: is there anything we could cache here to make asdf load tools from some cache instead?

NicolasMassart (approved):
Great job on the build-test split and smart caching! I added questions but just curious me, not required for approval.

bitrise.yml
vinistevam: I think somehow is already doing that because this step is taking 3-10 seconds, but there are certainly more places to take advantage of caching, such as the `Yarn Setup`.  I've tried to see if I could do it quickly but `setup` does several things and I didn't manage to get to a stable use of cache. We will gain a lot properly caching this step but it is better to take it into a separate ticket.

bitrise.yml
vinistevam: If is ok, I can remove it in the ccache https://github.com/MetaMask/metamask-mobile/pull/7975

bitrise.yml
```diff
@@ -338,34 +384,163 @@ workflows:
...
+            #!/usr/bin/env bash
+            if [ "$TEST_SUITE" = "Regression" ]; then
+              TEST_SUITE="Regression"
+            else
+              TEST_SUITE="Smoke"
+            fi
+            node -v
+            export METAMASK_ENVIRONMENT='local'
+            export METAMASK_BUILD_TYPE='main'
+            IGNORE_BOXLOGS_DEVELOPMENT="true" FORCE_BUNDLING=true yarn test:e2e:android:bitrise:run --testNamePattern="$TEST_SUITE"
```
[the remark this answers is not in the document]
vinistevam: if it's ok, I'm planning to remove it in the cache PR https://github.com/MetaMask/metamask-mobile/pull/7897#discussion_r1423699002

matthewwalsh0 (approved):

bitrise.yml
Gudahtt: To clarify, the cache was never meant to skip `yarn setup`. It was only intended to save us from having to download dependencies. Every other part of setup was run every time, we did not attempt to skip past those. We did confirm it worked when it was first setup.

I would still expect the concerns in the comment to apply here. That is, on rare occasions, this method of installation would result in a different directory structure within `node_modules` that would not be easily replicated locally, and this in turn could result in different behavior than you'd see on a fresh install. This is more of a concern with the extension because LavaMoat policies rely on the exact directory structure of `node_modules`, maybe that's less problematic here because LavaMoat isn't used yet, but I'm not sure.

I'd also expect this to break sometimes when the Node.js version changes, as some packages have setup steps that differ depending on Node.js versions, and now that would get cached.

PR #7975, opened 2023-12-01
test: Implement `CCache` to e2e builds

OGPoyraz:
## **Description**

This pull request aims to incorporate the `ccache` module for compiler caching in end-to-end (e2e) tests, as suggested in the [React Native / build speed documentation](https://reactnative.dev/docs/build-speed#use-a-compiler-cache).

## Cache Mechanism

Ccache wraps C++ compilers, stores compilation results, and skips recompilation if the intermediate result is already cached. More details about `ccache` can be found [here](https://ccache.dev/). Essentially, `ccache` stores cache files under the `./ccache` folder in the working directory (configured using `ccache --set-config cache_dir=$(pwd)/ccache` in the scripts). The `ccache` folder is then saved using Bitrise's `save-cache` step script.

## Cache Saving Criteria

There are two types of caches saved:

1. **`main` branch cache:** Used when there are no changes in native modules/dependencies. The identification of native changes is detailed in issue #XXXX.

2. **Branches with changes in native modules/dependencies:** Created in the first run when the build is executed for a specific branch. Subsequent branch builds benefit from this cache.

## Cache Download/Usage Logic

Refer to the `Check native folder diff for iOS/Android` step for precise behavior details. In summary:
- For `main` builds, cache download is skipped (`SKIP_CCACHE_RESTORE` set to true), but the `main` cache is saved for future builds.
- For non-main builds, differences in native folders against the `main` branch are checked. If there are differences, a cache is created for this branch; otherwise, the `main` cache is used.

After defining the cache, the build continues as usual.

## Workflow Integration

This PR targets e2e test builds exclusively, aiming to accelerate development.

## Example for iOS & Android

Since tests are flaky, exact event times are noted. Please disregard the `Duration` in the build links.

### IOS build without CCache (first run - cached at the end of build)
Link for the build: https://app.bitrise.io/build/502ebf4a-a99b-4173-b1f9-56dea1a8927c
```
Build started: 09:29:10
Build succeed: 09:35:51
Total build time: **6 min 41 sec**
```

### IOS build with CCache (second run - retrieved cache at the beginning of build)
Link for the build: https://app.bitrise.io/build/1cb385ca-b2a2-42dc-8d2e-fa833f186bb5
```
Build started: 09:58:22
Build succeed: 10:01:26
Total build time: **3 min 04 sec**
```

#### CCache statistics on the second iOS run
Note: %50 of hits expected because we also retrieved cache statistics from first run. 
```
Cacheable calls:    6096 / 6098 (99.97%)
  Hits:             3048 / 6096 (50.00%)
    Direct:         3048 / 3048 (100.0%)
    Preprocessed:      0 / 3048 ( 0.00%)
  Misses:           3048 / 6096 (50.00%)
Uncacheable calls:     2 / 6098 ( 0.03%)
Local storage:
  Cache size (GiB):  0.6 /  5.0 (12.21%)
  Hits:             3048 / 6096 (50.00%)
  Misses:           3048 / 6096 (50.00%)
```

**IOS build performance improvement: 3 min 37 sec** 

---

### Android build without CCache (first run - cached at the end of build)
Link for the build: https://app.bitrise.io/build/51e80079-4309-47c9-9146-3b8e9fc6f154
```
Build started: 10:47:48
Build succeed: 11:11:05
Total build time: **23 min 17 sec**
```

### Android build with CCache (second run - retrieved cache at the beginning of build)
Link for the build: https://app.bitrise.io/build/2f61ba97-c65a-4118-b4f5-dc06c08ccf5f 
```
Build started: 11:50:04
Build succeed: 12:11:50
Total build time: **21 min 46 sec**
```

#### CCache statistics on the second Android run
```
cache hit (direct)                  1480
cache hit (preprocessed)               0
cache miss                          1480
cache hit rate                     50.00 %
called for link                      664
called for preprocessing             144
no input file                          8
cleanups performed                     0
files in cache                      4440
cache size                           1.6 GB
max cache size                       5.0 GB
```
**Android build performance improvement: 1 min 33 sec** 

---

## **Related issues**

https://github.com/MetaMask/mobile-planning/issues/1344 

## **Manual testing steps**

N/A

## **Screenshots/Recordings**

N/A


android/build.gradle
```diff
@@ -32,3 +32,4 @@ buildscript {
     }
 	}
 }
+
```
OGPoyraz, line 35: This is an intentional change to trigger branch based cache. I will remove it later.

ios/MetaMask/AppDelegate.m
```diff
@@ -127,3 +127,4 @@ - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 }
 
 @end
+
```
OGPoyraz, line 130: This is an intentional change to trigger branch based cache. I will remove it later.

bitrise.yml
```diff
@@ -437,10 +488,54 @@ workflows:
           title: Set Env Path for caching deps & Ruby install
       - cocoapods-install@2:
           is_always_run: false
+      - script@1: 
+          title: Check native folder diff for iOS
+          inputs:
+            - content: |-
+                #!/usr/bin/env bash
```
sethkfman, line 495: Can we move these scripts out into the scripts directory of the code base?

ccache.conf
```diff
@@ -0,0 +1,6 @@
+export CCACHE_SLOPPINESS=clang_index_store,file_stat_matches,include_file_ctime,include_file_mtime,ivfsoverlay,pch_defines,modules,system_headers,time_macros
```
sethkfman, line 1: Can you add a comment to this file to explain it's purpose?

bitrise.yml
```diff
@@ -365,10 +365,48 @@ workflows:
             - profile: pixel_5
       - wait-for-android-emulator@1: {}
       - save-gradle-cache@1: {}
+      - script@1: 
+          title: Check native folder diff for Android
+          inputs:
+            - content: |-
+                #!/usr/bin/env bash
```
sethkfman, line 372: Move this script out to the scripts directory maybe place it in a sub folder cache

bitrise.yml
vinistevam: That is a great point. I've created a script `check-native-folders-diff.sh` and abstracted a bit to support `iOS` and `Android`.

ccache.conf
vinistevam: Added more comments.

bitrise.yml
vinistevam: Done, as I mentioned above now we are using one script `check-native-folders-diff.sh` to support both OS's.

scripts/check-native-folders-diff.sh
matthewwalsh0, line 1: As this script doesn't reference any specific native folders plus the check itself isn't really the goal, would a more explicit name such as `set-ccache-envs.sh` be more helpful?

scripts/check-native-folders-diff.sh
```diff
@@ -0,0 +1,42 @@
...
+
+    # Check if there are no differences
+    if [ -z "$differences" ]; then
+      # If no differences, use the main cache and skip cache upload
+      echo "No differences detected, we will use main cache"
+      envman add --key SKIP_CCACHE_UPLOAD --value true
+      envman add --key USE_MAIN_CCACHE --value true
+    else
+      # If there are differences, write them to the output file
+      echo "$differences" > $output_file
```
matthewwalsh0, line 27: I understand we're writing this to a file, so we can checksum it in Bitrise and use that checksum to generate a key for the cache:

`'android-ccache-{{if getenv "USE_MAIN_CCACHE" }}main{{else}}{{ checksum "android-diff.txt"}}{{end}}'`

But could we simplify this process and encapsulate more of the logic in this script if we just generate the checksum here using something like `sha512sum` and then we could even create the cache key here and put it in an env such as `CCACHE_KEY`.

This would mean we don't have to create any additional files, we wouldn't have to duplicate the key logic 4 times in `bitrise.yml`, and we wouldn't need the `USE_MAIN_CCACHE` env as we can set the cache key in either scenario.

We could even add a third argument to this script such as `prefix` so we can pass in `android` or `ios`.

bitrise.yml
```diff
@@ -437,10 +465,37 @@ workflows:
...
+      - restore-cache@1:
+          title: Restore CCache
+          run_if: '{{not (enveq "SKIP_CCACHE_RESTORE" "true")}}'
+          inputs:
+          - key: 'ios-ccache-{{if getenv "USE_MAIN_CCACHE" }}main{{else}}{{ checksum "ios-diff.txt"}}{{end}}'
       - script@1:
           inputs:
             - content: |-
                 #!/usr/bin/env bash
+                ccache --config-path $(pwd)/ccache.conf
```
matthewwalsh0, line 495: Is this line intentionally missing in the android workflow?

Is it worth moving these ccache setup lines into an additional script such as `cache/setup-ccache.sh` to avoid the duplication?

We could move the `check-native-folders-diff.sh` into the `cache` folder also as Seth suggested.

ccache.conf
```diff
@@ -0,0 +1,19 @@
...
+export CCACHE_FILECLONE=true
+
+# Enable dependency tracking to ensure accurate caching
+export CCACHE_DEPEND=true
+
+# Enable inode caching for faster cache lookups
+export CCACHE_INODECACHE=true
+
+# Set the maximum size for the Ccache cache directory to 5 gigabytes
+export CCACHE_MAXSIZE=5G
```
matthewwalsh0, line 16: Is this a meaningful value for our scenario?

And is there any capacity limits on the Bitrise cache itself across all our builds?

bitrise.yml
vinistevam: Great point created additional script and moved both to `scripts/cache`.

ccache.conf
vinistevam: You are right it's not relevant here, ccache by default is 5G. I thought it was here because Bitrise does not have a hard limit but they don't recommend having caches without any limit.

scripts/check-native-folders-diff.sh
vinistevam: Nice, using `sha512sum` and setting the key based on the difference instead of creating the file. I will test to see if we can remove the `USE_MAIN_CCACHE` because this looks more like a fallback to me.

bitrise.yml
```diff
@@ -380,6 +398,14 @@ workflows:
...
           title: Detox Build & Test
           is_always_run: false
+      - save-cache@1:
+          title: Save CCache
+          run_if: '{{not (enveq "SKIP_CCACHE_UPLOAD" "true")}}'
+          inputs:
+          - key: 'android-ccache-{{if getenv "USE_MAIN_CCACHE" }}main{{else}}{{ getenv "CCACHE_KEY" }}{{end}}'
+          - paths: |-
+              ccache
+      - save-gradle-cache@1: {}
```
vinistevam, line 408: Repositioned the Gradle save operation after the build process.

scripts/cache/set-cache-envs.sh
```diff
@@ -0,0 +1,41 @@
...
+
+    # Check if there are no differences
+    if [ -z "$differences" ]; then
+      # If no differences, use the main cache and skip cache upload
+      echo "No differences detected, we will use main cache"
+      envman add --key SKIP_CCACHE_UPLOAD --value true
+      envman add --key USE_MAIN_CCACHE --value true
+    else
+      # Generate a checksum for the differences and set it as CCACHE_KEY
+      local checksum=$(echo "$differences" | sha512sum | awk '{print $1}')
```
matthewwalsh0, line 26: This is really great but my original intent was to go even deeper and set the entire ccache key used by Bitrise to avoid the duplicated logic and encapsulate more in this script.

Rather than set `USE_MAIN_CCACHE` we could just update the `CCACHE_KEY` accordingly, plus we could add a `platform` argument to this script which would provide the `android` or `ios` string.

This would mean we could go from:

```
inputs:
            - key: 'android-ccache-{{if getenv "USE_MAIN_CCACHE" }}main{{else}}{{ getenv "CCACHE_KEY" }}{{end}}'
```

to: 

```
inputs:
            - key: '{{ getenv "CCACHE_KEY" }}'
```

vinistevam: Sorry, now I see.
I pushed the changes now `set-cache-env.sh` is responsible for building the `CCACHE_KEY`. That is a great improvement, thanks for the input.

matthewwalsh0 (dismissed):

scripts/build.sh
```diff
@@ -467,7 +467,7 @@ buildAndroidRunE2E(){
 	then
 		source $ANDROID_ENV_FILE
 	fi
-	cd android && ./gradlew assembleProdDebug app:assembleAndroidTest -DtestBuildType=debug && cd ..
+	cd android && ./gradlew assembleProdDebug app:assembleAndroidTest -DtestBuildType=debug -x app:createBundleQaDebugJsAndAssets -x app:createBundleFlaskDebugJsAndAssets --build-cache --parallel && cd ..
```
vinistevam, line 470: After introducing the 'flask' build type, @Andepande and I noticed a significant increase in Android build times, approximately an additional 7 minutes. Upon investigation using the profile report of Gradle tasks, it became apparent that unnecessary bundle tasks, including those for 'flask', were being executed despite not being utilized in our current workflow.

To address this, I've excluded these unnecessary tasks from the build process, resulting in successful Bitrise test runs (https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/82c50886-bc28-45b8-b7f7-c11e6e0fefdf) reducing the build time from 30min to 18min. While this approach has improved build times, I acknowledge that there may be more optimal solutions. I welcome any suggestions or feedback on the proposed removal of -x app:createBundleQaDebugJsAndAssets -x app:createBundleFlaskDebugJsAndAssets and would appreciate insights on potential optimizations.

image

vinistevam: @Cal-L @sethkfman I would greatly appreciate your insights on this PR as a whole, with a specific focus on this modification I've made.  :pray:

vinistevam: Perfect, excluded the appropriate gradle tasks for each of the `buildAndroid` functions.

dbrans (approved):

matthewwalsh0 (approved):

sethkfman (approved):
LGTM

Issue #7903, opened 2023-11-22
[BUG] Fix the TokenRateController API


sethkfman:
### What is this about?

This issue is to replace the CoinGecko API on the [TokenRates controller](https://github.com/MetaMask/core/blob/fda42694d534e1a49cf97d072178615a63ad732b/packages/assets-controllers/src/TokenRatesController.ts#L141). The API should be replaced with the public PriceAPI written by the codefi team which is documented [here](https://price-api.metafi.codefi.network/docs#/Spot%20Prices/PriceController_getSpotPriceByTokenAddressAsync).

### Scenario

- GIVEN a user has an account and assets in their wallet
- WHEN they navigate to the wallet view
- AND they review their assets 
- AND they have a network connection
- THEN there is a fiat balance displayed

### Technical Details

- Create a TokenRatesController patch that replaces the  [CoinGecko API](https://github.com/MetaMask/core/blob/fda42694d534e1a49cf97d072178615a63ad732b/packages/assets-controllers/src/TokenRatesController.ts#L141)
   - Replace the `getTokenPriceURL` to uses the PriceAPI call GET
[/v1/chains/{chainId}/spot-prices/{tokenAddress}](https://price-api.metafi.codefi.network/docs#/Spot%20Prices/PriceController_getSpotPriceByTokenAddressAsync)
   - Remove the `getSupportedVsCurrencies` this should not be needed
   - Return a hardcode list for `getPlatformsURL` 
```
  1: 'ethereum',
  10: 'optimistic-ethereum',
  25: 'cronos',
  56: 'binance-smart-chain',
  57: 'syscoin',
  66: 'okex-chain',
  70: 'hoo-smart-chain',
  82: 'meter',
  88: 'tomochain',
  100: 'xdai',
  106: 'velas',
  122: 'fuse',
  128: 'huobi-token',
  137: 'polygon-pos',
  250: 'fantom',
  288: 'boba',
  321: 'kucoin-community-chain',
  324: 'zksync',
  361: 'theta',
  1088: 'metis-andromeda',
  1284: 'moonbeam',
  1285: 'moonriver',
  8453: 'base',
  8545: 'shiden network',
  10000: 'smartbch',
  42161: 'arbitrum-one',
  42220: 'celo',
  42262: 'oasis',
  43114: 'avalanche',
  59144: 'linea',
  333999: 'polis-chain',
  1313161554: 'aurora',
  1666600000: 'harmony-shard-0', 

```

[...]

PR #7991, opened 2023-12-01
fix: patch for token rates controller with coin gecko endpoint

tommasini:
## **Description**
Patch for update to new price api the fetch the token rates to deprecate the coin gecko endpoint

* Patch TokenRatesController of assets-controlllers v10
* Improved the number of requests made by network change/account change
* Increased pooling to 30 minutes
* Implemented chunks of 20 tokens, if a user have more than 20 tokens we will do two requests to the coin gecko endpoint

[
Branch of core repo for reference](https://github.com/MetaMask/core/tree/mobile-v-7.12.1?rgh-link-date=2023-12-04T20%3A07%3A52Z)
## **Related issues**

Fixes: #

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Screenshots/Recordings**
https://recordit.co/bu1PUlm68m


patches/@metamask+assets-controllers+7.0.0.patch
```diff
@@ -492,180 +492,440 @@ index 4ed4990..df0524b 100644
...
+      * Fetches a pairs of token address and native currency.
+      *
 diff --git a/node_modules/@metamask/assets-controllers/dist/TokenRatesController.js b/node_modules/@metamask/assets-controllers/dist/TokenRatesController.js
-index e3f81e9..b8e07f6 100644
+index e3f81e9..10994ef 100644
 --- a/node_modules/@metamask/assets-controllers/dist/TokenRatesController.js
 +++ b/node_modules/@metamask/assets-controllers/dist/TokenRatesController.js
-@@ -1,4 +1,11 @@
- "use strict";
-+// PATCH NOTES:
```
Gudahtt, line 525: Are these changes tracked on any branch in the core repo? It'd be nice to preserve some notes about what the patch is doing as well

tommasini: Let me create it! I didn't do it for this version :(

tommasini: `mobile-v-7.12.1` [created](https://github.com/MetaMask/core/tree/mobile-v-7.12.1)

Gudahtt: Thanks!

For other reviewers, this compare URL can be used to see the changes: https://github.com/MetaMask/core/compare/09a23c20d5209f7f037284c655a4ec5e1b4f693f...mobile-v-7.12.1

CHANGELOG.md
```diff
@@ -1,6 +1,170 @@
 # Changelog
 
 ## Current Main Branch
+- [#7939](https://github.com/MetaMask/metamask-mobile/pull/7939): chore: build bump 1214 RC4
```
Gudahtt, line 4: Are these changes intentional? Seems like they may have been a mistake

android/app/build.gradle
```diff
@@ -137,8 +137,8 @@ android {
         applicationId "io.metamask"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 1214
-        versionName "7.12.0"
+        versionCode 1220
```
Gudahtt, line 140: Is this change related to the token rates controller patch?

patches/@metamask+assets-controllers+7.0.0.patch
```diff
@@ -492,180 +492,440 @@ index 4ed4990..df0524b 100644
...
++    if (typeof state === "function" ? receiver !== state || !f : !state.has(receiver)) throw new TypeError("Cannot write private member to an object whose class did not declare it");
++    return (kind === "a" ? f.call(receiver, value) : f ? f.value = value : state.set(receiver, value)), value;
++};
++var _TokenRatesController_instances, _TokenRatesController_pollState, _TokenRatesController_updateTokenList, _TokenRatesController_stopPoll, _TokenRatesController_poll;
+ Object.defineProperty(exports, "__esModule", { value: true });
+ exports.TokenRatesController = void 0;
+ const base_controller_1 = require("@metamask/base-controller");
+ const controller_utils_1 = require("@metamask/controller-utils");
+ const crypto_compare_1 = require("./crypto-compare");
++const fast_deep_equal_1 = __importDefault(require("fast-deep-equal"));
```
Gudahtt, line 625: I see that this is in the dependency tree already because other packages depend on it, but it's not a dependency of either metamask-mobile or assets-controllers. Maybe a good idea to add it to metamask-mobile as an explicit dependency so that we can be certain it will be present.

We can add a note next to the exclusion in the depcheck config to explain what it's for as well (i.e. that it's used by this patch)

patches/@metamask+assets-controllers+7.0.0.patch
```diff
@@ -492,180 +492,440 @@ index 4ed4990..df0524b 100644
...
++    stop() {
++        __classPrivateFieldGet(this, _TokenRatesController_instances, "m", _TokenRatesController_stopPoll).call(this);
++        __classPrivateFieldSet(this, _TokenRatesController_pollState, PollState.Inactive, "f");
+     }
+     /**
+      * Fetches a pairs of token address and native currency.
+@@ -148,11 +172,34 @@ class TokenRatesController extends base_controller_1.BaseController {
+      * @returns The exchange rates for the given pairs.
+      */
+     fetchExchangeRate(chainSlug, vsCurrency) {
```
Gudahtt, line 789: It looks like this was async, but the async keyword has been removed. Thoughts on preserving that? I know that you removed the `await` usages, but it's still returning a Promise so it still should be marked as async.

Gudahtt: Actually... no, it's not returning a Promise. It's returning an array of Promises. I don't think this is quite right.

It looks like this will make a fetch call for each batch without waiting for the previous one to resolve, but that means that as the number of tokens increases we'll start hitting the per-IP rate limit immediately.

We should be requesting the batches in sequence, one at a time. And then if we're resolving the Promises in this function, we may as well return a single Promise with the list of responses (rather than an array of Promises, each of which resolves to a response).

tommasini: Addressed and using a `for of` now! Thanks for catching this!

patches/@metamask+assets-controllers+7.0.0.patch
tommasini: Added! Thanks Mark! Yes definitely we should have it on our package json for we to be sure we have it present

tommasini: I removed because it was failing on depcheck, should we workaround that check and add it anyway?

sethkfman:
This issue was address by the [hot fix](https://github.com/MetaMask/metamask-mobile/pull/7997). Additional work is being tracked to address this with the PriceAPI in this ticket #7903

PR #7972, opened 2023-12-01
chore: Replace local code fence transform with `@metamask/build-utils`

rekmarks:
## **Description**

In https://github.com/MetaMask/metamask-mobile/pull/7734, code fence support was added to MetaMask Mobile by essentially copypasting the implementation from the extension. This PR replaces the local code fence implementation with a version that was extracted into its own package.

The external package implements code fence parsing and removal, as well as linting transformed files. The consumer is responsible for integrating the fence parser into the given build system, and configuring the ESLint instance used for linting. Consequently, only the Metro transform and the ESLint instance initialization remain in this repository.

## **Related issues**

- https://github.com/MetaMask/core/pull/3577
- https://github.com/MetaMask/metamask-extension/pull/22033

## **Manual testing steps**

1. Check that different build types are built correctly

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #8012, opened 2023-12-06
fix: Rename old code fences

rekmarks:
## **Description**

Code fences were recently broken due to the merging of #7972. This PR fixes them.

## **Related issues**

- #7972 

## **Manual testing steps**

1. Observe that CI passes for this PR but not on `main`
2. Confirm that all code fences are named `ONLY_INCLUDE_IF` and not `ONLY_INCLUDE_IN`


owencraston (commented):
Can we change all of these to `///: BEGIN:ONLY_INCLUDE_IN(snaps)` It was an oversight on my end when I first added these but I really think they all belong under the flask flag and I was going to change this in a follow up PR anyway. The only one that should be flask should be the [aboutMetaMaskTitle](https://github.com/MetaMask/metamask-mobile/blob/be66c2a2d1ce40b72324eb18b18f158022cd1404/app/components/Views/Settings/index.tsx#L181-L184) instance. Other than that it looks good and I will approve after you push.

owencraston (approved):
Never mind, since main is broken lets merge this and rename it later.

PR #8080, opened 2023-12-12
feat(ramp): add sell quick amounts with gas estimations

wachunei:
## **Description**

This PR adds quick amounts buttons as percentages to the Sell Build Quote Keyboard. The percentages and max button will calculate a new amount Amount and  always be set to the minimum of (Amount, Balance - Gas Fees) if the selected asset is the native asset, enforcing the Amount never to be over the required gas fees to be paid.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to Sell flow
2. Select a native asset (eg: Ethereum in mainnet)
3. Open the keyboard and test the buttons

### **After**

https://github.com/MetaMask/metamask-mobile/assets/1024246/9ae959de-1e49-4ceb-8963-bb42da839f3b


wachunei:
@SocketSecurity ignore @consensys/on-ramp-sdk@1.25.6

package.json
```diff
@@ -166,7 +166,7 @@
   },
   "dependencies": {
     "@consensys/ledgerhq-metamask-keyring": "^0.0.5",
-    "@consensys/on-ramp-sdk": "1.25.4",
+    "@consensys/on-ramp-sdk": "1.25.6",
```
NicolasMassart, line 169: Q: does this feature really require onramp upgrade?

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    engine: {
+      backgroundState: {
+        GasFeeController: GasFeeState,
+      },
+    },
+  });
+}
+
+describe('useGasPriceEstimation', () => {
+  it('should return null if gasLimit is 0', async () => {
```
NicolasMassart, line 45: Q: when and why gaslimit could be zero? Could you add some comments?

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
```diff
@@ -0,0 +1,299 @@
...
+  });
+
+  it('should default to 21000 as gasLimit', async () => {
+    const { result } = renderHookWithProvider(
+      () => useGasPriceEstimation({ estimateRange: 'low' }),
+      {
+        state: setGasFeeControllerState({
+          gasEstimateType: 'eth_gasPrice',
+          gasFeeEstimates: {
+            gasPrice: '2',
```
NicolasMassart, line 114: C: try to use constants to name the values to make it clear. Here I'm not sure why 2, 20000 and 1e9. I can guess, but I should not have to.

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
```diff
@@ -0,0 +1,299 @@
...
+  describe('gasEstimateType is "fee-market"', () => {
+    it('should return gasPrice if gasEstimateType is "fee-market" without estimateRange', async () => {
+      const { result } = renderHookWithProvider(
+        () => useGasPriceEstimation({ gasLimit: 21000 }),
+        {
+          state: setGasFeeControllerState({
+            gasEstimateType: 'fee-market',
+            gasFeeEstimates: {
+              low: {
+                suggestedMaxFeePerGas: '1',
```
NicolasMassart, line 133: Q: why is this a number as string? If they are predetermined fixed values, should we have a constant list or an enum? Also having them named would be easier to understand.

app/components/UI/Ramp/common/components/QuickAmounts.tsx
```diff
@@ -48,6 +51,13 @@ const Amount: React.FC<AmountProps> = ({
       accessible
       {...props}
     >
+      {value === 1 && isNative ? (
```
NicolasMassart, line 54: Q: can you clarify/comment in the code what this value === 1 specific case is?

app/components/UI/Ramp/buy/Views/BuildQuote/BuildQuote.tsx
```diff
@@ -175,6 +179,11 @@ const BuildQuote = () => {
   const { limits, isAmountBelowMinimum, isAmountAboveMaximum, isAmountValid } =
     useLimits();
 
+  const gasPriceEstimation = useGasPriceEstimation({
+    gasLimit: isBuy ? 0 : 21000,
+    estimateRange: 'high',
```
NicolasMassart, line 184: C: should you use a constant or enum here instead of a hardcoded string?

app/components/UI/Ramp/buy/Views/BuildQuote/BuildQuote.tsx
```diff
@@ -175,6 +179,11 @@ const BuildQuote = () => {
   const { limits, isAmountBelowMinimum, isAmountAboveMaximum, isAmountValid } =
     useLimits();
 
+  const gasPriceEstimation = useGasPriceEstimation({
+    gasLimit: isBuy ? 0 : 21000,
```
NicolasMassart, line 183: C: could you make the values less magic? Here it's probably a basic value transfer gas, so you could do create a constant to make it easier to read.
Consider the similar comment for other values and other files, even 0, not sure why we have zero gas in some calls.

NicolasMassart (commented):
Nice feature, I really missed it and did not realise it could be added this way. Good job.

I added comments as I understood the overall code but still find it difficult to exactly get what cases we actually take into account, specifically related to gaslimit, because of the use of numbers for which I have to guess the origin instead of clearly being told what they are for.
I suggest you add some code comments to explain why and/or use explicit constant names to prevent the magic numbers that could not be obvious for other devs. Try to imagine asking yourself what these numbers would mean if you just landed on this code without any knowledge of the project or crypto.
Thanks!

package.json
wachunei: Yes, the amount type changes in the SDK methods so it can take both number and string types

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
wachunei: It is 0 when the Build Quote view is used by the Buy flow instead of the sell flow.

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
wachunei: They are just test numbers, do you have an example for these? basically if the number is 2, we expect the result to be `gasLimit * gasPrice`, both in wei, so `gasLimit * 2 * 1e9` (Gwei to wei)

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
wachunei: Because that's the type of the GasFeeController https://github.com/MetaMask/core/blob/main/packages/gas-fee-controller/src/GasFeeController.ts 

These are basic example values to use for the test.

app/components/UI/Ramp/common/components/QuickAmounts.tsx
wachunei: it is 100%, the "MAX" button.

app/components/UI/Ramp/buy/Views/BuildQuote/BuildQuote.tsx
wachunei: it is typed string, it is already a constant, if you place anything different than the typed values it will error. There's no benefit at all to store it in a constant and use it.

wachunei:
@NicolasMassart Thank you for your review. I have addressed some of the comments already in the latest commit.

bkirb:
@wachunei LGTM I finished verifying this change ✅

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
NicolasMassart: I mean, 21000 is not just an example, it's a value that reflects something. to me it's the base gas amount for a value tx. But here it's just dropped without explanation. So having it as a constant would help understanding the code.
Just an example:
```ts
  /** Some comment to explain why you chose 21000 for this test
  */
  it('should default to 21000 as gasLimit', async () => {
    // comment to explain why 2 and not any other value
    const expectedGasFeeEstimate=2
    const defaultValueTxGasAmount=21000
    const gweiToWeiMultiplier=1e9
    const { result } = renderHookWithProvider(
      () => useGasPriceEstimation({ estimateRange: 'low' }),
      {
        state: setGasFeeControllerState({
          gasEstimateType: 'eth_gasPrice',
          gasFeeEstimates: {
            gasPrice: expectedGasFeeEstimate.toString(10),
          },
        }),
      },
    );
expect(result.current?.estimatedGasFee.toString(10)).toEqual(
      `${defaultValueTxGasAmount * expectedGasFeeEstimate * gweiToWeiMultiplier}`,
    );
```

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
NicolasMassart: can you comment this in the code directly please?

app/components/UI/Ramp/common/components/QuickAmounts.tsx
NicolasMassart: Could you comment to make it obvious or use a constnat named explicitely?

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
wachunei: It is there already in the jsdoc https://github.com/MetaMask/metamask-mobile/pull/8080/files#diff-b41da70a85d7b8cd6c8217e42fbdeb645d31a8d40f0a357a4e19db3b5e53b4b3R20-R28

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
wachunei: I see now what you mean! I will refactor the tests

app/components/UI/Ramp/common/components/QuickAmounts.tsx
wachunei: This is done already in the updated changes https://github.com/MetaMask/metamask-mobile/pull/8080/files#diff-65b6ab695baddc2f801f16425b1784c6306ee8e44f40b8afa22a209ce57555b6R55-R58

app/components/UI/Ramp/common/hooks/useGasPriceEstimation.test.ts
NicolasMassart, line 45:
```suggestion
  it('should return null if gasLimit is 0', async () => {
  //gasLimit is 0 when the Build Quote view is used by the Buy flow instead of the sell flow.
```

NicolasMassart (dismissed):
I approve this changes for merge.

However I believe we should carefully refactor the test to make sure it's very clear what all the values are and what the test case is.
Not doing it doesn't break the code but doing it would ,make the code even more easy to understand.
Please consider this suggestion before merging.

wachunei:
> I approve this changes for merge.
> 
> However I believe we should carefully refactor the test to make sure it's very clear what all the values are and what the test case is. Not doing it doesn't break the code but doing it would ,make the code even more easy to understand. Please consider this suggestion before merging.

Done in cf910739e7dbd2edd5f12b4d9a73ca71e7fe9adf, thank you very much for your review

NicolasMassart (commented):
Looks good to me.
Thank you so much, I love reading clear code like this!

NicolasMassart (approved):

PR #8202, opened 2024-01-03
fix: Isolate logic to expose js env variables sooner

tommasini:
## **Description**
When creating a new terminal and trying to run any of the building commands like `yarn setup` `yarn watch` `yarn start:android` and `yarn start:ios`

It would give an error that the environment variables are not loaded

The problem was that we were checking if the variables were loaded before they were actually loaded.

The solution is to load the variables as one of the first steps if they exist.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


owencraston:
This is a good solution but I am wondering if we should have better defaults for these env variables? I know that for METAMASK_BUILD_TYPE we can default to `main` and `METAMASK_ENVIRONMENT` should probably be local? Then in the build.sh we can simply just have check if these values exist and if not use these defaults?

scripts/build.sh
```diff
@@ -542,6 +540,7 @@ checkAuthToken() {
 checkParameters "$@"
 
 printTitle
+prebuild
```
owencraston, line 543: is this really needed on every run? perhaps we can extract the sourcing of env files into a new function and only call this to avoid the building of static assets on every start?

sethkfman (approved):
LGTM

PR #8328, opened 2024-01-18
fix: Transaction Details view shows inaccurate balance on Mobile

### LINKED ISSUE #7787: [Bug]: Transaction Details view shows inaccurate balance
### Describe the bug

A user has a discrepancy in the USDT token balance on his transaction details page and the main view in the MM app. 

After a swap, the user received a value according to Etherscan, but MM transaction details view shows a different value.

device info: Android
app version: 7.8.0

A similar issue was reported for PEPE. In this case it was confirmed that the main view was showing the correct balance.

### Expected behavior

Main view and token details view show the same and accurate balance for tokens

### Screenshots/Recordings

_No response_

### Steps to reproduce

TBD

_No response_

### Version

7.8.0

### Build type

Other (please specify exactly where you obtained this build in "Additional Context" section)

### Device

NA

### Operating system

Android

### Additional context

[Slack thread](https://consensys.slack.com/archives/C8RSKCNCD/p1698417235602149)

### Severity

Medium severity; unclear how many users are affected (if this is token specific)
Accurate value still displayed on main view

salimtb:
## **Description**
users have discrepancy in the USDT token balance on his tx insight page in the MM app, they are seeing 33.22 USDT but it should be 5.43 USDT ( see the pictures below ) 

**the problem is due to the fact that we use a hexadecimal value to create a big number object, whereas the function balanceToFiat expects a decimal value. to calculate the conversion fiat**

## **Related issues**

Fixes: #7787 

## **Manual testing steps**

1. Go to the main page of the app and click on your USDT token.
2. scroll down and click on "Received USDT"
3. check the amount section

### **Before**

https://drive.google.com/file/d/1766q37UBdndVqk0cNG8eVCBf-mCsCm2j/view?usp=sharing

### **After**

https://drive.google.com/file/d/1l_sC7CK67WbjQmtLpbFUqiqj4iie2Fqv/view?usp=sharing

### DIFF
--- a/app/components/UI/TransactionElement/utils.js
+++ b/app/components/UI/TransactionElement/utils.js
@@ -34,6 +34,7 @@ import { isSwapsNativeAsset } from '../Swaps/utils';
 import { toLowerCaseEquals } from '../../../util/general';
 import Engine from '../../../core/Engine';
 import { isEIP1559Transaction } from '@metamask/transaction-controller';
+import { convertHexToDecimal } from '@metamask/controller-utils';
 
 const { getSwapsContractAddress } = swapsUtils;
 
@@ -281,7 +282,7 @@ export function decodeIncomingTransfer(args) {
     selectedAddress,
   } = args;
 
-  const decimalAmount = parseInt(value, 16).toString();
+  const decimalAmount = convertHexToDecimal(value);
   const amount = toBN(decimalAmount);
   const token = { symbol, decimals, address: contractAddress };
 
--- a/app/components/UI/TransactionElement/utils.js
+++ b/app/components/UI/TransactionElement/utils.js
@@ -265,7 +265,7 @@ function getCollectibleTransfer(args) {
   return [transactionElement, transactionDetails];
 }
 
-function decodeIncomingTransfer(args) {
+export function decodeIncomingTransfer(args) {
   const {
     tx: {
       transaction: { to, from, value },
@@ -281,7 +281,8 @@ function decodeIncomingTransfer(args) {
     selectedAddress,
   } = args;
 
-  const amount = toBN(value);
+  const decimalAmount = parseInt(value, 16).toString();
+  const amount = toBN(decimalAmount);
   const token = { symbol, decimals, address: contractAddress };
 
   const renderTokenAmount = token
@@ -299,6 +300,7 @@ function decodeIncomingTransfer(args) {
       exchangeRate,
       currentCurrency,
     );
+
     renderTokenFiatNumber = balanceToFiatNumber(
       fromTokenMinimalUnit(amount, token.decimals) || 0,
       conversionRate,
--- a/app/components/UI/TransactionElement/utils.test.js
+++ b/app/components/UI/TransactionElement/utils.test.js
@@ -1,0 +1,57 @@
+import { decodeIncomingTransfer } from './utils';
+describe('decodeIncomingTransfer', () => {
+  it('should decode an incoming transfer', () => {
+    // Arrange
+    const args = {
+      tx: {
+        transaction: {
+          to: '0x77648f1407986479fb1fa5cc3597084b5dbdb057',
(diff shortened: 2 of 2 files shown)


app/components/UI/TransactionElement/utils.js
```diff
@@ -281,7 +281,8 @@ function decodeIncomingTransfer(args) {
     selectedAddress,
   } = args;
 
-  const amount = toBN(value);
+  const decimalAmount = parseInt(value, 16).toString();
+  const amount = toBN(decimalAmount);
```
salimtb, line 285: the problem is due to the fact that we use a hexadecimal value to create a big number object, whereas it expects a decimal value.

tommasini:
Is there an issue already created on github that we can link  for this bug?

salimtb:
> Is there an issue already created on github that we can link for this bug?

hey @tommasini ,
you can find here the [github](https://github.com/MetaMask/metamask-mobile/issues/7787) issue link

tommasini:
Thanks @salimtb , could we have some recordings with this working on some other networks? Like polygon, avalanche and bsc?

salimtb:
> Thanks @salimtb and thanks for looking into this! Could we have some recordings with this working on some other networks? Like polygon, avalanche and bsc?

i just added another screenshot on polygon on Polygon network

tommasini (dismissed):
LGTM!
Wondering if it would be cool to switch the conversion by this parseInt(value, 16).toString() by this [function](https://github.com/MetaMask/core/blob/235777c3feaf33a10ca8c01762e1c4d24adc478e/packages/controller-utils/src/util.ts#L456) it should work the same way! But with more safety I would say

app/components/UI/TransactionElement/utils.js
```diff
@@ -281,7 +281,8 @@ function decodeIncomingTransfer(args) {
     selectedAddress,
   } = args;
 
-  const amount = toBN(value);
+  const decimalAmount = parseInt(value, 16).toString();
```
tommasini, line 284: Do you think it would be cool to use this -https://github.com/MetaMask/core/blob/235777c3feaf33a10ca8c01762e1c4d24adc478e/packages/controller-utils/src/util.ts#L456

salimtb: nice catch , i just updated this line :)

app/components/UI/TransactionElement/utils.js
digiwand: I thought BN accepts hexadecimals as a string with a caveat 
> String values in hexadecimal literal form, e.g. '0xff' or '0xFF' (but not '0xfF'), are valid

ref: https://mikemcl.github.io/bignumber.js/

salimtb: hey @digiwand ,
sorry for the lack of precision, we use the bigNumber from the `ethereumjs-util` package.

digiwand: oh thanks for the explanation! I was confused since we utilize both packages in our repo. I see it imported now in the [controller-util](https://github.com/MetaMask/core/blob/235777c3feaf33a10ca8c01762e1c4d24adc478e/packages/controller-utils/src/util.ts#L456)

digiwand: I'm unable to locate where this version doesn't support hex strings during construction

salimtb: the `BN` object is always returning an hex value if you give an hex value as input, even if you specify the base ( 10 for example ) ,  i made this conclusion from some tests that i made.
let me try to find it in the doc.

salimtb: in the step after , to get the conversion to $ we're using `balanceToFiat` function , this one should take a `decimal` , but we was sending a hex because the `amount` that we got from `BN` was in `hex`

tommasini (approved):
lGTM

PR #1507, opened 2020-04-22
[Sentry] Bugfix/import account view

estebanmino:
**Description**

Fixes https://github.com/MetaMask/metamask-mobile/issues/1474

I think is safe to merge without QA

**Checklist**

* [ ] There is a related GitHub issue
* [ ] Tests are included if applicable
* [ ] Any added code is fully documented

**Issue**

Resolves #???

PR #8414, opened 2024-01-26
chore: add logs to identify root cause of issue reported in #1507

gantunesr:
## **Description**

This PR add logs help to collect more metrics in Sentry and provide more insights regarding the root cause of the issue described in https://github.com/MetaMask/mobile-planning/issues/1507.

## **Related issues**

Relates to https://github.com/MetaMask/mobile-planning/issues/1507

## **Manual testing steps**

There should not be any impact in the UI/UX. The goal of this PR is to detect possible edge cases in the app where the `selectedAddress` is not set.

## **Screenshots/Recordings**

There should not be any impact in the UI/UX


tommasini:
I believe yes, but I just to check with you!
When this error is thrown we are sure that it will be caught by the error boundary component?

Gudahtt:
I think we'll need to patch the keyring controller as well, since it calls `syncIdentities` with an empty array. That would result in this error being thrown.

gantunesr:
Thanks @Gudahtt!

I think it's safe to just comment out the line, I'll run some test on my end to verify it. Let me know what you think

gantunesr:
@tommasini I believe that too but I have not checked it yet. I'm still researching the issue and have not tested the current code of this PR.

Gudahtt:
>I think it's safe to just comment out the line, I'll run some test on my end to verify it. Let me know what you think

I think you're right, I'll try and verify this as best I can later though

Gudahtt:
@gantunesr I've taken another look, and I definitely think that `updateIdentities([])` call is not useful and safe to comment out.

Could you update this PR again when you get the chance, and I can review it again?

gantunesr:
Sure @Gudahtt, I just now returned to work on this code, but why is that? The next 2 lines updates the identities array to the account fetched from `eth-keyring-controller`

Gudahtt:
That method is replacing the existing vault with a new one, so it is supposed to replace all identities as well, as part of that process.

Not sure if that answers the question; could you clarify which part seems confusing if not?

gantunesr:
Right, I was under the impression that it would replace a repeated identity with a new one, but in reality it keeps the old one so it's right to clean `identities` before providing a new array of addresses

gantunesr:
@Gudahtt I decided to revert the patch because I was unsure of the impact. The `preferences-controller` patch would cause a crash under certain edge cases and I want to take a deeper look at the `keyring-controller` method. I have just kept this logs to monitor them in Sentry

owencraston (dismissed):

app/components/Views/ManualBackupStep1/index.js
```diff
@@ -81,6 +82,10 @@ const ManualBackupStep1 = ({ route, navigation, appTheme }) => {
             setView(CONFIRM_PASSWORD);
           }
         } catch (e) {
+          Logger.error(
+            'Error trying to recover SRP from keyring-controller',
+            e,
```
ccharly, line 87: Feel free to ignore this one:

Sorry for the noobie question, is it safe to log out an error here? This function seems quite sensitive. I've quickly checked (really quickly) and I think that most errors are "safe", but is it really future-proof though?

gantunesr: Good observation @ccharly, I can't say for sure if it's 100% safe, I can think of some usual cases that are ok but I could be missing some edge ones.

I don't think this needs to be future proof since it's not my goal to keep these logs in the long term, just for a couple of releases to get more data on the issue, but for the sake of "Better safe than sorry" I'll remove the error from the log.

gantunesr: Remove in [2215f51](https://github.com/MetaMask/metamask-mobile/pull/8414/commits/2215f51c400c8efabfb332114955b111df6b07ff)

montelaidev (approved):

PR #8471, opened 2024-01-30
feat: a hook to use Segment metrics

NicolasMassart:
## **Description**

- make MetaMetrics.getInstance synchronous and add an async init function + readiness check
- destructurable object hook return value to use metrics in components without having to use useRef and useEffect to set the instance
- interface for the destructurable object
- HOC (High Order Component) wrapper for class components
- new and updated unit tests

## **Related issues**

fixes https://github.com/MetaMask/mobile-planning/issues/1500

## **Manual testing steps**

none

### **Before**

NA

### **After**

NA

### DIFF
--- a/app/components/hooks/useMetrics/useMetrics.test.tsx
+++ b/app/components/hooks/useMetrics/useMetrics.test.tsx
@@ -1,0 +1,119 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import {
+  DataDeleteResponseStatus,
+  DataDeleteStatus,
+  IMetaMetricsEvent,
+} from '../../../core/Analytics';
+import MetaMetrics from '../../../core/Analytics/MetaMetrics';
+import useMetrics from './useMetrics';
+
+jest.mock('../../../core/Analytics/MetaMetrics');
+
+jest.mock('react-native/Libraries/Interaction/InteractionManager', () => ({
+  runAfterInteractions: (callback: () => Promise<void>) => callback(),
+}));
+
+const expectedDataDeletionTaskResponse = {
+  status: DataDeleteResponseStatus.ok,
+};
+
+const expectedDataDeleteStatus = {
+  deletionRequestDate: undefined,
+  dataDeletionRequestStatus: DataDeleteStatus.unknown,
+  hasCollectedDataSinceDeletionRequest: false,
+};
+
+const expectedDate = '20/04/2024';
+
+const expectedDataDeleteRegulationId = 'TWV0YU1hc2t1c2Vzbm9wb2ludCE';
+
+const mockMetrics = {
+  trackEvent: jest.fn(),
+  trackAnonymousEvent: jest.fn(),
+  enable: jest.fn(() => Promise.resolve()),
+  addTraitsToUser: jest.fn(() => Promise.resolve()),
+  group: jest.fn(),
+  reset: jest.fn(() => Promise.resolve()),
+  flush: jest.fn(() => Promise.resolve()),
+  createDataDeletionTask: jest.fn(() =>
+    Promise.resolve(expectedDataDeletionTaskResponse),
+  ),
+  checkDataDeleteStatus: jest.fn(() =>
+    Promise.resolve(expectedDataDeleteStatus),
+  ),
+  getDeleteRegulationCreationDate: jest.fn(() => expectedDate),
+  getDeleteRegulationId: jest.fn(() => expectedDataDeleteRegulationId),
+  isDataRecorded: jest.fn(() => true),
+  isEnabled: jest.fn(() => true),
+};
+
+(MetaMetrics.getInstance as jest.Mock).mockResolvedValue(mockMetrics);
+
+describe('useMetrics', () => {
+  it('tracks events', async () => {
+    const { result, waitForNextUpdate } = renderHook(() => useMetrics());
+
+    await waitForNextUpdate();
+
+    const event: IMetaMetricsEvent = {
+      category: 'test',
+    };
+
+    let deletionTaskId,
+      dataDeleteStatus,
+      deletionDate,
+      regulationId,
+      isDataRecorded,
+      isEnabled;
+
+    await act(async () => {
+      result.current.trackEvent(event);
+      result.current.trackAnonymousEvent(event);
+      await result.current.enable(true);
+      await result.current.addTraitsToUser({});
+      result.current.group('test', {});
+      await result.current.reset();
+      await result.current.flush();
+      deletionTaskId = await result.current.createDataDeletionTask();
+      dataDeleteStatus = await result.current.checkDataDeleteStatus();
+      deletionDate = result.current.getDeleteRegulationCreationDate();
+      regulationId = result.current.getDeleteRegulationId();
+      isDataRecorded = result.current.isDataRecorded();
+      isEnabled = result.current.isEnabled();
+    });
+
+    expect(mockMetrics.trackEvent).toHaveBeenCalledWith(
+      event.category,
+      {},
+      true,
+    );
+    expect(mockMetrics.trackAnonymousEvent).toHaveBeenCalledWith(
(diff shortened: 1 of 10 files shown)


app/components/hooks/useMetrics/useMetrics.ts
```diff
@@ -0,0 +1,65 @@
...
+function useMetrics() {
+  const [metrics, setMetrics] = useState<IMetaMetrics>();
+
+  useEffect(() => {
+    const setupMetaMetrics = async () => {
+      setMetrics(await MetaMetrics.getInstance());
+    };
+
+    setupMetaMetrics();
+  }, []);
```
wachunei, line 16: How often are we expecting this to be unset when `trackEvent`, `trackAnonymousEvent` are called? I'm thinking this might be a very rare edge case.

NicolasMassart: Do you mean we should add a readiness check?

app/components/hooks/useMetrics/useMetrics.ts
```diff
@@ -0,0 +1,318 @@
...
+ *   isDataRecorded,
+ *   isEnabled,
+ * } = useMetrics();
+ */
+function useMetrics() {
+  const [metrics, setMetrics] = useState<IMetaMetrics>();
+
+  useEffect(() => {
+    const setupMetaMetrics = async () => {
+      setMetrics(await MetaMetrics.getInstance());
```
Cal-L, line 45: @NicolasMassart You mentioned that there's a risk to the instance not being defined before events are recorded and potential events being dropped. To optimize for this, we can expose an init method on the MetaMetrics class, which we can then trigger at an earlier point in the app. This should solve the initial concern while making it flexible enough for us to use depending when metrics is enabled or not.

Cal-L (commented):
Left some comments

app/components/hooks/useMetrics/useMetrics.ts
NicolasMassart: I changed the way we use metametrics so that's not an issue anymore

app/components/hooks/useMetrics/useMetrics.ts
```diff
@@ -0,0 +1,121 @@
...
+    addTraitsToUser: MetaMetrics.getInstance().addTraitsToUser,
+    createDataDeletionTask: MetaMetrics.getInstance().createDataDeletionTask,
+    checkDataDeleteStatus: MetaMetrics.getInstance().checkDataDeleteStatus,
+    getDeleteRegulationCreationDate:
+      MetaMetrics.getInstance().getDeleteRegulationCreationDate,
+    getDeleteRegulationId: MetaMetrics.getInstance().getDeleteRegulationId,
+    isDataRecorded: MetaMetrics.getInstance().isDataRecorded,
+    isInitialized: MetaMetrics.getInstance().isInitialized,
+    isEnabled: MetaMetrics.getInstance().isEnabled,
+    init: MetaMetrics.getInstance().init,
```
wachunei, line 117: I see there must be cases that some code needs to reach the MetaMetrics instance, and doing it by using the hook across the codebase is the right way, but this seems like it is too coupled? I'm thinking maybe we just expose `trackEvent, trackAnonymousEvent, getInstance` instead of remapping each instance method to the hook? wdyt?

app/components/hooks/useMetrics/withMetricsAwareness.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ComponentClass } from 'react';
+import useMetrics from './useMetrics';
+import { IUseMetricsHook } from './useMetrics.types';
+
+const withMetricsAwareness =
+  (
+    Children: ComponentClass<{
+      metrics: IUseMetricsHook;
```
Cal-L, line 8: Let's make this a type as well so we can use it here and in other components when using withMetricsAwareness

Cal-L (commented):
Left some comments

app/components/hooks/useMetrics/useMetrics.ts
NicolasMassart: Great feedback!
Here's  what was my reasoning for choosing the mapping and not exposing the instance directly in the hook:
- I wanted to prevent the MetaMetrics class to embed component related code. So no effects or hooks or states or `InteractionManager.runAfterInteractions` for instance. Keeping it as pure as possible. Clean split between what we could call a controller and the more react native related side. 
- Hook is what makes the link with component context: adding lifecycle over metrics, for instance adding the `InteractionManager.runAfterInteractions` over trackEvent.
- If I expose the instance directly from the hook, in addition to wrapped function like trackEvent, it means devs will have the ability use the instance trackEvent. The later being dangerous as it's not taking the lifecycle specific behaviour.
- I wanted to expose only a subset of features that user will really use. Group, Reset and Alias for instance are not useful for the moment, so not exposing them seems the thing to do. They are usable through the MetaMetrics instance directly though. MetaMetrics class implements the full Segment SDK logic.
- The hook makes it far easier to test components and spy/mock the functions as we have some component tests that check the events are tracked. Which is a good thing we should do more.
- for advanced usage, we could directly use the MetaMetrics instance without going through the hook.
- hook greatly simplifies the event tracking calls migration, and given the amount of calls we have, and it can not really be done with search/replace, that will help speeding the process.

I know, probably it would be fine to directly forward instance. Maybe. This is just the thoughts I took onto account for my choice. Note that this is evolving.
I would be happy to update if that's really not ideal. In terms of maintainability, I believe this is not a big deal as the functions are tied to Segment SDK and I don't thing we will have new features often.
Anyway, even if it's the case, the mapping is still valid unless the signatures change. Code is tested and documented, so any change should be easy and not so impactful.

app/components/hooks/useMetrics/withMetricsAwareness.tsx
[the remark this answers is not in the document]
NicolasMassart: Could you clarify what you would want me to make a type with? Could you suggest a change using the ````suggestion` comment ?

app/core/Analytics/MetaMetrics.ts
```diff
@@ -462,19 +461,39 @@ class MetaMetrics implements IMetaMetrics {
...
+   * const metrics = MetaMetrics.getInstance();
+   * await metrics.init() && metrics.enable();
+   *
+   * @remarks Instance has to be initialized before being used
+   * Calling initialize multiple times will not reinitialize the instance
+   *
+   * @returns Promise indicating if MetaMetrics was initialized or not
+   */
+  init = async (): Promise<boolean> => {
+    if (this.#isInitialized) return true;
```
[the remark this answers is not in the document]
NicolasMassart: What do you mean? using Instance doesn't tell if it's initialized or not. Instance is defined as soon as you create it. But init may not have been async called. Could you show me a suggested change please?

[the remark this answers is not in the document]
NicolasMassart: Returning early is to prevent useless computation in case it's already initialised. Do you mean something different?

app/components/hooks/useMetrics/useMetrics.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ),
+  checkDataDeleteStatus: jest.fn(() =>
+    Promise.resolve(expectedDataDeleteStatus),
+  ),
+  getDeleteRegulationCreationDate: jest.fn(() => expectedDate),
+  getDeleteRegulationId: jest.fn(() => expectedDataDeleteRegulationId),
+  isDataRecorded: jest.fn(() => true),
+  isInitialized: jest.fn(() => false),
+  isEnabled: jest.fn(() => true),
+  init: jest.fn(() => Promise.resolve(false)),
```
Cal-L, line 47: Remove init and isInitialized from tests

Cal-L (approved):
LGTM

PR #8542, opened 2024-02-07
chore: simplify ccache cache logic

vinistevam:
## **Description**

The mix of the build logic and the fact that every push in the main triggers a pipeline in bitrise prevents the ccache from caching the main build files, and subsequent PRs get the benefit of the native dependencies previously cached in `main`.

This PR simplifies the approach by saving the native modules whenever something changes so every branch can benefit from the native dependencies cached.

- [Example pipeline presenting the issue](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/ac5b23e0-3481-4906-93b9-f76f58c12014)

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1514

## **Manual testing steps**

1. Run smoke and regression pipeline


OGPoyraz (dismissed):

bitrise.yml
```diff
@@ -625,7 +625,7 @@ workflows:
             - key: ipa-cache-{{ .CommitHash }}
       - script@1:
           title: Run detox build
-          run_if: '{{not (enveq "BITRISE_CACHE_HIT" "exact")}}'
+          run_if: '{{or (not (enveq "BITRISE_CACHE_HIT" "exact")) (enveq "CURRENT_BRANCH" "main")}}'
```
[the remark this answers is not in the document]
vinistevam: 1. yes, it was designed to do it and is not currently doing it because I've introduced `ipa-cache-{{ .CommitHash }}` which skips the build step in case you run the same commit twice.

2. I will fix it, previously the logic of it was in the `bitrise.yml` and I missed adding that into the script.

3. Indeed, this would simplify things. I can modify the logic and test it in the pipeline. Using this approach would take more time in the first run but after that, it would take advantage of the previous cache in case no native dependencies are changed in between. I think the first approach was meant to take advantage of the fact that native modules rarely change but you are right on isolating the cache by branch and not depending on what is happening in `main`.


Thanks for raising those questions, I will try to simplify taking what you suggested in point 3.

vinistevam:
Pipelines:
- Smoke [using ccache](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/0ea0476d-cc29-4eb9-bda3-487d4767c2ca)
- Smoke [without ccache](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/f044d6af-9255-45a1-8cc5-33138ad05157)

scripts/cache/set-cache-envs.sh
```diff
@@ -5,31 +5,11 @@ set_cache_envs() {
   local current_branch=$(git rev-parse --abbrev-ref HEAD)
```
matthewwalsh0, line 5: Do we need this variable now as we don't do anything with it other than echo it?

scripts/cache/set-cache-envs.sh
```diff
@@ -5,31 +5,11 @@ set_cache_envs() {
...
-    else
-      # Generate a checksum for the differences and set it as CCACHE_KEY
-      local checksum=$(echo "$differences" | sha512sum | awk '{print $1}')
-      envman add --key CCACHE_KEY --value "$prefix_ccache_key-$checksum"
-      echo "Checksum (CCACHE_KEY) set to $prefix_ccache_key-$checksum"
-    fi
-  fi
+  echo "Current branch: $current_branch"
+  # Generate a checksum for the native dependencies and set it as CCACHE_KEY
+  local checksum=$(find "$folder" -type f -exec sha512sum {} \; | sha512sum | awk '{print $1}')
```
matthewwalsh0, line 10: To clarify, is this generating checksums of each file in the folder, then generating a checksum of that list of checksums?

scripts/cache/set-cache-envs.sh
```diff
@@ -5,31 +5,11 @@ set_cache_envs() {
...
+  envman add --key CCACHE_KEY --value "$prefix_ccache_key-$checksum"
+  echo "Checksum (CCACHE_KEY) set to $prefix_ccache_key-$checksum"
```
matthewwalsh0, line 12: Would it be clearer to echo the checksum and the cache key separately rather than referring to it as the checksum cache key?

Or just to refer to it as the ccache key?

scripts/cache/setup-ccache.sh
```diff
@@ -1,7 +1,14 @@
...
+os_name=$(uname -s)
+echo "Operating system: $os_name"
 
+# Check if the OS is macOS
+if [ "$os_name" == "Darwin" ]; then
+  # Set up ccache configuration for iOS
+  ccache --config-path "$(pwd)/ccache.conf"
+fi
 # Set up ccache configuration
-ccache --config-path "$(pwd)/ccache.conf"
```
matthewwalsh0, line 4: Does this fail on other operating systems or cause any harm? Ideally the less branches the better.

bitrise.yml
```diff
@@ -582,6 +588,12 @@ workflows:
     after_run:
       - _notify_failure_on_slack
     steps:
+      - script@1:
+          title: Check native folder diff for iOS
```
matthewwalsh0, line 592: Maybe more explicit now to say `Generating ccache key using native folder checksum`? As there is no diff anymore.

bitrise.yml
```diff
@@ -615,16 +621,22 @@ workflows:
...
-          run_if: '{{not (enveq "SKIP_CCACHE_RESTORE" "true")}}'
           inputs:
             - key: '{{ getenv "CCACHE_KEY" }}'
+      - script@1:
+          title: Set skip ccache upload
+          run_if: '{{ enveq "BITRISE_CACHE_HIT" "exact" }}'
+          inputs:
+            - content: |-
+                #!/usr/bin/env bash
+                envman add --key SKIP_CCACHE_UPLOAD --value "true"
```
matthewwalsh0, line 632: Just to confirm, we now always use the CCache if it's available, even on `main` builds, but we only save it back to Bitrise if it's new as it should never change.

Should we update the PR description to clarify this if so?

scripts/cache/set-cache-envs.sh
vinistevam: Removed.

scripts/cache/set-cache-envs.sh
vinistevam: yeap, the command calculates the SHA-512 hash of all files in a specified folder and then calculates the SHA-512 hash of the concatenated hashes of those files.

scripts/cache/set-cache-envs.sh
vinistevam: You are right, I changed as you suggested to echo the checksum and Cache key separately.

scripts/cache/setup-ccache.sh
vinistevam: `--config-path` is not available in Linux but it won't cause any harm just log that this option does not exist. I'm removing it to simplify this PR. If necessary we can add in a separate PR.

bitrise.yml
vinistevam: Exactly, we only upload if something has changed in the native modules.
Updated PR description and title.

vinistevam:
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/7dee2ecc-59b5-4ed3-b81f-11a9670a1a57

matthewwalsh0 (approved):

OGPoyraz (approved):

PR #8546, opened 2024-02-07
Improve caching of node_modules and yarn cache dir on bitrise

dbrans:


Gudahtt:
What is the purpose of caching `node_modules`? I avoided that when implementing this caching step initially because it was susceptible to breaking in a number of different scenarios. For example, if the Node.js version changes and we had a dependency that was compiled for the previous Node.js version, it might break in CI but not locally because of the bad cache. Or it might break during the install script step (i.e. when `yarn allow-scripts` is run), if one of our dependencies has an install script that is only meant to be run once. And we can't skip running install scripts either, because they might have side-effects in directories outside of `node_modules`.

We could add the Node.js version to the cache key to eliminate that variable, but the install script problem, I didn't see a straightforward solution for. I don't think the time saved unzipping the packages in `.yarn/cache` is worth making the cache unreliable.

I realize that the `node_modules` cache wasn't introduced in this PR, but I thought it was worth bringing up again here since we're further entrenching that pattern here.

PR #8578, opened 2024-02-14
fix: Fix/1539 patch root file with controller data

Cal-L:
## **Description**

This PR creates a migration that reconstructs the old root persisted structure. In the migration, it reconsolidates the controllers back into the root data. This solution is intended to be compatible when upgrading from any version to 7.16.0, especially for users who have already upgraded to 7.15.0.

## **Related issues**

Fixes:

## **Manual testing steps**

- Install 7.14.0 release and upgrade to 7.16.0
- Install 7.15.0 release and upgrade to 7.16.0
- Upgrade to 7.15.0 from 7.14.0 on Android and notice the vault corruption flow. Recover the account, then upgrade to 7.16.0

The end behavior for all three scenarios is that the app should behave as expected and no data should be loss
For testing support, please reach out to @MarioAslau or @Cal-L 

### **After**

7.14.0 -> 7.16.0 is unaffected since the root architecture doesn't change and the migration just returns the original state

iOS
Going from 7.14 -> 7.15 -> 7.16. Same as going from 7.15 -> 7.16
https://github.com/MetaMask/metamask-mobile/assets/10508597/1a80c32f-cc57-4a5f-9e5b-fb7052595434

Going from 7.14 -> 7.16
https://github.com/MetaMask/metamask-mobile/assets/10508597/1e5b9c8e-19c1-43f3-a05b-c4b2af47357d

Android
Same as iOS

The video below shows that the app continues to function as expected when upgrading from 7.15.0 data -> 7.16.0 with the patch (Android)
https://github.com/MetaMask/metamask-mobile/assets/10508597/a98256f6-b855-45a4-acfe-3bac891b9e16


app/store/migrations/028.ts
```diff
@@ -0,0 +1,150 @@
...
+  },
+];
+
+/**
+ * Migrate back to use the old root architecture (Single root object)
+ *
+ * @param {unknown} state - Redux state
+ * @returns
+ */
+export default async function migrate(state: unknown) {
```
Gudahtt, line 62: This migration is async, but I thought our migration system did not support async migrations. I'm also concerned about setting the precedent of performing async migrations at this phase in general; it could negatively affect performance. These are not quick operations.

Have you considered moving this to the storage API instead? e.g. when we read from state, read both the old and new state structures, and use whichever is newer. We could also conditionally do this based on the migration version, so that it is only a one-time cost that users have to pay.

Cal-L: Redux persist types doesn't indicate async operations but digging into their source code, we can see that migrate can be async 
image. Agreed with the performance - however, it should only run once if the migration is needed and the app didn't show any significant slowdown while testing.

Regarding where to place this logic - we considered placing this in the `MigratedStorage` in `persistConfig.ts` but decided that it's cleaner to isolate it into a migration since it's performing a migration for a specific version (28). We didn't want any lingering code in the `MigratedStorage` that would have to be managed. The migration also happens before the `MigratedStorage` calls so we can better ensure that the data is populated in the persisted files before any gets and sets are called.

Cal-L: Would love to continue the convo on optimizing migrations/storage/and performance. Will work with Mario to create and share a doc around these topics ✌️

sethkfman:
Android failed to upgrade without an issue here: https://recordit.co/JGDL1vWBqE

sethkfman (approved):
LGTM

app/store/migrations/028.ts
Gudahtt: Ah OK, sounds good. I had only looked at the types, that explains why it looked wrong.

PR #8625, opened 2024-02-19
fix: ethQuery is not defined when refresh is called

tommasini:
## **Description**
We were experiencing `undefined is not an object (evaluating 'inputBn.toString')` this warning multiple times when opening the app, because `this.ethQuery` it was not defined yet at the `AccountTrackerController` when .

It was added a condition to only update the balance if it was possible to get balance from chain.

Core PR: https://github.com/MetaMask/core/pull/3933

## **Related issues**

Fixes:

## **Manual testing steps**

1. Open the app
2. Shouldn't have the warning `undefined is not an object (evaluating 'inputBn.toString')`
3.

### **After**

e

Smoke E2E test builds: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/a075edf7-676e-4d24-8b0d-014890863b6d 

Regression E2E test builds: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/30f56c4e-3811-432c-83f0-139ea5b16ff5 

### DIFF
--- a/patches/@metamask+assets-controllers+9.2.0.patch
+++ b/patches/@metamask+assets-controllers+9.2.0.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.d.ts b/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.d.ts
-index ad327da..56a1ae3 100644
+index ad327da..8df74ba 100644
 --- a/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.d.ts
 +++ b/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.d.ts
 @@ -1,6 +1,7 @@
@@ -47,7 +47,7 @@ index ad327da..56a1ae3 100644
      /**
       * Sets a new provider.
 diff --git a/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.js b/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.js
-index 18802c9..002f27c 100644
+index 18802c9..4b0d158 100644
 --- a/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.js
 +++ b/node_modules/@metamask/assets-controllers/dist/AccountTrackerController.js
 @@ -18,6 +18,7 @@ const async_mutex_1 = require("async-mutex");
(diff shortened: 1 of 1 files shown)


Gudahtt:
Good catch! This bug seems to be upstream as well. Could you fix this in core?

Gudahtt:
This does make me question how we're handling engine initialization as well though:

* Why are we attempting to get the balance before initialization? Perhaps we should show a loading indicator until after initialization has completed
* Initialization is happening in a disjointed manner; we're triggering it separately for each controller, and then some initialization steps are run at the end separately (`configureControllersOnNetworkChange` and `startPolling`). We should be initializing in one place, and awaiting any async steps so that we can know for certain when it has finished.
* The 500ms pause here is slowing down the response to network switching by 500ms. This is unnecessary now that the network controller has been updated. At one point it may have been a workaround do the fact that there was no event for when the provider has been updated, but we have that now (the `networkDidChange` event).
* Some of the steps in `configureControllersOnNetworkChange` seem questionable; I would audit those. Each controller should react to network changes on its own, we shouldn't need to manually configure anything. It's possible that some older controllers still need that, but I am doubtful.

tommasini:
> Why are we attempting to get the balance before initialization? Perhaps we should show a loading indicator until after initialization has completed
Definitely we should have a loading indicator! We can definitely seeing an zero balance for some time, and if the network is slower it will definitely seems like it's not getting the balance: https://recordit.co/BpyTp0wqjM

> The 500ms pause here is slowing down the response to network switching by 500ms. This is unnecessary now that the network controller has been updated. At one point it may have been a workaround do the fact that there was no event for when the provider has been updated, but we have that now (the networkDidChange event).

Need to debug further but after removing this timeout, it seems that the token rates are not updating on the UI when network changes or the tokens are being detected

> Some of the steps in configureControllersOnNetworkChange seem questionable; I would audit those. Each controller should react to network changes on its own, we shouldn't need to manually configure anything. It's possible that some older controllers still need that, but I am doubtful.

[These](https://github.com/MetaMask/metamask-mobile/blob/542af2c540a84851f12ec37851547f273a80d0db/app/core/Engine.ts#L1186) detect function should be inside a promise.all for performance reasons right? 
I need to understand why when we remove the timeout the networkDidChange is still not working for at least the tokens (only thing I tested), probably good to create a ticket for follow-up, wdyt?

Gudahtt:
Definitely, all of that stuff can be looked into as separate PRs/tickets, I just jotted them down here because I noticed it when looking at this PR.

tommasini:
Core PR: https://github.com/MetaMask/core/pull/3933

tommasini:
Forgot to paste here, issue created: https://github.com/MetaMask/mobile-planning/issues/1566

Cal-L (approved):
Lgtm

PR #8759, opened 2024-02-28
feat: Accounts controller integration

owencraston:
## **Description**

- In order to make way for account snaps/multi chain in the future, Metamask is migrating from address based accounts to account managed/created by the [accounts controller](https://github.com/MetaMask/core/tree/main/packages/accounts-controller). This account will be references through an account id and will hold information like the selected address, keyring type and account names (there is other information as well). This change will pave the way for holding snap accounts (which are available on extension already)
3. What is the improvement/solution?
- This change installs and sets up the appropriate actions/event listeners inside the Engine.ts file
- This change also runs a migration which derives the correct account controller information from the preferences controller. This migration runs on the first load to properly setup the data required in the accounts controller. See `032.ts` for more information.
- I created a helper method on the Engine class called `setSelectedAddress` that updates both the AccountsController and PreferencesController. this method ensures that the selected address is always up to date and in sync. There should be `NO MORE direct calls to PreferencesController.setSelectedAddress()` in the U.I. 
- This change DOES NOT interfere with the U.I. Changing the U.I to reference the accounts controller will take place in a subsequent change. 

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/305
Equivalent extension PR: https://github.com/MetaMask/metamask-extension/pull/21437

### Setup
1. checkout branch: `accounts-controller`
2. run `yarn setup && yarn watch:clean` 
3. in a new terminal run the app via `yarn start:ios` or inside xcode/android studio
4. optional: download [flipper](https://fbflipper.com/) debugger to make inspecting the logs easier
<details>
  <summary>5. add the following logs to app/components/Nav/App/index.js</summary>

```jsx
  // TODO remove this useEffect before merging
  useEffect(() => {
    // eslint-disable-next-line no-console
    console.log(
      'Accounts/ AccountController',
      JSON.stringify(Engine.context.AccountsController.state, null, 2),
    );
    // eslint-disable-next-line no-console
    console.log(
      'Accounts/ PreferencesController',
      JSON.stringify(Engine.context.PreferencesController, null, 2),
    );
  }, []);
```
</details>

### Testing
1. create a wallet
2. In your logs (terminal or flipper) search for `accounts`

- [x] 3. Notice that there are two logs, one for the accounts controller state and one for the preferences controller... your accounts controller log should have the default state

```
Accounts/ AccountController {
  "internalAccounts": {
    "accounts": {},
    "selectedAccount": ""
  }
}
```
and the preferences controller should have no selectedAddress or identities
```
Accounts/ PreferencesController {
  "defaultConfig": {},
  "defaultState": {
      "selectedAddress": "",
     "identities": {},
...
```
4. create a new account (you should have 2n accounts now)
6. swipe away the app to clear the state and reopen the app (this will trigger the logs again)
7. inspect the logs

- [ ] 8. you should have two accounts inside the AccountsControlller state that match the identities inside the preferences controller

```
Accounts/ AccountController {
  "internalAccounts": {
    "accounts": {
      "7acd47d4-9dd0-4740-842a-31d0ec10f63f": {
        "id": "7acd47d4-9dd0-4740-842a-31d0ec10f63f",
        "address": "0xbedd94abb29f6147481297f05152b18ca8b2633d",
        "options": {},
        "methods": [
          "personal_sign",
          "eth_sign",
          "eth_signTransaction",
          "eth_signTypedData_v1",
          "eth_signTypedData_v3",
          "eth_signTypedData_v4"
        ],
        "type": "eip155:eoa",
        "metadata": {
          "name": "Account 1",
          "keyring": {
            "type": "HD Key Tree"
          },
          "lastSelected": 1709793104305
        }
      },
      "4ca3eac2-9f08-43cd-a39d-3befee58f83b": {
        "id": "4ca3eac2-9f08-43cd-a39d-3befee58f83b",
        "address": "0xb8217d8111997fdaa0578da8f1d13692b90fc246",
        "options": {},
        "methods": [
          "personal_sign",
          "eth_sign",
          "eth_signTransaction",
          "eth_signTypedData_v1",
          "eth_signTypedData_v3",
          "eth_signTypedData_v4"
        ],
        "type": "eip155:eoa",
        "metadata": {
          "name": "Account 2",
          "keyring": {
            "type": "HD Key Tree"
          },
          "lastSelected": 1709793198158
        }
      }
    },
    "selectedAccount": "4ca3eac2-9f08-43cd-a39d-3befee58f83b"
  }
}
```

```
Accounts/ PreferencesController {
...
"internalState": {
    "featureFlags": {},
    "identities": {
      "0xbEDd94abB29F6147481297F05152B18Ca8b2633D": {
        "address": "0xbEDd94abB29F6147481297F05152B18Ca8b2633D",
        "name": "Account 1",
        "importTime": 1709793104318
      },
      "0xB8217d8111997fDAA0578Da8f1D13692B90Fc246": {
        "address": "0xB8217d8111997fDAA0578Da8f1D13692B90Fc246",
        "name": "Account 2",
        "importTime": 1709793198278
      }
    },
    "selectedAddress": "0xB8217d8111997fDAA0578Da8f1D13692B90Fc246",
...
```
- the `selectedAddress` inside the PreferencesController should match the address for the selectedAccount inside the AccountsController. Since the accounts controller uses account id's instead of the address as a key, you will need to cross reference the account id its properties such as account name and account address.

### Caveats
- Make sure to look at the `internalState` object within the PreferencesController as there are other weird state values that are out of sync and do not reflect the true state of the application.
- You may notice that the addresses inside the AccountsController are all lower cases where as the addresses inside the PreferencesController are checksummed. This is okay since in in the following change we will simply convert the addresses back to checksum in the U.I.

### **Before**

- N/A

### **After**

https://github.com/MetaMask/metamask-mobile/assets/22918444/1d8b5eb8-f82f-4b31-9f03-888e33ea07c5

## QA
I personally QA'd this build and tested the following flows. For these test I used [this QA build](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/9aff59d2-4970-4567-9745-563282487abc).

#### Import wallet on a fresh build
- After importing the wallet I added an account
- changed the account names to custom names
- [x]  closed the app and reopened it to ensure that the selected address was correct along with the custom name
- [x] connected account to [test dapp](https://metamask.github.io/test-dapp/)
- [x] was able to perform signs on the test dapp (personal sign, sign typed data v1-4)
- [x] disconnected account and reconnected
- [x] all addresses and account names were correct during these operations

Video:

https://github.com/MetaMask/metamask-mobile/assets/22918444/95a92721-234b-4e8d-846d-fc66e5b09428

#### Wallet Update Path (testing migration)
- imported a wallet on an older build (7.18) with funds
- added a new account with custom names
- added a QR hardware wallet to this account
- updated the app to my version using the account controller
- [x] after update all accounts were still present with correct names
- [x] funds were all present with correct balances
- [x] selected different account, reloaded the app and successfully found that the selected address was still correct
- [x] connected all 3 account to [test dapp](https://metamask.github.io/test-dapp/)
- [x] was able to perform signs on the test dapp (personal sign, sign typed data v1-4) using a QR hardware wallet 
- [x] disconnected all 3 accounts from the test dapp
- [x] during this time all account names and addresses were correct
- [x] changed my account name, reloaded the app and found that the new name was preserved
- [x] changed chain, the accounts remained correct

Video:

https://github.com/MetaMask/metamask-mobile/assets/22918444/cbccc9de-54ef-4a0e-961f-2ca505c97776

#### Dapp connection and send flow
- using the wallet from the above QA session (wallet containing 2 metamask accounts and 1 qr hardware wallet)
- I connected the QR hardware wallet to the [test dapp](https://metamask.github.io/test-dapp/)
- [x] was able to perform signs on the test dapp (personal sign, sign typed data v1-4) using a QR hardware wallet 
- disconnected the wallet
- reconnected the QR wallet to the test dapp
- went to perform a send 
- [x] the selected address was correctly my non QR account
- [x] went through the send flow with the address and names being correct
- I did not send since I did not have enough funds to cover gas costs

Video:

https://github.com/MetaMask/metamask-mobile/assets/22918444/6a11e340-61a4-498d-a2ad-f2acffd6c20b

#### Import account, remove account, re import account
- Using the wallet from the above steps (imported wallet with QR)
- deleted wallet
- re imported wallet
- [x] account was restored correctly
- [x] I was able to successfully connect and perform signs on two different accounts after reimporting

Video:

https://github.com/MetaMask/metamask-mobile/assets/22918444/dfbcbf64-9a57-4d92-bf2d-2815b9a4ec85

https://github.com/MetaMask/metamask-mobile/assets/22918444/e9798023-5d39-458f-a2bd-ff8e9f65c351


owencraston:
@SocketSecurity ignore npm/@metamask/obs-store@9.0.0

app/components/Nav/App/index.js
```diff
@@ -254,6 +254,18 @@ const App = ({ userLoggedIn }) => {
     }
   };
 
+  // TODO remove this useEffect before merging
```
owencraston, line 257: Used for testing but will remove before merging.

app/core/Engine.ts
```diff
@@ -225,21 +232,22 @@ type GlobalActions =
...
+  | KeyringControllerActions
+  | AccountsControllerActions;
 type GlobalEvents =
   | ApprovalControllerEvents
   | CurrencyRateStateChange
   | GasFeeStateChange
   | TokenListStateChange
   | NetworkControllerEvents
   | PermissionControllerEvents
-  | KeyringControllerEvents
```
owencraston, line 236: this was a duplicate as you can see from line 235 above.

app/core/Engine.ts
```diff
@@ -733,12 +798,45 @@ class Engine {
...
+
+            const internalAccountCount = internalAccounts.length;
+
+            const accountTrackerCount = Object.keys(
+              accountTrackerController.state.accounts || {},
+            ).length;
+
+            captureException(
+              new Error(
+                `Attempt to get permission specifications failed because their were ${accounts.length} accounts, but ${internalAccountCount} identities, and the ${keyringTypesWithMissingIdentities} keyrings included accounts with missing identities. Meanwhile, there are ${accountTrackerCount} accounts in the account tracker.`,
```
ccharly, line 836:
```suggestion
                `Attempt to get permission specifications failed because there were ${accounts.length} accounts, but ${internalAccountCount} identities, and the ${keyringTypesWithMissingIdentities} keyrings included accounts with missing identities. Meanwhile, there are ${accountTrackerCount} accounts in the account tracker.`,
```

app/core/Permissions/specifications.js
```diff
@@ -83,17 +83,23 @@ export const getCaveatSpecifications = ({ getIdentities }) => ({
...
  *   getAllAccounts: () => Promise<string[]>,
+ *   getInternalAccounts: () => import('@metamask/keyring-api').InternalAccount[],
+ *   captureKeyringTypesWithMissingIdentities: (internalAccounts?: import('@metamask/keyring-api').InternalAccount[], accounts?: string[]) => void,
  * }} options - Options bag.
  * @param options.getAllAccounts - A function that returns all Ethereum accounts
  * in the current MetaMask instance.
- * @param options.getIdentities - A function that returns the
- * `PreferencesController` identity objects for all Ethereum accounts in the
+ * @param options.getInternalAccounts - A function that returns the
+ * `AccountsController` internalAccount objects for all accounts in the
```
ccharly, line 92: Is it me or the doc line are a bit mixed up? Shouldn't it be "in the current Metamask instance" here?

app/core/Permissions/specifications.js
```diff
@@ -126,7 +132,43 @@ export const getPermissionSpecifications = ({ getAllAccounts }) => ({
 
     methodImplementation: async (_args) => {
       const accounts = await getAllAccounts();
-      return accounts;
+      const internalAccounts = getInternalAccounts();
+
+      return accounts.sort((firstAddress, secondAddress) => {
+        const firstAccount = internalAccounts.find(
+          (internalAccount) =>
+            internalAccount.address.toLowerCase() ===
```
ccharly, line 140: Same comment than earlier, maybe use `toChecksumAddress` here?

app/core/Permissions/specifications.js
```diff
@@ -126,7 +132,43 @@ export const getPermissionSpecifications = ({ getAllAccounts }) => ({
...
+            firstAddress.toLowerCase(),
+        );
+
+        const secondAccount = internalAccounts.find(
+          (internalAccount) =>
+            internalAccount.address.toLowerCase() ===
```
ccharly, line 146: Same.

app/core/Permissions/specifications.js
```diff
@@ -126,7 +132,43 @@ export const getPermissionSpecifications = ({ getAllAccounts }) => ({
 
     methodImplementation: async (_args) => {
       const accounts = await getAllAccounts();
-      return accounts;
+      const internalAccounts = getInternalAccounts();
+
+      return accounts.sort((firstAddress, secondAddress) => {
```
ccharly, line 137: Purely cosmetic small improvement, what about something like this (not tested):

```ts
const [firstAccount, secondAccount] = _.map([firstAddress, secondAddress], (address) => {
    const checksumedAddress = toChecksumAddress(address);
    const internalAccount = internalAccounts.find(
      (internalAccount) =>
        toChecksumAddress(internalAccount.address) === checksumedAddress
    );

    if (!internalAccount) {
        captureKeyringTypesWithMissingIdentities(internalAccounts, accounts);
        throw new Error(`Missing identity for address: "${address}".`);
    }

    return internalAccount;
});
```

This way, we only have conditions for the sort after we verified those "preconditions" (like having an internal account for a given address).

WDYT?

app/core/Permissions/specifications.js
```diff
@@ -169,7 +211,12 @@ function validateCaveatAccounts(accounts, getIdentities) {
       );
     }
 
-    if (!identities[address.toLowerCase()]) {
+    if (
+      !internalAccounts.some(
+        (internalAccount) =>
+          internalAccount.address.toLowerCase() === address.toLowerCase(),
```
ccharly, line 217: Same comment for `.toChecksumAddress` here.

app/core/Permissions/specifications.test.js
```diff
@@ -105,19 +140,45 @@ describe('PermissionController specifications', () => {
         });
 
         it('rejects addresses that have no corresponding identity', () => {
-          const getIdentities = jest.fn().mockImplementationOnce(() => ({
-            '0x1': true,
-            '0x3': true,
-          }));
+          const getInternalAccounts = jest.fn().mockImplementationOnce(() => [
```
ccharly, line 143: What about adding a:
```ts
const baseEoaAccount = {
  type: EthAccountType.Eoa,
  options: {},
  methods: [...Object.values(EthMethod)],
};
```

And use it for every test accounts we add with `...baseEoaAccount,`.

WDYT?

app/core/Permissions/specifications.js
ccharly: Also, after re-reading some other code, I think than using other names like `accountA` or `accountB`, might be more suitable here (since we're sorting "any" account, and not only the first and second.

app/store/migrations/033.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      backgroundState: {
+        PreferencesController: {
+          ...preferenceState,
+        },
+      },
+    },
+  };
+}
+
+describe('Migration #032', () => {
```
ccharly, line 90: Shouldn't it be `#33`? (Given the filename and looks like other migrations use `#XY` form, without a leading 0)

app/store/migrations/033.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      const uniqueAddressChar = address[2]; // Assuming 0x prefix is present, so actual third character is at index 2
+      const hashBase = '012345678987654';
+      return Promise.resolve(hashBase + uniqueAddressChar);
+    });
+  });
+
+  afterEach(() => {
+    sha256Spy.mockRestore();
+  });
+  describe('createDefaultAccountsController', () => {
```
ccharly, line 106: Having a newline here would be nice.

app/core/Permissions/specifications.js
ccharly: And finally. add a comment to explain in which order we sort the accounts. To my understanding, it's ordered by `.lastSelected` (highest `.lastSelected` get sorted in front, right?)

app/store/migrations/033.ts
```diff
@@ -0,0 +1,126 @@
...
+interface Identity {
+  name: string;
+  address: string;
+  lastSelected?: number;
+}
+
+export default async function migrate(state: unknown) {
+  if (!isObject(state)) {
+    captureException(
+      new Error(`Migration 32: Invalid root state: '${typeof state}'`),
```
ccharly, line 21: 32 -> 33 (everywhere in this file I guess)

app/store/migrations/033.ts
```diff
@@ -0,0 +1,126 @@
...
+  state: Record<string, any>,
+) {
+  const selectedAddress =
+    state.engine.backgroundState.PreferencesController?.selectedAddress;
+
+  const selectedAccount = Object.values<InternalAccount>(
+    state.engine.backgroundState.AccountsController.internalAccounts.accounts,
+  ).find(
+    (account: InternalAccount) =>
+      account.address.toLowerCase() === selectedAddress.toLowerCase(),
```
ccharly, line 117: `.toChecksumAddress`? :D

app/store/migrations/033.ts
```diff
@@ -0,0 +1,126 @@
...
+  ) as InternalAccount;
+
+  if (selectedAccount) {
+    state.engine.backgroundState.AccountsController.internalAccounts = {
```
ccharly, line 121: Couldn't we just use `....internalAccounts.selectedAccount = selectedAccount.id ?? '';` here?

app/store/migrations/033.test.ts
```diff
@@ -0,0 +1,281 @@
...
+      },
+    },
+  };
+}
+
+describe('Migration #032', () => {
+  let sha256Spy: jest.SpyInstance;
+
+  beforeEach(() => {
+    sha256Spy = jest.spyOn(Aes, 'sha256').mockImplementation((address) => {
```
ccharly, line 94: Isn't it redundant with what we have in the `testSetup.js` file?

patches/@metamask+accounts-controller+10.0.0.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/node_modules/@metamask/accounts-controller/dist/AccountsController.js b/node_modules/@metamask/accounts-controller/dist/AccountsController.js
+index 2d7cd3a..d01b558 100644
+--- a/node_modules/@metamask/accounts-controller/dist/AccountsController.js
++++ b/node_modules/@metamask/accounts-controller/dist/AccountsController.js
+@@ -60,7 +60,8 @@ class AccountsController extends base_controller_1.BaseController {
+             state: Object.assign(Object.assign({}, defaultState), state),
+         });
+         _AccountsController_instances.add(this);
+-        this.messagingSystem.subscribe('SnapController:stateChange', (snapStateState) => __classPrivateFieldGet(this, _AccountsController_instances, "m", _AccountsController_handleOnSnapStateChange).call(this, snapStateState));
++        // Snaps controller is not instanciated in the production MetaMask Mobile app (Flask only). For this reason we cannot subscriobe to this state change.
```
ccharly, line 10:
```suggestion
+        // Snaps controller is not instantiated in the production MetaMask Mobile app (Flask only). For this reason we cannot subscribe to this state change.
```

I'm not 100% sure for the `instantiated` here, I guess it's UK vs US form here? 🙈

But there was a typo for `subscriobe`!

montelaidev:
The accounts controller doesn't seem to update the selected account when changing accounts in the accounts picker

app/core/Engine.ts
```diff
@@ -484,15 +504,20 @@ class Engine {
       config: {
         provider: networkController.getProviderAndBlockTracker().provider,
         chainId: networkController.state.providerConfig.chainId,
+        selectedAddress:
+          initialState.AccountsController?.internalAccounts?.accounts[
```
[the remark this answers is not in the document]
owencraston: yeah good call. I'll add this in the latest commit.

app/core/Engine.ts
owencraston: good catch

app/core/Engine.ts
```diff
@@ -733,12 +798,45 @@ class Engine {
...
+            accountsController.listAccounts.bind(accountsController),
+          captureKeyringTypesWithMissingIdentities: (
+            internalAccounts = [],
+            accounts = [],
+          ) => {
+            const accountsMissingIdentities = accounts.filter(
+              (address) =>
+                !internalAccounts.some(
+                  (account) =>
+                    account.address.toLowerCase() === address.toLowerCase(),
```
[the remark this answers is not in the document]
owencraston: I wrote it this way to match the logic [inside extension](https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/metamask-controller.js#L1140). What is the benefit of computing the `.toLowerCase` outside of the `.some`? or are you suggesting that we use checkSum instead of lowercase?

app/core/Permissions/specifications.js
owencraston: 🤦 yes,I guess I lost my train of thought mid sentence. what else is new 🤣 . Fixed in the latest commit.

app/core/Permissions/specifications.test.js
owencraston: good idea, especially since this object gets used a lot.

app/store/migrations/033.test.ts
owencraston: ah yes, it used to be 32 but I had to change it in a rebase and forgot to modify this. good catch. ❤️

app/store/migrations/033.test.ts
owencraston: Guto and I paired on this and we were having issues getting the test setup to actually work so we added this. I went back and removed it just now and it seems like everything works without it so I will remove. I am not sure what changed but that is a nice win. I will remove it in the latest commit. cc @gantunesr

app/store/migrations/033.ts
owencraston: rebase issues 🤦

app/store/migrations/033.ts
owencraston: I am following the same pattern set out [in the extension](https://github.com/MetaMask/metamask-extension/blob/b3b5ef664c3b64a0cbc9cb7d418ea77930f52dcc/app/scripts/migrations/105.ts#L116). Perhaps we need to discuss the correct pattern as a group. cc @montelaidev

app/store/migrations/033.ts
owencraston: yes, addressed in the latest commit.

patches/@metamask+accounts-controller+10.0.0.patch
owencraston: Nope, leave it to a European to be better at english spelling than a Canadian.

app/core/Permissions/specifications.js
owencraston: I sent you a follow up discussion on slack.

app/store/migrations/033.ts
owencraston: fixed

app/core/Engine.ts
```diff
@@ -1663,4 +1780,8 @@ export default {
       logErrors?: boolean;
     } = {},
   ) => instance?.rejectPendingApproval(id, reason, opts),
+  setSelectedAddress: (address: string) => {
```
owencraston, line 1783: @montelaidev I added this method and replaced all direct calls to `PreferencesController.setSelectedAddress(address);` in the U.I to address the selected address being out of sync. It seems to working well now.

app/core/Permissions/specifications.test.js
```diff
@@ -4,15 +4,55 @@ import {
...
   unrestrictedMethods,
 } from './specifications';
+import { EthAccountType, EthMethod } from '@metamask/keyring-api';
+import { SnapCaveatType } from '@metamask/snaps-utils';
 
 describe('PermissionController specifications', () => {
+  const baseEoaAccount = {
+    type: EthAccountType.Eoa,
+    options: {},
+    methods: [...Object.values(EthMethod)],
```
montelaidev, line 14: We can't really use all the EthMethods now, because they would also contain 4337 methods. The extension would also have to be updated as well.

app/core/Engine.ts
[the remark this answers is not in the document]
ccharly: Having it outside makes it (in theory) a bit faster since we don't need t re-compute the checksum address for `address` during each iterations of the `some` (since `address` do not change here, but `account.address` does change though)

app/core/Permissions/specifications.test.js
owencraston: I updated it to explicitly state each eth method.

app/core/Engine.ts
```diff
@@ -484,15 +507,19 @@ class Engine {
       config: {
         provider: networkController.getProviderAndBlockTracker().provider,
         chainId: networkController.state.providerConfig.chainId,
+        selectedAddress:
+          internalAccounts?.accounts[internalAccounts?.selectedAccount]
+            ?.address ?? '',
```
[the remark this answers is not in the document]
owencraston: updated.

app/core/Permissions/specifications.js
owencraston: I computed the lowerCase value outside of the .some for the performance gain but have kept it as lowercase until I get some feedback from @montelaidev on this.

app/core/Engine.ts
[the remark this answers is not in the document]
owencraston: moved it outside of the .some but kept the values lowercase as we discussed on slack.

app/store/migrations/036.ts
```diff
@@ -0,0 +1,124 @@
...
+import { NativeModules } from 'react-native';
+const Aes = NativeModules.Aes;
+
+interface Identity {
+  name: string;
+  address: string;
+  lastSelected?: number;
+}
+
+export default async function migrate(state: unknown) {
```
tommasini, line 18: Right now we introduced async migrations, this state will be a promise, it will needed to be resolved for we have the values:
You can follow this [example](https://github.com/MetaMask/metamask-mobile/pull/8952/files#diff-4765cfe076490c6bc280062ec607c7ae60e9e0fb842e5aa75d6b8c24f0da9ea6R13)

app/store/migrations/033.ts
owencraston: Kept the values lowercase as per our discussion on slack.

app/core/Engine.ts
```diff
@@ -1529,6 +1628,18 @@ class Engine {
       }
     }
   }
+
+  // This should be used instead of directly calling PreferencesController.setSelectedAddress or AccountsController.setSelectedAccount
```
tommasini, line 1632: Are we going with this format further because we do not want to use the listeners on the controllers? 
I don't know if my question makes sense but, why PreferencesController or AccountsControler do not have listener that changes the internal state when one of them change? (And we use only one further on the mobile repo)

tommasini (changes requested):
Left a question and a comment <3

app/core/Engine.ts
owencraston: My reasoning here was that listeners have a lot more side effects than directly calling a function. In my eyes this is easier to maintain/reason about than parsing through a massive list of listeners. Also the extension uses the same pattern in conjunction with listeners so they are maintaining both patterns fairly well. this is a valid point though.

app/store/migrations/036.ts
owencraston: resolved in the [latest commit](https://github.com/MetaMask/metamask-mobile/pull/8759/commits/e499d81c40ef20ceaf4602cbcd142eef67bad0f4).

app/core/Engine.ts
tommasini: Thanks for the explanation! <3

app/store/migrations/036.ts
```diff
@@ -0,0 +1,125 @@
...
+    captureException(
+      new Error(`Migration 35: Invalid root state: '${typeof state}'`),
+    );
+    return state;
+  }
+
+  if (!isObject(state.engine)) {
+    captureException(
+      new Error(
+        `Migration 35: Invalid root engine state: '${typeof state.engine}'`,
```
tommasini, line 30: Can we update the migration number on the exception please?  And also if possible can we have this test cases on the migration test file as we have on older migrations?

tommasini (changes requested):
One last comment 🙏

app/store/migrations/036.ts
owencraston: 🤦 fixed

tommasini (dismissed):
COOOL, awesome work!! LGTM!

owencraston:
There are no U.I changes that can be QA'd here so I will advocate for this PR to not enter QA but when I eventually migrate all of the U.I to reference the accounts controller in a [subsequent change ](https://github.com/MetaMask/accounts-planning/issues/235)we can QA those. That being said I will leave this decision up to @plasmacorral and @gantunesr

owencraston:
[QA builds](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/b2e2cfc0-94f9-46f4-9517-ce2f1f5fb3da)

tommasini (dismissed):
LGTM

owencraston:
[New QA builds](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/11117a34-077a-42c4-ba95-0c69f3845c85)

app/store/migrations/037.ts
```diff
@@ -0,0 +1,147 @@
...
+        name: identity.name,
+        lastSelected: identity.lastSelected ?? undefined,
+        keyring: {
+          // This is default HD Key Tree type because the keyring is encrypted
+          // during migration, the type will get updated when the during the
+          // initial updateAccounts call.
+          type: 'HD Key Tree',
+        },
+      },
+      methods: [...Object.values(EthMethod)],
```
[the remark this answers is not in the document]
owencraston: true because this wont apply to 4437 accounts.... I set them to this now

```
    methods: [
      EthMethod.PersonalSign,
      EthMethod.Sign,
      EthMethod.SignTransaction,
      EthMethod.SignTypedDataV1,
      EthMethod.SignTypedDataV3,
      EthMethod.SignTypedDataV4,
      EthMethod.PrepareUserOperation,
      EthMethod.PatchUserOperation,
      EthMethod.SignUserOperation,
    ],
```

app/store/migrations/037.ts
```diff
@@ -0,0 +1,147 @@
...
+    );
+    const internalAccount = findInternalAccountByAddress(state, firstAddress);
+    if (internalAccount) {
+      state.engine.backgroundState.AccountsController.internalAccounts.selectedAccount =
+        internalAccount.id;
+      state.engine.backgroundState.PreferencesController.selectedAddress =
+        internalAccount.address;
+    }
+    return;
+  }
```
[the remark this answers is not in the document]
owencraston: We are handling this case above on this [line](https://github.com/MetaMask/metamask-mobile/pull/8759/files#diff-5ab5ef6bbca2dbda1ff1de6c748147b422c97b044b3c9d32a08ed048f0ef30d2R72) by simply creating the default state for the accounts controller. I added a test for this case in [this](https://github.com/MetaMask/metamask-mobile/pull/8759/commits/daadd0c96215fbd8fc35d250ca5cdea0aba8e455) commit.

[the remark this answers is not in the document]
owencraston: I also don't think we should set the selected address in the preferences controller to an empty string since it can have a lot of unintended side effects.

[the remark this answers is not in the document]
montelaidev: After discussion with @owencraston, the identities shouldn't be empty if there is a selectedAddress.

app/store/migrations/037.ts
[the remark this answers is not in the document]
montelaidev: we should exclude 
 ```EthMethod.PrepareUserOperation,
      EthMethod.PatchUserOperation,
      EthMethod.SignUserOperation,```
      
 These are the 4337 methods

montelaidev (dismissed):

Cal-L (approved):
LGTM

PR #8827, opened 2024-03-04
feat: preinstalled mobile snaps - [1 of 10]

Jonathansoufer:
## **Description**

Within the scope of [Notifications Project](https://github.com/MetaMask/mobile-planning/issues/1495), and due restrictions on both Apple and Google store, the way Snaps will be delivered to mobile users will be in a pre installed approach. Notifications will rely on [MM auth](https://github.com/consensys-vertical-apps/metamask-auth-sdk/tree/main) 
 as a snap. 

Since preinstalled snaps isn't a  feature available on mobile yet, this PR implements a way to enable MetaMask mobile to read from FS an embedded/preInstalled snap, so we can embed auth snap into Mobile App by default.

## **Related issues**

Fixes: [Jira Ticket](https://consensyssoftware.atlassian.net/browse/NOTIFY-467?atlOrigin=eyJpIjoiNjdiY2QwOGNmZjY3NGQwYjg3YzM3ZTg1NDI1N2RiZWQiLCJwIjoiaiJ9)

## **Manual testing steps**

1. Go to this Snaps page and check the existence of a installed snap in a fresh installation.

### **After**

Screenshot 2024-03-14 at 11 11 47


app/util/test/testSetup.js
```diff
@@ -20,6 +20,11 @@ jest.mock('react-native', () => {
   return originalModule;
 });
 
+jest.mock('../../lib/snaps/preinstalled-snaps', () =>
+  // eslint-disable-next-line no-console
+  console.log("do nothing since we aren't testing the pre installed snaps"),
```
NicolasMassart, line 25: Q: we are not testing the snap but could we test the getPreinstalledSnap function or the object? Making the function a util and then use it in app/lib/snaps/preinstalled-snaps.ts could help at least test the function implem.

owencraston:
I am not sure if we will be able to solve this issue here but when I navigate to the test snap site (https://metamask.github.io/snaps/test-snaps/1.1.0/) it does not show the bip32 snap as installed. I am wondering if this will cause issues in the future where the state of the browser and the local snaps are out of sync and we direct people to the browser to perform certain actions. thoughts?
Simulator Screenshot - iPhone 13 Pro - 2024-04-16 at 10 52 55

app/lib/snaps/preinstalled-snaps.ts
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable no-console */
```
owencraston, line 1: any ideas as to how we could make this more dynamic. If we add more pre bundled snaps in the future this code will have to change a lot and it will get very repetitive. I was able to come up with something


```ts
/* eslint-disable no-console */
/* eslint-disable @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports, import/no-commonjs*/
import { assert } from '@metamask/utils';
import getPreinstalledSnap from '../../util/snaps/getPreinstalledSnap';

// List of pre installed snaps
const snapModules = ['@metamask/bip32-example-snap'];

async function loadSnapAssets(snapName: string) {
  let bundleJs, manifestJson, icon;
  let bundlePath = 'dist/bundle.js';
  let iconPath = 'images/icon.svg';

  try {
    manifestJson = await import(
      `../../../node_modules/${snapName}/snap.manifest.json`
    );
    assert(
      manifestJson,
      new TypeError(`ManifestJson not found for ${snapName}`),
    );

    bundlePath = manifestJson.source.location.npm.filePath;
    bundleJs = await import(`../../../node_modules/${snapName}/${bundlePath}`);
    assert(bundleJs, new TypeError(`BundleJs not found for ${snapName}`));

    iconPath = manifestJson.source.location.npm.iconPath;
    try {
      icon = await import(`../../../node_modules/${snapName}/${iconPath}`);
    } catch (error) {
      console.warn(
        `Icon not found for ${snapName} at ${iconPath}, using default icon.`,
      );
      icon = ''; // what is the default?
    }
  } catch (error) {
    console.error(`Error loading assets for ${snapName}: ${error}`);
    throw error;
  }

  return {
    bundleJs,
    manifestJson,
    icon,
    bundlePath,
    iconPath,
  };
}

export async function loadPreinstalledSnaps() {
  const snapsPromises = snapModules.map(async (snapName) => {
    const { manifestJson, icon, bundleJs, iconPath, bundlePath } =
      await loadSnapAssets(snapName);
    return getPreinstalledSnap(snapName, manifestJson, [
      {
        path: iconPath,
        value: icon,
      },
      {
        path: bundlePath,
        value: JSON.stringify(bundleJs),
      },
    ]);
  });

  const PREINSTALLED_SNAPS = await Promise.all(snapsPromises);
  return PREINSTALLED_SNAPS;
}

const PREINSTALLED_SNAPS = await loadPreinstalledSnaps();

export default PREINSTALLED_SNAPS;

```

The issue is that this requires top level await which I do not think is currently supported on RN (we should verify this). Do you see a way we could implement something like what I have above? One option would be to change the `preinstalledSnaps?: PreinstalledSnap[];` to intake a generator function?

owencraston: I tried using dynamic require here but it did not work with variables.

Jonathansoufer:
> I am not sure if we will be able to solve this issue here but when I navigate to the test snap site (https://metamask.github.io/snaps/test-snaps/1.1.0/) it does not show the bip32 snap as installed. I am wondering if this will cause issues in the future where the state of the browser and the local snaps are out of sync and we direct people to the browser to perform certain actions. thoughts? Simulator Screenshot - iPhone 13 Pro - 2024-04-16 at 10 52 55

Good catch @owencraston , but I think that wouldn't be a problem since in the first call, that RPC would note the installation and proceed instantly as shown below.

https://github.com/MetaMask/metamask-mobile/assets/44679989/afa92a44-a832-4c3c-aadf-082bcd289610

app/lib/snaps/preinstalled-snaps.ts
Jonathansoufer: Indeed dynamic require/imports doesn't work as we want on RN. To achieve what you coded above, we would need to change `preinstalledSnaps` asap and I'm afraid that would impact Notifications' deadlines immensely. So, I came up with a **not so nice** approach but delivers the ability to work with more than just one embedded snap. We can revisit this theme after notifications and try to come up with a better solution as proposed by you.

owencraston (dismissed):

yarn.lock
```diff
@@ -4739,7 +4751,11 @@
   resolved "https://registry.yarnpkg.com/@noble/hashes/-/hashes-1.1.5.tgz#1a0377f3b9020efe2fae03290bd2a12140c95c11"
   integrity sha512-LTMZiiLc+V4v1Yi16TD6aX2gmtKszNye0pQgbaLqkvhIqP7nVsSaJsWloGQjJfJ8offaoP5GtX3yY5swbcJxxQ==
 
+<<<<<<< HEAD
```
legobeat, line 4754: Merge-conflict markers commited in lockfile. If you remove both versions and run `yarn install && yarn deduplicate`, it should do the right thing.

legobeat (dismissed):

app/lib/snaps/preinstalled/bip32/index.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports, import/no-commonjs*/
+import { assert } from '@metamask/utils';
+const bundleJs = require('../../../../../node_modules/@metamask/bip32-example-snap/dist/bundle.js');
```
legobeat, line 3: Is there a reason why these need relative path and can't resolve normally from package names? If so could we leave a comment as to why?
```suggestion
const bundleJs = require('@metamask/bip32-example-snap/dist/bundle.js');
```

(same on line below).

package.json
```diff
@@ -174,6 +175,7 @@
     "@metamask/transaction-controller": "8.0.1",
     "@metamask/utils": "^8.1.0",
     "@ngraveio/bc-ur": "^1.1.6",
+    "@noble/secp256k1": "^1.7.1",
```
legobeat, line 178: https://github.com/MetaMask/metamask-mobile/actions/runs/8908666381/job/24464750583?pr=8827

Jonathansoufer: Added by mistake. Removed.

app/lib/snaps/preinstalled/bip32/index.ts
Jonathansoufer: Fixed.

package.json
```diff
@@ -141,6 +141,7 @@
     "@metamask/approval-controller": "3.5.0",
     "@metamask/assets-controllers": "^18.0.0",
     "@metamask/base-controller": "^4.1.1",
+    "@metamask/bip32-example-snap": "^2.1.1",
```
FrederikBolding, line 144: Why are we installing the BIP32 example Snap?

app/lib/snaps/preinstalled-snaps.ts
```diff
@@ -0,0 +1,19 @@
+import getPreinstalledSnap from '../../util/snaps/getPreinstalledSnap';
+import bip32 from './preinstalled/bip32';
+
+const { name, manifestJson, icon, bundleJs } = bip32;
+
+const PREINSTALLED_SNAPS = Object([
```
FrederikBolding, line 6:
```suggestion
const PREINSTALLED_SNAPS = Object.freeze([
```

app/lib/snaps/preinstalled-snaps.ts
```diff
@@ -0,0 +1,19 @@
+import getPreinstalledSnap from '../../util/snaps/getPreinstalledSnap';
+import bip32 from './preinstalled/bip32';
+
+const { name, manifestJson, icon, bundleJs } = bip32;
```
FrederikBolding, line 4: This entire file can probably just mirror the extension now: https://github.com/MetaMask/metamask-extension/blob/2380a8df87d1dd4a62a9cfafd0bc00ddd77c11c4/app/scripts/snaps/preinstalled-snaps.ts#L1-L8

Jonathansoufer: Cool! I didn't know about the evolution on extention. Changed to mirro this implementation then,.

package.json
Jonathansoufer: This was the package we were using as example while we didn't have `@metamask/message-signing-snap`. Removed.

FrederikBolding (dismissed):

yarn.lock
Jonathansoufer: Fixed.

legobeat: @Jonathansoufer Still needs a `yarn deduplicate`: https://github.com/MetaMask/metamask-mobile/actions/runs/8976319546/job/24652990789?pr=8827

Jonathansoufer: just ran again.

FrederikBolding (approved):

owencraston (approved):
Looks good! Great work.

PR #8837, opened 2024-03-04
feat: Add RPC Method Tracking to Dapp JSON RPC Engine pipeline

jiexi:
## **Description**

Currently in Mobile, we do not directly track RPC Method calls in metrics. It would be great to have the same insights on RPC method calls that we currently do in Extension. This PR adds a simplified version of the [RPCMethodTrackingMiddleware found in Extension](https://github.com/MetaMask/metamask-extension/blob/0cd76e5b6a3fe4b491348ac0c348a27a82368a32/app/scripts/lib/createRPCMethodTrackingMiddleware.js#L129) to the Mobile Dapp JSON RPC Engine pipeline.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2169

## **Manual testing steps**

1. Use Flipper or some other software that monitors app console logs and network events
2. Start the iOS simulator build
3. Open a Dapp (such as the [test dapp](https://metamask.github.io/test-dapp/)) in the  in-app browser
4. Use the Dapp
5. Filter the app console logs for `TRACK event saved` AND/OR Check the network tab for `Inpage Provider`: `Provider Method Called` events that include the method name and origin
  - methods like `eth_sign` should not be rate limited
  - methods like `eth_requestAccounts` should be rate limited one tracked event every 5 minutes
  - most other methods take on the default behavior of being sampled randomly with a 0.1% chance, meaning it will be very difficult to get these ones to trigger events 
  - finally, there is a max 10 events per sliding 5 minute window that all events above are gated by

### **After**

https://github.com/MetaMask/metamask-mobile/assets/918701/4bef5012-894b-4112-b6a3-6b83f91a3a1b


app/util/middlewares/rpcMethodTracking.js
```diff
@@ -0,0 +1,132 @@
...
+    const shouldTrackEvent =
+      // Don't track if the rate limit has been hit
+      !isRateLimited &&
+      // Don't track if the global rate limit has been hit
+      !isGlobalRateLimited &&
+      // Don't track if the user isn't participating in metametrics
+      userParticipatingInMetaMetrics === true;
+
+    if (shouldTrackEvent) {
+      metrics.trackEvent(MetaMetricsEvents.PROVIDER_METHOD_CALLED, {
```
NicolasMassart, line 109: C: note that trackEvent, when used with the metrics instance directly (not the hook) is synchronous and on the same thread. So if you have any deed for it to be async, you have to encapsulate it into something like `InteractionManager.runAfterInteractions`

app/util/middlewares/rpcMethodTracking.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      const res = {
+        error: null,
+      };
+      const { executeMiddlewareStack, next } = getNext();
+      const handler = createHandler();
+      handler(req, res, next);
+      await executeMiddlewareStack();
+      expect(mockMetrics.trackEvent).not.toHaveBeenCalled();
+    });
+  });
```
NicolasMassart, line 76: C: this only tests the metametrics class is correctly applying its optin policy. You should not test that, it's already tested in metametrics tests.
Here you should base your test on the rate limit only, not the user metametrics optin.

app/util/middlewares/rpcMethodTracking.test.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  describe('MetaMetrics is enabled', () => {
```
NicolasMassart, line 78: C: same here, you should not have a describe based on metametrics enabled or not. This is out of your test scope.

app/util/middlewares/rpcMethodTracking.js
```diff
@@ -0,0 +1,132 @@
...
+    // Get the participateInMetaMetrics state to determine if we should track
+    // anything. This is extra redundancy because this value is checked in
+    // the metametrics controller's trackEvent method as well.
+    const userParticipatingInMetaMetrics = metrics.isEnabled();
+
+    const shouldTrackEvent =
+      // Don't track if the rate limit has been hit
+      !isRateLimited &&
+      // Don't track if the global rate limit has been hit
+      !isGlobalRateLimited &&
```
NicolasMassart, line 104: Q: what is the rate limit for exactly?
Is it for limiting the RPC requests, or only limiting the event tracking?
I'm not sure to understand, that's why I'm asking, but in case this is only for limiting tracking, there's better ways to do it.
You implemented this like it's on the extension, but on mobile we already use the official Segment SDK and this [SDK has flush policies and plugins](https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/) that you should use instead of adding some mechanism over it.
You can implement a specific policy for your use case, including method specific behaviours, and [attach it to the Segment client dynamically](https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/#adding-or-removing-policies). It would only require a new attach/detach function on metametrics to expose the sdk functions.

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -384,6 +385,10 @@ export class BackgroundBridge extends EventEmitter {
     // metadata
     engine.push(createOriginMiddleware({ origin }));
     engine.push(createLoggerMiddleware({ origin }));
+
+    // tracking
+    engine.push(createRPCMethodTrackingMiddleware({ metrics }));
```
NicolasMassart, line 390: Q: is there any reason you inject the metrics instance at this level instead of retrieving the instance directly in the middleware? It's a singleton, so whatever where you call it. As you don't seem to use it in app/core/BackgroundBridge/BackgroundBridge.js directly, and you mock it in tests anyway, I would advise to get the instance in app/util/middlewares/rpcMethodTracking.js directly

NicolasMassart (changes requested):
I have doubts about the whole thing as it looks to me that this is all to rate limit tracking events.
But there's a big difference with the extension here as mobile uses the official Segment SDK and not a home made one. So you already have flush policies and plugins that should be the way to deal with custom rate limitations.

PR #8844, opened 2024-03-05
feat: Add `hideSubmitButton` prop to templates

OGPoyraz:
## **Description**

This PR aims to add `hideSubmitButton` property to confirmation templates. Also hiding action buttons will be possible if `onSubmit` and `onCancel` is not provided. 

## **Related issues**

Fixes: https://github.com/orgs/MetaMask/projects/47/views/1?pane=issue&itemId=55391209

Similar task is resolved in extension https://github.com/MetaMask/metamask-extension/pull/23085

## **Manual testing steps**

To manually test this, either you should add `hideSubmitButton` to true in to your approval request (via code) 
Or temporarily set `hideSubmitButton` to true in your local at `app/components/Views/confirmations/components/Approval/TemplateConfirmation/Templates/index.ts` `values` constant. (See the recorded video by doing this)

### **After**

https://github.com/MetaMask/metamask-mobile/assets/7644512/ddad90e6-1e15-40a0-9e1c-887bbdc10ebc

### DIFF
--- a/app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
+++ b/app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
@@ -49,15 +49,6 @@ const TemplateConfirmation = ({
     [approvalRequest, onConfirm, onCancel, colors],
   );
 
-  const showActions =
-    !templatedValues.hideSubmitButton &&
-    Boolean(
-      templatedValues.onConfirm ||
-        onConfirm ||
-        templatedValues.onCancel ||
-        onCancel,
-    );
-
   useEffect(() => {
     // Handles the cancellation logic
     const handleOnCancel = () => {
@@ -70,16 +61,18 @@ const TemplateConfirmation = ({
     };
   }, [templatedValues.onCancel, onCancel, templatedValues]);
 
-  const buttons = [
-    {
+  const buttons = [];
+
+  if (!templatedValues.hideSubmitButton) {
+    buttons.push({
       variant: ButtonVariants.Primary,
       label: templatedValues.confirmText ?? strings('template_confirmation.ok'),
       size: ButtonSize.Lg,
       onPress: templatedValues.onConfirm ?? onConfirm,
-    },
-  ];
+    });
+  }
 
-  if (!templatedValues.onlyConfirmButton) {
+  if (!templatedValues.hideCancelButton) {
     buttons.push({
       variant: ButtonVariants.Secondary,
       label:
@@ -92,7 +85,7 @@ const TemplateConfirmation = ({
   return (
     <View style={styles.root}>
       <TemplateRenderer sections={templatedValues.content} />
-      {showActions && (
+      {buttons.length > 0 && (
         <View style={styles.actionContainer}>
           <BottomSheetFooter
             buttonsAlignment={ButtonsAlignment.Horizontal}
--- a/app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
+++ b/app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
@@ -49,6 +49,15 @@ const TemplateConfirmation = ({
     [approvalRequest, onConfirm, onCancel, colors],
   );
 
+  const showActions =
+    !templatedValues.hideSubmitButton &&
+    Boolean(
+      templatedValues.onConfirm ||
+        onConfirm ||
+        templatedValues.onCancel ||
+        onCancel,
+    );
+
   useEffect(() => {
     // Handles the cancellation logic
     const handleOnCancel = () => {
@@ -83,12 +92,14 @@ const TemplateConfirmation = ({
   return (
     <View style={styles.root}>
       <TemplateRenderer sections={templatedValues.content} />
-      <View style={styles.actionContainer}>
-        <BottomSheetFooter
-          buttonsAlignment={ButtonsAlignment.Horizontal}
-          buttonPropsArray={buttons}
-        />
-      </View>
+      {showActions && (
+        <View style={styles.actionContainer}>
+          <BottomSheetFooter
+            buttonsAlignment={ButtonsAlignment.Horizontal}
+            buttonPropsArray={buttons}
+          />
+        </View>
+      )}
     </View>
   );
 };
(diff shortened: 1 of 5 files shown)


app/components/Views/confirmations/components/Approval/TemplateConfirmation/Templates/index.ts
```diff
@@ -14,6 +14,7 @@ export interface ConfirmationTemplateValues {
   onConfirm?: (opts?: AcceptOptions) => void;
```
matthewwalsh0, line 14: Is it worth keeping these in alphabetical order now we have a few?

app/components/Views/confirmations/components/Approval/TemplateConfirmation/Templates/index.ts
```diff
@@ -14,6 +14,7 @@ export interface ConfirmationTemplateValues {
...
   onlyConfirmButton?: boolean;
```
matthewwalsh0, line 15: Since this does the same for cancel, is it worth renaming this to `hideCancelButton` for consistency?

Meaning both are visible by default but you can hide each one with an optional property.

We don't have many templates in mobile so should be easy to update any references.

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
```diff
@@ -49,6 +49,15 @@ const TemplateConfirmation = ({
     [approvalRequest, onConfirm, onCancel, colors],
   );
 
+  const showActions =
+    !templatedValues.hideSubmitButton &&
+    Boolean(
```
matthewwalsh0, line 54: The default `onConfirm` and `onCancel` are required properties meaning we can assume they are always present.

In which case, can we just set `showActions` by just checking `hideSubmitButton` or `hideCancelButton` is `false`?

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
```diff
@@ -49,6 +49,15 @@ const TemplateConfirmation = ({
     [approvalRequest, onConfirm, onCancel, colors],
   );
 
+  const showActions =
+    !templatedValues.hideSubmitButton &&
```
matthewwalsh0, line 53: We're currently only checking here in `showActions` which is responsible for hiding the button container, but do we need to also check when populating `buttons` below similar to `onlyConfirmButton`?

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
```diff
@@ -83,12 +92,14 @@ const TemplateConfirmation = ({
   return (
     <View style={styles.root}>
       <TemplateRenderer sections={templatedValues.content} />
-      <View style={styles.actionContainer}>
-        <BottomSheetFooter
-          buttonsAlignment={ButtonsAlignment.Horizontal}
-          buttonPropsArray={buttons}
-        />
-      </View>
+      {showActions && (
```
matthewwalsh0, line 95: As we already have snapshot tests, is it worth one for this scenario where no buttons are visible?

app/components/Views/confirmations/components/Approval/TemplateConfirmation/Templates/index.ts
OGPoyraz: Fixed in https://github.com/MetaMask/metamask-mobile/pull/8844/commits/b5fb540e43619ba9b1e24b511ede710f0d76f706

app/components/Views/confirmations/components/Approval/TemplateConfirmation/Templates/index.ts
OGPoyraz: Renamed in https://github.com/MetaMask/metamask-mobile/pull/8844/commits/b5fb540e43619ba9b1e24b511ede710f0d76f706

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
OGPoyraz: I decided to remove this constant and check `buttons` instead 
https://github.com/MetaMask/metamask-mobile/pull/8844/commits/b5fb540e43619ba9b1e24b511ede710f0d76f706

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
OGPoyraz: I agree fixed in https://github.com/MetaMask/metamask-mobile/pull/8844/commits/b5fb540e43619ba9b1e24b511ede710f0d76f706

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
OGPoyraz: Added tests in https://github.com/MetaMask/metamask-mobile/pull/8844/commits/4328973e1f268984b8fc6fd93cef5418eb362344

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
```diff
@@ -16,7 +16,10 @@ import { useAppThemeFromContext } from '../../../../../../util/theme';
 import { AcceptOptions, ApprovalRequest } from '@metamask/approval-controller';
 
 export interface TemplateConfirmationProps {
-  approvalRequest: ApprovalRequest<any>;
+  approvalRequest: ApprovalRequest<any> & {
+    hideCancelButton?: boolean;
```
matthewwalsh0, line 20: These two properties come from the template rather than the component properties so don't believe we need these.

app/components/Views/confirmations/components/Approval/TemplateConfirmation/TemplateConfirmation.tsx
```diff
@@ -61,16 +64,18 @@ const TemplateConfirmation = ({
...
       onPress: templatedValues.onConfirm ?? onConfirm,
-    },
-  ];
+    });
+  }
 
-  if (!templatedValues.onlyConfirmButton) {
+  if (!templatedValues.hideCancelButton) {
     buttons.push({
       variant: ButtonVariants.Secondary,
```
matthewwalsh0, line 80: Low priority now, and there won't be a use case with just a cancel button, but I wonder if we should change the variant to primary if it's the only button?

OGPoyraz: That make sense if only CTA is a cancel button here. But also not quite sure if it's against the UX rules since cancel action is always `Secondary` type in design system. 
Should we avoid adding an extra condition, given that there is no practical use for it?

vinistevam (approved):

matthewwalsh0 (approved):

PR #8862, opened 2024-03-06
Fix: Exclude `devDependencies` from `attribution.txt`

Gudahtt:
## **Description**

The generated file `attribution.txt` was intended to exclude all `devDependencies`, but a few snuck their way in anyway. The `build:attribution` script we use to generate that file has been updated to ensure that `devDependencies` are excluded.

The old script was deleting `devDependencies` from `package.json` in an attempt to exclude them, but this wasn't working properly because the package we use to find all licenses was looking on-disk for them, not walking through the dependency graph starting at our manifest.

The script now explicitly deletes `node_modules` and re-installs with only production dependencies, so they are the only ones present on-disk for the license checker to find.

The script now leaves the project in a state where only production dependencies are installed, so a reminder to run `yarn setup` was added to the end of the script

## **Related issues**

Fixes https://github.com/MetaMask/MetaMask-planning/issues/2182

## **Manual testing steps**

- Run `yarn build:attribution`
- See that `acorn-jsx` is not included (this is an example dependency of a `devDependency` that was included previously)
- Remember to re-run `yarn setup` after testing

## **Screenshots/Recordings**

N/A


Gudahtt:
@SocketSecurity ignore npm/@lavamoat/aa@3.1.5
@SocketSecurity ignore npm/@lavamoat/allow-scripts@2.5.1

These are our packages

Gudahtt:
@SocketSecurity ignore npm/@lavamoat/preinstall-always-fail@2.0.0

This is our package

Gudahtt:
@SocketSecurity ignore npm/cmd-shim@6.0.2
@SocketSecurity ignore npm/node-gyp@9.4.1
@SocketSecurity ignore npm/json-parse-even-better-errors@3.0.1

These new authors appear to be an npm bot and a maintainer

scripts/generate-attributions.sh
```diff
@@ -4,10 +4,47 @@ set -e
...
+  local tmp="package.json_temp"
+  jq 'del(.devDependencies)' package.json > "$tmp"
+  mv "$tmp" package.json
+
+  rm -rf "${PROJECT_DIRECTORY}/node_modules"
+  yarn --pure-lockfile
+
+  cd "${PROJECT_DIRECTORY}/scripts/generate-attributions"
+
+  # Install sub-project that just contains attribution generation package
```
Gudahtt, line 34:
```suggestion
  # Install sub-project that just contains attribution generation package
  # so that it can be used without installing `devDependencies`
```

scripts/generate-attributions.sh
```diff
@@ -4,10 +4,48 @@ set -e
...
+  git checkout package.json
+
+  # This condition is true when this script is run in an interactive shell
+  # See https://serverfault.com/a/146747
+  if [[ $- == *i* ]]; then
+    echo "Re-run yarn setup to restore development dependencies"
+  fi
+}
+
+main "${@}"
```
mcmire, line 51: I'm getting an "unbound variable" error when attempting to run this script. Changing this to the following seems to solve this problem:

``` suggestion
main "$@"
```

This may be a Bash 3 -> Bash 4 thing (macOS still uses Bash 3). 

If this script is intended to be run only in CI, then it doesn't matter, but I thought I'd note that.

mcmire (dismissed):
When I run the script on my computer I see different results for `attributions.txt`. It seems that `fsevents` gets added, `@parcel/watcher-linux-x64-glibc` gets removed, and `@parcel/watcher-linux-x64-musl` is replaced with `@parcel/watcher-darwin-arm64`. This makes sense to be as these changes are most likely due to a different platform (macOS). If this script is only intended to be run in CI, then that seems fine.

Regardless, I did not see any dev dependencies.

scripts/generate-attributions.sh
rickycodes: I don't think macOS even uses bash, does it? I think they switched to z shell. Running this locally on my machine worked as expected:

`
$ echo "$SHELL"
`
`
/bin/bash
`

looks like I am on bash 5

Gudahtt: Thanks for trying this out! I will test this and investigate why this is failing. Probably your suggestion makes sense but it's not obvious to me why.

Ricky that is a good point, but this script has Bash in the shebang so it should still be running in a Bash shell.

Gudahtt: I wasn't able to get to the bottom of this, but it seems to work fine as `"$@"` and it still passes ShellCheck, so it seems like a good solution to me!

Gudahtt:
>This makes sense to be as these changes are most likely due to a different platform (macOS). If this script is only intended to be run in CI, then that seems fine.

Exactly, this script will only be run on CI typically, so I think having platform differences in output is OK in this case

mcmire (approved):
LGTM!

Gudahtt:
CI errors appear to be unrelated to this PR and are non-blocking

PR #8952, opened 2024-03-15
feat: Network Controller update to v12.1.2

tommasini:
## **Description**

- Update network controller to version 12.1.2
- re-applied the patch
- created a migration to clean the storage of networkDetails and networkStatus properties of NetworkController that were deprecated
- removed ts-expect-error from engine from controllers that were expecting network controller v12
- update selectors that were dependents of the deprecated properties

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1555

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Screenshots/Recordings**

Add/remove/switch networks: https://recordit.co/LQvO3ZAtkQ

Before migrations and network controller update:
https://recordit.co/KwkE7QUjl6

After migrations and network controller update:
https://recordit.co/Bl1BkZVK1F
  

app/core/Engine.ts
```diff
@@ -546,10 +542,11 @@ class Engine {
       onNetworkStateChange: (listener) =>
         this.controllerMessenger.subscribe(
           AppConstants.NETWORK_STATE_CHANGE_EVENT,
+          //@ts-expect-error It's expected, it needs the gas fee controller to be updated a minor version
           listener,
         ),
       getCurrentNetworkEIP1559Compatibility: async () =>
-        await networkController.getEIP1559Compatibility(),
+        (await networkController.getEIP1559Compatibility()) ?? false,
```
Cal-L, line 549: Is there a reason why this would undefined?

e2e/fixtures/fixture-builder.js
```diff
@@ -135,15 +135,37 @@ class FixtureBuilder {
               ],
             },
             NetworkController: {
```
Cal-L, line 137: Is `networkId` missing from this state?

app/selectors/networkController.ts
```diff
@@ -63,7 +71,11 @@ export const selectNetworkConfigurations = createSelector(
 export const selectLegacyNetwork = createSelector(
```
Cal-L, line 71: Is this still considered deprecated?

app/util/test/initial-background-state.json
```diff
@@ -69,11 +69,11 @@
       "chainId": "0x1",
       "ticker": "ETH"
     },
```
Cal-L, line 71: Is `networkId` needed?

app/store/migrations/024.test.js
```diff
@@ -66,16 +66,20 @@ describe('Migration #24', () => {
 
     expect(newState.engine.backgroundState.NetworkController).toStrictEqual({
       networkConfigurations: {},
-      networkDetails: {
```
Cal-L, line 69: Is this test failing and are these changes needed? Seems like we should be leaving it alone especially if it's an old migration and should be testing using legacy data structure

app/store/migrations/036.ts
```diff
@@ -0,0 +1,124 @@
+import { captureException } from '@sentry/react-native';
+import { isObject, hasProperty } from '@metamask/utils';
+import { NetworkState, NetworkStatus } from '@metamask/network-controller';
+import { InfuraNetworkType } from '@metamask/controller-utils';
+
+/**
+ * This migration removes networkDetails and networkStatus property
+ * This migration add a new property `networkMetadata` to the NetworkController (Still under investigation if it's needed)
```
Cal-L, line 8: Could you elaborate on the investigation? I see `networksMetadata` used in the `Engine.ts` file.

app/store/migrations/036.ts
```diff
@@ -0,0 +1,124 @@
...
+  ) {
+    captureException(
+      new Error(
+        `Migration 36: Invalid NetworkController networkDetails state: '${typeof networkControllerState.networkDetails}'`,
+      ),
+    );
+    return state;
+  }
+
+  if (networkControllerState.networkDetails) {
```
Cal-L, line 65: The order of the logic starting from here until the end can probably be re-ordered in a way that it checks for pre-existing NetworkController fields first, then mutates/migrates to the new states. Otherwise it's possible to run into a scenario where the NetworkController has some new data as well as old data. In other words, we should try to ensure an absolute migration or fail all together without mutating the data.

Cal-L (commented):
Left some comments

app/store/migrations/036.ts
tommasini: Ah, it's needed, good catch, networksMetadata needs migration if not when the user updates the app and do not have values there it would crash on that line on Engine file

app/core/Engine.ts
tommasini: Yeah, if it can't determine the compatibility: https://github.com/MetaMask/core/blob/3a7c4743fc3e752ec3e8ad5fcbc1582074f83c02/packages/network-controller/src/NetworkController.ts#L970

app/selectors/networkController.ts
tommasini: I believe yes! Because it's very recent since we remove the property `network` from the state, it was introduced on 7.9.0! I think if we do not have any users on that version we can start to think to deprecate this!

app/store/migrations/024.test.js
tommasini: Hmmm, we can update the migration to not use our initialBackgroundState to keep this data. I can update that

app/util/test/initial-background-state.json
tommasini: It was already there, and it was a nice to have since it help us know that the property exists on the network controller state data I would say!
What do you think?

e2e/fixtures/fixture-builder.js
tommasini: Yes!! Will add it!

Cal-L (approved):
LGTM

PR #8991, opened 2024-03-19
fix: add Referer to Dapp Viewed events

NicolasMassart:
## **Description**

- add Referrer to Dapp Viewed events
- update unit test
- fix unit test names according to guidelines for consistency with new test

## **Related issues**

Fixes https://github.com/MetaMask/mobile-planning/issues/1601

## **Manual testing steps**

```Gherkin
Feature: Send Dapp Viewed with Referrer

  Scenario: Open dapp from in app browser
    Given the app is installed and setup
    And user goes into in app browser
    And user loads a dapp (must be a dapp you can connect with your wallet like https://portfolio.metamask.io/)
  
    When user connects dapp to Metamask
    And approves connection
  
    Then `Dapp Viewed` event is sent to Mixpanel
    And has `Referrer` property filled with the domain scheme+name, example https://portfolio.metamask.io/

  Scenario: Open dapp from external browser
    Given the app is installed and setup
    And user goes into external browser
    And user loads a dapp (must be a dapp you can connect with your wallet like https://portfolio.metamask.io/)
  
    When user connects dapp to Metamask using WalletConnect
    And approves connection
  
    Then `Dapp Viewed` event is sent to Mixpanel
    And has `Referrer` property filled with the domain scheme+name, example https://portfolio.metamask.io/
```

### **Before**

image

### **After**

image

### DIFF
--- a/app/util/metrics/trackDappViewedEvent/index.ts
+++ b/app/util/metrics/trackDappViewedEvent/index.ts
@@ -2,6 +2,7 @@ import { store } from '../../../store';
 import { selectIdentities } from '../../../selectors/preferencesController';
 import { addToViewedDapp } from '../../../actions/browser';
 import { MetaMetrics, MetaMetricsEvents } from '../../../core/Analytics';
+import { prefixUrlWithProtocol } from '../../browser';
 
 /**
  * Tracks Dapp viewed event
@@ -29,6 +30,7 @@ const trackDappViewedEvent = ({
   store.dispatch(addToViewedDapp(hostname));
 
   MetaMetrics.getInstance().trackEvent(MetaMetricsEvents.DAPP_VIEWED, {
+    Referrer: prefixUrlWithProtocol(hostname),
     is_first_visit: isFirstVisit,
     number_of_accounts: numberOfWalletAccounts,
     number_of_accounts_connected: numberOfConnectedAccounts,
--- a/app/util/metrics/trackDappViewedEvent/index.test.ts
+++ b/app/util/metrics/trackDappViewedEvent/index.test.ts
@@ -44,7 +44,7 @@ describe('trackDappViewedEvent', () => {
     jest.clearAllMocks();
   });
 
-  it('should track with isFirstVisit = true', () => {
+  it('tracks as first visit when dapp hostname not in history', () => {
     mockGetState.mockImplementation(() => ({
       browser: {
         visitedDappsByHostname: {},
@@ -59,6 +59,7 @@ describe('trackDappViewedEvent', () => {
     }));
 
     const expectedMetrics = {
+      Referrer: 'https://uniswap.org',
       is_first_visit: true,
       number_of_accounts: 2,
       number_of_accounts_connected: 1,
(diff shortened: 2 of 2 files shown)


MarioAslau (approved):
Looks great @NicolasMassart ! Thank you for the context

app/util/metrics/trackDappViewedEvent/index.ts
```diff
@@ -29,6 +30,7 @@ const trackDappViewedEvent = ({
   store.dispatch(addToViewedDapp(hostname));
 
   MetaMetrics.getInstance().trackEvent(MetaMetricsEvents.DAPP_VIEWED, {
+    Referrer: prefixUrlWithProtocol(hostname),
```
Cal-L, line 33: This property structure may have deviated a bit since the original discussion. We may want to update this to `referrer: {url: prefixUrlWithProtocol(hostname)}` instead. @gauthierpetetin Could you confirm this - https://github.com/MetaMask/mobile-planning/issues/1300#issuecomment-1893696980

Cal-L (commented):
Let's wait for @gauthierpetetin to confirm

app/util/metrics/trackDappViewedEvent/index.ts
NicolasMassart: This is the structure on Extension because it's part of the context data. But here it's an event property so adding structure is not going to make it detected by Mixpanel as if it was a context data. And given mobile is not only for browser (extension obviously is), having it as context doesn't make sense or it would have to be passed on all events and not only dapp viewing in browser.

Cal-L: Ah ok. Thanks for clarifying

Cal-L (approved):
LGTM

app/util/metrics/trackDappViewedEvent/index.ts
gauthierpetetin: In Extension, [here's](https://github.com/MetaMask/metamask-```
extension/blob/develop/app/scripts/background.js#L485-L487) how they emit the event:
    controller.metaMetricsController.trackEvent({
      event: MetaMetricsEventName.DappViewed,
      category: MetaMetricsEventCategory.InpageProvider,
      referrer: {
        url: origin,
      },
      properties: {
        is_first_visit: false,
        number_of_accounts: numberOfTotalAccounts,
        number_of_accounts_connected: numberOfConnectedAccounts,
      },
    });
```
which results in `Referrer` property appearing in MixPanel.

Our goal is to have the same property as Extension appearing in MixPanel.
However, what I understand from @NicolasMassart , is that it wouldn't work if we emitted the event with `referrer: {url: prefixUrlWithProtocol(hostname)}` structure and that we have to emit it as such instead: `Referrer: prefixUrlWithProtocol(hostname)`.

NicolasMassart: Clarifying the final decision and options we have in case we change our mind:

- The extension referrer is provided directly in the context and it makes sense because extension is on a web browser.
- Mixpanel transforms (makes pretty to read) this context object automatically.
- Mobile could add the referrer to context too (segment SDK plugin to write) but it would be sent to all events or require additional computation to each event tracked to prevent this data to be added when not relevant.
- a transform could be added to Mixpanel on the server side but requires to modify the server config.
- Adding the data directly in properties with the correct name provides the same result on Mixpanel, and is simpler.
- we can improve to any of these options in case data team require that we keep the exact same data structure.

gauthierpetetin: 👍🙏

Issue #9006, opened 2024-03-20
[Bug]: Switching networks when I'm in the browser doesn't change the active chain, I get: `Active chain is 0xa but received 0x1`


seaona:
### Describe the bug

**Problem:** whenever I am on a test dapp, if I switch networks and I try to perform a signature I get the error `Active chain is 0xa but received 0x1`.
It seems that this only happens if I change networks when I'm on the wallet browser, but it does not happen if I switch networks when I'm on the wallet view.

This was working before and the exact same signature is working for Extension doing the same flow of switching networks.

### Screenshots/Recordings

![Screenshot from 2024-03-20 12-07-33](https://github.com/MetaMask/metamask-mobile/assets/54408225/4246eb3b-f88c-4062-8f23-21cc681fbf16)

https://github.com/MetaMask/metamask-mobile/assets/54408225/c17a2c39-0b97-4554-8a95-7f10e703e44c

### Steps to reproduce

1. Select mainnet network
2. Go to the test-dapp
3. Connect the wallet
4. Click Malicious Permit --> see signature is triggered
5. Switch networks from the browser window
6. Click Malicious Permit --> see error

(This works normally if I switch networks from the wallet home view)

### Version

I've seen this on mainnet but not in 7.19

### Build type

None

### Device

Pixel 6

### Operating system

Android

PR #9021, opened 2024-03-21
fix: Network not updating when changing account connected the first time on a DAPP

tommasini:
## **Description**
Updating the conditional case on `BackgroundBridge` `onStateUpdate` function to trigger notifyChainChanged, when the chainId changes and when the network changes, it will allow for both properties being sent when they are called more than once and they are miss matching. That was happening on the first switch of the network and causing the bug.

This fixes this issue: https://github.com/MetaMask/metamask-mobile/issues/9006
## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Screenshots/Recordings**

https://recordit.co/FEycCcnZ92

### DIFF
--- a/app/selectors/networkController.ts
+++ b/app/selectors/networkController.ts
@@ -5,8 +5,6 @@ import {
   NetworkState,
   NetworkStatus,
 } from '@metamask/network-controller';
-import { getDecimalChainId } from '../util/networks';
-import { Hex } from '@metamask/utils';
 
 const selectNetworkControllerState = (state: RootState) =>
   state?.engine?.backgroundState?.NetworkController;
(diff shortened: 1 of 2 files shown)


app/selectors/networkController.ts
```diff
@@ -66,16 +68,18 @@ export const selectNetworkClientId = createSelector(
...
-    const { networkId, selectedNetworkClientId, networksMetadata } =
+  selectChainId,
+  (networkControllerState: NetworkState, chainId: Hex) => {
+    const { selectedNetworkClientId, networksMetadata } =
       networkControllerState;
+
     return networksMetadata?.[selectedNetworkClientId].status !==
       NetworkStatus.Available
       ? 'loading'
-      : networkId;
```
Gudahtt, line 79: Do you know why this was undefined? This is not the same value as `chainId`, this might end up breaking some dapps

tommasini: Thanks Mark! After zoom call, we found an answer (Thanks Mark again! ahah <3 )

Cal-L (commented):
Could you elaborate on how the network controller bump affected this logic?

tommasini:
Yes, the NetworkController changed the logic on how the state was being updated on version 12, and that seems to be impacting how the dapp receive data, what's happening is that the dapp is receiving data with the networkId and chainId on different times, because the networkId at certain point in time is null when the network changes (Because it's still being fetched) and because of that was not updating on the first change.


On NetworkController V13 (Next Major version) the core repo will not expose the networkId in the network state, so where we need this `networkId` data we will need to fetch it
This how extension does it: https://github.com/MetaMask/metamask-extension/blob/4604c23100adf406aea764e19fa3087f5263bfdd/app/scripts/metamask-controller.js#L2773

Cal-L (approved):
LGTM

PR #9065, opened 2024-03-26
feat: Update network controller to version 13.0.1

tommasini:
## **Description**

Updated network controller to version 13.0.1. Also created migration to delete networkId property from the network controller state
Added an inpageProvider property with a network id property to our internal property,
Created the reducer and the action for that property with createSlice of redux toolkit
Added  dispatch to the engine event subscriber `networkdStateChange` to update networkId 
Added dispatch to the Engine file with subscribing an event `networkWillChange` to update networkId to "loading"
Updated the patch of network controller
Updated test cases and initial states with new data structure with the networkId property

## **Related issues**

Fixes https://github.com/MetaMask/mobile-planning/issues/1624

## **Manual testing steps**
QA Build: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/4b1752b4-2490-4eee-b0f9-68a347562ccd

1. Go to this page...
2.
3.

## **Screenshots/Recordings**

https://github.com/MetaMask/metamask-mobile/assets/46944231/7d80d89e-8c4c-485e-8806-65e11419934e

Latest recording with new actions (while updating migration from main with account imported on optimism)

https://github.com/MetaMask/metamask-mobile/assets/46944231/d290680a-ae53-4f3e-b752-1ded27964954


omridan159 (dismissed):

andreahaku (dismissed):
LGTM

package.json
```diff
@@ -151,7 +151,7 @@
     "@metamask/keyring-api": "^4.0.0",
     "@metamask/keyring-controller": "^8.1.0",
     "@metamask/logging-controller": "^1.0.1",
-    "@metamask/network-controller": "12.2.0",
+    "@metamask/network-controller": "13.0.1",
```
Gudahtt, line 154: This is not something we need to change here, but in general I'd recommend using a `^` range instead of pinning dependencies, unless there is a specific reason we have for pinning on a specific version (I've had to do that in mobile for controllers in the past because of unintended breaking changes in non-major releases). Using a range of versions can assist with deduplicating dependencies, and can make it easier to bring in minor/patch updates.

Just commenting for future reference. Here it's fine, `^13.0.1` would resolve to `13.01` anyway and we're updating this again soon.

yarn.lock
```diff
@@ -4146,7 +4146,27 @@
...
+    "@metamask/swappable-obj-proxy" "^2.1.0"
+    "@metamask/utils" "^6.2.0"
+    async-mutex "^0.2.6"
+    eth-block-tracker "^7.0.1"
+    eth-rpc-errors "^4.0.2"
+    immer "^9.0.6"
+    json-rpc-engine "^6.1.0"
+    uuid "^8.3.2"
+
+"@metamask/network-controller@^12.0.0", "@metamask/network-controller@^12.1.2", "@metamask/network-controller@^12.2.0":
```
Gudahtt, line 4169: This update will introduce a peer dependency warning because various other controllers are still expecting the network controller to be v12, but instead they'll be communicating with v13.

However the only breaking change in v13 is the removal of the `networkId` state, and I've checked to confirm that none of these controllers are using that state. So I think we're good here.

app/util/networks/index.js
```diff
@@ -501,3 +507,42 @@ export const getBlockExplorerTxUrl = (
...
+    return `${value}`;
+  } else if (isStrictHexString(value)) {
+    return `${convertHexToDecimal(value)}`;
+  } else if (typeof value === 'string' && /^\d+$/u.test(value)) {
+    return value;
+  }
+  throw new Error(`Cannot parse as a valid network ID: '${value}'`);
+}
+
+export const deprecatedGetNetworkId = async () => {
```
Gudahtt, line 530: Nit: Would be great to have a JSDoc comment here as well, and unit tests

app/core/SDKConnect/AndroidSDK/AndroidService.ts
```diff
@@ -359,15 +359,13 @@ export default class AndroidService extends EventEmitter2 {
             NetworkController: NetworkController;
           }
         ).NetworkController;
-        const networkId = networkController.state.networkId ?? 1; // default to mainnet;
-        // transform networkId to 0x value
-        const hexChainId = `0x${networkId.toString(16)}`;
```
Gudahtt, line 364: Huh, this was never correct. Glad we're fixing this!

app/core/Engine.ts
```diff
@@ -1494,7 +1518,7 @@ class Engine {
       const fiatBalance = this.getTotalFiatAccountBalance() || 0;
       const totalFiatBalance = fiatBalance.ethFiat + fiatBalance.ethFiat;
 
-      return totalFiatBalance > 0 || tokenFound || nfts.length > 0;
+      return totalFiatBalance > 0 || tokenFound || nfts?.length > 0;
```
Gudahtt, line 1521: Hmm, what motivated this change? I wouldn't expect this `nfts` variable to be impacted by the network controller update

app/core/Engine.ts
```diff
@@ -1318,6 +1331,17 @@ class Engine {
             currentChainId = state.providerConfig.chainId;
           }, 500);
         }
+
+        try {
+          const networkId = await deprecatedGetNetworkId();
```
Gudahtt, line 1336: `deprecatedGetNetworkId` uses the provider, but the provider will not be updated until the `NETWORK_DID_CHANGE` event. It's still pointing at the previous network.  So this might end up incorrectly setting the ID, maybe only in some cases since this is async.

Have you tried using NETWORK_DID_CHANGE here? I know last time we talked about this, switching to that event ended up breaking something, but I can't recall the details.

Gudahtt:
Great work! Mostly looks good, I left some comments/suggestions but the only real concern is what I asked about in [this comment](https://github.com/MetaMask/metamask-mobile/pull/9065#discussion_r1541816835), as it may introduce a race condition where the `networkId` is sometimes incorrect.

app/core/Engine.ts
Gudahtt: Note the `setTimeout` a few lines above; I think that was used for this exact reason, to prevent the provider from being used before it's updated. That `setTimeout` was introduced before the `networkDidChange` event existed. The event is the more reliable way to know when the provider is ready.

package.json
tommasini: Will proceed, like this on the future! Yeah I think I was not pragmatic with that decision, I was taking too much in line the each controller version of the major core versions

tommasini: Thanks Mark!

yarn.lock
tommasini: Thanks for reviewing that Mark!

tommasini: I also added a note on the Engine file for the typescript expect that mismatching error, the next PRs will be to update that others controllers

christopherferreira9:
All good for SDK

app/core/SDKConnect/AndroidSDK/AndroidService.ts
tommasini: Yeahhhh

app/core/Engine.ts
tommasini: I believe it was because NETWORK_DID_CHANGE didn't follow the network id changes on the background bridge so it didn't update to the dapp

But I think here we can easily change for NETWORK_DID_CHANGE and that will not cause any difference on behaviour

app/core/Engine.ts
tommasini: I thought nfts could be undefined, but I just rechecked and it's not truth so it's not needed! Great catch!

app/core/Engine.ts
Gudahtt: Sounds good! We should be able to eliminate that `setTimeout` at the same time, which would eliminate a pre-existing race condition (rapid network switching would cause controllers to have the wrong provider set)

tommasini: Addressed it on this commit: https://github.com/MetaMask/metamask-mobile/pull/9065/commits/b9ae3c437b2724f843a9fa4786e1cefd430b945e

app/util/networks/index.js
tommasini: Added those on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/6210b092f695873d28cc2c9a31321c60bf7d5ca7)

Thanks Mark!

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -38,6 +36,19 @@ const EventEmitter = require('events').EventEmitter;
...
 import { getPermittedAccounts } from '../Permissions';
+import { NetworkStatus } from '@metamask/network-controller';
+
+const legacyNetworkId = () => {
+  const { networksMetadata, selectedNetworkClientId } =
+    store.getState().engine.backgroundState.NetworkController;
+
+  const { networkId } = store.getState().networkOnboarded;
+
+  return networksMetadata?.[selectedNetworkClientId].status !==
```
Gudahtt, line 47: We should also ensure that `networkId` is set here (i.e. if `networkId` is `undefined`, we want to return `'loading'` here rather than `undefined`.)

app/core/Engine.ts
```diff
@@ -1321,6 +1334,22 @@ class Engine {
       },
     );
 
+    this.controllerMessenger.subscribe(
```
Gudahtt, line 1337: Hmm, this may still leave us in an invalid state _shortly after_ a switch, or if we switch to a network that fails to respond correctly to `net_version`.

Perhaps we should clear the `networkId` state when a switch begins, so that it's not set to the wrong value during the switch, or after a failure. e.g.

```suggestion
    this.controllerMessenger.subscribe(
      'NetworkController:networkWillChange',
      () => {
        store.dispatch(setNetworkId(''));
      }
    );

    this.controllerMessenger.subscribe(
```

app/core/Engine.ts
```diff
@@ -1306,7 +1319,7 @@ class Engine {
 
     this.controllerMessenger.subscribe(
       AppConstants.NETWORK_STATE_CHANGE_EVENT,
-      (state: NetworkState) => {
+      async (state: NetworkState) => {
```
Gudahtt, line 1322: Nit: It seems that this change is unnecessary

```suggestion
      (state: NetworkState) => {
```

app/util/networks/index.js
```diff
@@ -501,3 +507,46 @@ export const getBlockExplorerTxUrl = (
...
+    return `${value}`;
+  } else if (isStrictHexString(value)) {
+    return `${convertHexToDecimal(value)}`;
+  } else if (typeof value === 'string' && /^\d+$/u.test(value)) {
+    return value;
+  }
+  throw new Error(`Cannot parse as a valid network ID: '${value}'`);
+}
+/**
+ * This function is only needed to get the networkId to support some legacy networks
```
Gudahtt, line 530: Nit: Here is my attempt to clarify the purpose of this; it's not for legacy networks, but for dapps that are using deprecated provider properties/events.

```suggestion
 * This function is only needed to get the `networkId` to support the deprecated `networkVersion` provider property and the deprecated `networkChanged` provider event.
 *
```

app/core/BackgroundBridge/BackgroundBridge.js
tommasini: Great catch! This way if the fetch of networkId fails we are always safe!

app/actions/onboardNetwork/index.ts
```diff
@@ -38,3 +38,8 @@ export const showNetworkOnboardingAction = ({
   nativeToken,
   showNetworkOnboarding,
 });
+
+export const setNetworkId = (networkId: string) => ({
```
Gudahtt, line 42: Nit: This action creator is not very Redux-y. Typically with Redux we want to model actions (and action creators) after _events_, rather than using them to set state. This tends to result in more easily maintained code, establishing clearer boundaries for what should live inside v.s. outside the reducer.

Here for example, the action is event-like ("NETWORK_ID_UPDATED"), but the action creator is named as a setter. We could instead name this `networkIdUpdated`, to align more with the action. See here for more on this guideline: https://redux.js.org/style-guide/#model-actions-as-events-not-setters

app/core/Engine.ts
Gudahtt: Though it may be more aligned with Redux best practices to create a new action + action creator for this case (e.g. `networkChanged` and `NETWORK_CHANGED`) rather than re-using `setNetworkId` this way.

app/reducers/networkSelector/index.ts
```diff
@@ -71,6 +73,11 @@ function networkOnboardReducer(
           ...state.networkOnboardedState,
         },
       };
+    case 'NETWORK_ID_UPDATED':
```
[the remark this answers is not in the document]
tommasini: Yeah I agree, that was my first confusion when I was choosing a place for that to live, and since the naming of the already reducers/actions related with networks were confused I went to the "easy" path of using one related to network already! But I will create one name like "legacyProvider" and will have the property networkId inside

app/actions/onboardNetwork/index.ts
tommasini: Thanks!!!! Will create a new action, reducer and improve it with these guidelines!

app/core/Engine.ts
tommasini: Ahhh yes!!! This will make it bullet proof!

app/reducers/networkSelector/index.ts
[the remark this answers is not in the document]
tommasini: Addressed on this latest commit!

app/core/Engine.ts
```diff
@@ -1298,6 +1314,29 @@ class Engine {
...
     );
 
+    this.controllerMessenger.subscribe(
+      'NetworkController:networkDidChange',
+      async () => {
+        try {
+          const networkId = await deprecatedGetNetworkId();
+          store.dispatch(networkIdUpdated(networkId));
+        } catch (error) {
+          captureException(
```
Gudahtt, line 1324: Perhaps it would be better to log this error to the console rather than send it to Sentry? Generally we should only use Sentry for _actionable_ things, like errors that we would then investigate and fix. If there is a failure case that we expect to happen, we should handle it in the best way we can and move on.

This catch block will be run for any chains that fail to implement `net_version`, and in any cases where the user has a network disruption. We cannot do anything to resolve people's network issues or RPC issues. Perhaps if we think this is associated with a significant problem, we could add metrics here to track this. But otherwise I'd suggest just logging a message so that we know this happened when debugging/investigating some other issue.

tommasini: Yeah, I added to help on investigation if there was some most used networks that fail to implement net_version, but probably it's premature and we only should add the sentry catchException if we think it's a bug that we want to investigate! Thanks Mark!

Will change it to a log!

tommasini:
Hmm something is weird at store level
I will need to understand why this is happening 
```
"networkProvider":"{\"networkProvider\":{\"networkId\":\"\"},\"networkId\":\"43114\"}","_persist":"{\"version\":37,\"rehydrated\":true}"}
```

tommasini:
Addressed on this last commit

app/core/Engine.ts
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/bb6b1ff095896ea9df93b0189e331aaeb1542bf6)

tommasini:
Re-tested branch, and added new recording to the description with an update from main (applying the migration) with 2 accounts (one imported via private key) and optimism network selected 
The store data format is fine, after playing around sometime and it applies the migrations as expected
(On this case I copied the data after some testing and I was on polygon network)
```
"experimentalSettings":"{\"securityAlertsEnabled\":true}","networkProvider":"{\"networkId\":\"137\"}","_persist":"{\"version\":37,\"rehydrated\":true}"}
```

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+
+    it('returns the same string if it is numeric', () => {
+      expect(convertNetworkId('123')).toEqual('123');
+    });
+
+    it('throws an error for non-numeric, non-hexadecimal strings', () => {
+      expect(() => convertNetworkId('abc')).toThrow(
+        "Cannot parse as a valid network ID: 'abc'",
+      );
+    });
```
NicolasMassart, line 366: This seems to be the same to me than the next test "throws an error for non-numeric strings":
- `"abc"` is not a number but a string and can't be evaluated as base 10 number.
- `"123abc"` is not a number but a string and can't be evaluated as base 10 number.

- `"abc"` is hex value missing `0x`
- `"123abc"` is hex value missing `0x`

Meaning that both test are testing exactly the same thing.

app/actions/networkProvider/index.ts
```diff
@@ -0,0 +1,8 @@
+export const networkIdUpdated = (networkId: string) => ({
+  type: 'NETWORK_ID_UPDATED',
+  networkId,
+});
+export const networkIdWillUpdate = () => ({
+  type: 'NETWORK_WILL_UPDATE',
+  networkId: '',
```
Gudahtt, line 7: Nit: Since this action is only used to clear the `networkId` state, we don't need to pass along any value here. The reducer knows that this action means to clear the state.

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+      expect(() => convertNetworkId(NaN)).toThrow(
+        "Cannot parse as a valid network ID: 'NaN'",
+      );
+    });
+
+    it('throws an error for non-numeric strings', () => {
+      expect(() => convertNetworkId('123abc')).toThrow(
+        "Cannot parse as a valid network ID: '123abc'",
+      );
+    });
```
NicolasMassart, line 378: see previous comment

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+    beforeEach(() => {
+      // Reset mocks before each test
+      jest.clearAllMocks();
+      // Setup default behavior for mocked functions
+      Engine.controllerMessenger.call.mockReturnValue({
+        sendAsync: mockSendAsync,
+      });
+    });
+
+    it('should resolve with the correct network ID', async () => {
```
NicolasMassart, line 393:
```suggestion
    it('resolves with the correct network ID', async () => {
```

app/actions/networkProvider/index.ts
```diff
@@ -0,0 +1,8 @@
+export const networkIdUpdated = (networkId: string) => ({
```
Gudahtt, line 1: Nit: It would be great to have JSDoc comments for these

app/actions/networkProvider/index.ts
Gudahtt, line 1: Nit: What motivated the name `networkProvider`? Personally I was thinking of this state as being for the `inpageProvider`. That's the name of the provider API that this state is used for. I am curious to know how you were thinking of it though

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+      expect(convertNetworkId('123')).toEqual('123');
+    });
+
+    it('throws an error for non-numeric, non-hexadecimal strings', () => {
+      expect(() => convertNetworkId('abc')).toThrow(
+        "Cannot parse as a valid network ID: 'abc'",
+      );
+    });
+
+    it('throws an error for NaN values', () => {
```
NicolasMassart, line 368:
```suggestion
    it('throws for NaN values', () => {
```

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+    it('converts a hexadecimal string to a decimal string', () => {
+      // Assuming convertHexToDecimal works correctly for '0x1' and returns 1
+      expect(convertNetworkId('0x1')).toEqual('1');
+    });
+
+    it('returns the same string if it is numeric', () => {
+      expect(convertNetworkId('123')).toEqual('123');
+    });
+
+    it('throws an error for non-numeric, non-hexadecimal strings', () => {
```
NicolasMassart, line 362: You can't throw anything else than an error, so writing "throws an error" is redundant
```suggestion
    it('throws for non-numeric, non-hexadecimal strings', () => {
```

app/actions/networkProvider/index.ts
Gudahtt, line 1: Suggestion: Are you familiar with `@reduxjs/toolkit`? It is my understanding that we are migrating to using this library for reducers/actions/action creators, rather than writing them by hand. This library has a few convenience methods that make it easier to create new Redux state in a way that follows best-practices, and even uses `Immer` internally to ensure that we don't accidentally mutate state.

I see that it has already been partially used for the `engineReducer`, in `app/core/redux/slices/engine`. That file follows the best practice of grouping the reducer in the same file as the selectors and action creators (see here: https://redux.js.org/style-guide/#structure-files-as-feature-folders-with-single-file-logic). Though it doesn't use the `createSlice` utility.

Perhaps we could use the same format as the `engineReducer` here and follow the newer pattern, or use `createSlice` instead like we use for newer Redux slices in the extension.

app/actions/networkProvider/index.ts
Gudahtt: We could also consider making the reducer set this to `loading` rather than an empty string, so that we could simplify the `BackgroundBridge`. If this value represents the `networkVersion` we use in the inpage provider, that would seem appropriate

app/core/Engine.ts
```diff
@@ -1298,6 +1314,28 @@ class Engine {
...
     );
 
+    this.controllerMessenger.subscribe(
+      'NetworkController:networkDidChange',
+      async () => {
+        try {
+          const networkId = await deprecatedGetNetworkId();
+          store.dispatch(networkIdUpdated(networkId));
+        } catch (error) {
+          Logger.error(
```
Gudahtt, line 1324: `Logger.error` will still send the error to Sentry, unfortunately. I am not sure what the recommended practice is for logging an error without sending it to Sentry.

e2e/fixtures/fixture-builder.js
```diff
@@ -585,6 +585,7 @@ class FixtureBuilder {
         '@MetaMask:UserTermsAcceptedv1.0': 'true',
         '@MetaMask:WhatsNewAppVersionSeen': '6.5.0',
       },
+      networkId: '',
```
Gudahtt, line 588: It's curious that this is not nested under a `networkProvider` property. is this meant to be at this level of the fixture?

NicolasMassart (commented):
Nice work!
I'm missing some subtleties on the patch part, if you could tell me more.
Also don't forget to check [our guidelines](https://github.com/MetaMask/contributor-docs/blob/main/docs/unit-testing.md#use-it-to-specify-the-desired-behavior-for-the-code-under-test)!

app/core/Engine.ts
tommasini: Ow, you are right! My bad! Probably `console.error`

app/actions/networkProvider/index.ts
tommasini: When I choose `networkProvider`, I was thinking on internal network provider properties (like the networkId), or others that we could need to save outside the NetworkController, but I don't have a stronger reason to keep that name.
Never thought about `inpageProvider` , it is definitely more suitable! Will make the change! Thanks Mark!

e2e/fixtures/fixture-builder.js
tommasini: I missed this change of when creating the new actions and reducer, and then solving the underlying stored data issue, good catch Mark! Thanks!

app/actions/networkProvider/index.ts
tommasini: I'm not 100% familiar but it's something that I looked into on the past and we have been talking about.

Yeah it's indeed better for start to using it instead of creating the reducers/actions manually, will try to follow the pattern! Cool! Thanks Mark!

tommasini: I remember talking with @MarioAslau and there was some issue around using the createSlice, but I don't remember which one, probably something cool to re-check!

app/store/migrations/033.ts
```diff
@@ -43,6 +41,17 @@ export default async function migrate(stateAsync: unknown) {
     return state;
   }
 
+  if (!isObject(networkControllerState.providerConfig)) {
+    captureException(
+      new Error(
+        `Migration 33: NetworkController providerConfig not found: '${JSON.stringify(
```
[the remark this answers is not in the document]
tommasini: Yeah this is a very good point! I will change it! Thanks Nico!

app/store/migrations/037.ts
```diff
@@ -0,0 +1,59 @@
...
+        `Migration 37: Invalid NetworkController state error: '${typeof networkControllerState}'`,
+      ),
+    );
+    return state;
+  }
+
+  if (!networkControllerState.networkId) {
+    captureException(
+      new Error(
+        `Migration 37: Invalid NetworkController networkId not found: '${JSON.stringify(
```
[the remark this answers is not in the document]
tommasini: Completely agree! Thanks Nico!

app/util/networks/index.test.ts
tommasini: Yeah I see, I left the last one because it was showing that if it had mixed numbers and letters on a string was not trying to convert, but yeaj it's the same as the string one, removed it!
Thanks Nico!

patches/@metamask+network-controller+13.0.1.patch
```diff
@@ -60,20 +60,20 @@ index 596ed70..5bd7d6f 100644
...
 +_NetworkController_ethQuery = new WeakMap(), _NetworkController_infuraProjectId = new WeakMap(), _NetworkController_trackMetaMetricsEvent = new WeakMap(), _NetworkController_previousProviderConfig = new WeakMap(), _NetworkController_providerProxy = new WeakMap(), _NetworkController_provider = new WeakMap(), _NetworkController_blockTrackerProxy = new WeakMap(), _NetworkController_destroyed = new WeakMap(), _NetworkController_autoManagedNetworkClientRegistry = new WeakMap(), _NetworkController_instances = new WeakSet(), _NetworkController_refreshNetwork = function _NetworkController_refreshNetwork() {
      return __awaiter(this, void 0, void 0, function* () {
          this.messagingSystem.publish('NetworkController:networkWillChange');
-         this.update((state) => {
+         __classPrivateFieldGet(this, _NetworkController_instances, "m", _NetworkController_applyNetworkSelection).call(this);
```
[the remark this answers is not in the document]
tommasini: This line 79 is not part of the actual applied code of the patch, it's just the rest of the file

app/util/networks/index.test.ts
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/util/networks/index.test.ts
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+      );
+    });
+
+    it('throws an error for NaN values', () => {
+      expect(() => convertNetworkId(NaN)).toThrow(
+        "Cannot parse as a valid network ID: 'NaN'",
+      );
+    });
+
+    it('throws an error for non-numeric strings', () => {
```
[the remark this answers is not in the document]
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/util/networks/index.test.ts
```diff
@@ -339,4 +345,78 @@ describe('network-utils', () => {
...
+      mockSendAsync.mockImplementation((_, callback) => {
+        callback(new Error('Failed to fetch network ID'), null);
+      });
+
+      await expect(deprecatedGetNetworkId()).rejects.toThrow(
+        'Failed to fetch network ID',
+      );
+    });
+
+    it('should throw an error if the provider has not been initialized', async () => {
```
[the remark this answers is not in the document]
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/store/migrations/037.ts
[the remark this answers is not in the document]
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/store/migrations/037.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      expect(newState).toStrictEqual(state);
+      expect(mockedCaptureException).toHaveBeenCalledWith(expect.any(Error));
+      expect(mockedCaptureException.mock.calls[0][0].message).toBe(
+        errorMessage,
+      );
+    });
+  }
+
+  it('Should have deleted networkId from Network Controller and copied to networkId property', async () => {
```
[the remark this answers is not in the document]
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/store/migrations/033.ts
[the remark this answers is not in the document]
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -38,6 +36,21 @@ const EventEmitter = require('events').EventEmitter;
...
 import { getPermittedAccounts } from '../Permissions';
+import { NetworkStatus } from '@metamask/network-controller';
+
+const legacyNetworkId = () => {
+  const { networksMetadata, selectedNetworkClientId } =
+    store.getState().engine.backgroundState.NetworkController;
+
+  const { networkId } = store.getState().networkProvider;
+
+  if (!networkId) return 'loading';
```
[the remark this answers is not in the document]
tommasini: Yes definitely! Thanks Nico!
Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

e2e/fixtures/fixture-builder.js
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/actions/networkProvider/index.ts
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44)

app/actions/networkProvider/index.ts
tommasini: Added JS docs to the new slice file 🚀

app/actions/networkProvider/index.ts
tommasini: It was addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/e4235d1ce48e74fe7ed729022dc8eace13ab5e44), it's more simplified, now we do not check if the networkId is null, if it's available we return the networkId if not we return "loading", wdyt?

app/store/migrations/037.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        },
+      }),
+      errorMessage:
+        "Migration 37: Invalid NetworkController networkId not found: 'null'",
+      scenario: 'providerConfig is invalid',
+    },
+  ];
+
+  for (const { errorMessage, scenario, state } of invalidStates) {
+    it(`should capture exception if ${scenario}`, async () => {
```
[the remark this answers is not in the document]
tommasini: Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9065/commits/21ef7c5cd06ad8f3616ea740d35c4964a2b8bd7e)

Gudahtt (dismissed):
LGTM!

NicolasMassart (dismissed):
Looks good to me.

NicolasMassart (dismissed):
Looks good to me.

NicolasMassart (dismissed):
Looks good to me.

tommasini:
Uprade path from version 7.19.0 with networks imported

https://github.com/MetaMask/metamask-mobile/assets/46944231/23f4ccd1-8959-43f9-9ae2-6bd5ad8c1d95

christopherferreira9:
all good with the sdk now 💪

NicolasMassart (approved):
Looks good to me.

PR #9070, opened 2024-03-27
feat: remove selectSelectedAddress in favour of selectSelectedInternalAccount

owencraston:
## **Description**
In an effort to make MetaMask multi chain we are migrating from an address based accounts to account id based account. In [this](https://github.com/MetaMask/metamask-mobile/pull/8759) previous PR we integrated the accounts controller which stores metadata like the account name, address and supported methods. This pr migrates all the use of `selectSelectedAddress` to use the accounts controller as the source of truth for the selected address.
- created `selectSelectedInternalAccount` which returns the entire account object for the currently selected address
- created a `selectSelectedInternalAccountAddressAsChecksum` that returns just the selected address instead of the entire accounts object
- deprecated `selectSelectedAddress`
- removed all use of `selectSelectedAddress` in favour of `selectSelectedInternalAccountAddressAsChecksum` or `selectSelectedInternalAccount`
    - For screens that only use the address then we will use the `selectSelectedInternalAccountAddressAsChecksum`
    - If the screen is using more than just the address (the account name for example) then we will select the entire account with `selectSelectedInternalAccount`
- modified all the screen tests to mock the accounts controller state
    - I created some test helpers to make mocking the accounts controller
        - `createMockInternalAccount` and `createMockUUIDFromAddress` in `app/selectors/accountsController.test.ts`

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2548

#### Fresh install
1. Launch wallet
2. create a new account
4. once on the wallet view, add a new account
5. create a custom name for your account(s)
6. swipe away the app and re open
7. login
8. the selected account should be same one that was previously selected
9. all addresses should be in [checksum format](https://coincodex.com/article/2078/ethereum-address-checksum-explained/)
10. connect the portfolio dapp via the portfolio button on the home screen. 
11. Click connect on the portfolio dapp
12. The first address in the address picker should be match the selected address on the wallet view and be in checksum format

#### Upgrade path

1. Use an existing wallet that has multiple accounts created/imported
2. Ideally change the names of your accounts so that they have custom names
3. "upgrade" your wallet to this branch
4. the selected account should be the same as it was on the previous branch
5. all of the account data should remain the same (names and balances)

### **Before**

N/A

### **After**

https://github.com/MetaMask/metamask-mobile/assets/22918444/52938128-9fbc-48a8-a8bf-0a0f4d7b2b14


chris6090 (approved):
kriseng83@gmail.com

app/components/Nav/App/index.js
```diff
@@ -107,6 +107,7 @@ import generateUserSettingsAnalyticsMetaData from '../../../util/metrics/UserSet
 import OnboardingSuccess from '../../Views/OnboardingSuccess';
 import DefaultSettings from '../../Views/OnboardingSuccess/DefaultSettings';
 import BasicFunctionalityModal from '../../UI/BasicFunctionality/BasicFunctionalityModal/BasicFunctionalityModal';
+import { toChecksumAddress } from 'ethereumjs-util';
```
tommasini, line 110: Should we use the controller utils [toChecksumHexAddress](https://github.com/MetaMask/core/blob/01fb17da36e8a7b9b27f7cff4c200eb6bd6b004d/packages/controller-utils/src/util.ts#L255C17-L255C37) instead?

app/components/Nav/App/index.js
```diff
@@ -312,12 +313,21 @@ const App = ({ userLoggedIn }) => {
...
-      const { PreferencesController } = Engine.context;
-      const selectedAddress = PreferencesController.state.selectedAddress;
+      const { AccountsController } = Engine.context;
+      const { selectedAccount } = AccountsController.state.internalAccounts;
+
+      const selectedInternalAccount =
+        AccountsController.state.internalAccounts.accounts[selectedAccount] ||
+        null;
+
+      const selectedAddress = selectedInternalAccount?.address;
```
tommasini, line 323: This two lines could be?
```suggestion
      const selectedAddress = AccountsController.state.internalAccounts?.accounts?.[selectedAccount]?.address;
```

tommasini, line 323: These two lines could be
This two lines could be?

```suggestion
      const selectedAddress = AccountsController.state.internalAccounts?.accounts?.[selectedAccount]?.address;

```

app/components/UI/AccountApproval/index.test.tsx
```diff
@@ -14,6 +34,13 @@ jest.mock('../../../core/Engine', () => ({
     },
     KeyringController: {
       getAccountKeyringType: () => Promise.resolve('HD Key Tree'),
+      state: {
```
tommasini, line 37: To make sure I understand!
The snapshot needed to be updated because now we have on the state an address on the keyrings mock data, that way it would be rendering the account image.

It's not because of the change of the selector itself, that should keep the same behaviour, right?

app/components/UI/OnboardingWizard/Step3/index.js
```diff
@@ -134,15 +128,8 @@ Step3.propTypes = {
...
-  selectedAddress: selectSelectedAddress(state),
-  identities: selectIdentities(state),
-});
-
 const mapDispatchToProps = (dispatch) => ({
   setOnboardingWizardStep: (step) => dispatch(setOnboardingWizardStep(step)),
 });
 
-export default connect(mapStateToProps, mapDispatchToProps)(Step3);
+export default connect(mapDispatchToProps)(Step3);
```
tommasini, line 135: Hmm I think we need to do something like 
```suggestion
export default connect(null, mapDispatchToProps)(Step3);
```

If i'm reading correctly the connect signature: https://react-redux.js.org/api/connect

app/components/Nav/App/index.js
owencraston: Clean

app/components/UI/OnboardingWizard/Step3/index.js
owencraston: thank you. addressed in the latest commit.

app/components/UI/AccountApproval/index.test.tsx
owencraston: Screenshot 2024-04-29 at 6 26 07 PM

Without this state mock the test was failing. In general nothing is changing in any of these screens. The only difference is where we are getting the selectedAddress information from (accounts controller vs preferences controller). Because we are not getting all this information from the accounts controller the tests needed the mock data in state.

app/components/Nav/App/index.js
owencraston: Yeah we can use this since it is just a wrapper around `toChecksumAddress` that we control. I went ahead and updated all of the pr to use this method.

ios/Podfile.lock
```diff
@@ -1028,7 +1028,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   Base64: cecfb41a004124895a7bcee567a89bae5a89d49b
-  boost: 57d2868c099736d80fcd648bf211b4431e51a558
+  boost: 7dcd2de282d72e344012f7d6564d024930a6a440
```
legobeat, line 1031: Why is this here?

owencraston: good catch. this was generated from yarn setup and probably needs to be committed to main but I will remove it from this PR.

owencraston: after rebasing on main this change disappeared!

app/components/Views/AccountActions/AccountActions.test.tsx
```diff
@@ -107,7 +103,7 @@ describe('AccountActions', () => {
     fireEvent.press(getByTestId(SHARE_ADDRESS));
 
     expect(Share.open).toHaveBeenCalledWith({
-      message: '0xe7E125654064EEa56229f273dA586F10DF96B0a1',
+      message: '0xC4966c0D659D99699BFD7EB54D8fafEE40e4a756',
```
montelaidev, line 106: I think we can import the MOCK_ADDRESS_2 and use the value here instead of just changing the string. The test would break if MOCK_ADDRESS_2 changes in the future.

app/components/Views/EditAccountName/EditAccountName.test.tsx
```diff
@@ -6,11 +6,40 @@ import { fireEvent } from '@testing-library/react-native';
...
+        ],
+        type: 'eip155:eoa',
+      },
+    },
+    selectedAccount: '30313233-3435-4637-b839-383736353430',
+  },
+};
+
 jest.mock('../../../core/Engine', () => ({
   setAccountLabel: () => mockEngineSetAccountLabel,
```
montelaidev, line 44: We should test if the mockPreferencesSetAccountLabel and mockAccountsControllerSetAccountName has been called. Also try to it the edge case where both controllers are not called when the account id doesn't exist.

app/components/Views/Asset/index.test.js
```diff
@@ -19,6 +39,13 @@ jest.mock('../../../core/Engine.ts', () => ({
   context: {
     KeyringController: {
       getQRKeyringState: async () => ({ subscribe: () => ({}) }),
+      state: {
+        keyrings: [
+          {
+            accounts: ['0xe64dD0AB5ad7e8C5F2bf6Ce75C34e187af8b920A'],
```
gantunesr, line 45:
```suggestion
            accounts: [MOCK_ADDRESS],
```

gantunesr (commented):
I have a left some nit comments about the naming of some testing variables, for example,

- I see the use of numbering for some variables like `internalAccount1`, `internalAccount2`, etc. and then the same numbering pattern is not used for `MOCK_ADDRESS`. We should consider also having it like `MOCK_ADDRESS_1`, `MOCK_ADDRESS_2`, etc.
- The follow this convention regarding the use of abbreviations: "_Treat abbreviations like acronyms in names as whole words, i.e. use loadHttpUrl, not ~loadHTTPURL~, unless required by a platform name (e.g. XMLHttpRequest)._". In reference to names that have `UUID`.

app/selectors/preferencesController.ts
```diff
@@ -17,6 +17,9 @@ export const selectIpfsGateway = createSelector(
     preferencesControllerState.ipfsGateway,
 );
 
+/**
+ * @deprecated use selectSelectedInternal or selectSelectedInternalAccountAddressAsChecksum account from selectors/accountsController.ts
+ */
 export const selectSelectedAddress = createSelector(
   selectPreferencesControllerState,
   (preferencesControllerState: PreferencesState) =>
```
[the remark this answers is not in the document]
owencraston: I am hesitant to do this since it implies that we are getting the value from the preferences controller when that is not the case. We want to shift people away from preferences controller for this type of information as much as possible. I think there is a good argument to be made here that we could delete the `selectSelectedAddress` selector entirely in this PR. What do you think?

app/components/hooks/useAccounts/useAccounts.ts
```diff
@@ -128,11 +128,14 @@ const useAccounts = ({
       }: { accounts: string[]; type: KeyringTypes } = keyring;
       for (const index in accountAddresses) {
         const checksummedAddress = toChecksumAddress(accountAddresses[index]);
-        const isSelected = selectedAddress === checksummedAddress;
+        const isSelected =
+          toChecksumAddress(selectedInternalAccount.address) ===
```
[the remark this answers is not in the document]
owencraston: good call.

app/components/UI/AccountOverview/index.test.tsx
```diff
@@ -4,8 +4,32 @@ import AccountOverview from './';
...
 import Engine from '../../../core/Engine';
+import {
+  createMockInternalAccount,
+  createMockUUIDFromAddress,
+} from '../../../selectors/accountsController.test';
+import { AccountsControllerState } from '@metamask/accounts-controller';
 
 const mockedEngine = Engine;
+
+const MOCK_ADDRESS = '0x76cf1CdD1fcC252442b50D6e97207228aA4aefC3';
```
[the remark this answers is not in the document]
owencraston: I kept this test as is to make the changes minimal but the mock was required. do you think more testing is needed here?

app/components/Views/AccountActions/AccountActions.test.tsx
owencraston: I intentionally did not use the MOCK_ADDRESS variables here so that the test would fail if the value changed. My logic for doing this is that it makes the person maintaining the test more aware of what is going on? Perhaps this is not the best strategy though.

app/components/Nav/App/index.js
```diff
@@ -312,12 +313,19 @@ const App = ({ userLoggedIn }) => {
...
     if (prevNavigator.current || !navigator) return;
     const appTriggeredAuth = async () => {
-      const { PreferencesController } = Engine.context;
-      const selectedAddress = PreferencesController.state.selectedAddress;
+      const { AccountsController } = Engine.context;
+      const { selectedAccount } = AccountsController.state.internalAccounts;
+
+      const selectedAddress =
+        AccountsController.state.internalAccounts?.accounts?.[selectedAccount]
+          ?.address;
```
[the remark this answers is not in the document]
owencraston: we cannot since the redux state may not be initialized yet on the first load.

app/components/Views/EditAccountName/EditAccountName.test.tsx
owencraston: For some reason this test is not working for me.

```ts
  it('should call both setAccountLabel in PreferencesController and AccountsController when saving a new account name', () => {
    const { getByTestId } = renderComponent(mockInitialState);
    const input = getByTestId('account-name-input');
    const saveButton = getByTestId('save-button');

    fireEvent.changeText(input, 'New Account Name');
    fireEvent.press(saveButton);

    expect(mockPreferencesSetAccountLabel).toHaveBeenCalled();
    expect(mockAccountsControllerSetAccountName).toHaveBeenCalled();
  });
```

app/components/Views/Asset/index.test.js
owencraston: No clue why but when I do the test fails with this error...

```
    TypeError: Cannot read properties of undefined (reading 'toLowerCase')

      182 |   return keyrings.find((keyring) =>
      183 |     keyring.accounts
    > 184 |       .map((account) => account.toLowerCase())
          |                                 ^
      185 |       .includes(address.toLowerCase()),
      186 |   );
      187 | }
```

app/selectors/accountsController.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  internalAccounts: {
+    accounts: {
+      [expectedUUID]: internalAccount1,
+      [expectedUUID2]: internalAccount2,
+    },
+    selectedAccount: expectedUUID2,
+  },
+};
+
+describe('Accounts Controller Selectors', () => {
```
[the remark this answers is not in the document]
owencraston: Good call, I added some error handling inside the selectors to throw in these cases as well as report the error to sentry. Also I updated the tests to handle these cases. We do not need to worry about the `selectedAccount` being undefined since it is type checked to be always defined. the real issue would be if the id does not match any of the accounts. My latest commit handles this.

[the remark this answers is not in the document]
owencraston: This should never happen but adding error logging is a good measure to put in place in case it does start to happen.

app/selectors/preferencesController.ts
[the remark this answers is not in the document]
gantunesr: If removing/deprecating this method is possible then we should do it, but I see no risk in read operation with the state of both controllers are in sync. The bigger risk is in write operations.

app/components/Views/AccountActions/AccountActions.test.tsx
gantunesr: I think it depends on the purpose of the test. For this case it should be ok to use `MOCK_ADDRESS_2` directly instead of the explicit value.

app/selectors/accountsController.test.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export function createMockUuidFromAddress(address: string): string {
+  const fakeShaFromAddress = Array.from(
+    { length: 16 },
+    (_, i) => address.charCodeAt(i) || 0,
+  );
+  return uuid({
+    random: fakeShaFromAddress,
+  });
+}
```
gantunesr, line 18: Since this method is widely used across the codebase, we should consider moving it to a test helpers directory

gantunesr (commented):
Looking good! Left a few comments

app/selectors/preferencesController.ts
```diff
@@ -17,6 +17,9 @@ export const selectIpfsGateway = createSelector(
     preferencesControllerState.ipfsGateway,
 );
 
+/**
+ * @deprecated use selectSelectedInternal or selectSelectedInternalAccountChecksummedAddress account from selectors/accountsController.ts
+ */
```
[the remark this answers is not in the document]
owencraston: In my opinion we should leave this as is (marked as deprecated) until we are 100% finished with the migration. Although this method should not be used from now on, if we encounter an issue with the accounts controller state then we can instruct people to use this as an alternative. Also as you mentioned there is no real risk in people reading the state from a different source.

app/selectors/accountsController.test.ts
owencraston: good call. I will move it into `app/utils/test/`

app/components/Views/EditAccountName/EditAccountName.test.tsx
owencraston: I am having trouble getting this test working. Since I will be replacing this method very soon in favour of [data syncing](https://github.com/MetaMask/metamask-mobile/pull/9564) I am not sure it really requires testing at this moment.

gantunesr (dismissed):

app/components/UI/Ramp/Views/OrderDetails/OrderDetails.test.tsx
```diff
@@ -46,9 +51,26 @@ type DeepPartial<BaseType> = {
...
+
+const MOCK_ACCOUNTS_CONTROLLER_STATE: AccountsControllerState = {
+  internalAccounts: {
+    accounts: {
+      [expectedUuid]: internalAccount1,
+    },
+    selectedAccount: expectedUuid,
+  },
+};
+
```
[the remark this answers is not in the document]
owencraston: @wachunei thanks so much for the review. I actually already created a [default mock](https://github.com/MetaMask/metamask-mobile/pull/9070/files#diff-709f964fc8ad71a4b1d46a21b00298d0f480f300e1daca2a28340babc7702e9dR57) that is imported into several files that do not require any custom logic. The reason I did not use that for this file (and other similar files) was because the address as needed in multiple mocks. My reason for not creating a single `createMockAccountsControllerState` function is because in order to fully test the accounts controller state, the accountIds are needed. See [selectors/accountsController.test.ts](https://github.com/MetaMask/metamask-mobile/pull/9070/files#diff-db57224f38dc51e2d622d498ed71656bd1cd507159c261e1c2ff6a6ef819a77dR35) as an example. By exposing the ids it enables more testing coverage in the future since all accounts will be referenced by id in the future.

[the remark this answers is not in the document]
owencraston: That being said I do not feel strongly about this and will happily make your suggested changes if you prefer.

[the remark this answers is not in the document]
owencraston: @wachunei I have addressed your feedback in my latest commit.

owencraston:
Full e2e regression has also passed: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/e2c81e5c-22aa-4f3a-89ff-47b9384c134e

wachunei (dismissed):

app/components/Views/confirmations/Send/index.test.tsx
```diff
@@ -1,5 +1,27 @@
 import { renderScreen } from '../../../../util/test/renderWithProvider';
 import Send from '.';
+import {
+  MOCK_ACCOUNTS_CONTROLLER_STATE,
+  MOCK_ADDRESS_1,
+} from '../../../../util/test/accountsControllerTestUtils';
+
+// const MOCK_ADDRESS = '0x76cf1CdD1fcC252442b50D6e97207228aA4aefC3';
```
tommasini, line 8: Can we remove this comments or are we leaving those here for some reason?

tommasini (dismissed):
LGTM!

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -1487,7 +1487,11 @@
 				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
-				OTHER_LDFLAGS = "$(inherited)";
+				OTHER_LDFLAGS = (
```
tommasini, line 1490: Is this changes needed for any reason? If not can we revert those?

tommasini (dismissed):
LGTM!

wachunei (approved):
LGTM

wachunei (dismissed):
LGTM

jpuri (dismissed):

gantunesr (dismissed):

jpuri (dismissed):

vivek-consensys:
Issue found on both Android and iOS using the latest [bitrise build](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/22308fd2-4d01-43d5-ac27-15e8b3b94d60), steps followed:- 

1. Create new wallet
2. Skip SRP section
4. Rename account name
5. Add new account
6. Rename second account
7. Delete wallet
8. Import another existing wallet using SRP
9. Click accounts
10. Error displayed


https://github.com/MetaMask/metamask-mobile/assets/106310394/a7c0ecaa-072d-4d74-9efb-7cf4cde29b76

image

owencraston:
New qa build: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/61898e26-3bd0-4895-954c-1b6f2919ab74

owencraston:
@vivek-consensys I was able to address this issue in another [fix for the release](https://github.com/MetaMask/metamask-mobile/pull/9764).

jpuri (dismissed):

owencraston:
Since this PR has started, the selectors were cherry picked into main and used as a fix in the latest RC. You can find these PRs [here](https://github.com/MetaMask/metamask-mobile/pull/9753) and [here](https://github.com/MetaMask/metamask-mobile/pull/9733). This is why you do not see `app/selectors/accountsController.ts` as a new file addition in this PR anymore.

Cal-L (dismissed):
Code LGTM but let's wait until E2E passes

Cal-L (dismissed):
Lgtm

owencraston:
@plasmacorral New [builds](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/a3d393f6-9ef3-412c-a452-0230f8c3fb5a) for you sir.

owencraston:
Latest QA build: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/f29a3218-cee5-4703-be52-aa845537d830

owencraston:
Latest QA build: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/942717b3-3692-43e9-a9a5-304ccc88e00d

plasmacorral:
During comprehensive QA testing of the migration from production v7.23.0 to the current branch, I encountered a consistent issue across all tested devices when attempting to change the user's password under specific conditions. This issue seems to be introduced by changes in this PR, as it was not observed in the latest QA builds of the main branch or the latest production version.

Devices Tested:

iPhone 13 mini (iOS 15.6.1) - Face ID
Samsung a515f (Android 12) - Fingerprint
iPhone Xs (iOS 17.4.1) - Face ID
Pixel 5a (Android 14) - Password

Pre-Migration Configuration:
Accounts: 3 HD, 1 imported (with ENS name), 1 Ledger
Dapp Connections: Various, including ENS domains, DeFi Saver, PancakeSwap, and OpenSea

Successful tests:
- All accounts, labels, and dapp connections migrated
- Adding/removing accounts
- transaction editing to switch from account
- testing dapp signatures
- Sign-In with Ethereum (SIWE)
- Legacy Send
- EIP 1559 transactions
- EIP 1559 transactions without gas
- deploying ERC-20, ERC-721, and ERC-1155 tokens 
- revealing secrets
- switching accounts

Observation: Attempting to change the password results in an error on the first attempt, across all devices, when an imported account is present. The error is bypassed upon selecting "Try Again," and the password change then succeeds.

Reproduction Steps:

1. Import an SRP.
2. Import an account by private key.
3. Navigate to Settings > Security & Privacy and select "Change Password."
4. Follow the prompts to attempt a password change.
5. Observe the error on the first attempt, success upon retry.

This issue was consistently reproducible on this branch with an imported account present, suggesting a potential regression or bug introduced by the PR's changes.

Screenshot: 
Screenshot 2024-06-06 at 7 03 31 PM

Given the specific conditions under which this issue occurs, it may be related to how imported accounts are handled during the password change process in the updated codebase.

plasmacorral:
Will open a unique issue for the edge case observed per slack thread and discussion with team, and work on merging this as-is and address the error in a subsequent PR.

owencraston:
@plasmacorral I have addressed your qa issue. 

https://github.com/MetaMask/metamask-mobile/assets/22918444/b1eff8e7-4ca7-4412-8429-8386e6c33788

Cal-L (approved):
LGTM

plasmacorral:
Confirmed password change with imported account present no longer presents an error.  Conducted migration test with QA build of commit d467a2b6e26f99eb164dfea9a63153d1b97be205 and Account labels and Dapp connections persisted.

PR #9216, opened 2024-04-12
fix: correct `SubjectType` for origins connecting via the `BackgroundBridge`

FrederikBolding:
## **Description**

The `SubjectType` for origins connecting via the `BackgroundBridge` was incorrectly set to `SubjectType.Snap` for all origins. This PR corrects it to assuming that all origins are of type `SubjectType.Website`. 

This is important in order to enforce that Snap-specific RPC methods don't leak into the possession of dapps.


Jonathansoufer (approved):
LGTM!

PR #9231, opened 2024-04-15
chore: ensure gemfile versions

Cal-L:
## **Description**

This PR updates the project to ensure that versions listed in the `GemFile` are respected when running `yarn setup`. The `yarn setup` command will automatically install the versions as specified in the `GemFile`. The `Gems` section in the `environment.md` file has also been updated to instruct users to install `bundler` and to run the `bundle install` command. With the correct versions, contributors should be able to run `yarn setup` successfully. Lastly, an optional `--build-ios` flag can be passed to the `yarn setup` command, which forces iOS specific steps to run on non-macOS devices

## **Related issues**

Fixes:

## **Manual testing steps**

1. Uninstall cocoapods `gem uninstall cocoapods -v 1.15.2`
2. Run `gem list` and see that cocoapods v1.15.2 does not exist
3. Run `yarn setup` and see that it passes (means cocoapods also successfully installed pods)
4. Run `gem list` to see that cocoapods v1.15.2 was auto installed

### Bundle install command installs the correct `cocoapods` gem as specified in the `Gemfile`
image

### Running yarn setup after bundle install results in successful setup (this also automatically calls `bundle install`)
image


docs/readme/environment.md
```diff
@@ -65,21 +65,20 @@ npm install -g yarn
...
-     - Check `Show Package Details` option below the tools list to show available versions
-     - Locate `NDK (Side-by-side)` option in the tools list
-     - Check NDK version `24.0.8215888`
-     - Locate `CMake` option in the tools list
-     - Check CMake version `3.22.1`
-     - Click "Apply" or "OK" to download
-   - Finally, start the emulator from Android Studio:
-     - Open "Virtual Device Manager"
-     - Launch emulator for "Pixel 5 <relevant API version mentioned in [React Native Getting Started](https://reactnative.dev/docs/environment-setup#installing-dependencies)>"
+Set environment variable `JAVA_HOME=/Applications/Android Studio.app/Contents/jbr/Contents/Home` to use java version shipped on the Android Studio App
```
legobeat, line 68:
```suggestion
Set environment variable `JAVA_HOME=/Applications/Android Studio.app/Contents/jbr/Contents/Home` to use the Java version shipped with Android Studio.
```

docs/readme/environment.md
```diff
@@ -65,21 +65,20 @@ npm install -g yarn
...
+
+- Go to Settings > Languages & Frameworks > Android SDK
```
legobeat, line 70:
```suggestion
```suggestion
- Set environment variable `JAVA_HOME=/Applications/Android Studio.app/Contents/jbr/Contents/Home` to use the Java version shipped with Android Studio
- Go to Settings > Languages & Frameworks > Android SDK
```

scripts/setup.mjs
```diff
@@ -107,6 +107,31 @@ const ppomBuildTask = {
...
 }
 
+const gemInstallTask = {
+  title: 'Install gems',
+  task: (_, task) => task.newListr([
+    {
+      title: 'Install gems using bundler',
+      task: async (_, podInstallTask) => {
+        const isOSX = process.platform === 'darwin';
+        if (!isOSX) {
```
legobeat, line 117: These are installing fine on Linux - should we consider installing them there to as a path towards making `pod install` work outside of Mac as well?

Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
-ruby '>= 3.0.0'
+ruby '3.1.4'
```
legobeat, line 4: Why pin ruby to specific patch version? 
```suggestion
ruby '>= 3.1.4'
```

Gemfile
```diff
@@ -1,7 +1,7 @@
...
 
-gem 'cocoapods', '~> 1.12'
+gem 'cocoapods', '1.12.1'
```
legobeat, line 6: The specific version is already effectively enforced through `Gemfile.lock`. Is there a reason to pin specifically to `1.12.1` here or could we leave it at `1.12.x`?

scripts/setup.mjs
Cal-L: Probably better to save until used outside of macos. It'll save a bit of time from installation

Gudahtt: Maybe we could make it optional, using a flag?

Likely most of the time, non-macOS devs will prefer to skip this step because they aren't doing any iOS development. Lego I know you've been in situations recently where you'd have liked to run this step just to update `Podfile` / `Podfile.lock` but I think those cases are the exception rather than the norm.

Gemfile
Cal-L: Good point, reverted the changes. Also ran `bundle update` to bump to the latest versions and updated Gemfile.lock

scripts/setup.mjs
Cal-L: Since gems should only be used by Xcode, which is specific to MacOS, I've left the logic alone for skipping this step. @Gudahtt I did follow up with adding an optional flag `--skip-ios` for Android only development 👍

scripts/setup.mjs
```diff
@@ -2,14 +2,17 @@ import fs from 'fs';
 import { $ } from 'execa';
 import { Listr } from 'listr2';
 
+const IS_OSX = process.platform === 'darwin';
+// Skip iOS specific steps. Useful when only developing on Android only.
+const SKIP_IOS = process.argv.slice(2)?.[0] === '--skip-ios';
```
legobeat, line 7:
```suggestion
const SKIP_IOS = process.argv.slice(2)?.[0] === '--skip-ios';
```
I think we may want to add further flags to this script to allow disabling/disabling further steps. Maybe a general CLI api will have flags named differently and this would become legacy. Adding just a single other optional flag would also make this warrant something like `yargs`.

What do you say about making this an env var for now, which should be a bit more flexible in that sense?

```suggestion
const SKIP_IOS = process.env['MMBUILD_SKIP_IOS'];
```

legobeat: On the other hand, what about this?
- https://github.com/MetaMask/metamask-mobile/pull/9561

Cal-L: Agreed, your env var scales better 👍 . The PR looks good as well, just left a comment

legobeat (dismissed):

legobeat (approved):

Cal-L:
E2E passed on last non readme change

Daniel-Cross (approved):
Looks good!

Gudahtt (approved):
LGTM!

tommasini (approved):

PR #9208, opened 2024-04-11
feat: enable basic functionality on onboarding & settings

frankvonhoven:
## **Description**

This adds the MVP of the Basic Functionality feature

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2211

## **Manual testing steps**
(Full testing steps from QA available in ticket above)

* #### Create a new Wallet
  a. Skip SRP
  - [ ] Should go to Onboarding Success Screen --> Default Settings --> Disable Basic Functionality

* #### Create a new Wallet
  a. Complete SRP
  - [ ] Should go to Onboarding Success Screen --> Default Settings --> Disable Basic Functionality

* #### Import a Wallet
  - [ ] Should go to Onboarding Success Screen --> Default Settings --> Disable Basic Functionality

* #### Settings Screen
  a. Open app to Wallet view
  b. Go to Settings > Privacy
  c. Scroll down to "General" > Basic functionality
  - [ ] Should be able to toggle Basic functionality
  - [ ] If OFF, should see Banner Alert on Wallet View

### **New Wallet + SRP**

https://github.com/MetaMask/metamask-mobile/assets/141057783/3c7e907e-c3f6-4596-adf9-905e6a870cd9

### **New Wallet + Skip SRP**

https://github.com/MetaMask/metamask-mobile/assets/141057783/266f7dfe-0636-4bb7-ae60-4402fc810565

### **Import Wallet**

https://github.com/MetaMask/metamask-mobile/assets/141057783/bceb80cc-8eb8-4148-8f9a-fd6cd1c12b40

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've clearly explained what problem this PR is solving and how it is solved.
- [x] I've linked related issues
- [x] I've included manual testing steps
- [x] I've included screenshots/recordings if applicable
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.
- [x] I’ve properly set the pull request status:
  - [ ] In case it's not yet "ready for review", I've set it to "draft".
[...]

PR #9286, opened 2024-04-17
feat(swaps): enable Base for swaps

nikoferro:
## **Description**

This PR enables Base for swaps. Related to this, it also upgrades `@metamask/swaps-controller` to `v8.0.0`. See `package.json` and `yarn.lock` changes

## **Related issues**

Depends on `@metamask/swaps-controller` to `v8.0.0`

## **Manual testing steps**

1. Enable swaps on base feature flag
2. Change network to base
3. Get a quote for a Swap


legobeat (dismissed):

legobeat (dismissed):

legobeat (dismissed):
LGTM

tommasini:
Hey @nikoferro, awesome work! I see dedupe issues, can you solve those, please?

app/store/sagas/index.ts
```diff
@@ -15,6 +15,11 @@ import Engine from '../../core/Engine';
 import Logger from '../../util/Logger';
 import LockManagerService from '../../core/LockManagerService';
 import AppConstants from '../../../app/core/AppConstants';
+import { XMLHttpRequest as _XMLHttpRequest } from 'xhr2';
+
```
tommasini, line 19: I don't understand why we need to check if XMLHttpRequest is undefined, could you elaborate why is this needed on the sagas file?

package.json
```diff
@@ -337,6 +337,7 @@
     "uuid": "^8.3.2",
     "valid-url": "1.0.9",
     "vm-browserify": "1.1.2",
+    "xhr2": "^0.2.1",
```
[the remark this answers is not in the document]
nikoferro: done! 🎉

app/store/sagas/index.ts
nikoferro: hey @tommasini i think you saw this in slack, but for the sake of keeping the record here: As part of the upgrade, it seems we are removing (in v8) a no longer needed global polyfill for XMLHttpRequest. Everything was smooth until https://github.com/MetaMask/metamask-mobile/pull/9208 got merged, since it seems to rely on that same polyfill. This is a fix for this new issue generated by this upgrade

tommasini: Thanks for the explanation! 
I'm trying to have more context on this and learn something from it

Because what I see that it's happening here it is that we are changing the XMLHttpRequest global variable of react native, that uses a polyfill that needs to work with IOS and Android with one [node.js based](https://www.npmjs.com/package/xhr2), I don't want to block this PR, but at the same time I don't feel confident that I'm know the enough to know if this change is okay

I looked into the RN source code and I can't understand why this would be undefined.

What gives me more confidence is to see that the XMLHttpRequest is only used internally on test/debug folders
image

legobeat:
@nikoferro Could you run a `yarn deduplicate` and commit the results?

app/store/sagas/index.ts
Gudahtt: Hmm, this is confusing for me as well.

Could you help me understand how this relates to adding Base support, or to the swap controller update?

legobeat: #9208 introduced test code with this inadvertent implicit dependency on the old version of the swaps controller. So those references need to be addressed for the swaps controller to be upgraded after that.

legobeat:
> @nikoferro Could you run a `yarn deduplicate` and commit the results?

I went ahead and sorted this in 6f8d028e3a005f522456aaa8f546eba27d599314.

app/store/sagas/index.ts
nikoferro: @Gudahtt @legobeat `xhr2` was added in order to polyfill `XMLHttpRequest`. As @legobeat pointed out #9208 (an unrelated PR) depended on this. For some reason it seems the older `swaps-controller` version was exposing this as a global module. 

The `typeof` check was added because i cannot assign the value directly, since (probably due to a misconfiguration of the linter), the linter understands that `XMLHttpRequest` is already a global variable.

I could have started looking even deeper into the linter configuration on why this is happening but its leading to a never ending extending scope of this PR

legobeat (dismissed):
LGTM.

Perhaps `xhr2` should actually be removed alongside a refactor of those unrelated tests in a separate PR.

legobeat (dismissed):

legobeat (approved):

tommasini (approved):
Approving this PR now since the E2E are passing
We should indeed address this XMLHTMLREQUEST overwritten on a PR separated from this one

PR #9307, opened 2024-04-18
fix: Refactor/9083 logger class

### LINKED ISSUE #9083: Refactor `Logger` class to take only Error types
### What is this about?

We're currently seeing issues where unrelated errors are being grouped together, stack traces being lost, and errors pointing to the wrong files. Ex. https://metamask.sentry.io/auth/login/metamask/?next=%2Fissues%2F4699473996%2F Most of these can be attributed to the `Logger` class, which we are using to capture errors for Sentry. Specifically, as @Gudahtt mentions:

```
A stack trace is created when you construct an Error. But this Logger accepts an error string, and constructs the error on behalf of the caller. All Errors captured this way end up pointing back at the Logger class. So the first frame of every error captured this way is pointing at the wrong code.
What's worse is that the engines used by React Native (JSC and Hermes) do not support async stack traces, and this static logger uses an async logger internally. This effectively discards the remainder of the stack trace, leaving only the single frame pointing at the logger itself. So 100% of the stack trace is lost.
```

Slack thread discussing issue - https://consensys.slack.com/archives/C02U025CVU4/p1711575708407019

At a minimum, we need to update the `Logger` class to only accept Error types so that errors are reported correctly and stack traces are preserved. If the task is too large, we can deprecate what is currently used and create a new method that only accepts only `Error` types

### Scenario

- GIVEN A user is using the app
- WHEN A crash occurs, which uses the `Logger` class
- THEN An error should be reported to `Sentry`
- AND The error should be pointing to the correct file and contain the correct stack traces

### Design

_No response_

### Technical Details

- Update the `Logger` class to only accept `Error` types
- The `error` method needs to be updated to take only `Error` types
- Figure out way to still be able to check for metrics enabled (Create hook?)

### Threat Modeling Framework

_No response_

### Acceptance Criteria

Refer to scenario above

- [ ] Engineering (needed in most cases)
- [ ] Design
- [ ] Product
- [ ] Security
- [ ] Legal
- [ ] Marketing
- [ ] Other (please specify)

### References

_No response_

Cal-L:
## **Description**

This PR refactors the `Logger` class's `error` method to take `Error` object. While we updated the first parameter of the `error` method to be type `Error`, the business logic still handles string inputs to prevent breakage (in case there are strings still being passed). Furthermore, this PR converts places where strings were being passed into the `error` method to pass `Error` objects instead.

The reason for only allowing for `Error` objects as the input is because Sentry stack traces relies on the `Error` object being created at the origin of the error instead of being reconstructed in the `Logger` class. After these changes, we should see the correct stack trace for errors reported via `Logger` class.

## **Related issues**

Fixes: #9083 

## **Manual testing steps**

- I'd recommend setting up your own Sentry instance for testing
- Once own Sentry instance is setup, update the `sentry.properties` file with your project values
- Ensure `SENTRY_DISABLE_AUTO_UPLOAD` is "false"
- Hardcode an error anywhere in the app to call `Logger.error` and pass in `Error` object
- In the next line, hardcode an error anywhere in the app to call `Logger.error` and pass in `string` object
- Build app in release mode
- Trigger the errors
- Notice the two errors show in Sentry
- The error reported with a string will be missing the trace of the original error
- The error reported with an error will show the correct trace of the original error

### **Before**

Stack trace doesn't show origin of error since `Logger.error` was passed a `string`
image

### **After**

Stack trace shows origin of error after `Logger.error` is passed an `Error` object
image


app/util/test/testSetup.js
```diff
@@ -274,6 +274,7 @@ jest.mock('react-native-default-preference', () => ({
 // eslint-disable-next-line import/no-commonjs
 require('react-native-reanimated/lib/module/reanimated2/jestUtils').setUpTests();
 global.__reanimatedWorkletInit = jest.fn();
+global.__DEV__ = false;
```
Cal-L, line 277: Unit tests should run under non dev conditions

Cal-L:
E2E failure not related to this PR

wachunei (approved):
LGTM!

NicolasMassart (approved):
Looks good to me.
Please fill the PR description properly for ruture reference and ease of search and tracking:
- issue link
- manual testing (even if just running a unit test and checking it works)
- check boxes in checklist, just to show you considered each item

Cal-L:
> Looks good to me. Please fill the PR description properly for ruture reference and ease of search and tracking:
> 
> * issue link
> * manual testing (even if just running a unit test and checking it works)
> * check boxes in checklist, just to show you considered each item

Thanks for the reminder! Updated ticket description to match just that 👍

PR #9310, opened 2024-04-18
chore: Update yargs-parser version

NicolasMassart:
## **Description**

Force yargs-parser version to latest non vulnerable

## **Related issues**

Fixes https://github.com/MetaMask/mobile-planning/issues/1616
Fixes https://github.com/MetaMask/metamask-mobile/security/dependabot/44

## **Manual testing steps**

- yarn setup should succeed
- Smoke e2e should succeed as old yargs-parser was used by Appium

### **Before**

NA

### **After**

NA


package.json
```diff
@@ -122,6 +122,7 @@
     "react-native-ble-plx": "3.1.2",
     "**/babel-runtime/regenerator-runtime": "^0.13.8",
     "@metamask/transaction-controller/@metamask/controller-utils": "^5.0.0",
+    "yargs-parser": "~21.1.1",
```
legobeat, line 125:
```suggestion
    "yargs-parser": "^21.1.1",
```

We're already breaking majors by forcing v21 for packages expecting v11... Why not be a bit less strict in the pinning? ^^

legobeat:
Even if (to be verified) forcing all current references of `yargs-parser` onto a single version is safe, that may not be the case in the future.
Instead of adding a global resolution for `yargs-parser`, could we add explicit resolutions for each of the ones we currently want to enforce? E.g. `appium>dappium-tizen-driver>yargs>yargs-parser`

legobeat:
What is the blocker for upgrading to Appium v2? The legacy use comes through `appium@1.22.3` (2022-03-20), with `2.0.0` released 2023-07-05 and latest available `2.5.4`.

`appium@2.0.0` removes transitive dependency on abandoned [`appium-tizen-driver`](https://github.com/Samsung/appium-tizen-driver) beta.

NicolasMassart:
> What is the blocker for upgrading to Appium v2?

We would like to get rid of appium as soon as we can in favor of Detox, so I have to check how much work moving to Appium 2 requires but if it's too much maybe it's not worth it as this is just a dev dependency for e2e testing.

PR #9318, opened 2024-04-18
feat: Update Keyring Controller to v^13, Accounts Controller to v^11 and Preferences Controller v^8

tommasini:
## **Description**

Update preferences controller to v^8

 Keyring controller to v^13 ✅

 Accounts Controller to v^11✅
 Preferences controller v^8✅
 * Downgrade controller utils to interact with linea sepolia, only function used by preferences controller of controller-utils is `toChecksumHexAddress` and the function is the same on[ v^5](https://github.com/MetaMask/core/blob/add71d194d75a3c720c935f7947d06ee950bd7c7/packages/controller-utils/src/util.ts#L255) and on [v^8](https://github.com/MetaMask/core/blob/a24869b8e7e518f6682058d3eda4c41f0317ccc8/packages/controller-utils/src/util.ts#L263C17-L263C37)
 * Updated flow of create account and import account to update the `ecnryptionLib` that way on the future we can create or import and we will not run the flow  `[recreateVaultWithSamePassword](https://github.com/MetaMask/metamask-mobile/blob/fcfa9d1bd03e16ba18f3e9919b85a854614c2377/app/core/Authentication/Authentication.ts#L110)`, since it was running unnecessarily
 * Updated controllers initialization that had preferencesOnStateChange argument on their constructor (Mostly assets-controllers and PPOM), to have the address checksummed, on this current version PreferencesController starts to listen to KeyringController and that address is now on lower case format, which means that we need to update to a checksummed format for those specific scenarios for the assets controllers keep their data and tokens never disappear from the user for missmatching user address format
* Updated keyring controller patch with the same as extension to prevent the HD wallet disappear, please check the details on this core branch: extension-keyring-controller-v13-patch

## **Related issues**

* Fixes https://github.com/MetaMask/mobile-planning/issues/1689
* Fixes https://github.com/MetaMask/mobile-planning/issues/1697
* Fixes https://github.com/MetaMask/mobile-planning/issues/1693

## **Manual testing steps**

1. Create new wallet
2. Import existing wallet
3. Import private key
4. Upgrade testing 

## **Screenshots/Recordings**

Upgrade from main with Binance Smart Chain Selected, imported account with private key and tokens imported (Multiple chains added, and tokens imported on them also NFT imported on mainnet)

https://github.com/MetaMask/metamask-mobile/assets/46944231/ba6da46f-e923-45c8-b940-53044bf912e9


tommasini:
I don't understand why is this file on the changes, for future reviewers let me know if you feel this is a blocker for approve this PR
image

app/core/Engine.ts
```diff
@@ -378,18 +381,26 @@ class Engine {
...
-    const preferencesController = new PreferencesController(
-      {},
-      {
+    const preferencesController = new PreferencesController({
+      messenger: this.controllerMessenger.getRestricted<
+        'PreferencesController',
+        never,
+        'KeyringController:stateChange'
+      >({
+        name: 'PreferencesController',
```
Gedoxa, line 390: cba14cf fix typescript issue for misaligned versions of preferences controller on engine and introduce the key for prefereces state change event on EngineService since now Preferences Controller extends BaseController v2

Gedoxa (approved):

scripts/ios/bundle-js-and-sentry-upload.sh
Gudahtt, line 1: Better to revert this if it's not something you intended to change here

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -111,7 +111,8 @@ export class BackgroundBridge extends EventEmitter {
       AppConstants.NETWORK_STATE_CHANGE_EVENT,
       this.sendStateUpdate,
     );
-    Engine.context.PreferencesController.subscribe(this.sendStateUpdate);
+    // eslint-disable-next-line no-console
+    console.log('ENTER this.sendStateUpdate', this.sendStateUpdate);
```
Gudahtt, line 115: Leftover console log here:
```suggestion
```

app/components/Views/ChoosePassword/index.js
```diff
@@ -494,18 +487,6 @@ class ChoosePassword extends PureComponent {
...
     }
-
-    // Reset preferencesControllerState
-    preferencesControllerState = PreferencesController.state;
-
-    // Set preferencesControllerState again
-    await PreferencesController.update(preferencesControllerState);
-    // Reselect previous selected account if still available
-    if (hdKeyring.accounts.includes(selectedAddress)) {
-      Engine.setSelectedAddress(selectedAddress);
```
Gudahtt, line 505: Hmm, do wel want to set the selected address in this case? The above process would reset the selected address to the first account

scripts/ios/bundle-js-and-sentry-upload.sh
tommasini: I don't intent this to be here, but tbh I tried to revert it and I was not being able to, will retry it!

app/components/Views/ChoosePassword/index.js
tommasini: I tested it, and it doesn't seem to me that it was needed, because this ChoosePassword screen is used when [importing an account via SRP or creating a new one](https://github.com/MetaMask/metamask-mobile/blob/c11dd2ec00946c3c56aa2c99e12aa3cfbeff3b5d/app/components/Nav/App/index.js#L180) or [backing up a recently created one](https://github.com/MetaMask/metamask-mobile/blob/d2fb83fdf4a8bdd35332ccb808cbfa271c393608/app/components/Nav/Main/MainNavigator.js#L621)! So it seems unnecessary because the only of this flows that will have 2 accounts is if we have an account imported after creating a new one that it's not backed up.

It don't know why ChoosePassword is on the [setPasswordFlow](https://github.com/MetaMask/metamask-mobile/blob/d2fb83fdf4a8bdd35332ccb808cbfa271c393608/app/components/Nav/Main/MainNavigator.js#L617), because it seems that it we are never navigating to it. The setPasswordFlow navigator is only used when the user tries to back up after the creation of the account (with the alert or the modal to remind to backup)

To confirm my thoughts I recorded the following test scenario
* Create a new account without backing up
* Import wallet with private key
* Backup wallet with imported account selected
* After backup the selected account should be the imported one

https://github.com/MetaMask/metamask-mobile/assets/46944231/ae75b4f4-02e3-4185-9fbc-e6ea4b3b6874

Gudahtt: Hmm OK. Well, worst case if this does introduce a regression where the selected account switches unexpectedly, this is a very minor problem! Removing this sounds good to me.

app/core/Authentication/Authentication.ts
```diff
@@ -351,6 +351,10 @@ class AuthenticationService {
       await this.storePassword(password, authData?.currentAuthType);
       await AsyncStorage.setItem(EXISTING_USER, TRUE);
       await AsyncStorage.removeItem(SEED_PHRASE_HINTS);
+      const encryptionLib = await AsyncStorage.getItem(ENCRYPTION_LIB);
+      if (encryptionLib !== ORIGINAL) {
```
Gudahtt, line 355: What is the purpose of `ENCRYPTION_LIB`? I thought it was to trigger the vault to be re-encrypted if a user was updating from an older version of the wallet.

If we're overwriting it here every time, it would not serve that purpose anymore. To prevent the re-encryption on login for new users, we'd want to only set this if it's _empty_.

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -111,7 +111,6 @@ export class BackgroundBridge extends EventEmitter {
       AppConstants.NETWORK_STATE_CHANGE_EVENT,
       this.sendStateUpdate,
     );
-    Engine.context.PreferencesController.subscribe(this.sendStateUpdate);
```
Gudahtt, line 114: It looks like this subscription was meant to update the provider state for dapps. I believe this is still something we want to do. Was there some other reason this was removed?

app/core/Engine.ts
```diff
@@ -378,18 +382,26 @@ class Engine {
...
+        allowedEvents: ['KeyringController:stateChange'],
+      }),
+      state: {
         ipfsGateway: AppConstants.IPFS_DEFAULT_GATEWAY_URL,
         useTokenDetection:
           initialState?.PreferencesController?.useTokenDetection ?? true,
         // TODO: Use previous value when preferences UI is available
```
Gudahtt, line 398: Hmm, what does this comment mean? Does the change you're making here (by adding `...initialState.PreferencesController`) fulfill this TODO?

I am not sure what this "Preferences UI" is that is referred to here

app/core/Engine.ts
```diff
@@ -419,7 +431,20 @@ class Engine {
 
     const assetsContractController = new AssetsContractController({
       onPreferencesStateChange: (listener) =>
-        preferencesController.subscribe(listener),
+        this.controllerMessenger.subscribe(
+          `AccountsController:selectedAccountChange`,
```
Gudahtt, line 435: Why are we listening to `AccountsController` state changes here rather than `PreferencesController`?

app/store/migrations/040.ts
```diff
@@ -0,0 +1,16 @@
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import {
+  ENCRYPTION_LIB,
+  EXISTING_USER,
+  ORIGINAL,
+} from '../../constants/storage';
+
+export default async function migrate(state: unknown) {
```
Gudahtt, line 8: Nit: We should include a comment with all migrations to explain the purpose

Gudahtt: It seems that there are no unit tests yet for this either

patches/@metamask+preferences-controller+8.0.0.patch
```diff
@@ -0,0 +1,166 @@
...
+\ No newline at end of file
+diff --git a/node_modules/@metamask/preferences-controller/dist/index.js b/node_modules/@metamask/preferences-controller/dist/index.js
+index e953704..4410655 100644
+--- a/node_modules/@metamask/preferences-controller/dist/index.js
++++ b/node_modules/@metamask/preferences-controller/dist/index.js
+@@ -15,4 +15,5 @@ var __exportStar = (this && this.__exportStar) || function(m, exports) {
+ };
+ Object.defineProperty(exports, "__esModule", { value: true });
+ __exportStar(require("./PreferencesController"), exports);
++__exportStar(require("./constants"), exports);
```
Gudahtt, line 164: Hmm, this looks like a new addition to this patch. What is the reason for exposing this? We don't export this on `core` at the moment.

app/core/Engine.ts
tommasini: Hmm I'm not sure, but good point let me verify!

patches/@metamask+preferences-controller+8.0.0.patch
tommasini: Ahhh amazing catch, this is not needed anymore, I forgot to revert it!

app/core/Engine.ts
tommasini: I was getting that subscribe was deprecated, and the goal was to have selectedAddress in sync with the selectedAccount of Accounts Controller, since `onPreferencesChange` property was mandatory for the constructors of assets controllers and PPOM

image

app/core/BackgroundBridge/BackgroundBridge.js
tommasini: Hmmm, the reason was because PreferencesController.subscribe was deprecated, let me debug this deeper, the accounts on the dapps seemed to be changing fine after this removal, but let me comeback to this.
image

Gudahtt: Oh I see - so with that release, we should be using the controller messenger instead of that subscribe function.

State change subscriptions are not deprecated, the API has just changed

app/store/migrations/040.ts
tommasini: This migraiton will be removed

app/core/Authentication/Authentication.ts
tommasini, line 355:
```suggestion
      if (!encryptionLib) {
```

app/core/Engine.ts
tommasini, line 435:
```suggestion
          `PreferencesController:stateChange`,
```

app/core/BackgroundBridge/BackgroundBridge.js
tommasini: Changed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/82ac01d04f861546e6f00aa89e724ee3de2820e8)

patches/@metamask+preferences-controller+8.0.0.patch
tommasini: This was reverted on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/82ac01d04f861546e6f00aa89e724ee3de2820e8)

app/store/migrations/040.ts
tommasini: Removed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/82ac01d04f861546e6f00aa89e724ee3de2820e8)

app/components/Views/ChoosePassword/index.js
```diff
@@ -433,7 +428,7 @@ class ChoosePassword extends PureComponent {
    * @param password - Password to recreate and set the vault with
    */
   recreateVault = async (password, authType) => {
```
Cal-L, line 430: TODO in follow up PRs:
This code is most likely not needed since ChoosePassword is only used within the onboarding stack. Another question is - Do we need to recreate the vault in the event where wallet creation fails?

app/core/BackgroundBridge/BackgroundBridge.js
tommasini: Re added with the new API change on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/34f0d8414866a876e77a85ae28c5a4715d581674)

And tested: 

https://github.com/MetaMask/metamask-mobile/assets/46944231/64e846b8-5929-48fb-b4bd-db7f4929a4b8

app/core/Engine.ts
```diff
@@ -378,18 +382,26 @@ class Engine {
...
+      >({
+        name: 'PreferencesController',
+        allowedEvents: ['KeyringController:stateChange'],
+      }),
+      state: {
         ipfsGateway: AppConstants.IPFS_DEFAULT_GATEWAY_URL,
         useTokenDetection:
           initialState?.PreferencesController?.useTokenDetection ?? true,
         // TODO: Use previous value when preferences UI is available
         useNftDetection: false,
```
Cal-L, line 398: TODO in future:
May be able to remove these in favor of spread in the future. However, the default PreferencesController values do not match the defaults that we specified here

app/core/Engine.ts
```diff
@@ -419,7 +431,21 @@ class Engine {
 
     const assetsContractController = new AssetsContractController({
       onPreferencesStateChange: (listener) =>
```
Cal-L, line 432: Can we create a reusable callback to be used in other controllers as well?

app/core/Engine.ts
tommasini: Yeah you are right! removed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/34f0d8414866a876e77a85ae28c5a4715d581674)

app/core/Engine.ts
```diff
@@ -1091,10 +1179,16 @@ class Engine {
...
               networkController.state.providerConfig.chainId;
+
+            const showIncomingTransactions =
+              preferencesController?.state?.showIncomingTransactions;
+
             return Boolean(
-              preferencesController?.state?.showIncomingTransactions?.[
-                currentHexChainId
-              ],
+              Object.keys(showIncomingTransactions).includes(currentHexChainId)
```
Cal-L, line 1186: Isn't this the same logic as before?

Cal-L (commented):
Left some comments

app/core/Engine.ts
tommasini: Yes it is! But this way typescript doesn't complain because we can typecast it with the `ShowIncomingTransactionKey `

owencraston (dismissed):
I logged the state of the preferences controller and accounts controller and went through some basic flows of adding new account, importing an account with funds and everything looked good.

app/store/migrations/039.ts
```diff
@@ -0,0 +1,16 @@
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import {
+  ENCRYPTION_LIB,
+  EXISTING_USER,
+  ORIGINAL,
+} from '../../constants/storage';
+
+export default async function migrate(state: unknown) {
```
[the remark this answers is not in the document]
tommasini: After talk about this we will bring this question for internal discussion and to be addressed on a future PR

app/core/Authentication/Authentication.ts
```diff
@@ -351,6 +351,10 @@ class AuthenticationService {
       await this.storePassword(password, authData?.currentAuthType);
       await AsyncStorage.setItem(EXISTING_USER, TRUE);
       await AsyncStorage.removeItem(SEED_PHRASE_HINTS);
+      const encryptionLib = await AsyncStorage.getItem(ENCRYPTION_LIB);
```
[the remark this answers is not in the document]
tommasini: Addressed on this [PR](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/ae46654bcc12f3f2dd5e52ee99a5a975db79185a)

app/core/Authentication/Authentication.ts
```diff
@@ -105,7 +112,7 @@ class AuthenticationService {
     await KeyringController.submitPassword(password);
     const encryptionLib = await AsyncStorage.getItem(ENCRYPTION_LIB);
     const existingUser = await AsyncStorage.getItem(EXISTING_USER);
-    if (encryptionLib !== ORIGINAL && existingUser) {
+    if (!encryptionLib && existingUser) {
```
Gudahtt, line 115: Hmm, shouldn't this still be `encryptionLib !== ORIGINAL` ?

```suggestion
    if (encryptionLib !== ORIGINAL && existingUser) {
```

We want to recreate the vault if the encryption library used was anything other than the one we expect, which is `ORIGINAL`.

Though I know we aren't aware of any cases where this is set to any other value, I didn't go digging in the git history to search for such examples, so leaving this as-is seemed safer

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -325,7 +329,7 @@ export class BackgroundBridge extends EventEmitter {
       AppConstants.NETWORK_STATE_CHANGE_EVENT,
       this.sendStateUpdate,
     );
-    Engine.context.PreferencesController.unsubscribe(this.sendStateUpdate);
```
Gudahtt, line 328: I think we want a a `Engine.controllerMessenger.unsubscribe` call here to replace this, so that we don't get double-notifications after a disconnect

app/core/Engine.ts
```diff
@@ -419,7 +430,21 @@ class Engine {
...
+        this.controllerMessenger.subscribe(
+          //@ts-expect-error assets-controllers need to be bump to v^26
+          //assets-controllers preferences controller version is misaligned with the app repo (v8 != v4)
+          `${preferencesController.name}:stateChange`,
+          (preferencesControllerState: PreferencesState) => {
+            //@ts-expect-error assets-controllers need to be bump to v^26
+            //assets-controllers preferences controller version is misaligned with the app repo (v8 != v4)
+            listener({
+              ...preferencesControllerState,
+              selectedAddress: toChecksumHexAddress(
```
Gudahtt, line 442: Hmm, why is this step needed? The `selectedAddress` should already be checksummed

app/core/Engine.ts
```diff
@@ -419,7 +430,21 @@ class Engine {
 
     const assetsContractController = new AssetsContractController({
       onPreferencesStateChange: (listener) =>
-        preferencesController.subscribe(listener),
+        this.controllerMessenger.subscribe(
+          //@ts-expect-error assets-controllers need to be bump to v^26
+          //assets-controllers preferences controller version is misaligned with the app repo (v8 != v4)
```
Gudahtt, line 435: The type error here is not caused by an incompatibility between the assets-controllers package and the preferences-controller package. The type error is on the global controller messenger, because it needs to be updated to include the types from the preferences controller package.

The types that need updating are `GlobalActions` and `GlobalEvents` - they should include all actions and events from all controllers. This update is bringing in new actions/events from the PreferencesController

app/store/migrations/025.js
```diff
@@ -1,4 +1,4 @@
-import { ETHERSCAN_SUPPORTED_CHAIN_IDS } from '@metamask/preferences-controller';
+import { ETHERSCAN_SUPPORTED_CHAIN_IDS } from '@metamask/preferences-controller/dist/constants';
```
Gudahtt, line 1: Hmm, so we do need this constant exported after all.

We should follow up on `core` to ensure this is accessible directly from these packages, as these file references may not work with later core versions. Since the other PreferencesController constants are only used internally, maybe it would make sense to move this constant to an already-public set of constants, like the controller-utils constants. Or we could export just this one constant.

tommasini: Yeah, I noticed, when the unit tests failed after removing it!

I created this [PR](https://github.com/MetaMask/core/pull/4233) on Core to follow up

app/core/Engine.ts
tommasini: Yeah you are right, it's not!
Since the addresses on identities of PreferencesController are being converted [here](https://github.com/MetaMask/core/blob/a24869b8e7e518f6682058d3eda4c41f0317ccc8/packages/preferences-controller/src/PreferencesController.ts#L326) to checksummed
And when we call [setSelectedAddress](https://github.com/MetaMask/core/blob/a24869b8e7e518f6682058d3eda4c41f0317ccc8/packages/preferences-controller/src/PreferencesController.ts#L359) as well, and it seems that are the places where selected address ios being changed!

Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/2cf05a6eb5308b0f111a417b087af5fd5244cb47)

app/core/BackgroundBridge/BackgroundBridge.js
tommasini: Yeah, great catch as always!
Addressed [here](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/2cf05a6eb5308b0f111a417b087af5fd5244cb47)

app/core/Engine.ts
tommasini: Thanks Mark! Followed with this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/1e071834d3b90c8a613c1ca67860d1d4e57551fd)
Also thanks for the explanation on this!

app/core/Engine.ts
```diff
@@ -247,11 +248,15 @@ type GlobalEvents =
   | SignatureControllerEvents
   | KeyringControllerEvents
   | PPOMControllerEvents
-  | AccountsControllerEvents;
+  | AccountsControllerEvents
+  | PreferencesControllerEvents;
```
Gudahtt, line 252: Could you add `PreferencesControllerEvents` to `GlobalActions` as well? All actions and events must be listed

app/core/Engine.ts
```diff
@@ -419,7 +431,15 @@ class Engine {
 
     const assetsContractController = new AssetsContractController({
       onPreferencesStateChange: (listener) =>
-        preferencesController.subscribe(listener),
+        this.controllerMessenger.subscribe(
+          `${preferencesController.name}:stateChange`,
+          (preferencesControllerState: PreferencesState) => {
+            listener({
+              ...preferencesControllerState,
+              selectedAddress: preferencesControllerState.selectedAddress,
```
Gudahtt, line 439: This line appears to be redundant

app/core/Engine.ts
tommasini: Added here: https://github.com/MetaMask/metamask-mobile/pull/9318/commits/c1daadff6c6f4d83575aec1923a9006596d3b7d9

app/core/Engine.ts
tommasini: Ah, yeah, it's true, is redundant everywhere, we are already spreading the preferencesControllerState
Cleaned [here](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/03edd382566e07b85982915cd2f701cf1a4cd46e)

app/core/Engine.ts
```diff
@@ -509,9 +541,17 @@ class Engine {
...
-        preferencesController.subscribe(listener),
+        this.controllerMessenger.subscribe(
+          `${preferencesController.name}:stateChange`,
+          (preferencesControllerState: PreferencesState) => {
+            //@ts-expect-error assets-controllers need to be bump to v^26
+            //assets-controllers preferences controller version is misaligned with the app repo (v8 != v4)
+            listener({
+              ...preferencesControllerState,
+            });
+          },
```
Gudahtt, line 553: Nit: I don't think we need to shallow clone the state here either. Something like this might be simpler:

```suggestion
          //@ts-expect-error assets-controllers need to be bump to v^26
          //assets-controllers preferences controller version is misaligned with the app repo (v8 != v4)
          listener,
```

Gudahtt (dismissed):
LGTM!

owencraston (dismissed):

app/core/Engine.ts
```diff
@@ -1091,10 +1118,16 @@ class Engine {
...
               networkController.state.providerConfig.chainId;
+
+            const showIncomingTransactions =
+              preferencesController?.state?.showIncomingTransactions;
+
             return Boolean(
-              preferencesController?.state?.showIncomingTransactions?.[
-                currentHexChainId
-              ],
+              Object.keys(showIncomingTransactions).includes(currentHexChainId)
```
Gudahtt, line 1126: I don't think we need the `ShowIncomingTransactionKey` type, or any cast. There is a TypeScript error here because TypeScript doesn't like us to directly reference object properties that may not exist. But we have another type-safe way of doing his: `hasProperty`.

e.g. we could return something like this instead that is type safe and requires no extra assertions or types:

```
return (
              hasProperty(
                preferencesController?.state?.showIncomingTransactions,
                currentChainId,
              ) &&
              Boolean(
                preferencesController?.state?.showIncomingTransactions?.[
                  currentHexChainId
                ],
              )
            );
```

Gudahtt (dismissed):
LGTM! Just one remaining non-blocking suggestion

app/core/Engine.ts
tommasini: Amazing, addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/28265f035e763523ee6ebc85e32707bf070becde)

app/core/Engine.ts
tommasini: Amazing result! Amazing callout <3 

Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/28265f035e763523ee6ebc85e32707bf070becde)

Cal-L (dismissed):
LGTM! Great work

app/core/Engine.ts
```diff
@@ -356,6 +360,22 @@ class Engine {
...
+    /**
+     * Subscribes a listener to the state change events of Preferences Controller.
+     *
+     * @param listener - The callback function to execute when the state changes.
+     */
+    const onPreferencesStateChange = (
+      // preferencesState is typed as any because if not will give an ts error on every assets-controllers using this function
+      // because assets-controllers preferences controller version is misaligned with the app repo (v8 != v4)
+      // and because of that preferencesState would be two different objects
+      listener: (preferencesState: any) => void,
```
Gudahtt, line 372: Please use `ts-expect-error` here rather than `any`. `any` is much more dangerous and impacts a much wider area of the codebase.

See here for more information: https://github.com/MetaMask/contributor-docs/blob/1809b07f00ebd6cf8383732dafb6f3cb9a8d1f77/docs/typescript.md#avoid-any

tommasini: You are right Mark, this way when we upgrade assets controllers, typescript will flag for us to remove the ts-expect-error!
Thanks!!
Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9318/commits/869cd172851ec9304f86d998cffbe20420477c29)

Gudahtt (approved):

owencraston (approved):

Cal-L (approved):
lgtm

plasmacorral:
Ran migration tests with a QA build of v7.18.0 (1268) to a QA build of commit 869cd172851ec9304f86d998cffbe20420477c29 and noted some accounts that were lost:  Have before and after state logs available on 3 unique devices.  All three devices lost the Ledger account, and 2 of 3 devices(both android and iOS) lost multiple QR accounts.

Daniel-Cross (approved):

MarioAslau:
This PR merge was reverted, branch was restored and should not be deleted yet !

gantunesr:
The changes are now part of this PR https://github.com/MetaMask/metamask-mobile/pull/9508

PR #9258, opened 2024-04-16
feat: notifications settings UI - [6 of 10]

Jonathansoufer:
## **Description**

This PR implements Notifications settings and wizard opt in UI.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to Settings page and click on Notifications
2. Enable Notifications settings.
3.

### **After**

https://github.com/MetaMask/metamask-mobile/assets/44679989/a1ecc2c0-ba65-4866-a902-1efc77ecb58a

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [x] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [x] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #9346, opened 2024-04-22
feat: notifications LIST screen UI - [9 of 10]

Jonathansoufer:
## **Description**

This PR implements a List screen for Notifications following designs from [Figma](https://www.figma.com/file/XcAVSI5z4myb1DtlbIHw7U/ProfileID-(UUID)-mobile?type=design&node-id=906-14213&mode=design&t=QXtJD7nofUBR9M7b-0)

This screen as all notification related code will be behind feature flag until ALL the code being merged. This PR is part of a series aiming to make the review process easier.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Enable Notifications from NavBar Bell icon or Settings --> Notifications -> Enable
2. Click on the NavBar Bell icon again
3.

### **After**

https://github.com/MetaMask/metamask-mobile/assets/44679989/75873986-66bf-4f2e-bed4-980a911c7eb9


app/components/Nav/Main/MainNavigator.js
```diff
@@ -700,9 +706,9 @@ const MainNavigator = () => (
     />
     <Stack.Screen name="AddBookmarkView" component={AddBookmarkView} />
     <Stack.Screen name="OfflineModeView" component={OfflineModeView} />
+    <Stack.Screen name="NotificationsView" component={NotificationsModeView} />
```
tommasini, line 709: Can we use the constant instead?
```suggestion
    <Stack.Screen name={Routes.NOTIFICATIONS.VIEW} component={NotificationsModeView} />
```

locales/languages/en.json
```diff
@@ -374,7 +374,7 @@
     "no_available_tokens": "Don't see your token?",
     "add_tokens": "Import tokens",
     "are_you_sure_exit": "Are you sure you want to exit?",
-    "search_information_not_saved":"Your search information will not be saved.",
+    "search_information_not_saved": "Your search information will not be saved.",
```
tommasini, line 377: Can we submit to main these translations on a different PR? To make sure that the translations are requested before the next release cut?

tommasini (commented):
This PR looks amazing, just left some comment, and if you firstly could submit a new PR for translations and then address the others it would be awesome.
I'm requesting this because translations need some time to be handled by other team and between reviews and changes we have a bigger change that they are handled before the next release cut

locales/languages/en.json
Jonathansoufer: All changes NOT related to notifications are probably some `linting` since I didn't touched anything but this. Cherry picking the translation just for that wouldn't be necessary in my humble opinion, don't you agree?

app/components/UI/Notification/Empty/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,65 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Empty should render correctly 1`] = `
+<View
+  style={
+    Object {
```
legobeat, line 6: As of the upgrade to Jest v29, snapshots will need to be regenerated (or, this worked fine for me: `rg -l -g '*.snap' 'Object\s*\{|Array\s*\[' | xargs sed -i -e 's#Object\s*{#{#g' -e 's#Array\s*\[#[#g'`

app/components/UI/Notification/Row/Content.tsx
```diff
@@ -0,0 +1,72 @@
...
+  createdAt,
+  value,
+  styles,
+}: NotificationContentProps) {
+  const useRenderTwoPartsTitles = useCallback(() => {
+    const parts = title?.toLowerCase().includes('sent')
+      ? title?.split(/(.*?to)/g)
+      : title?.toLowerCase().includes('received')
+      ? title?.split(/(.*?from)/g)
+      : [];
```
NicolasMassart, line 34: You could safeguard this part by first testing title is defined and then rely on this without the redundant optional chaining.
Returning directly in case of undefined or null title.
Also, the ternary here is not the most readable structure. Try a classic if/else, it sometimes makes it more readable with the exact same efficiency.

Finally, be careful, your regex are vulnerable to redos attack (as indficated by Sonar). This may be changed to more simple regex without backtracking or direct substring manipulation.

Example suggestion for all these changes:
```suggestion
if (!title) return <Text color={TextColor.Alternative} variant={TextVariant.BodySM}>{title}</Text>;

    const lowerTitle = title.toLowerCase();
    let parts;

    if (lowerTitle.includes('sent')) {
      parts = title.split(/(\bto\b)/g);
    } else if (lowerTitle.includes('received')) {
      parts = title.split(/(\bfrom\b)/g);
    } else {
      parts = [];
    }
```

app/components/UI/Notification/Row/Actions.tsx
```diff
@@ -0,0 +1,58 @@
...
+      return Linking.openURL(link.linkUrl);
+    }
+
+    if (action?.actionUrl) {
+      //  TODO: Handle deeplinks from notifications.
+      // const handledByDeeplink = SharedDeeplinkManager.parse(action.actionUrl, {
+      //   origin: AppConstants.DEEPLINKS.ORIGIN_NOTIFICATION,
+      //   onHandled: () => (navigation as any).pop(2),
+      // });
+    }
```
NicolasMassart, line 41: Remove this dead code and create an issue for adding the deeplink handling if is has to be provided in another PR.

app/components/UI/Notification/Row/Actions.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { Linking } from 'react-native';
+import Button, {
+  ButtonVariants,
+} from '../../../../component-library/components/Buttons/Button';
+import { IconName } from '../../../../component-library/components/Icons/Icon';
+//  TODO: Handle deeplinks from notifications.
+// import SharedDeeplinkManager from '../../../../core/DeeplinkManager/SharedDeeplinkManager';
+// import AppConstants from '../../../../core/AppConstants';
```
NicolasMassart, line 9: See Comment on lines +35 to +41

app/components/UI/Notification/Row/Actions.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    actionUrl: PREFIXES[ACTIONS.SEND],
+    isExternal: false,
+  };
+  const link = {
+    linkText: 'Learn more',
+    linkUrl: 'https://metamask.io',
+    isExternal: true,
+  };
+
+  it('renders correctly', () => {
```
NicolasMassart, line 27: Try to avoid this test name as it's not actually testing rendering is correct but only that it hasn't changed since last time.
```suggestion
  it('matches snapshot', () => {
```

app/components/UI/Notification/Row/Actions.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const { toJSON } = renderWithProvider(
+      <NotificationActions action={action} link={link} styles={styles} />,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('renders correctly with LINK only', () => {
```
NicolasMassart, line 34:
```suggestion
  it('matches snapshot with LINK only', () => {
```

app/components/UI/Notification/Row/Actions.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  it('renders link title correctly', () => {
+    const { getByText } = renderWithProvider(
+      <NotificationActions link={link} styles={styles} />,
+    );
+
+    expect(getByText('Learn more')).toBeTruthy();
+  });
+
+  it('Calls Linking.openURL when link CTA is clicked', async () => {
```
NicolasMassart, line 49:
```suggestion
  it('calls Linking.openURL when link CTA is clicked', async () => {
```

app/components/UI/Notification/Row/Actions.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('renders correctly with LINK only', () => {
+    const { toJSON } = renderWithProvider(
+      <NotificationActions link={link} styles={styles} />,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('renders link title correctly', () => {
```
NicolasMassart, line 41:
```suggestion
  it('renders link title', () => {
```

app/components/UI/Notification/List/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+  loading,
+}: NotificationsList) => {
+  const theme = useTheme();
+  const { colors } = theme;
+  const styles = createStyles(theme);
+  const { trackEvent } = useMetrics();
+
+  const onPress = useCallback(
+    (item) => {
+      //TODO: details will be implemented on a separete PR
```
NicolasMassart, line 44: This should be a PR comment instead of a code comment

app/components/UI/Notification/Empty/index.test.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Empty from './';
+import renderWithProvider from '../../../../util/test/renderWithProvider';
+
+describe('Empty', () => {
+  it('should render correctly', () => {
```
NicolasMassart, line 6:
```suggestion
  it('matches snapshot', () => {
```

app/components/UI/Notification/Row/Content.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const createdAt = '2024-04-26T16:35:03.147606Z';
+  const description = {
+    text: 'We are excited to announce the launch of our brand new website and app!',
+    asset: {
+      name: 'Ethereum',
+    },
+  };
+  const value = '100.00';
+
+  it('renders correctly', () => {
```
NicolasMassart, line 22:
```suggestion
  it('matches snapshot', () => {
```

app/components/UI/Notification/Row/Icon.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const walletNotification = {
+    notificationType: TRIGGER_TYPES.ERC20_SENT,
+    badgeIcon: IconName.Send2,
+    imageUrl:
+      'https://github.com/MetaMask/brand-resources/blob/master/SVG/SVG_MetaMask_Icon_Color.svg',
+  };
+
+  it('renders correctly with Wallet type', () => {
```
NicolasMassart, line 47:
```suggestion
  it('matches snapshot with Wallet type', () => {
```

app/components/UI/Notification/Row/Icon.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+        imageUrl={walletNotification.imageUrl}
+        badgeIcon={walletNotification.badgeIcon}
+        styles={styles}
+      />,
+      { state: mockInitialState },
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('renders correctly with Announcement type', () => {
```
NicolasMassart, line 60:
```suggestion
  it('matches snapshot with Announcement type', () => {
```

app/components/UI/Notification/Row/Root.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+const children = <></>;
+const styles = {
+  wrapper: {},
+  trashIconContainer: {},
+};
+const onDismiss = jest.fn();
+const handleOnPress = jest.fn();
+
+describe('NotificationRoot', () => {
+  it('renders correctly', () => {
```
NicolasMassart, line 16:
```suggestion
  it('matches snapshot', () => {
```

app/components/Views/Notifications/index.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+describe('NotificationsView', () => {
+  let navigation: NavigationProp<ParamListBase>;
+
+  beforeEach(() => {
+    navigation = {
+      navigate: jest.fn(),
+    } as unknown as NavigationProp<ParamListBase>;
+  });
+
+  it('should render correctly', () => {
```
NicolasMassart, line 33:
```suggestion
  it('matches snapshot', () => {
```

app/components/Views/Notifications/index.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+        notifications={MOCK_NOTIFICATIONS}
+      />,
+      {
+        state: mockInitialState,
+      },
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should render correctly with no notifications', () => {
```
NicolasMassart, line 47:
```suggestion
  it('matches snapshot without notifications', () => {
```

NicolasMassart (changes requested):
Nice overall work!
I added some suggestions on tests naming and implementation to match guidelines and good practices that may have CI stability and  security implications.

app/components/UI/Notification/Row/Icon.tsx
```diff
@@ -0,0 +1,79 @@
...
+      style={styles.badgeWrapper}
+    >
+      {notificationType.toLowerCase().includes('eth') ? (
+        <NetworkMainAssetLogo style={styles.ethLogo} />
+      ) : (
+        <RemoteImage
+          source={{
+            uri:
+              imageUrl ||
+              'https://token.api.cx.metamask.io/assets/nativeCurrencyLogos/ethereum.svg',
```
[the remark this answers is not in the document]
Jonathansoufer: Fixed.

app/components/UI/Notification/Row/Actions.tsx
Jonathansoufer: Fixed.

app/components/UI/Notification/Row/Actions.tsx
Jonathansoufer: Fixed.

app/components/UI/Notification/Row/Content.tsx
Jonathansoufer: Good approach, I implemented a mix of it.

app/components/UI/Notification/Row/Icon.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    nftPlaceholder: {},
+    assetLogo: {},
+    assetPlaceholder: {},
+    badgeWrapper: {},
+    ethLogo: {},
+  };
+
+  const announcementNotification = {
+    notificationType: TRIGGER_TYPES.FEATURES_ANNOUNCEMENT,
+    imageUrl:
```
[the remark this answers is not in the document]
Jonathansoufer: Fixed.

app/components/UI/Notification/Row/Icon.tsx
```diff
@@ -0,0 +1,78 @@
...
+  imageUrl?: string;
+}
+function NotificationIcon({
+  notificationType,
+  styles,
+  badgeIcon,
+  imageUrl,
+}: NotificationIconProps) {
+  const customStyles = () => {
+    if (notificationType.toLowerCase().includes('erc721' || 'erc1155')) {
```
NicolasMassart, line 30: As indicated by Sonar, this can't work: in JS, the || operator always return the first true item. And because items here are non empty strings, they are always considered as true. So the first one 'erc721' is always returned and the second one 'erc1155' is never considered in the includes. One way to fix that would be to match against `some` on an array (at least one element in the array passes) of notification types like the following:
```suggestion
    const types = ['erc721', 'erc1155'];
    if (types.some(type => notificationType.toLowerCase().includes(type))) {
```
This is not the only way of course. Feel free to adjust.

NicolasMassart (dismissed):

app/components/UI/Notification/Row/Icon.tsx
Jonathansoufer: Good catch. Fixed.

app/core/Analytics/MetaMetrics.events.ts
```diff
@@ -1401,6 +1410,23 @@ const legacyMetaMetricsEvents = {
...
+  WALLET_NOTIFICATIONS: generateOpt(
+    EVENT_NAME.WALLET_NOTIFICATIONS,
+    ACTIONS.SELECTS_WALLET_NOTIFICATIONS,
+    DESCRIPTION.NOTIFICATIONS,
+  ),
+  ANNOUCEMENTS_NOTIFICATIONS: generateOpt(
+    EVENT_NAME.ANNOUCEMENTS_NOTIFICATIONS,
+    ACTIONS.SELECTS_ANNOUCEMENTS_NOTIFICATIONS,
+    DESCRIPTION.NOTIFICATIONS,
+  ),
```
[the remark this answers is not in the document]
Jonathansoufer: These are used to track which tab the user are when selecting notifications list to see, but I moved them out of legacy to the new events list.

Cal-L:
Regarding PR description

- I can see in current prod that the notification button is behind a feature flag. Does this mean that users shouldn't be able to interact with the changes once it's merged?
- Can you attach a Figma link (useful for reviewers to have a point of reference)?
- Usually a good way to split the description is by sectioning out by the screen/components. Ex:
  - Turn on notifications page -> new page + explanation of functionality
  - Notifications toggle page -> new page + explanation of functionality
      - How each toggle affects the behavior of the app based on toggle state
  - Notifications list page -> new page + explanation of functionality
      - All -> shows all notification types - What are all of the possible notification types?
      - Wallet -> shows wallet based notifications -> What are all of the possible notification types here?
      - Announcements -> shows announcement notifications -> What are all of the possible notification types here?

Cal-L (changes requested):
Turn on notifications page's Learn more link needs a space to the left of itself

Cal-L (changes requested):
Turn on notifications page's Learn more link needs a space to the left of itself

Cal-L (commented):
The Notifications toggle page should probably have an X on the top right since it isn't mean to go back. The dismiss behavior should be a slide down (current slides diagonally)

Cal-L (changes requested):
The Notifications toggle page should probably have an X on the top right since it isn't mean to go back. The dismiss behavior should be a slide down (current slides diagonally)

Cal-L (changes requested):
Are the sub toggles from the Notifications page functional yet and if so, can we provide videos/description on app behavior when those toggles are on/off

Cal-L (changes requested):
The Allow notifications toggle should remain off until a user taps Yes (currently, it slides right but back to the left)

Cal-L (changes requested):
The Notifications list page seems to also have extra padding below the header or is that by design?

Cal-L (changes requested):
The row items on the notifications list on the Wallet tab seems misaligned vertically. For example, the icon and badge floats more towards the bottom whereas the middle content floats towards the top. Should this all be center aligned? More of a design comment: It'll scale better if we align the icon+badge towards the top with a bit of top padding so that it appears center aligned with wallet notifications. Applying that same approach to other notification types such as announcements, it will balance the row items out. After that, the height of each announcement will only be determined by the body content, which the icon+badge remains fixed against the top.

Cal-L (changes requested):
Unless by design, the trash icon should either be center aligned vertically or aligned against the top, similar to the icon+badge

Cal-L (changes requested):
Is the settings icon on the Notifications list page hooked up or is just a placeholder for now?

Jonathansoufer:
> Is the settings icon on the Notifications list page hooked up or is just a placeholder for now?

Settings was implemented [here](https://github.com/MetaMask/metamask-mobile/pull/9258), and

Jonathansoufer:
> Turn on notifications page's Learn more link needs a space to the left of itself

This is fixed and merged on main already.

Jonathansoufer:
> Are the sub toggles from the Notifications page functional yet and if so, can we provide videos/description on app behavior when those toggles are on/off

Not. They'll be calling profile sync, but for now they're just being saved on local storage. This PR just implements UI

Jonathansoufer:
> The Notifications list page seems to also have extra padding below the header or is that by design?

Fixed.

Jonathansoufer:
> The row items on the notifications list on the Wallet tab seems misaligned vertically. For example, the icon and badge floats more towards the bottom whereas the middle content floats towards the top. Should this all be center aligned? More of a design comment: It'll scale better if we align the icon+badge towards the top with a bit of top padding so that it appears center aligned with wallet notifications. Applying that same approach to other notification types such as announcements, it will balance the row items out. After that, the height of each announcement will only be determined by the body content, which the icon+badge remains fixed against the top.

Good one! Fixed.

Jonathansoufer:
> Unless by design, the trash icon should either be center aligned vertically or aligned against the top, similar to the icon+badge

Fixed.

Jonathansoufer:
> The Allow notifications toggle should remain off until a user taps Yes (currently, it slides right but back to the left)

Fixed.

app/components/UI/Notification/Empty/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+} from '../../../../component-library/components/Texts/Text';
+import { strings } from '../../../../../locales/i18n';
+import { createStyles } from './styles';
+import Icon, {
+  IconColor,
+  IconName,
+  IconSize,
+} from '../../../../component-library/components/Icons/Icon';
+const Empty = ({ testID }: { testID?: string }) => {
+  const styles = createStyles();
```
Cal-L, line 14: If nothing is dynamic in these styles, it doesn't need to be a function. Return `StyleSheet.create` directly instead

app/components/UI/Notification/List/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+        </NotificationRow.Root>
+      );
+    },
+    [onPress, styles],
+  );
+
+  const renderList = useCallback(
+    (list, idx) => (
+      <FlatList
+        // eslint-disable-next-line
```
Cal-L, line 133: Is there a reason why this is ignored by lint?

app/components/UI/Notification/List/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+      />
+    ),
+    [combinedLists, renderNotificationRow, styles.list],
+  );
+
+  return (
+    <View style={styles.container}>
+      {loading ? (
+        <View style={styles.loaderContainer}>
+          <ActivityIndicator color={colors.icon.default} size="large" />
```
Cal-L, line 158: @brianacnguyen What color would you suggest for Activity the activity indicator?

app/components/UI/Notification/List/styles.ts
```diff
@@ -0,0 +1,141 @@
...
+    tabStyle: {
+      paddingBottom: 0,
+      paddingVertical: 8,
+    },
+    tabBar: {
+      borderColor: colors.background.default,
+    },
+    textStyle: {
+      ...(typography.sBodyMD as TextStyle),
+      fontWeight: '500',
```
Cal-L, line 40: Is fontWeight needed? If so, the typography should be updated instead of fontWeight

app/util/notifications/methods/index.ts
```diff
@@ -36,24 +37,27 @@ interface ViewOnEtherscanProps {
...
   };
   details: Record<string, any>;
 }
 
 export const sortNotifications = (
   notifications: Notification[],
 ): Notification[] =>
-  notifications.sort((a, b) =>
+  notifications?.sort((a, b) =>
     a.createdAt > b.createdAt ? -1 : b.createdAt > a.createdAt ? 1 : 0,
```
Cal-L, line 61:
```suggestion
    b.createdAt - a.createdAt
```

app/components/Views/Notifications/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    >
+      <Icon
+        name={IconName.Setting}
+        size={IconSize.Lg}
+        style={createStyles().icon}
+      />
+    </TouchableOpacity>
+  ),
+  headerLeft: () => (
+    <TouchableOpacity onPress={() => navigation.goBack()}>
```
Cal-L, line 162: I'd recommend navigating to the actual page rather than going back. This prevent the risk of double tapping (which can go back twice)

app/components/Views/Notifications/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+    setLoading(true);
+    InteractionManager.runAfterInteractions(() => {
+      filterNotifications(selectedAddress);
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedAddress]);
+
+  useFocusEffect(
+    React.useCallback(() => {
+      if (!isNotificationEnabled) navigation.navigate('WalletView');
```
Cal-L, line 109: What is the scenario where we'd want to automatically navigate to the wallet screen automatically?

app/components/Views/Notifications/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+          ) === index,
+      );
+
+      setAllTransactions(allNotificationsSorted);
+
+      /**
+       * Based on a sorted and deduplicated list of notifications, filter notifications by type to populate different tabs
+       */
+
+      allNotificationsSorted?.filter((notification) => {
```
Cal-L, line 80: Use for each instead since this isn't actually filtering

app/components/Views/Notifications/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+      const annoucements: FeatureAnnouncementRawNotification[] = [];
+
+      /**
+       * Sort notifications by time and remove duplicates
+       */
+      const allNotificationsSorted = sortNotifications(
+        allNotifications,
+      )?.filter(
+        (notification, index, self) =>
+          self.findIndex(
```
Cal-L, line 69: Instead of `findIndex`, It's probably more performant to keep track IDs in an object and return false if the ID already exists in the object

app/components/UI/Notification/Row/Actions.tsx
```diff
@@ -0,0 +1,45 @@
...
+}: NotificationActionsProps) {
+  function handleCTAPress() {
+    if (link?.linkUrl) {
+      return Linking.openURL(link.linkUrl);
+    }
+  }
+
+  return (
+    <Button
+      testID="notification-actions-button"
```
Cal-L, line 35: Move into a constants file

Cal-L (commented):
Left some comments

app/components/UI/Notification/Empty/index.tsx
Jonathansoufer: Ok. No a problem. Fixed.

app/components/UI/Notification/List/index.tsx
Jonathansoufer: This shouldn't be a blocker, but in any case I changed to follow the pattern used in different place within the app which is `colors.primary.default`

app/components/UI/Notification/List/styles.ts
Jonathansoufer: again, I'm just following existing patterns available and used throughout the app, like [here](https://github.com/MetaMask/metamask-mobile/blob/dcdb03a95203e43d50dbe1dbb4d06226ba77dcd1/app/components/Views/Wallet/index.tsx#L77), [here](https://github.com/MetaMask/metamask-mobile/blob/476532d7f9c8d3699b4c591be744618e74f9136a/app/components/Views/RevealPrivateCredential/RevealPrivateCredential.tsx#L278), and [here](https://github.com/MetaMask/metamask-mobile/blob/f7466da745ad9db800abfe1a4e72639ba8e30598/app/components/Views/AddAsset/AddAsset.tsx#L96). Not an issue IMO.

app/components/Views/Notifications/index.tsx
Jonathansoufer: The Scenario is when users tap on bell icon with isNotificationEnabled equals true, they will see a list of notifications but also on the top right corner an settings icon allowing them to go to the settings and disable notifications. If they disable notifications can tap back, we shouldn't show them the notifications list (which will be still available on navigation stack) but bypass it directly to Wallet main page.

app/components/Views/Notifications/index.tsx
Jonathansoufer: Fixed.

app/components/Views/Notifications/index.tsx
Jonathansoufer: Refactored.

app/components/Views/Notifications/index.tsx
Jonathansoufer: Same approach used [here](https://github.com/MetaMask/metamask-mobile/blob/39f43bce87b2f9c686732e62f6fd52ce6b900683/app/components/Views/TransactionsView/index.js#L75), [here](https://github.com/MetaMask/metamask-mobile/blob/39f43bce87b2f9c686732e62f6fd52ce6b900683/app/components/Views/Asset/index.js#L352) before.

app/util/notifications/methods/index.ts
Jonathansoufer: This wouldn't work since createdAt is Date type, which would throw the "The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type" problem. I approached slightly different now mixing your suggestion and handling Date.

app/components/Views/Notifications/index.tsx
Cal-L: Is this the cause for the horizontal dismiss animation? Would we be able to have a direct dismiss by passing the condition to the screen that we're tapping back on instead?

app/components/UI/Notification/List/styles.ts
Cal-L: There are many areas in the app that contain less than ideal implementation. Those are areas that we need to refactor. Moving forward, let's remove the need to manage an additional piece of typography unless the actual styles provided in Figma isn't available.

app/components/Views/Notifications/index.tsx
Jonathansoufer: Moved logic to Settings.

Cal-L (dismissed):
LGTM. Since this feature isn't live yet, it's hard to imagine what the final flow will look like. We can review those future questions in later PRs.

Cal-L (approved):
LGTM

Cal-L:
Approved from code perspective, will defer to @Jonathansoufer regarding QA

PR #9390, opened 2024-04-24
feat(ramp): add ramp intent

wachunei:
## **Description**

This PR adds a new feature to implement the [Ramp's Buy & Sell Intent 2024 Proposal](https://gist.github.com/wachunei/f53eceb50816de789e4dfc0850ca4b34). This feature allows ramp flows to capture data from the query params of a link or deeplink. 

The properties to be parsed in this PR are:
- `chainId`: adds or changes to the passed network. In case the chainId is not supported, the Network switcher will prompt the user to select a supported network
- `address`: selects the token address if it is available, if not will default to the current behavior
- `amount`: fills the amount if possible, if not will default to 0

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/RAMPS-1511

## **Manual testing steps**

Use the MVP demo to generate links https://h2rywd.csb.app/
This demo app is set to use USA - California as region and Debit/Credit as payment method.

1. Go to the demo https://h2rywd.csb.app/
2. Generate link
3. Click on generated link, the intent properties must behave as in the descriptions

### **After**

https://github.com/MetaMask/metamask-mobile/assets/1024246/22ca9f4c-e743-4688-af1d-77cba29ef9d9

https://github.com/MetaMask/metamask-mobile/assets/1024246/fd92be99-fb6d-4072-a468-43ac4da82e62


app/components/UI/Ramp/utils/parseRampIntent.test.ts
```diff
@@ -0,0 +1,58 @@
+import parseRampIntent from './parseRampIntent';
+
+describe('parseRampIntent', () => {
```
georgeweiler, line 3: One additional test case might be nice to include: If the params are partial. What do you think?

We have this validation case to check when they are _all_ missing:

```
if (
!rampIntent.address &&
!rampIntent.chainId &&
!rampIntent.amount &&
!rampIntent.currency
) {
return undefined;
}
```

but `parseRampIntent` should still work if even 1 param is available. Maybe something like this?


```
it('should return a RampIntent object with only defined fields', () => {
  const pathParams = { address: '0x1234567890', amount: undefined };
  const result = parseRampIntent(pathParams);
  expect(result).toEqual({ address: '0x1234567890', chainId: '1' });
}); 
```

wachunei: Yes, good catch! will update

wachunei: Added in 395645ad5

Ericfav (approved):

Ericfav (approved):

app/components/UI/Ramp/deeplink/handleRampUrl.test.ts
```diff
@@ -12,23 +14,73 @@ describe('handleRampUrl', () => {
     navigation = {
       navigate: jest.fn(),
     } as unknown as NavigationProp<ParamListBase>;
+
+    (handleRedirection as jest.Mock).mockClear();
   });
 
-  it('should navigate to BUY route when rampType is BUY', () => {
+  it('should call handle redirection with the paths', () => {
```
NicolasMassart, line 21: Match [naming guidelines](https://github.com/MetaMask/contributor-docs/blob/main/docs/unit-testing.md#-use-describe-to-group-tests-under-scenarios)

```suggestion
  it('handles redirection with the paths', () => {
```

app/components/UI/Ramp/deeplink/handleRampUrl.test.ts
```diff
@@ -12,23 +14,73 @@ describe('handleRampUrl', () => {
...
     });
+    expect(handleRedirection).toHaveBeenCalledWith(
+      ['somePath'],
+      { as: 'example' },
+      RampType.BUY,
+      navigation,
+    );
+  });
+
+  it('should navigate to Buy route when rampType is BUY, redirectPaths length is 0 and query params do not have allowed fields', () => {
```
NicolasMassart, line 35:
```suggestion
  it('navigates to Buy route when rampType is BUY, redirectPaths length is 0 and query params do not have allowed fields', () => {
```

app/components/UI/Ramp/deeplink/handleRampUrl.test.ts
```diff
@@ -12,23 +14,73 @@ describe('handleRampUrl', () => {
...
+      rampPath: '?as=example',
+      rampType: RampType.BUY,
+      navigation,
+    });
+    expect(handleRedirection).not.toHaveBeenCalled();
     expect(navigation.navigate).toHaveBeenCalledWith(Routes.RAMP.BUY);
   });
 
-  it('should navigate to SELL route when rampType is SELL', () => {
+  it('should navigate to Sell route when rampType is SELL, redirectPaths length is 0 and query param do not have allowed fields', () => {
```
NicolasMassart, line 45:
```suggestion
  it('navigates to Sell route when rampType is SELL, redirectPaths length is 0 and query param do not have allowed fields', () => {
```

app/components/UI/Ramp/deeplink/handleRampUrl.test.ts
```diff
@@ -12,23 +14,73 @@ describe('handleRampUrl', () => {
...
-      rampPath: '/somePath?as=example',
+      rampPath: '?as=example',
       rampType: RampType.SELL,
       navigation,
     });
+    expect(handleRedirection).not.toHaveBeenCalled();
     expect(navigation.navigate).toHaveBeenCalledWith(Routes.RAMP.SELL);
   });
+
+  it('should navigate to Buy route when rampType is BUY, redirectPaths length is 0 and query param is intent', () => {
```
NicolasMassart, line 55:
```suggestion
  it('navigates to Buy route when rampType is BUY, redirectPaths length is 0 and query param is intent', () => {
```

app/components/UI/Ramp/deeplink/handleRampUrl.test.ts
```diff
@@ -12,23 +14,73 @@ describe('handleRampUrl', () => {
...
+    expect(navigation.navigate).toHaveBeenCalledWith(Routes.RAMP.BUY, {
+      screen: Routes.RAMP.GET_STARTED,
+      params: {
+        chainId: '1',
+        address: '0x123456',
+      },
+    });
+  });
+
+  it('should navigate to Sell route when rampType is SELL, redirectPaths length is 0 and query param is intent', () => {
```
NicolasMassart, line 71:
```suggestion
  it('navigates to Sell route when rampType is SELL, redirectPaths length is 0 and query param is intent', () => {
```

app/components/UI/Ramp/deeplink/handleRampUrl.ts
```diff
@@ -9,16 +16,34 @@ interface RampUrlOptions {
...
+    switch (rampType) {
+      case RampType.BUY:
+        navigation.navigate(...createBuyNavigationDetails(rampIntent));
+        break;
+      case RampType.SELL:
+        navigation.navigate(...createSellNavigationDetails(rampIntent));
+        break;
+    }
+  } catch (error) {
+    Logger.error(
```
NicolasMassart, line 44: Q: does this really have to be logged to sentry?
- Is this expected to happen often?
- Is this out of our control?
- Can not you do something about it, as providing fallback or automatically retry or show an error screen?
- 
If you answer yes to at least one of these, it may be better to move to analytics with `trackEvent` or `trackErrorAsAnalytics`.

app/components/UI/Ramp/deeplink/handleRedirection.test.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('handleRedirection', () => {
+  let navigation: NavigationProp<ParamListBase>;
+
+  beforeEach(() => {
+    navigation = {
+      navigate: jest.fn(),
+    } as unknown as NavigationProp<ParamListBase>;
+  });
+
+  it('should navigate to TRANSACTIONS_VIEW route when first path is "activity"', () => {
```
NicolasMassart, line 17:
```suggestion
  it('navigates to TRANSACTIONS_VIEW route when first path is "activity"', () => {
```

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.test.ts
```diff
@@ -0,0 +1,29 @@
+import getRedirectPathsAndParams from './getRedirectPathAndParams';
+
+describe('getRedirectPathsAndParams', () => {
+  it('should return empty values for empty paths', () => {
```
NicolasMassart, line 4:
```suggestion
  it('returns empty values for empty paths', () => {
```

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.ts
```diff
@@ -0,0 +1,19 @@
+const dummyProtocol = 'ramp:';
```
NicolasMassart, line 1: Q: could you use another term than "dummy" ?
To me it means substitute for the real thing or placeholder data with the same format as real data but an actual one.
This is actually used. Not standardised outside of MM but not useless or placeholder either.

NicolasMassart (commented):
first pass review. Will continue in a second pass as the PR is a bit big

app/components/UI/Ramp/utils/parseAmount.ts
```diff
@@ -0,0 +1,59 @@
+// This regex matches a string that is a number with an optional decimal part.
+const regex = /^(\d*)(\.\d+)?$/;
```
NicolasMassart, line 2: Q: is this regex redos proof?

app/components/UI/Ramp/utils/parseRampIntent.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  it('should return undefined if pathParams do not contain necessary fields', () => {
```
NicolasMassart, line 4:
```suggestion
  it('returns undefined if pathParams do not contain necessary fields', () => {
```

app/components/UI/Ramp/utils/parseRampIntent.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const pathParams = {};
+    const result = parseRampIntent(pathParams);
+    expect(result).toBeUndefined();
+  });
+
+  it('should return a RampIntent object if pathParams contain necessary fields', () => {
```
NicolasMassart, line 10:
```suggestion
  it('return a RampIntent object if pathParams contain necessary fields', () => {
```

app/components/UI/Ramp/utils/parseRampIntent.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const result = parseRampIntent(pathParams);
+    expect(result).toEqual({
+      address: '0x1234567890',
+      chainId: '1',
+      amount: '10',
+      currency: 'usd',
+    });
+  });
+
+  it('should default to chainId 1 if token address is defined but chainId is not', () => {
```
NicolasMassart, line 26:
```suggestion
  it('defaults to chainId 1 if token address is defined but chainId is not', () => {
```

app/components/UI/Ramp/utils/parseRampIntent.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const result = parseRampIntent(pathParams);
+    expect(result).toEqual({
+      address: '0x1234567890',
+      chainId: '1',
+      amount: '10',
+      currency: 'usd',
+    });
+  });
+
+  it('should return a RampIntent object with only defined fields', () => {
```
NicolasMassart, line 42:
```suggestion
  it('returns a RampIntent object with only defined fields', () => {
```

app/components/UI/Ramp/utils/parseRampIntent.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const result = parseRampIntent(pathParams);
+    expect(result).toEqual({
+      address: '0x1234567890',
+      chainId: '76',
+      amount: '10',
+      currency: 'usd',
+    });
+  });
+
+  it('should return a RampIntent object with only defined values', () => {
```
NicolasMassart, line 59:
```suggestion
  it('returns a RampIntent object with only defined values', () => {
```

app/components/UI/Ramp/deeplink/handleRampUrl.ts
wachunei: I think so. This is not supposed to happen as nothing in our control should throw (`getRedirectPathsAndParams`, `handleRedirection`, `parseRampIntent`). This is to catch any other error there might be to avoid crashing the app. At this point there is no UI so there no feedback to display. 

With this information what do you think we should do?

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.ts
wachunei: It is just used to form a valid url for `URL` constructor. It is not used before of after. it can be anything `anything:`. It is not useless but is kind of a placeholder. What variable name do you suggest?

app/components/UI/Ramp/hooks/useCryptoCurrencies.ts
```diff
@@ -55,19 +56,37 @@ export default function useCryptoCurrencies() {
    */
   useEffect(() => {
     if (cryptoCurrencies) {
+      if (intent?.address) {
+        const intentAsset = cryptoCurrencies.find(
+          (token) =>
+            token.address.toLowerCase() === intent.address?.toLowerCase(),
```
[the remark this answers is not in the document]
wachunei: No

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.tsx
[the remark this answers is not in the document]
wachunei: I agree, will add it to the list of files that I'm writing tests for.

app/components/UI/Ramp/utils/parseAmount.ts
wachunei: Yes, I kept that and removed the others and this commit [`d1d02ec` (#9390)](https://github.com/MetaMask/metamask-mobile/pull/9390/commits/d1d02ec0c502a98b3683c64a5ef0322baba7869d)

app/components/UI/Ramp/sdk/index.tsx
```diff
@@ -179,6 +187,9 @@ export const RampSDKProvider = ({
   const [selectedRegion, setSelectedRegion] = useState(INITIAL_SELECTED_REGION);
   const [unsupportedRegion, setUnsupportedRegion] = useState<Region>();
 
+  //** Ramp Intent */
```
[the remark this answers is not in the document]
wachunei: Removed in fc630481d

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.tsx
```diff
@@ -132,16 +133,18 @@ function NetworkSwitcher() {
...
+  const navigateToGetStarted = useCallback(() => {
+    navigation.navigate(Routes.RAMP.GET_STARTED, { chainId: undefined });
+  }, [navigation]);
 
-  const switchToMainnet = useCallback((type: 'mainnet' | 'linea-mainnet') => {
-    const { NetworkController } = Engine.context;
-    NetworkController.setProviderType(type as NetworkType);
-  }, []);
+  const switchToMainnet = useCallback(
+    (type: 'mainnet' | 'linea-mainnet') => {
```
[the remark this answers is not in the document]
wachunei: Not now, these are the arguments for the `setProviderType` function. Although it has been deprecated and it must be refactored in an upcoming PR, if there's not one already open. It is like this in [main branch now](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.tsx#L143)
image

app/components/UI/Ramp/Views/GetStarted/GetStarted.tsx
```diff
@@ -69,7 +85,10 @@ const GetStarted: React.FC = () => {
 
   useEffect(() => {
     if (getStarted) {
-      if (!isNetworkRampSupported) {
+      if (
+        !isNetworkRampSupported ||
+        (params?.chainId && params.chainId !== selectedChainId)
```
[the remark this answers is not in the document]
wachunei: This condition redirects the user to the Network Switcher view, a view with all the supported networks by Ramp.
If the current chainId is `1` and the user clicks `metamask://buy?chainId=56`, then they will be redirected to the networks switcher. The network switcher view will handle this redirection: if the network is not added, it will prompt the "Add network". If the network is added, it will switch to that network and continue the intent with the rest of the params if they are present (token and amount).

app/components/UI/Ramp/Views/GetStarted/GetStarted.tsx
```diff
@@ -42,6 +52,12 @@ const GetStarted: React.FC = () => {
     }
   }, [isBuy, selectedChainId, trackEvent]);
 
+  useEffect(() => {
+    if (params) {
+      setIntent(params);
+    }
+  }, [intent, params, setIntent]);
```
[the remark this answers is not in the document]
wachunei: We don't. Removed in 82bdd2547

app/components/UI/Ramp/hooks/useIntentAmount.ts
```diff
@@ -0,0 +1,66 @@
+import { useEffect } from 'react';
+import { type BN } from 'ethereumjs-util';
+import { useRampSDK } from '../sdk';
+import parseAmount from '../utils/parseAmount';
+import { toTokenMinimalUnit } from '../../../../util/number';
+import Logger from '../../../../util/Logger';
+import { FiatCurrency } from '@consensys/on-ramp-sdk';
+
+export default function useIntentAmount(
```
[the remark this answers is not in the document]
wachunei: Added in 46d98a2ec

app/components/UI/Ramp/utils/parseAmount.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    ${'123456'}                | ${0}          | ${'123456'}
+    ${'000123456'}             | ${undefined}  | ${'123456'}
+    ${'000123456'}             | ${3}          | ${'123456'}
+    ${'00123456'}              | ${2}          | ${'123456'}
+    ${'0123456'}               | ${1}          | ${'123456'}
+    ${'0000123456'}            | ${0}          | ${'123456'}
+    ${'12.012345678912345678'} | ${18}         | ${'12.012345678912345678'}
+    ${'12.012345678912345678'} | ${17}         | ${'12.01234567891234567'}
+    ${'asdas'}                 | ${3}          | ${undefined}
+    ${'0.0.0'}                 | ${undefined}  | ${undefined}
```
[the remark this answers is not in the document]
wachunei: It could happen if the amount set in the query params is set to those value. These test are covering these lines:

```ts
    const match = regex.exec(amount);
    if (!match) {
      return;
    }
```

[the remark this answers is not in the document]
NicolasMassart: Yes, but I mean why '0.0.0' would be a value passed in query, why not "my little pony" then? Is there a real risk a version formatted value is passed or does just any value could be passed? If so, you can't test all values that would not match.

app/components/UI/Ramp/Views/GetStarted/GetStarted.tsx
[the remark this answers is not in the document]
NicolasMassart: Thanks! that could be an interesting comment to add in the code.

app/components/UI/Ramp/utils/parseAmount.test.ts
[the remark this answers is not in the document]
wachunei: You're right. "0.0.0" is a representative of any value, including "my little pony". I know I can't test all the values, this is just one case to cover for that branching of the code.

app/components/UI/Ramp/Views/GetStarted/GetStarted.tsx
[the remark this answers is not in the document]
wachunei: Yes! Added in 9a566fafe

app/components/UI/Ramp/deeplink/handleRampUrl.ts
wachunei: >So I would say that if there's nothing directly throwing here, remove the try/catch. Otherwise, imagine, we could fill all our code with try catch everywhere...

We use `navigation.navigate` which throws when the navigator is not set up.
However, this is the outermost boundary between Ramps code domain and mobile code and we want to have a try/catch here in particular as part of [isolating the experiences](https://docs.google.com/document/d/1m8ahcx4G5zTz1eGhvsZJUms7uri2YHGyELq4YIw-eK8/edit#heading=h.1ef8pgy3bn9y) and preventing them from crashing the app.

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    ${'///example//long/path'}                       | ${[['example', 'long', 'path'], undefined]}
+    ${'///example//long/path?'}                      | ${[['example', 'long', 'path'], undefined]}
+    ${'?chainId=1'}                                  | ${[[], { chainId: '1' }]}
+    ${'?chainId=1&amount=4.20'}                      | ${[[], { chainId: '1', amount: '4.20' }]}
+    ${'//example?chainId=1'}                         | ${[['example'], { chainId: '1' }]}
+    ${'//example?chainId=1&amount=4.20'}             | ${[['example'], { chainId: '1', amount: '4.20' }]}
+    ${'//example/path?chainId=1&amount=4.20'}        | ${[['example', 'path'], { chainId: '1', amount: '4.20' }]}
+    ${'////example////path?chainId=1&amount=4.20&&'} | ${[['example', 'path'], { chainId: '1', amount: '4.20' }]}
+  `(
+    'should return the expected value for rampPath $rampPath',
```
wachunei, line 24:
```suggestion
    'returns the expected value for rampPath $rampPath',
```

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.ts
wachunei: Very good questions, let me explain:

You are correct, the `rampPath` string is **not** a URL, but it was part of one. This string is the result from the handler of deeplinks and universal links. For the [former](https://github.com/MetaMask/metamask-mobile/blob/7caff7bc06023086afc8c0d583d34ec5b213eb5b/app/core/DeeplinkManager/ParseManager/handleMetaMaskDeeplink.ts), the URL was [one of these](https://github.com/MetaMask/metamask-mobile/blob/7caff7bc06023086afc8c0d583d34ec5b213eb5b/app/core/DeeplinkManager/ParseManager/handleMetaMaskDeeplink.ts#L126-L142):

* `metamask://buy<rampPath>`
* `metamask://sell<rampPath>`
* `metamask://buy-crypto<rampPath>`
* `metamask://sell-crypto<rampPath>`

For the [latter](https://github.com/MetaMask/metamask-mobile/blob/feat/ramp-intent/app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts), [one of these](https://github.com/MetaMask/metamask-mobile/blob/feat/ramp-intent/app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts#L99-L108): 

* `https://metamask.app.link/buy<rampPath>`
* `https://metamask.app.link/buy-crypto<rampPath>`
* `https://metamask.app.link/sell<rampPath>`
* `https://metamask.app.link/sell-crypto<rampPath>`

>if it's not a URL but it's just handy to use the URL tools to split the string parts... then maybe you could encapsulate this into a type that may use URL internally but that would clarify that you do not expect this to directly be a URL.

This is what I wanted to express with `rampPath` type. it is just the raw `pathname` of the `URL` object resulting from combining any protocol with the ramp path (```new URL(`${dummyProtocol}${rampPath}`)```).

> Given the unit test, I see you use the params and path as in a URL, but at the same time there's all these /// at the begining that I don't really understand. So not really a URL. I would definitely favour a specific object to handle that even if its implementation uses URL under the hood for ease.

All of these `///` and weird cases come from the fact that we would be parsing the input from external links that might be malformed. Thankfully the `URL` handles that perfectly. 

My main question is, in more detail and concretly, what do you mean with "encapsulate this into a type" or "a specific object"? the `getRedirectPathsAndParams` itself is meant to be the function that handles this logic.

app/components/UI/Ramp/hooks/useIntentAmount.ts
```diff
@@ -30,9 +41,9 @@ export default function useIntentAmount(
           try {
             valueAsNumber = Number(parsedAmount);
           } catch (numberError) {
-            Logger.error(
-              numberError as Error,
+            console.error(
```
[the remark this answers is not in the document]
wachunei: The linter allows it, the `console.error` will still display the error in development, but not surface it in the prod build, so we still have visibility during development instead of just burying it.

app/components/UI/Ramp/deeplink/handleRampUrl.ts
NicolasMassart: Is there a fallback we could have if navigator throws? If it happens be we can't do anything about it, even if it's sent to Sentry, maybe just catching and prevent crash is enough?

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.ts
NicolasMassart: in this case, if the protocol is really just dummy to make the URL work, then maybe simply:
```suggestion
const dummyProtocol = 'dummy:';
```

app/components/UI/Ramp/deeplink/handleRampUrl.ts
wachunei: Sending it to Sentry would help us debug it faster, I prefer keeping it there rather than running in a situation like this <span>https://github.com/MetaMask/metamask-mobile/pull/8414</span> where we had to add it to investigate 🕵️

app/components/UI/Ramp/deeplink/utils/getRedirectPathAndParams.ts
wachunei: works for me! Changed in a92b3be99

NicolasMassart (dismissed):
Looks good to me.

infiniteflower (dismissed):
Looks good on the Swaps side

infiniteflower (approved):

NicolasMassart (approved):
Looks good to me.

PR #9435, opened 2024-04-29
feat: sdk async persistence  and protocol upgrade

abretonc7s:
## **Description**

- Full sdk protocol async communication and wallet connect sdk update.
- Allow for much faster and stable connection with both sdk and wc.
- Simplify the AccountConnect modal since wc now requests via hostname (vs id before).

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


andreahaku (dismissed):
LGTM

andreahaku (dismissed):
LGTM

andreahaku (approved):
LGTM

app/components/Views/AccountConnect/AccountConnect.tsx
```diff
@@ -119,14 +118,15 @@ const AccountConnect = (props: AccountConnectProps) => {
...
-  const [hostname, setHostname] = useState<string>(origin);
-
-  const urlWithProtocol = prefixUrlWithProtocol(hostname);
   const sdkConnection = SDKConnect.getInstance().getConnection({
     channelId: channelIdOrHostname,
   });
-  // Last wallet connect session metadata
-  const wc2Metadata = useSelector((state: RootState) => state.sdk.wc2Metadata);
+  const hostname =
+    origin ?? channelIdOrHostname.indexOf('.') !== -1
```
tommasini, line 125: Heyyy, could you leave a comment explaining the difference between the origin and the channelIdOrHostname?

I took some time to understand, and I'm not sure, but origin variable here is only related to the in app browser right?

channelIdOrHostname is regarding wc and sdk right?
Also it would be cool to understand easier the indexOf('.') on this case I can't really understand what makes this diff to choose sdkConnection (That could be actually an emtpy string, so it wouldn't appear the website we are trying connect on that case?) or the channelIdOrHostname, (should we show the channelId to the user? If it's indeed an id it wouldn't be cool to show it to the user right?) 

Please let me know if I'm miss understanding something on this logic

christopherferreira9:
# Test Plan
In the backward compatibility testing we refer to:
- Old Wallet -> Any wallet with sdk-communication-layer < 0.20.2
- New Wallet -> Wallet using sdk-communication-layer version 0.20.2

Combinations tested:
Wallet | SDK | Result
-- | -- | --
Old Wallet | Old SDK | ✅
Old Wallet | New SDK |✅
New Wallet | Old SDK | ✅
New Wallet | New SDK | ✅

## Test Cases:
### Async disconnection 1
Given that a user connects to MetaMask from a dapp
And Kills MetaMask and reopens again
Then the Connection list should be empty


### Async disconnection 2
Given that a user connects to MetaMask from a dapp
And Kills/refreshes the dapp and reopens again
And the user removes the connection from MetaMask
Then the Connection list should be empty

### Wallet Queue management (simulating a connection/Wallet crash)
Given that a user connects to MetaMask from a dapp
And the user triggers a sign request followed by suddenly killing MM
Then upon opening MetaMask manually the session should resume and a sign request should appear

## Dapp types where tests were performed:
- NodeJS
- ReactNative
- React & NextJS
- iOS (socket version)
- Unity desktop (socket version)

app/core/SDKConnect/Connection/Connection.ts
```diff
@@ -164,6 +173,7 @@ export class Connection extends EventEmitter2 {
     this.rpcQueueManager = rpcQueueManager;
     // batchRPCManager should be contained to current connection
     this.batchRPCManager = new BatchRPCManager(id);
+    this.protocolVersion = protocolVersion ?? 1;
```
tommasini, line 176: Curious question, which version the protocol can have and where can I find the differences between them?

app/core/SDKConnect/handlers/handleSendMessage.ts
```diff
@@ -77,41 +77,44 @@ export const handleSendMessage = async ({
...
-            // Ignore temporarily until next stage of permissions system implementation
-            DevLogger.log(`[handleSendMessage] error goBack()`, _e);
-          }
-          connection.navigation?.navigate(Routes.MODAL.ROOT_MODAL_FLOW, {
-            screen: Routes.SHEET.RETURN_TO_DAPP_MODAL,
-          });
-        }
-      }
       return;
+      // const currentRoute = connection.navigation?.getCurrentRoute()?.name;
```
tommasini, line 81: Any reason for we want to leave this code commented here?

tommasini (approved):
After talking with Chris, the connection with empty string as a origin with the sdk will be addressed with other polishing on a next PR, with this being said, this PR LGTM!

app/components/Views/AccountConnect/AccountConnect.tsx
andreahaku: `channelIdOrHostname.indexOf('.') !== -1` is to check that it's not `channelId` as it can't contain "."
the user should not be shown with the channelId (it's a UUID) as it is something that is not really visible anywhere and only used by the SDK to create a socket.io room.

christopherferreira9: > I'm not sure if this would be one of the source of this [issue](https://github.com/MetaMask/mobile-planning/issues/1731)

It is not related to that issue. This only relates to the connection modal and not with any other confirmation screens.
What Andrea said plus us not being able to validate the origin outside the in-app browser scope may result in an empty string depending on what the dapp sends us. 
This and other UI details will be addressed in a future PR.

app/core/SDKConnect/Connection/Connection.ts
christopherferreira9: Shared internal document directly.

PR #9441, opened 2024-04-29
chore: Convert initial background state to typed background state

tommasini:
## **Description**

A small step forward into an application with better typescript coverage.
Convert all files  to use backgroundState from initial-root-state file instead of the initial-background-state json file (unit test files)

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1734

## **Manual testing steps**

1. Go to this page...
2.
3.


brianacnguyen (dismissed):
DS approved

app/core/Engine.test.js
```diff
@@ -43,22 +43,14 @@ describe('Engine', () => {
...
-        ...backgroundState.KeyringController,
-        vault: {
-          cipher: 'mock-cipher',
-          iv: 'mock-iv',
-          lib: 'original',
-        },
-      },
+    initialBackgroundState = {
+      ...initialBackgroundState,
     };
```
Gudahtt, line 50: Nit: We can get rid of this part since we're no longer overwriting any part of this object

```suggestion
    initialBackgroundState = {
      ...initialBackgroundState,
    };
```

app/util/test/initial-background-state.json
```diff
@@ -137,7 +132,71 @@
   },
   "SmartTransactionsController": {
     "smartTransactionsState": {
-      "liveness": true
+      "smartTransactions": {
```
Gudahtt, line 135: Could we update this to be closer to the initial state? It can be found here: https://github.com/MetaMask/smart-transactions-controller/blob/d5fc8da68ff4030bddf3b10b39878a7e4273cb38/src/SmartTransactionsController.ts#L149

This fixture is meant to be totally blank (or as close to it as possible), not to have any built-in example data.

For the `undefined` properties, using `false` instead of `undefined` is a good idea though, since `undefined` isn't supposed to be allowed in controller state (the controller uses it only because it's outdated).

app/util/test/initial-root-state.ts
```diff
@@ -7,8 +7,9 @@ import { initialState as initialSmartTransactions } from '../../core/redux/slice
 import { initialState as transactionMetrics } from '../../core/redux/slices/transactionMetrics';
 import initialBackgroundState from './initial-background-state.json';
 
-// Cast because TypeScript is incorrectly inferring the type of this JSON object
-const backgroundState: EngineState = initialBackgroundState as any;
+// Cast as unknown since there are still mock variables that do not correspond with the existing types
```
Gudahtt, line 10: Nit: Here is an alternative explanation that is more specific:
```suggestion
// A cast is needed here because we use enums in some controllers, and TypeScript doesn't consider
// the string value of an enum as satisfying an enum type.
```

app/core/Engine.test.js
tommasini: Thanks for the review! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9441/commits/fc7803f14af17455211386b9515c73197cf63b8a)

app/util/test/initial-background-state.json
tommasini: Thanks for the review! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9441/commits/fc7803f14af17455211386b9515c73197cf63b8a)

app/util/test/initial-root-state.ts
tommasini: This was a learning lesson for me! I didn't know was because of the enums! Thanks!
Addressed on this commit: [`fc7803f` (#9441)](https://github.com/MetaMask/metamask-mobile/pull/9441/commits/fc7803f14af17455211386b9515c73197cf63b8a)

tommasini: I did more reading about this for learning purposes
Found this [article](https://dev.to/ivanzm123/dont-use-enums-in-typescript-they-are-very-dangerous-57bh) that brings some cool info about enums

It seems that the only solution we would have for every scenario would indeed convert to typed objects instead of enums
```
const MyConstants: Record<string, string> = {
  Value1: "VALUE_1",
  Value2: "VALUE_2",
};
```
But I see disadvantages there:
* Reduced Type Safety: While you can achieve a level of type safety using Record types or specific object typings, it's generally less strict compared to enums. You won't get errors when accessing non-existent properties, for example.
* No IntelliSense for Values: Unlike enums, where code editors can provide autocompletion for members, objects do not offer autocompletion for their values, only for their keys.

Not seeing a perfect solution for this right now!

--- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- 
Ignore this next part, it was me deep diving and taking notes where the enums live, if we want to create an issue to address this we have some work done here, if not ignore it 
--- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- 
Deep diving into the controllers we use enums in:
* [AddressBookController](https://github.com/MetaMask/core/blob/040abb5816207066e5bff07e655f54625afc0c91/packages/address-book-controller/src/AddressBookController.ts#L29)
* [assetsUtils](https://github.com/MetaMask/core/blob/174be6d06a0ea7d1efc315ae8507a3f5c14b25b7/packages/assets-controllers/src/assetsUtil.ts#L116)
* [assets controllers constants](https://github.com/MetaMask/core/blob/7b32741a8243bbeeb48e35f366f795a12a4de370/packages/assets-controllers/src/constants.ts#L1)
* [NftDetectionController](https://github.com/MetaMask/core/blob/4a6cd1a1778c41314ea8bffde2a924a5253d0430/packages/assets-controllers/src/NftDetectionController.ts#L164)
* [TokenRatesController](https://github.com/MetaMask/core/blob/b7462df72749ea54e8082c30792420f0a265fb62/packages/assets-controllers/src/TokenRatesController.ts#L76)
* [build-utils](https://github.com/MetaMask/core/blob/0b75d3794a64c0ef37de1f3a323fb4386120edb3/packages/build-utils/src/transforms/remove-fenced-code.ts#L17)
* [Controller utils constants](https://github.com/MetaMask/core/blob/80ff2fbb1fb38f577b4a5c7e3778550c4682202d/packages/controller-utils/src/constants.ts#L121)
* [Controller utils types](https://github.com/MetaMask/core/blob/4e03239443efea467d625d5c2167da9dbf82a2ef/packages/controller-utils/src/types.ts#L55)
* [Keyring Controller constants](https://github.com/MetaMask/core/blob/ffd16d4a778933ee257b6f7d28f308c280ad3d4b/packages/keyring-controller/src/constants.ts#L1)
* [Keyring Controller](https://github.com/MetaMask/core/blob/ffd16d4a778933ee257b6f7d28f308c280ad3d4b/packages/keyring-controller/src/KeyringController.ts#L53)
* [Logging controller (multiple files)](https://github.com/MetaMask/core/blob/83d37dcf390342a8ec6fb0c3a9886267a890e1cb/packages/logging-controller/src/logTypes/EthSignLog.ts#L6)
* [Network Controller ](https://github.com/MetaMask/core/blob/c8f508c03622ac047e419504329216707ccac5c7/packages/network-controller/src/constants.ts#L4)
* [Permission controller](https://github.com/MetaMask/core/blob/843a7940a4af63cb4bad215057d97bae4787d0a4/packages/permission-controller/src/Permission.ts#L351)
* Phishing Controller
* Transaction Controller

matthewwalsh0 (dismissed):

app/components/Views/confirmations/Send/index.test.tsx
```diff
@@ -21,7 +21,7 @@ const initialState = {
   settings: {},
   engine: {
     backgroundState: {
-      // ...initialBackgroundState,
+      // ...backgroundState,
```
NicolasMassart, line 24: is it worth changing this? it's commented so better remove it completely

app/util/test/initial-background-state.json
NicolasMassart, line 1: Q: why the changes in the json?

Prithpal-Sooriya (dismissed):

app/util/test/initial-background-state.json
tommasini: We discussed that earlier: https://github.com/MetaMask/metamask-mobile/pull/9441#discussion_r1636281135

We should have the mock data closest possible to the default state!

app/components/Views/confirmations/Send/index.test.tsx
tommasini: Removed [here](https://github.com/MetaMask/metamask-mobile/pull/9441/commits/44c8a0a378968b330a0054022bd31db049514b22)

wachunei (dismissed):
Ramp changes LGTM

NicolasMassart (commented):
Sonar should exclude test utils from test coverage
@sethkfman can we do that?

omridan159 (approved):

Gudahtt (approved):

sethkfman (approved):
LGTM

PR #9459, opened 2024-04-30
chore(devDeps): bump `@babel/core` to `^7.24.5`

legobeat:
## **Description**

- devDeps: @babel/core@^7.23.2->^7.24.5
- ~chore: regenerate `ppom.html.js`~

## **Related issues**
- Broken out from #9454
  - https://github.com/MetaMask/metamask-mobile/pull/9454#discussion_r1584720941
- #9475

#### Blocking
- #9454
  - #9452

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #9454, opened 2024-04-30
chore: Update Jest to v29

legobeat:
_This is #7160 rebased on `main` with changes to snapshots reapplied_

---
Jest has been updated to v29, along with related dependencies. This change comes with a number of new features, such as the new `jest.replaceProperty()` function.

The only breaking change affecting us is a change to the snapshot format. All snapshots have been updated.

See here for the migration guide: https://jest-archive-august-2023.netlify.app/docs/upgrading-to-jest29/

See here for a full list of changes: https://github.com/jestjs/jest/blob/main/CHANGELOG.md#2900

#### Blocked by
- [x] #9459 

#### Blocking
- #9452

#### Previously 
- #7160
- #6277 
- #6876 
- #8960 

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #9475, opened 2024-04-30
chore: Fix handling of generated ppom files

legobeat:
## **Description**

- Add generated `app/lib/ppom/ppom.html.js` to `.gitignore` and remove file from source control
- Infer webpack build mode for ppom subproject from `NODE_ENV` env var
  - This means that dev builds will not be minified (like the production build being ignored here is)
- Add `clean:ppom` package script which cleans generated ppom files
  - Run `clean:ppom` as part of `clean` package script
- ci: Run `yarn setup` (includes ppom-build step in addition to the steps in `setup:node`) instead of `yarn setup:node` before GitHub Actions unit tests 
  
## **Related issues**

- #9459
- https://github.com/MetaMask/metamask-mobile/pull/9454#discussion_r1584720941

## **Manual testing steps**
- Run `rm app/lib/ppom/ppom.html.js && NODE_ENV=production yarn setup`
  * Verify that `app/lib/ppom/ppom.html.js` is identical to previous version, e.g. by comparing output of `sha256sum app/lib/ppom/ppom.html.js`
- Run `rm app/lib/ppom/ppom.html.js && NODE_ENV=development yarn setup`
  * Verify that `app/lib/ppom/ppom.html.js` is not minified


jpuri (approved):

.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: yarn
-      - run: yarn setup:node
+      - run: yarn setup
```
Gudahtt, line 92: Hmm, so I had introduced `setup:node` here to speed up installation in CI by skipping iOS-specific setup steps. I found this really useful locally as a Linux user as well, because I wasn't benefiting from the iOS setup steps.

It seems like you've identified a bug with `setup:node` here, in that it was missing that PPOM generation step. Rather than workaround that problem, perhaps we could fix `setup:node`? Or at least delete it, since it's no longer a reliable setup command because it's missing that step.

legobeat: Thanks for pointing it out. There is also some redundancy between `setup` and `setup:node`.

Another way to go about it would be to extend `setup:node` with the PPOM steps, and change `setup.mjs` to replace the redundant steps with a call to `setup:node`.

Gudahtt: >Or make setup.mjs take some flag determining which steps to run.

This seems ideal to me! I would have suggested this in the PR where setup.mjs was run; unfortunately I hadn't read that script until later.

leotm: as a sidenote, think `pod:install` is redundant, so we can delete it too right?

https://github.com/MetaMask/metamask-mobile/blob/712d6ad62c0edcf17fa6e654b7c3850e20a19e29/package.json#L10

i didn't spot `setup:node`, so i've been manually deleting this part to skip iOS 😅

https://github.com/MetaMask/metamask-mobile/blob/712d6ad62c0edcf17fa6e654b7c3850e20a19e29/scripts/setup.mjs#L113-L127

Issue #9492, opened 2024-05-01
Do not implement RPC methods ahead of the permission middleware


rekmarks:
Ref: https://github.com/MetaMask/metamask-extension/issues/24331

In the extension, some RPC methods, e.g. `wallet_switchEthereumChain` and `wallet_addEthereumChain`, are implemented ahead of the permission middleware. This makes it difficult to keep track of our restricted vs. unrestricted methods, while also literally circumventing our permission system. We are going to stop doing this in the extension (see above issue), and, once mobile has implemented the permission middleware, we shouldn't do it here either. Instead, we should reorder our `json-rpc-engine` middleware stack such that every method hits the permission middleware, and all unrestricted methods have to be enumerated in an unrestricted methods list.

rekmarks:
This was undone by #9699. It should be reimplemented per https://github.com/MetaMask/metamask-extension/pull/24472.

Issue #1723, opened 2020-07-27
Receipt callback is not called on Android Metamask Browser


PowerRyder:
When using web3.eth.sendTransaction, the transaction is mined but Reciept callback is called on Android Metamask Browser v0.2.19. On Pc it works fine.

rickycodes:
closing because this is a duplicate of #1469

PR #9521, opened 2024-05-03
fix: Fix/1723 add permission middleware

Cal-L:
## **Description**

This PR changes the following:
- Adds the missing permission middleware to the JsonRpcEngine stack. All RPC methods will now be validated by the PermissionController before it can reach the RPCMethodMiddleware (user facing RPC method handlers)
- Removes metadata from the permitted accounts in the PermissionController subject state. The result returned by `eth_accounts` handler should contain a list of permitted addresses. A migration is included in the changes to migrate the data.

## **Related issues**

Fixes:

#9492 & [#1723](https://github.com/MetaMask/mobile-planning/issues/1723)

### PermissionController state should be successfully migrated
- GIVEN I have a previous version of the app installed
- AND I have accounts 1 and 2 connected to `uniswap.org`
- AND Account 1 is active on the dapp
- WHEN I upgrade the app to the version with this PR's changes
- AND I log in, access the browser tab, see uniswap tab opened
- THEN I should see accounts 1 and 2 are connected with Account 1 being active
- AND I should be able to freely connect or disconnect accounts without any issues

### Un-permitted restricted RPC method should be blocked by permission middleware
- GIVEN I have no accounts connected to the dapp `metamask.github.io/test-dapp`
- WHEN I tap `ETH_ACCOUNTS` button
- THEN I should see an unauthorized error

### Permitted restricted RPC method should be allowed by permission middleware
- GIVEN I have no accounts connected to the dapp `metamask.github.io/test-dapp`
- AND I connect Account 1 to the dapp
- WHEN I tap `ETH_ACCOUNTS` button
- THEN I should see my active+permitted account address

### WC should still work and be able to trigger transaction
- GIVEN I select WC option from uniswap.org
- AND I use the QR reader to scan
- THEN I should receive the connect account prompt
- WHEN my account is connected the the dapp
- THEN I submit a transaction in the dapp
- THEN I should see the transaction prompt in the mobile app

### When upgrading with PermissionController state migration
https://github.com/MetaMask/metamask-mobile/assets/10508597/5ed6e44a-345a-4b07-9fac-8d48959fc7ca

### When calling restricted+un-permitted RPC method
https://github.com/MetaMask/metamask-mobile/assets/10508597/e2b03084-e0cb-42a0-9cab-226a92b6254b

### When calling restricted+permitted RPC method
https://github.com/MetaMask/metamask-mobile/assets/10508597/1478e79a-b99b-4780-8b6e-34143dec3661

### Interaction using WC to connect and prompt transaction
https://github.com/MetaMask/metamask-mobile/assets/10508597/fd12f0bb-e648-4fbf-a83b-c049b3b5e3df


app/core/WalletConnect/extractApprovedAccounts.ts
```diff
@@ -7,20 +7,9 @@ export const extractApprovedAccounts = (
...
-              'address' in account
-            ) {
-              return account.address;
-            }
-            return undefined;
-          })
-          .flat();
+    ?.map((caveat) => {
+      if (Array.isArray(caveat?.value)) {
+        return caveat.value.flat();
```
rekmarks, line 12: It should no longer be possible for `value` to be nested, right? That being the case, we should omit the `.flat()`.

app/core/Permissions/specifications.js
```diff
@@ -60,7 +60,8 @@ export const getCaveatSpecifications = ({ getInternalAccounts }) => ({
 
     decorator: (method, caveat) => async (args) => {
       const allAccounts = await method(args);
-      const res = caveat.value.filter(({ address }) => {
+      const res = caveat.value.filter((address) => {
+        // TODO: Should addresses in allAccounts be checksummed instead of lowercase?
```
rekmarks, line 64: In an ideal world, yes, but we've historically always lowercased them, and shouldn't do anything about that in this PR, anyway.

app/core/Permissions/index.ts
```diff
@@ -100,24 +94,19 @@ export const addPermittedAccounts = (
...
-
-  const newSortedAccounts = orderBy<SelectedAccount>(
-    [...existing.value, ...selectedAccounts],
-    'lastUsed',
-    'desc',
-  );
+  const newPermittedAccounts = [
+    ...addresses,
+    ...existingPermittedAccountAddresses,
+  ];
```
rekmarks, line 100: Recommend defensively deduping this (assumes addresses are cased the same, which for us should mean lowercased):
```suggestion
  const newPermittedAccounts = Array.from(new Set([
    ...addresses,
    ...existingPermittedAccountAddresses,
  ]));
```

app/store/migrations/040.ts
```diff
@@ -0,0 +1,58 @@
...
+          if (
+            isObject(ethAccounts) &&
+            hasProperty(ethAccounts, 'caveats') &&
+            Array.isArray(ethAccounts.caveats)
+          ) {
+            ethAccounts.caveats = ethAccounts.caveats.map((caveat) => ({
+              ...caveat,
+              value: caveat.value.map(
+                ({ address }: { address: string }) => address,
+              ),
```
rekmarks, line 49: Should the new address array be sorted by `lastUsed` from the previous caveat value?

app/core/Permissions/index.ts
```diff
@@ -68,7 +64,7 @@ export const switchActiveAccounts = (hostname: string, accAddress: string) => {
   }
   let newAccounts = [...existingAccounts];
   newAccounts.splice(accountIndex, 1);
-  newAccounts = [{ address: accAddress, lastUsed: Date.now() }, ...newAccounts];
+  newAccounts = [accAddress, ...newAccounts];
```
rekmarks, line 67: Consider defensively deduping this per my suggestion for `newPermittedAccounts` below. If also deduping this, consider extracting the `Array.from(new Set(...))` rigmarole into a function.

app/components/UI/AccountSelectorList/AccountSelectorList.types.ts
```diff
@@ -5,12 +5,7 @@ import { FlatListProps } from 'react-native';
 // External dependencies.
 import { Account, UseAccounts } from '../../hooks/useAccounts';
 
-export interface SelectedAccount {
-  address: string;
-  lastUsed?: number;
-}
-
-export type SelectedAccountByAddress = Record<string, SelectedAccount>;
+export type SelectedAccountByAddress = Record<string, string>;
```
rekmarks, line 8: What are the keys of this type? I'm asking because accounts are usually just identified by their addresses.

app/store/migrations/040.ts
Cal-L: The caveat value (permitted account addresses) is already sorted by `lastUsed`, so no need to sort here

rekmarks: Ahh, I see that now. Word 👍

app/util/general/index.js
```diff
@@ -160,3 +160,31 @@ export const deepJSONParse = ({ jsonString, skipNumbers = true }) => {
...
+  }
+  // Check if every argument is an array
+  arrays.forEach((array, index) => {
+    if (!Array.isArray(array)) {
+      throw new TypeError(`Argument at position ${index} is not an array.`);
+    }
+  });
+
+  // Flatten the arrays
+  const flattenedArray = arrays.flat();
```
rekmarks, line 184: We should probably mention the flattening in the signature of the function, or state that we expect arrays to be flattened.

I think my preference would be the latter, since we're currently only flattening to the default depth of 1 ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)), and there's in any case no way for us to know what the "right" depth is without the consumer telling us, in which case they can flatten their arrays themselves.

app/util/general/index.js
```diff
@@ -160,3 +160,31 @@ export const deepJSONParse = ({ jsonString, skipNumbers = true }) => {
 
   return parsedObject;
 };
+
+/**
+ * Generates an array with unique items from a list of arrays
```
rekmarks, line 165:
```suggestion
 * Generates an array of referentially unique items from a list of arrays.
```

app/util/general/index.js
```diff
@@ -160,3 +160,31 @@ export const deepJSONParse = ({ jsonString, skipNumbers = true }) => {
...
+ *
+ * @param  {...Array} arrays - A list of arrays
+ * @returns {Array} - Returns an array with unique items
+ * @throws {Error} - Throws if arrays is not defined
+ * @throws {TypeError} - Throws if any of the arguments is not an array
+ */
+export const uniqueList = (...arrays) => {
```
rekmarks, line 172:
```suggestion
export const getUniqueList = (...arrays) => {
```

app/util/general/index.test.ts
```diff
@@ -190,3 +191,29 @@ describe('deepJSONParse function', () => {
...
+      expectedErrorMessage,
+    );
+  });
+  it('should return an array with unique items', async () => {
+    const testArray = ['0x1', '0x2'];
+    const testArray2 = ['0x2', '0x3'];
+    const expectedArray = ['0x1', '0x2', '0x3'];
+    expect(uniqueList(testArray, testArray2)).toEqual(expectedArray);
+  });
+  it('should return the same array if all arrays from the arguements are the same', async () => {
```
rekmarks, line 214:
```suggestion
  it('should return the same array if all arrays from the arguments are the same', async () => {
```

app/util/general/index.js
```diff
@@ -160,3 +160,31 @@ export const deepJSONParse = ({ jsonString, skipNumbers = true }) => {
...
+  if (arrays.length === 0) {
+    throw new Error('At least one array must be defined.');
+  }
+  // Check if every argument is an array
+  arrays.forEach((array, index) => {
+    if (!Array.isArray(array)) {
+      throw new TypeError(`Argument at position ${index} is not an array.`);
+    }
+  });
```
rekmarks, line 181: To make this error more useful during debugging, it'd be better if it also included the value or at least the type of the value in its message.

app/core/Permissions/index.ts
```diff
@@ -86,27 +87,27 @@ export const addPermittedAccounts = (
...
-    ...addresses,
-    ...existingPermittedAccountAddresses,
-  ];
+  // No change in permitted account addresses
+  if (
+    newPermittedAccountsAddresses.length ===
+    existingPermittedAccountAddresses.length
+  ) {
+    return existingPermittedAccountAddresses[0];
+  }
```
rekmarks, line 105: Ah, I didn't realize we threw an error in this case before. We should probably `console.error()` here instead; although reapproving a permitted account is harmless, it's still a problem that should be fixed somewhere.

app/core/Permissions/specifications.js
rekmarks: I'd remove this TODO, there are existing issues for it in the extension, e.g.:
- https://github.com/MetaMask/metamask-extension/issues/4869
- https://github.com/MetaMask/metamask-extension/issues/7985

app/util/general/index.js
rekmarks: Doh; I completely misinterpreted what this line was for. You can revert the change where you specify that arrays must be flattened; it's enough to say that the items / elements must be referentially unique.

rekmarks (approved):
LGTM!

Cal-L:
Also ran `pr_regression_e2e_pipeline` - https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/08cea1fe-b7be-4331-9d83-47fcb67b7934

Cal-L:
E2E passed from non-JSdoc change

christopherferreira9:
SDK & WC remain untouched ✅

tommasini (approved):

PR #9566, opened 2024-05-06
fix: Fix Engine context types

Gudahtt:
## **Description**

The `context` property of `Engine` was typed as `any`, effectively disabling type checking for most controller interactions. The `any` has been removed, restoring controller type checking.

## **Related issues**

This problem was introduced in #6749

## **Manual testing steps**

N/A

## **Screenshots/Recordings**

N/A


app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -128,7 +128,7 @@ export class BackgroundBridge extends EventEmitter {
 
     try {
       const pc = Engine.context.PermissionController;
-      const controllerMessenger = Engine.context.controllerMessenger;
+      const controllerMessenger = Engine.controllerMessenger;
```
Gudahtt, line 131: `controllerMessenger` was being added to `Engine.context`, but was only used in this one place. To simplify the types, I have removed it. The same instance was already accessible as `Engine.controllerMessenger`.

app/core/Engine.ts
```diff
@@ -287,6 +287,51 @@ export interface EngineState {
...
+  PreferencesController: PreferencesController;
+  PPOMController: PPOMController;
+  TokenBalancesController: TokenBalancesController;
+  TokenListController: TokenListController;
+  TokenDetectionController: TokenDetectionController;
+  TokenRatesController: TokenRatesController;
+  TokensController: TokensController;
+  TransactionController: TransactionController;
+  SignatureController: SignatureController;
+  ///: BEGIN:ONLY_INCLUDE_IF(snaps)
```
Gudahtt, line 318: The fences were missing from this type, so I added them here. Everything else about snaps is fenced, including other types.

app/core/Engine.ts
```diff
@@ -1207,42 +1224,37 @@ class Engine {
...
-          storageBackend: new RNFSStorageBackend('PPOMDB'),
-          securityAlertsEnabled:
-            initialState.PreferencesController?.securityAlertsEnabled ?? false,
-          state: initialState.PPOMController,
-          nativeCrypto: Crypto as any,
-        });
-        controllers.push(ppomController as any);
-      } catch (e) {
-        Logger.log(`Error initializing PPOMController: ${e}`);
-        return;
```
Gudahtt, line 1244: This return statement was raising a type error, because it meant the constructor would exit early and not finish initializing all instance properties (in particular, the `this.context` property).

cc @jpuri , was there a particular reason we wanted to exit early here? Do we expect initialization to fail in production?

app/core/Engine.ts
```diff
@@ -1273,9 +1285,7 @@ class Engine {
         ...context,
         [controller.name]: controller,
       }),
-      {
-        controllerMessenger: this.controllerMessenger,
```
Gudahtt, line 1277: This is that 2nd reference to the `controllerMessenger` that I referenced in the other comment: https://github.com/MetaMask/metamask-mobile/pull/9566/files#r1592923348

It's removed here

tommasini (dismissed):
🔥

app/core/Engine.ts
```diff
@@ -1418,12 +1434,12 @@ class Engine {
...
     const networkProvider = NetworkController.state.providerConfig;
-    const conversionRate =
+    let conversionRate =
       CurrencyRateController.state?.currencyRates?.[networkProvider?.ticker]
-        ?.conversionRate === null
-        ? 0
-        : CurrencyRateController.state?.currencyRates?.[networkProvider?.ticker]
-            ?.conversionRate;
+        ?.conversionRate;
+    if (conversionRate === null) {
```
Gudahtt, line 1440: This is equivalent to what the nested ternary was doing, but it's now a bit easier to read, and TypeScript's type inference is now able to understand that the value cannot be null.

app/core/Ledger/Ledger.test.ts
```diff
@@ -13,63 +14,96 @@ import Engine from '../../core/Engine';
...
-    getKeyringsByType: mockGetKeyringsByType,
-    getAccounts: mockGetAccounts,
-    persistAllKeyrings: mockPersistAllKeyrings,
-    signTypedMessage: mockSDignTypedMessage,
-    addNewAccountForKeyring: mockAddNewAccountForKeyring,
-  };
+  let ledgerKeyring: LedgerKeyring;
+  let ledgerKeyringClone: LedgerKeyring;
+
+  beforeEach(() => {
```
Gudahtt, line 36: I replaced the module-scoped mocks with mocks that get reset in-between each test, as I ran into some dependencies between tests when trying to fix the types here.

app/core/Ledger/Ledger.test.ts
```diff
@@ -134,17 +172,16 @@ describe('Ledger core', () => {
...
+        MockEngine.context.KeyringController.persistAllKeyrings,
+      ).toHaveBeenCalled();
     });
   });
 
   describe('getDeviceId', () => {
     it('should return deviceId', async () => {
       const value = await getDeviceId();
-      expect(value).toBe('deviceId');
+      expect(value).toBe('deviceIdClone');
```
Gudahtt, line 184: This was a strange one. `getDeviceId` uses `getKeyringsByType` internally which clearly uses the id `deviceIdClone`, but the test was passing previously with the assertion for `deviceId`.

This is because the mock got manipulated by earlier tests. The `should add a keyring if none could be found` test changed the `getKeyringsByType` mock to return an empty array. This led to `getDeviceId` creating a new keyring, which is how we got `deviceId` as the device ID here.

When the mocks were updated to be reset between tests, this changes back to `deviceIdClone` here.

Ideally we'd have coverage for both paths, but this lack of coverage is a pre-existing problem and I'd like to keep this PR more focused on types if possible. Coverage is not worse here at least.

app/core/RPCMethods/RPCMethodMiddleware.test.ts
```diff
@@ -451,13 +444,31 @@ describe('getRpcMethodMiddleware', () => {
 
   describe('wallet_requestPermissions', () => {
     it('can requestPermissions for eth_accounts', async () => {
+      const mockOrigin = 'example.metamask.io';
+      const mockPermission: Awaited<
```
Gudahtt, line 448: I updated the mocks here to be compatible with the expected types

app/core/RPCMethods/utils.ts
```diff
@@ -2,8 +2,11 @@ import { query } from '@metamask/controller-utils';
 import Engine from '../Engine';
 
 export const polyfillGasPrice = async (method: string, params: any[] = []) => {
-  const { TransactionController } = Engine.context;
-  const data = await query(TransactionController.ethQuery, method, params);
+  const ethQuery = Engine.controllerMessenger.call(
```
Gudahtt, line 5: We're getting the `ethQuery` instance from a different controller here than before, but it's effectively the same. The TransactionController got its `ethQuery` instance from the `NetworkController`, so this is just more direct.

I had to change this because the `ethQuery` instance on the TransactionController is private.

app/core/RPCMethods/wallet_watchAsset.ts
```diff
@@ -87,6 +87,7 @@ const wallet_watchAsset = async ({
     asset: {
       address,
       symbol: finalTokenSymbol,
+      // @ts-expect-error TODO: Fix decimal type
```
Gudahtt, line 90: This seems pretty broken. A number is expected, but we're passing in a string

app/core/Engine.ts
legobeat: Would this work here? Would remove the need to make `conversionRate` mutable.

(GH won't allow suggestion here)

```
    const conversionRate =
      CurrencyRateController.state?.currencyRates?.[networkProvider?.ticker]
        ?.conversionRate ?? 0;
```

alternatively, less clean but still nicer IMO:
```
    const controllerRate =
      CurrencyRateController.state?.currencyRates?.[networkProvider?.ticker]
        ?.conversionRate;
    const conversionRate = controllerRate === null
      ? controllerRate
      : 0;
```

Gudahtt: Sure, those both seem reasonable, I'll try them out

Gudahtt: The first suggestion worked, updated here: bbe3f7df8

legobeat (approved):

app/core/Engine.ts
jpuri: Previous version of PPOMController was initialising PPOM instance in constructor and there was a probability that it would fail - in which case it should not stop app initialisation. But there is a change in PPOMController constructor code now. It should be ok to get rid of try/catch bock.

Gudahtt: Great to hear! Thanks for clarifying

app/core/RPCMethods/wallet_watchAsset.ts
tommasini: The experience for the user seems to be working but this is indeed weird

tommasini (approved):

PR #9508, opened 2024-05-02
feat: bump controllers related accounts logic

gantunesr:
## **Description**

**The original PR is https://github.com/MetaMask/metamask-mobile/pull/9318. It was reverted and now the changes are being introduced in this new PR.**

This PR updates several controllers that handle logic related to the wallet and accounts.

- Keyring controller to v13
- Accounts Controller to v11
- Preferences controller v8
 
Some noteworthy points for this changes,

 - Downgrade controller utils to interact with linea sepolia, only function used by preferences controller of controller-utils is `toChecksumHexAddress` and the function is the same on[ v^5](https://github.com/MetaMask/core/blob/add71d194d75a3c720c935f7947d06ee950bd7c7/packages/controller-utils/src/util.ts#L255) and on [v^8](https://github.com/MetaMask/core/blob/a24869b8e7e518f6682058d3eda4c41f0317ccc8/packages/controller-utils/src/util.ts#L263C17-L263C37)
 - Updated flow of create account and import account to update the `ecnryptionLib` that way on the future we can create or import and we will not run the flow  `[recreateVaultWithSamePassword](https://github.com/MetaMask/metamask-mobile/blob/fcfa9d1bd03e16ba18f3e9919b85a854614c2377/app/core/Authentication/Authentication.ts#L110)`, since it was running unnecessarily
 - Updated controllers initialization that had preferencesOnStateChange argument on their constructor (Mostly assets-controllers and PPOM), to have the address checksummed, on this current version PreferencesController starts to listen to KeyringController and that address is now on lower case format, which means that we need to update to a checksummed format for those specific scenarios for the assets controllers keep their data and tokens never disappear from the user for missmatching user address format
- Updated keyring controller patch with the same as extension to prevent the HD wallet disappear, please check the details on this core branch: extension-keyring-controller-v13-patch

## **Related issues**

* Fixes https://github.com/MetaMask/mobile-planning/issues/1689
* Fixes https://github.com/MetaMask/mobile-planning/issues/1697
* Fixes https://github.com/MetaMask/mobile-planning/issues/1693

## **Manual testing steps**

1. Create new wallet
2. Import existing wallet
3. Import private key
4. Upgrade testing 

## **Screenshots/Recordings**

Upgrade from main with Binance Smart Chain Selected, imported account with private key and tokens imported (Multiple chains added, and tokens imported on them also NFT imported on mainnet)

[...]

PR #9570, opened 2024-05-07
chore: Update `@metamask/keyring-controller` to v16

Gudahtt:
## **Description**

The package `@metamask/keyring-controller` has been updated from v13 to v16. See here for the changelog: https://github.com/MetaMask/core/blob/%40metamask/keyring-controller%4016.0.0/packages/keyring-controller/CHANGELOG.md#1600

### Breaking KeyringController changes

Most of the changes did not require any code changes in mobile. The breaking changes in v14 were mainly type changes that didn't break anything for us, and the changes in v15 only impacted the extension (it changed the meaning of the optional `accountCount` parameter to `addNewAccount`, which mobile doesn't use).

The main breaking change is that as of v16, a number of methods no longer return a copy of the `KeyringController` state. We were never relying on this return value, but we had to adjust how a few of these methods were called or mocked to accommodate this change.

### withKeyring method

v16 of the KeyringController introduced a new method for performing custom keyring operations, called `withKeyring`. This is meant as an alternative approach to using `getKeyringForAccount` or `getKeyringsByType` to get a keyring instance, then using that keyring instance directly.

The `withKeyring` method helps with getting the keyring instance, and it ensures that the KeyringController is locked while the operation is performed, and it ensures that the KeyringController state is updated when the operation is finished. This gets rid of a bunch of boilerplate in our QR and Ledger integrations, and adds additional protection against race conditions and non-atomic state changes.

It's not strictly required as part of this update, but this was implemented in the same PR because I was concerned that the other changes in v16 might reveal some pre-existing race conditions that using `withKeyring` would prevent.

### Peer dependency warnings

This update introduces three new `peerDependencies` warnings, each of which I have audited and determined are safe to temporarily ignore.

* `warning "@metamask/keyring-controller > @metamask/keyring-api@6.1.0" has incorrect peer dependency "@metamask/providers@>=15 <17".`
  * The `@metamask/keyring-api` package uses types from the `@metamask/providers` package as part of the dapp client. The dapp client is not used by the `KeyringController`.
* `warning " > @metamask/accounts-controller@11.0.0" has incorrect peer dependency "@metamask/keyring-controller@^13.0.0".`
  * The `@metamask/accounts-controller` package uses the `KeyringTypes` type from `@metamask/keyring-controller`. This type has not changed between v13 and v16.
* `warning " > @metamask/preferences-controller@8.0.0" has incorrect peer dependency "@metamask/keyring-controller@^13.0.0".`
  * The `@metamask/preferences-controller` package uses the `KeyringController` state and `stateChange` event. These have not changed between v13 and v16.

## **Related issues**

Fixes https://github.com/MetaMask/accounts-planning/issues/441

## **Manual testing steps**

Affected workflows may include:
* Lock
* Unlock
* Upgrade app
* Add account
* Remove account
* Connect hardware wallet
* All hardware wallet interactions

Regression testing in these areas would be recommended. I would also advise doing some regression testing in these areas, then restart your device, then unlock it again (to ensure the keyring data was correctly persisted).

### **Before**

N/A

### **After**

TODO


Gudahtt:
This is close to done, I think I need to adopt `withKeyring` before this is completely ready though.

app/core/Ledger/Ledger.ts
```diff
@@ -61,12 +60,12 @@ export const unlockLedgerDefaultAccount = async (
...
 
-  const keyring = await getLedgerKeyring();
-
-  if (isAccountImportReq) {
-    await keyringController.addNewAccountForKeyring(keyring);
-  }
-  const address = await keyring.getDefaultAccount();
+  const address = await withLedgerKeyring(async (keyring: LedgerKeyring) => {
+    if (isAccountImportReq) {
+      await keyringController.addNewAccountForKeyring(keyring);
```
mikesposito, line 65: This will cause a deadlock, we can use `keyring.addAccounts(1)` instead since we don't use the idempotency feature of the `KeyringConctroller.addNewAccountForKeyring` method.

```suggestion
      await keyring.addAccounts(1);
```

(Probably it makes sense to remove `addNewAccountForKeyring` as a method from KeyringController in the future, since KeyringController doesn't even check if the passed keyring instance belongs to the controller)

Gudahtt: Ah great catch! Right, because `withKeyring` operations can be used to call write keyring methods, but not write KeyringController methods.

Gudahtt: I have committed this change, and added a warning to the JSDoc for this method to warn against this: https://github.com/MetaMask/metamask-mobile/pull/9570/commits/fede87fa5a554e3c207268b5e4be013b89306a89

app/core/Ledger/Ledger.ts
```diff
@@ -80,39 +79,36 @@ export const unlockLedgerDefaultAccount = async (
...
  * Get DeviceId from Ledger Keyring
  *
  * @returns The DeviceId
  */
-export const getDeviceId = async (): Promise<string> => {
-  const ledgerKeyring = await getLedgerKeyring();
-  return ledgerKeyring.deviceId;
-};
+export const getDeviceId = async (): Promise<string> =>
+  await withLedgerKeyring(async (keyring: LedgerKeyring) => keyring.deviceId);
```
mikesposito, line 111: Unrelated to this PR, we could make `withKeyring` more efficient at controller level by persisting keyrings only if the operation results in a change of the serialised version of the keyrings.

This would avoid unnecessary encryption steps in cases like this one, where the operation just selects a value

Gudahtt:
@SocketSecurity ignore npm/@spruceid/siwe-parser@2.1.0

New author seems OK, nothing concerning in the recent published packages by this person. They appear to be a member of the same team, an active maintainer.

app/core/Ledger/Ledger.ts
```diff
@@ -126,7 +122,9 @@ export const ledgerSignTypedMessage = async (
   },
   version: SignTypedDataVersion,
 ): Promise<string> => {
-  await getLedgerKeyring();
+  await withLedgerKeyring(async (_keyring: LedgerKeyring) => {
+    // This is just to trigger the keyring to get created if it doesn't exist already
+  });
```
mikesposito, line 127: This is probably also unrelated to this PR as the behavior is unchanged, but is the `ledgerSignTypedMessage` function necessary? 

Shouldn't a user be unable to sign a message with a ledger address in the `from` attribute if the keyring was not present in the first place (also because it would require the user to select it first)? Also, how do we know that the keyring added will have that address in the `form` attribute if it is not yet present?

This is to say, we could probably simply route to `KeyringController.signTypedMessage` as we do with other keyrings

Gudahtt: I guess the only relevant difference is that this function will also create the keyring if it doesn't exist yet.

It's unclear to me why we'd want to create the keyring as part of this interaction though, or as part of any interaction frankly.

app/components/UI/Transactions/index.js
```diff
@@ -552,7 +552,7 @@ class Transactions extends PureComponent {
   };
 
   signLedgerTransaction = async (transaction) => {
-    const ledgerKeyring = await getLedgerKeyring();
+    const deviceId = getDeviceId();
```
Gudahtt, line 555: Missing `await`

```suggestion
    const deviceId = await getDeviceId();
```

app/components/Views/ConnectHardware/SelectHardware/index.tsx
```diff
@@ -25,6 +25,7 @@ import {
 } from '../../../../util/theme';
 import { getNavigationOptionsTitle } from '../../../UI/Navbar';
 import { useMetrics } from '../../../../components/hooks/useMetrics';
+import LedgerKeyring from '@consensys/ledgerhq-metamask-keyring';
```
Gudahtt, line 28: Prefer type import:
```suggestion
import type LedgerKeyring from '@consensys/ledgerhq-metamask-keyring';
```

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -26,6 +26,8 @@ import Logger from '../../../util/Logger';
 import { removeAccountsFromPermissions } from '../../../core/Permissions';
 import { safeToChecksumAddress } from '../../../util/address';
 import { useMetrics } from '../../../components/hooks/useMetrics';
+import { KeyringTypes } from '@metamask/keyring-controller';
+import { MetaMaskKeyring as QRKeyring } from '@keystonehq/metamask-airgapped-keyring';
```
Gudahtt, line 30: Prefer type import:

```suggestion
import type { MetaMaskKeyring as QRKeyring } from '@keystonehq/metamask-airgapped-keyring';
```

app/components/Views/LedgerAccountInfo/index.tsx
```diff
@@ -32,6 +32,7 @@ import ledgerDeviceDarkImage from '../../../images/ledger-device-dark.png';
 import ledgerDeviceLightImage from '../../../images/ledger-device-light.png';
 import { MetaMetricsEvents } from '../../../core/Analytics';
 import { useMetrics } from '../../../components/hooks/useMetrics';
+import LedgerKeyring from '@consensys/ledgerhq-metamask-keyring';
```
Gudahtt, line 35: Prefer type import:
```suggestion
import type LedgerKeyring from '@consensys/ledgerhq-metamask-keyring';
```

app/core/Ledger/Ledger.test.ts
```diff
@@ -49,65 +42,15 @@ describe('Ledger core', () => {
...
-    ]);
-    mockKeyringController.persistAllKeyrings.mockResolvedValue(false);
+    mockKeyringController.withKeyring.mockImplementation(
+      // @ts-expect-error The Ledger keyring is not compatible with our keyring type yet
+      (_selector, operation) => operation(ledgerKeyring),
+    );
     mockKeyringController.signTypedMessage.mockResolvedValue('signature');
   });
 
-  describe('addLedgerKeyring', () => {
```
Gudahtt, line 71: I should add tests for `withLedgerKeyring` to replace these

app/core/Vault.js
```diff
@@ -1,45 +1,34 @@
 import Engine from './Engine';
 import Logger from '../util/Logger';
 import { KeyringTypes } from '@metamask/keyring-controller';
-import { getLedgerKeyring } from './Ledger/Ledger';
+import { withLedgerKeyring } from './Ledger/Ledger';
 
 /**
  * Restores the QR keyring if it exists.
```
Gudahtt, line 7: Add JSDoc parameter, and further explanation of what this does

app/core/Vault.js
```diff
@@ -1,45 +1,34 @@
...
+  try {
+    await KeyringController.restoreQRKeyring(serializedQrKeyring);
+  } catch (e) {
+    Logger.error(e, 'error while trying to get qr accounts on recreate vault');
   }
 };
 
 /**
  * Restores the Ledger keyring if it exists.
  */
```
Gudahtt, line 21: Add JSDoc parameter, and further explanation of what this does

app/core/Ledger/Ledger.test.ts
Gudahtt: Added in 82713f66f

app/core/Vault.js
Gudahtt: Updated here: 45082f45e

app/core/Vault.js
Gudahtt: Update here: 45082f45e

app/core/Ledger/Ledger.ts
gantunesr: I also fail to see the logic behind creating a keyring for an operation in case it doesn't exist, this is a particular pattern for hardware wallets that we should revert. If a keyring is missing then something failed somewhere earlier in the flow

gantunesr (commented):
LGTM, just left some minor comments

app/core/Vault.js
```diff
@@ -139,3 +138,27 @@ export const recreateVaultWithSamePassword = async (
...
+
+/**
+ * Checks whether the given keyring type exists in the given state.
+ *
+ * @param {KeyringControllerState} state - The KeyringController state.
+ * @param {KeyringTypes} type - The keyring type to check for.
+ * @returns Whether the type was found in state.
+ */
+function hasKeyringType(state, type) {
+  return state.keyrings.some((keyring) => keyring.type === type);
```
Gudahtt, line 150: I think this line is causing the e2e test failures

Gudahtt:
```suggestion
  return state?.keyrings?.some((keyring) => keyring.type === type);
```

app/core/Ledger/Ledger.ts
legobeat: Maybe @stanleyyconsensys has some context here?

legobeat (dismissed):

app/core/Vault.js
```diff
@@ -94,16 +87,22 @@ export const recreateVaultWithNewPassword = async (
...
   await KeyringController.createNewVaultAndRestore(newPassword, seedPhrase);
 
-  await restoreQRKeyring(qrKeyring);
-  await restoreLedgerKeyring(ledgerKeyring);
+  if (serializedQrKeyring !== undefined) {
+    await restoreQRKeyring(serializedQrKeyring);
+  }
+  if (serializedLedgerKeyring !== undefined) {
+    await restoreLedgerKeyring(serializedLedgerKeyring);
+  }
```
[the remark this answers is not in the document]
mikesposito: However, that's something that we could also do in a separate PR to tackle all QR keyring interactions

mikesposito (dismissed):
Two small comments, but this looks good!

app/core/Ledger/Ledger.ts
gantunesr: This was done as part of the Keystone integration and then copied into Ledger. It was just a bad solution for a bug that we should find a better solution

Gudahtt:
Oops, I meant to add a recording before marking this as ready for review. I can't test the hardware wallet stuff, but I can show that I've tested HD account and imported account stuff at least (I have already manually tested that).

I'll fix these conflicts and post that recording later today

gantunesr (dismissed):

tommasini (dismissed):
Ow cool this is how it looks like a patch of ESM build, the patch LGTM! 
Cool to see that mjs is the compiled code from ecmascript files cjs is for common js

Curious question, why it creates a chunk for js and a chunk for mjs?

plasmacorral:
I've conducted migration tests on the [QA build of this branch](https://app.bitrise.io/build/86e07e87-db82-483b-a361-6a552ef0b98f?tab=artifacts) from the following sources:
- QA build of version 7.12.0 (build 1202), initially believed to be the release-approved version but later identified as RC 1.
- QA build of version 7.20.1 (build 1314).

The testing was performed across four different physical devices:
- iPhone Xs running iOS 17.4.1
- iPhone 13 mini running iOS 15.6.1
- Samsung a515f running Android 12 (security patch dated Jan 1, 2024)
- Pixel 5a running Android 14 (security patch dated Apr 5, 2024)

### Observations:

#### Critical Issues Identified:
1. **Ledger Transaction Issue**: Encountered an inability to transact using Ledger across all devices. This was tested using a test dApp, as well as attempts to sign into Opensea and create a Safe via defisaver.com. Instead of the expected `Looking for device` message, an `Unexpected error occurred` message was displayed. This issue was not present in the QA build of the main commit (80fb30f938b555aed3870857e3432c608872743a), suggesting a potential regression in this build. **This is likely a blocker.**
Ledger Transaction Issue

2. **Private Key Import Error**: Attempts to import a private key resulted in a `Something went wrong` message, although the account did appear in the account list. This issue was consistent across iOS and Android devices. Notably, importing a private key via QR code was successful in the QA build of the main commit (80fb30f938b555aed3870857e3432c608872743a) and with build 1326 from TestFlight, despite a previously reported invalid deeplink error.
Private Key Import Error

3. **QR Account Labels Issue**: An existing issue with QR account labels was observed, which was not introduced in this build. This issue was previously identified and has been reported in a new issue [here](https://github.com/MetaMask/metamask-mobile/issues/9592).

### Positive Outcomes:

- Successful migration testing on all 4 devices with HD, Imported, Ledger, and QR accounts, preserving all accounts and custom labels.
- Ability to add and remove Imported and Hardware accounts post-migration.
- Functionality to add HD accounts and rename accounts post-migration was intact.
- Password change, SRP reveal, and private key reveal functionalities worked as expected on all devices.
- Transactions using Keystone for SIWE, Sign permit, and typed data, as well as deployment of ERC20, ERC712, and ERC1155 contracts were successful.
- Post-testing device reboots did not affect the presence of accounts or custom labels upon unlocking.

### Recommendation:
For future iterations, incorporating a custom build number could facilitate a broader range of migration path tests, especially considering the current build (1317) limits testing migration from version 7.22.0 (builds >1321), which could affect a significant user base upon release.

app/core/Ledger/Ledger.ts
```diff
@@ -60,15 +62,12 @@ export const unlockLedgerDefaultAccount = async (
...
-
-  if (isAccountImportReq) {
-    // @ts-expect-error The Ledger keyring isn't compatible with our keyring type yet
-    await keyringController.addNewAccountForKeyring(keyring);
-  }
-  const address = await keyring.getDefaultAccount();
+  const address = await withLedgerKeyring(async (keyring: LedgerKeyring) => {
+    if (isAccountImportReq) {
+      await keyring.addAccounts(1);
+    }
```
mikesposito, line 68: Nit: Looking at [LedgerKeyring](https://github.com/Consensys/ledgerhq-metamask-keyring/blob/97b340cc20695a748369d32eb6a086c4d253b0b6/src/index.ts#L165), `getDefaultAccount` already adds an account in case there is none, so I don't think there is a need to pass `isAccountImportReq` to this function since it could potentially lead to having two accounts involuntarily.

However, again, this had this behavior already before this PR, so no need to address this here

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -227,12 +227,12 @@ const RootRPCMethodsUI = (props) => {
...
         // For Ledger Accounts we handover the signing to the confirmation flow
         if (isLedgerAccount) {
-          const ledgerKeyring = await getLedgerKeyring();
+          const deviceId = await getDeviceId();
 
           props.navigation.navigate(
             ...createLedgerTransactionModalNavDetails({
               transactionId: transactionMeta.id,
-              deviceId: ledgerKeyring.deviceId,
+              deviceId,
```
mikesposito, line 235: Hmm, I suspect that this `deviceId` will always be an empty string, since it is supposed to be passed to the keyring constructor or through `deserialize` (by the consumer), but it seems like it has always been like this, so probably unrelated to the Ledger signing issues we are seeing

mikesposito:
The import account flow still uses the old return that used to include the mem state: (`app/util/address/index.js`)

The function `importAccountFromPrivateKey`:
```typescript
  const { importedAccountAddress } =
    await KeyringController.importAccountWithStrategy('privateKey', [pkey]);
```

Should use, instead:
```typescript
  const importedAccountAddress =
    await KeyringController.importAccountWithStrategy('privateKey', [pkey]);
```

This would explain the error during private key import

mikesposito:
Nit (here or another PR): we could make [this function](https://github.com/MetaMask/metamask-mobile/blob/58c377ad4690a448d17bbf315f1a1882b6a5ab89/app/util/importAdditionalAccounts.js#L32-L33) to recreate primary accounts with balance "without update" better with `withKeyring`

app/components/Nav/Main/RootRPCMethodsUI.js
mikesposito: Ah, no. The `deviceId` comes from here: https://github.com/MetaMask/metamask-mobile/blob/9d977a7eec96dc9aece5ab8ef09faa13fbd36000/app/components/Views/LedgerConnect/index.tsx#L327-L328

mikesposito:
> Ledger Transaction Issue: Encountered an inability to transact using Ledger across all devices.

@plasmacorral could you confirm that the Ledger device connection (to add an account) was successful, thus indicating a problem related to transaction/message signing and not to the general ability to connect to the device?

plasmacorral:
> > Ledger Transaction Issue: Encountered an inability to transact using Ledger across all devices.
> 
> @plasmacorral could you confirm that the Ledger device connection (to add an account) was successful, thus indicating a problem related to transaction/message signing and not to the general ability to connect to the device?

Yes, I can. Observed no problems in pairing. The issue occurs when looking for the device to sign. To quickly see this on a zero balance address, try a v4 typed data signature with the test dapp.

app/core/Ledger/Ledger.ts
mikesposito: Looking at `ledgerSignTypedMessage` again, it appears unused in the codebase

app/components/Views/confirmations/Approval/index.js
```diff
@@ -469,14 +469,14 @@ class Approval extends PureComponent {
 
       // For Ledger Accounts we handover the signing to the confirmation flow
       if (isLedgerAccount) {
-        const ledgerKeyring = await getLedgerKeyring();
+        const deviceId = getDeviceId();
```
mikesposito, line 472: Missing `await` here:

```suggestion
        const deviceId = await getDeviceId();
```

This should fix the signing issue with Ledger

app/core/Vault.js
```diff
@@ -94,16 +87,22 @@ export const recreateVaultWithNewPassword = async (
   const hdKeyring = KeyringController.state.keyrings[0];
   const existingAccountCount = hdKeyring.accounts.length;
 
-  const ledgerKeyring = await getLedgerKeyring();
-  const qrKeyring = (
-    await KeyringController.getKeyringsByType(KeyringTypes.qr)
-  )[0];
+  const serializedLedgerKeyring = hasKeyringType(KeyringTypes.ledger)
+    ? getSerializedKeyring(KeyringTypes.ledger)
```
mikesposito, line 91: Missing `await` 

```suggestion
    ? await getSerializedKeyring(KeyringTypes.ledger)
```

app/core/Vault.js
```diff
@@ -94,16 +87,22 @@ export const recreateVaultWithNewPassword = async (
...
+    : undefined;
+  const serializedQrKeyring = hasKeyringType(KeyringTypes.qr)
+    ? getSerializedKeyring(KeyringTypes.qr)
```
mikesposito, line 94: Missing `await` 

```suggestion
    ? await getSerializedKeyring(KeyringTypes.qr)
```

mikesposito:
@SocketSecurity ignore npm/@metamask/keyring-api@6.0.0

This is our package

mikesposito:
I'm unable to test this locally, but [135e572](https://github.com/MetaMask/metamask-mobile/pull/9570/commits/135e572569c27ae5393639959bb43036c894884c) should fix the issue with Ledger (1) and [c5bd3db](https://github.com/MetaMask/metamask-mobile/pull/9570/commits/c5bd3db4ad2f12e03b21f1367c99ca1967b8b5a8) should fix the error when importing an account (2) 

> QR Account Labels Issue: An existing issue with QR account labels was observed, which was not introduced in this build.

I assume we want to fix (3) separately instead

vivek-consensys:
@mikesposito @gantunesr I am unable to verify the personal sign using the test dapp on iPhone 13 (iOS 14.3.1), which was tested using the [bitrise build](https://app.bitrise.io/app/be69d4368ee7e86d/build/8922f487-b3c7-4c8b-a495-fdbf705a04fb/artifact/a2ddfbbf309d24b7/p/c866f05221740ad8073b4799f664af2c). The verify button does not result in any of the recovery results appearing:- 

image

app/core/Vault.js
```diff
@@ -94,16 +87,22 @@ export const recreateVaultWithNewPassword = async (
...
-  const ledgerKeyring = await getLedgerKeyring();
-  const qrKeyring = (
-    await KeyringController.getKeyringsByType(KeyringTypes.qr)
-  )[0];
+  const serializedLedgerKeyring = hasKeyringType(KeyringTypes.ledger)
+    ? await getSerializedKeyring(KeyringTypes.ledger)
+    : undefined;
+  const serializedQrKeyring = hasKeyringType(KeyringTypes.qr)
+    ? await getSerializedKeyring(KeyringTypes.qr)
+    : undefined;
```
mikesposito, line 95: This function is most likely failing to recreate hardware keyrings when changing password because `hasKeyringType` always returns false, because we should pass the KeyringController instance state:

```suggestion
  const serializedLedgerKeyring = hasKeyringType(
    KeyringController.state,
    KeyringTypes.ledger,
  )
    ? await getSerializedKeyring(KeyringTypes.ledger)
    : undefined;
  const serializedQrKeyring = hasKeyringType(
    KeyringController.state,
    KeyringTypes.qr,
  )
    ? await getSerializedKeyring(KeyringTypes.qr)
    : undefined;
```

mikesposito: Alternatively, we can mimic `getSerializedKeyring` by selecting the `KeyringController.state` in `hasKeyringType` itself

mikesposito:
> QR accounts are being duplicated when closing the app and reopening

Hi @vivek-consensys, looking at your screen recording it looks like QR accounts are not being duplicated during login (app restart). Were they duplicated in another step (e.g. QR Sync/sign), or is it just happening inconsistently?

Gudahtt:
>When changing the password, the hardware accounts are removed.

The latest commit should fix this problem

vivek-consensys:
> > When changing the password, the hardware accounts are removed.
> 
> The latest commit should fix this problem

This has been fixed using latest build, screen recording shows:

https://github.com/MetaMask/metamask-mobile/assets/106310394/88e90101-caec-4029-a394-470cf50c81bd

vivek-consensys:
2 issues found:- 

1. When updating the account name, then changing password, the updated account name is not reflected

https://github.com/MetaMask/metamask-mobile/assets/106310394/641cdb79-dcf1-47c7-a9c0-b358bffedc47

2. When the user clicks on Add hardware wallet -> QR-based -> Continue -> Only this time (camera permission) -> once the camera opens go back to home and add QR hardware wallet -> The camera won't open with no errors

https://github.com/MetaMask/metamask-mobile/assets/106310394/159bec03-9b1c-4ee0-9c16-c8568f754786

Gudahtt:
Thanks @vivek-consensys ! Could you verify whether those bugs exist already on `main`? The connection between that behavior and this PR is not clear to me. We don't change anything close to the hardware wallet connect flow in the UI, and the KeyringController doesn't store those account labels.

Gudahtt:
Actually, for #2, @mikesposito has found a possible explanation (the permission failure might cause a rollback, which causes some existing keyring event subscriptions to be stale). We have a solution, and I will push it up shortly

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -118,15 +120,15 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
   }, []);
 
   useEffect(() => {
-    let memStore: any;
-    KeyringController.getQRKeyringState().then((_memStore: any) => {
```
mikesposito, line 122: This step used to add the QRKeyring under the hood, is it added somewhere else before approaching this view?

Gudahtt: Good question. This subscription should work regardless, but yes we should avoid any functional change here.

Gudahtt: Addressed in https://github.com/MetaMask/metamask-mobile/pull/9570/commits/03673ea5079c0d9e9c41134e22f9a5e87e1ac97e

Technically there is still a functional change here, in the order of events. But the subscription should be valid even if it gets setup first, so this change in order of events should have no impact.

mikesposito: ah right, we use `KeyringController.connectQRHardware` and `KeyringController.unlockQRHardware` in this same view, which also create the keyring

mikesposito: I agree that the subscription will be valid regardless, since it is on the messenger

Gudahtt:
@SocketSecurity ignore npm/@metamask/snaps-sdk@4.0.1

This is our package

app/components/UI/QRHardware/withQRHardwareAwareness.tsx
```diff
@@ -22,6 +22,11 @@ const withQRHardwareAwareness = (
     };
 
     useEffect(() => {
+      // This ensures that a QR keyring gets created if it doesn't already exist.
+      // This is intentionally not awaited (the subscription still gets setup correctly if called
```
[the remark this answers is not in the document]
mikesposito: In any case, the subscription will be valid regardless, and the keyring will be effectively created in this same view when calling `connectQRHardware` and `unlockQRHardwareWalletAccount`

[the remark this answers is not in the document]
mikesposito: I'm not sure how `async-mutex` works under the hood, but my concern was that if a Promise that should lock a mutex is not awaited it would never be able to acquire the mutex, so essentially never executed (and never resolving). What you are suggesting instead is that it would acquire the mutex regardless when available and execute as usual, but it would not block this `useEffect`?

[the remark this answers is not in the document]
Gudahtt: Yes exactly. In general it's not possible for a Promise-returning function to know if that Promise is awaited or not. All Promises are executed eagerly. So this will run just as if it were awaited, except that it won't block.

Gudahtt:
To recap about these two issues: https://github.com/MetaMask/metamask-mobile/pull/9570#issuecomment-2114959664

I don't know the cause of the first, but I suspect it may be unrelated to this PR. For the second, we hope to have fixed this in the latest commit.

vivek-consensys:
@Gudahtt it seems the second issue still persists with the latest commits using this [bitrise build ](https://app.bitrise.io/build/5b027afe-cb5f-442b-bf62-212c7cd4ba72)


https://github.com/MetaMask/metamask-mobile/assets/106310394/3fa55ab8-62dd-4adc-a912-088cf5024eeb

Gudahtt:
>2. When the user clicks on Add hardware wallet -> QR-based -> Continue -> Only this time (camera permission) -> once the camera opens go back to home and add QR hardware wallet -> The camera won't open with no errors


>@Gudahtt it seems the second issue still persists with the latest commits using this [bitrise build](https://app.bitrise.io/build/5b027afe-cb5f-442b-bf62-212c7cd4ba72)

I noticed that this second video is different from the first. In the first, the camera doesn't open at all. In the second, the camera opens briefly and says "scanning", then closes shortly after, bringing you back to the previous screen (and no QR code was scanned).

Still broken, but not in quite the same way

app/components/UI/QRHardware/withQRHardwareAwareness.tsx
```diff
@@ -23,15 +22,20 @@ const withQRHardwareAwareness = (
...
-      const { KeyringController } = Engine.context as any;
-      KeyringController.getQRKeyringState().then((store: any) => {
-        keyringState.current = store;
-        keyringState.current.subscribe(subscribeKeyringState);
-      });
+      // This ensures that a QR keyring gets created if it doesn't already exist.
+      // This is intentionally not awaited (the subscription still gets setup correctly if called
+      // before the keyring is created).
+      // TODO: Stop automatically creating keyrings
+      Engine.context.KeyringController.getOrAddQRKeyring();
```
Gudahtt, line 29: So this is a bit different functionally than before. A closer equivalent would be something like:

```TypeScript
const isSubscribedRef = useRef<boolean>();
...
useEffect(() => {
  (async () => {
    await Engine.context.KeyringController.getOrAddQRKeyring();
    isSubscribedRef.current = true;
    Engine.controllerMessenger.subscribe(
      'KeyringController:qrKeyringStateChange',
      subscribeKeyringState,
    );
  })();
  return () => {
    if (isSubscribedRef.current) {
      isSubscribedRef.current = false;
      Engine.controllerMessenger.unsubscribe(
        'KeyringController:qrKeyringStateChange',
        subscribeKeyringState,
      );
    }
  };
}, []);
```

The difference is that this delays the subscription in a similar manner, until after we've confirmed that we have a QR Keyring. Maybe this skips a problematic state update. Why would this matter? No idea. But it's a possible difference at least.

I'm hesitant to even suggest testing this because I don't think this is a likely explanation for the bug, nor a good change in general. Probably the bug is related to some change in the KeyringController itself, which we'd get more insight into by adding logging and doing further debugging.

Gudahtt:
The issue found with connecting QR hardware wallets should be resolved as of https://github.com/MetaMask/metamask-mobile/pull/9570/commits/edf67f23ecf9a40fcaf7b7872dc2a2d135f14a10

Ready for testing again @vivek-consensys @gantunesr

vivek-consensys:
> 2 issues found:-
> 
> 1. When updating the account name, then changing password, the updated account name is not reflected
> 
>  Screen_Recording_20240516_114208_MetaMask-QA.mp4 
> 2. When the user clicks on Add hardware wallet -> QR-based -> Continue -> Only this time (camera permission) -> once the camera opens go back to home and add QR hardware wallet -> The camera won't open with no errors
> 
>  Screen_Recording_20240516_113338_MetaMask-QA.mp4

First issue still persists with the new [Bitrise build](https://app.bitrise.io/build/9bbb17b5-3c6b-40a2-857d-61f057f0fdd1?tab=artifacts). Second issue is fixed. Screen recordings below: 

1.  

https://github.com/MetaMask/metamask-mobile/assets/106310394/0b333f48-4add-4e0a-9da1-6a3cd1b40139

2. 


https://github.com/MetaMask/metamask-mobile/assets/106310394/ffb668ca-dd3b-4950-92cf-bdfca4d7148b

Gudahtt:
I have investigated the one remaining issue, which is that account labels (perhaps just hardware wallet accounts? unsure) disappear after changing password.

It looks like we have never had any logic for preserving these account labels, but they are preserved right now in production because they reside in the PreferencesController, which doesn't notice that the accounts were removed then re-added. The PreferencesController became synchronized with the KeyringController in `@metamask/preferences-controller@7.0.0`,  which was introduced to mobile recently in https://github.com/MetaMask/metamask-mobile/pull/9508. I suspect that PR introduced this problem, and that it will be found on `main` as well (and in 7.23.0).

Rather than stash and restore this data, I'd recommend that we address this bug by updating how "reset password" works to preserve the vault during the process. Currently we throw away then recreate the vault to change the password, but we don't need to do that. @mikesposito recently introduced a new KeyringController method for changing the password without discarding any state, in this PR: https://github.com/MetaMask/core/pull/4279. This bug will be resolved when we introduce that functionality.

In any case, since it was not introduced by this PR, I don't think it should block. All issues resolved, @vivek-consensys please take another look when you get the chance 🙏

vivek-consensys:
Testing the [bitrise build](https://app.bitrise.io/build/041e1bd3-a36e-4bff-902f-cc27ce1e0fe4?tab=artifacts), found an issue where the balance is not reflecting on the homepage and the account list for a hardware device (Ledger)

https://github.com/MetaMask/metamask-mobile/assets/106310394/03c4fa30-e230-4351-89cd-2dc92a512e9a

vivek-consensys:
> I have investigated the one remaining issue, which is that account labels (perhaps just hardware wallet accounts? unsure) disappear after changing password.
> 
> It looks like we have never had any logic for preserving these account labels, but they are preserved right now in production because they reside in the PreferencesController, which doesn't notice that the accounts were removed then re-added. The PreferencesController became synchronized with the KeyringController in `@metamask/preferences-controller@7.0.0`, which was introduced to mobile recently in #9508. I suspect that PR introduced this problem, and that it will be found on `main` as well (and in 7.23.0).
> 
> Rather than stash and restore this data, I'd recommend that we address this bug by updating how "reset password" works to preserve the vault during the process. Currently we throw away then recreate the vault to change the password, but we don't need to do that. @mikesposito recently introduced a new KeyringController method for changing the password without discarding any state, in this PR: [MetaMask/core#4279](https://github.com/MetaMask/core/pull/4279). This bug will be resolved when we introduce that functionality.
> 
> In any case, since it was not introduced by this PR, I don't think it should block. All issues resolved, @vivek-consensys please take another look when you get the chance 🙏

As per @Gudahtt comment above on the remaining issue, there are no further issues found on this PR.

gantunesr (approved):

Gudahtt:
I hate to bypass the unit test quality gate, but the alternatives seem worse in this case. The main two places where coverage has dropped here are `Vault.js` and `app/components/Views/ConnectQRHardware/index.tsx`. Both files have extensive interactions with the `KeyringController`. However I had trouble using the `KeyringController` in a unit test because I was unable to override the built-in keyring builders for the Simple and HD keyrings, and mocking the `KeyringController` interactions proved to be too complex (I got it working locally, but the tests were horrifically complicated and fragile, and probably would do more harm than good).

Given the urgency of this update, I'm going to recommend that we bypass the quality gate check for now. But I have a PR on core to address the issue I ran into, and will follow up soon with unit tests for `Vault.js` to make up for this.

OGPoyraz (approved):

Gudahtt:
E2E tests are failing, but the failures are from `main`, not introduced by this PR.

e.g. see this run from `main`: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/512fae62-73ae-4d5c-a082-ec220e4a4de8

tommasini (approved):

PR #9627, opened 2024-05-14
fix: fix/1723 broken sign verification on test dapp

Cal-L:
## **Description**

The PR that introduced permission controller middleware https://github.com/MetaMask/metamask-mobile/pull/9521, broke verifying personal sign on the test dapp. This PR resolves a few areas in the app that needed to use lowercased account addresses. Another note: Since permission middleware now handles `eth_accounts`, we migrated a condition to use the correct origin based on SDK, WC, or dapp in the `BackgroundBridge`. This code is a temporary fix and should be refactored in the future. To keep this PR smaller, we can follow up with removing `eth_accounts` from RPCMethodMiddleware in another PR.

## **Related issues**

Fixes: [#1723](https://github.com/MetaMask/mobile-planning/issues/1723), https://app.zenhub.com/workspaces/mobile-release-testing-6249e5242464b70013315a98/issues/gh/metamask/metamask-mobile/9628

## **Manual testing steps**

Before fix:
1. Connect to the MM test dapp
2. Trigger personal sign
3. Trigger verify
4. Results will not appear

After fix:
1. Connect to the MM test dapp
2. Trigger personal sign
3. Trigger verify
4. Results should show address used to sign
5. Switch/connect another account
6. Trigger personal sign
4. Results should show address used to sign

### **Before**

https://github.com/MetaMask/metamask-mobile/assets/10508597/e38a71b5-6ac0-40ea-b3de-36b38830e7c9

### **After**

Signing verification works even after switching accounts
https://github.com/MetaMask/metamask-mobile/assets/10508597/5c43af3b-7c9f-456b-a870-a5345027642b

WC connects and is able to trigger a transaction

https://github.com/MetaMask/metamask-mobile/assets/10508597/7009ac14-fa87-4ebd-88e0-022f2f30264e


Cal-L:
Will run E2E on release branch

app/core/Permissions/index.ts
```diff
@@ -176,7 +176,7 @@ export const getPermittedAccounts = async (
         hostname,
         RestrictedMethods.eth_accounts,
       );
-    return accounts;
+    return accounts.map((account: string) => account.toLowerCase());
```
Gudahtt, line 179: Nit: Hmm, we don't need to lower-case anything on the extension equivalent. I wonder where it's getting checksummed.

Gudahtt: Actually I'm not sure that this fixes the issue. It seems that the test dapp is using the `accountsChanged` event from the provider to populate its list of accounts, and this line doesn't impact what that account returns.

Instead that happens here: https://github.com/MetaMask/metamask-mobile/blob/3b1fc28ce879ab8078d4ac46d2b3af050d9b4777/app/core/BackgroundBridge/BackgroundBridge.js#L241

See that `approvedAccounts` is returned directly, and it starts with `selectedAddress`, which is expected to be checksummed (at least in how it's used - see the `selectedAddress.toLowerCase()` call in the `found` expression)

Gudahtt: I think the fix here is to return `permittedAccounts` directly in that `notifySelectedAddressChanged` function. `permittedAccounts` should already sort the selected address first.

Gudahtt:
I have tried testing this PR locally, and it does not appear to work. However I did confirm that #9521 seems to introduce the problem.

app/core/Permissions/index.ts
Gudahtt: I think I found the problem. The `BackgroundBridge` method `sendNotification` is called directly from the `BrowserTab` module, with the `accountsChanged` event.  It is getting that list of accounts from a selector that uses `getPermittedAccountsByHostname`, and that method doesn't do any normalization. That's what changed in #9521

Gudahtt: ^ That just explains why the `accountsChanged` event returns checksummed accounts. But you're right that `eth_accounts` also returns checksummed accounts.

This is because of the caveat specification, which differs from the one used for the extension. The mobile caveat `restrictReturnedAccounts` will return the caveat value directly, and that value is an array of checksummed addresses. It should instead do what the extension version of this specification does, which is return the original result but _filter_ by the caveat value.

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -389,7 +389,11 @@ export class BackgroundBridge extends EventEmitter {
     // Add PermissionController middleware
     engine.push(
       Engine.context.PermissionController.createPermissionMiddleware({
-        origin,
+        // FIXME: This condition migrates `eth_accounts` from RPCMethodMiddleware so that both WC and SDK are compatible with the permission middleware.
+        // This is not a long term solution. BackgroundBridge should be not contain hardcoded logic pertaining to WC, SDK, or browser.
+        origin: this.isWalletConnect
```
Gudahtt, line 394: This is a bit confusing. Why do we have a double-ternary here if there are only two possible values that we want to set? And why set it here, rather than above on like 364 where origin is declared? As-is, the origin would be inconsistent with permissions everywhere else in the RPC pipeline now. And arguably "less correct", since the SDK passes in a constant for the URL, which would give our middleware the impression that all SDK connections are from the same source.

Gudahtt: Perhaps we can leave this for a different PR. The SDK permission system integration seems to be in a bad state right now for multiple reasons.

app/core/Permissions/specifications.js
```diff
@@ -59,13 +59,16 @@ export const getCaveatSpecifications = ({ getInternalAccounts }) => ({
...
         const addressToCompare = address.toLowerCase();
-        return allAccounts.includes(addressToCompare);
+        const isPermittedAccount = allAccounts.includes(addressToCompare);
+        if (isPermittedAccount) {
+          permittedAccounts.push(addressToCompare);
+        }
       });
-
-      return res;
+      return permittedAccounts;
```
Gudahtt, line 71: Nit: We should just copy what the extension does here. It's much more concise:

```
const result = await method(args);
return result.filter((account) => caveat.value.includes(account));
```

app/components/UI/AccountSelectorList/AccountSelectorList.tsx
```diff
@@ -154,7 +154,12 @@ const AccountSelectorList = ({
         : CellVariant.Select;
       let isSelectedAccount = isSelected;
       if (selectedAddresses) {
-        isSelectedAccount = selectedAddresses.includes(address);
+        const lowercasedSelectedAddresses = selectedAddresses.map(
```
Gudahtt, line 157: Hmm, how is this change related to the signature verification bug fix?

Cal-L: This piece was affected since we updated `getPermittedAccountsByHostname` to return lowercased account addresses. `selectedAddresses`, which was derived from `getPermittedAccountsByHostname` used to contain checksummed addresses. The logic here renders the highlight around the active account for the dapp.

image

tommasini:
Should we target main and then cherry pick it to the release?

app/core/BackgroundBridge/BackgroundBridge.js
Cal-L: I carried over the double-ternary from `eth_accounts` as is from RPCMethodMiddleware https://github.com/MetaMask/metamask-mobile/blob/953086c7a8fcb8a03b48d5a5b1f697f97e5f9d82/app/core/RPCMethods/RPCMethodMiddleware.ts#L300C18-L300C73 since I didn't want to change the existing functionality of WC/SDK. Prior to the introduction of PS middleware, the origin used through the json rpc engine stack didn't contain the double-ternary condition either so I didn't want to risk how origin was handled prior to being handled by eth_accounts. I'll defer to @abretonc7s for context behind setting up the double-ternary this way.

app/core/Permissions/specifications.js
Gudahtt: > Since result aka getAllAccounts isn't ordered in the same manner

Hmm, it should be ordered in this manner. That's what we're doing here: https://github.com/MetaMask/metamask-mobile/blob/953086c7a8fcb8a03b48d5a5b1f697f97e5f9d82/app/core/Permissions/specifications.js#L137

We're sorting by "last selected", so the selected account (or most recently selected) will always be first.

app/core/BackgroundBridge/BackgroundBridge.js
Gudahtt: Gotcha, OK. Well in the interest if minimizing changes for now, perhaps we could do this more concisely with a single ternary: `isMMSDK ? this.channelId : origin`

app/core/Permissions/specifications.js
Cal-L: Discussed that mobile has decoupled wallet address vs dapp permitted addresses. "Active" address for dapps is determined by order of permitted addresses in PS

app/components/UI/AccountSelectorList/AccountSelectorList.tsx
Cal-L: Acknowledged

app/core/BackgroundBridge/BackgroundBridge.js
Cal-L: Will update!

Cal-L: Updated

app/core/Permissions/index.ts
Cal-L: Resolving this in favor of threads below, which addresses some of the questions

Cal-L:
Closing in favor of https://github.com/MetaMask/metamask-mobile/pull/9656

PR #9748, opened 2024-05-24
fix: QR scan issue

mikesposito:
## **Description**

Alternative solution to QR scan issue on #9570 

## **Related issues**

Related to #9570

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Screenshots/Recordings**

https://github.com/MetaMask/metamask-mobile/assets/34438276/a208bbdb-958b-4458-abc8-aa5a58f7335b


app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -101,6 +96,14 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
   const [existingAccounts, setExistingAccounts] = useState<string[]>([]);
 
+  useEffect(() => {
+    KeyringController.connectQRHardware(0).then(
+      (connectedAccounts: Awaited<ReturnType<QRKeyring['getFirstPage']>>) => {
+        setAccounts(connectedAccounts);
+      },
+    );
+  }, [KeyringController]);
+
```
mikesposito, line 106: This one is only used to show the pagination view directly instead of the scanner in case there's a QR device connected already

mikesposito: ^ actually, it is also used to show the pagination view right after `submitQRCryptoHDKey` is called.

mikesposito: Though, I suspect there is a more elegant way of doing this

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -115,55 +118,24 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
     });
   }, [KeyringController]);
 
-  const subscribeKeyringState = useCallback((storeValue: any) => {
-    setQRState(storeValue);
-  }, []);
```
mikesposito, line 120: Using the internal keyring state to open and close the modal seems unnecessarily complex and confusing.

We can open the modal at the button click, and close it on scan completion/failure without any input coming from the keyring

app/components/Views/ConnectQRHardware/index.tsx
mikesposito: In `[09ae100](https://github.com/MetaMask/metamask-mobile/pull/9748/commits/09ae100d9d422ffa88e7b8878f7dbbbc77333479)` this uses a clearer approach (I believe) using `withKeyring` - so that we progressively abandon KeyringController's qr-related methods

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -79,16 +79,13 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
   const { trackEvent } = useMetrics();
   const styles = createStyles(colors);
 
-  const KeyringController = useMemo(() => {
+  const KeyringController = useMemo<
+    typeof Engine.context.KeyringController
```
mikesposito, line 83: this missing generic type was making `KeyringController` an `any`

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -115,59 +128,29 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
-
   const onConnectHardware = useCallback(async () => {
     trackEvent(MetaMetricsEvents.CONTINUE_QR_HARDWARE_WALLET, {
       device_type: 'QR Hardware',
     });
     resetError();
-    const _accounts = await KeyringController.connectQRHardware(0);
-    setAccounts(_accounts);
-  }, [KeyringController, resetError, trackEvent]);
+    showScanner();
```
mikesposito, line 136: I would say this is the main change: the scanner is opened directly here, instead of waiting for an event coming from the Keyring

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -115,59 +123,29 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
   const onScanSuccess = useCallback(
-    (ur: UR) => {
+    async (ur: UR) => {
       hideScanner();
       trackEvent(MetaMetricsEvents.CONNECT_HARDWARE_WALLET_SUCCESS, {
         device_type: 'QR Hardware',
       });
       if (ur.type === SUPPORTED_UR_TYPE.CRYPTO_HDKEY) {
-        KeyringController.submitQRCryptoHDKey(ur.cbor.toString('hex'));
+        await KeyringController.submitQRCryptoHDKey(ur.cbor.toString('hex'));
```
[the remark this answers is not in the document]
Gudahtt: Here's the full path that I traced, for more details on this:

`getFirstPage` isn't defined on the child class, but it's on the base here: https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/base-eth-keyring/src/BaseKeyring.ts#L345

which calls `__getPage` here: https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/base-eth-keyring/src/BaseKeyring.ts#L404

I first looked at the `readKeyring` call. That seems like the more likely culprit. When I glanced at `__getNormalPage` it didn't seem to be doing anything that related to the camera. It seems like that's post-initialization, and relies upon the key being present locally already

`readKeyring` is here: https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/base-eth-keyring/src/BaseKeyring.ts#L222

It calls `await this.getInteraction().readCryptoHDKeyOrCryptoAccount();`.

This `getInteraction` is how the camera interactions work I believe. It's defined on the child class here: [https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7[…]4ff5/packages/metamask-airgapped-keyring/src/MetaMaskKeyring.ts](https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/metamask-airgapped-keyring/src/MetaMaskKeyring.ts#L24)

The "MetamaskInteractionProvider` is a separate class in this package, here: [https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7[…]s/metamask-airgapped-keyring/src/MetaMaskInteractionProvider.ts](https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/metamask-airgapped-keyring/src/MetaMaskInteractionProvider.ts)

Here is the `readCryptoHDKeyOrCryptoAmount` method that was referenced earlier: [https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7[…]s/metamask-airgapped-keyring/src/MetaMaskInteractionProvider.ts](https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/metamask-airgapped-keyring/src/MetaMaskInteractionProvider.ts#L60)

See how it uses a Promise constructor that only resolves or rejects based on events. I think those events are the key to this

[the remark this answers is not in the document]
Gudahtt: Sorry, to clarify: I think this solution does avoid the deadlock, but it doesn't ensure that the connection would actually work. We'd have no deadlock, but we'd have an unresolved Promise instead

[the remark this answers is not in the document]
mikesposito: Hmm I think I follow what you are saying: if the promise constructed [here](https://github.com/MetaMask/metamask-mobile/blob/10ff6589f883288179b657088e177e50147b4661/app/components/Views/ConnectQRHardware/index.tsx#L109) is on a keyring instance that is different (due to restore or other) from the instance [we are submitting the HDKey to](https://github.com/MetaMask/metamask-mobile/blob/10ff6589f883288179b657088e177e50147b4661/app/components/Views/ConnectQRHardware/index.tsx#L141), then the first promise we were awaiting for would not resolve (because [this event](https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/metamask-airgapped-keyring/src/MetaMaskInteractionProvider.ts#L87) would be emitted/listened by different instances)

If what I understood is correct, could we call `getFirstPage` and `submitCryptoHDKey` [here](https://github.com/MetaMask/metamask-mobile/blob/10ff6589f883288179b657088e177e50147b4661/app/components/Views/ConnectQRHardware/index.tsx#L141) wrapped by the same `withKeyring` call so that they will always be the same instance?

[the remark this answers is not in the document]
mikesposito: ^ I committed this idea so that it is clearer what I mean to say: bc286cc5caa8690a998c2c329abe6ae4617f09e1

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -101,6 +93,27 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
+      },
+      { createIfMissing: true },
+    );
+  }, [KeyringController]);
+
+  useEffect(() => {
+    // in case there is a QR keyring already connected, this will
+    // show the first page of accounts once the view is loaded
+    connectFirstPage();
+  }, [connectFirstPage]);
```
Gudahtt, line 115: It looks like this will initiate a connection upon render. Could you elaborate on what this is for?

app/components/Views/ConnectQRHardware/index.tsx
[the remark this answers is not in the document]
Gudahtt: Yeah that looks right to me! Good idea.

I left a few other comments, but overall this seems viable to me. Though I can't say for sure because I don't understand how the keyring works internally or why they'd choose to implement this internal-event-based flow.

app/components/Views/ConnectQRHardware/index.tsx
mikesposito: If there is already a QR Keyring with an HDKey (already paired) we want to show the paged account list.

I'm noticing now though that on `main` this happens after tapping on "Connect" even if the keyring is already connected

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -101,6 +93,27 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
 
+  const connectFirstPage = useCallback(async () => {
+    await KeyringController.withKeyring(
+      { type: KeyringTypes.qr },
+      // @ts-expect-error QRKeyring is not yet compatible with EthKeyring type
+      async (keyring: QRKeyring) => {
+        if ((await keyring.serialize()).initialized) {
+          // the first page of accounts is obtained only if
+          // the keyring is already initialized (e.g. there is already an HDKey)
+          setAccounts(await keyring.getFirstPage());
```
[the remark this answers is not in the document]
mikesposito: Shouldn't this promise resolve immediately if `initialized` is `true` (already paired Keyring)?

app/components/Views/ConnectQRHardware/index.tsx
mikesposito: To keep `main`'s behavior this should happen on button tap instead of right after rendering - even though it doesn't make much sense to tap on "Connect" to show an already connected device

app/components/Views/ConnectQRHardware/index.tsx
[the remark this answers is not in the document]
mikesposito: What I want to say is: in the QRKeyring implementation it looks like `this.readKeyring` is not called when `initialized` is `true`: https://github.com/KeystoneHQ/keystone-airgaped-base/blob/e7df2021cfd688bd0365aed7cc1a31cad8c74ff5/packages/base-eth-keyring/src/BaseKeyring.ts#L406C7-L406C32

So it would not end up in the odd promise waiting for events to be fired, but would return the page directly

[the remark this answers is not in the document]
Gudahtt: Ah I see, I missed that, thanks

PR #9718, opened 2024-05-22
chore: upgrade snaps-controller and adapts its usage

Jonathansoufer:
## **Description**

This PR upgrades the snap-controller version to its latest version.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [x] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [x] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #9779, opened 2024-05-29
chore: Update accounts controller v14

tommasini:
## **Description**

This PR updates accounts controller to v^14.

qa build: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/96496314-9441-4329-bd43-fc9e4f1e3dcd

This PR also includes this PR: https://github.com/MetaMask/metamask-mobile/pull/9832

## Blocked by

- [x] https://github.com/MetaMask/metamask-mobile/pull/9718
  This PR is dependent on the snaps controller update [PR](https://github.com/MetaMask/metamask-mobile/pull/9718)
- [x] https://github.com/MetaMask/metamask-mobile/pull/9570
  This PR is dependent on the keyring controller update [PR](https://github.com/MetaMask/metamask-mobile/pull/9570)

## Fixes
- https://github.com/MetaMask/mobile-planning/issues/1777

## **Manual testing steps**

1. Import account via SRP
2. Import account via private key
3. Import tokens 

## **Screenshots/Recordings**

* Import with private key

image

* Importing with SRP

https://github.com/MetaMask/metamask-mobile/assets/46944231/d829d4d6-6e28-404f-b1bb-e0ed3ec6f986


app/util/test/accountsControllerTestUtils.ts
```diff
@@ -59,21 +58,23 @@ export const internalAccount2 = createMockInternalAccount(
   'Account 2',
 );
 
+const internalAcccountsAccounts = {
+  [expectedUuid]: internalAccount1,
+  [expectedUuid2]: internalAccount2,
+};
+
 // used as a default mock for other tests
-export const MOCK_ACCOUNTS_CONTROLLER_STATE: AccountsControllerState = {
```
owencraston, line 63: Nitpick: I want to keep these type explicit in case there is an update in the data type that we expect form the package we can find the issue very easily before it fails on CI.

owencraston:
we should also bump the version of the [keyring-api](https://github.com/MetaMask/keyring-api/releases). I think we will need [a minimum of 6.1.1](https://github.com/MetaMask/core/pull/4262/files#diff-11700fa807e8d925b649057d40f964047063e26926ce62fc5f09210de947dbffR47). This update introduces an extra field on the InternalAccount type of importTime. This will allow us to 100% migrate off of preferences controller.

app/util/test/accountsControllerTestUtils.ts
tommasini: Reverted <3

package.json
```diff
@@ -149,7 +149,7 @@
     "@metamask/etherscan-link": "^2.0.0",
     "@metamask/gas-fee-controller": "^13.0.0",
     "@metamask/key-tree": "^9.0.0",
-    "@metamask/keyring-api": "^4.0.0",
+    "@metamask/keyring-api": "^6.0.0",
```
owencraston, line 152: I believe version 14 of the accounts controller is using `^6.1.1`

tommasini: Tried to bump because it was not a major version change, but it seems it is ^6.0.0: https://github.com/MetaMask/core/blob/3227bd51bc4eca0531a53f4be6468404e6a5f032/packages/accounts-controller/package.json#L47

app/store/migrations/036.ts
```diff
@@ -115,6 +115,7 @@ async function createInternalAccountsForAccountsController(
       options: {},
       metadata: {
         name: identity.name,
+        importTime: 123,
```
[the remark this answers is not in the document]
tommasini: Addressed! You are super right! Thanks for the catch!

owencraston (dismissed):

montelaidev (changes requested):
We will need to create a new migration to add `importTime` to existing users that have already migrated past `036`. 

An alternative would be to call `updateAccounts` on the accounts controller and it will populate the value.

yarn.lock
```diff
@@ -3633,6 +3633,23 @@
     immer "^9.0.6"
     uuid "^8.3.2"
 
+"@metamask/accounts-controller@^14.0.0":
```
legobeat, line 3636: v11 is still used by `@metamask/assets-controllers`.

I guess this also needs to be coupled with an update of `@metamask/assets-controllers` to either v29 or v30?

```
$ nvs -p @metamask/accounts-controller @metamask/assets-controllers | tail
26.0.0  2024-03-04T18:44:07.058Z        >=16.0.0        ^11.0.0
27.0.0  2024-03-12T19:47:40.443Z        >=16.0.0        ^12.0.0
27.0.1  2024-03-13T11:26:43.777Z        >=16.0.0        ^12.0.0
27.1.0  2024-03-19T08:49:30.180Z        >=16.0.0        ^12.0.0
27.2.0  2024-03-19T18:15:55.457Z        >=16.0.0        ^12.0.0
28.0.0  2024-04-17T15:00:35.910Z        >=16.0.0        ^13.0.0
29.0.0  2024-05-01T13:38:25.311Z        >=16.0.0        ^14.0.0
30.0.0  2024-05-29T20:08:33.688Z        >=16.0.0        ^14.0.0
31.0.0  2024-05-31T18:08:03.582Z        >=16.0.0        ^15.0.0
32.0.0  2024-05-31T19:50:52.965Z        ^18.18 || >=20  ^16.0.0
```

legobeat (dismissed):

montelaidev:
> We will need to create a new migration to add `importTime` to existing users that have already migrated past `036`.
> 
> An alternative would be to call `updateAccounts` on the accounts controller and it will populate the value.

Also, i think the importTime from the identity in the preferences controller should be used if its available

yarn.lock
tommasini: But since it's a dependency of assets controller and not a peer dependency, is this a blocker?

Assets controllers will still be using the accounts controller v11, do you think it should be added as a peer dependency instead?

tommasini:
> Also, i think the importTime from the identity in the preferences controller should be used if its available

Ah read that right now, will update the migration with it

yarn.lock
legobeat: The dependency is only referenced in `TokenDetectionController` Looking here:
https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/TokenDetectionController.ts

The actual code of the dependency is not actually used - it's just pulled in for types and interacting with an externally provided `AccountsController`.

So unless we have them matched here, we will have an `assets-controller` implementet towards and expecting an `accounts-controller@11` but actually getting `accounts-controller@14`.

If it has been verified that there are no compatibility issues with mixing v11/v14.

(

So yes, I do think it should be in `peerDependencies`. At which point the `dependencies` entry can be removed, as the expected types will be provided through the `peerDependencies`.

tommasini: Thanks for this answer and thanks for validating that there are no compatibility issues!
With this being said I'm WIP on the [assets controllers v30 update ](https://github.com/MetaMask/metamask-mobile/pull/9832)as we speak and that [version](https://github.com/MetaMask/core/blob/6f563ffb714db410666ec518e76e1806581367c8/packages/assets-controllers/package.json#L91) already have accounts-controller as a peer dependency

app/store/migrations/043.test.ts
```diff
@@ -0,0 +1,199 @@
...
+        .accounts,
+    ).forEach((accountId) => {
+      if (!newState.engine) {
+        expect(true).toBeFalsy();
+        return;
+      }
+      expect(
+        newState.engine.backgroundState.AccountsController.internalAccounts
+          .accounts[accountId].metadata.importTime,
+      ).toBeDefined();
```
montelaidev, line 175: We can use `expect.any(Number)` here

app/store/migrations/043.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  it('should add importTime from identities or default to the current date', () => {
+    // Mocked Date.now since jest is not aware of date.
+    jest.spyOn(Date, 'now').mockReturnValue(new Date('2023-01-01').getTime());
+    const newState: Partial<RootState> = migration(
+      oldState,
+    ) as Partial<RootState>;
+    if (!newState.engine) {
+      expect(true).toBeFalsy();
```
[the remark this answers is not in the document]
tommasini: Was a workaround that I've found to the test fail if the engine is not defined and to deal with the lint and typescript issues to check if the engine object was defined inside of newState, since it can be undefined. It will never happen!

app/store/migrations/043.ts
```diff
@@ -0,0 +1,110 @@
...
+    return state;
+  }
+
+  Object.keys(accountsControllerState.internalAccounts.accounts).forEach(
+    (accountId) => {
+      if (
+        !accountsControllerState.internalAccounts.accounts[accountId].metadata
+          .importTime
+      ) {
+        Object.keys(preferencesControllerState.identities).forEach(
```
tommasini, line 89: I'm wondering if there is any reason for the identities to be empty objects and if we should take this into account and write a fallback if it isn't, it seems safer but at the same time, it seems dead code if it's not possible such scenario

tommasini: This outdated because the 043 file represents other migration but this is still present on 044.ts file

app/store/migrations/044.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    // Mocked Date.now since jest is not aware of date.
+    jest.spyOn(Date, 'now').mockReturnValue(new Date('2023-01-01').getTime());
+    const newState: Partial<RootState> = migration(
+      oldState,
+    ) as Partial<RootState>;
+    if (!newState.engine) {
+      expect(true).toBeFalsy();
+      return;
+    }
+    Object.keys(
```
montelaidev, line 164:
```suggestion
    const newState: Pick<RootState, 'engine'> = migration(oldState) as Pick<
      RootState,
      'engine'
    >;
```

I think we can just use pick instead of partial and then we don't need to this ```
if (!newState.engine) {
      expect(true).toBeFalsy();
      return;
    }
```

File: app/store/migrations/043.ts (hunk 7 above)
Comment (montelaidev) [reply app/store/migrations/043.ts]: Yes we should still check because its possible for the user to reset their vault via forget password after running migration 036. In this case the identities should be an empty object.

File: app/store/migrations/044.test.ts (hunk 8 above)
Comment (tommasini) [reply app/store/migrations/044.test.ts:164]: You are right!  Thanks for the knowledge!

Comment (tommasini) [reply app/store/migrations/044.test.ts]: addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9779/commits/febbb833db08dd5cdbe2906df9919742657d5e0e)

File: app/store/migrations/043.ts (hunk 7 above)
Comment (tommasini) [reply app/store/migrations/043.ts]: addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9779/commits/febbb833db08dd5cdbe2906df9919742657d5e0e)

**Review by montelaidev — DISMISSED** [review]

**Review by owencraston — DISMISSED** [review]

**Review by Cal-L — DISMISSED** [review]
LGTM

File: yarn.lock [hunk 9]
```diff
@@ -3573,6 +3573,23 @@
...
```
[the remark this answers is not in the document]
tommasini: I can merge this PR into the assets controllers one: https://github.com/MetaMask/metamask-mobile/pull/9832
Or I can merge both on the same day, resolving by changing on the assets controllers one the migration for number 48*
I would like to keep those separate for now to make life easier for the reviewers on the assets controllers one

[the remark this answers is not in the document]
tommasini: Pointed https://github.com/MetaMask/metamask-mobile/pull/9832 to this branch so we can proceed with easier reviews and at the same time keep this ones that we already have! Thanks for the suggestions Lego!

wachunei (approved):

infiniteflower (approved):

jpuri (approved):

sleepytanya (approved):

sethkfman (approved):
LGTM

PR #4281, opened 2022-05-11
Feature/4086 ps link component

Cal-L:
**Description**

#4086
<div style='display:inline-block'>
image
image
</div>

**Checklist**

* [ ] There is a related GitHub issue
* [ ] Tests are included if applicable
* [ ] Any added code is fully documented

**Issue**

Progresses #???

PR #9857, opened 2024-06-05
fix: create nft auto detection modal and remove nft polling logic

sahar-fehri:
## **Description**
This PR removes any code that triggers polling on NftDetectionController.

Calling `detectNfts()` function is now only triggered when the user clicks on the NFT tab.

It also Enables NFTDetection by default for new users.

>[!NOTE]
> this PR has an asset-cpntroller patch from this [core PR #4281](https://github.com/MetaMask/core/pull/4281)
> This patch keeps NFTDetection controller extending the polling controller while the PR makes it extend BaseController but it makes sure the polling is not triggered anymore in the client code.
> It will be fully updated in newer versions of assets-controllers
> I removed what triggers the `start()` of the polling, and now instead, mobile calls `detectNfts()` directly.
> Then i made updates on the fct `getOwnerNfts` so that it only fetches information for a specific cursor instead of looping through all user nfts.
> The loop is now being done inside the `detectNfts()` fct, and we basically fetch the first page of NFTs and save it to state so its available to clients to view.

**New functionalities highlight:**

_1- Enable NFT autodetection by default to new users.
2- Old users who have the NFT autodetection off will see the new NFT detection modal to encourage them to enable the modal.
3- When users click on the banner notice, they are no longer directed to settings and instead we enable the NFT detection after they click on “enable nft autodetection” and we show the Toast. We wanted to reduce friction and we removed that redirection to settings.
4- The code should not do any of the 3mins polling in the background anymore._

## **Related issues**

Related: https://github.com/MetaMask/core/pull/4281

## **Manual testing steps**

1. Import MM application
2. Go to settings => Security and privacy and you should see that the NFT detection is enabled.
3. Go back to home page and click on NFT tab; you should be able to see your NFTs being added.
4. Switch to another account that has NFTs
5. Click on NFTs tab and you should also see your NFTs being added to sate.
6. Go to settings => Security and privacy and turn OFF the NFT detection toggle.
7. You should see a new modal. Clicking on "allow" button should re-enable the NFT detection toggle.

### **Before**

https://github.com/MetaMask/metamask-mobile/assets/10994169/83e86852-3455-4b7c-bfc6-a658dc20c1b0

### **After**

Testing a fresh wallet import:

https://github.com/MetaMask/metamask-mobile/assets/10994169/331668b1-3a27-493f-a648-3568e4ba67c2

Testing an upgrade scenario:

First build on main, where you should see the NFT auto detection disabled by default if the user did not enable it

Then build on this PR:
You should see the new NFT detection modal and clicking on `allow` button should enable the NFT detection in the settings.

https://github.com/MetaMask/metamask-mobile/assets/10994169/6bcb77f5-13ef-429e-a1f5-4d907145ee30

NFT detection banner with new Toast:

https://github.com/MetaMask/metamask-mobile/assets/10994169/2f7465c0-4c42-409b-bfe4-ed70d9e1ca2c


e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
cortisiko, line 1: I would also add another test where NFT auto-detection is off by default, and in the not detection modal, I would tap Allow. Then I would verify that the NFts are appearing correctly. The wallet that we uses the default fixture has 2 NFTs (721 and 1155) on mainnet

e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  beforeAll(async () => {
+    jest.setTimeout(170000);
+    await TestHelpers.reverseServerPort();
+  });
+
+  it('should show nft detection when nftDetection is off', async () => {
+    await withFixtures(
+      {
+        fixture: new FixtureBuilder()
+          .withGanacheNetwork()
```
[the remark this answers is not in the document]
cortisiko: Furthermore, perhaps this is a bug or not but with token detection off, when you start on mainnet, the NFT detection bottom sheet does not appear but if you switch between a custom network back to mainnet the detection sheet appears. If this is not a bug then you would need to adjust the test steps: start on ganache, switch to mainnet to trigger the bottom sheet

cortisiko (changes requested):
Alright, @sahar-fehri, I left some feedback. Furthermore here are the bugs that I  noticed: 

- With the orangefox the below 1155 contracts are not auto imported
https://opensea.io/assets/ethereum/0xd07dc4262BCDbf85190C01c996b4C06a461d2430/77675
 https://opensea.io/assets/ethereum/0x3778541C3ABb6668eCB3eCC407388837a4457d5D/1 

  in addition to this ENS NFT: https://opensea.io/assets/ethereum/0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85/52013207599245702848910397253902912447697817810483152369346578845112926213225 


- There is a bug when you try to manually import an NFT. 
 For example, you can try using the orangefox test account, go to polygon and  try to import this NFT: contract address: 0x79161C741379c1bCC815FD73D37B91cD8E159B6c and TokenID: 72

sahar-fehri:
> Alright, @sahar-fehri, I left some feedback. Furthermore here are the bugs that I noticed:
> 
> * With the orangefox the below 1155 contracts are not auto imported
>   https://opensea.io/assets/ethereum/0xd07dc4262BCDbf85190C01c996b4C06a461d2430/77675
>   https://opensea.io/assets/ethereum/0x3778541C3ABb6668eCB3eCC407388837a4457d5D/1
>   in addition to this ENS NFT: https://opensea.io/assets/ethereum/0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85/52013207599245702848910397253902912447697817810483152369346578845112926213225
> * There is a bug when you try to manually import an NFT.
>   For example, you can try using the orangefox test account, go to polygon and  try to import this NFT: contract address: 0x79161C741379c1bCC815FD73D37B91cD8E159B6c and TokenID: 72

Hey @cortisiko ! Thanks for the feedback! 
Regarding this NFT that is not showing up https://opensea.io/assets/ethereum/0xd07dc4262BCDbf85190C01c996b4C06a461d2430/77675
I'm checking with the team on this, it does not show because its marked as spam, im checking why it's the case.

For this one https://opensea.io/assets/ethereum/0x3778541C3ABb6668eCB3eCC407388837a4457d5D/1
Im actually able to see it being detected and added to state (screenshot)
Screenshot 2024-06-18 at 15 00 43

For the ENS NFT: https://opensea.io/assets/ethereum/0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85/52013207599245702848910397253902912447697817810483152369346578845112926213225
That NFT does not look like it is owned by orangeFox account. When i call [ownerOf](https://etherscan.io/address/0x57f1887a8bf19b14fc0df6fd9b2acc9af147ea85#readContract) function on the contract for this tokenId `52013207599245702848910397253902912447697817810483152369346578845112926213225` i get this address [0x378f19ddbF56ecd512C364FbA1F271A3F571C12f](https://etherscan.io/address/0x378f19ddbF56ecd512C364FbA1F271A3F571C12f) as being the current owner of that NFT.

Please lemme know if i missed something! 🙏 


For the bug on polygon mainnet after adding contract address: 0x79161C741379c1bCC815FD73D37B91cD8E159B6c and TokenID: 72; What was the behavior you experienced?

I was able to add the NFT when i import manually (screenshot) 
Screenshot 2024-06-18 at 15 06 23

brianacnguyen (dismissed):

e2e/specs/assets/nft-detection-modal.spec.js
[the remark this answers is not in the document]
sahar-fehri: I have updated the test to start on ganache and the switch to mainnet 🙏 
The new modal should only appear on mainnet because as you mentioned NFT detection today is only on mainnet and it should appear when the NFT detection toggle is off.

I was not able to reproduce the behavior you described: 
`when you start on mainnet, the NFT detection bottom sheet does not appear but if you switch between a custom network back to mainnet the detection sheet appears.`

If you have a video to provide that could help me reproduce, but if you start on mainnet and the NFT detection toggle is OFF, then you should see the modal. If it is ON, then it is expected to not see the modal unless you turn the toggle off.

sahar-fehri:
> ## [https://bitrise.io/](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/69eb2a84-9704-44eb-bce4-7090135bfa8d) **Bitrise**
> ❌❌❌ `pr_smoke_e2e_pipeline` failed on Bitrise! ❌❌❌
> 
> Commit hash: [98d1813](https://github.com/MetaMask/metamask-mobile/commit/98d18138884ee0a0e83d2267f83050c9c323c6f8) Build link: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/69eb2a84-9704-44eb-bce4-7090135bfa8d
> 
> Note
> 
> * You can kick off another `pr_smoke_e2e_pipeline` on Bitrise by removing and re-applying the `Run Smoke E2E` label on the pull request

e2e failing because of the new privacy policy toast, This [PR](https://github.com/MetaMask/metamask-mobile/pull/10021) addresses that

e2e/specs/assets/nft-detection-modal.spec.js
sahar-fehri: I added a test for that. 🙏

cortisiko:
hey @sahar-fehri thanks for clarifying my questions. To follow up here:

> For the bug on polygon mainnet after adding contract address: 0x79161C741379c1bCC815FD73D37B91cD8E159B6c and TokenID: 72; What was the behavior you experienced?
> 
> I was able to add the NFT when i import manually (screenshot)


I am no longer experiencing the import bug. However, when I kill and relaunch the app after importing the NFT, that same NFT disappears from my wallet.


https://github.com/MetaMask/metamask-mobile/assets/12821081/2eff1c85-57f9-4620-88d9-2dd7994d5565


> Hey @cortisiko ! Thanks for the feedback!
> Regarding this NFT that is not showing up https://opensea.io/assets/ethereum/0xd07dc4262BCDbf85190C01c996b4C06a461d2430/77675
> I'm checking with the team on this, it does not show because its marked as spam, im checking why it's the case.


I came across the same [issue](https://app.screencast.com/CZXKcrPETFmdp) using a different account.

e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        await NetworkEducationModal.tapGotItButton();
+
+        await Assertions.checkIfVisible(NftDetectionModal.container);
+        await NftDetectionModal.tapCancelButton();
+        // Check that we are on the wallet screen
+        await WalletView.isVisible();
+
+        // Go to NFTs tab and check that the banner is visible
+        await WalletView.tapNftTab();
+        await TestHelpers.checkIfElementByTextIsVisible(
```
cortisiko, line 51: Use the assertions class for this. You can do `Assertions.checkIfTextIsDisplayed('string')`

e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,94 @@
...
+          enContent.wallet.nfts_autodetection_desc,
```
cortisiko, line 52: ahh, so similar to how you would create a testID selector object, you would do the same for String selectors. See here for example: https://github.com/MetaMask/metamask-mobile/blob/feat/call-nft-detection-on-components/e2e/selectors/Common.selectors.js#L23

e2e/specs/assets/nft-detection-modal.spec.js
cortisiko: https://github.com/MetaMask/metamask-mobile/blob/feat/call-nft-detection-on-components/e2e/specs/confirmations/send-erc721.spec.js#L61

e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        // Check that we are on the wallet screen
+        await WalletView.isVisible();
+
+        // Go to NFTs tab and check that the banner is NOT visible
+        await WalletView.tapNftTab();
+        await TestHelpers.checkIfNotVisible(
+          enContent.wallet.nfts_autodetection_desc,
+        );
+
+        const tesNFTOnMainnet = "collectible-Life's A Joke #2875-2875";
```
cortisiko, line 89: Consider declaring this variable at the beginning of the test.

e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        await TestHelpers.checkIfVisible(tesNFTOnMainnet);
```
cortisiko, line 90: Use the assertions class for this. You can do `Assertions.checkIfTextIsDisplayed(tesNFTOnMainnet)`

e2e/specs/assets/nft-detection-modal.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        await NetworkEducationModal.tapGotItButton();
+
+        await Assertions.checkIfVisible(NftDetectionModal.container);
+        await NftDetectionModal.tapAllowButton();
+        // Check that we are on the wallet screen
+        await WalletView.isVisible();
+
+        // Go to NFTs tab and check that the banner is NOT visible
+        await WalletView.tapNftTab();
+        await TestHelpers.checkIfNotVisible(
```
cortisiko, line 85: Use the assertions class for this. You can do `Assertions.checkIfTextIsNotDisplayed('string')`

e2e/specs/assets/nft-detection-modal.spec.js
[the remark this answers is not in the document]
cortisiko: Okay great, thanks for updating! Regarding the issue I described, seems more like an edge case. I can no longer reproduce the issue.

cortisiko (changes requested):

e2e/specs/assets/nft-detection-modal.spec.js
sahar-fehri: Fixed 🙏

sahar-fehri:
> e across the same [issue](https://app.screencast.com/CZXKcrPETFmdp) using a different account.

Hey @cortisiko ! I was not able to reproduce the behavior in the video. The NFT is displayed correctly after relaunching app.
Also i checked the [issue](https://app.screencast.com/CZXKcrPETFmdp) you mentioned, i see that our API is only returning one NFT for that account ( the LifesAJokeNFT). [Opensea](https://opensea.io/0x76cf1CdD1fcC252442b50D6e97207228aA4aefC3) also is only returning one NFT for that account.

I also checked the ownership on the contract:
contract: 0xB66a603f4cFe17e3D27B87a8BfCaD319856518B8
tokenId: 100059584189867566852894599154934721552567499366058460268305346163854916190212

The account 0x76cf1CdD1fcC252442b50D6e97207228aA4aefC3 does not hold that NFT

Screenshot 2024-06-19 at 09 46 46

app/components/UI/CollectibleDetectionModal/index.tsx
```diff
@@ -13,21 +27,29 @@ const styles = StyleSheet.create({
...
-      screen: 'SecuritySettings',
-      params: {
-        scrollToDetectNFTs: true,
-      },
-    });
+const CollectibleDetectionModal = () => {
+  const { colors } = useTheme();
+  const { toastRef } = useContext(ToastContext);
+  const showToastAndEnableNFtDetection = () => {
+    void (async () => {
```
tommasini, line 34: Please solve lint issue

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { useMetrics } from '../../../components/hooks/useMetrics';
+import { MetaMetricsEvents } from '../../../core/Analytics';
+import { selectChainId } from '../../../selectors/networkController';
+import { useSelector } from 'react-redux';
+import { selectDisplayNftMedia } from '../../../selectors/preferencesController';
+
+const NFTAutoDetectionModal = () => {
+  const { styles } = useStyles(styleSheet, {});
+  const sheetRef = useRef<BottomSheetRef>(null);
+  const walletImage = require('../../../images/wallet-alpha.png'); // eslint-disable-line
```
tommasini, line 31: Should we do the require outside or directly on the prop of the image instead? it seems out of context do a require here but probably it's personal, feel free to change it or not!

app/components/Views/Wallet/index.tsx
```diff
@@ -328,13 +364,9 @@ const Wallet = ({
   useEffect(
     () => {
       requestAnimationFrame(async () => {
-        const {
-          TokenDetectionController,
-          NftDetectionController,
-          AccountTrackerController,
-        } = Engine.context as any;
+        const { TokenDetectionController, AccountTrackerController } =
+          Engine.context as any;
```
tommasini, line 368: Do we still need this `as any`? If it's not too much outside of the scope it would be cool to remove it

app/components/Views/Wallet/index.tsx
```diff
@@ -328,13 +364,9 @@ const Wallet = ({
...
         TokenDetectionController.detectTokens();
-        NftDetectionController.detectNfts();
```
tommasini, line 337: If we do not detect the NFTs it will not automatically add if the user adds an nft? Thinking out loud, not sure now how the detect works on the controller level

app/components/Views/Wallet/index.tsx
```diff
@@ -383,14 +415,23 @@ const Wallet = ({
...
 
   const onChangeTab = useCallback(
-    (obj) => {
+    async (obj) => {
       if (obj.ref.props.tabLabel === strings('wallet.tokens')) {
         trackEvent(MetaMetricsEvents.WALLET_TOKENS);
       } else {
         trackEvent(MetaMetricsEvents.WALLET_COLLECTIBLES);
+        // Call detect nfts
+        const { NftDetectionController } = Engine.context as any;
```
tommasini, line 424: Can we try not use any?

app/core/Engine.ts
```diff
@@ -448,7 +448,7 @@ class Engine {
         ipfsGateway: AppConstants.IPFS_DEFAULT_GATEWAY_URL,
         useTokenDetection:
           initialState?.PreferencesController?.useTokenDetection ?? true,
-        useNftDetection: false,
+        useNftDetection: true, // set this to true to enable nft detection by default to new users
```
tommasini, line 451: I imagine this was a product decision, could you share the thread via DM where this was discussed?

tommasini (changes requested):
Heyy, awesome work! Left some comments!

app/component-library/components/Toast/Toast.tsx
```diff
@@ -187,6 +187,19 @@ const Toast = forwardRef((_, ref: React.ForwardedRef<ToastRef>) => {
...
       }
+      case ToastVariants.Icon: {
+        const { iconName, iconColor, backgroundColor } = toastOptions;
+        return (
+          <Avatar
+            variant={AvatarVariant.Icon}
+            name={iconName}
+            iconColor={iconColor}
+            backgroundColor={backgroundColor}
+            //size={AvatarSize.Md}
```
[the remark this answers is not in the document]
sahar-fehri: Removed!

app/components/Views/Wallet/index.tsx
sahar-fehri: This `useEffect` will be triggered when the user is on the wallet component. I have removed calling `detectNft()` here because instead i want to only call `detectNfts()` when the user clicks on the NFT tab. (now triggered inside `onChangeTab`)

app/core/Engine.ts
sahar-fehri: Yes it was. I cannot find a thread regarding this but it is something that we have discussed internally to be done as part of the removal of the 3 mins polling logic;
cc @alfeng6 🙏

app/components/UI/CollectibleDetectionModal/index.tsx
sahar-fehri: Sorry i did notice that the pipeline is failing with a lint issue here, did i miss something?

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.tsx
sahar-fehri: I have pushed that require up before the function declaration!

app/components/UI/CollectibleDetectionModal/index.tsx
tommasini: It was flagged hmmm will resolve this

app/components/Views/Wallet/index.tsx
tommasini: Awesome thanks!

patches/@metamask+assets-controllers+26.0.0.patch
```diff
@@ -897,10 +897,10 @@ index d5c60db..adef299 100644
   * Controller that passively polls on a set interval for NFT auto detection
```
tommasini, line 897: I'm working on the update of Assets controllers to v30 could you update this branch on core with this change? 
Branch: `patch/mobile-assets-controllers-v30'`

brianacnguyen (dismissed):

cortisiko:
After discussing with Sahar, we notice the bug I reported [here](https://github.com/MetaMask/metamask-mobile/pull/9857#issuecomment-2177291176) exists on main. A [ticket](https://github.com/MetaMask/metamask-mobile/issues/10057) was created.

cortisiko (dismissed):
this is ✅ from an e2e perspective.

app/components/UI/CollectibleContracts/index.test.tsx
```diff
@@ -495,4 +495,124 @@ describe('CollectibleContracts', () => {
     expect(spyOnUpdateNftMetadata).toHaveBeenCalledTimes(0);
     expect(spyOnDetectNfts).toHaveBeenCalledTimes(1);
   });
+
+  it('Should show spinner if nfts are still being fetched', async () => {
```
NicolasMassart, line 499: Follow contribution guidelines for test naming please.
```suggestion
  it('Should show spinner if nfts are still being fetched', async () => {
```

app/components/UI/CollectibleContracts/index.test.tsx
```diff
@@ -495,4 +495,124 @@ describe('CollectibleContracts', () => {
...
+    };
+    const { queryByTestId } = renderWithProvider(<CollectibleContracts />, {
+      state: mockState,
+    });
+
+    const spinner = queryByTestId('spinner');
+    expect(spinner).not.toBeNull();
+  });
+
+  it('Should not show spinner if nfts are not still being fetched', async () => {
```
NicolasMassart, line 559:
```suggestion
  it('Does not show spinner if nfts are not still being fetched', async () => {
```

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+describe('NFT Auto detection modal', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('should render correctly', () => {
+    const { toJSON } = renderComponent(initialState);
+    expect(toJSON()).toMatchSnapshot();
+  });
+
+  it('should call setUseNftDetection and setDisplayNftMedia when clicking on allow button with nftDisplayMedia initially off', () => {
```
NicolasMassart, line 61:
```suggestion
  it('calls setUseNftDetection and setDisplayNftMedia when clicking on allow button with nftDisplayMedia initially off', () => {
```

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      },
+    });
+    const allowButton = getByTestId('allow');
+
+    fireEvent.press(allowButton);
+    expect(setUseNftDetectionSpy).toHaveBeenCalled();
+    expect(setDisplayNftMediaSpy).toHaveBeenCalled();
+  });
+
+  it('should call setDisplayNftMedia when clicking on allow button if displayNftMedia if on', () => {
```
NicolasMassart, line 78:
```suggestion
  it('calls setDisplayNftMedia when clicking on allow button if displayNftMedia if on', () => {
```

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const { getByTestId } = renderComponent(initialState);
+    const allowButton = getByTestId('allow');
+
+    fireEvent.press(allowButton);
+    expect(setUseNftDetectionSpy).toHaveBeenCalled();
+    expect(setDisplayNftMediaSpy).not.toHaveBeenCalled();
+  });
+
+  it('should not call setUseNftDetection nor setDisplayNftMedia when clicking on not right now button', () => {
```
NicolasMassart, line 87:
```suggestion
  it('Does not call setUseNftDetection nor setDisplayNftMedia when clicking on not right now button', () => {
```

app/components/Views/Wallet/index.tsx
```diff
@@ -267,6 +283,31 @@ any) => {
     });
   }, [navigate, providerConfig.chainId, trackEvent]);
 
+  const checkNftAutoDetectionModal = useCallback(() => {
+    const isOnMainnet = isMainNet(providerConfig.chainId);
+    if (!useNftDetection && isOnMainnet && !isNFTAutoDetectionModalOpened) {
```
NicolasMassart, line 288: Q: i'm not sure to understand properly here: it seems to me that the modal will show every time we go on the wallet mainnet if we haven't allowed nft detection.
Is this right? If so, I would advise not to do that as it's going to be very annoying...
It may be ok in the security settings, even I fi would consider as a user that I should be able to cancel modal and never see it again. But on the wallet, it's dangerous.
that why I doubt I understood this properly and need you to confirm please.

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+        const { PreferencesController } = Engine.context;
+        if (!displayNftMedia) {
+          PreferencesController.setDisplayNftMedia(true);
+        }
+        PreferencesController.setUseNftDetection(true);
+        trackEvent(MetaMetricsEvents.NFT_AUTO_DETECTION_MODAL_ENABLE, {
+          chainId,
+        });
+      } else {
+        trackEvent(MetaMetricsEvents.NFT_AUTO_DETECTION_MODAL_DISABLE, {
```
NicolasMassart, line 50: Q: here you don't set anything for the user not to have this modal after cancel each time they land on the app?

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+        {() => <NFTAutoDetectionModal />}
+      </Stack.Screen>
+    </Stack.Navigator>,
+    { state },
+  );
+describe('NFT Auto detection modal', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('should render correctly', () => {
```
NicolasMassart, line 56:
```suggestion
  it('render matches snapshot', () => {
```

app/components/Views/Wallet/index.tsx
sahar-fehri: Hey @NicolasMassart  👋 , It won't show everytime. The modal will appear only once, and the user will not see it again.
The user can should to dismiss or allow the NFT autodetection, but regardless, the modal will appear only once.
The check for `isNFTAutoDetectionModalOpened` should prevent that

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.tsx
sahar-fehri: If i understood your question correctly, you are asking why im not setting anything when the user cancels so the modal does not appear again?
If so, The dispatch here https://github.com/MetaMask/metamask-mobile/pull/9857/files#diff-6cf3f4e9a74706082cca6f2c08153710158e6a971d42c6f7179ce4721e70bacaR292 should prevent the modal from appearing again. Regardless if the user clicks on cancel or allow

app/components/Views/Wallet/index.tsx
NicolasMassart: I think the name "isNFTAutoDetectionModalOpened" is what confused me. Something like "isNFTAutoDetectionModalViewed" would be clearer for me. "opened" it's not clear if it's just checking if it's currently already open or not or if it opened once before. I thought it was a value to make sure you don't open two at the same time.
I would advise for something that indicates the user actually already viewed it once.

patches/@metamask+assets-controllers+30.0.0.patch
NicolasMassart, line 1: I'm not sure how to review a big patch like this.
Is there a PR with your changes on the source code on the core repos I could check?
This "compiled" code is not meant to be read by humans.

sahar-fehri: Yes! this is the core PR https://github.com/MetaMask/core/pull/4281 🙏 
Note: This patch keeps NFTDetection controller extending the polling controller while the PR makes it extend BaseController but it makes sure the polling is not triggered anymore in the client code.
It will be fully updated in newer versions of assets-controllers

app/components/Views/Wallet/index.tsx
sahar-fehri: Yes, totally, apologies for the confusion will update the naming to `isNFTAutoDetectionModalViewed` 🙏

bergeron (dismissed):
tested all the scenarios successfully.  new user, existing user, the modal, sending requests when clicking the nft tab, verifying http requests are respecting cache

patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,237 @@ index 0dc70ec..461a210 100644
...
+@@ -72,11 +75,6 @@ var NftDetectionController = class extends _pollingcontroller.StaticIntervalPoll
+       const { selectedAddress: previouslySelectedAddress, disabled } = this.config;
+       if (selectedAddress !== previouslySelectedAddress || !useNftDetection !== disabled) {
+         this.configure({ selectedAddress, disabled: !useNftDetection });
+-        if (useNftDetection) {
+-          this.start();
+-        } else {
+-          this.stop();
+-        }
+       }
```
sahar-fehri, line 63: Removed this to remove triggering the start of the polling when users enable the NFT auto detection toggle

patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,237 @@ index 0dc70ec..461a210 100644
...
++        Version: '1'
+       }
+-      const newNfts = nftApiResponse.tokens.filter(
+-        (elm) => elm.token.isSpam === false && (elm.blockaidResult?.result_type ? elm.blockaidResult?.result_type === "Benign" /* Benign */ : true)
+-      );
+-      nfts = [...nfts, ...newNfts];
+-    } while (next = nftApiResponse.continuation);
+-    return nfts;
++    });
++    return nftApiResponse;
```
sahar-fehri, line 107: The function getOwnerNfts no longer loops through all user NFTs and return them, but only fetches for a specific cursor.

patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,237 @@ index 0dc70ec..461a210 100644
...
+-          image_thumbnail_url && { imageThumbnail: image_thumbnail_url },
+-          image_original_url && { imageOriginal: image_original_url },
+-          kind && { standard: kind.toUpperCase() },
+-          lastSale && { lastSale },
+-          attributes && { attributes },
+-          topBid && { topBid },
+-          rarityRank && { rarityRank },
+-          rarityScore && { rarityScore },
+-          collection && { collection }
++    const updateKey = `${chainId}:${userAddress}`;
```
sahar-fehri, line 163: Added this to avoid having two ongoing NFT fetching at the same time.

patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,237 @@ index 0dc70ec..461a210 100644
...
++        apiNfts = resultNftApi.tokens.filter((elm) =>
++            elm.token.isSpam === false &&
++            (elm.blockaidResult?.result_type? elm.blockaidResult?.result_type === BlockaidResultType.Benign: true),
+         );
+-        await this.addNft(contract, token_id, {
+-          nftMetadata,
+-          userAddress,
+-          source: "detected" /* Detected */,
+-          networkClientId
++        const addNftPromises = apiNfts.map(async (nft) => {
```
sahar-fehri, line 194: Now detectNfts function fetches the first page of user NFTs and saves it to state so it is available to display instead of waiting to fetch all user NFTs before saving to state.

NicolasMassart (dismissed):
Looks good to me.
Thanks a lot for all your effort making this review clear and solid!

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+        displayNftMedia: true,
+      },
+    },
+  },
+};
+
+const Stack = createStackNavigator();
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const renderComponent = (state: any = {}) =>
```
tommasini, line 43: Most likely this any can be typed 
We shouldn't use any anymore, it stricts all the type safety we need

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.tsx
```diff
@@ -0,0 +1,103 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
```
tommasini, line 1: If we use the require inside the image component so we still need all this disables?

app/components/Views/Wallet/index.tsx
```diff
@@ -117,9 +125,11 @@ const Wallet = ({
   navigation,
   storePrivacyPolicyShownDate,
   shouldShowNewPrivacyToast,
+  currentRouteName,
   storePrivacyPolicyClickedOrClosed,
-}: // TODO: Replace "any" with type
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+  showNftFetchingLoadingIndicator,
+  hideNftFetchingLoadingIndicator,
+}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tommasini, line 132: We shouldn't use any anymore, could you please type instead?

app/components/Views/Wallet/index.tsx
```diff
@@ -254,6 +265,11 @@ any) => {
   const networkName = useSelector(selectNetworkName);
 
   const networkImageSource = useSelector(selectNetworkImageSource);
+  const useNftDetection = useSelector(selectUseNftDetection);
+  const isNFTAutoDetectionModalViewed = useSelector(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (state: any) => state.security.isNFTAutoDetectionModalViewed,
```
tommasini, line 271: Same, please use RootState type instead of any

e2e/selectors/Modals/NftDetectionModal.selectors.js
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/prefer-default-export
```
tommasini, line 1: I think we can remove this disable

tommasini (changes requested):

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
sahar-fehri: I thought in a test file its okay to keep the any, otherwise it will complain for missing properties that may not be relevant to the test, what do you think?

tommasini: I think we can use Partial (https://www.typescriptlang.org/docs/handbook/utility-types.html)#partialtype if we just want some of the properties 🙏🏼

sahar-fehri: Updated, we might want to circle back at some point in the future to the rest of test files to cleanup the "any" 🙏

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.tsx
sahar-fehri: After discussion this we decided to keep the disable since it would be needed if we move the require inside the <Image>

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
tommasini: Definitely! We need to improve the Typescript coverage of the mobile app in general and set quality gates!
Step by step we reach there! These small efforts helps! Thank you a lot for addressing it!

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </Stack.Screen>
+    </Stack.Navigator>,
+    { state },
+  );
+describe('NFT Auto detection modal', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('render matches snapshot', () => {
+    const { toJSON } = renderComponent(initialState as Partial<RootState>);
```
tommasini, line 57: Interesting, do we need to cast this here? Just curious

tommasini (dismissed):
LGTM! Feel free to resolve the comments, or if you do any changes regarding them ping me to re approve it!
There is a lint issue on CollectibleDetectionModal

app/components/Views/NFTAutoDetectionModal/NFTAutoDetectionModal.test.tsx
sahar-fehri: I added the cast because otherwise it would say something about `missing properties  AccountTrackerController, AddressBookController, AssetsContractController,,,, and 22 more.`

tommasini: What if where you initialise the initial state you say that is `Partial<RootState>`, would still need to cast this everywhere?

```
const initialState: Partial<RootState> = {
  engine: {
    backgroundState: {
      PreferencesController: {
        displayNftMedia: true,
      },
    },
  },
};
```

sahar-fehri: Yes, `Partial<RootState>` will only make optional the first level of the type properties. Would work perfectly with simple objects but since the state is nested, it would still ask for missing state properties.

I wanted to avoid creating a new type for it, but i guess we can create a "deep partial" type, to avoid having to add any properties not necessarily relevant to the test.

tommasini: Thanks!! I was not aware it was only the first level of depth! Thanks for sharing that knowledge!

With this being said I will resolve this conversation since it's a test file, and will leave up to your decision!

tommasini (dismissed):

cortisiko (approved):
🌮 🌮 🌮 🌮 🌮 🌮

tommasini (approved):

PR #9961, opened 2024-06-12
feat: adding search to the network bottomsheet as part of the networks management UI redesign 

EtherWizard33:
## **Description**
Adding search to the network bottomsheet as part of the networks management UI redesign. Here is a 2 minute overview demo: https://www.loom.com/share/dc29313a5cd34e9893c76a0caeba0fc1

Here is a link to the related Figma that includes the adding of the search box to the bottom sheet:
https://www.figma.com/design/76R5BvAVubUhWaN2Ld7MAv/Default-Networks?node-id=1303-9439&t=yhxfHHCsG0CimP0w-0

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2495

## **Manual testing steps**

1. Add a the following env var that will be used as a feature flag to enable to the feature export MM_NETWORK_UI_REDESIGN_ENABLED="1"
2. From the header in the home screen, click the network selector
3. The network bottom sheet should come up and you should see the search box
4. Type ava in the search box and it should filter out all other networks and show only avalanche
5. For a demo please see the video in the description above


tommasini:
Probably background of the container of the button shouldn't be visible like this? It is like that on the figma file?
image

tommasini:
Can we have recordings on dark mode as well?

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -64,16 +64,20 @@ import { TESTNET_TICKER_SYMBOLS } from '@metamask/controller-utils';
...
   const [showPopularNetworkModal, setShowPopularNetworkModal] = useState(false);
   const [popularNetwork, setPopularNetwork] = useState(undefined);
   const [showWarningModal, setShowWarningModal] = useState(false);
+  const [searchString, setSearchString] = useState('');
   const { navigate } = useNavigation();
   const theme = useTheme();
   const { trackEvent } = useMetrics();
   const { colors } = theme;
-  const styles = createStyles(colors);
+  const styles = createStyles(colors || mockTheme.colors);
```
tommasini, line 80: Does this fallback for `mockTheme.colors` have any occasion where it get's triggered?

app/components/Views/NetworkSelector/NetworkSelector.styles.ts
```diff
@@ -37,6 +42,59 @@ const createStyles = (colors: any) =>
       color: colors.primary.default,
       marginTop: 1,
     },
+    networkListContainer: {
+      height: isNetworkUiRedesignEnabled ? '100%' : undefined,
```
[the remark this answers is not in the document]
EtherWizard33: 100% is just to used as fixed height, it could be 90% or 80%.
It necessary so that the size does not change while searching, when there is only 2 search results, the window changed height based on the content.

Btw, since it is under the feature flag, it does not affect the current version, only this bottom sheet and only if the feature flag is enabled.

app/components/Views/NetworkSelector/NetworkSelector.tsx
EtherWizard33: Yes you are right, I think it can be removed, it was inspired by another component who used it.
I can remove it.

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -165,8 +169,23 @@ const NetworkSelector = () => {
...
+      name.toLowerCase().includes(networkName.toLowerCase()),
+    );
+
+    return searchResult;
+  };
+
   const renderMainnet = () => {
     const { name: mainnetName, chainId } = Networks.mainnet;
+
+    if (
```
tommasini, line 183: I see this 4 conditions for render or not the networks, I'm wondering if we can create a util function to be called instead, with a name that help us identify better what's going on on that condition, let me know your thoughts

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -310,18 +358,51 @@ const NetworkSelector = () => {
...
+          renderTitle('networks.enabled_networks')}
         {renderMainnet()}
         {renderLineaMainnet()}
         {renderRpcNetworks()}
         {isNetworkUiRedesignEnabled &&
+          searchString.length === 0 &&
           renderTitle('networks.additional_networks')}
         {isNetworkUiRedesignEnabled && renderAdditonalNetworks()}
-        {renderTestNetworksSwitch()}
+        {searchString.length === 0 && renderTestNetworksSwitch()}
```
tommasini, line 405: I will assume that's the intention of the feature since the issue says:

"As the user types into the search input, the list of networks dynamically updates to display only those networks that match the search query. This applies to both existing added networks and popular networks available for addition."

Although if the user have the test networks enabled, he could be looking for a test network, food for thoughts

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -1,6 +1,6 @@
 // Third party dependencies.
```
tommasini, line 1: It would be cool if you see value in adding a test scenario for this new feature

Also the snapshot is not failing because this is using a env variable to be on and off, and on the snapshot data this is off, just questioning if that will be updated on the last issue of this redesign

I'm okay with adding now or later the unit tests for it!

EtherWizard33:
> Probably background of the container of the button shouldn't be visible like this? It is like that on the figma file? image

Hi Tomas, I had a look at it and it has always been like this, before or after this PR. The figma itself doesnt seem to show example specific to this. I had a look and it doesnt seem to be a container but the button itself, I tried a few things like overflow:'hidden' in case it could be a quick win but it didnt change it.

app/components/Views/NetworkSelector/NetworkSelector.tsx
EtherWizard33: Hi Tomas, yes I have given it some thoughts, right now test nets are shown during search if the test net search is enabled, otherwise it would be showing them when they are not enabled, assuming it's devs using test nets they most likely should know to toggle the switch.

app/components/Views/NetworkSelector/NetworkSelector.tsx
EtherWizard33: yes you are right, this is now fixed, marking as resolved

app/components/Views/NetworkSelector/NetworkSelector.tsx
EtherWizard33: Yes, this is intended, the team plans to add the tests after the features under the feature flag are complete, and before release. Glad you noticed, marking as resolved for now since it seems we are on the same page :)

app/components/Views/NetworkSelector/NetworkSelector.tsx
EtherWizard33: done, marking as resolved

app/components/Views/NetworkSelector/NetworkSelector.styles.ts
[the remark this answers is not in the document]
EtherWizard33: Given the details shared above, I'm going ahead and marking this as resolved, trusting it aligns well with your expectations. Cheers!

salimtb (approved):

EtherWizard33:
> Can we have recordings on dark mode as well?

Oi Tomas! 
Here is the darkmode:

darkmode-network-search

tommasini (approved):
Awesome! Thanks for all the answers! LGTM!

PR #9997, opened 2024-06-14
fix: synchronize accounts name

tommasini:
## **Description**

Migration to synchronise accounts name based on identities names of preferences controller

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### DIFF
--- a/app/store/migrations/044.test.ts
+++ b/app/store/migrations/044.test.ts
@@ -92,7 +92,7 @@ describe('Migration #44', () => {
         },
       }),
       errorMessage:
-        "FATAL ERROR: Migration 44: Invalid AccountsController state error: 'null'",
+        "FATAL ERROR: Migration 44: Invalid AccountsController state error: 'object'",
       scenario: 'AccountsController state is invalid',
     },
     {
@@ -104,7 +104,7 @@ describe('Migration #44', () => {
         },
       }),
       errorMessage:
-        "FATAL ERROR: Migration 44: Invalid AccountsController internalAccounts state error: 'null'",
+        "FATAL ERROR: Migration 44: Invalid AccountsController internalAccounts state error: 'object'",
       scenario: 'AccountsController internalAccounts state is invalid',
     },
     {
@@ -117,7 +117,7 @@ describe('Migration #44', () => {
         },
       }),
       errorMessage:
-        "FATAL ERROR: Migration 44: Invalid PreferencesController state error: 'null'",
+        "FATAL ERROR: Migration 44: Invalid PreferencesController state error: 'object'",
       scenario: 'PreferencesController state is invalid',
     },
     {
@@ -130,7 +130,7 @@ describe('Migration #44', () => {
         },
       }),
       errorMessage:
-        "FATAL ERROR: Migration 44: Invalid PreferencesController identities state error: 'null'",
+        "FATAL ERROR: Migration 44: Invalid PreferencesController identities state error: 'object'",
       scenario: 'PreferencesController identities state is invalid',
     },
     {
@@ -147,8 +147,58 @@ describe('Migration #44', () => {
         },
       }),
       errorMessage:
-        "FATAL ERROR: Migration 44: Invalid AccountsController entry with id: '92c0e479-6133-4a18-b1bf-fa38f654e293', type: 'object'",
+        "FATAL ERROR: Migration 44: Invalid AccountsController account entry with id: '92c0e479-6133-4a18-b1bf-fa38f654e293', type: 'object'",
       scenario: 'AccountsController accounts account state is invalid',
+    },
+    {
+      state: merge({}, initialRootState, {
+        engine: {
+          backgroundState: {
+            PreferencesController: { identities: null },
+            AccountsController: {
+              internalAccounts: {
+                accounts: {
+                  '92c0e479-6133-4a18-b1bf-fa38f654e293': { metadata: null },
+                },
+              },
+            },
+          },
+        },
+      }),
+      errorMessage:
+        "FATAL ERROR: Migration 44: Invalid AccountsController account metadata entry with id: '92c0e479-6133-4a18-b1bf-fa38f654e293', type: 'object'",
+      scenario: 'AccountsController accounts account state is invalid',
+    },
+    {
+      state: merge({}, initialRootState, {
+        engine: {
+          backgroundState: {
+            PreferencesController: { identities: null },
+            AccountsController: {
+              internalAccounts: {
+                accounts: { '92c0e479-6133-4a18-b1bf-fa38f654e293': null },
+              },
+            },
+          },
+        },
+      }),
+      errorMessage:
+        "FATAL ERROR: Migration 44: Invalid AccountsController account entry with id: '92c0e479-6133-4a18-b1bf-fa38f654e293', type: 'object'",
+      scenario: 'AccountsController accounts account state is invalid',
+    },
+    {
+      state: merge({}, initialRootState, {
+        engine: {
+          backgroundState: {
+            PreferencesController: {
+              identities: { [mockChecksummedInternalAcc1]: null },
+            },
+          },
+        },
+      }),
+      errorMessage:
+        "FATAL ERROR: Migration 44: Invalid PreferencesController identity entry with type: 'object'",
+      scenario: 'PreferencesController identities identity state is invalid',
     },
   ];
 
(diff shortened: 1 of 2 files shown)


app/store/migrations/044.ts
```diff
@@ -0,0 +1,134 @@
...
+    return state;
+  }
+
+  const accountsControllerState =
+    state.engine.backgroundState.AccountsController;
+
+  if (!isObject(accountsControllerState)) {
+    captureException(
+      new Error(
+        `FATAL ERROR: Migration 44: Invalid AccountsController state error: '${JSON.stringify(
```
Gudahtt, line 30: I know this is a common practice in our migrations, but I think directly including corrupted state contents in error messages should be discouraged for reasons of privacy and Sentry error comprehensibility. In general, including any "unknown data" in error messages should be discouraged.

In cases like this, we can provide informative data by including the type of the data instead. Whatever is enough to explain why the check failed. This is a type check, so `typeof accountsControllerState` should be enough to explain to us _how_ the state is corrupted without revealing anything else, and without introducing a new unique Sentry error per user.

app/store/migrations/044.ts
```diff
@@ -0,0 +1,134 @@
...
+  const accounts = accountsControllerState.internalAccounts.accounts;
+  const identities = preferencesControllerState.identities;
+  Object.keys(accounts).forEach((accountId) => {
+    const account = accounts[accountId];
+    if (isObject(account) && isObject(account.metadata)) {
+      if (Object.keys(identities).length) {
+        Object.keys(identities).forEach((identityAddress) => {
+          if (
+            toChecksumHexAddress(identityAddress) ===
+            toChecksumHexAddress(account.address as string)
```
Gudahtt, line 114: Hmm, we shouldn't need casts anywhere here, unless we've failed to validate some part of the state

app/store/migrations/044.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+  const accounts = accountsControllerState.internalAccounts.accounts;
+  const identities = preferencesControllerState.identities;
+  Object.keys(accounts).forEach((accountId) => {
+    const account = accounts[accountId];
+    if (isObject(account) && isObject(account.metadata)) {
+      if (Object.keys(identities).length) {
+        Object.keys(identities).forEach((identityAddress) => {
+          if (
+            toChecksumHexAddress(identityAddress) ===
```
Gudahtt, line 113: Nit: Perhaps easier to use `toLowerCase` to ensure they are comparable. It's faster than `toChecksumHexAddress`

app/store/migrations/044.ts
```diff
@@ -0,0 +1,134 @@
...
+      new Error(
+        `FATAL ERROR: Migration 44: Invalid PreferencesController identities state error: '${preferencesControllerState.identities}'`,
+      ),
+    );
+    return state;
+  }
+  const accounts = accountsControllerState.internalAccounts.accounts;
+  const identities = preferencesControllerState.identities;
+  Object.keys(accounts).forEach((accountId) => {
+    const account = accounts[accountId];
```
Gudahtt, line 108: Nit: Slight readability improvement:

```suggestion
  Object.entries(accounts).forEach(([accountId, account]) => {
```

app/store/migrations/044.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (isObject(account) && isObject(account.metadata)) {
```
Gudahtt, line 109: Nit: Better to do these validation checks separately, before we begin mutating state, so that we leave the state unmodified if it's corrupted somehow in an unexpected way. If we _partially_ migrate corrupted state, it would be much more confusing to figure out how to proceed.

Gudahtt, line 109: The `isObject(account.metadata)` check here does not appear to be covered by the validation steps. If this check is false, it means that the user's state is invalid but we would not get a Sentry error for it, and we might still proceed with mutating state despite the corruption.

Our general strategy is to detect corruption and report on it, then bail with no migrations. If we ignore corruption, we may fail to discover it until it's too late. And if we mutate despite corruption, it may be more difficult to determine how to fix the issue once we discover this.

app/store/migrations/044.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (Object.keys(identities).length) {
+        Object.keys(identities).forEach((identityAddress) => {
```
Gudahtt, line 111: Nit: Slight readability improvement, we can use `Object.entries` here as well to get the identity and the address at one time

app/store/migrations/044.ts
tommasini: It's because `toChecksumHexAddress` receives a string and we validate that account is an object but we do not validate that address is part of that object and it is a string hmmm let me test something

app/store/migrations/044.ts
tommasini: You are right! I need to change here and in other PRs that I have, thanks Mark!

app/store/migrations/044.ts
tommasini:
```
  if (
      isObject(account) &&
      isObject(account.metadata) &&
      typeof account.address === 'string'
    ) {
      if (Object.keys(identities).length) {
        Object.keys(identities).forEach((identityAddress) => {
          if (
            toChecksumHexAddress(identityAddress) ===
            toChecksumHexAddress(account.address)
          ) {
```

Tried something like this, but it seems I'm missing how to typesafe a string inside an object hmm I'm missing something

app/store/migrations/044.ts
tommasini: you are totally right, it's missing from the validation steps! I'm adding it as we speak

app/store/migrations/044.test.ts
```diff
@@ -0,0 +1,194 @@
...
+);
+
+const oldState = {
+  engine: {
+    backgroundState: {
+      PreferencesController: {
+        identities: {
+          [mockChecksummedInternalAcc1]: {
+            address: mockChecksummedInternalAcc1,
+            name: 'Account 1',
```
Gudahtt, line 29: If this state is meant to represent the affected users, probably it would be better to have non-default names here, but use default names in the AccountsController state.

e.g. here I'd expect a custom name like "First", "Second", etc.. And then for the AccountsController, it would be "Account 1", "Account 2", etc.

app/store/migrations/044.test.ts
```diff
@@ -0,0 +1,194 @@
...
+
+      expect(newState).toStrictEqual(state);
+      expect(mockedCaptureException).toHaveBeenCalledWith(expect.any(Error));
+      expect(mockedCaptureException.mock.calls[0][0].message).toBe(
+        errorMessage,
+      );
+    });
+  }
+
+  it('should set name property of accounts of accounts controller to name property of identities of Preferences Controller', () => {
```
Gudahtt, line 167: Probably we should have a test case for unaffected users (e.g. users who have synchronized names already)

Gudahtt (commented):
Great work! Functionally this looks great to me, comments are all suggested improvements to error handling, types, tests, etc.

tommasini:
Thanks for the review Mark, all comments were addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9997/commits/e6943436554bee26bacb2fd3dd23d94352b93980) except one that I will keep the conversation opened!

app/store/migrations/044.test.ts
tommasini: Added on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9997/commits/140afd4379daa9871949f827cfc9999448659cc4)

app/store/migrations/044.test.ts
tommasini: Great catch!

Added on this [commit](https://github.com/MetaMask/metamask-mobile/pull/9997/commits/140afd4379daa9871949f827cfc9999448659cc4)

app/store/migrations/044.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (
+      isObject(account) &&
+      isObject(account.metadata) &&
+      typeof account.address === 'string'
+    ) {
+      if (Object.keys(identities).length) {
+        Object.entries(identities).forEach(([identityAddress, identity]) => {
+          if (
+            identityAddress.toLowerCase() ===
+            (account.address as string).toLowerCase()
```
Gudahtt, line 147: There would be a way to get rid of these casts I think, but the main problem here seems to be that TypeScript is forgetting about the inferred type because this is in scope of a different function. I am not really sure why this is happening. Perhaps we can investigate this further later.

Gudahtt (approved):
LGTM!

app/store/migrations/044.ts
tommasini: Deal! Let's do it! We will have more migrations to give more thoughts to this

owencraston (approved):

PR #10063, opened 2024-06-21
chore: Refactor Snaps integration

FrederikBolding:
## **Description**

Refactors a few parts of the Snaps integration in mobile and fixes some problems:
- Fixes RPC methods being called by dapps - which were broken due to moving the PermissionController
- Move PermissionController middleware to execute before all other middlewares when handling requests from Snaps
- Simplify NPM implementation significantly by using a base class `BaseNpmLocation` which handles most of the complexity
- Support `snap_getFile`, which was broken due to a missing hook
- Add proper feature flags for basic functionality and allowing local Snaps


Jonathansoufer (dismissed):
LGTM!

Jonathansoufer (approved):

PR #10066, opened 2024-06-21
feat: Snaps Custom UI v0

FrederikBolding:
## **Description**
Very much WIP


app/components/Approvals/Snaps/SnapDialogApproval/SnapDialogApproval.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { styles } = useStyles(stylesheet, {});
+
+  const onConfirm = useCallback(async () => {
+    if (!approvalRequest) return;
+
+    await Engine.acceptPendingApproval(approvalRequest.id, true as any);
+  }, [approvalRequest]);
+
+  const onReject = useCallback(async () => {
+    if (!approvalRequest) return;
```
Daniel-Cross, line 31: We've been having a discussion recently on the Mobile Team regarding the usage of `useCallback` and `useMemo` as there is generally an overuse of these hooks in an attempt to make the code more efficient. However this is not really the case, and every time we use these, we're asking the device to allocate system memory to watch the dependencies in case they change. In essence, we're introducing memory leaks into our code in the name of improvements when we haven't had any performance issues around these functions to begin with. In this instance, we're asking the `useCallback` to watch a boolean. This could be re-rendered thousands of times and it wouldn't impact performance in anyway.

Thats a long winded way of basically saying, can you please remove the usage of `useCallback` on these functions as it's overkill and not needed.

I recently wrote some docs on this if you're interested in knowing more: https://www.notion.so/metamask-consensys/React-Native-Contributor-Guidelines-127f86d67d6880adb3f4e9ec784c0d36

app/components/UI/Snaps/SnapUIRenderer/SnapUIRenderer.jsx
```diff
@@ -0,0 +1,69 @@
...
+  // sections are memoized to avoid useless re-renders if one of the parents element re-renders.
+  const sections = useMemo(
+    () =>
+      content &&
+      mapToTemplate({
+        map: {},
+        element: content,
+      }),
+    [content],
+  );
```
Daniel-Cross, line 37: Is this performance intensive? Re-rendering is also garbage collecting. The longer the app is open, the more memory is allocated, introducing performance issues. A string being re-rendered multiple times is much better than allocating device memory.

app/components/UI/Snaps/SnapUIRenderer/SnapUIRenderer.jsx
```diff
@@ -0,0 +1,69 @@
...
+        </SnapInterfaceContextProvider>
+      </Box>
+  );
+};
+
+// SnapUIRenderer is memoized to avoid useless re-renders if one of the parents element re-renders.
+export const SnapUIRenderer = memo(
+  SnapUIRendererComponent,
+  (prevProps, nextProps) => isEqual(prevProps, nextProps),
+);
```
Daniel-Cross, line 63: Same here. Is this really necessary or are we trying to optimise before we've had performance issues?

Daniel-Cross (changes requested):
Looking good, nice and clean. Just a couple of changes needed around the usage of `useCallback` and `useMemo` hooks.

app/components/Approvals/Snaps/SnapDialogApproval/SnapDialogApproval.tsx
FrederikBolding: No strong preference on the hook use in this component, feel free to make any changes you see fit!

app/components/UI/Snaps/SnapUIRenderer/SnapUIRenderer.jsx
FrederikBolding: This is performance intensive, at least enough that we have decided to use memoization. This function maps a JSX blob  to our templating format, including some hashing to derive deterministic React keys. All of which is required to render the custom UI deterministically.

app/components/UI/Snaps/SnapUIRenderer/SnapUIRenderer.jsx
```diff
@@ -0,0 +1,69 @@
...
+const SnapUIRendererComponent = ({
+  snapId,
+  isLoading = false,
+  interfaceId,
+}) => {
+  const interfaceState = useSelector(
+    (state) => getMemoizedInterface(state, interfaceId),
+    // We only want to update the state if the content has changed.
+    // We do this to avoid useless re-renders.
+    (oldState, newState) => isEqual(oldState.content, newState.content),
```
FrederikBolding, line 23: This is also somewhat unusual, but required due to the way state management is handled in custom UI.

We want to avoid the renderer re-rendering as much as possible and instead let the components themselves re-render where possible.

The context is initialized with the state on the first render and from then onwards the context manages the state for each component. The state values are synced to the controller as well, but it has proven faster to keep a "cache" locally in the form of `SnapInterfaceContext`.

PR #10110, opened 2024-06-25
chore: disable protect modal in development

dbrans:
## **Description**
This PR disables the "Protect your wallet" blocking modal when NODE_ENV is 'development'
image

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### DIFF
--- a/app/components/UI/DrawerView/index.js
+++ b/app/components/UI/DrawerView/index.js
@@ -87,7 +87,7 @@ import { selectSelectedInternalAccountChecksummedAddress } from '../../../select
 import { createAccountSelectorNavDetails } from '../../Views/AccountSelector';
 import NetworkInfo from '../NetworkInfo';
 import { withMetricsAwareness } from '../../../components/hooks/useMetrics';
-import { isDevelopment } from 'app/util/env';
+import { isDevelopment } from '../../../util/env';
 
 const createStyles = (colors) =>
   StyleSheet.create({
--- a/app/components/UI/DrawerView/index.js
+++ b/app/components/UI/DrawerView/index.js
@@ -87,6 +87,7 @@ import { selectSelectedInternalAccountChecksummedAddress } from '../../../select
 import { createAccountSelectorNavDetails } from '../../Views/AccountSelector';
 import NetworkInfo from '../NetworkInfo';
 import { withMetricsAwareness } from '../../../components/hooks/useMetrics';
+import { isDevelopment } from 'app/util/env';
 
 const createStyles = (colors) =>
   StyleSheet.create({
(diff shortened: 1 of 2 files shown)


app/components/UI/DrawerView/index.js
```diff
@@ -949,6 +949,11 @@ class DrawerView extends PureComponent {
   };
 
   renderProtectModal = () => {
+    const NODE_ENV = 'NODE_ENV';
```
matthewwalsh0, line 952: Can all this be encapsulated in a util such as `isDevBuild` assuming one doesn't already exist?

dbrans: Great idea. Done.

tommasini:
I don't know if it's a good idea to hide this always when it's development, we identified recently bug on that modal, if we apply this we will always miss if any bug is present during development 
I think we could create a variable for this but one specific 
Open to more thoughts

dbrans:
@tommasini I think this could be a good improvement: default the modal to off during development. Developers who need to debug it can just comment out a line as they do now. What do you think?

tommasini:
> I think this could be a good improvement: default the modal to off during development. Developers who need to debug it can just comment out a line as they do now. What do you think?
If I understood correctly you are suggesting that the default behaviour is to hide the protect modal to funded not backed up accounts.
I am not sure just because we don't have anything educating the engineers that this was a feature, some of the new engineers probably don't know it, and if we turn this on by default for not showing the protect modal, I'm afraid they can only discover when they create a bug on it! Sorry, I am just trying to think how this could go wrong, If more team members (QA team included) are okay with this, I'm OK with moving this further!

app/util/env.ts
```diff
@@ -0,0 +1,4 @@
+const NODE_ENV = 'NODE_ENV';
+
+// eslint-disable-next-line import/prefer-default-export
+export const isDevelopment = () => process.env[NODE_ENV] === 'development';
```
NicolasMassart, line 4: S: we may not realy need a const on this as it's only used once and in s very specific scope (envs).
I suggest the following:
```suggestion
export const isDevelopment = () => process.env['NODE_ENV'] === 'development';
```
Also prefer default export is not a rule anymore.

legobeat:
If it's too onerous for devs to force this behavior, perhaps that's an indication it is also too arduous for many users?

Could it perhaps be toggled via "Settings" as a runtime setting, rather than as build-time configuration?

NicolasMassart:
Just going back on this, I have an idea: the issue is currently that there's no way to escape this modal an once you have it you don't have other choice than saving SRP.
What about a button to ignore for some time, like os upgrade, but have this forced after say 10 ignores. That would allow devs enough time to test things.
The modal would show each time we open the foreground the app on the wallet view.
What do you think?

NicolasMassart:
Do we close this one @gauthierpetetin ?

gauthierpetetin:
@NicolasMassart can you please post a message in #metamask-mobile-dev channel to inform other engineers you'll close this in 24 hours if no one disagrees?

PR #10133, opened 2024-06-27
fix: Fix BaseControllerV1 state rehydration

### LINKED ISSUE #10057: [Bug]: NFT disappears after killing and relaunching the app
### Describe the bug

There is currently a bug on main and the release relating to NFTs disappearing. The bug happens when you:

manually import an NFT
Kill and relaunch the app
Upon landing on the wallet, the NFT disappears. 

### Expected behavior

NFTs should not disappear from your account when you kill and relaunch the app 

### Screenshots/Recordings

https://github.com/MetaMask/metamask-mobile/assets/12821081/657081df-08a2-414e-a4fd-e558ef56f86c

### Steps to reproduce

Import wallet that has NFTs
Manually import the NFT do not use NFT auto detection
Kill and relaunch the app
Notice the NFT is no longer in your wallet

_No response_

### Version

7.26 RC

### Build type

None

### Device

iOS and Android

### Operating system

Android

### Additional context

Credits to Sahara for going back in the commit history to reproduce:

On commit: 2384af6f6 => I can reproduce this bug => PR: https://github.com/MetaMask/metamask-mobile/pull/9570/files
On commit 5db49f187 => I cannot reproduce this bug

### Severity

_No response_

Gudahtt:
## **Description**

We were failing to properly restore persisted state for some BaseControllerV1 controllers. This resulted in state being lost when the application was restarted. The only confirmed affected controller so far is the `NftController`.

The context here is that for some older controllers, rather than passing the initial state into the controller directly when it is constructed, we call the `update` function directly. This has always been a bad practice (controllers should update their own state), and we're forced to fix this for controllers that we update to `BaseControllerV2` (because directly updating state like this is no longer possible. But this method of rehydrating controller state was still relied upon in some cases.

This was broken recently in #9570 when a condition was added to fix a type error. The condition was meant to check that the controller had a `subscribe` function. Unfortunately this `hasProperty` check only looks at owned properties, not inherited properties, and the `subscribe` function was inherited from the base class. It has been updated to use the `in` operator instead, which does look up the entire prototype chain.

## **Related issues**

Fixes #10057

## **Manual testing steps**

See #10057

### **Before**

See #10057

### **After**

https://github.com/MetaMask/metamask-mobile/assets/2459287/b7cf3e09-086b-4964-8180-e58195969e17

### DIFF
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -1340,7 +1340,8 @@ class Engine {
     for (const controller of controllers) {
       if (
         hasProperty(initialState, controller.name) &&
-        hasProperty(controller, 'subscribe') &&
+        // Use `in` operator here because the `subscribe` function is one level up the prototype chain
+        'subscribe' in controller &&
         controller.subscribe !== undefined
       ) {
         // The following type error can be addressed by passing initial state into controller constructors instead


FrederikBolding (approved):

tommasini (approved):

tommasini:
Awesome finding!

PR #10142, opened 2024-06-27
fix: memoize token list

OGPoyraz:
## **Description**

This PR aims to address performance issue on simulations happens only on mainnet.

## **Manual testing steps**

There should be no performance discrepancies for simulations between mainnet and any other network. 

### DIFF
--- a/app/components/hooks/DisplayName/useTokenList.test.ts
+++ b/app/components/hooks/DisplayName/useTokenList.test.ts
@@ -1,3 +1,4 @@
+import React from 'react';
 import { type TokenListMap } from '@metamask/assets-controllers';
 import { selectChainId } from '../../../selectors/networkController';
 import { selectUseTokenDetection } from '../../../selectors/preferencesController';
@@ -69,6 +70,17 @@ describe('useTokenList', () => {
     selectUseTokenDetectionMock.mockReturnValue(true);
     selectTokenListMock.mockReturnValue(TOKEN_LIST_MOCK);
     isMainnetByChainIdMock.mockReturnValue(true);
+
+    const memoizedValues = new Map();
+    jest.spyOn(React, 'useMemo').mockImplementation((factory, deps) => {
+      const depsKey = (deps as []).join('|');
+      if (memoizedValues.has(depsKey)) {
+        return memoizedValues.get(depsKey);
+      }
+      const newValue = factory();
+      memoizedValues.set(depsKey, newValue);
+      return newValue;
+    });
   });
 
   it('returns normalized STATIC_MAINNET_TOKEN_LIST if token detection is disabled and chain is mainnet', () => {
--- a/app/components/hooks/DisplayName/useTokenList.ts
+++ b/app/components/hooks/DisplayName/useTokenList.ts
@@ -1,3 +1,4 @@
+import React from 'react';
 import { type TokenListMap } from '@metamask/assets-controllers';
 import contractMap from '@metamask/contract-metadata';
 
@@ -7,7 +8,7 @@ import { selectUseTokenDetection } from '../../../selectors/preferencesControlle
 import { selectTokenList } from '../../../selectors/tokenListController';
 import { isMainnetByChainId } from '../../../util/networks';
 
-function normalizeTokenAddresses(tokenMap: TokenListMap) {
+export function normalizeTokenAddresses(tokenMap: TokenListMap) {
   return Object.keys(tokenMap).reduce((acc, address) => {
     const tokenMetadata = tokenMap[address];
     return {
@@ -25,11 +26,15 @@ export default function useTokenList(): TokenListMap {
   const chainId = useSelector(selectChainId);
   const isMainnet = isMainnetByChainId(chainId);
   const isTokenDetectionEnabled = useSelector(selectUseTokenDetection);
-  const tokenList = useSelector(selectTokenList) || [];
+  const tokenList = useSelector(selectTokenList);
 
-  if (!isTokenDetectionEnabled && isMainnet) {
-    return NORMALIZED_MAINNET_TOKEN_LIST;
-  }
+  const memoizedTokenList = React.useMemo(() => tokenList || [], [tokenList]);
+  const normalizedTokenList = React.useMemo(() => {
+    if (!isTokenDetectionEnabled && isMainnet) {
+      return NORMALIZED_MAINNET_TOKEN_LIST;
+    }
+    return normalizeTokenAddresses(memoizedTokenList);
+  }, [isTokenDetectionEnabled, isMainnet, memoizedTokenList]);
 
-  return normalizeTokenAddresses(tokenList);
+  return normalizedTokenList;
 }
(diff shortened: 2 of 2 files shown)


app/components/hooks/DisplayName/useTokenList.test.ts
```diff
@@ -69,6 +70,17 @@ describe('useTokenList', () => {
     selectUseTokenDetectionMock.mockReturnValue(true);
     selectTokenListMock.mockReturnValue(TOKEN_LIST_MOCK);
     isMainnetByChainIdMock.mockReturnValue(true);
+
+    const memoizedValues = new Map();
+    jest.spyOn(React, 'useMemo').mockImplementation((factory, deps) => {
```
OGPoyraz, line 75: This implementation is required in order to make use of `useMemo` in `testEnvironment: 'jest-environment-node'`. If we don't add it throws `TypeError: Cannot read properties of null (reading 'useMemo')`

digiwand (dismissed):

sleepytanya:
Works perfectly!
Bitrise build https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/677ffba8-5a0a-4e3b-b7bf-f783da9bba84?tab=artifacts
Samsung S24+, Android 14
iOS 15, iOS version 17.15.1

iOS:
https://github.com/MetaMask/metamask-mobile/assets/104780023/e32af1bd-fbc7-479b-8836-cb5e4ee7ba25


https://github.com/MetaMask/metamask-mobile/assets/104780023/a4004e83-d827-4480-824c-ff45f4ec791b


https://github.com/MetaMask/metamask-mobile/assets/104780023/f3104e80-674b-4c68-aac6-31829d08974f

Android:


https://github.com/MetaMask/metamask-mobile/assets/104780023/9d71d6a5-4831-4d65-a152-79314400ffc7


https://github.com/MetaMask/metamask-mobile/assets/104780023/7171cd02-0706-4ed2-9cd7-f5df1a3236db

app/components/hooks/DisplayName/useTokenList.ts
```diff
@@ -25,11 +26,15 @@ export default function useTokenList(): TokenListMap {
   const chainId = useSelector(selectChainId);
   const isMainnet = isMainnetByChainId(chainId);
   const isTokenDetectionEnabled = useSelector(selectUseTokenDetection);
-  const tokenList = useSelector(selectTokenList) || [];
+  const tokenList = useSelector(selectTokenList);
 
-  if (!isTokenDetectionEnabled && isMainnet) {
-    return NORMALIZED_MAINNET_TOKEN_LIST;
-  }
+  const memoizedTokenList = React.useMemo(() => tokenList || [], [tokenList]);
```
matthewwalsh0, line 31: No need for `React.` as I think the standard is to import all hooks directly.

app/components/hooks/DisplayName/useTokenList.ts
```diff
@@ -25,11 +26,15 @@ export default function useTokenList(): TokenListMap {
...
+  const normalizedTokenList = React.useMemo(() => {
+    if (!isTokenDetectionEnabled && isMainnet) {
+      return NORMALIZED_MAINNET_TOKEN_LIST;
+    }
+    return normalizeTokenAddresses(memoizedTokenList);
+  }, [isTokenDetectionEnabled, isMainnet, memoizedTokenList]);
 
-  return normalizeTokenAddresses(tokenList);
+  return normalizedTokenList;
```
matthewwalsh0, line 39: Could we skip a step and just return the above `useMemo` directly?

app/components/hooks/DisplayName/useTokenList.ts
```diff
@@ -25,11 +26,15 @@ export default function useTokenList(): TokenListMap {
...
   const isTokenDetectionEnabled = useSelector(selectUseTokenDetection);
-  const tokenList = useSelector(selectTokenList) || [];
+  const tokenList = useSelector(selectTokenList);
 
-  if (!isTokenDetectionEnabled && isMainnet) {
-    return NORMALIZED_MAINNET_TOKEN_LIST;
-  }
+  const memoizedTokenList = React.useMemo(() => tokenList || [], [tokenList]);
+  const normalizedTokenList = React.useMemo(() => {
+    if (!isTokenDetectionEnabled && isMainnet) {
```
matthewwalsh0, line 33: Could we combine these two booleans into a single dependency to reduce the number of re-renders? Something like `shouldUseStaticList`?

app/components/hooks/DisplayName/useTokenList.ts
matthewwalsh0, line 31: Is this not redundant since it's returning itself based on itself changing?

matthewwalsh0: If the goal is to only fire when the token list data changes, we should use a `createDeepEqualSelector` to compare the actual values, rather than firing on every new reference.

OGPoyraz: I remember that I've added because of the linter. So the logical experssion was making it to change it on every render. But now noticed that we don't really need to assign empty object to tokenList. It already sets an empty object in initial state. 
Hence I am removing this and the logical expression.

app/components/hooks/DisplayName/useTokenList.ts
OGPoyraz: Done

app/components/hooks/DisplayName/useTokenList.ts
OGPoyraz: Done

app/components/hooks/DisplayName/useTokenList.ts
OGPoyraz: Good call, done.

app/components/hooks/DisplayName/useTokenList.ts
```diff
@@ -25,11 +26,13 @@ export default function useTokenList(): TokenListMap {
   const chainId = useSelector(selectChainId);
   const isMainnet = isMainnetByChainId(chainId);
   const isTokenDetectionEnabled = useSelector(selectUseTokenDetection);
-  const tokenList = useSelector(selectTokenList) || [];
+  const tokenList = useSelector(selectTokenList);
```
matthewwalsh0, line 29: This selector could still be causing us to normalize more than we need to, given my understanding is any token list controller state update will fire that selector as it's not a deep equal.

But if this fixes the initial performance issue, then that can be a future optimization.

matthewwalsh0 (approved):

app/components/hooks/DisplayName/useTokenList.ts
OGPoyraz: But isn't normalization happen in the `useMemo` function, which will not hit on every render? 
With current changes it only hits normalization if there is a change in the selected `tokenList` which is something we want.

matthewwalsh0: That's exactly what we want, but this is a standard selector so I think it returns a fresh reference every time any of the token list controller state changes, which will also fire the normalization even if the content of the array is the same.

It could be state updates are rare in the token list controller so it's a non-issue, but just wanted to highlight it.

OGPoyraz: I see what you mean now, thanks for the explanation Matthew <3

PR #10164, opened 2024-06-28
fix: Stop running fencing logic on `node_modules`

FrederikBolding:
## **Description**

Stops running fencing logic on `node_modules`, mirroring the extension implementation. There is no reason for us to run fencing on code coming from `node_modules` anyway. This may have a potential positive performance impact as well during build-time.


Jonathansoufer (dismissed):

Jonathansoufer (approved):
LGTM!

PR #10189, opened 2024-07-01
feat: Revamp Snap connection screen

FrederikBolding:
## **Description**

This PR reworks the Snaps connection screen to add the Snap icon and simplify the implementation. It also tweaks the fencing to allow connecting to Snaps with only the `preinstalled-snaps` flag on.


Jonathansoufer (approved):
LGTM!

PR #10214, opened 2024-07-02
feat: Partially local WebView for Snaps

FrederikBolding:
## **Description**

Replaces the remote WebView with a locally bundled WebView. This WebView still uses a remote iframe for execution, but the initial load will be using all local code. Dependant on https://github.com/MetaMask/snaps/pull/2528

Also moves the `SnapsExecutionWebView` to the App index to have it available even when the application is locked, also prevents issues with breaking the execution environment when locking the app.


Jonathansoufer (dismissed):
LGTM!

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -9,6 +9,9 @@ import { createStyles } from './styles';
 import { WebViewInterface } from '@metamask/snaps-controllers/dist/types/services/webview/WebViewMessageStream';
 import { WebViewError } from '@metamask/react-native-webview/lib/WebViewTypes';
 import { PostMessageEvent } from '@metamask/post-message-stream';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
tommasini, line 13: If the error is expected, can we change for the usage of ts-expect-error and mention why it's expected?
We can also create a type for it on `declarations` file

Jonathansoufer: This is added to ignore the TS warning about `Cannot find module '@metamask/snaps-execution-environments/dist/browserify/webview/index.html' or its corresponding type declarations.` since it doesn't get resolved.

tommasini (dismissed):

tommasini (approved):
LGTM! It's missing  a QA label, since this is not covered by E2E I would suggest add the label "no qa needed"

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -9,6 +9,8 @@ import { createStyles } from './styles';
 import { WebViewInterface } from '@metamask/snaps-controllers/dist/types/services/webview/WebViewMessageStream';
 import { WebViewError } from '@metamask/react-native-webview/lib/WebViewTypes';
 import { PostMessageEvent } from '@metamask/post-message-stream';
+// @ts-expect-error Can't type a distritibuted html file
```
naugtur, line 12: have you tried typing it inline with jsdoc syntax? (not sure if it works in ts files tho)

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -89,13 +89,11 @@ export class SnapsExecutionWebView extends Component {
...
             }
-            source={{
-              uri: SNAPS_EE_URL,
-            }}
+            source={WebViewHTML}
             onMessage={this.onWebViewMessage}
             onError={this.onWebViewError}
             onLoadEnd={this.onWebViewLoad}
-            originWhitelist={['https://execution.metamask.io*']}
+            originWhitelist={['*']}
```
naugtur, line 96: (non-blocking) That's a security regression. You sure this can't be avoided?
Not a lot is lost, but if 2 layers of security fail, this used to stop the successful attacker from navigating elsewhere

app/lib/snaps/SnapsExecutionWebView.tsx
Jonathansoufer: Yep, unfortunately that doesn't work.

app/lib/snaps/SnapsExecutionWebView.tsx
Jonathansoufer: unfortunately this is a requirement when webview is using local source. Without this, the whole feature is useless.

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -9,6 +9,8 @@ import { createStyles } from './styles';
 import { WebViewInterface } from '@metamask/snaps-controllers/dist/types/services/webview/WebViewMessageStream';
 import { WebViewError } from '@metamask/react-native-webview/lib/WebViewTypes';
 import { PostMessageEvent } from '@metamask/post-message-stream';
+// @ts-expect-error Can't type a distritibuted html file
+import WebViewHTML from '@metamask/snaps-execution-environments/dist/browserify/webview/index.html';
```
naugtur, line 13: (should) make it an official export that doesn't depend on the `dist/browserify` part of the path so that changing the path is not a breaking change.

Also, since it's being built anyway, there's nothing stopping you from generating a JS or JSON file with the html string embedded in it that's easy to import and type without relying on specific bundler features.

naugtur (approved):

Jonathansoufer (approved):
Approved on behalf of Snaps team since we got already two reviewers from other teams.

naugtur:
@FrederikBolding @Jonathansoufer 
Here's what people seem to be using to make the html string be considered a secure context: https://github.com/webview-crypto/react-native-webview-crypto/blob/master/index.js#L176

PR #10241, opened 2024-07-03
chore: add Sentry snapshot with masked data

tommasini:
## **Description**

This PR introduces snapshot with masked data for metamask mobile.
Here is an example of masked data on sentry

```
{"accounts": {"reloadAccounts": false}, "alert": {"autodismiss": null, "content": null, "data": null, "isVisible": false}, "bookmarks": [], "browser": {"activeTab": null, "favicons": [], "history": [], "tabs": [], "visitedDappsByHostname": {}, "whitelist": []}, "collectibles": {"favorites": {}, "isNftFetchingProgress": false}, "engine": {"backgroundState": {"AccountTrackerController": [Object], "AccountsController": [Object], "AddressBookController": [Object], "ApprovalController": [Object], "AssetsContractController": [Object], "CurrencyRateController": [Object], "GasFeeController": [Object], "KeyringController": [Object], "LoggingController": [Object], "NetworkController": [Object], "NftController": [Object], "NftDetectionController": "object", "PPOMController": [Object], "PermissionController": [Object], "PhishingController": [Object], "PreferencesController": [Object], "SmartTransactionsController": [Object], "SnapController": [Object], "SubjectMetadataController": [Object], "SwapsController": [Object], "TokenBalancesController": "object", "TokenDetectionController": [Object], "TokenListController": [Object], "TokenRatesController": [Object], "TokensController": [Object], "TransactionController": [Object]}}, "experimentalSettings": {"securityAlertsEnabled": true}, "fiatOrders": "object", "infuraAvailability": {"isBlocked": false}, "inpageProvider": {"networkId": "1"}, "legalNotices": {"newPrivacyPolicyToastClickedOrClosed": true, "newPrivacyPolicyToastShownDate": null}, "modals": {"collectibleContractModalVisible": false, "dappTransactionModalVisible": false, "networkModalVisible": false, "receiveAsset": undefined, "receiveModalVisible": false, "shouldNetworkSwitchPopToWallet": true, "signMessageModalVisible": true}, "navigation": {"currentBottomNavRoute": "Wallet", "currentRoute": "Login"}, "networkOnboarded": {"networkOnboardedState": {}, "networkState": {"nativeToken": "", "networkType": "", "networkUrl": "", "showNetworkOnboarding": false}, "switchedNetwork": {"networkStatus": false, "networkUrl": ""}}, "notification": {"notification": {"notificationsSettings": [Object]}, "notifications": []}, "onboarding": {"events": []}, "privacy": {"approvedHosts": {}, "revealSRPTimestamps": []}, "rpcEvents": {"signingEvent": {"eventStage": "idle", "rpcName": ""}}, "sdk": {"approvedHosts": {}, "connections": {}, "dappConnections": {}, "wc2Metadata": undefined}, "security": {"allowLoginWithRememberMe": false, "automaticSecurityChecksEnabled": false, "dataCollectionForMarketing": null, "hasUserSelectedAutomaticSecurityCheckOption": false, "isAutomaticSecurityChecksModalOpen": false, "isNFTAutoDetectionModalViewed": false}, "settings": {"basicFunctionalityEnabled": true, "hideZeroBalanceTokens": false, "lockTime": 30000, "primaryCurrency": "ETH", "searchEngine": "DuckDuckGo", "useBlockieIcon": true}, "signatureRequest": "object", "smartTransactions": {"optInModalAppVersionSeen": null}, "swaps": "object", "transaction": "object", "transactionMetrics": "object", "user": {"ambiguousAddressEntries": "object", "appTheme": "os", "backUpSeedphraseVisible": false, "gasEducationCarouselSeen": false, "initialScreen": "", "isAuthChecked": false, "loadingMsg": "", "loadingSet": false, "passwordSet": true, "protectWalletModalVisible": false, "seedphraseBackedUp": true, "userLoggedIn": true}, "wizard": {"step": 1}}
```

image

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1843

## **Manual testing steps**

1. Go to this page...
2.
3.


tommasini:
The failed quality gate do not seem accurate `maskKey === undefined` it's needed and not redundant!

NicolasMassart:
> The failed quality gate do not seem accurate `maskKey === undefined` it's needed and not redundant!

The `maskKey === undefined` check might not be necessary if you want to handle it in the same way as `false`. So if you use `maskKey == false` both `false` and `undefined` values will be covered. At least that's my understanding.

tommasini:
That would mean that `maskKey == false` would cover for both? It's not what it seems, but most likely `!maskKey` would cover for both, but since there is an else scenario, Im not sure if we should do that, we are not accounting for key that are `null` or empty strings or 0, those values would fall into the else and throw an error!

image

I bring that method from [extension](https://github.com/MetaMask/metamask-extension/blob/8e00507f9c98ad18294448f3fdedcf92b19a00b3/shared/modules/object.utils.js#L23)

app/util/sentry/utils.js
```diff
@@ -134,6 +408,16 @@ function rewriteReport(report) {
...
+
+    const appState = store?.getState();
+    if (appState) {
+      const maskedState = maskObject(appState, SENTRY_STATE);
+      report.contexts.appState = maskedState;
+    }
+
+    if (!report.contexts.appState) {
+      report.contexts.appState = {};
+    }
```
NicolasMassart, line 420:
```suggestion
    if (appState) {
      const maskedState = maskObject(appState, SENTRY_STATE);
      report.contexts.appState = maskedState || {};
    }
```

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+ * with dynamic keys, and applies a rule (either `true`, `false`, or a
+ * sub-mask`) to every property in that object.
+ *
+ * If a property is excluded, its type is included instead.
+ *
+ * @param {object} object - The object to mask
+ * @param {{[key: string]: object | boolean}} mask - The mask to apply to the object
+ */
+export function maskObject(object, mask) {
+  if (!object) return {};
```
NicolasMassart, line 371: S: could you test the mask param too? Or maybe have a default value for this function mask param?

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+  let maskAllProperties = false;
+  if (Object.keys(mask).includes(AllProperties)) {
+    if (Object.keys(mask).length > 1) {
+      throw new Error('AllProperties mask key does not support sibling keys');
+    }
+    maskAllProperties = true;
+  }
```
NicolasMassart, line 378: Q: could you explain what this part is for exactly? Some code comment?

app/util/sentry/utils.js
```diff
@@ -6,6 +6,239 @@ import DefaultPreference from 'react-native-default-preference';
...
+      },
+      AddressBookController: {
+        addressBook: false,
+      },
+      ApprovalController: {
+        approvalFlows: false,
+        pendingApprovals: false,
+        pendingApprovalCount: false,
+      },
+      AssetsContractController: {},
```
NicolasMassart, line 45: Q: what will the mask be for this with {} ?

app/util/sentry/utils.js
```diff
@@ -6,6 +6,239 @@ import DefaultPreference from 'react-native-default-preference';
...
+      },
+      NftController: {
+        allNftContracts: false,
+        allNfts: false,
+        ignoredNfts: false,
+      },
+      NftDetectionController: false,
+      PPOMController: {
+        storageMetadata: [],
+        versionInfo: [],
```
NicolasMassart, line 86: Q: what will the mask be for this with [] ?

app/util/sentry/utils.test.ts
```diff
@@ -126,4 +129,225 @@ describe('captureSentryFeedback', () => {
       expectedUserFeedback,
     );
   });
+  describe('maskObject', () => {
+    it('mask our fixture of intial root state', () => {
```
NicolasMassart, line 133: S: Could you test a state with actual values that would cover the specific cases you test in the util function? See how "accounts" for instance masks with a real value. I feel the fixture root state is too empty to provide an accurate masking test. You have to at least test it of course, but adding a more realistic one would help.

Also, just some naming:
```suggestion
    it('masks initial root state fixture', () => {
```

app/util/sentry/utils.test.ts
```diff
@@ -126,4 +129,225 @@ describe('captureSentryFeedback', () => {
...
+        signatureRequest: 'undefined',
+        smartTransactions: {
+          optInModalAppVersionSeen: null,
+        },
+        swaps: 'undefined',
+        transaction: 'undefined',
+        transactionMetrics: 'object',
+        user: {},
+        wizard: undefined,
+      });
```
NicolasMassart, line 350: S: as the output you are testing is a big generated JSON, and it's not really useful to have it visible in the test itself (you just copy/pasted the initial return and no one reads each value and removing it would make the test more readable), that's a good use case for using snapshots. Snapshots allow to compare a json, it's not necessarily just for UIs. This seems a good use case for it to me here.

```suggestion
      expect(maskedState).toMatchSnapshot();
```

app/util/sentry/utils.js
tommasini: I don't replicated that on the test examples, but yes , I will be adding unit tests to proof this!
image

app/util/sentry/utils.test.ts
tommasini: Yeah you are totally right! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/26ba148811d8c108dfe5ebb850e4a5fe17d9a232)

app/util/sentry/utils.js
tommasini: Refactored on this commit: https://github.com/MetaMask/metamask-mobile/pull/10241/commits/26ba148811d8c108dfe5ebb850e4a5fe17d9a232

app/util/sentry/utils.test.ts
tommasini: Refactored on this commit: https://github.com/MetaMask/metamask-mobile/pull/10241/commits/26ba148811d8c108dfe5ebb850e4a5fe17d9a232

app/util/sentry/utils.js
tommasini: Addressed on this commit: https://github.com/MetaMask/metamask-mobile/pull/10241/commits/26ba148811d8c108dfe5ebb850e4a5fe17d9a232

app/util/sentry/utils.js
tommasini: >  if (Object.keys(mask).includes(AllProperties)) {

This is looking if the keys on the mask object contains the Symbol '*'

> if (Object.keys(mask).length > 1) {
If the it includes the Symbol '*' and includes more than one key, the iteration that happens after only support one key being '*' as we can see here when we are accessing the object on mask
```
    const maskKey = maskAllProperties ? mask[AllProperties] : mask[key];
```

If the value on the mask is true we can sho the values of the state (we do not obfuscate them)

>     } else if (maskKey && typeof maskKey === 'object') {
If it's an object, it calls function again since it's goes deeper on the object tree

>  } else if (maskKey === undefined || maskKey === false) {
If it's undefined or false we pass the type of the object to that property, obfuscating their real value

app/util/sentry/utils.js
tommasini: If it an object what will happen is:

Will enter the maskObject function, will fall into the second condition: `else if (maskKey && typeof maskKey === 'object')` then it will fall into the first condition : `if (maskKey === true) {` and show an empty object on sentry

app/util/sentry/utils.js
tommasini: Same [here](https://github.com/MetaMask/metamask-mobile/pull/10241#discussion_r1666999564)

app/util/sentry/utils.js
tommasini: Taking the example of assetsContracController we can see that it do not appear on Sentry 
image

app/util/sentry/utils.js
```diff
@@ -6,6 +6,239 @@ import DefaultPreference from 'react-native-default-preference';
...
+
+/**
+ * This symbol matches all object properties when used in a mask
+ */
+const AllProperties = Symbol('*');
+
+// This describes the subset of background controller state attached to errors
+// sent to Sentry These properties have some potential to be useful for
+// debugging, and they do not contain any identifiable information.
+export const SENTRY_STATE = {
```
NicolasMassart, line 19: S: Now that I read this a bit more I feel it could be called `sentryStateMask` for more clarity (and also avoid uppercase const names, that's not in guidelines and usually more for env vars.)
```suggestion
export const sentryStateMask = {
```

app/util/sentry/utils.test.ts
```diff
@@ -126,4 +128,360 @@ describe('captureSentryFeedback', () => {
       expectedUserFeedback,
     );
   });
+
+  describe('maskObject', () => {
+    const rootState = {
```
NicolasMassart, line 133: Q: how to make sure this will be in sync with our real state? it will have to be updated each time the real rootState will be updated in the app.
Should we move this to TS to be able to implement RootState from app/reducer and ensure a link between both forcing des to update this mask test too?

tommasini: Not sure if I understood this part "Should we move this to TS to be able to implement RootState"
But rootState it's meant to be a default state, default state doesn't have data, we want here to see if the mask is working as expected and we need to mock data for that to happen!

app/util/sentry/utils.js
```diff
@@ -6,6 +6,239 @@ import DefaultPreference from 'react-native-default-preference';
...
+
+/**
+ * This symbol matches all object properties when used in a mask
+ */
+const AllProperties = Symbol('*');
+
+// This describes the subset of background controller state attached to errors
+// sent to Sentry These properties have some potential to be useful for
+// debugging, and they do not contain any identifiable information.
+export const sentryStateMask = {
```
jpcloureiro, line 19: We can remove the keys with `false`, making the mask object smaller and easier to maintain

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+ * according to that sub-mask. The "AllProperties" symbol is used for objects
+ * with dynamic keys, and applies a rule (either `true`, `false`, or a
+ * sub-mask`) to every property in that object.
+ *
+ * If a property is excluded, its type is included instead.
+ *
+ * @param {object} object - The object to mask
+ * @param {{[key: string]: object | boolean}} mask - The mask to apply to the object
+ */
+export function maskObject(object, mask = true) {
```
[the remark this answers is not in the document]
jpcloureiro: We could also improve the naming of the `object` parameter.

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+  return Object.keys(object).reduce((state, key) => {
+    const maskKey = maskAllProperties ? mask[AllProperties] : mask[key];
+    if (maskKey === true) {
+      state[key] = object[key];
+    } else if (maskKey && typeof maskKey === 'object') {
+      state[key] = maskObject(object[key], maskKey);
+    } else if (maskKey === undefined || maskKey === false) {
+      state[key] = typeof object[key];
+    } else {
+      throw new Error(`Unsupported mask entry: ${maskKey}`);
```
jpcloureiro, line 388: Can't we just ignore unsupported mask entries?

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+  let maskAllProperties = false;
+  if (Object.keys(mask).includes(AllProperties)) {
+    if (Object.keys(mask).length > 1) {
+      throw new Error('AllProperties mask key does not support sibling keys');
+    }
+    maskAllProperties = true;
+  }
+  return Object.keys(object).reduce((state, key) => {
+    const maskKey = maskAllProperties ? mask[AllProperties] : mask[key];
+    if (maskKey === true) {
```
jpcloureiro, line 381: Lets give names to these conditions.

Example
```
const shouldPrintValue = maskKey === true
const shouldIterateSubMask = Boolean(maskKey) && typeof maskKey === 'object'
const shouldPrintType = maskKey === undefined || maskKey === false
```

jpcloureiro (changes requested):

app/util/sentry/utils.test.ts
```diff
@@ -126,4 +128,360 @@ describe('captureSentryFeedback', () => {
...
+      transactionMetrics: {
+        metricsByTransactionId: {},
+      },
+    };
+
+    it('masks initial root state fixture', () => {
+      const maskedState = maskObject(rootState, sentryStateMask);
+
+      expect(maskedState).toMatchSnapshot();
+    });
```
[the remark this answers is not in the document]
jpcloureiro: have a submask with 

```
{
  [AllProperties]: false,
  enabled: true
}
```

app/util/sentry/utils.js
jpcloureiro: Something along these lines would help better understand the code

```js
// This function applies a mask to an object, returning a new object where the properties
// are either copied from the original object, replaced with their type, or recursively masked.
function maskObject(object, mask = true) {
  // If the input object is null or undefined, return null.
  if (!object) return null;

  // Flag to indicate if all properties should be masked.
  let maskAllProperties = false;

  // Check if the mask contains the special AllProperties key.
  if (Object.keys(mask).includes(AllProperties)) {
    // If AllProperties key exists alongside other keys, throw an error.
    if (Object.keys(mask).length > 1) {
      throw new Error('AllProperties mask key does not support sibling keys');
    }
    // Set flag to true to mask all properties.
    maskAllProperties = true;
  }

  // Reduce the object's keys to a new masked state.
  return Object.keys(object).reduce((state, key) => {
    // Determine the mask for the current key.
    const maskKey = maskAllProperties ? mask[AllProperties] : mask[key];
    
    if (maskKey === true) {
      // If the mask is true, copy the property from the original object.
      state[key] = object[key];
    } else if (maskKey && typeof maskKey === 'object') {
      // If the mask is an object, recursively apply maskObject to the property.
      state[key] = maskObject(object[key], maskKey);
    } else if (maskKey === undefined || maskKey === false) {
      // If the mask is undefined or false, replace the property with its type.
      state[key] = typeof object[key];
    } else {
      // If the mask has an unsupported value, throw an error.
      throw new Error(`Unsupported mask entry: ${maskKey}`);
    }

    // Return the accumulated state.
    return state;
  }, {});
}


```

app/util/sentry/utils.js
tommasini: We need to keep it, because keeping it shows the type, and since we can't look at the data at least we try to understand if it's corrupted by the type
By removing it we will not log anything on that entry

app/util/sentry/utils.js
tommasini: Thanks for the review!!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/38fd89d0add8f15464854fb84e967b492c45b0f0)

app/util/sentry/utils.js
tommasini: Thanks for the review!!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/38fd89d0add8f15464854fb84e967b492c45b0f0)

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+ * with dynamic keys, and applies a rule (either `true`, `false`, or a
+ * sub-mask`) to every property in that object.
+ *
+ * If a property is excluded, its type is included instead.
+ *
+ * @param {object} object - The object to mask
+ * @param {{[key: string]: object | boolean}} mask - The mask to apply to the object
+ */
+export function maskObject(object, mask = true) {
+  if (!object) return null;
```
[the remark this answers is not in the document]
tommasini: Thanks for the review!!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/38fd89d0add8f15464854fb84e967b492c45b0f0)

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+ * symbol. `true` implies the property should be included, and `false` will
+ * exclude it. A sub-mask implies the property should be further masked
+ * according to that sub-mask. The "AllProperties" symbol is used for objects
+ * with dynamic keys, and applies a rule (either `true`, `false`, or a
+ * sub-mask`) to every property in that object.
+ *
+ * If a property is excluded, its type is included instead.
+ *
+ * @param {object} object - The object to mask
+ * @param {{[key: string]: object | boolean}} mask - The mask to apply to the object
```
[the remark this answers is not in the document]
tommasini: Thanks for the review!!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/38fd89d0add8f15464854fb84e967b492c45b0f0)

app/util/sentry/utils.test.ts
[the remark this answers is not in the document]
tommasini: Thanks for the review!!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/38fd89d0add8f15464854fb84e967b492c45b0f0)

app/util/sentry/utils.js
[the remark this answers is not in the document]
tommasini: Thanks for the review!!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/38fd89d0add8f15464854fb84e967b492c45b0f0)

app/util/sentry/utils.js
```diff
@@ -117,6 +350,47 @@ function removeSES(report) {
...
+export function maskObject(object, mask = true) {
+  if (!object) return null;
+  let maskAllProperties = false;
+  if (Object.keys(mask).includes(AllProperties)) {
+    if (Object.keys(mask).length > 1) {
+      throw new Error('AllProperties mask key does not support sibling keys');
+    }
+    maskAllProperties = true;
+  }
+  return Object.keys(object).reduce((state, key) => {
```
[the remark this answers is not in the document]
tommasini: Thanks for the review!! Addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/ac2e0396566a0acccbba58676edb4e6f7f72fdad)

tommasini:
@jpcloureiro Thanks for the review!!!! I believe I addressed/answered everything!

app/util/sentry/utils.js
jpcloureiro: Should log the type of keys not defined in the mask

```js
const shouldPrintType = maskKey === undefined || maskKey === false;

if (shouldPrintType) {
  maskedObject[key] = typeof objectToMask[key];
}
```

jpcloureiro: For the sub-masks that only have keys with `false`, we could use the `[AllProperties]: false`

For the ones that have siblings as `true`, it is safe to remove the keys with `false`

This could be refactored with an enum to cover more cases. I'll leave it up to the author to decide if it is worth the effort.

tommasini: You are right! Thank you so much, is much better now, addressed on this [commit](https://github.com/MetaMask/metamask-mobile/pull/10241/commits/6ecf7433bd9e195454d4c9ce977b728e0fedf111)

jpcloureiro (dismissed):
lgtm

NicolasMassart (dismissed):
Looks good to me.

NicolasMassart (approved):
Looks good to me.

PR #9652, opened 2024-05-16
chore(ramp): upgrade sdk to 1.28.1

wachunei:
## **Description**

This PR upgrades the Ramp SDK to version 1.28.1. 

This version changes the networks chain id type from number to string.
There must be a removal of all of the casting it was done to perform this change keeping backwards compatibility. This will be addressed in an upcoming PR.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/RAMPS-1613

## **Manual testing steps**

@bkirb Please test this ~~against Staging and Production environments since they are different as we transition from chainIds numbers to strings~~ Update: These environments are not different anymore

1. Ramps Flows (Buy & Sell) must not be impacted.

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #10262, opened 2024-07-05
chore(ramp): upgrade sdk to 1.28.3 and remove chainId casting

wachunei:
## **Description**

This PR upgrades Ramp SDK version to 1.28.3, it updates `chainId` mock data to be `string` and removes the casting of its type, finishing what was pending from https://github.com/MetaMask/metamask-mobile/pull/9652

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/RAMPS-1733

## **Manual testing steps**

1. Ramps Flows (Buy & Sell) must not be impacted.


app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.test.tsx
```diff
@@ -66,7 +66,7 @@ const mockedNetworksDetails = [
   },
 ];
 
-function render(Component: React.ComponentType, chainId?: string) {
+function render(Component: React.ComponentType, chainId?: `0x${string}`) {
```
Cal-L, line 69: We can remove `chainId` since it's never used

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.tsx
```diff
@@ -62,7 +62,7 @@ function NetworkSwitcher() {
   const error = errorFetchingNetworks || errorFetchingNetworksDetail;
   const rampNetworks = useMemo(() => {
     const activeNetworkDetails: Network[] = [];
-    // TODO(ramp, chainId-string): filter supportedNetworks by EVM compatible chains (chainId are strings of decimal numbers)
+    // TODO(ramp, btc): filter supportedNetworks by EVM compatible chains (chainId are strings of decimal numbers)
```
Cal-L, line 65: Could you elaborate on the change and its meaning?

app/components/UI/Ramp/utils/index.test.ts
Cal-L, line 1: Would be good to add a test that passes a number instead of string as chainId for `isNetworkRampSupported`.

Cal-L (commented):
Left some comments

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.test.tsx
wachunei: It is used in line 83 and 367 of this file.

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.tsx
wachunei: `supportedNetworks` used to have only values like `"1"`, `"56"`, etc... but soon they will have [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-2.md) string values like `"bip122:000000000019d6689c085ae165831e93"`, so we need to filter those cases out of this functionality.

We are currently excluding it in practice with the try/catch in the line 68, but we must improve it once we start the effort to support those chain ids.

app/components/UI/Ramp/utils/index.test.ts
wachunei: The goal was to have them there only while we did the transition from number to string, now that we deliver strings only it does not make sense to have it there. If it keep one in there then we will get a TypeScript error.

Cal-L: For tests, it's fine if we add a ts-expect comment to the test since we expect that the input is incorrect.

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.test.tsx
Cal-L: Oh, what I meant was that there are no instances where `chainId` is passed into `render` and therefore is never used.

wachunei: There is one in line 367 https://github.com/MetaMask/metamask-mobile/pull/10262/files/861c904420ff6ee86d25de3311e14f4c1775eb88#diff-7819aaffe352000e97942bf6d71d96e08c2759a523cc7085ca4677b3f7f40992R367

app/components/UI/Ramp/utils/index.test.ts
wachunei: What I mean is passing a number will not work, since it is not the format we expect, the same way as it wont work if I pass an object or a hex string.

Cal-L: Yep, I think we're talking about the same thing. The unexpected format was the test that I was recommending to add so that we can ensure the function doesn't crash if someone passes a non string for whatever reason. TS should catch it but just to be safe

wachunei: I understand now. I added them here 3f981d715

Cal-L: Nice thanks!

Cal-L (dismissed):
LGTM

wachunei:
@Cal-L need your approval again please, a conflict resolution dismissed it 🤓

bkirb:
LGMT for QA ✅

tommasini (approved):

app/components/UI/Ramp/Views/NetworkSwitcher/NetworkSwitcher.test.tsx
tommasini, line 69: NIT: should we use `Hex` from `@metamask/utils` ?

wachunei: We should, will update in an upcoming PR.

PR #10264, opened 2024-07-08
fix: skip blockaid validations for users internal accounts

jpuri:
## **Description**

Skip blockaid validation for users internal accounts

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2633

## **Manual testing steps**

1. Make a send transaction to user's own account
2. Check that blockaid validation is not done

## **Screenshots/Recordings**
NA

- [X] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Extension Coding Standards](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/CODING_GUIDELINES.md).
- [X] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.


app/lib/ppom/ppom-util.ts
```diff
@@ -80,6 +81,22 @@ async function validateRequest(req: PPOMRequest, transactionId?: string) {
     return;
   }
 
+  if (req.method === 'eth_sendTransaction') {
```
matthewwalsh0, line 84: Would it add clarity to add a small local function such as `isInternalAccount(req)`?

matthewwalsh0 (dismissed):

app/lib/ppom/ppom-util.test.ts
```diff
@@ -110,6 +116,27 @@ describe('PPOM Utils', () => {
       expect(spyTransactionAction).toBeCalledTimes(0);
     });
 
+    it('should not validate if request is send to users own account ', async () => {
+      const spyTransactionAction = jest.spyOn(
+        TransactionActions,
+        'setTransactionSecurityAlertResponse',
+      );
+      MockEngine.context.PreferencesController.state.securityAlertsEnabled =
+        false;
```
digiwand, line 125:
```suggestion
        true;
```

I think this would be more effective if the setting were turned on which could call the validation, else we do not expect validate to run

app/lib/ppom/ppom-util.ts
```diff
@@ -61,6 +61,7 @@ async function validateRequest(req: PPOMRequest, transactionId?: string) {
     PPOMController: ppomController,
     PreferencesController,
     NetworkController,
+    AccountsController,
```
digiwand, line 64: [nit] could move to the top to preserve alphabetical order

digiwand (changes requested):
requesting minor update to test before merge

app/lib/ppom/ppom-util.test.ts
jpuri: Its `true` for all other tests, this particular test is check that validation is not done if its `false`.

jpuri: https://github.com/MetaMask/metamask-mobile/blob/f31edcacf86dba7c35664c96afbbd9b09664cb41/app/lib/ppom/ppom-util.test.ts#L25

app/lib/ppom/ppom-util.test.ts
```diff
@@ -110,6 +116,27 @@ describe('PPOM Utils', () => {
...
+      MockEngine.context.AccountsController.listAccounts = (() => [
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tommasini, line 127: We shouldn't use any anymore, could you change it and type it correctly?

jpuri: I fixed it

matthewwalsh0 (dismissed):

tommasini (dismissed):

matthewwalsh0 (dismissed):

app/lib/ppom/ppom-util.test.ts
```diff
@@ -118,6 +124,30 @@ describe('PPOM Utils', () => {
       expect(spyTransactionAction).toBeCalledTimes(0);
     });
 
+    it('should not validate if request is send to users own account ', async () => {
+      const spyTransactionAction = jest.spyOn(
+        TransactionActions,
+        'setTransactionSecurityAlertResponse',
+      );
+      MockEngine.context.PreferencesController.state.securityAlertsEnabled =
+        false;
```
digiwand, line 133:
```suggestion
```

app/lib/ppom/ppom-util.test.ts
digiwand: in this case, I think we can remove the line then. See suggestion https://github.com/MetaMask/metamask-mobile/pull/10264/files#r1676204298

It looks like the test above checks the validation is not done when this value is false `should not validate if preference securityAlertsEnabled is false`

I think in this test, `should not validate if request is send to users own account`, we want the feature to be enabled to properly check the feature does not run for internal accounts

jpuri: I am setting value to true here: https://github.com/MetaMask/metamask-mobile/blob/f31edcacf86dba7c35664c96afbbd9b09664cb41/app/lib/ppom/ppom-util.test.ts#L25

digiwand (approved):

tommasini (approved):

PR #10294, opened 2024-07-12
feat: create redux slice for featureFlags

frankvonhoven:
## **Description**
1. Implement LaunchDarkly on mobile app
2. Fetch flags and store in redux

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1652


app/util/featureFlags/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export default function launchDarklyURL(
+  metamaskBuildType = 'main',
+  metamaskEnvironment = 'production',
+): string {
+  const client = 'mobile';
+
+  let environment: string | undefined = environmentMapping[metamaskEnvironment];
+
+  if (!environment) {
```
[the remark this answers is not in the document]
frankvonhoven: As I'm looking at it now, these checks are unnecessary since there now there are default values. So `environment` will always exist as well as buildType.

app/core/redux/slices/featureFlags/index.ts
```diff
@@ -0,0 +1,65 @@
+import { PayloadAction, createSlice } from '@reduxjs/toolkit';
+
+export interface FeatureFlagsState {
+  featureFlags: object;
```
tommasini, line 4: This is awesome, so if I'm understand here it's where it's stored the featureFlags, can we have a commented typed object example, so the first implementing a feature flag have already where to add the type of it? I will say for example: "multichainV1", we should have a enum that can add the feature flag to typescript, wdyt?

app/reducers/index.ts
```diff
@@ -58,6 +59,9 @@ export interface RootState {
   engine: { backgroundState: EngineState };
   // TODO: Replace "any" with type
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  featureFlags: any;
```
tommasini, line 62: Based on the other comment, we want that type very much to exist and it could be an empty object for now here if there is no featureFlag created, does that make sense?

app/store/sagas/index.ts
```diff
@@ -124,8 +131,42 @@ export function* basicFunctionalityToggle() {
...
 }
 
+function arrayToObject(data: []): object {
+  return data.reduce((obj, current) => {
+    Object.assign(obj, current);
+    return obj;
+  }, {});
+}
+
+function* fetchFeatureFlags(): Generator {
```
tommasini, line 141: I never was a big sagas adopter, so I find always that I want something to learn here! Could you explain what made you have the decision to implement this with Sagas? Just asking for learning purposes!

tommasini:
We can update this branch with main to have the audit:ci pipeline fixed, also, could we add a PR description?

app/store/sagas/index.ts
frankvonhoven: Ah, actions and reducers are fine for manipulating the global state. 

But when we have side effects, like api calls, that's where sagas are useful. 

So rather than dirty our state changing logic (actions/reducers) we have sagas to handle logic that is not directly related to changing/updating state. 

Side note: In most apps there would be a "services" folder/file with all the api calls that the app would make, and the sagas would just call those. Like `fetchSomeData` and that would look like:
```
// app/services/services.ts

function fetchSomeData () {
   const response =  axios.get(someUrl)
    ... other logic - maybe error handling
   return response
}
```

app/reducers/index.ts
frankvonhoven: I think we can fix this 👍

app/core/redux/slices/featureFlags/index.ts
frankvonhoven: yeah I debated adding it because in theory we can add new flags in LD and they would be in the response, so it really is type `any`. But if we're going to use them via redux, we should know what's available. I think it's a better dev experience to type them. Good call :ty:

app/store/sagas/index.ts
tommasini: Thank you so much for the explanation Frank!! This answer was amazing! Always good to learn with you!

tommasini (dismissed):
Awesome! LGTM!

tommasini:
Ah it seems unit tests and tsc issues :(

app/core/AppConstants.ts
```diff
@@ -213,4 +213,8 @@ export default {
     'static.cx',
     'config-api.metamask.io/featureFlags',
   ],
+  FEATURE_FLAGS_API: {
+    BASE_URL: 'https://client-config.dev-api.cx.metamask.io',
```
sethkfman, line 217: Do you still want this pointing to `dev-api`?

frankvonhoven: UPDATED!

sethkfman (approved):
LGTM

PR #10350, opened 2024-07-19
feat: make add-team-label use the reusable workflow

itsyoboieltr:
## **Description**

Instead of copy-pasting the implementation of add-team-label to other repositories, such as metamask-extension and metamask-mobile, it should be possible to call the workflow directly from the github-tools repository, if we make it a reusable workflow. This is great, because if we make a change to the workflow in the github-tools repository, the changes propagate to other repositories as well.

More information about reusable workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows

This kind of aligns with https://github.com/MetaMask/metamask-extension/issues/25509, but not exactly. The shared library would still be needed nevertheless, as this is only sharing the workflow definition, but not the scripts that are run during the workflow. The scripts still need to exist in all the repositories (for now). Ideally, we would make the shared library an npm package that can be installed in the different repositories to allow for the sharing of scripts.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2760

## **Manual testing steps**

1. Have the workflow run the same as before, without it being defined in the repository.

## **Screenshots/Recordings**

Not applicable


package.json
```diff
@@ -83,7 +83,7 @@
     "deduplicate": "yarn yarn-deduplicate && yarn install",
     "set-version": "./scripts/set-versions.sh",
     "add-release-label-to-pr-and-linked-issues": "ts-node ./.github/scripts/add-release-label-to-pr-and-linked-issues.ts",
-    "add-team-label-to-pr": "ts-node ./.github/scripts/add-team-label-to-pr.ts",
+    "add-team-label": "ts-node ./.github/scripts/add-team-label.ts",
```
NicolasMassart, line 86: Q: Is it even useful to have it as a package.json script given it's never going to be run locally?

itsyoboieltr: @NicolasMassart 

You could run it locally for debugging/development. 

It also makes the workflow easier to read:
`yarn ts-node ./.github/scripts/add-team-label.ts`
vs
`yarn run add-team-label`

This is the current solution until we have a common library (as defined in https://github.com/MetaMask/metamask-extension/issues/25509). Once we have a common library which can be installed as a package, we can then remove the common scripts (like add-team-label.ts) and npm commands (like add-team-label) from the repositories.

.github/workflows/add-team-label.yml
```diff
@@ -7,31 +7,6 @@ on:
...
-
-      - name: Install dependencies
-        run: yarn --immutable
-
-      - name: Add team label to PR
-        id: add-team-label-to-pr
-        env:
-          RELEASE_LABEL_TOKEN: ${{ secrets.RELEASE_LABEL_TOKEN }}
-        run: yarn run add-team-label-to-pr
+    uses: metamask/github-tools/.github/workflows/add-team-label.yml@main
```
NicolasMassart, line 10: S: one last comment, could we make it target a specific commit instead of just the branch? This is useful to make sure that the behaviour is frozen and you don't have any surprise but also for security reasons it prevents someone to introduce bad behaviour on the original repo calling the action in case of corrupting github-tools repo.

itsyoboieltr: fixed

NicolasMassart (approved):
Looks good to me.
Thanks for the investigation on using the hash!

PR #10373, opened 2024-07-22
chore: Migrate default preference to mmkv

tommasini:
## **Description**

* Migration for migrate default preference data to mmkv 
* Refactor of the default preferences usage across the app

```
Before (Default Preferences)
{"@MetaMask:MetaMetricsId": "558de5e5-d382-4a9e-9f24-cce477116414", "@MetaMask:analyticsDataRecorded": "true", "@MetaMask:metricsOptIn": "agreed", "@MetaMask:onboardingWizard": "explored"}
```
```
After  (MMKV)
 MetaMetricsId 558de5e5-d382-4a9e-9f24-cce477116414
 analyticsDataRecorded true
 metricsOptIn agreed
 onboardingWizard explored
```
## **Related issues**

Fixes:

## **Manual testing steps**

1. install the app
2. login 
3. close onboarding flow
4. update the app with the code on this PR
5. the onboarding flow shouldn't appear


andreahaku (dismissed):
LGTM

app/core/Authentication/Authentication.test.ts
```diff
@@ -96,21 +114,18 @@ describe('Authentication', () => {
   it('should return a type AUTHENTICATION_TYPE.PASSWORD if the user exists and there are no available biometrics options but the password does not exist in the keychain', async () => {
     SecureKeychain.getSupportedBiometryType = jest.fn().mockReturnValue(null);
     await AsyncStorage.setItem(EXISTING_USER, TRUE);
+    SecureKeychain.getGenericPassword = jest.fn().mockReturnValue(null);
     const result = await Authentication.getType();
     expect(result.availableBiometryType).toBeNull();
-    expect(result.currentAuthType).toEqual(AUTHENTICATION_TYPE.REMEMBER_ME);
+    expect(result.currentAuthType).toEqual(AUTHENTICATION_TYPE.PASSWORD);
```
NicolasMassart, line 120: C: wow good catch!

app/store/migrations/050.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  it('migrates asyncStorage values to mmkv ', async () => {
+    for (const key in defaultPreferenceItems) {
+      await DefaultPreference.set(key, defaultPreferenceItems[key]);
+    }
+
+    await migrate({});
+
+    const keyValues = await DefaultPreference.getAll();
+
+    expect(Object.keys(keyValues).length).toBe(0);
```
NicolasMassart, line 29: This is not testing anything but your mock return value that is statically set to empty {}. You have to implement some sort of mock memory to make this check valuable and actually verify that migrate really removes items from the DefaultPreference.

You could check also that the asyncStorage eventually has the inital count of items that were in the DefaultPreference (excluding null values).

app/store/migrations/050.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * The goal of this migration is set all the data that was on DefaultPreference to MMKV
+ * and clean DefaultPreference data
+ * @param state
+ * @returns state
+ */
+export default async function migrate(state: unknown) {
+  const keyValues = await DefaultPreference.getAll();
+  if (keyValues) {
```
NicolasMassart, line 13: Is this test really needed? It [can't be undefined or null as it at least returns an empty map](https://github.com/kevinresol/react-native-default-preference/blob/16c1beb2dd203b417da1edecbeaf7819a99b66ba/android/src/main/java/com/kevinresol/react_native_default_preference/RNDefaultPreferenceModule.java#L85-L92), the `Object.keys(keyValues)` in the loop will then just not do any loop.

app/store/migrations/050.test.ts
```diff
@@ -0,0 +1,31 @@
+import migrate from './050';
+import DefaultPreference from 'react-native-default-preference';
+
+const defaultPreferenceItems: { [key: string]: string } = {
+  valueA: 'a',
+  valueB: 'true',
+  valueC: 'myValue',
```
NicolasMassart, line 7: add a null value to cover the migration code that checks for it.

app/store/migrations/index.test.ts
```diff
@@ -7,6 +7,11 @@ import {
 const defaultNodeEnv = process.env.NODE_ENV;
 jest.unmock('redux-persist');
 jest.mock('../../store', () => jest.fn());
+jest.mock('react-native-default-preference', () => ({
+  set: jest.fn(),
+  clear: jest.fn(),
+  getAll: jest.fn().mockReturnValue({}),
+}));
```
NicolasMassart, line 14: Q: why do you mock react-native-default-preference here and not StorageWrapper ? Not sure I get why this mock is needed.

NicolasMassart (commented):
Nice job on this, a lot of places to change indeed.
I commented some issue on the migration unit test as some parts are not testing anything to me (I can be wrong though) and suggested idea to make your migration coverage stronger.

Also I wonder if you could make something to deprecate the DefaultPreferences until we get rid of the dependency. Otherwise we get the risk that it's used again somewhere.
Thanks!

app/store/migrations/050.ts
tommasini: You are totally right! Amazing catch! Will remove that condition

app/store/migrations/050.test.ts
tommasini: Bue that will raise an TS issue with the `set` of DefaultPreferences, 
image

But I think it's worth it to add it and add a ts expect error since it's a test case

app/store/migrations/index.test.ts
tommasini: It was not recognizing the library when it was called [here](https://github.com/MetaMask/metamask-mobile/blob/04e0ff84157533a0e41acf720fcda78f452eb402/app/store/migrations/index.ts#L128) so we needed to mock it at that level

I believe we do not need to mock the StorageWrapper because we are already mocking the store [here](https://github.com/MetaMask/metamask-mobile/blob/5d519645209059964b295c5730f760e4a7e5336a/app/store/migrations/index.test.ts#L9), but I'm not 100% sure, I only discovered that the mock of react-native-default-preference mock was needed here because the unit tests were failing at that point that I shared above

app/core/Authentication/Authentication.test.ts
tommasini: Actually I'm confused with his change now tbh

app/store/migrations/050.test.ts
tommasini: I believe the best we can do is have a variable that we populate with the mocks, but we will not be actually testing anything as well.  I think this test is more to test the intentional behaviour because there is not much we can test since we are literally using other storage packages to do the behaviour we want!
Any ideas?

app/core/Authentication/Authentication.test.ts
tommasini: I will review this  now

app/store/migrations/050.test.ts
tommasini: Addressed [here](https://github.com/MetaMask/metamask-mobile/pull/10373/commits/c880ee512ac5ac6d9ea56b5b46fcc4d61b0ee13d)

app/store/migrations/050.ts
tommasini: Addressed [here](https://github.com/MetaMask/metamask-mobile/pull/10373/commits/c880ee512ac5ac6d9ea56b5b46fcc4d61b0ee13d)

app/store/migrations/050.test.ts
NicolasMassart: You can actually have a mock with memory. So the migration will use the mock and so you test the migration.

app/core/Authentication/Authentication.test.ts
tommasini: Ahhh reviewed, yeah this way is the aligned with the test case description! And the test was failing!

app/store/migrations/050.test.ts
NicolasMassart: You can actually have a mock with memory. So the migration will use the mock and so you test the migration.

tommasini: I think you are right! Let me try that!

NicolasMassart (approved):
Looks good to me.
Great work mate!

PR #10251, opened 2024-07-04
chore: remove remaining use of Preferences Controller for account information

owencraston:
## **Description**

1. What is the reason for the change?
- In previous issues we migrated all account information state to reference the the new accounts controller redux state. - This change furthers this effort and removes all remaining references to PreferencesController.identities and PreferencesController.selectedAddress. 
- All of the remaining references are non state values and thus are not going to be using the redux state implemented in the previous issues. Instead we will reference the Engine.context.AccountsController instance directly.
- This PR also removes `selectSelectedAddress` and `selectIdentities` from the available redux selectors to prevent future use of this data.
    - This is ideal since we eventually want to completely remove this data from the PreferencesController in favour of the AccountsController
- After this PR there should be `zero references to this data in the codebase`.

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/537

## **Review checklist**
- [x] confirm that there are no reference to `PreferencesController.identities` in the codebase
- [x] Confirm that there are no references to `PreferencesController.selectedAddress` in the codebase
- [x] Confirm that there are no references to `selectIdentities`
- [x] Confirm that there are no references to `selectSelectedAddress`

(There will still be use of these variables in the migration files that are migrating this data from PreferencesController to AccountsController. There should be no other references besides the migrations)

## **Manual testing steps**

1. Import a wallet ideally with funds
2. after onboarding, create a new account
5. edit then name of this new account
6. add a hardware wallet via ledger or QR
7. import a private key like orangefox.eth or another of your choosing
8. go through the send flow from an account with funds
9. EXPECT, all addresses, names and transaction data should match the account listed in the transaction
10. EXPECT, all addresses to be in checksum format (example 0xC4955C0d639D99699Bfd7Ec54d9FaFEe40e4D272 vs 0xc4955c0d639d99699bfd7ec54d9fafee40e4d272)
11. navigate to https://metamask.github.io/test-dapp/ in the mobile browser
12. connect one or more account
13. EXPECT, the connection modal should display the correct account information as well as the address in checksum format
14. perform a send on the test dapp
15. EXPECT, the send modal should display the correct account information and the address in checksum format
16. Disconnect the account from the test dapp
17. reconnect the account vai wallet connect
18. EXPECT, the connection should work and should display the correct account information as well as the address in checksum format

### **After**

Screenshot 2024-07-24 at 3 15 51 PM
Screenshot 2024-07-24 at 3 16 02 PM
Screenshot 2024-07-24 at 3 16 37 PM
[...]

PR #10404, opened 2024-07-25
fix: aggregated balance calculation

bergeron:
## **Description**

Fixes the calculation for how much the aggregated balance changed in the last 1 day.

The fiat price 1 day ago was calculated incorrectly.  For example it would calculate:

```json
{"ethFiat1dAgo": 1.2011985977032296, "ethFiat": 1.28, "pricePercentChange1d": -6.156359554435198}
```

But this is backwards.  If the price went down 6% in the last 24 hours, you would have had *more* fiat $ in the past, not less.   This broke further calculations. 

## **Related issues**
#10406 

## **Manual testing steps**

1. Open a wallet with just eth, no erc20 tokens (or hide them all)
2. Verify the aggregated fiat balance % matches the balance % for ETH in the token list

### **Before**

Simplest case: a wallet with just eth.  The aggregated balance should match ETH since it's the only token:

Screenshot 2024-07-24 at 6 46 20 PM

### **After**

Screenshot 2024-07-24 at 6 45 54 PM

Matches extension now:

Screenshot 2024-07-24 at 6 41 16 PM


app/component-library/components-temp/Price/AggregatedPercentage/AggregatedPercentage.tsx
```diff
@@ -35,7 +35,7 @@ const AggregatedPercentage = ({
   const amountChange = totalBalance - totalBalance1dAgo;
 
   const percentageChange =
-    ((totalBalance - totalBalance1dAgo) / totalBalance1dAgo) * 100 || 0;
+    ((totalBalance - totalBalance1dAgo) / totalBalance) * 100 || 0;
```
bergeron, line 38: The fact I had to switch the denominator may indicate more needs to be fixed

app/core/Engine.ts
```diff
@@ -1670,12 +1666,13 @@ class Engine {
...
+
           const tokenBalance1dAgo =
-            tokenBalanceFiat +
-              (tokenBalanceFiat *
-                tokenExchangeRates?.[item.address as `0x${string}`]
-                  ?.pricePercentChange1d) /
-                100 || tokenBalanceFiat;
+            tokenPricePercentChange1d !== undefined
+              ? tokenBalanceFiat / (1 + tokenPricePercentChange1d / 100)
+              : tokenBalanceFiat;
```
salimtb, line 1675: can we add a test for this `getTotalFiatAccountBalance` , here is my suggestion:
```
it('should calculate total fiat account balance correctly', () => {
    // const engine = Engine.init(mockState);
    const mockState = {
      CurrencyRateController: {
        currentCurrency: 'usd',
        currencyRates: {
          ETH: {
            conversionRate: 2000, // Example rate: 1 ETH = 2000 USD
          },
        },
      },
      PreferencesController: {
        selectedAddress: '0x123',
      },
      AccountTrackerController: {
        accountsByChainId: {
          '0x1': {
            '0x123': {
              balance: '1000000000000000000', // 1 ETH
            },
          },
        },
      },
      TokenBalancesController: {
        contractBalances: {},
      },
      TokenRatesController: {
        marketData: {
          '0x1': {
            '0x0000000000000000000000000000000000000000': {
              pricePercentChange1d: -4.867211076998867, // Example decrease of 4.867211076998867%
            },
          },
        },
      },
      TokensController: {
        tokens: [],
      },
      NetworkController: {
        state: {
          providerConfig: {
            chainId: '0x1',
            ticker: 'eth',
          },
        },
      },
    };

    const engine = Engine.init(mockState);

    // Mock store.getState() to return the above mock state
    store.getState = jest.fn(() => ({
      settings: {
        showFiatOnTestnets: true,
      },
      engine: {
        backgroundState: mockState,
      },
    }));

    const result = engine.getTotalFiatAccountBalance();

    expect(result.ethFiat).toBe(2000);
    expect(result.ethFiat1dAgo).toBeCloseTo(2102.42, 0); // Calculated as 2000 / (1 - 4.867211076998867 / 100)
    expect(result.tokenFiat).toBe(0);
    expect(result.tokenFiat1dAgo).toBe(0);
  });
});
```

app/core/Engine.ts
```diff
@@ -1670,12 +1666,13 @@ class Engine {
             decimalsToShow,
           );
 
+          const tokenPricePercentChange1d =
+            tokenExchangeRates?.[item.address as Hex]?.pricePercentChange1d;
```
salimtb, line 1670: can we handle the undefined case here ? my suggestion is :
```
 const ethPriceChange1d =
      tokenExchangeRates?.[toHexadecimal(chainId)]?.[
        zeroAddress() as `0x${string}`
      ]?.pricePercentChange1d ?? 0;

    // Calculate the ETH Fiat value from 1 day ago
    ethFiat1dAgo = ethFiat / (1 + ethPriceChange1d / 100);
    ```

**Comment by salimtb** [comment]
i think we need the same fix [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Engine.ts#L1673) , my suggestion is :
```
// Get the percentage change in token price over the last day
          const tokenPriceChange1d =
            tokenExchangeRates?.[item.address as `0x${string}`]
              ?.pricePercentChange1d ?? 0;
          // Calculate the token Fiat value from 1 day ago
          const tokenBalance1dAgo =
            tokenBalanceFiat / (1 + tokenPriceChange1d / 100);
```

**Comment by sahar-fehri** [comment]
Thanks for the fix! 🙏 
Tested this while looking at extension;
Noticed that extension total price percentage is not changing like mobile does as you import new tokens? Can someone else confirm?

https://github.com/user-attachments/assets/6efaa2a1-0aad-42e8-a1fb-35c330bd5b87

**Comment by salimtb** [comment]
> https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/c20cf4b5-baa3-4081-9401-b76e6ac23bff

the aggregated balance is only implemented on mobile for now

File: app/core/Engine.ts [hunk 4]
```diff
@@ -1634,24 +1635,19 @@ class Engine {
       );
     }
 
+    const ethPricePercentChange1d =
```
sethkfman, line 1638: Can you extract this function and write a unit test for it?

darkwing:
We need to add tests for this.

bergeron:
> i think we need the same fix [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Engine.ts#L1673) , 

Should be there already, it's fixed for the erc20 token calculation too.

brianacnguyen (commented):
Not sure how DS was tagged to review this, as it does not touch DS

app/core/Engine.ts
bergeron: added unit tests in https://github.com/MetaMask/metamask-mobile/pull/10404/commits/f3d3e74fb3d783bf8059979feffd8bda991bc954

salimtb (dismissed):

app/core/Engine.test.js
```diff
@@ -91,4 +93,151 @@ describe('Engine', () => {
...
+      const totalFiatBalance = engine.getTotalFiatAccountBalance();
+      expect(totalFiatBalance).toStrictEqual({
+        ethFiat: 0,
+        ethFiat1dAgo: 0,
+        tokenFiat: 0,
+        tokenFiat1dAgo: 0,
+      });
+    });
+
+    // TODO: Why do these tests work when running individually but not when running all tests?
```
sahar-fehri, line 129: not a blocker, maybe we can remove these TODOs since tests pass

sahar-fehri (dismissed):

app/core/Engine.test.js
bergeron: oops yeah.  they pass in parallel after destroying the engine after each test

sahar-fehri (dismissed):

salimtb (dismissed):

bergeron:
https://github.com/MetaMask/metamask-mobile/pull/10251 changed how accounts are represented.  trying to figure out how  to mock now

app/core/Engine.test.js
```diff
@@ -91,4 +93,154 @@ describe('Engine', () => {
...
+      );
+
+      expect(totalFiatBalance).toStrictEqual({
+        ethFiat,
+        ethFiat1dAgo: ethFiat / (1 + ethPricePercentChange1d / 100),
+        tokenFiat,
+        tokenFiat1dAgo,
+      });
+    });
+  });
```
gambinish, line 245: Wondering why the test file is in JS and the files it's testing `Engine.ts` is in TS?

app/core/Engine.ts
```diff
@@ -1660,24 +1659,19 @@ class Engine {
...
       const { contractBalances: tokenBalances } = TokenBalancesController.state;
-      const { marketData } = TokenRatesController.state;
-      const tokenExchangeRates = marketData[chainId];
       tokens.forEach(
         (item: { address: string; balance?: string; decimals: number }) => {
-          const exchangeRate =
-            tokenExchangeRates && item.address in tokenExchangeRates
-              ? tokenExchangeRates[item.address as Hex]?.price
-              : undefined;
+          const exchangeRate = tokenExchangeRates?.[item.address as Hex]?.price;
```
gambinish, line 1674: Is there a reason we are typecasting here rather than typing it out as `{ address: Hex }`

gambinish (approved):
Approving with a couple comments out of curiosity ✅

app/core/Engine.test.js
bergeron: I think it's just older files that were written before typescript was the standard.  I did attempt to convert this to typescript, but there were too many errors to deal with.

app/core/Engine.ts
bergeron: The TokenRatesController types it as string in its state, so it would have to be changed there:

https://github.com/MetaMask/core/blob/d292d7b1bbab9514c7e826a983ef87a2137a6e25/packages/assets-controllers/src/TokenRatesController.ts#L52

We do try to type address as hex, so it'd be a good change

Cal-L (approved):
LGTM

PR #10468, opened 2024-07-30
chore: Capture custom rpc url in `trackEvent`

gambinish:
## **Description**

Custom `rpcUrls` were not being emitted to MetaMetrics, only `chainId`. Task was to emit a `NETWORK_ADDED` event whenever a custom RPC url was added in settings.

Additionally, I noticed that custom RPCs were being tracked from `source: 'Popular network list'` rather than `Custom network form`. This event should probably also be emitted on "Add Network" action, rather than "Switching Networks" action, since switching networks technically is not required in order to "Add Network".

## **Related issues**

Fixes: MMASSETS-298

## **Manual testing steps**

1. Go to Settings view
2. Add Popular network, validate that event gets emitted by with `source: "Popular network list"`
3. Add Custom network, validate that event gets emitted by with `source: "Custom network form"`

## **Screenshots/Recordings**

Popular Network:

https://github.com/user-attachments/assets/1b3bc09b-fbcc-4505-bea6-36151c5f5694

Custom Network

https://github.com/user-attachments/assets/e23fb2df-a174-4eba-8ef9-e75a0f42ccc8


app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,7 +93,34 @@ const NetworkModals = (props: NetworkProps) => {
...
+      toHex('42161'),
+      toHex('56'),
+      toHex('8453'),
+      toHex('10'),
+      toHex('11297108109'),
+      toHex('137'),
+      toHex('324'),
+    ];
 
+    if (!popularNetworkIds.includes(chainId)) {
```
bergeron, line 108: I don't think this is determinable based on chain id.  A network being on the popular list does not prevent you from adding it via the custom form.

gambinish: Refactored to use `showPopularNetworksModal` boolean here: https://github.com/MetaMask/metamask-mobile/pull/10468/commits/52320079f715e9cedd639dfcce09b7da5ada09de

legobeat (changes requested):
Has analysis been done on the resistance of this scheme against rainbow-table attacks?
 
https://github.com/MetaMask/metamask-mobile/pull/10468/files#r1697698617

app/components/UI/NetworkModal/index.tsx
```diff
@@ -92,6 +96,22 @@ const NetworkModals = (props: NetworkProps) => {
...
+        source: 'Popular network list',
+        symbol: ticker,
+      });
+    } else {
+      // emit custom network
+      trackAnonymousEvent(MetaMetricsEvents.NETWORK_ADDED, {
+        chain_id: toHex(chainId),
+        source: 'Custom network form',
+        symbol: ticker,
+        rpcUrl: ethers.utils.sha256(ethers.utils.toUtf8Bytes(rpcUrl)),
```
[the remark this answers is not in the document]
gambinish: Thanks for pointing this out. I am going to link the companion PR on extensions so that this issue can be tracked together: https://github.com/MetaMask/metamask-extension/pull/26203/files#diff-8e09bedbf5f945a64e22a83416aeb4c5f890318acfa999b5ebedd7ff102b58a3R973

gambinish:
> From PR description:
> 
> > As a privacy step, we are hashing the rpcUrl when it gets sent, and emitted the event with trackAnonymousEvent
> 
> Where is this happening?

These are no longer being hashed, since we are only sending over chains that are popular and in the `safeChains` list. Follows the same strategy as the extensions PR that got merged, which is effectively this same implementation.

I can update the description.

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +104,22 @@ const NetworkModals = (props: NetworkProps) => {
 
   const addNetwork = async () => {
     const validUrl = validateRpcUrl(rpcUrl);
+    if (!showPopularNetworkModal && safeChains) {
```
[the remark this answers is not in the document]
gambinish: In the `useSafeChains` hook, `safeChains` is undefined before they get fetched.

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +104,26 @@ const NetworkModals = (props: NetworkProps) => {
...
+        rpcUrl,
+        safeChains,
+      );
+      // emit custom network
+      trackEvent(MetaMetricsEvents.NETWORK_ADDED, {
+        chain_id: toHex(safeChain.chainId),
+        source: 'Custom Network Added',
+        symbol: safeChain.nativeCurrency.symbol,
+        rpcUrl: safeRPCUrl,
+      });
```
NicolasMassart, line 118: Q: is there any privacy risk by associating these infos with the user metrics ID?
If so, you should make it an anonymous event for instance like the following:
```ts
      trackEvent(MetaMetricsEvents.NETWORK_ADDED, {
        propertie: {
          source: 'Custom Network Added',
        }
        sensitiveProperties: {
          chain_id: toHex(safeChain.chainId),
          symbol: safeChain.nativeCurrency.symbol,
          rpcUrl: safeRPCUrl,
        },
      });
```
In this example only the source is not an anonymous value and will be sent with both anonymous (with anonymous ID) and non-anonymous (with actual user unique ID) events.
The sensitiveProperties will only be sent with the anonymous (with anonymous ID) event.

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +104,26 @@ const NetworkModals = (props: NetworkProps) => {
...
+    } else {
+      // emit popular network
+      trackEvent(MetaMetricsEvents.NETWORK_ADDED, {
```
NicolasMassart, line 121: Q: same question here about anonymous event

NicolasMassart (commented):
Only questions is if you evaluated privacy risk and if so, is there's any need for anonymous event.
Otherwise looks good!

app/components/UI/NetworkModal/index.tsx
gambinish: Good call! There is a slightly different pattern on Extension so I was a bit unsure. I'll go ahead and make this change.

app/components/UI/NetworkModal/index.tsx
gambinish: We were already emitting this event non-anonymously previously as chainId isn't/shouldn't necessarily be personally identifiable like a custom rpc url is (ie a dev running a local eth node). Changes in this PR are to add telemetry for custom rpcs

So, similar to how we are doing in this [PR](https://github.com/MetaMask/metamask-extension/pull/26266) on extension, we only want to capture the rpc url as a sensitive property.

app/components/UI/NetworkModal/index.tsx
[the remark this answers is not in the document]
gambinish: Would you say that this change is strictly necessary for this PR to get merged? I suppose we could do a reverse lookup on `NETWORKS_CHAIN_ID` but I'm not sure how much value that would add to this from a privacy perspective (if any) since the chainId and ticker that get evaluated here only ever get sent from our popular network list anyway?

If I'm missing something here please lmk. Might need to spell this out a bit more if I'm not seeing a big blindspot.

NicolasMassart (dismissed):
Looks good to me.

app/components/UI/NetworkModal/index.tsx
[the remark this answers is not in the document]
legobeat: > since the chainId and ticker that get evaluated here only ever get sent from our popular network list anyway?

> [(they are being evaluated from the hardcoded popular networks list in the source code, rather than API response)](https://github.com/MetaMask/metamask-mobile/pull/10468/files#r1742443707)

It may be me who's missing something, because it looks like this will trigger also for user-added custom networks, where `chainId`, `rpcUrl`, and `ticker` will e free-form user-defined values. Your comments indicate that this is not the case. Could you help by pointing mere to how this is limited only to networks on the popular-networks list?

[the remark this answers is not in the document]
gambinish: Can you message me directly in Slack plz? Would like to get this resolved so that we can move on, rather than going back and forth here.

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +105,30 @@ const NetworkModals = (props: NetworkProps) => {
...
+        chain_id: toHex(safeChain.chainId),
+        source: 'Custom Network Added',
+        symbol: safeChain.nativeCurrency.symbol,
+        sensitiveProperties: {
+          rpcUrl: safeRPCUrl,
+        },
+      });
+    } else {
+      Logger.log('Unable to capture custom network');
```
legobeat, line 130:
```suggestion
      Logger.log('MetaMetrics - Unable to capture custom network');
```

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +105,30 @@ const NetworkModals = (props: NetworkProps) => {
 
   const addNetwork = async () => {
     const validUrl = validateRpcUrl(rpcUrl);
```
legobeat, line 107:
```suggestion
    const isValidUrl = validateRpcUrl(rpcUrl);
```

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +105,30 @@ const NetworkModals = (props: NetworkProps) => {
...
+        symbol: safeChain.nativeCurrency.symbol,
+        sensitiveProperties: {
+          rpcUrl: safeRPCUrl,
+        },
+      });
+    } else {
+      Logger.log('Unable to capture custom network');
+    }
 
     setNetworkAdded(validUrl);
```
legobeat, line 133:
```suggestion
    setNetworkAdded(isValidUrl);
```

app/components/UI/NetworkModal/index.tsx
```diff
@@ -91,6 +105,30 @@ const NetworkModals = (props: NetworkProps) => {
...
   const addNetwork = async () => {
     const validUrl = validateRpcUrl(rpcUrl);
+    if (showPopularNetworkModal) {
+      // emit popular network
+      trackEvent(MetaMetricsEvents.NETWORK_ADDED, {
+        chain_id: toHex(chainId),
+        source: 'Popular network list',
+        symbol: ticker,
+      });
+    } else if (!showPopularNetworkModal && safeChains) {
```
legobeat, line 115:
```suggestion
    } else if (safeChains) {
```

legobeat (approved):
LGTM!

NicolasMassart (approved):
Looks good to me.

PR #10469, opened 2024-07-30
fix: Fix fiat price precision

gambinish:
## **Description**

On mobile, the fiat values in these screenshots below should have 2 decimal precision. Meaning, that all supported fiat currencies should default to two decimal places, and correct currency format in the Portfolio Balance component.

If the default currency is either a unsupported fiat currency, or is a cryptocurrency, fallback to default behavior of 5 decimal place precision.

## **Related issues**

Initially reported here: https://consensys.slack.com/archives/C05PK980JBC/p1719260553801789

Fixes: MMASSETS-279

## **Manual testing steps**

1. Go to token page
2. Hard code `total` to handle edge cases you want to test for (11.2, 9.8883 etc)

### **Before**
Screenshot 2024-07-30 at 12 56 20 AM

### **After**

Screenshot 2024-07-30 at 12 53 37 AM


app/components/UI/Tokens/__snapshots__/index.test.tsx.snap
```diff
@@ -337,7 +337,7 @@ exports[`Tokens should hide zero balance tokens when setting is on 1`] = `
                                 }
                                 testID="total-balance-text"
                               >
-                                0 USD
```
gambinish, line 340: Unsure why snapshots don't match unit tests here. I can look into this more.

gambinish: resolved ✅

app/components/UI/Tokens/__snapshots__/index.test.tsx.snap
```diff
@@ -337,7 +337,7 @@ exports[`Tokens should hide zero balance tokens when setting is on 1`] = `
...
+                                $0.00
```
bergeron, line 340: Do we know why this went from `0 USD` to `$0.00`

gambinish: Let me verify

gambinish: Okay, so the reason is because in app, the `currencyCode` is a mapping of lowercase currency codes `usd` `jpy` etc.

In the snapshot test, they are being mocked as `USD` and so the `renderFiat` it was previously not evaluating as it was in app, and unit tests weren't catching this.

```
if (currencySymbols[currencyCode]) {
  return `${currencySymbols[currencyCode]}${fiatFixed}`;
}
return `${fiatFixed} ${currencyCode.toUpperCase()}`;
```

The new `renderPortfolioBalance` does a `toLowerCase` check and also tests against this edge case.

bergeron: Thanks for looking.  `$0.00` makes more sense anyway, as thats the format i see in the actual product before + after.

bergeron:
This does resolve the precision of the overall portfolio balance.  But the precision of the individual tokens can still be off.  Lets see if we can fix that one too.

image

image

gambinish:
Closing in favor of https://github.com/MetaMask/metamask-mobile/pull/10599

PR #10529, opened 2024-08-01
feat: parse tx logs to update nft state

sahar-fehri:
## **Description**

This PR adds a subscription to `TransactionController:transactionStatusUpdated` and scans the received transaction logs to see if it it contains any topics indicating that ERC721 or ERC1155 NFTs has been transferred on that transaction. If it is the case, we will call `NftController.addNft` to add the NFTs to the state which allows the user to see the new NFT after navigating to the NFT tab.

## **Related issues**

Fixes:

## **Manual testing steps**
To make this test we will need 2 accounts; a sender and a receiver. The sender account will need to have at least one NFT.

1. Click on navigation icon and go to opensea.testnet
2. Import an ERC721 NFT on the sender account and list it for sale.
3. Switch to the receiver account.
4. Click on navigation tab and click on buy item to buy the same NFT listed for sale by the Sender account.
5. Once you get the notification that the transaction is confirmed, click on NFT tab and you should be able to see the new NFT on the receiver account.
6. Switch back to the Sender account and Click on the NFT. Once clicked the "Send" button in the NFT details page should be disabled. When you go back to the main NFT page, the NFT should be removed from the list.

### **After**

https://github.com/user-attachments/assets/9f049d56-1987-491c-acfd-9fd3c9a416e7


app/components/UI/Navbar/index.js
```diff
@@ -1153,6 +1153,7 @@ export function getNftDetailsNavbarOptions(
               name={IconName.MoreVertical}
               size={IconSize.Lg}
               style={innerStyles.headerBackIcon}
+              testID="more"
```
salimtb, line 1156: can we call this test-id "more-button" instead ?

app/components/Views/NftDetails/NftDetails.test.tsx
```diff
@@ -1,31 +1,17 @@
...
-    useNavigation: () => ({
-      navigate: mockNavigate,
-      setOptions: mockSetOptions,
-    }),
-    useFocusEffect: jest.fn(),
-  };
-});
+import NftDetails from './';
+// eslint-disable-next-line import/no-namespace
+import * as allSelectors from '../../../../app/reducers/collectibles/index.js';
```
salimtb, line 9: Suggestion: instead of using global import then use spyOn , it's better to mock the whole module using `jest.mock`.
using jest mock have some advantage like avoid leaking mock btw tests, better readable tests and improve performence, here is an example:

```

// Mock the selectors directly
jest.mock('../../../../app/reducers/collectibles/index.js', () => ({
  collectiblesSelector: mockedCollectiblesSelector,
  collectibleContractsOwnedSelector: mockedCollectibleContractsOwnedSelector,
}));

// use it inside specific test
    mockedCollectibleContractsOwnedSelector.mockReturnValue(data);

```
link to the documentation: https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options

app/components/Views/NftDetails/NftDetails.test.tsx
```diff
@@ -140,18 +126,122 @@ jest.mock('../../../util/navigation/navUtils', () => ({
...
+    const { getByTestId } = renderComponent(initialState);
+    const sendButton = getByTestId('send');
+
+    fireEvent.press(sendButton);
+
+    await waitFor(() => {
+      expect(sendButton.props.disabled).toBe(true);
+    });
+
+    spyOnContracts.mockRestore();
```
salimtb, line 219: i'm seeing the use of mockRestore after several tests , this can be common by using afterEach:
```
  afterEach(() => {
    cleanup();
    jest.restoreAllMocks(); // This will restore all mocked functions after each test
  });
  ```

File: app/core/Engine.ts [hunk 4]
```diff
@@ -2278,7 +2407,10 @@ export default {
 
   init(state: Record<string, never> | undefined, keyringState = null) {
     instance = Engine.instance || new Engine(state, keyringState);
-    Object.freeze(instance);
+    // Added this check because Object.freeze was making instance immutable, which prevents jest to add Spys to the object.
+    if (process.env.NODE_ENV === 'production') {
+      Object.freeze(instance);
+    }
```
gambinish, line 2413: This makes me a bit nervous. Not sure what the downstream effects of this would be. There might be some CI or testing environment flags that we aren't aware of?

I'm noticing that we had spys implemented previously. Were they not spying previously? Could you help understand why this is necessary now?

If changing the unit tests is way too hard, I wonder if this might be safer 

```
  import { DEVELOPMENT } from "./AppConstants"
   if (process.env.NODE_ENV !== DEVELOPMENT) {
      Object.freeze(instance);
    }
```

What do you think?

app/core/Engine.ts
```diff
@@ -1750,6 +1770,115 @@ class Engine {
...
+            const parsedLog = iface.parseLog({
+              data: singleLog.data as string,
+              topics: singleLog.topics as unknown as string[],
+            });
+            allNftParsedLog.push({
+              contract: singleLog.address,
+              ...parsedLog,
+            });
+          } catch (err) {
+            // ignore
```
gambinish, line 1848: Did you meant to log the caught err here?

gambinish (changes requested):
Leaving a couple questions here. Also, if theses are dev accounts that you are using, could we share the accounts that you used in your recording? Will make testing easier on my end 🙇

PR #10599, opened 2024-08-13
fix: Fiat price precision

gambinish:
## **Description**

Replaces: https://github.com/MetaMask/metamask-mobile/pull/10469 which I felt was touching many unrelated sections of code which I felt to be out of scope for this task. This implementation keeps the scope of this fix to three main areas listed in the testing steps below. Happy to discuss this decision further if you'd like. Screenshots of initial reported bug is in the other PR.

On mobile, the fiat values in the screenshots from the other PR should have 2 decimal precision, not 1. Meaning, that all supported fiat currencies should default to the [Intl.Number,](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) where this bug is apparent and noticeable.

If the default currency is either a unsupported fiat currency, or is a cryptocurrency, fallback to default behavior of 5 decimal place precision.

## **Related issues**

Initially reported here: https://consensys.slack.com/archives/C05PK980JBC/p1719260553801789

Fixes: MMASSETS-279

## **Manual testing steps**

1. Go to token page
2. Hard code `total` to handle edge cases you want to test for (11.2, 9.8883 etc)
3. Can also add several different token types like `PEPE` for weird/small values
4. Change `PrimaryCurrency` toggle, change `currentCurrency`

Yen shouldn't have decimals, other supported currencies should be to 2 decimal places w/ currency symbol.
Crypto and unsupported currencies to 5 decimals precision, trailing currency code

Places to check:

1. MainBalance on wallet screen
Screenshot 2024-08-13 at 10 00 11 PM

2. TokenBalances on wallet screen
Screenshot 2024-08-13 at 10 00 37 PM

3. Balance on AssetOverview screen
Screenshot 2024-08-13 at 10 01 18 PM

## **Screenshots/Recordings**

https://github.com/user-attachments/assets/275479ff-a7bb-42c3-94cc-7274aca784a3


app/util/number/index.js
```diff
@@ -683,6 +707,83 @@ export function renderFiat(value, currencyCode, decimalsToShow = 5) {
...
+      decimalsToShow,
+    );
+    const belowThreshold = value < threshold && value !== 0;
+    if (belowThreshold) {
+      isBelowMinimum = true;
+      valueToReturn = `${threshold} ${currencyCode.toUpperCase()}`;
+    } else {
+      valueToReturn = `${fixedBalance} ${currencyCode.toUpperCase()}`;
+    }
+  }
```
gambinish, line 783: Should I break this outer if/else into two smaller functions? I think some additional code comments here might also makes sense.

salimtb:
can we double check the conversion value on the swap and send flow as well ? i just did the verification and all seems to be correct but i don't have enough funds to test it correclly

app/util/number/index.js
salimtb: yea i agree , maybe start by adding some comments first

gambinish: Added comments here: https://github.com/MetaMask/metamask-mobile/pull/10599/commits/a8aa7563c1854efcb8e8af0a93903bf8af21064d

Imo I think this is sufficiently clear. If you want me to break into smaller unit functions I can.

gambinish:
> can we double check the conversion value on the swap and send flow as well ? i just did the verification and all seems to be correct but i don't have enough funds to test it correclly and deeply. for the send flow i'm getting more than 2 decimals ( you can check the video below )
> 
>  send-swap.mov

@salimtb During standup, we decided to handle the fiat precision on Send/Swap flow in a separate PR/ticket. Here is the reasoning:

1. This behavior isn't a regression
2. After looking at the logic, it will be a bit more involved than the implementation in this task (and not directly related)
3. There may be some UX feedback that will be required that could block (eg what happens when sending less than `$0.01`?)

Ultimately we didn't feel like this bugfix needed to wait here and get stale while working on the Send/Swap flow.

For reference, here is the new ticket: https://consensyssoftware.atlassian.net/browse/MMASSETS-351

salimtb:
> > can we double check the conversion value on the swap and send flow as well ? i just did the verification and all seems to be correct but i don't have enough funds to test it correclly and deeply. for the send flow i'm getting more than 2 decimals ( you can check the video below )
> > send-swap.mov
> 
> @salimtb During standup, we decided to handle the fiat precision on Send/Swap flow in a separate PR/ticket. Here is the reasoning:
> 
> 1. This behavior isn't a regression
> 2. After looking at the logic, it will be a bit more involved than the implementation in this task (and not directly related)
> 3. There may be some UX feedback that will be required that could block (eg what happens when sending less than `$0.01`?)
> 
> Ultimately we didn't feel like this bugfix needed to wait here and get stale while working on the Send/Swap flow.
> 
> For reference, here is the new ticket: https://consensyssoftware.atlassian.net/browse/MMASSETS-351

ok that's fine for me , thnx for sharing

salimtb (dismissed):

sethkfman (dismissed):
LGTM

app/components/UI/Tokens/__snapshots__/index.test.tsx.snap
```diff
@@ -3684,21 +3614,7 @@ exports[`Tokens should show all balance tokens when hideZeroBalanceTokens settin
...
-                                      "fontSize": 16,
-                                      "fontWeight": "500",
-                                      "letterSpacing": 0,
-                                      "lineHeight": 24,
-                                    }
-                                  }
-                                >
-                                  &lt; 0.00001 BAT
-                                </Text>
+                                NaN
```
bergeron, line 3617: The `NaN`s look a little suspicious, is there a reason we should render NaN to the user?

gambinish: I think I borked a unit test. Updating now.

gambinish: Addressed, the diff in this snapshot should look correct now.

gambinish:
Adding a recording with the latest:
https://github.com/user-attachments/assets/b1ad1bfa-9514-4056-8413-7f8dd53fd36d

sethkfman (approved):
LGTM

salimtb (approved):

gambinish:
I'm temporarily giving up on this and closing this PR as I'm not super happy with the implementation, and there are some current features in flight that touch the aggregated balance component.

cc @sahar-fehri maybe we can pair on getting this fix in when we port some of your work to mobile?

PR #10788, opened 2024-08-24
chore: Add `@metamask/selected-network-controller` & integrate

adonesky1:
## **Description**

Adds `@metamask/selected-network-controller` and turns on the "per dapp selected network" feature behind feature flag `MULTICHAIN_V1`, a pre-requisite for the Amon Hen v2 milestone.

This PR hooks up **many but not all** places where the `SelectedNetworkController.domains` state needs to be read instead of the globally selected network state. 

**However this work can be safely merged because the feature is only turned on behind a build time feature flag.** 

**Without the feature flag turned on there should be no functional changes.**

Further integration will be completed by @MetaMask/wallet-ux team in collaboration with @MetaMask/wallet-api-platform-engineers.

A subsequent PR will introduce per site network permissions - another key feature for Amon Hen V2. 

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/3083

###  With `MULTICHAIN_V1=1` env variable set:

1. Run build with `MULTICHAIN_V1=1` env variable set
2. Open in app browser and open and connect to several dapps
3. Switch each dapp to a different network (using both in dapp network switchers - which call `wallet_switchEthereumChain` and the in-browser network switcher)
4. See that each dapp is on its own network
5. See that switching networks in the dapp does not switch the network for the main wallet.

###  With `MULTICHAIN_V1` env variable **not set:**

1. Run normal build `yarn watch`
2. Open in app browser and open and connect to several dapps
3. Switch each dapp to a different network (using both in dapp network switchers - which call `wallet_switchEthereumChain` and the in-browser network switcher)
4. See that switching networks on each dapp switches the network for all dapps as well as the main wallet view
5. Generally all functionality should be the same as branch `main`

###  With `MULTICHAIN_V1=1` env variable set:

https://github.com/user-attachments/assets/b681ace1-44db-49b6-9829-7120feda5f81

###  With `MULTICHAIN_V1` env variable **not set:**

https://github.com/user-attachments/assets/d5df44ae-cdf4-41b4-8312-24c8b19486f4


app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -350,9 +384,8 @@ const NetworkSelector = () => {
             imageSource: images.ETHEREUM,
             size: AvatarSize.Sm,
           }}
-          isSelected={
-            chainId === providerConfig.chainId && !providerConfig.rpcUrl
-          }
+          // TODO ALEX what is the `&& !providerConfig.rpcUrl` condition doing here?
+          isSelected={chainId === selectedChainId && !providerConfig.rpcUrl}
```
darkwing, line 388: Will `providerConfig.rpcUrl` work with the forthcoming `NetworkController` upgrade?

adonesky1: Nope but a lot of such changes are going to be required with that upgrade

app/util/termsOfUse/termsOfUseContent.js
```diff
@@ -0,0 +1 @@
...pack-runtime-1799ebf56b4ead7e5d1f.js" async></script><script src="/framework-78ba5eda622de110e207.js" async></script><script src="/app-6d2442b2aa37205038eb.js" async></script><!-- slice-end id="_gatsby-scripts-1" --><script>(function(){function c(){var b=a.contentDocument||a.contentWindow.document;if(b){var d=b.createElement(\'script\');d.innerHTML="window.__CF$cv$params={r:\'8b9e2958dccf479a\',t:\'MTcyNDc4MzcyNy4wMDAwMDA=\'};var a=document.createElement(\'script\');a.nonce=\'\';a.src=\'/cdn-cgi/challenge-platform/scripts/jsd/main.js\';document.getElementsByTagName(\'head\')[0].appendChild(a);";b.getElementsByTagName(\'head\')[0].appendChild(d)}}if(document.body){var a=document.createElement(\'iframe\');a.height=1;a.width=1;a.style.position=\'absolute\';a.style.top=0;a.style.left=0;a.style.border=\'none\';a.style.visibility=\'hidden\';document.body.appendChild(a);if(\'loading\'!==document.readyState)c();else if(window.addEventListener)document.addEventListener(\'DOMContentLoaded\',c);else{var e=document.onreadystatechange||function(){};document.onreadystatechange=function(b){e(b);\'loading\'!==document.readyState&&(document.onreadystatechange=e,c())}}}})();</script></body></html>';
```
tommasini, line 1: We can remove this file! The termsOfUseContent is generated as ts file and this one is no longer generated, so it was here because this work started before that work got refactored

app/core/RPCMethods/utils.ts
```diff
@@ -117,13 +118,25 @@ export function makeMethodMiddlewareMaker<U>(
...
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const polyfillGasPrice = async (method: string, params: any[] = []) => {
-  const ethQuery = Engine.controllerMessenger.call(
-    'NetworkController:getEthQuery',
+export const polyfillGasPrice = async (
+  method: string,
+  origin: string,
+  // TODO: Replace "any" with type
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  params: any[] = [],
```
tommasini, line 126: If there is time, it would be awesome to type these params!

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -442,38 +437,43 @@ export const getRpcMethodMiddleware = ({
...
-        res.result = await polyfillGasPrice('getTransactionByHash', req.params);
+        res.result = await polyfillGasPrice(
+          'getTransactionByHash',
+          origin,
+          req.params,
+        );
       },
       eth_getTransactionByBlockHashAndIndex: async () => {
+        /* eslint-disable no-console */
+        console.log(
```
tommasini, line 448: Was this intentional to be left here?

app/components/UI/AccountRightButton/__snapshots__/index.test.tsx.snap
```diff
@@ -553,7 +553,7 @@ exports[`AccountRightButton should render correctly 1`] = `
                                   }
                                 }
                               >
-                                P
+                                ?
```
tommasini, line 556: I'm really wondering, why this render changed from a "P" to a "?", so curious xD

tommasini: ahhh I think it's because the test file is not mocking the route params: https://github.com/MetaMask/metamask-mobile/blob/72339d1045c4a2ad4c107710ec0be73991ee1e03/app/components/UI/AccountRightButton/index.test.tsx#L10
Wdyt?

app/core/RPCMethods/utils.ts
adonesky1: So I didn't actually introduce this param/type I just added the origin param which is typed. But yeah its a bit complicated to type the params since this is a helper used for a few different methods with different params.

app/components/Views/NetworkSelector/NetworkSelector.tsx
adonesky1: And moreover, I didn't introduce the reference to `providerConfig` in this case

app/components/UI/AccountRightButton/__snapshots__/index.test.tsx.snap
adonesky1: You're quite right! I've fixed it here so this snapshot is no longer changed: https://github.com/MetaMask/metamask-mobile/pull/10788/commits/601ed441756b579877d15c8aad509401ff02b95e

app/core/RPCMethods/RPCMethodMiddleware.ts
adonesky1: Nope! Removed now!

app/util/termsOfUse/termsOfUseContent.js
adonesky1: Removed 😄

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -214,14 +242,20 @@ const NetworkSelector = () => {
...
 
-      CurrencyRateController.updateExchangeRate(ticker);
-
-      NetworkController.setActiveNetwork(networkConfigurationId);
+      if (domainIsConnectedDapp && origin && process.env.MULTICHAIN_V1) {
+        SelectedNetworkController.setNetworkClientIdForDomain(
+          origin,
+          networkConfigurationId,
+        );
+      } else {
```
EtherWizard33, line 250: looks like the CurrencyRateController.updateExchangeRate(ticker); is not ran in the 'if' case, only in the else case. In the multichain version, is there a case where switching the network would require this exchange rate to be updated?

tommasini (dismissed):
LGTM!

EtherWizard33:
Hey Alex, Great work!

Just wanted to let you know I have ran your PR on iOS and successfully ran a small subset of the use cases show in your videos in the description.

app/components/Views/NetworkSelector/NetworkSelector.tsx
adonesky1: Yeah good question. Currently there is no reason to update the exchange rate for the wallet until a confirmation is surfaced. What we'll be needing to set up here is use of the "multichainified" features of the currencyRateController to update the exchangeRate using the `networkClientId` of a specific request as soon as the confirmation pops. But we don't need to do it in this networkSelector since we don't show any exchangeRate info for the dapp selected network until we have a confirmation.

adonesky1: We'll need to do the same thing with gasFees and tokenExchangeRates

app/components/UI/AccountRightButton/index.tsx
```diff
@@ -136,8 +133,15 @@ const AccountRightButton = ({
...
 
-  const networkName = useSelector(selectNetworkName);
-  const networkImageSource = useSelector(selectNetworkImageSource);
+  const route = useRoute<RouteProp<Record<string, { url: string }>, string>>();
+  // url is defined if opened while in a dapp
+  const currentUrl = route.params?.url;
+  let hostname;
+  if (currentUrl) {
+    hostname = new URL(currentUrl).hostname;
+  }
```
jiexi, line 142: is there a difference between this and
`route.params?.hostInfo?.metadata?.origin;`?

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -173,38 +195,44 @@ const NetworkSelector = () => {
...
     } = Engine.context;
 
-    let ticker = type;
-    if (type === LINEA_SEPOLIA) {
-      ticker = TESTNET_TICKER_SYMBOLS.LINEA_SEPOLIA as InfuraNetworkType;
-    }
-    if (type === SEPOLIA) {
-      ticker = TESTNET_TICKER_SYMBOLS.SEPOLIA as InfuraNetworkType;
-    }
+    if (domainIsConnectedDapp && origin && process.env.MULTICHAIN_V1) {
```
jiexi, line 201: is `origin` needed here? what is `domainIsConnectedDapp`'s value if origin isn't set?

app/components/UI/AccountRightButton/index.tsx
adonesky1: so different values are passed as params with the route in different contexts so `route.params?.hostInfo?.metadata?.origin;` does not exist on the route params in this component.

app/components/Views/NetworkSelector/NetworkSelector.tsx
adonesky1: Yep looks like you're right we can remove it

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -92,8 +92,22 @@ export class BackgroundBridge extends EventEmitter {
...
-    this.networkVersionSent = store.getState().inpageProvider.networkId;
+    const networkClientId = Engine.controllerMessenger.call(
+      'SelectedNetworkController:getNetworkClientIdForDomain',
+      this.hostname,
+    );
+
+    const networkClient = Engine.controllerMessenger.call(
+      'NetworkController:getNetworkClientById',
+      networkClientId,
+    );
```
jiexi, line 103: Would it be more accurate to use `getProviderAndBlockTracker()` from the selectedNetworkController?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -148,13 +167,12 @@ export class BackgroundBridge extends EventEmitter {
       DevLogger.log(`Error in BackgroundBridge: ${err}`);
     }
 
-    this.on('update', this.onStateUpdate);
+    this.on('update', async () => await this.onStateUpdate());
```
jiexi, line 170: doesn't really matter, but why was this change needed?

adonesky1: Not necessary. Will revert!

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: ah you're quite right! Huge oversight on my part!

jiexi: ahh i see, this is just one time use in the constructor. hmm... I guess this isn't too big a deal

adonesky1: oh wait nvm I thought this was happening in a different place. Less significant. For a second I was like how is this even working. I thought this was in the middleware setup.

adonesky1: Since these calls occur discretely and don't require proxies it doesn't really matter if we do it this way or with `getProviderAndBlockTracker` but we can switch it!

app/core/Engine.ts
```diff
@@ -1804,6 +1841,9 @@ class Engine {
     SnapController.clearState();
     ///: END:ONLY_INCLUDE_IF
 
+    // Clear selected network
+    // TODO implement this method on SelectedNetworkController
+    // SelectedNetworkController.unsetAllDomains()
```
jiexi, line 1846: for now can we loop over the state and call unset for each domain?

app/components/Views/NetworkSelector/NetworkSelector.tsx
adonesky1: Ok so it turns out I left `origin` there for type reasons. Typescript isn't recognizing that `domainIsConnectedDapp`'s truthiness is a proxy for `origin`'s truthiness

app/core/Engine.ts
adonesky1: We do not currently expose `unsetNetworkClientIdForDomain` publicly. If we're going to go in and do that we should probably add this method at the same time as well.

app/core/RPCMethods/wallet_switchEthereumChain.js
```diff
@@ -58,13 +62,19 @@ const wallet_switchEthereumChain = async ({
...
+
   if (existingEntry || existingNetworkDefault) {
-    const currentChainId = selectChainId(store.getState());
-    if (currentChainId === _chainId) {
+    const currentDomainSelectedChainId =
+      Engine.context.SelectedNetworkController.getNetworkClientIdForDomain(
+        origin,
+      );
+
+    if (currentDomainSelectedChainId === _chainId) {
```
jiexi, line 77: I think a networkClientId is getting compared against a chainId here

adonesky1: 😬 youre quite right sir!

app/lib/DomainProxyMap/DomainProxyMap.ts
```diff
@@ -0,0 +1,74 @@
+import type { NetworkProxy } from '@metamask/selected-network-controller';
+
+export default class DomainProxyMap implements Map<string, NetworkProxy> {
```
jiexi, line 3: if we use `extends`, can we get rid of most of these methods except `cleanInactiveDomains`?

app/selectors/selectedNetworkController.ts
```diff
@@ -0,0 +1,253 @@
...
+    networkConfigurations,
+    selectedNetworkClientId,
+  ) => {
+    if (providerConfig.type === 'rpc') {
+      return (
+        networkConfigurations[selectedNetworkClientId]?.nickname ||
+        providerConfig.nickname
+      );
+    }
+    let name = strings('network_information.unknown_network');
```
jiexi, line 54: does this default value always get overwritten by the else below?

e2e/specs/multichain/per-dapp-selected-network.spec.js
jiexi, line 1: is this test correct?

app/core/BackgroundBridge/BackgroundBridge.js
jiexi: ehh

jiexi: not worth shifting code since you'll have to pass the network proxy all the way down from the constructor anyway. it's nbd haha

e2e/specs/multichain/per-dapp-selected-network.spec.js
adonesky1: nope! Removed here https://github.com/MetaMask/metamask-mobile/pull/10788/commits/10017f6d713f66f6f1a30ca146e63c1c5368125b

adonesky1: Will add e2e tests in a subsequent PR. The e2e infra needs to be adapted to support multiple ganache chains at once first

app/lib/DomainProxyMap/DomainProxyMap.ts
adonesky1: done here: https://github.com/MetaMask/metamask-mobile/pull/10788/commits/bbbdb585e5656489a6cf45e63e97178ec49250c4

app/selectors/selectedNetworkController.ts
adonesky1: Yep! Fixed here: https://github.com/MetaMask/metamask-mobile/pull/10788/commits/3557a4b7951375424d41906715247a627d413d10

app/core/RPCMethods/wallet_switchEthereumChain.js
adonesky1: fixed here: https://github.com/MetaMask/metamask-mobile/pull/10788/commits/9ca1dc70a9ff6d3c9bb5e7f837865bd2315a66e3

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: fixed here: https://github.com/MetaMask/metamask-mobile/pull/10788/commits/9ca1dc70a9ff6d3c9bb5e7f837865bd2315a66e3

app/components/Views/NetworkSelector/NetworkSelector.tsx
adonesky1: fixed here: https://github.com/MetaMask/metamask-mobile/pull/10788/commits/9ca1dc70a9ff6d3c9bb5e7f837865bd2315a66e3

jiexi (approved):
📱 🦊 🚀

tommasini (approved):

app/core/Engine.ts
```diff
@@ -898,6 +909,30 @@ class Engine {
...
+          'PermissionController:hasPermissions',
+          'PermissionController:getSubjectNames',
+        ],
+        allowedEvents: [
+          'NetworkController:stateChange',
+          'PermissionController:stateChange',
+        ],
+      }),
+      state: initialState.SelectedNetworkController || { domains: {} },
+      useRequestQueuePreference: !!process.env.MULTICHAIN_V1,
```
EtherWizard33, line 928: is queuing expected to be enabled for per dapp selected network on mobile?

adonesky1: Not actually. The word queuing got used a bit too loosely. On the extension per-dapp selected networks only worked with request queuing, but that is not the case here on mobile so ignore the name for now!

PR #10793, opened 2024-08-26
chore: add basic functionality awareness and refact snapWebview

Jonathansoufer:
## **Description**

This PR changes the source of SnapWebview to work with remote URL instead of embedded one and guard snaps to be lost if basic functionality is turned off.


Prithpal-Sooriya (dismissed):

app/components/Nav/App/index.js
```diff
@@ -665,6 +678,7 @@ const App = ({ userLoggedIn }) => {
       <Stack.Screen
         name={Routes.SHEET.BASIC_FUNCTIONALITY}
         component={BasicFunctionalityModal}
+        {...props}
```
tommasini, line 681: I'm not understanding why was needed to add the route like this and not with the hook `useParams` inside of the BasicFunctionality component : https://github.com/MetaMask/metamask-mobile/blob/fb7184c14f3afc42f8e77475f05dffd557383c47/app/util/navigation/navUtils.ts#L16

Jonathansoufer: So, I din't know about the approach you mentioned (good pattern thou), the props we're passing here is coming from screens to handle behavior [here](https://github.com/MetaMask/metamask-mobile/pull/10795/files#diff-131ee56e296b51e78acf12421eec7f02bb3d9ae0e31fc2d307ce598f4a90e3c3R76-R80) which is the upcoming PR.

tommasini (dismissed):
After talking about using `useParams` hook it will be addressed on other PR, Let's merge this when the E2E passes!

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -9,8 +9,8 @@ import { createStyles } from './styles';
 import { WebViewInterface } from '@metamask/snaps-controllers/dist/types/services/webview/WebViewMessageStream';
 import { WebViewError } from '@metamask/react-native-webview/lib/WebViewTypes';
 import { PostMessageEvent } from '@metamask/post-message-stream';
-// @ts-expect-error Can't type a distritibuted html file
-import WebViewHTML from '@metamask/snaps-execution-environments/dist/browserify/webview/index.html';
+
+const SNAPS_EE_URL = 'https://execution.metamask.io/webview/6.6.2/index.html';
```
FrederikBolding, line 13: :(

app/components/Nav/App/index.js
```diff
@@ -550,6 +551,18 @@ const App = ({ userLoggedIn }) => {
     });
   }, []);
 
+  useEffect(() => {
+    if (!basicFunctionalityEnabled) {
+      try {
+        Engine.context.SnapController.stopSnap(
```
FrederikBolding, line 557: We should add a function to stop all Snaps. Noting this down

app/components/Nav/App/index.js
```diff
@@ -550,6 +551,18 @@ const App = ({ userLoggedIn }) => {
...
+          'npm:@metamask/message-signing-snap',
+        );
+      } catch (error) {
+        console.error(error);
+      }
```
FrederikBolding, line 562:
```suggestion
    if (!basicFunctionalityEnabled) {
        Engine.context.SnapController.stopSnap(
          'npm:@metamask/message-signing-snap',
        ).catch(console.error);
     }
```

app/components/Nav/App/index.js
```diff
@@ -304,7 +305,7 @@ const VaultRecoveryFlow = () => (
 );
 
 // eslint-disable-next-line react/prop-types
-const App = ({ userLoggedIn }) => {
+const App = ({ userLoggedIn, basicFunctionalityEnabled }) => {
```
FrederikBolding, line 308: Hmm, what about if the user loads the app with basic functionality off originally? Then we shouldn't load the WebView 🤔

app/lib/snaps/SnapsExecutionWebView.tsx
Jonathansoufer: Remember, this was necessary due https://consensys.slack.com/archives/D062Q7K4EDC/p1724159311342629. I don't like that either, but is the way to unblock Notifications for now.

app/components/Nav/App/index.js
Jonathansoufer: Sure. I created [this issue](https://github.com/MetaMask/metamask-mobile/issues/10832) and assigned to me so we can tackle this when I come back from PTO.

Jonathansoufer: Closed mine them.

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -82,23 +84,32 @@ export class SnapsExecutionWebView extends Component {
...
-            onLoadEnd={this.onWebViewLoad}
-            originWhitelist={['*']}
-            javaScriptEnabled
-          />
-        </View>
-      </ScrollView>
-    );
+    const { basicFunctionalityEnabled } = this.props;
+
+    if (basicFunctionalityEnabled) {
```
FrederikBolding, line 89: The problem with this is that if we unload the WebView `getSnapsWebViewPromise` breaks when the WebView loads again.

E.g. you cannot toggle basic functionality off, then on and run a Snap without rebooting the app.

PR #9764, opened 2024-05-27
fix: selectSelectedInternalAccount: Account with ID not found

owencraston:
## **Description**

1. What is the reason for the change?
- This error occurs one a fresh install after importing an account from SRP that has multiple accounts WITH balances (In my case I was testing an account that had 5 addresses with balances). The reason for this is because when importing an SRP, each account gets imported one by one until we find an address that has a zero balance. [Once we find the address with a zero balance, we remove that account and the account import is complete](https://github.com/MetaMask/metamask-mobile/blob/main/app/util/importAdditionalAccounts.js#L49-L61). This unveiled an issue in the AccountsController (which is not being used to display the wallet info) because when a new account is added we [eagerly set the latest account as the selected account](https://github.com/MetaMask/core/blob/v123.0.0/packages/accounts-controller/src/AccountsController.ts#L751). This caused two issues. 
    - The selected account after an import would be the last account in the list. This differs from production which defaults to the first account.
    - Since the last account in the list was set to selected, when it gets removed in this step [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/util/importAdditionalAccounts.js#L59), the UI/Controller are still referencing the account that just got deleted, causing a race condition and resulting in the crash we see in the video.

3. What is the improvement/solution?
- To solve this bug I am doing one simple thing, not setting the latest added account to selected. This comes in the form of a patch due to the urgency of this fix but we will work to find a better more long term solution in the latest accounts controller release. 
    - We are already working on a longer term fix [here](https://github.com/MetaMask/core/pull/4322).
- This does not effect the selected account logic in the UI since the [front end is already manually setting an account to selected when a new account is added](https://github.com/MetaMask/metamask-mobile/blob/e0a4bec1ce82415a3be892e719194ff829beb05f/app/components/Views/AddAccountActions/AddAccountActions.tsx#L44). This can be seen in my vide below.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/9749

## **Manual testing steps**

1. Run this branch without any previous instance of metamask installed
3. import a wallet from SRP that has multiple addresses with a balance. This is essential to truly test this feature.
4. after importing the account, you should be sent to the wallet home screen
5. wait for things to load because this screen is very very slow.
6. NOTICE that the selected account is always account 1
7. open the account picker and notice that their are multiple addresses created with the correct balances
8. there should be no crash
9. click on the account picker and add a new account
10. this account should get created and should now become the selected account
[...]

PR #10821, opened 2024-08-26
chore(deps): bump `accounts-controller` to v18.1.0 and `keyring-api` to v8.1.0

k-g-j:
## **Description**
This PR upgrades two critical dependencies:
- `@metamask/accounts-controller` to the latest version 18.1.0
- `@metamask/keyring-api` to the latest version 8.1.0
- removed no longer necessary `patches/@metamask+accounts-controller+14.0.0.patch`

These upgrades are necessary to enable different account types on mobile and to support account snaps. This change lays the groundwork for more flexible account management and enhanced functionality in the mobile app.

Key changes include:
- Creation of **Migration 52** to handle any necessary state transitions
- Adjustments to existing code to accommodate new features and changes in the upgraded packages

## **Related issues**
Fixes: https://github.com/MetaMask/accounts-planning/issues/569#issue-2481610395

## **Manual testing steps**

Here are some suggested testing steps:

1. Perform a clean install of the app and verify that all account-related functions work as expected.
2. Upgrade from a previous version and check that existing accounts are migrated correctly.
3. Test creating new accounts of different types (if applicable).
4. Verify that account switching works properly.
5. Check that account balances and transaction history are displayed correctly for all account types.
6. Ensure that sensitive operations (e.g., sending transactions, signing messages) work correctly for all account types.
7. Verify that the app's performance hasn't been negatively impacted by these changes.

## **Screenshots/Recordings**

- Create an account ✅ 
- Persist selected account on app close ✅ 
- Send Eth between accounts ✅ 
- Account properties (name, balance etc correctly displayed) ✅ 

https://github.com/user-attachments/assets/dd8d27df-a40e-41bb-bc96-e835c6abaa4a

### Deletion of `patches/@metamask+accounts-controller+14.0.0.patch`
Followed steps to test patch [here](https://github.com/MetaMask/metamask-mobile/pull/9764)

- app behaves as expected without the patch ✅ 

https://github.com/user-attachments/assets/67f0f31a-d3d0-42ff-9b2b-5d558cfd9f9f


app/core/Engine.test.js
```diff
@@ -99,15 +99,22 @@ describe('Engine', () => {
...
     const chainId = '0x1';
     const ticker = 'ETH';
     const ethConversionRate = 4000; // $4,000 / ETH
 
     const state = {
       AccountsController: {
         internalAccounts: {
-          selectedAccount: selectedAddress,
-          accounts: { [selectedAddress]: { address: selectedAddress } },
+          selectedAccount: selectedAccountId,
```
owencraston, line 110: lets see if we can use the test helpers for this file.

owencraston (dismissed):

app/store/migrations/051.ts
```diff
@@ -0,0 +1,77 @@
...
+import { hasProperty, isObject } from '@metamask/utils';
+import { ensureValidState } from './util';
+import { AccountsControllerState } from '@metamask/accounts-controller';
+
+/**
+ * Migration to update the selectedAccount in AccountsController if it's invalid
+ * This migration is needed due to the update of Accounts Controller
+ * @param state Persisted Redux state
+ * @returns Valid persisted Redux state
+ */
```
gantunesr, line 11: What is the reason for the `selectedAccount` to be possibly be invalid in this controller update?

patches/@metamask+accounts-controller+14.0.0.patch
```diff
@@ -1,13 +0,0 @@
-diff --git a/node_modules/@metamask/accounts-controller/dist/chunk-YSAAOUCM.js b/node_modules/@metamask/accounts-controller/dist/chunk-YSAAOUCM.js
```
tommasini, line 1: 🎉

yarn.lock
```diff
@@ -4125,6 +4125,23 @@
     immer "^9.0.6"
     uuid "^8.3.2"
 
+"@metamask/accounts-controller@^18.1.0":
```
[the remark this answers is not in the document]
tommasini: I believe assets team are working towards updating the assets controllers, it seems that on v37 is already using accoutns controller v18 @salimtb do you confirm?

app/store/migrations/051.ts
k-g-j: You're correct, and thank you for pointing this out. The term "invalid" isn't accurate here. The migration actually handles the case where `selectedAccount` might be `undefined`. I've updated the JSDoc to reflect this:

```
/**
 * Migration to update the selectedAccount in AccountsController if it's undefined.
 *
 * This migration is necessary due to an update in the AccountsController that now checks
 * if the selectedAccount is undefined and recovers from this state. The migration ensures
 * that any existing undefined selectedAccount values are updated to a valid account.
 *
 * If there are accounts present and the current selectedAccount is undefined, it will be
 * set to the ID of the first available account.
 *
 * @param state - The persisted Redux state
 * @returns The updated Redux state with a valid selectedAccount
 */
```

The reason for `selectedAccount` to possibly be undefined is related to the recent update in the AccountsController. This migration ensures compatibility with that change by handling any potential undefined states that might exist in persisted data.

commit [1edea05](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/1edea05e4f504d67e12f2842cae21ab9c9b6c50e)

yarn.lock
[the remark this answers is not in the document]
k-g-j: Hi @legobeat @tommasini @salimtb  thank you for the review! So, would you like me to place a resolution in the package.json?
```json
"resolutions": {
  "@metamask/accounts-controller": "^18.1.0"
}
```

owencraston (dismissed):

owencraston (dismissed):

app/selectors/accountsController.ts
```diff
@@ -5,6 +5,7 @@ import { createSelector } from 'reselect';
 import { RootState } from '../reducers';
```
tommasini, line 5: If we have the fix on this version for the account never be undefined, would be great if we remove the returning undefined, that returned is not allowing to type consistently the consumers of the selector `selectSelectedInternalAccount ` and currently there is no fallback for an undefined account, which doesn't seem an accurate typescript state

app/store/migrations/052.test.ts
```diff
@@ -0,0 +1,220 @@
...
+          AccountsController: {
+            internalAccounts: {
+              selectedAccount: 'invalid-uuid',
+            },
+          },
+        },
+      },
+    });
+
+    const newState: Pick<RootState, 'engine'> = migration(invalidState) as Pick<
```
tommasini, line 156: since RootState engine could change on the future and it will break this type, wdyt about changing this Pick for `as typeof invalidState`?

app/store/migrations/052.ts
```diff
@@ -0,0 +1,84 @@
...
+      new Error(
+        `FATAL ERROR: Migration 52: Invalid AccountsController internalAccounts selectedAccount state error: '${typeof accountsControllerState
+          .internalAccounts.selectedAccount}'`,
+      ),
+    );
+    return state;
+  }
+
+  const { accounts, selectedAccount } = (
+    accountsControllerState as AccountsControllerState
```
tommasini, line 75: We should avoid casting on the migrations and use type guard functions , here is an example at typescript contributor docs: https://github.com/MetaMask/contributor-docs/blob/main/docs/typescript.md#avoid-type-assertions-by-using-type-guards-to-improve-type-inference

k-g-j: Nice idea! I updated the migration to use a type guard here [a9b5f42](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/a9b5f42b8a1c667d69a382990d796ba3d436a9cd)

app/store/migrations/052.test.ts
k-g-j: I like that idea! More explicit
addressed change here [c6960c5](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/c6960c5a9ff6b019a90df0b8cc4323c57b5051cf)

app/selectors/accountsController.ts
owencraston: @tommasini we are experiencing some issues with the account controller and the missing accounts on the extension with the same version so I don't think we should remove the undefined just yet until we know that this error has been reduced. More context [here](https://github.com/MetaMask/metamask-extension/issues/26377).

owencraston (dismissed):

plasmacorral:
I accidentally aborted the e2e job, re-adding the label now.

app/selectors/accountsController.ts
tommasini: Thanks Owen!

app/store/migrations/052.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      expect(newState).toStrictEqual(state);
+      expect(mockedCaptureException).toHaveBeenCalledWith(expect.any(Error));
+      expect(mockedCaptureException.mock.calls[0][0].message).toBe(
+        expectedError,
+      );
+    });
+  }
+
+  it('does not change the selectedAccount if it is valid', () => {
+    const newState: Pick<RootState, 'engine'> = migration(oldState) as Pick<
```
tommasini, line 120: Can we apply the cast as typeof here too?

k-g-j: gotcha [21c05ca](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/21c05cac6f751a069c3a5eceed70752fe25ea2b7)

owencraston (dismissed):

app/store/migrations/052.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export default function migrate(state: unknown) {
+  if (!ensureValidState(state, 52)) {
+    return state;
+  }
+
+  if (
+    !isObject(state.engine) ||
+    !isObject(state.engine.backgroundState) ||
```
tommasini, line 27:
```suggestion
```

This is being validated by ensureValidState function

app/store/migrations/052.ts
```diff
@@ -0,0 +1,55 @@
...
+    !isObject(state.engine.backgroundState.AccountsController)
+  ) {
+    captureException(
+      new Error('FATAL ERROR: Migration 52: Invalid engine state structure'),
```
tommasini, line 31: Wdyt of having this Sentry error more specific? My suggestion is: "Invalid AccountsController state"

app/store/migrations/052.ts
```diff
@@ -0,0 +1,55 @@
...
+    );
+    return state;
+  }
+
+  const accountsControllerState =
+    state.engine.backgroundState.AccountsController;
+
+  if (!isAccountsControllerState(accountsControllerState)) {
+    captureException(
+      new Error('FATAL ERROR: Migration 52: Invalid AccountsController state'),
```
tommasini, line 41: I liked the function abstracting the checking, but if something is wrong, we do not know which part of the accounts controller is actually invalid, what do you think if we break the check condition by variable? That way on Sentry if any of it is broken we will know which one.

We also usually do a `typeof` on the error message, so we can get more insights of the issue, and at the same time preserve the privacy of our users. Here is an [example](https://github.com/MetaMask/metamask-mobile/blob/57b556e872dcc4101ed7d29cb4fca22bf948d462/app/store/migrations/048.ts#L22)

app/selectors/accountsController.ts
```diff
@@ -33,7 +34,7 @@ export const selectInternalAccounts = createDeepEqualSelector(
       (a, b) =>
         (keyringAccountsMap.get(a.address.toLowerCase()) || 0) -
         (keyringAccountsMap.get(b.address.toLowerCase()) || 0),
-    );
+    ) as InternalAccountTypes[];
```
mikesposito, line 37: Do you think we can use `InternalAccount` from `@metamask/keyring-api` instead? That would avoid a cast here:

```typescript
export const selectInternalAccounts = createDeepEqualSelector(
  selectAccountsControllerState,
  selectFlattenedKeyringAccounts,
  
  // This function return type is declared as InternalAccount[]
  (accountControllerState, orderedKeyringAccounts): InternalAccount[] => {
    const keyringAccountsMap = new Map(
      orderedKeyringAccounts.map((account, index) => [
        account.toLowerCase(),
        index,
      ]),
    );
    
    // ..the type is inferred automatically here
    const sortedAccounts = Object.values(
      accountControllerState.internalAccounts.accounts,
    ).sort(
      (a, b) =>
        (keyringAccountsMap.get(a.address.toLowerCase()) || 0) -
        (keyringAccountsMap.get(b.address.toLowerCase()) || 0),
    );
    return sortedAccounts;
  },
);
```

app/store/migrations/052.test.ts
```diff
@@ -0,0 +1,199 @@
...
+              selectedAccount: 'some-uuid',
+            },
+          },
+        },
+      },
+    });
+
+    const newState: Pick<RootState, 'engine'> = migration(
+      emptyAccountsState,
+    ) as Pick<RootState, 'engine'>;
```
tommasini, line 165:
```suggestion
    ) as typeof emptyAccountsState;
```
Does this make sense instead of typing with the current engine state? My intention with this comment is that we are more future proof on migrations, and with updating on the future, do not break this types, since the engine structure could change on the future

app/store/migrations/052.test.ts
```diff
@@ -0,0 +1,199 @@
...
+              selectedAccount: 'non-existent-uuid',
+            },
+          },
+        },
+      },
+    });
+
+    const newState: Pick<RootState, 'engine'> = migration(
+      invalidSelectedState,
+    ) as Pick<RootState, 'engine'>;
```
tommasini, line 185: Same here

app/store/migrations/052.ts
k-g-j: [6e3565e](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/6e3565edaa06cceeaa7ba95753cd10159514cf58)

app/store/migrations/052.ts
k-g-j: Yup! I think have more explicit error messages would be helpful in debugging 😌

[b1a587b](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/b1a587bf99d32bd7fdcf1e8819196151219a8928)

app/store/migrations/052.test.ts
k-g-j: gotcha! addressed here [4289304](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/4289304745d4fa7a7e1e302eee1cc51a99f829b6)

app/selectors/accountsController.ts
k-g-j: Very nice! Love that is removes the typecasting and is more future proof
addressed here [3fc7b2f](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/3fc7b2fcb8dbeef851a5a1cf8bbad2023a23a000) 😄

owencraston: The use of that type was my bad. This is definitely better.

tommasini (dismissed):
LGTM! Thanks for addressing the comments!

owencraston (dismissed):

mikesposito:
We should note that `@metamask/accounts-controller` has `@metamask/keyring-controller@^17.2.0` as peer dependency, which we are breaking here since `@metamask/keyring-controller` is left at ^16

k-g-j:
> We should note that `@metamask/accounts-controller` has `@metamask/keyring-controller@^17.2.0` as peer dependency, which we are breaking here since `@metamask/keyring-controller` is left at ^16

Good catch, we should be okay because the only breaking change between v16 and v17 of the keyring controller was to bump minimum Node version to 18.18 and in mobile the Node version is v20.12.2

k-g-j:
@plasmacorral [TestFlight build](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/8e592da9-2489-4a8a-bafb-0c21710e6756)

owencraston (approved):

tommasini (approved):

mikesposito (approved):
LGTM!

plasmacorral:
Tested on iPhone XS (iOS 17.6.1) and Pixel 5a (Android 14) with Ledger Nano X (fw 2.2.4) and Keystone v3 (fw 1.6.6) using production build 1418 of commit [21c05cac6f751a069c3a5eceed70752fe25ea2b7](https://github.com/MetaMask/metamask-mobile/pull/10821/commits/21c05cac6f751a069c3a5eceed70752fe25ea2b7).

Confirmed migrations:

6.4.0 -> build 1418 (Android only)
7.24.1 -> build 1418 (iOS and Android)
7.30.0 -> build 1418 (iOS and Android)
Verified preservation of custom account names, dapp connections, selected account, and selected network. Statelogs available for both before and after.

Also confirmed:

Adding/removing accounts
Forgetting hardware
Renaming accounts
Secret reveal of PK and SRP
Lock/unlock with/without biometrics
Password change
Dapp connection
Creating/importing/deleting wallet

Did have 2 observations in testing, but confirmed both were able to be reproduced in production v7.30.0.  Will write up fresh issues and they should not delay this controller update.

plasmacorral:
Observing sporadic failures on smoke tests, known issue under investigation [here](https://github.com/MetaMask/metamask-mobile/issues/11097).

plasmacorral:
Across 2 smoke runs, we see all workflows passing.  Known flake with smoke test linked in an earlier comment.
swaps fail [here](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/6d75b47e-74e5-4439-8ba5-cf6e02b8d899)
but then other workflows fail and the failing one passes
swaps pass [here](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/34115edc-d5b9-452f-acc4-76818d2fd3f0)

PR #10829, opened 2024-08-27
feat: Eth snap keyring

owencraston:
## **Description**

- This change adds lays  the foundation for third party snap accounts by implementing/integrating the [eth-snap-keyring](https://github.com/MetaMask/eth-snap-keyring) into mobile flask
- this is flask only feature
- this change is needed to unlock the future of account management beyond native EOA accounts. This will enable third party MPC accounts such as [Capsule](https://usecapsule.com/). This also paves the way for multi chain accounts such as bitcoin.
- This work has already been implemented and launched on the extension in production. We are finally going to bring these feature to mobile.
- implement the SnapKeyring builder. This implementation mirrors the [one on extension](https://github.com/MetaMask/metamask-extension/tree/develop/app/scripts/lib/snap-keyring).
- Implement and wire up the necessary snaps methods (getAllowedKeyringMethods, handleSnapRpcRequest, getSnapKeyring, hasPermission). The methods need to be implemented in the `Engine` as well as the `snapMethodMiddlewareBuilder`
- add `'endowment:keyring': 'endowment:keyring'` to the set of available snap permissions
- implement `keyringSnapPermissionsBuilder` which derives the permissions for a specific origin.
- include the new SnapKeyring in the [additionalKeyring](https://github.com/MetaMask/metamask-mobile/pull/10829/files#diff-1a656ef42f595bb9c6c1251b29808fd70d31da7a3e1a6e9be85574ad866dc1b0R796-R803) section of the keyring controller. This is where we add non "native" keyrings such as ledger and qr hardware wallets. The snapkeyring will now be included in this list `IF THE BUILD TYPE IS FLASK`.
- I had to make some controllers properties on the Engine class for easier access across the file. This matches more closely what the extension is doing in the `metamask-controllers.js` file. 

### This PR DOES NOT
- Allow users to confirm or deny snap account addition/removal. For now we will accept all incoming snap account additions/removals. Those confirmations will be handled in these following changes. https://github.com/MetaMask/accounts-planning/issues/152 and https://github.com/MetaMask/accounts-planning/issues/151
- This PR does not support custom for account snaps
- This PR does not support asynchronous approvals 
- This PR does not implement the improved account list U.I for snap accounts. For now the snap accounts in the account list will look exactly like the rest of the accounts.
- This PR does not support granular permissions required in production since there seems to be come issues with the mobile snaps permissions in general. see https://github.com/MetaMask/metamask-mobile/issues/11029. For now we will eagerly grant keyring permissions to dapps since this is just a flask feature and is not doing into production.
- Restrict wallet usage based on account permissions. Even if an account does not have signing permissions these flows will be available in the U.I and likely break.

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/149

## **Manual testing steps**

1. run yarn setup on this branch
2. navigate to .js.env and set `export METAMASK_BUILD_TYPE=flask`
5. `source .js.env`
6. `yarn watch:clean`
7. create/import a wallet
8. navigate to settings/snaps
9. there should be only one pre installed snap
10. open the browser
11. navigate to `https://metamask.github.io/snap-simple-keyring/latest/`
12. click connect and accept all of the permissions
13. you should now see this snap in the settings/snaps list
14. click create account on the snap simple keyring
15. the call should succeed and return the account info as a json obkect
16. navigate to the home page and open the account list, the selected account should be a new account named `Snap Account 1`
17. repeat step 15 as many times as you want, each time the new account should be added
18. clicking list accounts on the SSK dapp should give you the entire list of accounts
19. extract an id value from the listed accounts and paste it into the Remove accounts field and click.
20. the call should succeed and return null
21. navigate back to the home page and open the account list, that account should no longer be there and another account should be the selected account
22. 

### **Before**

N/A

### **After**

https://github.com/user-attachments/assets/79b0a22d-2bf3-4124-883d-82b91ab0121f


app/core/Snaps/SnapsMethodMiddleware.ts
```diff
@@ -36,6 +37,11 @@ const snapMethodMiddlewareBuilder = (
       engineContext.PermissionController,
       origin,
     ),
+    hasPermission: engineContext.PermissionController.hasPermission(origin),
```
FrederikBolding, line 40:
```suggestion
    hasPermission: engineContext.PermissionController.hasPermission.bind(engineContext.PermissionController, origin),
```

This should fix the error you are seeing and bring this closer to the extension implementation!

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -451,7 +451,7 @@ export class BackgroundBridge extends EventEmitter {
       snapMethodMiddlewareBuilder(
         Engine.context,
         Engine.controllerMessenger,
-        origin,
+        this.url,
```
owencraston, line 454: This was needed in order to include the `https` aspect of the url in order to validate the protocol in the `keyringSnapPermissionsBuilder`. This relates to an overall issue documented in [this issue](https://github.com/MetaMask/metamask-mobile/issues/11029).

app/core/Engine.ts
```diff
@@ -830,10 +911,22 @@ class Engine {
...
+            keyringSnapPermissionsBuilder(origin),
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          hasPermission: (origin: string, target: any) =>
+            this.controllerMessenger.call<'PermissionController:hasPermission'>(
+              'PermissionController:hasPermission',
+              origin,
+              target,
+            ),
+          // TODO: Code fence this block
+          getSnapKeyring: this.getSnapKeyring.bind(this),
```
owencraston, line 924: For some reason when I added a code fence around this block it breaks. 

```
          getSnapKeyring: this.getSnapKeyring.bind(this),
          ///: END:ONLY_INCLUDE_IF
```

results in this error in the watchman server...

```
app/core/Engine.ts: Invalid fence structure in file "app/core/Engine.ts":
A valid fence consists of two fence lines, but the file contains an uneven number, "47", of fence lines.
```

app/core/SnapKeyring/KeyringSnapPermissions.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      KeyringRpcMethod.ApproveRequest,
+      KeyringRpcMethod.RejectRequest,
+    ]);
+  });
+
+  it.each([
+    '',
+    'null',
+    'sftp://some-dapp.com',
+    'http://some-dapp.com',
```
owencraston, line 114: There is an issue with the test suite where is does not go through the code fencing step before testing. Since we conditionally include `http` in the `ALLOWED_PROTOCOLS` for flask this test is failing. This seems like a wider issue with jest testing and code fencing on mobile.



```
const ALLOWED_PROTOCOLS: string[] = [
  'https:',
  'http:',
  ///: END:ONLY_INCLUDE_IF
];
```

owencraston: I have since removed `http` from the `ALLOWED_PROTOCOLS` to get the tests passing.

app/core/Snaps/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Passes a JSON-RPC request object to the SnapController for execution.
+ *
+ * @param {object} args - A bag of options.
+ * @param {string} args.snapId - The ID of the recipient snap.
+ * @param {string} args.origin - The origin of the RPC request.
+ * @param {string} args.handler - The handler to trigger on the snap for the request.
+ * @param {object} args.request - The JSON-RPC request object.
+ * @returns The result of the JSON-RPC request.
+ */
+export async function handleSnapRequest(
```
owencraston, line 13: This is used in both the Engine and `app/core/Snaps/SnapsMethodMiddleware.ts`

ios/Podfile.lock
```diff
@@ -1302,6 +1302,6 @@ SPEC CHECKSUMS:
   Yoga: 6f5ab94cd8b1ecd04b6e973d0bc583ede2a598cc
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
-PODFILE CHECKSUM: e4050300fc9c8d091b9c00e1486ad6e205c250c7
+PODFILE CHECKSUM: 3514934ac1830af4ca844ba018f38720a48d006b
```
owencraston, line 1305: this needs to be removed once this checksum change gets committed to main. It currently gets created with yarn setup.

jest.config.js
```diff
@@ -7,6 +7,7 @@ process.env.MM_FOX_CODE = 'EXAMPLE_FOX_CODE';
 
 process.env.SECURITY_ALERTS_API_ENABLED = 'true';
 process.env.SECURITY_ALERTS_API_URL = 'https://example.com';
+process.env.METAMASK_BUILD_TYPE = 'main';
```
owencraston, line 10: I added this in an attempt to get the code fencing to work in unit test but it did not work. I figured it is okay to leave but can remove it if people feel otherwise.

app/core/Engine.ts
Jonathansoufer: The error is expected since on line 823 there's a code fence opening, which only closes on line 928. When you tried to add a code fence inside another code fence (getSnapKeyring: this.getSnapKeyring.bind(this), is at line 924), the "parser" will find an Opening (line 823) and EXPECT to find a closing (line 928), and not another opening (line 924)). As i reviewed the code, there's no need of any adding of code fence in this block.

app/core/SnapKeyring/KeyringSnapPermissions.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      KeyringRpcMethod.FilterAccountChains,
+      KeyringRpcMethod.UpdateAccount,
+      KeyringRpcMethod.DeleteAccount,
+      KeyringRpcMethod.ExportAccount,
+      KeyringRpcMethod.ListRequests,
+      KeyringRpcMethod.GetRequest,
+      KeyringRpcMethod.ApproveRequest,
+      KeyringRpcMethod.RejectRequest,
+    ]);
+  });
```
Jonathansoufer, line 108: If known and unknown origins can call the same methods, what's the difference between them?

app/core/SnapKeyring/keyringSnapsPermissions.ts
```diff
@@ -0,0 +1,103 @@
...
+ * The origins of the Portfolio dapp.
+ */
+const PORTFOLIO_ORIGINS: string[] = [
+  'https://portfolio.metamask.io',
+  ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+  'https://dev.portfolio.metamask.io',
+  'https://stage.portfolio.metamask.io',
+  'https://ramps-dev.portfolio.metamask.io',
+  'https://portfolio-builds.metafi-dev.codefi.network',
+  ///: END:ONLY_INCLUDE_IF
```
Jonathansoufer, line 14: Could we handle in the future something like `https://*.portfolio.metamask.io` instead?

Jonathansoufer (dismissed):
Excellent work and a huge step closer to what will bring such innovation to our beloved MetaMask.  Kudos! Left some small comments.

app/core/Engine.ts
```diff
@@ -732,17 +817,17 @@ class Engine {
...
-    ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps)
+    ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps,keyring-snaps)
     /**
      * Gets the mnemonic of the user's primary keyring.
      */
     const getPrimaryKeyringMnemonic = () => {
       // TODO: Replace "any" with type
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const [keyring]: any = keyringController.getKeyringsByType(
+      const [keyring]: any = this.keyringController.getKeyringsByType(
```
k-g-j, line 830: why do we need to use the `this` keyword here now?

app/core/Engine.ts
```diff
@@ -732,17 +817,17 @@ class Engine {
...
+      keyringBuilders: additionalKeyrings,
     });
 
-    ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps)
+    ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps,keyring-snaps)
     /**
      * Gets the mnemonic of the user's primary keyring.
      */
     const getPrimaryKeyringMnemonic = () => {
       // TODO: Replace "any" with type
```
k-g-j, line 828: Is this TODO possible yet?

app/core/Engine.ts
```diff
@@ -870,7 +963,7 @@ class Engine {
       // @ts-expect-error Typecast permissionType from getPermissionSpecifications to be of type PermissionType.RestrictedMethod
       permissionSpecifications: {
         ...getPermissionSpecifications({
-          getAllAccounts: () => keyringController.getAccounts(),
+          getAllAccounts: () => this.keyringController.getAccounts(),
```
k-g-j, line 966: same comment as above - why is the `this` keyword needed now when before it was working without it?

app/core/SnapKeyring/KeyringSnapPermissions.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    const permissions = keyringSnapPermissionsBuilder(
+      // TODO: Replace `any` with type
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      origin as any,
+    );
+    expect(permissions()).toStrictEqual([]);
+  });
+});
+
+describe('isProtocolAllowed', () => {
```
k-g-j, line 133: NIT: in the extension for testing functions with the function name we do `"#isProtocolAllowed"`

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,113 @@
...
+ * - `removeAccount`: Initiates the process of removing an account with user confirmation and handling the user input.
+ */
+export const snapKeyringBuilder = (
+  controllerMessenger: SnapKeyringBuilderMessenger,
+  getSnapController: () => SnapController,
+  persistKeyringHelper: () => Promise<void>,
+  removeAccountHelper: (address: string) => Promise<unknown>,
+): { (): SnapKeyring; type: string } => {
+  const builder = () => new SnapKeyring(getSnapController(), {
+    addressExists: async (address) =>
```
k-g-j, line 40: Can we add some JSDOC comments to these methods please so it is clearer was `redirectUser` or `saveState` is doing for people not as familiar with the snap keyring system as we are?

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,113 @@
...
+      Logger.log(
+        `SnapKeyring: addAccount called with \n
+        - address: ${address} \n
+        - handleUserInput: ${handleUserInput} \n
+        - snapId: ${snapId} \n
+        - accountNameSuggestion: ${accountNameSuggestion} \n
+        - displayConfirmation: ${displayConfirmation}`,
+      );
+
+      // approve everything for now
```
k-g-j, line 74: NIT: give some reasoning here such as "approve everything for now _because_ we have not implemented snap account confirmations yet"

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,113 @@
...
+      snapId: string,
+      handleUserInput: (accepted: boolean) => Promise<void>,
+    ) => {
+      Logger.log(
+        `SnapKeyring: removeAccount called with \n
+        - address: ${address} \n
+        - handleUserInput: ${handleUserInput} \n
+        - snapId: ${snapId} \n`,
+      );
+      // approve everything for now
```
k-g-j, line 105: same comment as above about providing reasoning

app/core/Engine.ts
owencraston: Okay thank you. I removed the todo.

app/core/SnapKeyring/KeyringSnapPermissions.test.ts
owencraston: @Jonathansoufer this is related to one of the big issues I was having with the permission builder. I posted a discussion in slack but I will repost it here. This is something I would greatly appreciate your guidance on. 

Here is the current code block for mobile.

```ts
export function keyringSnapPermissionsBuilder(
  origin: string,
): () => string[] {
  return () => {
    if (origin === 'metamask') {
      return METAMASK_ALLOWED_METHODS;
    }

    if (PORTFOLIO_ORIGINS.includes(origin)) {
      return PORTFOLIO_ALLOWED_METHODS;
    }
    return isProtocolAllowed(origin) ? WEBSITE_ALLOWED_METHODS : [];
  };
}
```
extension:

```ts
export function keyringSnapPermissionsBuilder(
  controller: SubjectMetadataController,
  origin: string,
): () => string[] {
  return () => {
    if (origin === 'metamask') {
      return METAMASK_ALLOWED_METHODS;
    }

    if (PORTFOLIO_ORIGINS.includes(origin)) {
      return PORTFOLIO_ALLOWED_METHODS;
    }

    const originMetadata = controller.getSubjectMetadata(origin);
    if (originMetadata?.subjectType === SubjectType.Website) {
      return isProtocolAllowed(origin) ? WEBSITE_ALLOWED_METHODS : [];
    }

    return [];
  };
 ```
as you can see we are not performing the `originMetadata?.subjectType` check because when I do this on mobile the response for getSubjectMetadata is undefined. I logged the state of the SubjectMetadataController and noticed that it only stored information by snapId not url so this logic would not work on mobile. For now I am returning the full bag of permissions if it is using https. I am curious what people think we should do here given these contraints?

File: app/core/SnapKeyring/SnapKeyring.ts (hunk 14 above)
Comment (owencraston) [reply app/core/SnapKeyring/SnapKeyring.ts:40]: @k-g-j these return methods are already documented in the jsdoc. 

Screenshot 2024-09-18 at 12 43 09 PM

Comment (owencraston) [reply app/core/SnapKeyring/SnapKeyring.ts:40]: Screenshot 2024-09-18 at 12 47 48 PM

File: app/core/SnapKeyring/keyringSnapsPermissions.ts (hunk 9 above)
Comment (owencraston) [reply app/core/SnapKeyring/keyringSnapsPermissions.ts:14]: I would need to think about this a bit more. I feel like there is some extra security benefits of being specific.  By adding the start method we are forced to use regex for the lookup which adds complexity and decreases performance. I am in favour of this simpler approach which is more verbose but arguably easier to maintain in the long run. I do not feel strongly about this though and we can definitely improve it in the future.

File: app/core/SnapKeyring/keyringSnapsPermissions.ts [hunk 17]
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Builds a function that returns the list of keyring methods an origin can
+ * call.
+ * - Here is the equivalent function on the extension: https://github.com/MetaMask/metamask-extension/blob/develop/app/scripts/lib/snap-keyring/keyring-snaps-permissions.ts#L96
+ * @param controller - Reference to the `SubjectMetadataController`.
+ * @param origin - The origin itself.
+ * @returns A function that returns the list of keyring methods an origin can
+ * call.
+ */
+export function keyringSnapPermissionsBuilder(
```
owencraston, line 90: @Jonathansoufer This is something I would greatly appreciate your guidance on. 

Here is the current code block for mobile.

```ts
export function keyringSnapPermissionsBuilder(
  origin: string,
): () => string[] {
  return () => {
    if (origin === 'metamask') {
      return METAMASK_ALLOWED_METHODS;
    }

    if (PORTFOLIO_ORIGINS.includes(origin)) {
      return PORTFOLIO_ALLOWED_METHODS;
    }
    return isProtocolAllowed(origin) ? WEBSITE_ALLOWED_METHODS : [];
  };
}
```
extension:

```ts
export function keyringSnapPermissionsBuilder(
  controller: SubjectMetadataController,
  origin: string,
): () => string[] {
  return () => {
    if (origin === 'metamask') {
      return METAMASK_ALLOWED_METHODS;
    }

    if (PORTFOLIO_ORIGINS.includes(origin)) {
      return PORTFOLIO_ALLOWED_METHODS;
    }

    const originMetadata = controller.getSubjectMetadata(origin);
    if (originMetadata?.subjectType === SubjectType.Website) {
      return isProtocolAllowed(origin) ? WEBSITE_ALLOWED_METHODS : [];
    }

    return [];
  };
 ```
as you can see we are not performing the `originMetadata?.subjectType` check because when I do this on mobile the response for getSubjectMetadata is undefined. I logged the state of the SubjectMetadataController and noticed that it only stored information by snapId not url so this logic would not work on mobile. For now I am returning the full bag of permissions if it is using https. I am curious what people think we should do here given these contsraints?

File: app/core/SnapKeyring/KeyringSnapPermissions.test.ts (hunk 13 above)
Comment (owencraston) [reply app/core/SnapKeyring/KeyringSnapPermissions.test.ts:133]: @k-g-j this is not what I am seeing on the extension.

Screenshot 2024-09-18 at 12 55 34 PM

File: app/core/Engine.ts (hunk 10 above)
Comment (owencraston) [reply app/core/Engine.ts:848]: The reason this. is now needed on a few controllers is because I modified the Engine class such that the `keyringController` `subjectMetadataController` and `snapController` are all class properties on the Engine. This differs from before where they were merely constants inside the class. The reason for this change is that is makes initializing these controller much easier. If these controllers remain as constants the order of initialization starts to matter. What I mean by this is that I cannot initialize the `snapKeyringBuilder` until I initialize the `snapsController` and I cannot initialize the `snapsController` until Initialize the `keyringController` and so on. Building it this way would require moving hundreds of lines around in the Engine and possibly running into circular dependencies. Having these "important" controllers as class variables ensures that we can access them across the file no matter when they are initialized. This pattern also matches more closely what is being done on the extension now.

File: app/core/Engine.ts (hunk 12 above)
Comment (owencraston) [reply app/core/Engine.ts:988]: see above comment.

File: app/core/BackgroundBridge/BackgroundBridge.js (hunk 2 above)
Comment (owencraston) [reply app/core/BackgroundBridge/BackgroundBridge.js:454]: @Jonathansoufer @Cal-L will this effect the notifications snap in any adverse way?

**Comment by owencraston** [comment]
I updated the human readable versions of the snaps permissions to include `snap_manageAccounts` and use updated copy for `endowment:keyring`. These values come from this [doc](https://www.notion.so/metamask-consensys/bac3299d2c5241c599d2e5e7986e72f7?v=ef742a61bd844435b7171bd2e90b447e). 

Before: 
Screenshot 2024-09-18 at 1 51 49 PM


After:
Screenshot 2024-09-18 at 1 47 32 PM

File: app/core/Snaps/SnapsMethodMiddleware.ts [hunk 18]
```diff
@@ -4,16 +4,25 @@ import {
...
+import { SnapRpcHookArgs } from '@metamask/snaps-utils';
 import { RestrictedMethods } from '../Permissions/constants';
+import { keyringSnapPermissionsBuilder } from '../SnapKeyring/keyringSnapsPermissions';
+import { SnapId } from '@metamask/snaps-sdk';
+import { EngineContext } from '../Engine';
+import { handleSnapRequest } from './utils';
 
+export function getSnapIdFromRequest(
+  request: Record<string, unknown>,
+): string | null {
```
legobeat, line 16:
```suggestion
): SnapId | null {
```

Would remove the need to typecast the result for usage as `SnapId`

app/core/Snaps/SnapsMethodMiddleware.ts
```diff
@@ -51,6 +65,26 @@ const snapMethodMiddlewareBuilder = (
...
+      const snapId = getSnapIdFromRequest(request);
+
+      if (!snapId) {
+        throw new Error(
+          'snapMethodMiddlewareBuilder handleSnapRpcRequest: Invalid snap request: snapId not found',
+        );
+      }
+
+      return await handleSnapRequest(controllerMessenger, {
+        snapId: snapId as SnapId,
```
legobeat, line 82:
```suggestion
        snapId,
```
https://github.com/MetaMask/metamask-mobile/pull/10829/files#r1765861509

app/core/Engine.ts
```diff
@@ -733,11 +760,69 @@ class Engine {
...
+        `${approvalController.name}:startFlow`,
+        `${approvalController.name}:endFlow`,
+        `${approvalController.name}:showSuccess`,
+        `${approvalController.name}:showError`,
+        `${phishingController.name}:testOrigin`,
+        `${phishingController.name}:maybeUpdateState`,
+        `KeyringController:getAccounts`,
+        'AccountsController:setSelectedAccount',
+        'AccountsController:getAccountByAddress',
+        'AccountsController:setAccountName',
```
ccharly, line 801: Nit/Cosmetic: We might wanna want to keep the "style" consistent here, so either we use plain string or we use template string everywhere, WDYT?

NOTE: Template string might not be possible here, since we instantiate the `KeyringController` after this, so I'd probably go with plain strings here

app/core/Engine.ts
ccharly: We also need this in some functions like `persistAndUpdateAccounts` which is referenced before the `KeyringController` instantiation.

We could use `let keyringController` probably, but having it on the `Engine` feels like more natural to me 👍.

app/core/Engine.ts
```diff
@@ -848,10 +929,21 @@ class Engine {
               origin,
             );
           },
+          getAllowedKeyringMethods: (origin: string) =>
+            keyringSnapPermissionsBuilder(origin),
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          hasPermission: (origin: string, target: any) =>
```
ccharly, line 935: Shouldn't this be `string` instead of `any` here? Looking at: https://github.com/MetaMask/core/blob/main/packages/permission-controller/src/PermissionController.ts#L824-L828

```suggestion
          hasPermission: (origin: string, target: string) =>
```

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param controller - Instance of the MetaMask Controller.
+ * @param snapId - Snap ID to get accounts for.
+ * @returns The addresses of the accounts.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const getAccountsBySnapId = async (engine: any, snapId: SnapId) => {
+  const snapKeyring: SnapKeyring = await engine.getSnapKeyring();
+  return await snapKeyring.getAccountsBySnapId(snapId);
+};
+
```
ccharly, line 19: This does not seem to be used anywhere?

Also, if you want to keep this for future PR, we could try to remove the use of `any` here, maybe with a type like: `{ getSnapKeyring: () => SnapKeyring }`?

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,115 @@
...
+        await controllerMessenger.call('KeyringController:getAccounts')
+      ).includes(address.toLowerCase()),
+
+    redirectUser: async (snapId: string, url: string, message: string) => {
+      Logger.log(
+        `SnapKeyring: redirectUser called with \n
+        - snapId: ${snapId} \n
+        - url: ${url} \n
+        - message: ${message} \n`,
+      );
```
ccharly, line 53: Since we are only logging here, we could add a comment to explain why there's no "real implementation", WDYT?

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,115 @@
...
+      Logger.log(
+        `SnapKeyring: addAccount called with \n
+        - address: ${address} \n
+        - handleUserInput: ${handleUserInput} \n
+        - snapId: ${snapId} \n
+        - accountNameSuggestion: ${accountNameSuggestion} \n
+        - displayConfirmation: ${displayConfirmation}`,
+      );
+
+      //approve everything for now because we have not implemented snap account confirmations yet
```
ccharly, line 76:
```suggestion
      // Approve everything for now because we have not implemented snap account confirmations yet
```

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -0,0 +1,115 @@
...
+    ) => {
+      Logger.log(
+        `SnapKeyring: removeAccount called with \n
+        - address: ${address} \n
+        - handleUserInput: ${handleUserInput} \n
+        - snapId: ${snapId} \n`,
+      );
+      // approve everything for now because we have not implemented snap account confirmations yet
+      await removeAccountHelper(address);
+      await handleUserInput(true);
```
ccharly, line 109: Just re-ordering things, since the comment is related to `handleUserInput` and `persistKeyringHelper`:

```suggestion
      await removeAccountHelper(address);

      // Approve everything for now because we have not implemented snap account confirmations yet
      await handleUserInput(true);
```

app/core/Snaps/SnapsMethodMiddleware.ts
ccharly: But we might need to typecast `snapId` here (which might be inferred as `string` while using:

```ts
return typeof snapId === 'string' ? snapId : null;
```

Would need to be changed to:

```ts
return typeof snapId === 'string' ? snapId as SnapId : null;
```

In anyway, I do agree with you, this function should return a `SnapId`, so typecast should happen here rather than elsewhere :)

app/core/Snaps/SnapsMethodMiddleware.ts
```diff
@@ -4,16 +4,25 @@ import {
...
 import { RestrictedMethods } from '../Permissions/constants';
+import { keyringSnapPermissionsBuilder } from '../SnapKeyring/keyringSnapsPermissions';
+import { SnapId } from '@metamask/snaps-sdk';
+import { EngineContext } from '../Engine';
+import { handleSnapRequest } from './utils';
 
+export function getSnapIdFromRequest(
+  request: Record<string, unknown>,
+): string | null {
+  const snapId = request.snapId;
```
ccharly, line 17: What about de-structuring here:

```suggestion
  const { snapId } = request;
```

app/core/Snaps/SnapsMethodMiddleware.ts
```diff
@@ -4,16 +4,25 @@ import {
...
     from extension https://github.dev/MetaMask/metamask-extension/blob/1d5e8a78400d7aaaf2b3cbdb30cff9399061df34/app/scripts/metamask-controller.js#L3830-L3861
     */
 const snapMethodMiddlewareBuilder = (
-  // TODO: Replace "any" with type
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  engineContext: any,
+  engineContext: EngineContext,
   // TODO: Replace "any" with type
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   controllerMessenger: any,
```
ccharly, line 28: We could fix this since you are now exporting the `Engine.ControllerMessenger` type?

app/core/Snaps/utils.ts
```diff
@@ -0,0 +1,23 @@
+import { ControllerMessenger } from '../Engine';
+import { HandleSnapRequestArgs } from './types';
+/**
```
ccharly, line 3:
```suggestion

/**
```

app/core/Snaps/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function handleSnapRequest(
+  controllerMessenger: ControllerMessenger,
+  args: HandleSnapRequestArgs,
+) {
+  return await controllerMessenger.call('SnapController:handleRequest', {
+    snapId: args.snapId,
+    origin: args.origin,
+    handler: args.handler,
+    request: args.request,
+  });
```
ccharly, line 22: We could just use `args` here, no?

```suggestion
  return await controllerMessenger.call('SnapController:handleRequest', args);
```

app/core/Engine.ts
owencraston: good catch

app/core/SnapKeyring/SnapKeyring.ts
owencraston: I brought this over from extension but realized it is not needed. I will remove it now and add it when the time comes.

app/core/SnapKeyring/SnapKeyring.ts
owencraston: I added a small TODO above the two logs.

app/core/SnapKeyring/SnapKeyring.ts
owencraston: I based this order off of what is already [in production on the extension](https://github.com/MetaMask/metamask-extension/blob/ed2850844e2af8381284468313c920b5c6e5084a/app/scripts/lib/snap-keyring/snap-keyring.ts#L365-L367). do you think we should still change the order?

app/core/Snaps/SnapsMethodMiddleware.ts
owencraston: clean

app/core/Snaps/SnapsMethodMiddleware.ts
owencraston: I actually did this in the engine with this code...

```ts
export type ControllerMessenger = ExtendedControllerMessenger<
  GlobalActions,
  GlobalEvents
>;
```


The problem was there is a type mismatch between what the controller messenger expects and what the snaps controller methods expect which is causing type errors when using the controller below. This is likely due to the fact that the controllers are not typed on the extension so some type mis matches snuck through. I can do this but I will need to ts-ignore on all of the lines that use the controller messenger.

app/core/SnapKeyring/keyringSnapsPermissions.ts
owencraston: after some more discussions with @gantunesr we think this is alright since this is just a flask feature and the permissions will change drastically with pre installed snaps. When it comes time to implement those we will need to iterate on this permission builder.

Cal-L (dismissed):
Engine LGTM from platform perspective

Cal-L (approved):
Engine LGTM

k-g-j (dismissed):

Jonathansoufer (dismissed):
Approved. I think the comms I left can be handled in a separated PR.

Cal-L (dismissed):
Engine LGTM

k-g-j (dismissed):

k-g-j (approved):

Jonathansoufer (approved):
LGTM!

Cal-L (approved):
LGTM

PR #10831, opened 2024-08-27
feat: Update Polygon from `MATIC` to `POL`

gambinish:
## **Description**

Adds POL token and ticker to networks configs. See [blog](https://polygon.technology/blog/save-the-date-matic-pol-migration-coming-september-4th-everything-you-need-to-know) for more details

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MMASSETS-362

## **Manual testing steps**

1. When adding Polygon Network from a network from, ticker should be POL
2. When switching to Polygon Mainnet, ticker should be POL (not MATIC)
3. Can import Polygon Network Token on Ethereum Mainnet
4. Can import Matic Network Token on Ethereum Mainnet (backward compatibility)
Note, that both MATIC and POL erc20 tokens can already be imported on Ethereum Mainnet, so this behavior should remain unchanged.

Additionally, when a user upgrades to this build, migration should run and overwrite ticker from `MATIC` to `POL`

Install a previous build (checkout main and `yarn && yarn setup && yarn watch` should be fine here) -> Polygon Network should show MATIC as ticker. You may need to remove and re-add Polygon Network, as migrations are incremental.
checkout to this branch, `yarn && yarn setup && yarn watch`, migration 051 should run -> Polygon Network should show POL as ticker, even when default selected Network was MATIC in previous build.

## **Screenshots/Recordings**

Screenshot 2024-08-27 at 11 11 31 PM


app/store/migrations/051.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  // update ticker to POL in providerConfig if chainId is 0x89 and ticker is MATIC
+  // needed if user is already on selectedNetworkId that maps to pre-existing MATIC ticker
+  if (
+    networkControllerState.providerConfig.chainId === '0x89' &&
+    networkControllerState.providerConfig.ticker === 'MATIC'
+  ) {
+    networkControllerState.providerConfig.ticker = 'POL';
+  }
```
gambinish, line 74: This seems to be needed in order to upgrade a user if their `selectedNetworkId` is set to a previously set ticker of `MATIC`.

This cause this weird state after migration, where the ticker was `POL` (correct) but it was being flagged as spam because the ticker of `providerConfig.<polygon-network-uuid>.ticker` was still `MATIC` (incorrect). This change updates it in `providerConfig` as well.

Though, I didn't have this issue in extension. Is this expected?

app/store/migrations/051.test.ts
```diff
@@ -0,0 +1,256 @@
...
+import initialRootState from '../../util/test/initial-root-state';
+import { captureException } from '@sentry/react-native';
+import { NetworkState } from '@metamask/network-controller';
+
+jest.mock('@sentry/react-native', () => ({
+  captureException: jest.fn(),
+}));
+const mockedCaptureException = jest.mocked(captureException);
+
+describe('Migration #51', () => {
```
sethkfman, line 12: @tommasini Can you take a look at the migration implementation?

app/components/Views/Settings/NetworksSettings/NetworkSettings/index.js
```diff
@@ -485,6 +485,13 @@ export class NetworkSettings extends PureComponent {
         (network) => network.chainId === parseInt(chainId),
       );
 
+      // This is a temporary hack to not include POL as a potential scam token while chainlist updates
+      // TODO: This can be safely removed once safeChainsList updates from MATIC to POL
```
tommasini, line 489: Created a [PR](https://github.com/ethereum-lists/chains/pull/5757) for we know when we can remove this temporary solution

app/store/migrations/051.ts
```diff
@@ -0,0 +1,79 @@
...
+          networkControllerState.providerConfig,
+        )}'`,
+      ),
+    );
+    return state;
+  }
+
+  // update network settings if chain 0x89 and ticker is MATIC
+  for (const networkConfiguration of Object.values(
+    networkControllerState.networkConfigurations as Record<
```
tommasini, line 53: Is there any reason why we need to cast this type?

app/store/migrations/051.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param state Persisted Redux state
+ * @returns
+ */
+export default function migrate(state: unknown) {
+  if (!ensureValidState(state, 51)) {
+    return state;
+  }
+
+  const networkControllerState = state.engine.backgroundState
+    .NetworkController as NetworkState;
```
tommasini, line 18: Same here, we shouldn't type it as NetworkState, we should avoid the typing, on the future, when we change the NetworkState, we will have trouble on this file because it doesn't fit anymore with the current variables, but since this is a migration, at this point this would always be true

This is something that is on past migrations, but the goal is to keep improving the migrations! 

So the best thing to do while we migrate data is every step be a safe step and not assuming by type casting that the variable it's there.

app/store/migrations/051.ts
bergeron: We'll probably have this on extension too, if we hotfix it.  In develop we're on a later version of the network controller that no longer has providerConfig, so there'd be nothing to do there.  But for a 12.1.x hotfix we would still have providerConfig.  You can test the migration against that branch https://github.com/MetaMask/metamask-extension/tree/Version-v12.1.0

tommasini: This sounds correct to me! If the providerConfig data is matic, we should change it to POL, not finding this issue on extension makes me think that the providerConfig ticker is overridden after by the ticker on networkConfigurations, but I'm just assuming so do not take my words for granted. 

On the mobile app, when the app opens, the Engine initializes and the [NetworkController](https://github.com/MetaMask/metamask-mobile/blob/5db2f60c7216e9fcd32896ad56fc51464e3c905a/app/core/Engine.ts#L495) as well, and it uses the data that is persisted on the device. I can see the NetworkController [constructor](https://github.com/MetaMask/core/blob/f8e343bde43a7510a1e2856abf15cdee004db179/packages/network-controller/src/NetworkController.ts#L592) uses the new state, and that will make providerConfig use the ticker that was persisted (which with this migration will be SOL instead of MATIC)

app/store/migrations/051.test.ts
```diff
@@ -0,0 +1,256 @@
...
+        engine: {
+          backgroundState: {
+            NetworkController: {
+              providerConfig: null,
+            },
+          },
+        },
+      }),
+      errorMessage:
+        "FATAL ERROR: Migration 51: NetworkController networkConfigurations not found: 'null'",
```
tommasini, line 120: This mesasge doesn't seem accurate, did you wanted to check if the providerConfig was not found?

app/store/migrations/051.ts
tommasini: Ahhh nevermind, my bad! I see we only change the MATIC ones

app/store/migrations/051.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      scenario: 'providerConfig is null',
+    },
+  ];
```
tommasini, line 123: In order to improve the migration, and we are removing the cast, do you think it would be a good pratice to check for the variable `ticker` `chaindId` on `networkConfigurations` and `providerConfig` objects?

app/store/migrations/051.ts
gambinish: @tommasini this will fail the `lint:tsc` check if not typecasted. Should I `@ts-ignore` this file?

tommasini: What errors will the typescript show?

tommasini: I'm recommending not casting and using type-safe functions on the go, like checking if chainId exists or the ticker variable exists inside of networkConfigurations and providerConfig objects 🙏

app/store/migrations/051.ts
```diff
@@ -0,0 +1,79 @@
+import { captureException } from '@sentry/react-native';
```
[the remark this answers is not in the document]
gambinish: Hmmm, this should only update the ticker if the `chainId` is `'0x89'` && the ticker is `MATIC`

If the user added the Polygon Network, with the custom ticker "polygon", it should remain unchanged. I believe I have unit tests to check for this case both in the `networkConfiguration` and `providerConfig`. Lmk if you have any suggestion on making that more clear?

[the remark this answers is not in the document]
gambinish: Am I missing something here? Please lmk if I am.

app/store/migrations/051.test.ts
gambinish: Good catch, updating.

gambinish: https://github.com/MetaMask/metamask-mobile/pull/10831/commits/5f01a999ad96d7ae3b3614b61d24944bfe881b6e

app/store/migrations/051.test.ts
gambinish: Removed typecasting here: https://github.com/MetaMask/metamask-mobile/pull/10831/commits/7548ddf36656c39524d63cf8ff10aabf356c53aa

app/store/migrations/051.ts
gambinish: Thanks for walking me through this. Removed typecasting here: https://github.com/MetaMask/metamask-mobile/pull/10831/commits/7548ddf36656c39524d63cf8ff10aabf356c53aa

app/store/migrations/051.ts
gambinish: Updated here: https://github.com/MetaMask/metamask-mobile/pull/10831/commits/7548ddf36656c39524d63cf8ff10aabf356c53aa 👍

app/store/migrations/051.test.ts
```diff
@@ -0,0 +1,257 @@
...
+          },
+        },
+      },
+    });
+
+    const newState = await migrate(modifiedOldState);
+
+    const customNotMaticPolygon = Object.values(
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
tommasini, line 193: Let's remove it here too

app/store/migrations/051.test.ts
```diff
@@ -0,0 +1,257 @@
...
+          },
+        },
+      },
+    });
+
+    const newState = await migrate(modifiedOldState);
+
+    expect(
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
tommasini, line 253: and here

gambinish: https://github.com/MetaMask/metamask-mobile/pull/10831/commits/ee4252cc2c99c0c92fd3e952910c6bd379c9babe

app/store/migrations/051.test.ts
gambinish: https://github.com/MetaMask/metamask-mobile/pull/10831/commits/ee4252cc2c99c0c92fd3e952910c6bd379c9babe

app/store/migrations/051.test.ts
```diff
@@ -0,0 +1,257 @@
...
+          },
+        },
+      },
+    });
+
+    const newState = await migrate(modifiedOldState);
+
+    const kovanNetwork = Object.values(
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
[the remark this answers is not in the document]
gambinish: ✅

wachunei (approved):
Ramp changes LGTM

tommasini (approved):

PR #10936, opened 2024-08-30
fix: Add missing allowedAction NetworkController:findNetworkClientIdByChainId to transactionController

jpuri:
## **Description**

Add missing allowedAction NetworkController:findNetworkClientIdByChainId to transactionController

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/10853

## **Manual testing steps**
NA

## **Screenshots/Recordings**
NA

## **Pre-merge author checklist**

- [X] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [X] I've completed the PR template to the best of my ability
- [X] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [X] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #11002, opened 2024-09-03
fix: infinite gas fee loading during send

### LINKED ISSUE #10895: [Bug]: Gas info not loading for send transactions initiated within the wallet
### Describe the bug

After updating to v7.29 and starting a send transaction within the app, some users are not able to confirm the transaction due to the gas info never loading. 

The issue happens to all networks, but only for send transactions started within the app. MM swaps and dapp proposed transactions seem to be unaffected. 

### Expected behavior

Gas info to load timely and users be able to confirm the transaction.

### Screenshots/Recordings

image

### Steps to reproduce

1. Update your app to v7.29
2. Start a send transaction
3. See the MM confirmation for the send transaction with the gas info loading forever

Note: not everyone is able to repro this so there's more steps to repro that we don't know of yet

_No response_

### Detection stage

In production (default)

### Version

7.29

### Build type

None

### Device

iphone

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

matthewwalsh0:
## **Description**

Additional fixes required to prevent infinite or delayed gas fee loading when creating transfers from the wallet.

The core of the problem is as follows:

1. Following the recent upgrade to version `35.0.0` of the `TransactionController`, the `networkClientId` is automatically set in the transaction metadata when adding a transaction.
2. This meant legacy transactions already in the wallet before upgrade did not have this value set.
3. When the `GasFeePoller` polls automatically in the `TransactionController`, it needs to determine the `networkClientId` of each transaction in order to retrieve the correct estimates from the `GasFeeController`.
4. If the `networkClientId` is not in the transaction metadata, this is derived using the `NetworkController:findNetworkClientIdByChainId` messenger action which was not defined as an allowed action for the `TransactionController` in the mobile client until a recent PR.
5. Send transactions are unique in that they exclusively used the gas fee estimates from the transaction metadata, and do not fallback to the network gas fee estimates provided by the `GasFeeController` state directly.
6. All of the above ultimately meant that if a user attempted to create a send transaction and had `unapproved` transactions in their state from a previous release, then the gas fee polling would consistently fail (due to the disallowed action) meaning transaction gas fee estimates were never updated and therefore the UI would never load and never allow confirmation of the transaction.

The core issue has unknowingly already been resolved by the PR #10936 

For added stability, this PR also:

1. Updates send confirmations to also fallback to the network gas fee estimates to increase their stability in the face of errors and not block transactions.
2. Clears all unapproved transactions on startup, since they are never displayed and provide no benefit while needlessly increasing state size, background processing, and callouts.

Further logging and modular error handling will also be added to the `TransactionController` going forward.

## **Related issues**

Fixes: #10895 

## **Manual testing steps**

To re-create the original issue:

1. Install or checkout version `7.27.0`.
2. Add a transaction but hard reset / close the app while the confirmation is visible.
4. Install or checkout version `7.28.0` or later.
5. Start a transfer.
6. Note that the gas fee never loads.

### DIFF
--- a/app/components/Views/confirmations/SendFlow/Confirm/index.js
+++ b/app/components/Views/confirmations/SendFlow/Confirm/index.js
@@ -112,9 +112,9 @@ import CustomGasModal from './components/CustomGasModal';
 import { ResultType } from '../../components/BlockaidBanner/BlockaidBanner.types';
 import { withMetricsAwareness } from '../../../../../components/hooks/useMetrics';
 import {
-  selectTransactionGasFeeEstimates,
   selectCurrentTransactionMetadata,
   selectCurrentTransactionSecurityAlertResponse,
+  selectGasFeeEstimates,
 } from '../../../../../selectors/confirmTransaction';
 import { selectGasFeeControllerEstimateType } from '../../../../../selectors/gasFeeController';
 import { createBuyNavigationDetails } from '../../../../UI/Ramp/routes/utils';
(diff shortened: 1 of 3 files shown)


OGPoyraz (approved):

digiwand (approved):

sethkfman (approved):
LGTM

sleepytanya:
https://github.com/user-attachments/assets/f227fcb4-666d-4bf7-832e-37e3eb85f290

PR #10880, opened 2024-08-29
chore: Enable linting Pods and re-organize setup.mjs file

Cal-L:
## **Description**

This PR refactors setup.mjs file with the purpose of:
1. Moves Gemfile into iOS directory, which is where it belongs since it only relates to iOS. We also updated everywhere that calls bundle install or exec so it's calling it in the ios directory
2. Run pod install in CI for detecting diffs in Podfile.lock

Because we're now running pod install in the CI, the total `ci.yml` job time will take longer. However, it still completes before some other tasks such as unit tests. And since it runs in parallel to those, we should not see an increase in total CI run time.

## **Related issues**

- https://github.com/MetaMask/metamask-mobile/issues/11293
- #9789

## **Manual testing steps**

- Running `yarn setup` still works normally when run locally
- CI passes when run
- CI fails when Podfile.lock is changed when pod install runs in the CI

### **After**

Example of pod file diff detected
https://github.com/MetaMask/metamask-mobile/actions/runs/10746121351/job/29806449687?pr=10880

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [x] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [x] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #11052, opened 2024-09-05
chore: fix concurrency issue in setup.mjs

legobeat:
- Targeting: #10880
- fix race-conditions in `setup.mjs`
- support supplying multiple cli args
  - removes need for `--diff` as it's equivalent to `--node --build-ios` 
- throw error when passing unsupported cli args


scripts/setup.mjs
```diff
@@ -120,7 +131,7 @@ const buildPpomTask = {
 const setupIosTask = {
   title: 'Set up iOS',
   task: async (_, task) => {
-    if (!BUILD_IOS && !IS_DIFF) {
+    if (!BUILD_IOS) {
```
legobeat, line 134: I guess this check is actually redundant considering the same condition is considered further down.

legobeat: Removing it.

scripts/setup.mjs
```diff
@@ -120,10 +131,6 @@ const buildPpomTask = {
 const setupIosTask = {
   title: 'Set up iOS',
   task: async (_, task) => {
-    if (!BUILD_IOS && !IS_DIFF) {
```
legobeat, line 123: I found it more managable to gather all preconditional checks into one place (at the bottom) rather than doing it inside the task itself.

PR #11087, opened 2024-09-06
fix: Intermittent Display Issue of Fiat Currency on Main Wallet View

### LINKED ISSUE #11035: [Bug]: Intermittent Display Issue of Fiat Currency on Main Wallet View
### Describe the bug

Main wallet view intermittently displays $0 fiat currency on account balance despite account holding ETH.

### Expected behavior

The fiat currency should consistently reflect the correct value corresponding to the amount of ETH in the account, even after the device is left on standby.

### Screenshots/Recordings

IMG_0054

### Steps to reproduce

1. Log in to the app on an iOS device Or Android.
2. Verify that the fiat currency is correctly displayed on the main wallet view.
3. Leave the device untouched on the main wallet view for several minutes.
4. Observe that the fiat currency intermittently shows $0

### Detection stage

During release testing

### Version

7.30.0

### Build type

Other (please specify exactly where you obtained this build in "Additional Context" section)

### Device

IPhone 14
Pixel 5

### Operating system

iOS
Android OS

### Additional context

Build 1410
Version 7.30.0

### Severity

Sev1

gambinish:
## **Description**

This bug is difficult to reproduce organically, but we suspect that the root cause is that `fetchExchangeRate` function to crypto-compare could sometimes be down, causing the function to throw an error prematurely. We were able to simulate this locally, and it caused the zero balance to occur.

In `CurrencyController` we `updateExchangeRate` in a finally block (so, even when there is an error in the fetchExchangeRates call). In this case, the Controller will returns nullish values, causing the balance to zero out.

This patch only updates state in the `try` block, after a successful response. We are maintaining existing error handling in the `catch` statement, and keeping the `releaseLock` in the finally to handle concurrent requests gracefully.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/11035

## **Manual testing steps**

1. Run app normally, to get initial balance
2. Throw error in `fetchExchangeRate` function
3. Balance should not zero out like it did before this patch.

### **Before**

https://github.com/user-attachments/assets/ac0fa1c2-9f4d-4e42-85f4-9b5edb8fc753

### **After**

https://github.com/user-attachments/assets/62c3fc40-9dfc-4573-97e5-6c2a1fabcec8


patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,40 @@ index 0dc70ec..461a210 100644
...
+diff --git a/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js b/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js
+index 45254ad..8a82cd7 100644
+--- a/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js
++++ b/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js
+@@ -101,11 +101,6 @@ var CurrencyRateController = class extends _pollingcontroller.StaticIntervalPoll
+         usdConversionRate = fetchExchangeRateResponse.usdConversionRate;
+         conversionDate = Date.now() / 1e3;
+       }
+-    } catch (error) {
+-      if (!(error instanceof Error && error.message.includes("market does not exist for this coin pair"))) {
```
bergeron, line 35: It'd help to have a PR to core first so this can be reviewed without looking at a patch.

The one behavior that looks different, is that when we get `market does not exist for this coin pair` - we used to update the state with null values.  But now it looks like it won't update state.  I don't know how the app will handle this.  At least test some networks where the native currency doesn't support pricing.

patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,40 @@ index 0dc70ec..461a210 100644
...
+         this.update(() => {
+           return {
+@@ -120,9 +115,15 @@ var CurrencyRateController = class extends _pollingcontroller.StaticIntervalPoll
+             currentCurrency
+           };
+         });
+-      } finally {
+-        releaseLock();
++      } catch (error) {
++        throw new Error("Error: updateExchangeRate failed to update state")
```
bergeron, line 49: The other different behavior is this explicit throw if the state update fails.  Is there a reason that was necessary? Should change as little as possible

patches/@metamask+assets-controllers+30.0.0.patch
```diff
@@ -23,6 +23,40 @@ index 0dc70ec..461a210 100644
        }
      };
      return acc;
+diff --git a/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js b/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js
+index 45254ad..8a82cd7 100644
+--- a/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js
++++ b/node_modules/@metamask/assets-controllers/dist/chunk-ELSMS5S7.js
```
bergeron, line 29: Based on other 2 comments, I'd suggest something like this:
https://github.com/MetaMask/core/pull/4662

patches/@metamask+assets-controllers+30.0.0.patch
gambinish: I can remove this catch, I suppose it's not necessary

patches/@metamask+assets-controllers+30.0.0.patch
gambinish: https://github.com/MetaMask/metamask-mobile/pull/11087/commits/4cf049e4cb610f088243843cbc56118866f77be2

bergeron:
testing this...

bergeron (approved):

tommasini:
Loud thoughts, would this be possible to be caused by [price api](https://github.com/MetaMask/core/blob/c1a8b82a6759813afa4620ad2fd11bebd9f106c9/packages/assets-controllers/src/token-prices-service/codefi-v2.ts#L422) as well used on TokenRatesController??

bergeron:
> Loud thoughts, would this be possible to be caused by [price api](https://github.com/MetaMask/core/blob/c1a8b82a6759813afa4620ad2fd11bebd9f106c9/packages/assets-controllers/src/token-prices-service/codefi-v2.ts#L422) as well used on TokenRatesController?

We looked at `TokenRatesController` and it seems to already have the behavior of only updating on success

tommasini (approved):

PR #11132, opened 2024-09-10
feat: Add performance tracing infrastructure

OGPoyraz:
## **Description**

- Add initial tracing and performance metrics infrastructure.
Add the `trace`, `endTrace` methods, to be used anywhere in the mobile code to generate a Sentry traces. Implementation is almost same as we have in the extension. The only difference is in extension `withIsolatedScope`, in mobile `withScope` is used because `@sentry/browser`  version mismatch in `@sentry/react-native`. This is also noted in the `trace.ts` file.
- Add developer options to settings (similar settings [added in extension](https://github.com/MetaMask/metamask-extension/blob/e858167bb1fee33d4d0be1ab2f8e1ce2b11c7945/ui/pages/settings/developer-options-tab/developer-options-tab.tsx#L50) ) 
This adds an easily accessible development action menu where devs wants to test some functionality locally. Specifically this PR adds generating nested traces in Sentry. 
- New environment variables:
`MM_ENABLE_SETTINGS_PAGE_DEV_OPTIONS` : Needs to be set to `true` in order to show `Developer Options` in the settings page
`MM_SENTRY_DSN_DEV` : This is explicitly set for `test-metamask` project and the expected value is in the `.js.env.example`

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1877

## **Manual testing steps**

1. Go to settings page tab `Developer Options`
2. Tab `Generate Trace Test`
3. Open `https://metamask.sentry.io/traces/?project=273496&statsPeriod=15m`, in this link you will be able to see traces that have been send to `test-metamask` in last 15 minutes
4. See `Developer Test` trace is triggered


app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
matthewwalsh0, line 1: This is great to have in mobile and useful for testing the trace support, but would it break up the code review to have it in its own PR?

app/components/Nav/Main/MainNavigator.js
```diff
@@ -312,6 +313,14 @@ const SettingsFlow = () => (
       component={AppInformation}
       options={AppInformation.navigationOptions}
     />
+    {process.env.MM_ENABLE_SETTINGS_PAGE_DEV_OPTIONS === 'true' && (
```
matthewwalsh0, line 316: Is there any kind of config abstraction in mobile, so a single place to reference and parse any environment variables?

app/util/sentry/trace.ts
matthewwalsh0, line 1: I appreciate the `sentry` directory already existed, but part of the benefit of these utility functions was to abstract and encapsulate any usage of Sentry for these to minimise coupling through the codebase.

app/util/sentry/trace.ts
```diff
@@ -0,0 +1,203 @@
...
+  DeveloperTest = 'Developer Test',
+  Middleware = 'Middleware',
+  NestedTest1 = 'Nested Test 1',
+  NestedTest2 = 'Nested Test 2',
+}
+
+const ID_DEFAULT = 'default';
+const OP_DEFAULT = 'custom';
+
+const tracesByKey: Map<string, PendingTrace> = new Map();
```
matthewwalsh0, line 20: This is perhaps less of a problem in the extension since the browser is more likely to be closed, especially in the case of the popup.

But since mobile may have a longer typical lifecycle, is it worth a `setTimeout` when adding to this to delete them after 5 minutes for example to minimise memory usage?

app/util/sentry/trace.ts
```diff
@@ -0,0 +1,203 @@
...
+
+interface PendingTrace {
```
matthewwalsh0, line 22: Is `interface` encouraged in mobile rather than `type`?

app/util/sentry/trace.ts
```diff
@@ -0,0 +1,203 @@
...
+  const duration = endTime - startTime;
+
+  Logger.log('Finished trace', name, id, duration, { request: pendingRequest });
+}
+
+function traceCallback<T>(request: TraceRequest, fn: TraceCallback<T>): T {
+  const { name } = request;
+
+  const callback = (span: Span | undefined) => {
+    Logger.log('Starting trace', name, request);
```
matthewwalsh0, line 88: I'm a big fan of `debug` and `createModuleLogger` for the colour coding and prefix support.

Something worth aligning across both clients and core I think.

app/util/sentry/utils.js
```diff
@@ -502,7 +505,8 @@ export function setupSentry() {
               }),
             ]
           : integrations,
-      tracesSampleRate: 0.04,
+      // Set tracesSampleRate to 1.0, as that ensures that every transaction will be sent to Sentry for development builds.
+      tracesSampleRate: __DEV__ ? 1.0 : 0.04,
```
matthewwalsh0, line 509: Worth nothing that mobile are uses `0.04` while extension uses `0.01`.

I assume a side effect of less traffic generated from mobile?

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { trace, TraceName } from '../../../../util/sentry/trace';
+
+function GenerateTrace() {
+  const theme = useTheme();
+  const { colors } = theme;
+  const styles = createStyles(colors);
+
+  const handleGenerateTraceTest = useCallback(async () => {
+    const sleep = (ms: number) =>
+      new Promise((resolve) => setTimeout(resolve, ms));
```
matthewwalsh0, line 23: This must be the most duplicated logic in every repository.

Is this a util already or should we bite the bullet and make one?

matthewwalsh0 (dismissed):

.js.env.example
```diff
@@ -83,3 +83,10 @@ export GOOGLE_SERVICES_B64=""
 #Notifications Feature Announcements
 export FEATURES_ANNOUNCEMENTS_ACCESS_TOKEN=
 export FEATURES_ANNOUNCEMENTS_SPACE_ID=
+
+# Enables the Settings Page - Developer Options
+export MM_ENABLE_SETTINGS_PAGE_DEV_OPTIONS="true"
+
+# The endpoint used to submit errors and tracing data to Sentry.
+# The below is the `test-metamask` project.
+# export MM_SENTRY_DSN_DEV=https://f59f3dd640d2429d9d0e2445a87ea8e1@sentry.io/273496
```
Cal-L, line 92: Let's replace the value with an empty string in this file

app/components/Nav/Main/MainNavigator.js
```diff
@@ -312,6 +313,14 @@ const SettingsFlow = () => (
...
+      <Stack.Screen
+        name="DeveloperOptions"
```
Cal-L, line 318: Define name in route file

app/components/Views/Settings/DeveloperOptions/DeveloperOptions.types.ts
```diff
@@ -0,0 +1,14 @@
+export interface Props {
```
Cal-L, line 1: Instead of defining the props navigation and route here, you can remove the props and use `useNavigation` hook in the component instead

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
```diff
@@ -0,0 +1,93 @@
...
+  ButtonVariants,
+  ButtonSize,
+  ButtonWidthTypes,
+} from '../../../../component-library/components/Buttons/Button';
+import { trace, TraceName } from '../../../../util/sentry/trace';
+
+function GenerateTrace() {
+  const theme = useTheme();
+  const { colors } = theme;
+  const styles = createStyles(colors);
```
Cal-L, line 19: replace with `useStyles` instead, which already provides theme under the hood

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
```diff
@@ -0,0 +1,93 @@
...
+        style={styles.accessory}
+      />
+    </>
+  );
+}
+
+export default function SentryTest() {
+  const theme = useTheme();
+  const { colors } = theme;
+  const styles = createStyles(colors);
```
Cal-L, line 79: Replace with `useStyles` instead

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  return (
+    <>
+      <Text
+        color={TextColor.Default}
+        variant={TextVariant.HeadingLG}
+        style={styles.heading}
+      >
+        Sentry
```
Cal-L, line 88:
```suggestion
        {'Sentry'}
```

This gives better formatting in IDE

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </Text>
+      {<GenerateTrace />}
```
Cal-L, line 90:
```suggestion
      <GenerateTrace />
```

app/components/Views/Settings/DeveloperOptions/index.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const store = mockStore(initialState);
+
+const setOptions = jest.fn();
+
+describe('DeveloperOptions', () => {
+  it('should render correctly', () => {
+    const wrapper = render(
+      <Provider store={store}>
```
Cal-L, line 27: Use `renderScreen` instead, which already wraps with Providers

app/components/Views/Settings/DeveloperOptions/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Props } from './DeveloperOptions.types';
+import createStyles from './DeveloperOptions.styles';
+import SentryTest from './SentryTest';
+
+const DeveloperOptions = ({ navigation, route }: Props) => {
+  const isFullScreenModal = route?.params?.isFullScreenModal;
+
+  const theme = useTheme();
+  const { colors } = theme;
+  const styles = createStyles(colors);
```
Cal-L, line 16: Use `useStyles` instead

app/util/sentry/utils.js
```diff
@@ -472,13 +472,16 @@ export function deriveSentryEnvironment(
...
+
   // Disable Sentry for E2E tests
-  if (isTest) {
+  if (isTest && !MM_SENTRY_DSN_DEV) {
     return;
   }
 
   const init = async () => {
-    const dsn = process.env.MM_SENTRY_DSN;
+    const dsn = __DEV__ ? MM_SENTRY_DSN_DEV : MM_SENTRY_DSN;
```
Cal-L, line 484: Cleaner to handle this in the build.sh script instead

Cal-L (commented):
Left some comments

app/components/Nav/Main/MainNavigator.js
OGPoyraz: Unfortunately I couldn't find a place to add a utility function for this. If you search `process.env.` these branches are everywhere, we can start one but it wont get notice by other devs. I think it should be handled at once to centralise it.

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
OGPoyraz: I don't think it's adding too much complexity to this PR. Also `SentryTest` is the first content of `DeveloperOptions` shall we keep this one for now?

app/util/sentry/trace.ts
OGPoyraz: Moved up in `app/util` directory and leave the name as is.

app/util/sentry/trace.ts
OGPoyraz: Added in https://github.com/MetaMask/metamask-mobile/pull/11132/commits/f4ce466260d239fccb02a9ea319a55739b55cf8b

app/util/sentry/trace.ts
OGPoyraz: Fixed in https://github.com/MetaMask/metamask-mobile/pull/11132/commits/e0026992f4db086977fb61bb3e9cb7f26f0d23d7

app/util/sentry/trace.ts
OGPoyraz: Done in https://github.com/MetaMask/metamask-mobile/pull/11132/commits/308f2e6497196e5a99b93c148b93c8aa3503294f

app/util/sentry/utils.js
OGPoyraz: It's probably that but not sure why is this 1/4 ratio.

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
OGPoyraz: Relocated this into very empty `app/util/testUtils` 😄 https://github.com/MetaMask/metamask-mobile/pull/11132/commits/9ed584db2ec6f77b61baa9b6dd4f375a24eeb5a4

.js.env.example
OGPoyraz: Is there any downside/reason leaving as is? Because we already show this value in extension example environments as well

app/components/Views/Settings/DeveloperOptions/index.tsx
OGPoyraz: Done

app/components/Views/Settings/DeveloperOptions/index.test.tsx
OGPoyraz: Done

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
OGPoyraz: Done

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
OGPoyraz: Done

app/components/Views/Settings/DeveloperOptions/SentryTest.tsx
OGPoyraz: Done

app/components/Views/Settings/DeveloperOptions/DeveloperOptions.types.ts
OGPoyraz: Done

app/components/Nav/Main/MainNavigator.js
OGPoyraz: Done

app/util/sentry/utils.js
OGPoyraz: Do you mean setting single `SENTRY_DSN `? Please also note that `MM_SENTRY_DSN_DEV` is needed in this context because I believe `isTest` is not working as intended.

.js.env.example
Cal-L: Since this is a mm dev is environment, we wouldn't want non mm contributors reporting to the same DSN. They should be using their own

app/util/sentry/utils.js
Cal-L: Yep, we can conditionally map the sentry dsn to the dev one in debug mode from the build.sh file. You might not need to couple it with isTest since that's used for e2e.

.js.env.example
OGPoyraz: Regardless of this PR it's already in public in extension, that's what I mean. I don't mind removing it for sure.

OGPoyraz: https://github.com/MetaMask/metamask-mobile/pull/11132/commits/3481ff35b23881fc5d23ea13ac1c5091082e411b

app/util/sentry/utils.js
OGPoyraz: Well that's the issue, `isTest` is returning `true` in dev environments unless we don't set `METAMASK_ENVIRONMENT=production`. 

I'm cautious about conditionally modifying `process.env` values, especially outside of test environments. This approach can lead to unpredictable behavior and make it harder to debug issues. Unless there's a specific requirement or a well-established convention, I think it's best to avoid dynamically changing environment variables. In our case, I believe it's more maintainable and predictable to keep the Sentry DSN configuration separate from the build process. If we need to switch between dev and prod DSNs, we can achieve this through other means, such as using separate configuration files or environment-specific settings.

Cal-L: I'm not sure about the issue with the `isTest` but let's keep it simple and just consider the debug environment for now. We can handle E2E conditions in another PR, preferably when the `isTest` is fixed. I see what you mean on how it could be harder to debug when the mechanism that handles setting the env vars is abstracted from the main code. Although the build.sh script may introduce complexities, the project was set up in a way to be reliant on it. At the same time, by using the intended script commands, the build.sh is also intended to produce predictable behavior. For example, one would expect that running `yarn start:ios:qa` would start a build using QA environment setup. There are definitely alternatives to setting up these environments such as creating multiple variants. However as it's set up today, the non-documented approach is to map the variables within each variant in `buildIos` in the build.sh script. For your changes, it would mean mapping the env vars in `buildIosSimulator` by adding a mapping function. You can reference `buildIosFlaskRelease` as a point of reference. The end result is: provided that you have the dev sentry DSN defined in your .js.env, running `yarn start:ios` should pick up the dev value.

OGPoyraz: @Cal-L I appreciate your clarification. While I'm not entirely opposed to this approach, I do worry that the growing number of small differences between our mobile and extension repositories is making life harder for developers. I understand that these are two distinct clients with varying requirements, but I also believe that we should strive for common conventions wherever possible, in that case it's not. 

As you pointed out that this won't be the first example done in mobile, I'll implement the changes as you've suggested.

https://github.com/MetaMask/metamask-mobile/pull/11132/commits/3cc85dc17eb387fa4e77da1d60883fc5b99679f3

Cal-L (dismissed):
Nice, LGTM. Feel free to reach out for another approval once the conflicts are fixed.

app/util/trace.ts
```diff
@@ -0,0 +1,217 @@
...
+  } finally {
+    if (!isPromise) {
+      finallyFn();
+    }
+  }
+
+  return undefined;
+}
+
+export function cleanupTraces() {
```
matthewwalsh0, line 210: Is this called anywhere yet or just available?

My assumption was we could handle this automatically by triggering an individual `setTimeout` for each one within `startTrace` and then store the timeout inside `PendingTrace` so we can call `clearTimeout` within `endTrace`?

app/util/trace.ts
```diff
@@ -0,0 +1,217 @@
...
+  Middleware = 'Middleware',
+  NestedTest1 = 'Nested Test 1',
+  NestedTest2 = 'Nested Test 2',
+}
+
+const ID_DEFAULT = 'default';
+const OP_DEFAULT = 'custom';
+export const TRACES_CLEANUP_INTERVAL = 5 * 60 * 1000; // 5 minutes
+
+export const tracesByKey: Map<string, PendingTrace> = new Map();
```
matthewwalsh0, line 25: It's difficult, but I'm a big fan of avoiding exporting variables purely for the sake of tests.

Ideally, any unit tests would just verify the side effects such as `end` being called on the `span` or a `clearTimeout` being called?

matthewwalsh0 (changes requested):

app/util/trace.ts
OGPoyraz: Changed the approach here: https://github.com/MetaMask/metamask-mobile/pull/11132/commits/69aa635b6986eb32c026e7e72d27680095a82b4a

app/util/trace.ts
OGPoyraz: That make sense, I've changed the tests to not use `tracesByKey`

matthewwalsh0 (approved):

vinistevam (approved):

leotm (approved):
lgtm

Screenshot 2024-09-18 at 6 52 31 PM

perhaps like https://github.com/MetaMask/metamask-extension/blob/develop/development/README.md#sentry

we should add a Sentry section here https://github.com/MetaMask/metamask-mobile/blob/main/docs/readme/testing.md

with heading e.g. `Generating a trace test` that mentions our new env vars, with the odd screenshot or two if it helps

PR #11144, opened 2024-09-10
test: E2E Mocking Setup For Detox Tests

Andepande:
Consolidated Mocking Research into Current Branch

## **Description**

This PR introduces the initial E2E API mocking test. When IsTest is true, the iOS/Android app redirects traffic to a proxy within the e2e folder for tests that require mocking. The mock server proxy forwards all traffic to its destination unless it matches specific mock requirements. The first test verifies that the suggestedGasApi falls back to the legacy endpoint when the EIP1559 API is down, displaying a legacy modal. The test then reinstates the service and asserts the modal change immediately.

## **Related issues**

Fixes:https://github.com/orgs/MetaMask/projects/60/views/16?pane=issue&itemId=70977379

## **Manual testing steps**

1. Run e2e Detox setup 
2. Have IOS and Android sim ready
3. For IOS run command `yarn detox test -c ios.sim.debug e2e/specs/wallet/suggestedGasApi.mock.spec.js;` for Android
run `yarn detox test -c android.emu.debug e2e/specs/wallet/suggestedGasApi.mock.spec.js`

### **After**

https://www.loom.com/share/64d322a3f50d4888b1b7a96107008918
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/ef9aaa7c-6247-4bf9-81fd-09bbaa76d7c8


e2e/pages/AccountListView.js
```diff
@@ -74,6 +75,7 @@ class AccountListView {
 
   async swipeToDismissAccountsModal() {
     await Gestures.swipe(this.title, 'down', 'fast', 0.6);
+    await Assertions.checkIfNotVisible(this.title);
```
cortisiko, line 78: is this needed? Assertions should live in the test file not the page object

e2e/specs/wallet/suggestedGasApi.mock.spec.js
```diff
@@ -0,0 +1,88 @@
...
+        await stopMockServer(); //stop mock server to reinstate suggested gas api service
+        await TestHelpers.delay(20000);
+        await Assertions.checkIfVisible(
+          TransactionConfirmView.editPriorityFeeSheetContainer,
+        );
+      },
+    );
+  });
+// Because we stop the server within the test, a try catch block here would stop the server if the test fails midway
+  afterAll(async () => {
```
cortisiko, line 78: it is good practice to put the hooks together (before hook followed by after all hook

e2e/specs/wallet/suggestedGasApi.mock.spec.js
```diff
@@ -0,0 +1,88 @@
...
+import TestHelpers from '../../helpers.js';
+
+describe(SmokeCore('Mock suggestedGasApi fallback to legacy gas endpoint  when EIP1559 endpoint is down'), () => {
+  let mockServer;
+  beforeAll(async () => {
+    jest.setTimeout(150000);
+    await TestHelpers.reverseServerPort();
+
+    mockServer = await startMockServer({ // Configure mock server
+      mockUrl: 'https://gas.api.cx.metamask.io/networks/1/suggestedGasFees',
```
cortisiko, line 27: perhaps have some data structure to store endpoints and then access the data in the test.

e2e/pages/Send/TransactionConfirmView.js
```diff
@@ -79,10 +85,14 @@ class TransactionConfirmationView {
     await Gestures.waitAndTap(this.cancelButton);
   }
 
-  async tapEstimatedGasLink() {
+  async tapEstimatedGasLink(index = 0) {
     await Gestures.swipe(this.transactionAmount, 'up', 'fast');
 
-    await Gestures.waitAndTap(this.estimatedGasLink);
+    if (index === 0) {
```
cortisiko, line 91: Mind explaining the conditional?

cortisiko (changes requested):
Good work on this @Andepande and @SamuelSalas. I left comments

e2e/pages/Send/TransactionConfirmView.js
SamuelSalas: This is because some issues we're having some issues with Detox on the transaction screen with the existing test and this new mock test. So i added this conditional so it work for both tests.

e2e/pages/AccountListView.js
SamuelSalas: Right now is no longer needed because we added a conditional to check if the platform is ios or android. We're having some issues with Android trying to tap the Actions button before the Accounts modal was closed.

shim.js
```diff
@@ -56,3 +56,28 @@ if (typeof localStorage !== 'undefined') {
 // If using the crypto shim, uncomment the following line to ensure
 // crypto is loaded first, so it can populate global.crypto
 // require('crypto')
+
+if (isTest) {
+  (async () => {
+    const { fetch: originalFetch } = global;
+    const MOCKTTP_URL = `http://${
+      // eslint-disable-next-line no-undef
+      Platform.OS === 'ios' ? 'localhost' : '10.0.2.2' // l
```
NicolasMassart, line 65: Suggestions:
- the `eslint-disable-next-line no-undef` to prevent global `Platform` to raise error may be replaced by a file level `/* global Platform */`
- remove or fix comment `// l`

shim.js
```diff
@@ -56,3 +56,28 @@ if (typeof localStorage !== 'undefined') {
...
+    }:8000`;
+
+    const isMockServerAvailable = await originalFetch(
+      // small healthcheck
```
NicolasMassart, line 69: Not useful comment. Small gives no real information and the path `/health-check` already says what it is.
```suggestion
```

e2e/specs/wallet/suggestedGasApi.mock.spec.js
```diff
@@ -0,0 +1,90 @@
...
+        await SendView.inputAddress(RECIPIENT);
+        await SendView.tapNextButton();
+        await AmountView.typeInTransactionAmount(AMOUNT);
+        await AmountView.tapNextButton();
+        await TransactionConfirmView.tapEstimatedGasLink(1);
+        await Assertions.checkIfVisible(
+          TransactionConfirmView.editPriorityModal,
+        );
+        await stopMockServer(); //stop mock server to reinstate suggested gas api service
+        await TestHelpers.delay(20000);
```
NicolasMassart, line 83: Why 20s pause?

e2e/mockserver/mockserver.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-console */
+import { getLocal } from 'mockttp';
+
+const mockServer = getLocal();
+
+export const startMockServer = async ({
+  mockUrl,
+  responseCode = 500,
+  responseBody = {},
+  port = 8000,
```
NicolasMassart, line 10: S: export a `const defaultPort=8000` to use here as default of the func and in shim.js too.

e2e/mockserver/mockserver.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-console */
+import { getLocal } from 'mockttp';
+
+const mockServer = getLocal();
+
+export const startMockServer = async ({
```
NicolasMassart, line 6: Q: should we also handle response headers? I guess we have some request with headers in the response.
We can default to no headers `{}`.

shim.js
```diff
@@ -56,3 +56,28 @@ if (typeof localStorage !== 'undefined') {
...
+      `${MOCKTTP_URL}/health-check`,
+    )
+      .then((res) => res.ok)
+      .catch(() => false);
+    // if mockserver is off we route to original destination
+    global.fetch = async (url, options) =>
```
NicolasMassart, line 75: Q: what about requests not using fetch? Is this mocking in case of Axios is used?

NicolasMassart (commented):
I like this mock!
Added some suggestions and questions.

e2e/specs/wallet/suggestedGasApi.mock.spec.js
SamuelSalas: The main reason is that we encounter an error with Detox when the app change the legacy suggested gas fee modal to the new  EIP1559 modal. So the pause is just to give the app enough time to change modals and continue with the testing.

e2e/mockserver/mockserver.js
SamuelSalas: We could handle the response headers, but in this case the main objective of the pr was to push a POC for how the mockserver would be implemented on the detox test. So we just wanted to show one scenario were the behavior of the app changed by mocking one response. We didn't want to add more than needed.

shim.js
Andepande: A: We experimented with several methods for rerouting network traffic and found that this approach is a lightweight solution that effectively routes most requests to the mock server for our use case. However, if we encounter situations where axios or other custom request methods are not intercepted, we can modify the code to better handle these cases or introduce conditional logic to accommodate different types of requests.

e2e/specs/wallet/suggestedGasApi.mock.spec.js
NicolasMassart: Could we await something instead of having a fixed 20s pause?

shim.js
```diff
@@ -60,13 +61,12 @@ if (typeof localStorage !== 'undefined') {
 if (isTest) {
   (async () => {
     const { fetch: originalFetch } = global;
+    const defaultMockPort = 8000;
```
NicolasMassart, line 64: C: my point was to reuse port const from e2e/mockserver/mockserver.js here too

e2e/specs/wallet/suggestedGasApi.mock.spec.js
NicolasMassart: Timeout is better. Thanks!

NicolasMassart (dismissed):
Looks good to me. Great work!

cortisiko (approved):

SamuelSalas (approved):

Issue #11072, opened 2024-09-05
[Bug]: "Hide Tokens without Balance" toggle does not hide tokens with 0 balance on the main wallet view


Andepande:
### Describe the bug

When the "Hide Tokens without Balance" toggle is enabled, tokens with a 0 balance are still visible in the main wallet view. 

### Expected behavior

 Tokens with a 0 balance should be hidden on the main wallet view 

### Screenshots/Recordings

<img width="472" alt="image" src="https://github.com/user-attachments/assets/4e0cadc9-765e-4e96-9e3f-5759553fb5c3">

![PHOTO-2024-09-12-10-47-49](https://github.com/user-attachments/assets/34186846-e3f3-487b-9977-a0d11eb2a203)

### Steps to reproduce

1. Open the wallet.
2. Ensure there are tokens with a 0 balance visible in the main wallet view.
3. Enable the "Hide Tokens without Balance" toggle.
4. Observe that tokens with a 0 balance are still shown.

### Detection stage

In production (default)

### Version

7.30.0

### Build type

Other (please specify exactly where you obtained this build in "Additional Context" section)

### Device

IOS & Android

### Operating system

iOS, Android

### Severity

sev2

PR #11172, opened 2024-09-12
fix: fix check token balance is zero

sahar-fehri:
## **Description**

When the setting to hide tokens when is zero is ON, we should not be able to see zero balances on tokens tab.

## **Related issues**

Fixes:
https://github.com/MetaMask/metamask-mobile/issues/11072
https://github.com/MetaMask/mobile-planning/issues/1869

## **Manual testing steps**

1. Click on setting => General and enable setting: Hide tokens without balance
2. Go to tokens tab and import any token that you do not have balance for
3. Check that you do not see the imported token on the main page

### **Before**

https://github.com/user-attachments/assets/1d18de27-90ca-4e10-9ca6-9b834d94a323

### **After**

https://github.com/user-attachments/assets/1c58346b-45a6-4b31-ab26-f82795d6aeda


app/util/lodash/index.ts
```diff
@@ -7,7 +7,11 @@
 // TODO: Replace "any" with type
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const isZero = (value: number | any): boolean => {
-  if (value?.toString && value?.toString?.() === '0') {
+
+  if (
+    value?.toString &&
+    (value?.toString?.() === '0' || value?.toString?.() === '0x0')
+  ) {
```
sahar-fehri, line 14: The token balances we get from `useTokenBalancesController` are in hex format; when the balance is zero it would be "0x0" not "0"; 
Was not sure what was the reason behind doing the "0" check; so i did not remove it and added the check for "0x0"

tommasini: Hmm, maybe because the old api request was answering 0 and this was an issue/ tech debt
Since I'm no t certain as well I guess maybe it's better than safe than sorry fi we do not see anything be impacted

tommasini: It would be awesome if we could understand if 0 is one possibility

tommasini (approved):

PR #11191, opened 2024-09-13
temp: add `Transaction` trace implementation

OGPoyraz:
## **Description**

- Adding `createTracingMiddleware` to start tracing for defined type of messages.
- Adding `trace` callback to `TransactionController`.
- End `Notification Display` trace after confirmation landed on users screen. (We don't name `Notification` for user confirmations in mobile, but the since the trace name is `Notification Display` in controller, we will leave it for now and refactor naming later)

Note on patch: 
The patch applies single commit to the current patch PR: https://github.com/MetaMask/core/pull/4692/files

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1876

## **Manual testing steps**

1. Set `MM_SENTRY_DSN_DEV` to developer Sentry dsn, see the value here https://github.com/MetaMask/metamask-extension/blob/develop/.metamaskrc.dist
2. Go test-dapp in app, trigger and accept test transaction
3. If you used the mentioned dsn above, your `Transaction` trace will appear on this link: https://metamask.sentry.io/traces/?project=273496&query=transaction%3ATransaction&source=traces&statsPeriod=1h

## **Screenshots/Recordings**

Example trace: https://metamask.sentry.io/performance/trace/1c4b2cfbe13640448ff5ac33feb13692/?node=txn-a3464e9139ad47c5bed4adbcfc756920&pageEnd&pageStart&project=273496&query=transaction%3ATransaction&source=traces&statsPeriod=1h&timestamp=1726219598.426


app/components/Views/confirmations/components/TransactionReview/index.js
```diff
@@ -361,6 +362,8 @@ class TransactionReview extends PureComponent {
         POLLING_INTERVAL_ESTIMATED_L1_FEE,
       );
     }
+
+    endTrace({ name: TraceName.NotificationDisplay, id: transaction.id });
```
OGPoyraz, line 366: We don't name Notification for user confirmations in mobile, but the since the trace name is Notification Display in controller, we will leave it for now and refactor naming later

app/core/createTracingMiddleware.test.ts
Cal-L, line 1: Let's place this and the original file in a folder named `createTracingMiddleware`

app/core/createTracingMiddleware.ts
```diff
@@ -0,0 +1,43 @@
+// Request and repsones are currently untyped.
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Cal-L, line 2: Let's remove the TS ignore and type the req and res. You can reference `eth_sendTransaction` on what that looks like

app/core/RPCMethods/eth_sendTransaction.ts
```diff
@@ -66,14 +67,19 @@ async function eth_sendTransaction({
   validateAccountAndChainId,
 }: {
   hostname: string;
-  req: JsonRpcRequest<unknown> & { method: 'eth_sendTransaction' };
+  req: JsonRpcRequest<unknown> & {
+    method: 'eth_sendTransaction';
+    traceContext?: unknown;
```
Cal-L, line 72: Does Sentry provide a type for this?

app/core/createTracingMiddleware.test.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
Cal-L, line 1: Let's type this file if we can

Cal-L (commented):
left some comments

app/components/Views/confirmations/components/TransactionReview/index.js
matthewwalsh0, line 366: Should we do this at the start of the function as there is some `await` calls before this point that may delay the timing?

app/core/RPCMethods/eth_sendTransaction.ts
matthewwalsh0: Technically it does, but part of the goal of the `trace` utility functions is to abstract the use of `Sentry` to reduce coupling.

But while there should be no need to manipulate or inspect the value itself, we do still define a `TraceContext` type equal to `unknown` within `trace.ts`, in case we update the type in future.

app/core/RPCMethods/eth_sendTransaction.ts
```diff
@@ -93,14 +99,22 @@ async function eth_sendTransaction({
     chainId: req.params[0].chainId,
   });
 
+  endTrace({ name: TraceName.Middleware, id: actionId });
+
   const { result, transactionMeta } = await sendTransaction(req.params[0], {
+    actionId: req.id as string,
```
matthewwalsh0, line 105: Is this related to this PR? Why do we need to add action ID support here?

app/core/createTracingMiddleware.ts
```diff
@@ -0,0 +1,43 @@
...
+    const traceName = METHOD_TYPE_TO_TRACE_NAME[method];
+
+    if (traceName) {
+      req.traceContext = await trace({
+        name: traceName,
+        id,
+        tags: METHOD_TYPE_TO_TAGS[method],
+      });
+
+      await trace({
```
matthewwalsh0, line 34: Do these two need `await` since they are just starting traces which should be synchronous?

matthewwalsh0 (changes requested):

Issue #10784, opened 2024-08-23
[EPIC] refactor metametrics


NicolasMassart:
### What is this about?

As MetaMetrics system evolved, we stacked multiple ways to define events. We even have a Legacy section in the events list.
Even on the new events, there's multiple ways to track events because they are structured differently. This forced us to build "conversion" functions.

### Technical Details

- [x] #11317
- [ ] #11570
- [ ] #11568
- [ ] #11569
- [x] align mobile and extension metrics use([1](#note-1))  
- [x] #11319
- [x] #11219
- [x] #11065
- [x] #12469
- [ ] #12468
- [x] #12728
- [x] #12688
- [x] #12913
- [x] #13204
- [x] https://github.com/MetaMask/mobile-planning/issues/2150

> [!NOTE]
> <span id="note-1">1: </span>[research done](https://github.com/MetaMask/mobile-planning/issues/1905), decision is not to align as none is an acceptable target, but refactor mobile the best way as possible and later align extension on it

### Threat Modeling Framework

What are we working on?

- event tracking cleanup and alignment with extension

What can go wrong?

- too many events tracked (not enough cleanup)
- not enough events tracked (too much cleanup)
- data tracked when they should not be (privacy)
- data analytics broken (dashboards and data store not updated accordingly)

What are we going to do about it?

- sync with data team and PMs of each feature (ext and mobile)
- clearly define the Segment tracking plan
- write unit tests
- review code and QA test all critical features

Did we do a good job?

- so far yes, but we will do even better

### Acceptance Criteria

- [ ] events list clarified and simplified, consistent and cleaned our of legacy
- [ ] all changes tested and test passes
- [ ] events properly received and usable on the analytics side (Segment/Mixpanel)

### Stakeholder review needed before the work gets merged

- [X] Engineering (needed in most cases)
- [ ] Design
- [X] Product
- [X] QA (automation tests are required to pass before merging PRs but not all changes are covered by automation tests - please review if QA is needed beyond automation tests)
- [ ] Security
- [ ] Legal
- [ ] Marketing
- [ ] Management (please specify)
- [ ] Other (please specify)

NicolasMassart:
See also https://github.com/MetaMask/metamask-mobile/pull/10545#issuecomment-2310285135

gauthierpetetin:
Linking this ticket here as well, to make sure that we adopt a similar approach in Extension and Mobile, whatever we agree on.
[...]

PR #11262, opened 2024-09-17
chore: Refactor event tracking method

NicolasMassart:
## **Description**

- introduces a tracking event [Builder](https://refactoring.guru/design-patterns/builder) to help dev on creation of complex events
- removes dependency to Segment SDK JsonMap by implementing our own
- updates to useMetrics hook
- deprecates legacy trackEvent function signature
- updates app/components/UI/NetworkModal/index.tsx file using the builder instead of trackEvent as an example
- updates unit tests that were checking for hook trackEvent params: the hook is now just a bypass and does no processing so metametrics trackEvent is called with same params.
- updates code doc

> [!NOTE]
> These changes are made in order to preserve existing mechanisms in parallel.
> This leads to strange naming like ITrackingEvent as the ideal IMetaMetricsEvent already exists.
> This can be dealt with later as it's well documented and following guidelines.

## **Related issues**

fixes #11319

## **Manual testing steps**

NA

## **Screenshots/Recordings**

NA

### **Before**

NA

### **After**

NA


wachunei:
With @NicolasMassart we had a call and aligned on the API this Event Builder can have, here are some examples of the expected behavior:

```ts

// Event Builder interface
const eventBuilder = createEventBuilder(EVENT_NAME)
  .addProperties({ ...})
  .addSensitiveProperties({ ... })
  .addProperties({ ... }) // These can be added multiple times

if (condition) {
  eventBuilder.addProperties({ ... })
  eventBuilder.removeProperties(['prop1', 'prop2'])
  // or inline
  // eventBuilder.addProperties({ ... }).removeProperties(['prop1', 'prop2'])
}

const event = eventBuilder.build();

// Suggested way to track this event
// We pass the built event to the trackEvent function
trackEvent(event);

// or inline
trackEvent(createEventBuilder(EVENT).addProperties({ ... }).build());

// And backwards compatible (for now?? deprecate it in the future).
trackEvent({ name: EVENT, properties: { ... }, sensitiveProperties: { ... } });

// What if I want to modify an already built event?
// We can revert it to a Builder state if we create a new builder from the event
createEventBuilder(event).addProperties({ ... }).build();

// ———

// Another way (initial suggestion) was to have a track method in the event object
// THIS WILL NOT BE SUPPORTED
event.track();

```

If there's a use case we detect, in the future the builder can be extended to support new features, for example:
```ts
eventBuilder.makeSensitive(['prop1', 'prop2']);
eventBuilder.makeNonSensitive(['prop1', 'prop2']);
```

app/components/hooks/useMetrics/useMetrics.ts
```diff
@@ -70,57 +68,19 @@ import { CombinedProperties } from '../../../core/Analytics/MetaMetrics.types';
...
-   * @example track with properties:
-   * const { trackEvent } = useMetrics();
-   * trackEvent(MetaMetricsEvents.BROWSER_SEARCH_USED, {
-   *       option_chosen: 'Browser Bottom Bar Menu',
-   *       number_of_tabs: undefined,
-   *     });
-   *
-   * @see MetaMetrics.trackEvent
-   */
-  const trackEvent = useCallback(
```
NicolasMassart, line 94: C: removing this to directly pass the metametrics function as the hook function.
The InteractionManager.runAfterInteractions here is useless given how the Segment SDK is implemented. Making it more simple.

app/components/hooks/useMetrics/useMetrics.types.ts
```diff
@@ -19,15 +23,30 @@ export interface IUseMetricsHook {
   isEnabled(): boolean;
   enable(enable?: boolean): Promise<void>;
   addTraitsToUser(userTraits: UserTraits): Promise<void>;
+
+  /**
+   * @deprecated use {@link trackEvent(ITrackingEvent, boolean)}
```
NicolasMassart, line 28: C: deprecating only this signature in favour of the new one that uses ITrackingEvent.

app/core/Analytics/MetaMetrics.ts
```diff
@@ -336,14 +339,14 @@ class MetaMetrics implements IMetaMetrics {
...
-   * Track an analytics event
+   * Send an analytics event to the Segment SDK track function
    *
    * @param event - Analytics event name
    * @param properties - Object containing any event relevant traits or properties (optional)
    * @param saveDataRecording - param to skip saving the data recording flag (optional)
    * @see https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/#track
    */
-  #trackEvent = (
+  #trackWithSdkClient = (
```
NicolasMassart, line 349: C: renaming function for more clarity

app/util/metrics/TrackError/trackErrorAsAnalytics.ts
```diff
@@ -6,7 +6,7 @@ import { InteractionManager } from 'react-native';
  * The goal is to log errors (that are not errors from our side) like “Invalid Password”.
  * An error like this generally means a user inserted the wrong password, so logging to sentry doesn't make sense.
  * But we still want to log this to analytics to be aware of a rapid increase which may mean it's an error from our side, for example, an error with the encryption library.
- * @param event the original event to track
```
NicolasMassart, line 9: This is just fixing the param name that was wrong, a bit out of scope but still metrics related and I was updating all comments.

app/core/Analytics/MetricsEventBuilder.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * represents the event tracking object produced by the event builder
+ */
+class TrackingEvent implements ITrackingEvent {
+  readonly #name: string;
+  #properties: JsonMap;
+  #sensitiveProperties: JsonMap;
+  #saveDataRecording: boolean;
+
+  constructor(event: IMetaMetricsEvent | ITrackingEvent) {
```
gambinish, line 17: Does the `IMetaMetricsEvent` interface need to be expanded to also contain an optional `sensitiveProperties` key?

app/components/UI/NetworkModal/index.tsx
```diff
@@ -106,26 +106,32 @@ const NetworkModals = (props: NetworkProps) => {
...
+      trackEvent(
+        createEventBuilder(MetaMetricsEvents.NETWORK_ADDED)
+          .addProperties({
+            chain_id: toHex(safeChain.chainId),
+            source: { anonymous: true, value: 'Custom Network Added' },
+            symbol: safeChain.nativeCurrency.symbol,
+          })
+          .addSensitiveProperties({ rpcUrl: safeRPCUrl })
+          .build(),
+      );
```
gambinish, line 134: I generally would prefer to call `createEventBuilder` separately, assign it's return value to a variable (e.g., `const complexEvent = createEventBuilder(...)`), and then pass that variable into `trackEvent`, rather than passing the function directly. This might help with readability.

Additionally, I wonder if we could wrap the existing `trackEvent` in a higher-order function (e.g., `trackComplexEvent`) to abstract some of these implementation details. This could help centralize the callback chain, so it doesn’t need to be repeated every time this pattern is used.

app/core/Analytics/MetricsEventBuilder.ts
```diff
@@ -0,0 +1,171 @@
...
+    this.#trackingEvent = new TrackingEvent(event);
+  }
+
+  /**
+   * Create a new event builder
+   * @param event the event for the builder to build
+   */
+  static createEventBuilder(event: IMetaMetricsEvent | ITrackingEvent) {
```
gambinish, line 101: Nit: Why did we decide on `createEventBuilder` here instead of `eventBuilder` or maybe `metricsEventBuilder`?

app/core/Analytics/MetricsEventBuilder.ts
NicolasMassart: No, it's the legacy one, it has to stay compatible with old codebase and so should not include this prop. It will be even completely removed in the future but it requires to go through all the codebase to replace the usage by new usage. Maybe something Devin AI can do.

app/components/UI/NetworkModal/index.tsx
NicolasMassart: Both call styles are possible. it's up to the dev and maybe future guidelines.

One place where you would like to build first in a variable is if you don't use the event instantly or add properties progressively. That's why both ways are possible.
But when it's creating the event and tracking it all in a row, I think inlining everything is simpler and not harder to read. I wish all the codebase was as easy to read tbh...

app/core/Analytics/MetricsEventBuilder.ts
NicolasMassart: It's the [factory pattern](https://refactoring.guru/design-patterns/factory-method).
You add a static function to your object to allow creating the object without having to call the constructor with `new`: only the object now has to know how to build itself.

The object instance is implementing the [builder pattern](https://refactoring.guru/design-patterns/builder).

`create[ObjectName]` is a standardised naming for this factory pattern.

app/core/Analytics/MetricsEventBuilder.ts
NicolasMassart: This PR is part of an epic #10784 and so removal should be done in a next issue #11569

app/components/UI/NetworkModal/index.tsx
gambinish: Ahh I see, yeah that makes sense 👍 

Autocomplete does make things easier. When I started I also didn't know about sensitive properties.

gambinish (approved):
LGTM! Thanks for those explanations.

Fyi I am not a codeowner on this, so looks like you'll still need to get that as well.

wachunei (approved):
Looks good to me

EtherWizard33 (approved):

PR #11318, opened 2024-09-19
feat: app event manager and attribution id parameters

abretonc7s:
## **Description**

When a user is redirected to MetaMask Mobile app thanks to a deep link, including an attributionId parameter:

metamask://open.browser/website_url?attributionId=xyz
(deep link above is just a placeholder)

Then MM Mobile app should attach attributionId as a property to an event emitted right after Mobile app opens.
Given Mobile app would open and directly redirect the user to the in-app browser, the "App Opened" event might be a good candidate.

## **Related issues**

Files: [jira issue: ](https://id.atlassian.com/login?continue=https%3A%2F%2Fid.atlassian.com%2Fjoin%2Fuser-access%3Fresource%3Dari%253Acloud%253Ajira%253A%253Asite%252F8831f492-a12c-460e-ac1b-400f1b09e935%26continue%3Dhttps%253A%252F%252Fconsensyssoftware.atlassian.net%252Fbrowse%252FSDK-18%253FatlOrigin%253DeyJpIjoiNzI2OWEwNTczYTZmNGEyZjgxYmFjYjkxMzJlZmE4MTYiLCJwIjoiaiJ9&application=jira)

## **Manual testing steps**

1. Open deeplink container url and check event on segment


app/core/DeeplinkManager/ParseManager/parseDeeplink.ts
```diff
@@ -41,6 +44,19 @@ function parseDeeplink({
...
+    const isDataCollectionForMarketingEnabled = security.dataCollectionForMarketing;
+    DevLogger.log(`DeepLinkManager:parse isDataCollectionForMarketingEnabled=${isDataCollectionForMarketingEnabled} attributionId=${attributionId}`);
+
+    if(attributionId && isDataCollectionForMarketingEnabled) {
+      MetaMetrics.getInstance().trackEvent(
+          MetaMetricsEvents.APP_OPENED,
+          { attributionId },
+          true,
+        );
+    }
```
gauthierpetetin, line 59: I see `App Opened` event is emitted in `parseDeeplink` function.
This means it will only be emitted when App is opened through deep link, right?

If so, we might want to update the location of it, because we want it to be emitted every time the App is opened (via deep link but also without deep link). More detail in @Cal-L 's [comment](https://consensys.slack.com/archives/C07EJHAT6JX/p1726259653287669) on Slack.

abretonc7s: but everything goes threw parseDeeplinks, even the qrcode scanning

gauthierpetetin: If I presse the app icon on my phone and the app opens, will it also go through `parseDeeplink` function?

abretonc7s: this works on hot or cold start.

gauthierpetetin:
Adding release-7.32.0 on this PR to make sure we cherry-pick it. cc @Cal-L @sethkfman for visibility

app/core/Analytics/MetaMetrics.events.ts
```diff
@@ -36,6 +36,9 @@ const ONBOARDING_WIZARD_STEP_DESCRIPTION: { [key: number]: string } = {
  * Analytics Tracking Events
  */
 enum EVENT_NAME {
+  // App
+  APP_OPENED = 'App Opened',
```
NicolasMassart, line 40: Q: would this be different from DAPP_VIEWED ?

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      return { remove: jest.fn() };
+    });
+    appStateManager = new AppStateManager();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('should subscribe to AppState changes on instantiation', () => {
```
NicolasMassart, line 59:
```suggestion
  it('subscribes to AppState changes on instantiation', () => {
```

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    expect(AppState.addEventListener).toHaveBeenCalledWith('change', expect.any(Function));
+  });
+
+  it('should track event when app becomes active and conditions are met', () => {
```
NicolasMassart, line 63: Q: what conditions exactly?
```suggestion
  it('tracks event when app becomes active and [THE ACTUAL CONDITIONS]', () => {
```

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    jest.advanceTimersByTime(2000);
+
+    expect(mockTrackEvent).toHaveBeenCalledWith(
+      MetaMetricsEvents.APP_OPENED,
+      { attributionId: 'test123' },
+      true
+    );
+  });
+
+  it('should not track event when data collection is disabled', () => {
```
NicolasMassart, line 80: Q: what are you testing exactly here?
The assert shows the track event is called.
If you are testing that no actual event is sent to the SDK when metrics are not optin, this is already done on the metametrics test level, no need to repeat it here.

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    jest.advanceTimersByTime(2000);
+
+    expect(mockTrackEvent).toHaveBeenCalledWith(
+      MetaMetricsEvents.APP_OPENED,
+      { attributionId: undefined },
+      true
+    );
+  });
+
+  it('should not track event when there is no deeplink', () => {
```
NicolasMassart, line 95: Q: what does not tracking mean here? I can read the assert toHaveBeenCalledWith so it means trackEvent is called.

Also naming guidelines:
```suggestion
  it('does not track event when there is no deeplink', () => {
```

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    jest.advanceTimersByTime(2000);
+
+    expect(mockTrackEvent).toHaveBeenCalledWith(
+      MetaMetricsEvents.APP_OPENED,
+      { attributionId: undefined },
+      true
+    );
+  });
+
+  it('should handle errors gracefully', () => {
```
NicolasMassart, line 110:
```suggestion
  it('handles app state change errors', () => {
```

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    jest.advanceTimersByTime(2000);
+
+    expect(Logger.error).toHaveBeenCalledWith(
+      expect.any(Error),
+      'AppStateManager: Error processing app state change'
+    );
+    expect(mockTrackEvent).not.toHaveBeenCalled();
+  });
+
+  it('should clean up the AppState listener on cleanup', () => {
```
NicolasMassart, line 125:
```suggestion
  it('cleans up the AppState listener', () => {
```

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    const mockRemove = jest.fn();
+    (AppState.addEventListener as jest.Mock).mockReturnValue({ remove: mockRemove });
+
+    appStateManager = new AppStateManager();
+    appStateManager.cleanup();
+
+    expect(mockRemove).toHaveBeenCalled();
+  });
+
+  it('should not process app state change when app is not becoming active', () => {
```
NicolasMassart, line 135:
```suggestion
  it('does not process app state change when app is backgrounded', () => {
```

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    mockAppStateListener('background');
+    jest.advanceTimersByTime(2000);
+
+    expect(mockTrackEvent).not.toHaveBeenCalled();
+  });
+
+  it('should not process app state change when app state has not changed', () => {
```
NicolasMassart, line 142:
```suggestion
  it('does not process app state change when app state is the same', () => {
```
Q: Does this also work for other states?

app/core/AppStateManager.ts
```diff
@@ -0,0 +1,61 @@
+import { AppState, AppStateStatus } from 'react-native';
+import { store } from '../store';
+import Logger from '../util/Logger';
+import { MetaMetrics, MetaMetricsEvents } from './Analytics';
+import extractURLParams from './DeeplinkManager/ParseManager/extractURLParams';
+import DevLogger from './SDKConnect/utils/DevLogger';
+
+class AppStateManager {
```
NicolasMassart, line 8: C: I'm not fond of the name "manager". It tells not much about what this class does exactly, it's mostly synonym to "does things". Could you find a more precise term? Maybe there's a term used in the design patterns that could match?

Maybe try something like "AppStateChangeListener"?

Also the only public method on this class is setCurrentDeeplink(), and all the state part is private. Maybe the class name should be about deeplink or the class does to many different things that you had to find a generic term like "manager".
See also my next comment about extracting attribution code. This may help clarifying the scope of this class.

app/core/AppStateManager.ts
```diff
@@ -0,0 +1,61 @@
...
+  private appStateSubscription: ReturnType<typeof AppState.addEventListener>;
+  private currentDeeplink: string | null = null;
+  private lastAppState: AppStateStatus = AppState.currentState;
+
+  constructor() {
+    this.lastAppState = AppState.currentState;
+    this.appStateSubscription = AppState.addEventListener('change', this.handleAppStateChange);
+  }
+
+  public setCurrentDeeplink(deeplink: string | null) {
```
NicolasMassart, line 18: Q: why the deeplink could be null? If the method is called, should it be when the deeplink exists?

app/core/AppStateManager.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+      DevLogger.log(`AppStateManager:: processAppStateChange:: sending event 'APP_OPENED' isMarketingEnabled=${isMarketingEnabled} attributionId=${attributionId}`);
+      MetaMetrics.getInstance().trackEvent(
+        MetaMetricsEvents.APP_OPENED,
+        { attributionId },
+        true
+      );
+    } catch (error) {
+      Logger.error(error as Error, 'AppStateManager: Error processing app state change');
+    }
```
NicolasMassart, line 53: S: Make nit modular. This attribution specific code could be extracted to a separate file to make the state management class scope restricted to managing states but not the consequences of state change.
You then link a specific process to happen on state change, here the attribution based on deeplink, and this is this part that will have the setCurrentDeeplink(), not the state change one.

NicolasMassart (changes requested):
The new class could benefit from being split in a way that matches the actual process more semantically.
Understanding what this does is not easy and I see a risk of this becoming a "catch-all" place under the synonym "manager" name.
Refactoring into a more modular code now, given it's still fairly simple, could prevent many issues later and make this part more readable.

app/core/Analytics/MetaMetrics.events.ts
deeeed: yes, DAPP_VIEWED is for in-app browser events. this is general to metamask app.

app/core/AppStateManager.test.ts
deeeed: we dont want to send the same event twice

NicolasMassart: I get it. Does this test also work if the state goes from background to background? Here we only test active.

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('subscribes to AppState changes on instantiation', () => {
+    expect(AppState.addEventListener).toHaveBeenCalledWith('change', expect.any(Function));
+  });
+
+  it('tracks event when app becomes active and conditions are met', () => {
```
NicolasMassart, line 63: S: could this name be more specific than "conditions are met". What conditions exactly?

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    jest.advanceTimersByTime(2000);
+
+    expect(mockTrackEvent).toHaveBeenCalledWith(
+      MetaMetricsEvents.APP_OPENED,
+      { attributionId: 'test123' },
+      true
+    );
+  });
+
+  it('does not track event when data collection is disabled', () => {
```
NicolasMassart, line 80: Q: is this redundant with the test already existing in metametrics to check if trackEvent is done when opt-out?

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    (store.getState as jest.Mock).mockReturnValue({
+      security: { dataCollectionForMarketing: true },
+    });
+
+    mockAppStateListener('active');
+    jest.advanceTimersByTime(2000);
+
+    expect(mockTrackEvent).toHaveBeenCalledWith(
+      MetaMetricsEvents.APP_OPENED,
+      { attributionId: undefined },
```
NicolasMassart, line 105: Q: should we send nothing at all instead of an undefined?

app/core/AppStateManager.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      true
+    );
+  });
+
+  it('handles errors gracefully', () => {
```
NicolasMassart, line 110:
```suggestion
  it('handles app state change processing errors', () => {
```

app/core/processAttribution.tsx
```diff
@@ -0,0 +1,14 @@
+import { store } from '../store';
+import extractURLParams from './DeeplinkManager/ParseManager/extractURLParams';
+
+export function processAttribution(currentDeeplink: string | null): string | undefined {
+  const state = store.getState();
+  const isMarketingEnabled = state.security.dataCollectionForMarketing;
+
+  if (isMarketingEnabled && currentDeeplink) {
+    const { params } = extractURLParams(currentDeeplink);
+    return params.attributionId;
```
NicolasMassart, line 10: C: if undefined, the return value from extractURLParams will be an empty string. Maybe you should align with your policy to return undefined.

app/core/AppStateEventListener.ts
```diff
@@ -34,15 +33,8 @@ class AppStateManager {
...
-      if (isMarketingEnabled && this.currentDeeplink) {
-          const { params } = extractURLParams(this.currentDeeplink);
-          attributionId = params.attributionId;
-      }
-      DevLogger.log(`AppStateManager:: processAppStateChange:: sending event 'APP_OPENED' isMarketingEnabled=${isMarketingEnabled} attributionId=${attributionId}`);
+      const attributionId = processAttribution(this.currentDeeplink);
+      DevLogger.log(`AppStateManager:: processAppStateChange:: sending event 'APP_OPENED' attributionId=${attributionId}`);
       MetaMetrics.getInstance().trackEvent(
         MetaMetricsEvents.APP_OPENED,
         { attributionId },
```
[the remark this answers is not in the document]
abretonc7s: this is the same thing...

app/core/processAttribution.tsx
abretonc7s: it('returns undefined when deeplink is undefined', () => {
    (store.getState as jest.Mock).mockReturnValue({
      security: { dataCollectionForMarketing: true },
    });

    const result = processAttribution(null);
    expect(result).toBeUndefined();
  });

app/core/AppStateEventListener.ts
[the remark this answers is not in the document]
abretonc7s: also you can't spread a string

NicolasMassart (dismissed):
Looks good to me.
Thanks for the very interesting discussion and for your care about the codebase!

NicolasMassart (dismissed):
Looks good to me.
Thanks for these last changes.

Cal-L (commented):
left a comment

app/components/Nav/App/index.js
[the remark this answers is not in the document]
deeeed: Initializing with the store definitely makes sense, but then there is also dependency to metametrics. I did play around initializing separately but then it also required changes on deeplink manager which was becoming a 
"complex" change for a small feature and went with this version.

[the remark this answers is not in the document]
deeeed: I can adjust on Monday.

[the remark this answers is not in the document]
abretonc7s: @Cal-L please take a look again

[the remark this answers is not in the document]
Cal-L: Nice, that's looking clean

Cal-L (approved):
LGTM

gauthierpetetin:
Fixes https://github.com/MetaMask/mobile-planning/issues/1522

PR #11321, opened 2024-09-19
fix: fix detect tokens performance

sahar-fehri:
## **Description**

This PR removes not needed calls in the codebase to `TokenDetectionController.detectTokens(); ` that was making the app slower;

Note: This performance issue is more noticeable on Android than IOS.

This PR remove this [call](https://github.com/MetaMask/metamask-mobile/blob/6fc1d9d3ae3fc2f7e447349f8b110e9520fc9e15/app/core/Engine.ts#L1698) in Engine.ts. This call is executed whenever the user switches networks; I think this is not needed since `TokenDetectionController` in core is already subscribed to `NetworkController:networkDidChange`. 

Similarly, This PR also also removes this [call](https://github.com/MetaMask/metamask-mobile/blob/6fc1d9d3ae3fc2f7e447349f8b110e9520fc9e15/app/components/Views/Wallet/index.tsx#L401) in wallet/index.tsx; 

I have also noticed that switching accounts in android is a bit slow; that was due to calling detectTokens function twice.
One call was made because TokenDetectionController is subscribed to `AccountsController:selectedAccountChange`
the other one was happening because of the subscription to `PreferencesController:stateChange` which is redundant and was happening because this [PR](https://github.com/MetaMask/core/pull/4219/files#diff-3336d65a4d2c1e5af90f22e16feb058f327960796b135308bca9ed8415418f5dR297) update was not patched;

When testing with simulator, after adding logs inside [onNetworkChange](https://github.com/MetaMask/metamask-mobile/blob/726ca826662ea6044cd5da671555c379e19a7864/app/components/Views/NetworkSelector/NetworkSelector.tsx#L191) 
This is how much it take to go from linea to Ethereum network on this PR vs main
Screenshot 2024-09-19 at 17 29 51

Screenshot 2024-09-19 at 18 27 51

Testing on physical:

https://github.com/user-attachments/assets/146cee68-75ba-459b-9e80-4b8f9bfb54b6

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1927

## **Manual testing steps**

(I found it easier to debug when you add console.log inside detect tokens function to make sure it is not being called multiple times)
1. Go to home page and select Linea network; then go back to Ethereum mainnet; switch back and forth multiple times and you should notice the switch is smoother now; you can check your console.log in your terminal and see that when you switch between networks; you only see it calling detectTokens once.
2. Switch accounts back and forth and it should also be faster; you can also check logs and make sure detectTokens is only triggered once when switching accounts

### **Before**

https://github.com/user-attachments/assets/721d1f56-6e1e-4940-9544-5d4b5c3232c0

### **After**

https://github.com/user-attachments/assets/c6287f8c-3091-4267-b82a-4c6bc02527fb

### DIFF
--- a/app/components/Views/Wallet/index.tsx
+++ b/app/components/Views/Wallet/index.tsx
@@ -388,9 +388,8 @@ const Wallet = ({
   useEffect(
     () => {
       requestAnimationFrame(async () => {
-        const { TokenDetectionController, AccountTrackerController } =
+        const { AccountTrackerController } =
           Engine.context;
-        TokenDetectionController.detectTokens();
         AccountTrackerController.refresh();
       });
     },
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -1678,7 +1678,6 @@ class Engine {
     const {
       AccountTrackerController,
       AssetsContractController,
-      TokenDetectionController,
       NetworkController,
       SwapsController,
     } = this.context;
@@ -1700,7 +1699,6 @@ class Engine {
       ).configuration.chainId,
       pollCountLimit: AppConstants.SWAPS.POLL_COUNT_LIMIT,
     });
-    TokenDetectionController.detectTokens();
     AccountTrackerController.refresh();
   }
 
--- a/patches/@metamask+assets-controllers+30.0.0.patch
+++ b/patches/@metamask+assets-controllers+30.0.0.patch
@@ -381,7 +381,7 @@ index ee6155c..06a3a04 100644
  };
  var NftDetectionController_default = NftDetectionController;
 diff --git a/node_modules/@metamask/assets-controllers/dist/chunk-HDI4L2DD.js b/node_modules/@metamask/assets-controllers/dist/chunk-HDI4L2DD.js
-index 76e3362..586b15e 100644
+index 76e3362..6f79cf7 100644
 --- a/node_modules/@metamask/assets-controllers/dist/chunk-HDI4L2DD.js
 +++ b/node_modules/@metamask/assets-controllers/dist/chunk-HDI4L2DD.js
 @@ -34,7 +34,7 @@ var STATIC_MAINNET_TOKEN_LIST = Object.entries(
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "metamask",
-  "version": "7.30.0",
+  "version": "v1.22.22",
```
sahar-fehri, line 3: 👀  removing this weird commit

tommasini (dismissed):

tommasini (commented):
detected some issues during testing, detected tokens not showing

sahar-fehri:
:wrench: looking into token detection issue

NicolasMassart (dismissed):

sahar-fehri:
About this [change](https://github.com/MetaMask/metamask-mobile/pull/11321/files#diff-57be48c81528a3189d712b45da56d116056f2871ca21f05b3158421d19909bc3R448):

When users switch accounts on android, the app was slow to switch; this happened because token detection was triggered
twice for `"PreferencesController:stateChange"` and  "`AccountsController:selectedAccountChange"`

After this core [PR](https://github.com/MetaMask/core/issues/4219); tokenDetection controller no longer checks if `isSelectedAddressChanged` to restart token detection. I have made this update in the patch, so we are making one less call to `detectTokens` and instead we are relying on the subscription to `AccountsController:selectedAccountChange` to restartTokenDetection.

This is where the bug reported happened, because while the event `AccountsController:selectedAccountChange` was triggered and we fetched the detected tokens; the app failed to add them to state.

This failure was happening because the current version of addDetectedTokens in TokensController in mobile is still relying on PreferencesController to fetch the selected address
` const { chainId: currentChain, selectedAddress: currentAddress } = this.config;`
The `currentAddress` we get in tokens controller today is a checksum address. 

Newer version of TokensController rely instead on this [event](https://github.com/MetaMask/core/pull/4219/files#diff-4de3568278666a6d7a1c5084b38802cc730ba3ce94b7fd90de875d758e462d99R240): `'AccountsController:selectedEvmAccountChange'` and the address is instead received from `AccountsController`; which is not a checksum address.

A proper fix would be to upgrade assets-controllers to at least v34.0.0;

I have initially started patching the following changes:

* TokenDetectionController:  instead of subscribing to `AccountsController:selectedAccountChange` => subscribe to `AccountsController:selectedEvmAccountChange`

* TokensController: instead of relying on preferencesController to configure selected address; allow for `AccountsController:getAccount`

It seemed like i have underestimated this patch, and patching AccountTrackerController would also be needed.

In this hot fix; i convert the address i receive in tokenDetectionController after subscribing to AccountsController:selectedAccountChange  to a checksum address;
Which allows tokens controller, when executing the function `addDetectedTokens`, to add them to state.

sethkfman (approved):
LGTM

tommasini (approved):

PR #11330, opened 2024-09-19
feat: disable wallet buttons for accounts that cannot sign transactions

k-g-j:
## **Description**

This PR disables certain buttons in the `WalletActions` component when the selected account cannot sign transactions. This change improves the user experience by preventing attempts to perform actions that require transaction signing when the account cannot do so.

1. The reason for the change is to prevent users from attempting actions that their current account cannot perform.
2. The improvement is that Buy, Sell, Send, Swap, and Bridge buttons are now disabled when the selected account cannot sign transactions, providing clear visual feedback to the user about available actions.

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/570

## **Manual testing steps**

1. Go to the [SSK](https://metamask.github.io/snap-simple-keyring/latest/) and create an account
2. Update the account using this JSON, replacing the address and ID fields with the new account information (this update step removes `eth_signTransaction` from the account's methods)
```json
{
    "id": "your new account ID",
    "address": "you new account address",
    "options": {},
    "methods": [
        "personal_sign",
        "eth_sign",
        "eth_signTypedData_v1",
        "eth_signTypedData_v3",
        "eth_signTypedData_v4"
    ],
    "type": "eip155:eoa"
}
```
3. Switch to the account that cannot sign transactions if not already there
4. Verify that the Buy, Sell, Send, Swap, and Bridge buttons are visually disabled with a message and not clickable
5. Click on the asset details view
6. Verify that the Buy, Sell, Send, Swap, and Bridge buttons are visually disabled and not clickable
7. Switch back to an account that can sign transactions
8. Verify that the buttons are now enabled and functional in wallet view and asset details view

### **Before**

Wallet view
 

Asset details view
simulator_screenshot_5FB98682-1B94-440A-9E12-BDE6D8B2B1C3

### **After**

Wallet view
Simulator Screenshot - iPhone 15 Pro - 2024-10-24 at 15 43 57

Asset details view
Simulator Screenshot - iPhone 15 Pro - 2024-10-24 at 15 44 06


app/components/UI/WalletAction/WalletAction.types.ts
```diff
@@ -1,14 +1,66 @@
...
+export interface WalletActionStrings {
+  title: string;
+  description: string;
+  disabledDescription: string;
+}
+
+export const walletActionStrings: Record<
+  WalletActionType,
+  WalletActionStrings
+> = {
```
gantunesr, line 24: I suggest a different naming for these types instead of using the word `string`

```suggestion
export interface WalletActionDetail {
  title: string;
  description: string;
  disabledDescription: string;
}

export const walletActionDetails: Record<
  WalletActionType,
  WalletActionDetail
> = {
```

app/components/UI/WalletAction/WalletAction.types.ts
```diff
@@ -1,14 +1,66 @@
...
+  [WalletActionType.Buy]: {
+    title: strings('asset_overview.buy_button'),
+    description: strings('asset_overview.buy_description'),
+    disabledDescription: strings('wallet.disabled_button.buy'),
```
gantunesr, line 28: Would it be better for these strings to also be inside `asset_overview` like the title and description?

app/components/Views/WalletActions/WalletActions.test.tsx
```diff
@@ -180,4 +187,48 @@ describe('WalletActions', () => {
 
     expect(mockNavigate).toHaveBeenCalled();
   });
+  it('should disable buttons when the account cannot sign transactions', () => {
```
gantunesr, line 190: nit: I know the rest of the file is using the previous convention but that shouldn't stop us of using the new one

```suggestion
  it('disables action buttons when the account cannot sign transactions', () => {
```

gantunesr (commented):
looks good overall, left some minor comments

app/components/UI/WalletAction/WalletAction.types.ts
k-g-j: renamed here [20e79f9](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/20e79f9f6747da06cc98ca81fd14d49cf471c11a)

app/components/UI/WalletAction/WalletAction.types.ts
k-g-j: good idea! [845c66f](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/845c66f17a2a2c164d0a942319d3f7995675b0c2)

app/components/UI/WalletAction/WalletAction.types.ts
```diff
@@ -1,14 +1,64 @@
...
+      title: strings('asset_overview.receive_button'),
+      description: strings('asset_overview.receive_description'),
+      disabledDescription: strings('asset_overview.disabled_button.receive'),
+    },
+  };
 
 export interface WalletActionProps {
-  actionTitle?: string;
-  actionDescription?: string;
+  actionType?: WalletActionType;
```
owencraston, line 56: why would this be optional if we are using it to derive the key information?

app/components/Views/WalletActions/WalletActions.tsx
```diff
@@ -54,6 +56,11 @@ const WalletActions = () => {
   const [isNetworkRampSupported] = useRampNetwork();
   const { trackEvent } = useMetrics();
 
+  const selectedAccount = useSelector(selectSelectedInternalAccount);
+  const canSignTransactions = selectedAccount?.methods?.includes(
```
owencraston, line 60: lets memoize this value with React.useMemo: https://react.dev/reference/react/useMemo

app/components/UI/WalletAction/WalletAction.types.ts
```diff
@@ -1,14 +1,64 @@
...
+  Receive = 'Receive',
+}
+
+export interface WalletActionDetail {
+  title: string;
+  description: string;
+  disabledDescription: string;
+}
+
+export const walletActionDetails: Record<WalletActionType, WalletActionDetail> =
```
owencraston, line 21: I like this a lot better.

app/components/Views/WalletActions/WalletActions.test.tsx
```diff
@@ -180,4 +187,48 @@ describe('WalletActions', () => {
 
     expect(mockNavigate).toHaveBeenCalled();
   });
+  it('disables action buttons when the account cannot sign transactions', () => {
+    const mockStateWithoutSigning: DeepPartial<RootState> = {
+      ...mockInitialState,
```
owencraston, line 192: Are all of these spreads necessary? Why cant you just modify the AccountsController section?

locales/languages/en.json
```diff
@@ -1237,6 +1238,14 @@
     "info": "Info",
     "swap": "Swap",
     "bridge": "Bridge",
+    "disabled_button": {
```
owencraston, line 1241: @AlexJupiter @eriknson what do you think of this copy?

app/components/UI/WalletAction/WalletAction.types.ts
k-g-j: It is currently optional in the implementation on the main branch [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/WalletAction/WalletAction.types.ts#L6) and making it non-optional would break [AssetDetailsActions](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/AssetDetails/AssetDetailsActions/AssetDetailsActions.tsx)

app/components/Views/WalletActions/WalletActions.tsx
k-g-j: smart thinking! [f5e18fb](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/f5e18fb493d7eb6311063ef787dbc0182ffa9d26)

app/components/Views/WalletActions/WalletActions.test.tsx
k-g-j: We do need to spread the whole state unfortunately, because the buttons such as the Buy and Swap buttons need to know the context of the Chain ID and other configurations [see this test](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/WalletActions/WalletActions.test.tsx#L101) for an example why

app/components/UI/WalletAction/WalletAction.types.ts
owencraston: true but now we are using this to derive the description which was previously a prop. I personally see no reason to make this optional. Do you?

k-g-j: Agreed, we should make it non-optional 
done here [8c2b833](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/8c2b833cb918dc60e8991e3f2c8e8544aefebd9e)

app/components/UI/WalletAction/WalletAction.tsx
```diff
@@ -8,44 +8,65 @@ import Text, {
...
-        <Text variant={TextVariant.BodyMD} style={styles.descriptionLabel}>
+        <Text
+          variant={TextVariant.BodyLGMedium}
+          style={disabled ? { color: colors.text.muted } : undefined}
+        >
+          {actionTitle}
+        </Text>
+        <Text
+          variant={TextVariant.BodyMD}
+          style={[
```
owencraston, line 65: passing inline styles like this mean that on every render a new array instance will be passed as a prop. because of this the virtual dom will not memoize the props for this component. can you try and extract these out so they are not computed inline? The same should be done for the rest of the files.

k-g-j: I believe I have made some improvement here in this commit [a9909ce](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/a9909ce0e9d22affaa69ed60173d8f326df456ef)

app/components/Views/AssetDetails/AssetDetailsActions/AssetDetailsActions.tsx
```diff
@@ -40,84 +40,78 @@ export const AssetDetailsActions: React.FC<AssetDetailsActionsProps> = ({
...
-            iconStyle={styles.icon}
-            containerStyle={styles.containerStyle}
-            {...generateTestId(Platform, TOKEN_OVERVIEW_BUY_BUTTON)}
-          />
-          <Text variant={TextVariant.BodyMD}>
-            {strings('asset_overview.buy_button')}
-          </Text>
-        </View>
+      {displayBuyButton &&
+        renderWalletAction(
```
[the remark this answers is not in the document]
k-g-j: removed! [88690e1](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/88690e1c50dc3996f419ac8125e25d878a85b0f9)

owencraston (dismissed):

plasmacorral:
Can confirm that observation 2 above can be reproduced on a flask build of main commit 9e3d2684acfaf193cc1781680beaf7ea29f5edb8.

owencraston (dismissed):

plasmacorral:
Observation 1 resolved in commit 91cb7d310b775b6d3f57875c713244fa1666f8f4
<div style="display: inline-block"></div>

Observation 2 is beyond the scope of this PR and was not introduced here.  Now tracked as [633](https://github.com/MetaMask/accounts-planning/issues/633).

Observation 3 can be addressed later once we are in production with the ability to disable the actions

plasmacorral:
Maybe only wallet view is in scope here?  These action buttons are presented and usable from the asset details pages, even for an account that does not support signing and has actions disabled from wallet view.

https://www.loom.com/share/68348eda25db4e2187a21c047a2eb32a

Is this something that should also be addressed within this PR, or do we take this incremental improvement and follow it with a subsequent PR to address asset detail pages?

plasmacorral:
I now see that action buttons are disabled on the asset details pages. However, there is no feedback when a user clicks on these muted buttons, which can create the impression that the mobile app is unresponsive.

Although the [recording](https://www.loom.com/share/000203123b90425b92255ba96a50baf9?sid=aab86031-f120-41e2-9d82-2ea0746a3503) doesn't show it well, repeatedly clicking these muted buttons provides no feedback, which can be frustrating for users.

I'm not sure if we have an established pattern for this on mobile; it might be a good question for our DS teammates.

In the Chrome extension with Flask, the cursor changes to a barred circle, and a tooltip appears without requiring a click, explicitly stating "Not supported on this account."

While we may not be able to fully replicate the extension behavior, we have an opportunity to make the user experience more explicit on the asset details page and avoid the app feeling unresponsive.
Screenshot 2024-10-10 at 9 37 36 PM

k-g-j:
> I'm not sure if we have an established pattern for this on mobile; it might be a good question for our DS teammates.

cc: @eriknson and @AlexJupiter what are your thoughts?

plasmacorral:
Having an issue with local builds of flask on this branch in commit ab171e4b881505a2abdc6611a3eb397394be6854

```
ERROR  Error: createSelector expects all input-selectors to be functions, but received the following types: [undefined, function memoized(), function memoized()], js engine: hermes
 ERROR  TypeError: Cannot read property 'store' of undefined, js engine: hermes
 ERROR  Error: createSelector expects all input-selectors to be functions, but received the following types: [undefined, function memoized(), function memoized()], js engine: hermes
 ERROR  TypeError: Cannot read property 'store' of undefined, js engine: hermes
```
Screenshot_20241018-174745

app/components/UI/WalletAction/WalletAction.tsx
```diff
@@ -1,51 +1,81 @@
...
+  );
+
+  const descriptionStyle = useMemo(
+    () => [
+      styles.descriptionLabel,
+      disabled ? { color: colors.text.muted } : undefined,
+    ],
+    [styles, disabled, colors],
+  );
+
```
brianacnguyen, line 58: These style changes are not needed (and also does not follow our disabled pattern. Instead, you can just update the container style to have  `opacity: 0.5`

brianacnguyen (changes requested):
1 small request

app/components/UI/WalletAction/WalletAction.tsx
k-g-j: okay! thanks for letting me me know :) 
I adjusted the styles to only use `opacity: 0.5` in this commit [153c801](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/153c801904870d58708663be384383abbbd098f0)

app/components/Views/AssetDetails/AssetDetailsActions/AssetDetailsActions.tsx
```diff
@@ -39,47 +41,51 @@ export const AssetDetailsActions: React.FC<AssetDetailsActionsProps> = ({
...
     <View style={styles.activitiesButton}>
       {displayBuyButton && (
         <View style={styles.buttonWrapper}>
           <WalletAction
             iconName={IconName.Add}
-            iconSize={AvatarSize.Lg}
             onPress={onBuy}
-            iconStyle={styles.icon}
-            containerStyle={styles.containerStyle}
+            {...walletActionProps}
```
owencraston, line 63: I THINK the spread operator here defeats the purpose of the memoization since spreading the object will create a new reference anyway. [See this stackoverflow thread](https://stackoverflow.com/questions/52959059/does-spread-operator-return-new-object-after-operation). I am not certain about this so please correct me if I am wrong.

owencraston: the same goes for the rest of the file.

app/components/Views/WalletActions/WalletActions.tsx
```diff
@@ -67,132 +77,149 @@ const WalletActions = () => {
...
-            actionTitle={strings('asset_overview.buy_button')}
-            actionDescription={strings('asset_overview.buy_description')}
+            actionType={WalletActionType.Buy}
             iconName={IconName.Add}
-            iconSize={AvatarSize.Md}
             onPress={onBuy}
-            iconStyle={styles.icon}
             actionID={WalletActionsModalSelectorsIDs.BUY_BUTTON}
+            iconStyle={styles.icon}
+            {...walletActionBaseProps}
```
owencraston, line 172: same question about spreading the memoized props.

owencraston (commented):
Looking very good and very nice 😁 ! The code looks good I just had a few comments/questions about all the memoization of props.

owencraston:
you will need to run `yarn jest -u pp/components/UI/AssetOverview/AssetOverview.test.tsx` to update the snapshot and get your unit tests passing.

app/components/UI/WalletAction/WalletAction.tsx
```diff
@@ -1,46 +1,59 @@
...
+      ...iconStyle,
+      backgroundColor: disabled ? colors.primary.muted : colors.primary.default,
+    }),
+    [disabled, colors, iconStyle],
+  );
+
   return (
     <TouchableOpacity
-      style={{ ...styles.base, ...containerStyle }}
+      style={[styles.base, containerStyle, disabled && styles.disabled]}
```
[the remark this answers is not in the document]
k-g-j: [5946bfb](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/5946bfb50a54bf9003b4ef85e56d2c47227203c4)

app/components/Views/AssetDetails/AssetDetailsActions/AssetDetailsActions.tsx
k-g-j: I THINK you are right

I updated the code in this commit [37077bf](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/37077bfc24ac5a8022b1e5d05f833d3ec17bff89) so we're not creating new object references unnecessarily, which was defeating the purpose of memoization. It also makes the code more explicit about which props are being passed to each `WalletAction` component.

app/components/Views/WalletActions/WalletActions.tsx
k-g-j: Same logic as your above comment and my point about creating new object references unnecessarily 
addressed here [37077bf](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/37077bfc24ac5a8022b1e5d05f833d3ec17bff89)

plasmacorral:
Tested on Pixel 5a with Android 14 with commit [f714dd1d56a1b22e79b6b53cc8a5668294a22eca](https://github.com/MetaMask/metamask-mobile/pull/11330/commits/f714dd1d56a1b22e79b6b53cc8a5668294a22eca). Buttons were enabled in the wallet view bottom bar and asset view when signing was enabled on the snap account, and disabled when signing permission was removed. Re-enabling signing restored the buttons. Checked on Eth mainnet, Sepolia, Base, and Arbitrum. No impact on button display or function for HD, imported, QR, and Ledger accounts.

app/components/UI/WalletAction/WalletAction.tsx
```diff
@@ -1,46 +1,65 @@
...
+  const actionStrings = actionType ? walletActionDetails[actionType] : null;
+  const actionTitle = actionStrings?.title ?? '';
+  const actionDescription = actionStrings?.description;
   const { colors } = lightTheme;
   const { styles } = useStyles(styleSheet, {});
+
+  const avatarStyle = useMemo(
+    () => ({
+      ...iconStyle,
+      backgroundColor: disabled ? colors.primary.muted : colors.primary.default,
```
brianacnguyen, line 39: this disabled change is not needed since the opacity would take care of color change

brianacnguyen (changes requested):
1 small request regarding avatar disabled

owencraston:
Do to difficulty rebasing this PR I am continuing the work over [here](https://github.com/MetaMask/metamask-mobile/pull/12145).

PR #11412, opened 2024-09-24
fix: upgrade assets-controllers to v31

sahar-fehri:
## **Description**

PR to upgrade assets-controllers to v31.0.0

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### **After**

Verifying NFT and Token detection:

https://github.com/user-attachments/assets/b9cda3a7-bfc9-4465-bca0-f348648cec8e


app/core/Engine.ts
```diff
@@ -2027,17 +2007,17 @@ class Engine {
     // TODO implement this method on SelectedNetworkController
     // SelectedNetworkController.unsetAllDomains()
     //Clear assets info
-    TokensController.update({
+    /*     TokensController.update({
```
tommasini, line 2010: Was this left commented intentionally?

app/core/EngineService/EngineService.ts
```diff
@@ -54,15 +54,19 @@ class EngineService {
         key: `${engine.context.AddressBookController.name}:stateChange`,
       },
       { name: 'AssetsContractController' },
-      { name: 'NftController' },
+      { name: 'NftController', key: 'NftController:stateChange' },
       {
         name: 'TokensController',
+        key: 'TokensController:stateChange',
```
tommasini, line 60: We could keep the same pattern
```suggestion
        key: `${engine.context.TokensController.name}:stateChange`,,
```

tommasini:
Can we provide recordings on the PR description? It would be also cool to have a core branch so the review of the patch is easier!

app/core/Engine.ts
sahar-fehri: Removed

sahar-fehri:
> Can we provide recordings on the PR description? It would be also cool to have a core branch so the review of the patch is easier!

Will add recordings in description 🙏  i have manually made the update in node-modules then used the patch command to create the patch file; i did not create a branch in core for the upgrade; is that necessary for each upgrade?

tommasini:
@sahar-fehri for me it's okay without the core branch, it's much smaller now and readable! Thanks!

app/core/Engine.ts
```diff
@@ -2010,18 +1990,6 @@ class Engine {
     // Clear selected network
     // TODO implement this method on SelectedNetworkController
     // SelectedNetworkController.unsetAllDomains()
-    //Clear assets info
-    TokensController.update({
```
tommasini, line 2014: I see that the resolution for the conversation was to removed this, although this is part of the vault corruption flow, that [function](https://github.com/MetaMask/metamask-mobile/blob/5275022825865dfb378cde69f58b4d33fcda1ca6/app/core/EngineService/EngineService.ts#L204) log can be checked at EngineService, I believe we want to keep the original behaviour of it if nothing pushes for the other way. 
I see that TokensController and NftController now extend BaseController v2 what motivate this change because their state can not longer be updated like this since now it's a protected function.
On that sense I highly recommend that we built reset methods on each of them to reset their state, like it's done on TokenBalancesController (It seems that function only lives on the patch and it was not pushed to the core repo, so most likely my fault when I updated assets controllers lately)

package.json
```diff
@@ -145,7 +145,7 @@
     "@metamask/accounts-controller": "^18.2.1",
     "@metamask/address-book-controller": "^6.0.1",
     "@metamask/approval-controller": "^7.0.1",
-    "@metamask/assets-controllers": "^30.0.0",
+    "@metamask/assets-controllers": "^31.0.0",
```
tommasini, line 148: I almost certain that this update needs to be together with PreferencesController update to v^12, because now assets-controllers is pointing to preferences-controller version^12 and that version doesn't have the `openseaEnabled` property applied, because that it's applied via patch to PreferencesController v^11 

With this being said, I'm almost certain that the display nft media toggle is not working as expected

gambinish: I can probably help with the `PreferencesController` update to v12 since I'm going to be in that controller working on a few new features soon.

I will follow up with the team today.

app/core/Engine.ts
gambinish: It looks like this is the PR to update TokensController to extend BaseController: https://github.com/MetaMask/core/pull/4304

And the corresponding issue: https://github.com/MetaMask/core/issues/4075

It looks like the main motivation was to remove the flattened `config` in favor of more specific `selectedAddress` and `provider` parameters.

Are you suggesting that we add a `reset` method to the `TokensController` that does the same thing as this diff, but internally?

package.json
gambinish: @tommasini I synced with @sahar-fehri and the team, and it looks like the PreferencesController may not need to be bumped here at this time (might make sense in a future PR though)

Did some testing of the  `openseaEnabled` flag, and it seems to be working fine. I think the naming convention is a little different on mobile vs extension but the behavior seems to be expected:

With disabled settings:

Screenshot 2024-10-07 at 9 40 48 AM
Screenshot 2024-10-07 at 9 41 09 AM

Pressing confirm toggles setting:

Screenshot 2024-10-07 at 9 41 38 AM
Screenshot 2024-10-07 at 9 42 33 AM

And images render:

Screenshot 2024-10-07 at 9 46 12 AM

Also tested edge case in a different wallet where some images are hosted on ipfs, while others are not (Pinata Gateway selected):

Screenshot 2024-10-07 at 9 49 06 AM

Screenshot 2024-10-07 at 9 49 28 AM

Gateway toggle off:

Screenshot 2024-10-07 at 9 53 28 AM

sahar-fehri: Hi @tommasini !
Preferences-controller v12 and latest have `openseaEnabled`; for some historic reason; mobile has been using a patched version of preferences-controller;
It would be great to align all clients and core to use same version of preferences controller but i believe it could be done in a separate PR;
Today, even if we upgrade preferences controller to latest with the same codebase of mobile, we would still need to do the patch for openseaEnabled and displayNftMedia unless we also refactor mobile codebase to use the appropriate fcts and state preferences

app/core/Engine.ts
tommasini: Yes 🙏🏼

tommasini: Ah I see it was implemented! Cool!

package.json
tommasini: This is super interesting: https://github.com/MetaMask/metamask-mobile/pull/11412/files#diff-96e2588d1d93ac60788df22c864a19d9cac5bcdf44951811c1fafe0d938397dd

How it was possible to create a patch for a dependency of a dependency? :o 

yeah, for historical reasons the `openSeaEnabled` was changed for `displayNftMedia`, was actually me on the past, and unfortunately, never finished that implementation on core :( 

I see, this is working because assets controllers dependency of preferences controller v^12 is only used to types and the actually logic used is the preferences controller v^11? 
Let me know if my question was confusing! If it's confirmed that it's working it's okay to me! Just trying to understand why, I'm definitely missing some knowledge here!

app/core/Engine.ts
sahar-fehri: Thnx Tomas, pushed here [c8611f9](https://github.com/MetaMask/metamask-mobile/pull/11412/commits/c8611f94bfed771f3636d65efe692e227929e606) 🙏

app/components/UI/AssetElement/index.test.tsx
```diff
@@ -17,7 +17,7 @@ describe('AssetElement', () => {
     balanceFiat: ' $1',
     logo: '',
     isETH: undefined,
-    balanceError: null,
+    hasBalanceError: false,
```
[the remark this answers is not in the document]
sahar-fehri: Added here [7baba63](https://github.com/MetaMask/metamask-mobile/pull/11412/commits/7baba63172487eadded37bdbb4fa2efa42f94bc8)

app/store/migrations/054.ts
```diff
@@ -0,0 +1,41 @@
...
+      }
+      delete token?.balanceError;
+      return token;
+    });
+    tokensControllerState.tokens = migratedTokens;
+  }
+
+  // Return the modified state
+  return state;
+}
```
gambinish, line 41: This migration looks good, and the tests are passing so I am going to approve this as is.

I'm still having some trouble verifying this upgrade path locally. Before we merge this PR, could we go through this together?

We should also double check with @salimtb to make sure that we don't have any blockers on his default RPC work before we merge, since I think that should probably take priority.

gambinish (dismissed):
Preemptively approving, but I think that we should sync up as a team before we get this merged 👍

app/store/migrations/054.test.ts
```diff
@@ -0,0 +1,521 @@
...
+
+      expect(newState).toStrictEqual(state);
+      expect(mockedCaptureException).toHaveBeenCalledWith(expect.any(Error));
+      expect(mockedCaptureException.mock.calls[0][0].message).toBe(
+        errorMessage,
+      );
+    });
+  }
+
+  it('should update hasBalanceError to false in tokensController state', async () => {
```
tommasini, line 70: NIT: we do not need to async/await the migrarion, since it is not asynchronous 🙏🏼

tommasini (dismissed):
LGTM! Just a small comment on the test file!

app/store/migrations/054.test.ts
sahar-fehri: Right! fixed

tommasini (approved):

cryptodev-2s (approved):
LGTM!

gambinish (approved):

salimtb (approved):

PR #11451, opened 2024-09-26
feat: Remove Account Snap Warning (Flask)

owencraston:
## **Description**

- This change adds an extra warning to users when trying to remove an account snap (https://metamask.github.io/snap-simple-keyring/latest/ for example) AND the user has snap accounts in their wallet
- Snap accounts are not managed/owned by metamask and therefore cannot be recovered in metamask. User's must manage and backup these accounts in the snaps UI.
- Adds an extra layer of friction when trying to remove an account snap and the user has snap accounts in their wallet
- This is a `FLASK ONLY` feature

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/155

## **Manual testing steps**

1. check this branch
2. open `.js.env`
3. ensure that `export METAMASK_BUILD_TYPE=flask` is set to flask (not main)
4. build the app on this branch
5. create/import a wallet
6. inside the mobile browser, open this URL: https://metamask.github.io/snap-simple-keyring/latest/
7. connect this snap
8. click create account, you can do this as many times as you want.
9. navigate back to the main wallet view
10. NOTICE there should be new accounts in your wallet with the name "Snap Account x"
11. Go to settings
12. navigate to snaps
13. locate the Snap simple keyring settings
14. scroll to the bottom of the page and click the red "Remove" button
15. Notice: new warning should pop up indicating that the user should backup these accounts before removing the snap
16. click continue
17. you should be prompted to type the name of the snap in the text input.
18. YOU SHOULD NOT BE ABLE TO REMOVE THE SNAP WITHOUT TYPING The CORRECT SNAP NAME
19. Once you type the correct snap name, you should be able to remove the snap
20. After clicking remove the snap should be removed.

#### Testing a non account snap
1. check this branch
2. open `.js.env`
3. ensure that `export METAMASK_BUILD_TYPE=flask` is set to flask (not main)
4. build the app on this branch
5. create/import a wallet
6. navigate to settings/snaps
7. click on the pre installed Message signing snap
8. click on the remove button at the bottom
9. `THERE SHOULD NOT BE A WARNING`

#### Testing removing a snap account without snap accounts in the wallet
1. check this branch
2. open `.js.env`
3. ensure that `export METAMASK_BUILD_TYPE=flask` is set to flask (not main)
4. build the app on this branch
5. create/import a wallet
6. inside the mobile browser, open this URL: https://metamask.github.io/snap-simple-keyring/latest/
7. connect this snap. Do not create any accounts this time.
8. Go to settings
9. navigate to snaps
10. locate the Snap simple keyring settings
11. scroll to the bottom of the page and click the red "Remove" button
12. There should be no warning pop up and the snap should be removed

### Extension Version

Screenshot 2024-10-01 at 5 38 58 PM
Screenshot 2024-10-01 at 5 39 14 PM
Screenshot 2024-10-01 at 5 39 20 PM

### **After**

with snap accounts:

https://github.com/user-attachments/assets/24aa2993-98fa-4d72-bc1f-b7f9c1aeb6ad

Screenshot 2024-10-10 at 3 26 46 PM

Without snap accounts:

https://github.com/user-attachments/assets/7cf42c03-dc6e-40fb-b490-fb0eb4d8ba83

### DIFF
--- a/app/components/Views/AccountActions/AccountActions.tsx
+++ b/app/components/Views/AccountActions/AccountActions.tsx
@@ -34,7 +34,12 @@ import { protectWalletModalVisible } from '../../../actions/user';
 import Routes from '../../../constants/navigation/Routes';
 import { AccountActionsModalSelectorsIDs } from '../../../../e2e/selectors/Modals/AccountActionsModal.selectors';
 import { useMetrics } from '../../../components/hooks/useMetrics';
-import { isHardwareAccount } from '../../../util/address';
+import {
+  isHardwareAccount,
+  ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+  isSnapAccount,
+  ///: END:ONLY_INCLUDE_IF
+} from '../../../util/address';
 import { removeAccountsFromPermissions } from '../../../core/Permissions';
 import ExtendedKeyringTypes, {
   HardwareDeviceTypes,
@@ -189,6 +194,50 @@ const AccountActions = () => {
     }
   }, [controllers.KeyringController]);
 
+  ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+
+  /**
+   * Remove the snap account from the keyring
+   */
+  const removeSnapAccount = useCallback(async () => {
+    if (selectedAddress) {
+      await controllers.KeyringController.removeAccount(selectedAddress as Hex);
+      await removeAccountsFromPermissions([selectedAddress]);
+      trackEvent(MetaMetricsEvents.ACCOUNT_REMOVED, {
+        accountType: keyring?.type,
+        selectedAddress,
+      });
+    }
+  }, [
+    controllers.KeyringController,
+    keyring?.type,
+    selectedAddress,
+    trackEvent,
+  ]);
+
+  const showRemoveSnapAccountAlert = useCallback(() => {
+    Alert.alert(
+      strings('accounts.remove_snap_account'),
+      strings('accounts.remove_snap_account_alert_description'),
+      [
+        {
+          text: strings('accounts.remove_account_alert_cancel_btn'),
+          style: 'cancel',
+        },
+        {
+          text: strings('accounts.remove_account_alert_remove_btn'),
+          onPress: async () => {
+            sheetRef.current?.onCloseBottomSheet(async () => {
+              await removeSnapAccount();
+              await selectFirstAccount();
+            });
+          },
+        },
+      ],
+    );
+  }, [removeSnapAccount, selectFirstAccount]);
+  ///: END:ONLY_INCLUDE_IF
+
   /**
    * Forget the device if there are no more accounts in the keyring
    * @param keyringType - The keyring type
@@ -306,6 +355,18 @@ const AccountActions = () => {
             testID={AccountActionsModalSelectorsIDs.REMOVE_HARDWARE_ACCOUNT}
           />
         )}
+        {
+          ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+          selectedAddress && isSnapAccount(selectedAddress) && (
+            <AccountAction
+              actionTitle={strings('accounts.remove_snap_account')}
+              iconName={IconName.Close}
+              onPress={showRemoveSnapAccountAlert}
+              testID={AccountActionsModalSelectorsIDs.REMOVE_SNAP_ACCOUNT}
+            />
+          )
+          ///: END:ONLY_INCLUDE_IF
+        }
       </View>
       <BlockingActionModal
         modalVisible={blockingModalVisible}
--- a/app/components/Views/AccountActions/AccountActions.tsx
+++ b/app/components/Views/AccountActions/AccountActions.tsx
@@ -229,13 +229,12 @@ const AccountActions = () => {
           onPress: async () => {
             sheetRef.current?.onCloseBottomSheet(async () => {
               await removeSnapAccount();
-              await selectFirstAccount();
             });
           },
         },
       ],
     );
-  }, [removeSnapAccount, selectFirstAccount]);
+  }, [removeSnapAccount]);
   ///: END:ONLY_INCLUDE_IF
 
   /**
--- a/app/components/Views/Snaps/KeyringSnapRemovalWarning/KeyringSnapRemovalWarning.tsx
+++ b/app/components/Views/Snaps/KeyringSnapRemovalWarning/KeyringSnapRemovalWarning.tsx
@@ -37,6 +37,7 @@ import {
   KEYRING_SNAP_REMOVAL_WARNING,
   KEYRING_SNAP_REMOVAL_WARNING_CANCEL,
   KEYRING_SNAP_REMOVAL_WARNING_CONTINUE,
+  KEYRING_SNAP_REMOVAL_WARNING_TEXT_INPUT
 } from './KeyringSnapRemovalWarning.constants';
 
 interface KeyringSnapRemovalWarningProps {
@@ -174,6 +175,7 @@ export default function KeyringSnapRemovalWarning({
               style={styles.input}
               value={confirmationInput}
               onChangeText={handleConfirmationInputChange}
+              testID={KEYRING_SNAP_REMOVAL_WARNING_TEXT_INPUT}
             />
             {error && (
               <Text variant={TextVariant.BodySM} style={styles.errorText}>
--- a/app/components/Views/Snaps/KeyringSnapRemovalWarning/KeyringSnapRemovalWarning.tsx
+++ b/app/components/Views/Snaps/KeyringSnapRemovalWarning/KeyringSnapRemovalWarning.tsx
@@ -1,0 +1,208 @@
+///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+import React, {
+  useEffect,
+  useRef,
+  useState,
+  useCallback,
+  useMemo,
+} from 'react';
+import { Snap } from '@metamask/snaps-utils';
+import BottomSheet, {
+  BottomSheetRef,
+} from '../../../../component-library/components/BottomSheets/BottomSheet';
+import Text, {
+  TextVariant,
+} from '../../../../component-library/components/Texts/Text';
+import { InternalAccount } from '@metamask/keyring-api';
+import BannerAlert from '../../../../component-library/components/Banners/Banner/variants/BannerAlert';
+import { BannerAlertSeverity } from '../../../../component-library/components/Banners/Banner';
+import BottomSheetHeader from '../../../../component-library/components/BottomSheets/BottomSheetHeader';
+import { ScrollView, View, TextInput } from 'react-native';
+import { useStyles } from '../../../hooks/useStyles';
+import stylesheet from './KeyringSnapRemovalWarning.styles';
+import { strings } from '../../../../../locales/i18n';
+import { KeyringAccountListItem } from '../components/KeyringAccountListItem';
+import { getAccountLink } from '@metamask/etherscan-link';
+import { useSelector } from 'react-redux';
+import { selectProviderConfig } from '../../../../selectors/networkController';
+import BottomSheetFooter, {
+  ButtonsAlignment,
+} from '../../../../component-library/components/BottomSheets/BottomSheetFooter';
+import {
+  ButtonProps,
+  ButtonSize,
+  ButtonVariants,
+} from '../../../../component-library/components/Buttons/Button/Button.types';
+import {
+  KEYRING_SNAP_REMOVAL_WARNING,
+  KEYRING_SNAP_REMOVAL_WARNING_CANCEL,
+  KEYRING_SNAP_REMOVAL_WARNING_CONTINUE,
+} from './KeyringSnapRemovalWarning.constants';
+
+interface KeyringSnapRemovalWarningProps {
+  snap: Snap;
+  keyringAccounts: InternalAccount[];
+  onCancel: () => void;
+  onClose: () => void;
+  onSubmit: () => void;
+}
+
+export default function KeyringSnapRemovalWarning({
+  snap,
+  keyringAccounts,
+  onCancel,
+  onClose,
+  onSubmit,
+}: KeyringSnapRemovalWarningProps) {
+  const [showConfirmation, setShowConfirmation] = useState(false);
+  const [confirmedRemoval, setConfirmedRemoval] = useState(false);
+  const [confirmationInput, setConfirmationInput] = useState('');
+  const [error, setError] = useState(false);
+  const { chainId } = useSelector(selectProviderConfig);
+  const { styles } = useStyles(stylesheet, {});
+  const bottomSheetRef = useRef<BottomSheetRef>(null);
+
+  useEffect(() => {
+    setShowConfirmation(keyringAccounts.length === 0);
+  }, [keyringAccounts]);
+
+  const validateConfirmationInput = useCallback(
+    (input: string): boolean => input === snap.manifest.proposedName,
+    [snap.manifest.proposedName],
+  );
+
+  const handleConfirmationInputChange = useCallback(
+    (text: string) => {
+      setConfirmationInput(text);
+      setConfirmedRemoval(validateConfirmationInput(text));
+    },
+    [validateConfirmationInput],
+  );
+
+  const handleContinuePress = useCallback(() => {
+    if (!showConfirmation) {
+      setShowConfirmation(true);
+    } else if (confirmedRemoval) {
+      try {
+        onSubmit();
+      } catch (e) {
+        setError(true);
+      }
+    }
+  }, [showConfirmation, confirmedRemoval, onSubmit]);
+
+  const cancelButtonProps: ButtonProps = useMemo(
+    () => ({
+      variant: ButtonVariants.Secondary,
+      label: strings(
+        'app_settings.snaps.snap_settings.remove_account_snap_warning.cancel_button',
+      ),
+      size: ButtonSize.Lg,
+      onPress: onCancel,
+      testID: KEYRING_SNAP_REMOVAL_WARNING_CANCEL,
+    }),
+    [onCancel],
+  );
+
+  const continueButtonProps: ButtonProps = useMemo(
+    () => ({
+      variant: ButtonVariants.Primary,
+      label: showConfirmation
+        ? strings(
+            'app_settings.snaps.snap_settings.remove_account_snap_warning.remove_snap_button',
+          )
+        : strings(
+            'app_settings.snaps.snap_settings.remove_account_snap_warning.continue_button',
+          ),
+      size: ButtonSize.Lg,
+      onPress: handleContinuePress,
+      isDisabled: showConfirmation && !confirmedRemoval,
+      isDanger: showConfirmation,
+      testID: KEYRING_SNAP_REMOVAL_WARNING_CONTINUE,
+    }),
+    [showConfirmation, confirmedRemoval, handleContinuePress],
+  );
+
+  const accountListItems = useMemo(
+    () =>
+      keyringAccounts.map((account, index) => (
+        <KeyringAccountListItem
+          key={index}
+          account={account}
+          snapUrl={getAccountLink(account.address, chainId)}
+        />
+      )),
+    [keyringAccounts, chainId],
+  );
+
+  return (
+    <BottomSheet
+      ref={bottomSheetRef}
+      isFullscreen={false}
+      onClose={onClose}
+      shouldNavigateBack={false}
+      testID={KEYRING_SNAP_REMOVAL_WARNING}
+    >
+      <View style={styles.container}>
+        <BottomSheetHeader>
+          <Text variant={TextVariant.HeadingMD}>
+            {strings(
+              'app_settings.snaps.snap_settings.remove_account_snap_warning.title',
+            )}
+          </Text>
+        </BottomSheetHeader>
+        <BannerAlert
+          severity={BannerAlertSeverity.Warning}
+          title={strings(
+            'app_settings.snaps.snap_settings.remove_account_snap_warning.banner_title',
+          )}
+        />
+        {showConfirmation ? (
+          <>
+            <Text variant={TextVariant.BodyMD} style={styles.description}>
+              {strings(
+                'app_settings.snaps.snap_settings.remove_account_snap_warning.remove_account_snap_alert_description_1',
+              )}
+              <Text variant={TextVariant.BodyMDBold}>
+                {snap.manifest.proposedName}
+              </Text>
+              {strings(
+                'app_settings.snaps.snap_settings.remove_account_snap_warning.remove_account_snap_alert_description_2',
+              )}
+            </Text>
+            <TextInput
+              style={styles.input}
+              value={confirmationInput}
+              onChangeText={handleConfirmationInputChange}
+            />
+            {error && (
+              <Text variant={TextVariant.BodySM} style={styles.errorText}>
+                {strings(
+                  'app_settings.snaps.snap_settings.remove_account_snap_warning.remove_snap_error',
+                  {
+                    snapName: snap.manifest.proposedName,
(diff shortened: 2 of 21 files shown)


app/util/test/accountsControllerTestUtils.ts
```diff
@@ -39,6 +39,37 @@ export function createMockInternalAccount(
...
+    id: createMockUuidFromAddress(address),
+    metadata: {
+      name: nickname,
+      importTime: 1684232000456,
+      keyring: {
+        type: 'Snap Keyring',
+      },
+      snap: {
+        id: 'npm:@metamask/snap-simple-keyring-snap',
+        name: 'MetaMask Simple Snap Keyring',
```
k-g-j, line 57:
```suggestion
        name: 'MetaMask Snap Simple Keyring',
```

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
```diff
@@ -42,9 +45,16 @@ const SnapSettings = () => {
...
+    isShowingSnapKeyringRemoveWarning,
+    setIsShowingSnapKeyringRemoveWarning,
+  ] = useState(false);
+
+  const [keyringAccounts, setKeyringAccounts] = useState<InternalAccount[]>([]);
+  const internalAccounts = useSelector(selectInternalAccounts);
+
   // TODO: Replace "any" with type
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   function getPermissionSubjects(state: any) {
```
k-g-j, line 60: We could consider using TypeScript's `Record` type instead of `any` for `getPermissionSubjects` and `getPermissions` functions. This would improve type safety and make the code more self-documenting. 

For example:

   ```typescript
   function getPermissionSubjects(state: Record<string, unknown>) {
     return (state.subjects as Record<string, unknown>) || {};
   }

   function getPermissions(state: Record<string, unknown>, origin: string) {
     return (getPermissionSubjects(state)[origin] as Record<string, unknown>)?.permissions;
   }

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
```diff
@@ -71,52 +81,106 @@ const SnapSettings = () => {
...
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const { SnapController } = Engine.context as any;
+    const { SnapController } = Engine.context;
     await SnapController.removeSnap(snap.id);
     navigation.goBack();
   }, [navigation, snap.id]);
 
+  const isKeyringSnap = Boolean(permissionsFromController?.snap_manageAccounts);
+  
+  useEffect(() => {
```
k-g-j, line 91: I think the `useEffect` for `setKeyringAccounts` could be optimized by moving the `getAccountsBySnapId` call outside the effect and using `useMemo` instead. 

```typescript
const keyringAccounts = useMemo(() => {
  if (isKeyringSnap) {
    const addresses = await getAccountsBySnapId(snap.id);
    return Object.values(internalAccounts).filter(
      (internalAccount) =>
        addresses.includes(internalAccount.address.toLowerCase()),
    );
  }
  return [];
}, [snap?.id, internalAccounts, isKeyringSnap]);
```

app/components/Views/Snaps/components/KeyringAccountListItem/KeyringAccountListItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+            {toChecksumHexAddress(account.address)}
+          </Text>
+        </View>
+        <View style={styles.buttonContainer}>
+          <ButtonIcon
+            testID={KEYRING_ACCOUNT_LIST_ITEM_BUTTON}
+            iconName={IconName.Export}
+            iconColor={IconColor.Primary}
+            size={ButtonIconSizes.Lg}
+            onPress={() => Linking.openURL(snapUrl)}
```
k-g-j, line 57: Could the `onPress` handler here for `ButtonIcon` be memoized with `useCallback` to prevent unnecessary re-renders? 

Something along the lines of . . . 

```typescript
  const handlePress = useCallback(() => {
    Linking.openURL(snapUrl);
  }, [snapUrl]);

  return (
    <ButtonIcon
      testID={KEYRING_ACCOUNT_LIST_ITEM_BUTTON}
      iconName={IconName.Export}
      iconColor={IconColor.Primary}
      size={ButtonIconSizes.Lg}
      onPress={handlePress}
    />
  );
}
```

app/components/Views/Snaps/KeyringSnapRemovalWarning/test/KeyringSnapRemovalWarning.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+    const textInput = getByTestId(KEYRING_SNAP_REMOVAL_WARNING_TEXT_INPUT);
+    expect(textInput).toBeTruthy();
+    fireEvent.changeText(textInput, mockSnapName);
+
+    await waitFor(() => {
+      expect(continueButton.props.disabled).toBe(false);
+    });
+  });
+
```
k-g-j, line 210: Let's test this edge case too
```suggestion

it('handles empty snap names correctly', () => {
  const { getByTestId } = renderWithProvider(
    <KeyringSnapRemovalWarning
      snap={{ ...mockSnap, manifest: { ...mockSnap.manifest, proposedName: '' } }}
      keyringAccounts={[]}
      onCancel={onCancelMock}
      onClose={onCloseMock}
      onSubmit={onSubmitMock}
    />
  );
  
  const textInput = getByTestId(KEYRING_SNAP_REMOVAL_WARNING_TEXT_INPUT);
  fireEvent.changeText(textInput, '');

  const continueButton = getByTestId(KEYRING_SNAP_REMOVAL_WARNING_CONTINUE);
  expect(continueButton.props.disabled).toBe(true);

  expect(textInput.props.value).toBe('');
  expect(continueButton.props.children[1].props.children).toBe('Remove Snap');
});

```

app/components/Views/Snaps/KeyringSnapRemovalWarning/KeyringSnapRemovalWarning.tsx
```diff
@@ -0,0 +1,210 @@
...
+  );
+
+  const handleContinuePress = useCallback(() => {
+    if (!showConfirmation) {
+      setShowConfirmation(true);
+    } else if (confirmedRemoval) {
+      try {
+        onSubmit();
+      } catch (e) {
+        setError(true);
```
k-g-j, line 90: Maybe it would be helpful with debugging if we added some type of more specific error messaging 

something like: 

```typescript
try {
  onSubmit();
} catch (error) {
  if (error instanceof NetworkError) {
    setErrorMessage('Network error occurred. Please try again.');
  } else if (error instanceof ValidationError) {
    setErrorMessage('Validation error: ' + error.message);
  } else {
    setErrorMessage('An unexpected error occurred.');
  }
}
```

k-g-j:
Passing initial manually testing! Good job 😄 

https://github.com/user-attachments/assets/904a11fd-14e4-412a-8c76-4f7819ea5b5c

app/util/test/accountsControllerTestUtils.ts
owencraston: I am going off of the [proposed name of the test](https://github.com/MetaMask/snap-simple-keyring/blob/main/packages/snap/snap.manifest.json#L4C20-L4C48) snap which is `MetaMask Simple Snap Keyring`.

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
owencraston: I tried to get proper typing working for this but it did not work since the response from the PermissionController is typed as unknown. Because of this even adding the Record<string, unknown> is going to require a cast of any or unknown since the return object could be {}. In this case you will get the error ```Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
  No index signature with a parameter of type 'string' was found on type '{}'```. Until we can get proper typing on this I say we leave it. At least with the TODO/eslint-disable we can easily find this messy code in an audit of all the any's in the codebase. If you feel strongly against this we can try to get it working.

app/components/Views/Snaps/components/KeyringAccountListItem/KeyringAccountListItem.tsx
owencraston: yes good catch.

app/components/Views/Snaps/KeyringSnapRemovalWarning/test/KeyringSnapRemovalWarning.test.tsx
owencraston: Thank you

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
owencraston: I like this suggestion however in practice it does not work. The issue here is that useMemo doesn't handle async functions in the way we might expect. When you pass an async function to useMemo, it doesn't await the result. Instead, it memoizes the Promise returned by the async function. This results in a return type of Promise<> for the keyringAccounts so we would need to handle the .then of that value all over this file. From what I can see the useeffect + state is the best way to handle this. Let me know if you find a solution. 
Screenshot 2024-10-07 at 8 06 48 PM

app/components/Views/Snaps/KeyringSnapRemovalWarning/KeyringSnapRemovalWarning.tsx
owencraston: Based on the extension implementation and designs it seems like we do not  want to render "random" error messages. Instead we render a pre defined error message that can be translated. Instead of setting the error message to be displayed, I will add some logs into this try/catch for better debugging.

Screenshot 2024-10-07 at 8 11 34 PM

k-g-j (dismissed):
Thank you for addressing my comments! I appreciate the changes 😄 

### Manual Testing

**with account snap**

https://github.com/user-attachments/assets/9280188d-aded-4fb9-a564-01d644d85734

**remove a snap account without snap accounts in the wallet**

https://github.com/user-attachments/assets/cdf7c55c-9a3e-4bae-9771-0db653223c7c

app/components/Views/Snaps/components/KeyringAccountListItem/KeyringAccountListItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Text, {
+  TextColor,
+  TextVariant,
+} from '../../../../../component-library/components/Texts/Text';
+import { strings } from '../../../../../../locales/i18n';
+import {
+  KEYRING_ACCOUNT_LIST_ITEM,
+  KEYRING_ACCOUNT_LIST_ITEM_BUTTON,
+} from './KeyringAccountListItem.constants';
+interface KeyringAccountListItemProps {
```
gantunesr, line 24: nit

```suggestion
} from './KeyringAccountListItem.constants';

interface KeyringAccountListItemProps {
```

app/components/Views/Snaps/components/KeyringAccountListItem/KeyringAccountListItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+  account: InternalAccount;
+  snapUrl: string;
```
gantunesr, line 26: The name of this attribute confused me on first inspection, I thought it was a link to the snap dapp. What do you think about renaming it to `blockExplorerUrl`?

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
gantunesr: An alternative could be wrapping the method inside a useCallback and that invoke it inside the useEffect, something like,

```ts
const fetchIdentities = useCallback(async () => {
  // ...
}, [...]);

useEffect(() => {
  if (isKeyringSnap) {
    fetchIdentities();
  }
}, [fetchIdentities, isKeyringSnap]);
```

locales/languages/en.json
```diff
@@ -1035,14 +1035,30 @@
...
         "remove_snap_section_title": "Remove Snap",
         "remove_snap_section_description": "This action will delete the snap, its data, and its granted permissions.",
-        "remove_button_label": "Remove {{snapName}}"
+        "remove_button_label": "Remove {{snapName}}",
+        "remove_account_snap_warning": {
+          "title": "Remove Snap",
+          "description": "Removing this Snap removes these accounts from MetaMask:",
+          "remove_account_snap_alert_description_1": "Type ",
+          "remove_account_snap_alert_description_2": " to confirm you want to remove this snap:",
+          "banner_title": "Be sure you can access any accounts created by the Snap via private keys before removing them",
```
gantunesr, line 1044: Is this copy correct? In extension we have: `Be sure you can access any accounts created by this Snap on your own before removing it`

app/components/Views/Snaps/components/KeyringAccountListItem/KeyringAccountListItem.tsx
owencraston: this is better. I made the change in the latest commit.

locales/languages/en.json
owencraston: @eriknson @AlexJupiter and I changed it based on these discussions in slack. The main reason for this is that we do not use tooltips on mobile (since there is no hover event) so we needed the message to be more clear.

app/core/SnapKeyring/utils/getAccountsBySnapId.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Get the addresses of the accounts managed by a given Snap.
+ *
+ * @param snapId - Snap ID to get accounts for.
+ * @returns The addresses of the accounts.
+ */
+export const getAccountsBySnapId = async (
+  snapId: SnapId,
+) => {
+  const snapKeyring: SnapKeyring = await Engine.getSnapKeyring() as SnapKeyring;
+  return await snapKeyring.getAccountsBySnapId(snapId);
```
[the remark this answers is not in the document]
owencraston: I couldn't find an easy way to do this since the response object from the keyring method we are using to get this information is typed as unknown. 

Screenshot 2024-10-09 at 5 30 54 PM

Jonathansoufer (dismissed):
Agreed w/ Guto's point on comments. Once its addressed, you're good to go. Amazing work as usual!

owencraston:
Addressed comments and it is still working as expected.

Simulator Screenshot - iPhone 16 Pro - 2024-10-09 at 17 35 40
Simulator Screenshot - iPhone 16 Pro - 2024-10-09 at 17 35 44

Jonathansoufer (dismissed):
Reapproving after comments addressed

Jonathansoufer (dismissed):
Reapproved.

locales/languages/en.json
gantunesr: the problem I see with this copy is that not all accounts will have a private key. For example, SCA will be controlled by a smart contract and MPC accounts will rely on private key shards

app/core/SnapKeyring/utils/getAccountsBySnapId.ts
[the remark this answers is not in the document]
gantunesr: hmm we should improve the return type of that method to not be `unknown[]`. Not a blocker for this PR.

gantunesr (dismissed):

locales/languages/en.json
gantunesr: @owencraston after discussing with @eriknson and @AlexJupiter it was decided to keep the extension copy and open a new issue to improve this UX to provide a better flow for us to understand how to backup their account according to the type.

k-g-j (dismissed):

locales/languages/en.json
owencraston: okay I made those changes to the copy in the latest commit. Here is a screenshot

Simulator Screenshot - iPhone 16 Pro - 2024-10-10 at 13 52 39

owencraston:
latest copy version compared to extension

Screenshot 2024-10-10 at 3 26 46 PM

Jonathansoufer (dismissed):
Re-approved due copy change.

k-g-j (dismissed):

plasmacorral:
A few observations that suggest this may not be ready to merge:

1. when more than 3 accounts are present then the title and warning are shifted up and not presented on screen 
<div style="display: inline-block"></div>

2. When more than 4 snap accounts are present (will vary based on screen size), the user cannot see or interact with the `Continue` CTA in the remove flow, due to the lack of scrolling support
RECORDING

3. When I disable the snap before removing it, the accounts are not removed.  If I reinstall the snap, I am stuck with multiple accounts in the picker that are not seen or supported with SSK v1.1.6
RECORDING
4. Long press and hold in the accounts picker does not offer the remove account option like we see for imported accounts.  
RECORDING

plasmacorral:
Looking at this again before I log off.  1 and 2 above are pretty much the same, but will leave them as-is to avoid creating confusion.

owencraston:
@plasmacorral addressed observation 4. Removing the snap now removes the accounts as well. Good catch.


https://github.com/user-attachments/assets/04b259f8-292c-4e86-b9e8-1a701e51276e

owencraston:
@plasmacorral I addressed the issue with the overflowing accounts list mentioned in observation 1 & 2.

https://github.com/user-attachments/assets/744f74b5-4d16-4601-ae3d-77514ef503d9

owencraston:
@mikesposito I have enabled removing individual snap accounts via the three dot menu. I made the copy the same as the copy on extension. 

https://github.com/user-attachments/assets/5350b42d-287a-4f80-9cac-abc409752f25

Copy comparison:

Screenshot 2024-10-17 at 3 47 16 PM

Simulator Screenshot - iPhone 16 Pro - 2024-10-17 at 15 47 00

owencraston:
@mikesposito I tested removing the snap/accounts when the snap is disabled and confirmed that we can remove the snap even when it is disabled. This matches the extension. 

https://github.com/user-attachments/assets/446c871c-f1b0-46d0-99c9-9fcdfdc1fbe4

k-g-j (dismissed):
LGTM! Good work.

preinstalled snap
https://github.com/user-attachments/assets/e7b1266c-d32a-4da5-8904-df3ed546e62b

with account snap
https://github.com/user-attachments/assets/f4d5748f-73e1-4323-b983-4c6639560fa6

without account snap
https://github.com/user-attachments/assets/c000d7d7-2159-450e-8ede-6376e585a3f0

plasmacorral:
Took a fresh look at commit bc77790657bf0a5ac29801e2fe5a71cb2cbab6cb on Android 14 with pixel 5a.

Blocking observation remaining-
6. Although the `Remove Snap` title is not pushed off screen now, I am unable to see any more than 2 of the accounts that I am being warned about removing.

See [recording](https://www.loom.com/share/f4ee5568f253411fa47337c330e79c06)

I think that should be resolved before we merge as the user could remove accounts that were not disclosed on-screen in the warning.

Not blocking-
7. The 3 dots to remove the account works well, thank you!  Although I was surprised to see that when I removed the selected account I was always defaulted to Account 1 being selected instead of the next account up.  When I long press and remove an account imported by private key the active account selected is the one immediately above the removed account.  I mention it because it felt odd, but I think may be of no consequence to the user.

Not blocking-
8. The copy being used for removing a single account could still probably use some finesse.  It mentions SRP or private key for this "imported" account.  Since some snap accounts may not have an SRP or a private key (like MPC), we may want to revisit this on both clients.  At minimum we should be consistent in refering to it as a snap account, rather than calling it imported.  That can be handled in a follow-on and should not delay this PR.

Not blocking/Not related to snap accounts-
9. Only now do I appreciate that an imported account cannot be removed from the 3 dots menu in wallet view.  This makes it now the outlier as a removable account that can only be removed through a long press and hold from the accounts picker.  We should at least add the option to remove under 3 dots menu while we consider further consistency with the tap and hold to remove pattern from the account picker.

owencraston:
@plasmacorral Thank you for those observations. I had only tested on iOS and did not notice the issue there. Turns out there were a few extra things I needed to do in order to get the scrolling on android to work inside of an animated view. I also tested on an iPhone mini and the heights/scrolling seem to work well.

Here is a video showing that observation 6 is resolved:

https://github.com/user-attachments/assets/747e176a-0480-492a-998f-681cc7f24a1e

Regarding observation 7. I included that piece of logic to set the selected account to the first account since that was the same logic we used to removing a hardware account from the 3 dot menu. However I do agree this is a weird pattern and differs from the long press logic so I removed that. Now the selected account will be account at index -1. 

Here is a video showing this behaviour.

https://github.com/user-attachments/assets/ea49ad95-5ec0-407d-b523-e0d792641502

Re observation 8. I agree this is not great copy however in order to be consistent with the extension I left it the same. we should probably overhaul this in the future.

Re observation 8: It seems that our removing account logic is all over the place for all sort of account types. Thank you for making that issue so we can resolve it in an isolated change. This will likely need design feedback.

plasmacorral:
Based in video posted [here](https://github.com/MetaMask/metamask-mobile/pull/11451#issuecomment-2428120669) it appears that the remaining blocker was resolved.  Moving this to QA pass.

k-g-j (approved):

app/components/Views/AccountActions/AccountActions.tsx
```diff
@@ -189,6 +194,49 @@ const AccountActions = () => {
     }
   }, [controllers.KeyringController]);
 
+  ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+
+  /**
+   * Remove the snap account from the keyring
+   */
+  const removeSnapAccount = useCallback(async () => {
```
Jonathansoufer, line 202: Is there any chance to a "dirty" state occurs here? Meaning, if something goes wrong while removing the account (through `KeyringController.removeAccount`, how we should handle and inform our users? Should be an "error" state somewhere?

owencraston: This is a valid concern. I dont think we have any designs or patterns for this now. If we look at the other methods for removing accounts they are identical and have no given us issues for a long time. Take removing a hardware account for example.

```
/**
   * Remove the hardware account from the keyring
   * @param keyring - The keyring object
   * @param address - The address to remove
   */
  const removeHardwareAccount = useCallback(async () => {
    if (selectedAddress) {
      await controllers.KeyringController.removeAccount(selectedAddress as Hex);
      await removeAccountsFromPermissions([selectedAddress]);
      trackEvent(MetaMetricsEvents.ACCOUNT_REMOVED, {
        accountType: keyring?.type,
        selectedAddress,
      });
    }
  }, [
    controllers.KeyringController,
    keyring?.type,
    selectedAddress,
    trackEvent,
  ]);
```

If we were take this on I think it should occur in its own isolated change since it would require some design thought.

owencraston: @plasmacorral created an [issue](https://github.com/MetaMask/accounts-planning/issues/653) to improve the consistency of removing accounts, perhaps we could take this on in that issue.

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
```diff
@@ -70,53 +81,118 @@ const SnapSettings = () => {
...
+      removeSnap();
+    }
+  }, [isKeyringSnap, keyringAccounts.length, removeSnap]);
+
+
+  const handleRemoveSnapKeyring = useCallback(() => {
+    try {
+      setIsShowingSnapKeyringRemoveWarning(true);
+      removeSnap();
+      setIsShowingSnapKeyringRemoveWarning(false);
```
Jonathansoufer, line 133: Not a big deal, but isn't this redundant if you're using `finally` anyway?

Jonathansoufer (approved):
Left some minor comments, but excellent work overall!

tommasini (approved):

Cal-L (approved):
Lgtm!

vinnyhoward (approved):

PR #11453, opened 2024-09-26
feat: Add signature tracing

OGPoyraz:
## **Description**

- Adding `createTracingMiddleware` to start tracing for defined type of messages.
- Adding `trace` callback to `SignatureController`.
- End `Notification Display` trace after signature confirmation landed on users screen.

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/1883

## **Manual testing steps**

1. Set `MM_SENTRY_DSN_DEV` to developer Sentry dsn, see the value here https://github.com/MetaMask/metamask-extension/blob/develop/.metamaskrc.dist
2. Go test-dapp in app, trigger and personal signature
3. If you used the mentioned dsn above, your `Signature` trace will appear on this link: https://metamask.sentry.io/traces/?project=273496&query=signature%3ATransaction&source=traces&statsPeriod=1h


pedronfigueiredo (dismissed):

vinistevam (dismissed):

jpuri (dismissed):

app/components/Approvals/SignatureApproval/SignatureApproval.tsx
```diff
@@ -14,6 +16,11 @@ const SignatureApproval = () => {
     });
   }, [onConfirm]);
 
+  useAsyncResult(async () => await endTrace({
```
matthewwalsh0, line 19: Why do we need `useAsyncResult` here as `endTrace` isn't asynchronous?

app/components/Approvals/SignatureApproval/SignatureApproval.tsx
```diff
@@ -14,6 +16,11 @@ const SignatureApproval = () => {
...
+      name: TraceName.NotificationDisplay,
+      id: approvalRequest?.requestData?.requestId,
+    }), [approvalRequest?.requestData?.requestId]);
```
matthewwalsh0, line 22: Is it worth defining a local variable such as `signatureRequestId`?

app/core/RPCMethods/RPCMethodMiddleware.ts
```diff
@@ -659,7 +680,12 @@ export const getRpcMethodMiddleware = ({
       eth_signTypedData_v4: async () => {
         const data = JSON.parse(req.params[1]);
         const chainId = data.domain.chainId;
-        PPOMUtil.validateRequest(req);
+
+        trace(
```
matthewwalsh0, line 684: Maybe not in this PR, but I wonder if we'd avoid a lot of duplication by using a generic function for all the signing methods?

matthewwalsh0 (changes requested):

app/components/Approvals/SignatureApproval/SignatureApproval.tsx
OGPoyraz: That's right, fixed

app/components/Approvals/SignatureApproval/SignatureApproval.tsx
OGPoyraz: Done

app/core/RPCMethods/RPCMethodMiddleware.ts
OGPoyraz: I tried that in the beginning but it was looking load of changes so didn't proceed. I definitely agree that it's worth doing some cleaning up on these rpc methods.

matthewwalsh0 (approved):

app/core/createTracingMiddleware/index.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export default function createTracingMiddleware() {
+  return async function tracingMiddleware(
+    req: JsonRpcRequest<unknown> & { traceContext?: unknown },
+    _res: PendingJsonRpcResponse<unknown>,
+    next: () => void,
+  ) {
+    const { id, method } = req;
+
```
andreahaku, line 27: you can validate `id`and do an early return in case it's not valid:
```
if (typeof id !== 'string') {
      return next();
    }

File: app/core/createTracingMiddleware/index.ts [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+    const traceName = METHOD_TYPE_TO_TRACE_NAME[method];
+
```
andreahaku, line 29: same thing for `traceName`:
```
if (!traceName) {
      return next();
    }

File: app/core/createTracingMiddleware/index.ts [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+    if (traceName) {
```
andreahaku, line 30: you can parallelize trace calls as they are indipendent so to optimise response time:

```
const [traceContext] = await Promise.all([
      trace({
        name: traceName,
        id,
      }),
      trace({
        name: TraceName.Middleware,
        id,
        parentContext: req.traceContext,
      })
    ]);

    req.traceContext = traceContext;

**Review by andreahaku — COMMENTED** [review]
some optimisations suggested

File: app/core/createTracingMiddleware/index.ts (hunk 4 above)
Comment (OGPoyraz) [reply app/core/createTracingMiddleware/index.ts:27]: We still want this `trace` action to be called even if the `id` is not defined here.

File: app/core/createTracingMiddleware/index.ts (hunk 6 above)
Comment (OGPoyraz) [reply app/core/createTracingMiddleware/index.ts:30]: Unfortunately this doesn't work as we want, because subsequent `trace` call (second one) needs parent context to be settled.

File: app/core/createTracingMiddleware/index.ts (hunk 5 above)
Comment (OGPoyraz) [reply app/core/createTracingMiddleware/index.ts:29]: I like early returns but doing this will add duplication of `next` call. 

```
if(!traceName) {
  return next();
}

/// actions

next();
```

andreahaku (approved):
LGTM

tommasini (approved):

PR #11620, opened 2024-10-04
fix: Fix unwanted `IncompleteAssetDisplayed` events

OGPoyraz:
## **Description**

This PR aims to stop `IncompleteAssetDisplayed` events appearing in the mix panel even the token is known by the API.

For more information please see the mixpanel link https://mixpanel.com/project/1979229/view/132079/app/events#H35NJuByTVqi

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/3362

## **Manual testing steps**

No QA needed

### DIFF
--- a/app/components/Views/confirmations/components/TransactionReview/index.js
+++ b/app/components/Views/confirmations/components/TransactionReview/index.js
@@ -593,7 +593,7 @@ class TransactionReview extends PureComponent {
                       primaryCurrency={primaryCurrency}
                       chainId={chainId}
                     />
-                    {useTransactionSimulations && (
+                    {useTransactionSimulations && transactionSimulationData && (
                       <View style={styles.transactionSimulations}>
                         <SimulationDetails
                           simulationData={transactionSimulationData}
--- a/app/components/UI/SimulationDetails/useBalanceChanges.ts
+++ b/app/components/UI/SimulationDetails/useBalanceChanges.ts
@@ -98,13 +98,12 @@ const fetchTokenExchangeRates = async (
   chainId: Hex,
 ) => {
   try {
-    const x = await fetchTokenContractExchangeRates({
+    return await fetchTokenContractExchangeRates({
       tokenPricesService: new CodefiTokenPricesServiceV2(),
       nativeCurrency,
       tokenAddresses,
       chainId,
     });
-    return x;
   } catch (err) {
     return {};
   }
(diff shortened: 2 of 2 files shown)


app/components/Views/confirmations/components/TransactionReview/index.js
```diff
@@ -593,7 +593,7 @@ class TransactionReview extends PureComponent {
                       primaryCurrency={primaryCurrency}
                       chainId={chainId}
                     />
-                    {useTransactionSimulations && (
+                    {useTransactionSimulations && transactionSimulationData && (
```
OGPoyraz, line 596: The issue arises due to a race condition in the `useBalanceChanges` hook. Initially, the hook is invoked with empty balance changes (an empty array). Subsequently, `simulationData` gets populated and the hook is called again. In the interim, the empty balance changes resolve and set `pending` to `false`. This results in the `IncompleteAssetEvent` being triggered because it mistakenly interprets the new contract information as having resolved.

I reviewed the `SimulationDetails` component and `simulationData` in the extension. It appears that `SimulationDetails` is never rendered when `simulationData` is undefined. This is likely because the `transactions` state changes do not occur immediately, unlike in the mobile app where the transactions `simulationData` is not defined in the beginning of the confirmation.

Please refer to the logs I added at the top of `SimulationDetails` in the extension, which show that `simulationData` is always defined.
Screenshot 2024-10-04 at 09 09 50

This fix is just imitating the behaviour of extension and fixes that race condition.

pedronfigueiredo (approved):

matthewwalsh0 (approved):

app/components/Views/confirmations/components/TransactionReview/index.js
EtherWizard33, line 596: nit: 'use' is usually reserverd for hooks, could be, areTransactionSimulationsUsed

EtherWizard33 (approved):

app/components/Views/confirmations/components/TransactionReview/index.js
OGPoyraz: The naming comes from preferences controller property so I don't think it make sense to change this at this point to keep the property name parities between extension and mobile.

It's not `reserved` phonetically it sounds like a hook, so I agree that it would've been named better.

PR #11714, opened 2024-10-09
chore(docs): Update ruby install instructions in environment setup readme

MajorLift:
## **Description**

The current ruby install instructions in the "Environment Setup" readme are vague, and may cause `gem` to be run with the system root installation.

Screenshot 2024-10-09 at 9 13 29 AM
> https://consensys.slack.com/archives/C02U025CVU4/p1728496388117329

This commit updates the instructions with commands that fix this issue.

### DIFF
--- a/docs/readme/environment.md
+++ b/docs/readme/environment.md
@@ -23,7 +23,16 @@ MacOS ships with an old ruby version that is incompatible with this project
 
 It is recommended to install a Ruby version manager such as [rbenv](https://github.com/rbenv/rbenv?tab=readme-ov-file#installation)
 
+```bash
+brew install rbenv
+```
+
 Install ruby version defined in the file `.ruby-version`
+
+```bash
+rbenv install $(cat .ruby-version)
+rbenv shell $(cat .ruby-version)
+```
 
 ### Gems
 
--- a/docs/readme/environment.md
+++ b/docs/readme/environment.md
@@ -30,6 +30,10 @@ brew install rbenv
 Install ruby version defined in the file `.ruby-version`
 
 ```bash
+rbenv install
+rbenv local
+
+# If the above fails, try this:
 rbenv install $(cat .ruby-version)
 rbenv shell $(cat .ruby-version)
 ```
--- a/docs/readme/environment.md
+++ b/docs/readme/environment.md
@@ -23,20 +23,7 @@ MacOS ships with an old ruby version that is incompatible with this project
 
 It is recommended to install a Ruby version manager such as [rbenv](https://github.com/rbenv/rbenv?tab=readme-ov-file#installation)
 
-```bash
-brew install rbenv
-```
-
 Install ruby version defined in the file `.ruby-version`
-
-```bash
-rbenv install
-rbenv local
-
-# If the above fails, try this:
-rbenv install $(cat .ruby-version)
-rbenv shell $(cat .ruby-version)
-```
 
 ### Gems
 
--- a/docs/readme/environment.md
+++ b/docs/readme/environment.md
@@ -19,11 +19,18 @@ brew install watchman
 
 ### Ruby
 
-MacOS ships with an old ruby version that is incompatible with this project
+MacOS ships with an old ruby version that is incompatible with this project.
 
-It is recommended to install a Ruby version manager such as [rbenv](https://github.com/rbenv/rbenv?tab=readme-ov-file#installation)
+It is recommended to install a ruby version manager, and use it to install the ruby version defined in the file `.ruby-version`.
 
-Install ruby version defined in the file `.ruby-version`
+For example, the following commands will [install `rbenv`](https://github.com/rbenv/rbenv?tab=readme-ov-file#installation) and use it to [install the configured ruby version](https://github.com/rbenv/rbenv#installing-ruby-versions):
+
+```bash
+brew install rbenv
+rbenv init
+rbenv install
+rbenv local  # if installation is successful, this will print the installed ruby version.
+```
 
 ### Gems
 


mikesposito (dismissed):

docs/readme/environment.md
```diff
@@ -23,8 +23,17 @@ MacOS ships with an old ruby version that is incompatible with this project
...
 
+```bash
+brew install rbenv
+```
+
 Install ruby version defined in the file `.ruby-version`
 
+```bash
+rbenv install $(cat .ruby-version)
+rbenv shell $(cat .ruby-version)
```
MajorLift, line 34:
```suggestion
rbenv install
rbenv local

# If the above fails, try this:
rbenv install $(cat .ruby-version)
rbenv shell $(cat .ruby-version)
```

joaoloureirop:
It is expected that a MetaMask contributor picks a version manager of their choice, and follows the version manager documentation to install and configure it on their system.

`rbenv` is mentioned as an example of a ruby version manager

We should keep from being prescriptive on how to configure tools that are not maintained by us.

MajorLift:
@joaoloureirop Would it help to add language emphasizing that the `rbenv` commands are examples? Or perhaps we could add more links to sections within the `rbenv` readme?

The primary target of this readme section probably isn't developers who have already set up ruby version managers themselves and don't need the instructions. Shouldn't unblocking non-users of Ruby be a priority?

legobeat:
FWIW there is also the option of the repository [`Dockerfile`](https://github.com/MetaMask/metamask-mobile/blob/main/scripts/docker/Dockerfile):

```
# build image with bundled node/ruby/etc
$ podman build --build-arg UID=$(id -u) --build-arg GID=$(id -g) -t localhost/mm-mobile-builder -f scripts/docker/Dockerfile .

# run shell in container
$ podman run --rm -it --user $(id -u):$(id -g) --userns=keep-id:uid=$(id -u),gid=$(id -g) -v "$(pwd)":/app -w /app -v ~/.yarn:/home/node/.yarn -v ~/.cache/yarn:/home/node/.cache/yarn mm-mobile-build:debian bash

# run ruby directly
$ podman run --rm -it --user $(id -u):$(id -g) --userns=keep-id:uid=$(id -u),gid=$(id -g) -v "$(pwd)":/app -w /app mm-mobile-build:debian ruby --version
```

(taken to the next level: https://github.com/legobeat/l7-devenv)

> Shouldn't it be a priority to unblock non-users of ruby by using the most obvious and unambiguous language possible?

I think it should also be a priority to steer developers towards secure solution and setups they can control. (No shade on rbenv or using it the proposed way)

I do think there are risks with the approach, though, specifically considering the  undeterministic mess that CocoaPods have become.. By steering towards a specific blessed setup, we kind of take responsibility for that being Good Enough.

So kind of agree with @joaoloureirop

> Would it help to add language emphasizing that the rbenv commands are examples? Or perhaps we could add more links to sections within the rbenv readme?

Both sound like nice ideas to me!

joaoloureirop:
> @joaoloureirop Would it help to add language emphasizing that the `rbenv` commands are examples? Or perhaps we could add more links to sections within the `rbenv` readme?
> 

We could expand this section to offer better guidance to new contributors unfamiliar with the tool. 
The challenge is how to keep it relevant to all the options available

> The primary target of this readme section probably isn't developers who have already set up ruby version managers themselves and don't need the instructions. Shouldn't it be a priority to unblock non-users of ruby by using the most obvious and unambiguous language possible?

The concern is with new contributors getting familiar with this tool since it will require intervention every time our `.ruby-version` changes.

We want to help new contributors as much as possible, and in this case, the best help redirecting them to the official documentation.

For example, if a new contributor decides to use the version manager mentioned in our docs they can follow the hyperlink provided and follow the installation guide, which is only two steps when using a package manager. 
**The most important part that is often overlooked is to setup the shell to load rbenv.**

Down the road when intervention is needed (install new ruby version, rehashing, removing...) , the official documentation is the best help they can get.

docs/readme/environment.md
```diff
@@ -19,11 +19,18 @@ brew install watchman
...
 
-Install ruby version defined in the file `.ruby-version`
+For example, the following commands will [install `rbenv`](https://github.com/rbenv/rbenv?tab=readme-ov-file#installation) and use it to [install the configured ruby version](https://github.com/rbenv/rbenv#installing-ruby-versions):
+
+```bash
+brew install rbenv
+rbenv init
+rbenv install
+rbenv local  # if installation is successful, this will print the installed ruby version.
+```
```
MajorLift, line 33: @joaoloureirop @legobeat Thanks for the feedback! I understand the need to avoid endorsing a specific version manager, but feel like the reservations around providing instructions...

> it will require intervention every time our .ruby-version changes.
> Down the road when intervention is needed (installing a new ruby version, rehashing, removing...), the official documentation is the best help they can get.

... might apply to any external package we don't manage but provide instructions for in our readmes.

I attempted an edit but I'm not sure how well it addresses your concerns. If either of you have suggestions around adding stronger disclaimer language or editing/removing the additions I'd be happy to apply them.

joaoloureirop: > ... might apply to any external package we don't manage but provide instructions for in our readmes.

Instructions should be provided in our readmes when the configuration differs from the default or when the official docs miss the installation/configuration guide.
Other than that, official docs should be the single source of truth.

the node version manager section is as vague as the ruby version manager section for the same reason.

joaoloureirop: What is the added value here that the official docs don't provide?

joaoloureirop: What is the added value here that the official docs don't already provide?

MajorLift: Providing immediate, explicit, unambiguous instructions would hopefully serve as a better first line of defense against troubleshooting threads like these, some of them quite huge:
- https://consensys.slack.com/archives/C02U025CVU4/p1729086334814599?thread_ts=1728571415.206109&cid=C02U025CVU4
- https://consensys.slack.com/archives/C02U025CVU4/p1724357838719499?thread_ts=1717610381.685139&cid=C02U025CVU4
- https://consensys.slack.com/archives/C02U025CVU4/p1722369858678939?thread_ts=1722019463.898419&cid=C02U025CVU4
- https://consensys.slack.com/archives/CBW7S9FSN/p1695241220126889?thread_ts=1695228967.143599&cid=CBW7S9FSN
- https://consensys.slack.com/archives/C02U025CVU4/p1728496388117329

If our internal developers (myself included) have this much trouble with this step I don't think the situation will be better for external contributors.

In the end I guess this comes down to the question of what's a higher priority for a readme: being correct or being helpful.

PR #11753, opened 2024-10-11
feat: initial setup for not globally selected chain ID (`useChainId` hook)

andreahaku:
## **Description**

Initial test support for not globally selected chain ID applied to assets (token and NFTs) lists.

This PR creates a React hook (`useChainId`) that replaces the `selectChainId` selector used to get the globally selected chain Id.

Currently, the network badge for assets is derived from the globally selected chain ID in the app. This global chain ID is shimmed via the new hook, effectively making it equal to the chain ID selected for the entire session or dapp. However, with the support of the hook and by adding the support of the chain ID at the single component (token and NFT) level we can move towards supporting a per-dapp chain ID.

## **Related issues**

Fixes: https://github.com/orgs/MetaMask/projects/85/views/2?pane=issue&itemId=82943990

## **Related PRs**
- https://github.com/MetaMask/metamask-mobile/pull/11753 (this one)
- https://github.com/MetaMask/metamask-mobile/pull/11781
- https://github.com/MetaMask/metamask-mobile/pull/11811
- https://github.com/MetaMask/metamask-mobile/pull/11813
- https://github.com/MetaMask/metamask-mobile/pull/11837

## **Manual testing steps**

1. Go to this page...
2.
3.

### **Before**
As **After**

### **After**

https://github.com/user-attachments/assets/adb84915-11f2-42d2-aad9-0b9ae088c19e

Simulator Screenshot - iPhone 12 Pro - 2024-10-11 at 14 25 55
Simulator Screenshot - iPhone 12 Pro - 2024-10-11 at 14 25 38
Simulator Screenshot - iPhone 12 Pro - 2024-10-11 at 14 25 26


sahar-fehri:
LGTM!

sahar-fehri (dismissed):

app/selectors/hooks.ts
```diff
@@ -0,0 +1,10 @@
+import { useSelector } from 'react-redux';
+import { selectChainId } from '../selectors/networkController';
+
+export const useChainId = () => {
```
matthewwalsh0, line 4: I understand the convenience of having this as re-usable hook, but does this not still couple our UX to a global or semi-global chain ID?

What logic do we predict could populate this?

What if we want to render two components at once on alternate chain IDs for example?

I can't speak for the rest of the mobile UX, but the pattern we're using in confirmations is relying on the subject or data of the confirmation as a source of truth.

For example, since signatures and transactions are created on a specific chain, so we can persist that in the controller state and read that chain where needed in the confirmation components.

andreahaku: this will be able to return the selected chain ID from the of permitted chain ID (and the list of permitted chain IDs if needed) on the first phase.

When per-dapp network permissions will be ready, this will return the per-dapp selected chain ID and potentially also the list of the per-dapp permitted networks if needed.

This is the logic behind the creation of this hook

andreahaku: it can return 'per-dapp' chain Id or the mobile app chain Id coming not a specific dapp but a general state

andreahaku: and in theory components (like token lists and/or NFTs list) can iterate and show assets from all the permitted networks this way as what I've been doing is to have single elements of those lists have a "chainId" prop.

matthewwalsh0: But won't that excessively couple us to a dApp and a selected chain, where as in the final form of multi-chain, there won't be dApp selected chains, only requests on alternate chains.

If we treat the chain ID as a global or context based thing, does that not hinder the readability and dynamic nature of where the chain ID is coming from?

If we're looking at a signature or transaction, let's just look at its saved chain ID.

If we're viewing a NFT or token, they will have persisted chain ID also?

Why store it in a context at all, if we can make it explicit with `props` or by reading it from the relevant state as needed?

PR #11801, opened 2024-10-15
fix: accounts not syncing between devices bug

Jonathansoufer:
## **Description**

This PR fixed an issue preventing mobile accounts to sync its states (enable/disable) with other clients (platform and extension). [NOTIFY-1218](https://consensyssoftware.atlassian.net/browse/NOTIFY-1218?atlOrigin=eyJpIjoiZTRlYTlkNjVjOGVjNDAyYTljNmM3YTg2NjBiYzllNGYiLCJwIjoiaiJ9)

## **Related issues**

Fixes:

## **Manual testing steps**

1. Import the same SRP on Mobile
2. Make sure all accounts are visible on both devices/platforms. If not, import all accounts
3. Switch Notifications state in account "A" on extension.
4. Go to Notifications Settings on mobile and see if the state is propagated.
5. Switch back on mobile
6. Go to extension and see if state is propagated. 

### **After**

https://github.com/user-attachments/assets/860e2e46-f08a-45fd-b965-a9948ef59630

[NOTIFY-1218]: https://consensyssoftware.atlassian.net/browse/NOTIFY-1218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Prithpal-Sooriya (dismissed):

mathieuartu (dismissed):

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -111,58 +87,24 @@ export function useRefetchAccountSettings(
...
-      setData(res);
-      dispatch(updateAccountState(res));
-    } catch {
-      setError('Failed to get account settings');
-    } finally {
-      setLoading(false);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  const memoAccounts = useMemo(() => accounts.map((account) => account.address),[accounts]);
```
Daniel-Cross, line 94: Question around this usage of `useMemo`, (and the `useCallback` in this file).

I know you didn't implement it but is it needed? 

If we're not having performance issues around these functions then it can be removed. We have a lot of `useMemo` and `useCallback` usage in the app and we're allocating device memory to each one. Removing any unnecessary usage would be great.

Daniel-Cross:
Looks good @Jonathansoufer, just a couple of comments.

Also, what's the purpose of the `skip-sonar-cloud` label? I see you're hitting the quality gate threshold, I would prefer it if we ran as many checks as possible against all changes.

app/util/notifications/hooks/useUpdateAccountSetting.tsx
```diff
@@ -13,15 +13,18 @@ export function useUpdateAccountSetting(address: string, refetchAccountSettings:
...
       setLoading(true);
       try {
+        // change the account state in the controller
         await switchAccountNotifications([address], state);
-        refetchAccountSettings();
-        refetch();
+        // refetch the account settings from the controller
+        await updateAndfetchAccountSettings();
+        // refetch the notifications from the controller
+        await refetch();
```
salimtb, line 21: Is it necessary to execute this sequentially, or can it be done in parallel using Promise.all? Doing it in parallel could potentially improve performance.

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -111,58 +87,24 @@ export function useRefetchAccountSettings(
...
-    } finally {
-      setLoading(false);
+        Engine.context.NotificationServicesController.checkAccountsPresence(
+          memoAccounts,
+        ).then((result) => {
+          dispatch(updateAccountState(result));
+          return result;
+        });
+    } catch (err) {
+      Logger.log(err, 'Failed to get account settings');
```
salimtb, line 105: I noticed the use of then() alongside a try-catch block for handling the asynchronous function. To maintain consistency and clarity, it might be a good idea to choose between using one of the following approaches:
Using then() and catch(): This will maintain the current promise-chaining style:
```
Engine.context.NotificationServicesController.checkAccountsPresence(memoAccounts)
  .then((result) => {
    dispatch(updateAccountState(result));
    return result;
  })
  .catch((err) => {
    Logger.log('Failed to get account settings:', err);
  });
```
Using async/await with try-catch: This will make the asynchronous handling more streamlined:

```
try {
  const result = await Engine.context.NotificationServicesController.checkAccountsPresence(memoAccounts);
  dispatch(updateAccountState(result));
  return result;
} catch (err) {
  Logger.log('Failed to get account settings:', err);
}
```
Both methods are valid, but it's important to stick to one approach for consistency

salimtb:
great work @Jonathansoufer , i just left two small comments regarding consistency and perf improvement

app/components/Views/Settings/NotificationsSettings/index.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable react-native/no-inline-styles */
```
tommasini, line 1: NIT: I think we can remove the inline style from this component, wdyt?

app/core/redux/slices/notifications/index.ts
```diff
@@ -24,8 +24,12 @@ const slice = createSlice({
       state,
       action: PayloadAction<NotificationsAccountsState>,
     ) => {
-      if (JSON.stringify(state) !== JSON.stringify(action.payload)) {
-        return action.payload;
+      // Check if the payload is not an empty object
+      if (Object.keys(action.payload).length > 0) {
+        // Only update if there are differences
+        if (JSON.stringify(state) !== JSON.stringify(action.payload)) {
```
tommasini, line 30: What do you think about using the `isEqual` function of lodash instead? Any strong opinions on why JSON.stringify would be better? 

Interesting, I found a [benchmark](https://www.measurethat.net/Benchmarks/Show/4089/0/lodashisequal-vs-jsonstringify-equality-comparison-for#latest_results_block) that shows that the stringify comparison is quicker!

```suggestion
        if (isEqual(state, action.payload)) {
```

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -1,20 +1,21 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 /* eslint-disable import/prefer-default-export */
```
tommasini, line 2: I believe both this disable rules can be removed
```suggestion

```

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -1,20 +1,21 @@
...
-  selectIsMetamaskNotificationsEnabled,
-  selectIsUpdatingMetamaskNotificationsAccount,
-} from '../../../selectors/notifications';
+import { useDispatch } from 'react-redux';
+
 import { updateAccountState } from '../../../core/redux/slices/notifications';
+import { Account } from '../../../components/hooks/useAccounts/useAccounts.types';
+import Logger from '../../../util/Logger';
+
+jest.mock('../../../util/Logger');
```
tommasini, line 17: I believe we do not want to mock logger at this file
```suggestion

```

Jonathansoufer: Good catch. Removed.

app/util/notifications/hooks/useSwitchNotifications.ts
Jonathansoufer: Removed.

app/core/redux/slices/notifications/index.ts
Jonathansoufer: Good change, but think that the correct change would be `if (!isEqual(state, action.payload)) {`. Changed

app/components/Views/Settings/NotificationsSettings/index.tsx
Jonathansoufer: Sure. Fixed.

app/util/notifications/hooks/useUpdateAccountSetting.tsx
Jonathansoufer: Good one! Changed.

app/util/notifications/hooks/useSwitchNotifications.ts
Jonathansoufer: This functions are really performance sensitive since it's dealing with a user (and the UI) changing account settings for Notifications and some user can have really huge number of accounts. I'd rather open an issue to look (myself included) in other places where we could reduce the usage of memos and callbacks for sure.

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -111,58 +87,24 @@ export function useRefetchAccountSettings(
...
+}, [dispatch, memoAccounts]);
 
-  return {
-    data,
-    initialLoading: loading,
-    error,
-    accountsBeingUpdated,
-    update,
-  };
+  return { updateAndfetchAccountSettings };
```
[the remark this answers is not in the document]
Jonathansoufer: Replied above already.

app/components/Views/Settings/NotificationsSettings/AccountsList.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        accountSettingsData={{}}
+        updateAndfetchAccountSettings={updateAndfetchAccountSettings}
+        isUpdatingMetamaskNotificationsAccount={[]}
+      />,
+      {
+        state: mockInitialState,
+      },
+    );
+
+    expect(updateAndfetchAccountSettings).toHaveBeenCalled();
```
tommasini, line 86: Wondering if we should begin a pattern to make sure the calls that we want to happen only once, they happen only once
```
toHaveBeenCalledTimes(1)
```

wdyt?

tommasini (dismissed):
LGTM

app/components/Views/Settings/NotificationsSettings/AccountsList.test.tsx
Jonathansoufer: Good idea since we could also prevent multiple undesirable calls. Changed so we can start implementing this NOW!

tommasini (approved):

Prithpal-Sooriya (approved):

andreahaku (approved):
LGTM

PR #11763, opened 2024-10-11
feat(3448): header update

vinnyhoward:
## **Description**

The header is being updated according to this [Figma design](https://www.figma.com/design/aMYisczaJyEsYl1TYdcPUL/Portfolio-View?m=auto&node-id=5019-59596&t=PAdxL1bg2Mk08dSk-1) (second to last screen). 

Because `WalletAccount` is being removed, the more icon or overflow icon will be moved to the `AccountSelectorList`. Here is the [PR for more info](https://github.com/MetaMask/metamask-mobile/pull/11823)

Regression Tests Passing (11/04/24): https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/6224a4f4-1c55-43cf-a622-ec6d91f42110?tab=workflows

## **Related issues**

Ticket: [#3448](https://github.com/MetaMask/MetaMask-planning/issues/3448)

## **Manual testing steps**

1. Wallet Home Page
   - Open the account picker and switch between accounts
   - Verify functionality remains the same as before

2. Copy Address
   - Use the copy address feature in the top right
   - Confirm it works as expected

3. Network Switcher
   - Switch between different networks
   - Ensure all networks load correctly and balances update

4. Overflow (Edit Account): This is on https://github.com/MetaMask/metamask-mobile/pull/11823
   - Tap on account picker and a bottom sheet will appear 
   - Click on the overflow icon (the vertical more icon) and edit your account just like before

## **Screenshots/Recordings**

| Light  | Dark  |
|:---:|:---:|
|light_mode_header_update|dark_mode_header_update|

### **Before**

| Light  | Dark  |
|:---:|:---:|
|light_mode_original|dark_mode_original|

### **After**

| Light  | Dark  |
|:---:|:---:|
|light_mode_screenshot|dark_mode_screenshot|

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
[...]

PR #11851, opened 2024-10-17
feat: extend PickerBase component functionality

vinnyhoward:
## **Description**

This PR enhances the PickerBase component to provide more flexibility and control to developers, particularly in relation to the dropdown icon styling and positioning. These changes are necessary to accommodate the [Header Update](https://www.figma.com/design/aMYisczaJyEsYl1TYdcPUL/Portfolio-View?m=auto&node-id=5019-59596&t=PAdxL1bg2Mk08dSk-1) design requirements.

### Key Changes:

1. **Dropdown Icon Size Control**: Developers can now specify the size of the dropdown icon, with a fallback to ensure consistent behavior.

2. **Dropdown Icon Spacing**: Added the ability to control the spacing between the dropdown icon and other elements within the component.

3. **Backward Compatibility**: These changes are designed to be non-breaking, maintaining compatibility with existing implementations.

### Benefits:

- Increased flexibility for custom designs
- Better alignment with the new Header Update requirements
- Improved developer control over component styling

## **Related issues**

Related: [#11763](https://github.com/MetaMask/metamask-mobile/pull/11763)

## **Manual testing steps**

1. View any instance of `PickerBase` being used. For example the `NetworkPicker` in header in the home screen

## **Screenshots/Recordings**

| Before  | After  |
|:---:|:---:|
|original|updated|

### **Before**

NA

### **After**

NA


app/component-library/components/Pickers/PickerBase/PickerBase.types.ts
```diff
@@ -13,9 +14,20 @@ export interface PickerBaseProps extends TouchableOpacityProps {
...
    */
   children: React.ReactNode;
+  /**
+   * Icon size.
+   */
+  iconSize?: IconSize;
+  /**
+   * Dropdown icon styles.
+   */
+  dropdownIconStyle?: ViewStyle;
```
brianacnguyen, line 24: Can you point me to when these icon size and style needs to be changed?

brianacnguyen (commented):
Question regarding the arrow down icon changes

app/component-library/components/Pickers/PickerBase/PickerBase.types.ts
vinnyhoward: Yup! Its part of the redesign effort we have here. This is the screen I am working on (the second to last one) [header redesign](https://www.figma.com/design/aMYisczaJyEsYl1TYdcPUL/Portfolio-View?m=auto&node-id=5019-59596&t=PAdxL1bg2Mk08dSk-1)

And here is feedback from Amanda on what I [needed to change](https://www.figma.com/design/aMYisczaJyEsYl1TYdcPUL/Portfolio-View?node-id=5936-13830&t=A82DJ33SallzMj6N-4)

brianacnguyen: thank you. can you tell me why you decided to add another prop instead of just changing the size?

vinnyhoward: It is because I need control of the distance of the picker icon. Here it is without the props of `dropdownIconStyle` as you can see it doesn't match the Figma design and is much to far. Changing the styles of the children does nothing to change the distance of the picker
| Without  | With  |
|:---:|:---:|
|Screenshot 2024-10-22 at 11 29 37 AM|Screenshot 2024-10-22 at 11 33 44 AM|

(this screenshot is pulled from my original [PR](https://github.com/MetaMask/metamask-mobile/pull/11763) of the updated picker base in use)

Introducing the `dropdownIconStyle` doesn't change any UI since its default styles is what it was before but adds the flexibility to change it if needed

brianacnguyen: can you try passing gap to the component style?

vinnyhoward: Just tried it and no luck. The icon doesn't behave like I want it too

georgewrmarshall (approved):
I agree that we should allow access to all child elements within our components. We usually handle this by spreading a prop object to the child components instead of exposing each prop individually, as you're doing with `dropdownIconStyle`. It might be more flexible to use `dropdownIconProps` and spread them to the `<Icon {...dropdownIconProps} />`, allowing props to be overridden, but that would require somewhat complex types which is out of the scope of this PR. Approving to unblock.

vinnyhoward:
> I agree that we should allow access to all child elements within our components. We usually handle this by spreading a prop object to the child components instead of exposing each prop individually, as you're doing with `dropdownIconStyle`. It might be more flexible to use `dropdownIconProps` and spread them to the `<Icon {...dropdownIconProps} />`, allowing props to be overridden, but that would require somewhat complex types which is out of the scope of this PR. Approving to unblock.

That makes sense. I can do a follow up after the header redesign is finished to allow that. Thanks for reviewing and approving🫡 @georgewrmarshall

PR #11988, opened 2024-10-23
chore: upgrade assets-controllers v37

sahar-fehri:
## **Description**

Updates assets-controllers to v37

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/components/Base/RemoteImage/index.js
```diff
@@ -67,23 +67,36 @@ const RemoteImage = (props) => {
...
-    }
-  }, [props.source.uri, ipfsGateway]);
+  const [resolvedIpfsUrl, setResolvedIpfsUrl] = useState(false);
 
-  const uri = resolvedIpfsUrl || source.uri;
+  const uri =
+    resolvedIpfsUrl ||
+    (source.uri === undefined || source.uri?.startsWith('ipfs')
+      ? ''
+      : source.uri);
```
sahar-fehri, line 77: Now that `getFormattedIpfsUrl` is `async`, the `resolvedIpfsUrl` is being resolved slightly late resulting in
passing uri as `ipfs://example` which triggers "No suitable image URL loader for ipfs://..." error.

I am also setting the uri to `""` instead of `undefined` in the case where `source.uri`  is undefined

app/core/EngineService/EngineService.ts
```diff
@@ -53,7 +53,6 @@ class EngineService {
         name: 'AddressBookController',
         key: `${engine.context.AddressBookController.name}:stateChange`,
       },
-      { name: 'AssetsContractController' },
       { name: 'NftController', key: 'NftController:stateChange' },
```
sahar-fehri, line 56: Removing AssetsContractController from here because it is no longer required for state changes because it is now stateless

app/components/UI/Ramp/Views/BuildQuote/__snapshots__/BuildQuote.test.tsx.snap
```diff
@@ -11839,7 +11839,7 @@ exports[`BuildQuote View renders correctly 2`] = `
                                               onLoadEnd={[Function]}
                                               source={
                                                 {
-                                                  "uri": undefined,
+                                                  "uri": "",
                                                 }
```
sahar-fehri, line 11843: These snapshots are updated because of this change https://github.com/MetaMask/metamask-mobile/pull/11988/files#diff-a78d4aae7fcc002ba295de7ac42c733134d60ed085982abd0ccab79dff0fb7d8R75

When uri is `undefined`, the [Image.getSize](https://github.com/MetaMask/metamask-mobile/blob/60bdb968e75fe246ab23b8b754d4440dc037019b/app/components/Base/RemoteImage/index.js#L108) fct will throw

sahar-fehri:
@SocketSecurity ignore npm/@metamask/eth-json-rpc-provider@1.0.1
@SocketSecurity ignore npm/@metamask/eth-sig-util@6.0.2

app/components/Base/RemoteImage/index.js
tommasini: Are we displaying a fallback image? I remember we had a background for such cases

app/core/Engine.test.ts
```diff
@@ -71,7 +71,13 @@ describe('Engine', () => {
     const engine = Engine.init({});
     const initialBackgroundState = engine.datamodel.state;
 
-    expect(initialBackgroundState).toStrictEqual(backgroundState);
+    // AssetsContractController is stateless in v37 resulting in an undefined state
```
tommasini, line 74: If is stateless, should we remove it from the `bacgkroundState` object instead and also free the memory was using on the users devices by deleting it with a migration?

sahar-fehri: I ve tested this;  For the `initialBackgroundState` to match the `bacgkroundState`; i would need to remove AssetsContractController from this controllers array https://github.com/MetaMask/metamask-mobile/blob/60bdb968e75fe246ab23b8b754d4440dc037019b/app/core/Engine.ts#L1435
And also remove AssetsContractController from initialBackgroundState https://github.com/MetaMask/metamask-mobile/blob/60bdb968e75fe246ab23b8b754d4440dc037019b/app/util/test/initial-background-state.json#L24

The test would pass but the Engine.context would have `undefined` for AssetsContractController which would break the flows that call AssetsContractController example here https://github.com/MetaMask/metamask-mobile/blob/60bdb968e75fe246ab23b8b754d4440dc037019b/app/util/address/index.ts#L609

cc @cryptodev-2s

tommasini: But I still don't think it should be on the redux state right? It shouldn't be on backgroundState if doesn't have data to it right?


I see as well that version 36 didn't had data at AssetsContractController state. I see, okok, I think this is okay, but, it seems a key that it's persisted to the storage with no usage at all, which is weird, what do you think?

tommasini (approved):

cryptodev-2s (approved):

gambinish (approved):

Prithpal-Sooriya (approved):

app/components/Base/RemoteImage/index.js
sahar-fehri: Noted; I think this is unrelated to this upgrade PR, we can check this separately

app/core/Engine.test.ts
sahar-fehri: I agree its weird to have the state as undefined; but did not find yet a good to fix this without breaking the context; 
also did not think it is a blocker as long as we are able to call functions on assetsContractController

amitabh94 (approved):

PR #12163, opened 2024-11-04
feat(ci): Expo [Outdated, moved to https://github.com/MetaMask/metamask-mobile/pull/12244]

andrepimenta:
Moved work to here because of incompatibilities with libraries installed at the same time as expo:
https://github.com/MetaMask/metamask-mobile/pull/12244


ios/.xcode.env.local
Cal-L, line 1: Remove file

andrepimenta: Oops! Thanks!

PR #12171, opened 2024-11-04
feat: multichain polling hook

bergeron:
## **Description**

Adds a `usePolling` hook that handles starting and stopping polling loops for polling controllers.  This will be used to poll across chains, and eventually make more polling UI based so we're only polling data when UI components require it.


app/components/Views/Root/index.js
```diff
@@ -85,9 +86,11 @@ const ConnectedRoot = () => {
     <SafeAreaProvider>
       <ThemeContext.Provider value={theme}>
         <ToastContextWrapper>
-          <ErrorBoundary view="Root">
-            <App />
-          </ErrorBoundary>
+          <AssetPollingProvider>
```
tommasini, line 89: What do you think of instead adding the provider at the root level, adding it only for authenticated routes?

bergeron: That sounds like it makes sense.  Could you help me understand what that would look like?

bergeron: I experimented with putting it in the `MainFlow` of `app/components/Nav/Main/index.js`.  But then it calls the hook every time I navigate to a different page in the UI.  Whereas keeping it in this root view its only called once, which may be better for perf?

bergeron: Will add the provider in a future PR

app/components/hooks/usePolling.ts
```diff
@@ -0,0 +1,55 @@
...
+  [usePollingOptions.input && JSON.stringify(usePollingOptions.input)]);
+
+  // stop all polling on dismount
+  useEffect(() => () => {
+      for (const token of pollingTokens.current.values()) {
+        usePollingOptions.stopPollingByPollingToken(token);
+      }
+    },
+    // Intentionally empty to trigger on dismount
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tommasini, line 51: NIT: We should avoid surpressing this lint rule accordingly to react documentation
image

They also have documentation with strategies to avoid this.
 https://react.dev/learn/removing-effect-dependencies#removing-unnecessary-dependencies
 
 This is just a nit pick, because we are seeing this lint rule surpressing growing across the code base but, nothing was decided or not how to proceed around this.

tommasini (approved):

app/components/hooks/usePolling.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// A hook that manages multiple polling loops of a polling controller.
+// Callers provide an array of inputs, and the hook manages starting
+// and stopping polling loops for each input.
+const usePolling = <PollingInput>(
+  usePollingOptions: UsePollingOptions<PollingInput>,
+) => {
+
+  const pollingTokens = useRef<Map<string, string>>(new Map());
```
frankvonhoven, line 16: Why a `Map` vs just a `{}`?

PR #12178, opened 2024-11-05
chore: upgrade assets-controllers to v41

sahar-fehri:
## **Description**

Bump assets-controllers to v41

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### **After**

https://github.com/user-attachments/assets/83ae8f0b-173d-4bb2-9e95-f5cfcd82a9d0


app/core/Engine.ts
```diff
@@ -704,8 +704,18 @@ export class Engine {
...
+      ? Object.values(networkController.state.networkConfigurationsByChainId)
+      : [];
+    const nativeCurrencies = [
+      ...new Set(
+        Object.values(networkConfigurations).map((n) => n.nativeCurrency),
+      ),
+    ];
     currencyRateController.startPolling({
-      networkClientId: networkController.state.selectedNetworkClientId,
+      nativeCurrencies,
```
bergeron, line 718: It probably makes sense for this to continue being the called with the ticker of the current network.  We'll instead want to use the polling hooks to multichain this, so it responds to networks being added and removed. 

We could also consider combining this upgrade with https://github.com/MetaMask/metamask-mobile/pull/12171 to actually multichain it.  Which I think would make it no longer necessary to call `updateExchangeRate` in these other places

sahar-fehri: updated here [44e937f](https://github.com/MetaMask/metamask-mobile/pull/12178/commits/44e937fc069748e411bec15d1443636a7b9f1e48)

bergeron (dismissed):

app/components/UI/Tokens/index.tsx
```diff
@@ -75,14 +75,23 @@ const Tokens: React.FC<TokensI> = ({ tokens }) => {
...
+    selectNetworkConfigurations,
+  );
   const hideZeroBalanceTokens = useSelector(
     (state: RootState) => state.settings.hideZeroBalanceTokens,
   );
 
   const tokenExchangeRates = useSelector(selectContractExchangeRates);
   const currentCurrency = useSelector(selectCurrentCurrency);
   const conversionRate = useSelector(selectConversionRate);
+  const nativeCurrencies = [
```
bergeron, line 88: This should probably also only poll on the current network's ticker.  Otherwise we'll have multiple polling loops making redundant requests.  One loop for all tickers here and one loop for the current ticker from engine.ts.

We can make it actually multichain once we have https://github.com/MetaMask/metamask-mobile/pull/12171

bergeron: This existing code isn't using polling correctly anyway.  By calling start without a corresponding stop, it would always be polling on that chain even if you were to switch to a different one.  It'd be better to make a one off call to `updateExchangeRate`.

app/core/Engine.ts
```diff
@@ -704,8 +704,14 @@ export class Engine {
       }),
       state: initialState.CurrencyRateController,
     });
+    const currentNetworkConfig =
+      networkController.getNetworkConfigurationByNetworkClientId(
+        networkController?.state.selectedNetworkClientId,
+      );
     currencyRateController.startPolling({
```
bergeron, line 711: It's also not clear to me that any one is stopping and starting a new polling loop when the user switches to a network with a different native currency.  But we're just about to multichain this next, so it can stay for now.

bergeron (dismissed):

app/components/UI/Tokens/index.tsx
bergeron: https://github.com/MetaMask/metamask-mobile/pull/12178/commits/9b67d0ae6241b6cb3a9e49d48f8c60e6faa7892e

bergeron (approved):

tommasini (approved):
LGTM!

darkwing (approved):

PR #12237, opened 2024-11-07
chore: Remove GoogleService files from git cache

gambinish:
## **Description**

Even though they are gitignore, `ios/GoogleService-Info.plist` and `ios/GoogleServices/GoogleService-Info.plist` were cached before they got added. So, they would not be ignored correctly by the gitignore.

This PR is the result of:

```
git rm --cached ios/GoogleServices/GoogleService-Info.plist
git rm --cached ios/GoogleService-Info.plist
```

Tested locally and it seems to work as we need it to. Can you please also verify when reviewing?

## **Related issues**

Fixes:

## **Manual testing steps**

1. Remove existing GoogleServices files from your local
2. yarn && yarn setup
3. yarn start:ios

You will see the new files get generated: `GoogleService-Info.plist has been created successfully.`
They should no longer be tracked in git: `git status`


owencraston:
I am worried about how removing this will effect the external contributor experience. Here is a snippet from the readme...

```
Before running the app, keep in mind that MetaMask uses FCM (Firebase Cloud Message) to empower communications. Based on this, as an external contributor you would preferably need to provide your own FREE Firebase project config file with a matching client for package name io.metamask, and update your google-services.json file in the android/app or GoogleService-Info.plist file in the ios directory.

External Contributors In case you don't have FCM account, you can use ./android/app/google-services-example.json for Android or ./ios/GoogleServices/GoogleService-Info-example.plist for iOS and follow the steps below to populate the correct environment variables in the .env files (.ios.env, .js.env, .android.env), adding GOOGLE_SERVICES_B64_ANDROID or GOOGLE_SERVICES_B64_IOS variable depending on the environment you are running the app (ios/android).
```

Perhaps we should move these into an example file or add the suffix.example and modify the readme instructions to instruct users to copy over these files in the correct paths. Alternatively if we are confident that the `yarn:start:ios` will generate this file for us then we can ommit this change.

gambinish:
> Perhaps we should move these into an example file or add the suffix.example and modify the readme instructions to instruct users to copy over these files in the correct paths. Alternatively if we are confident that the `yarn:start:ios` will generate this file for us then we can ommit this change.

I've never had to copy these files, and have been relying purely on the yarn start:ios script to do everything. Have you been copying the files every time?

owencraston (approved):
Confirmed that removing the GoogleService-Info.plist from git cache stops them from being tracked by git. The readme changes look good but since we still have the example files in the code base (I did not realize this before I made my last comment) I think its okay to revert your changes to the readme.

PR #12283, opened 2024-11-13
fix: ensure unstake max will unstake all user shares

nickewansmith:
## **Description**

1. This PR addresses a bug where when unstaking entire balance staked causes 1 share to be left in the vault that cannot be unstaked from the UI.  
2. The solution is for mobile only and is a quick fix, note that the issue currently exists in portfolio as well and will need to be fixed properly. We solve the issue by calling the getShares contract method to get the total user shares instead of converting the total assets into shares with the convertToShares contract method. We add the interface for the getShares contract method on the fly as it is not part of the contract abi in the stake-sdk where we get the contract instance from.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/STAKE-867

## **Manual testing steps**

1. Stake some amount of ETH or have some ETH staked already
2. Unstake All of the ETH
3. When the transaction is successful, there should be no ETH listed for the Staked Ethereum asset value. It should be exactly 0 and not < .0001
4. If there is an account that has 1 share left, it should now be able to exit that 1 share.

### **Before**

https://github.com/user-attachments/assets/4d235027-642b-4e4b-af50-7ae798ad47af

### **After**

https://github.com/user-attachments/assets/b9e9df65-a1e9-4d58-8bf5-3239b5007f41


amitabh94 (approved):
LGTM!

PR #11579, opened 2024-10-02
feat: 1940 Add custom traces

MarioAslau:
## **Description**

Screenshot 2024-10-11 at 20 29 04
Screenshot 2024-10-16 at 00 01 02
Screenshot 2024-10-15 at 17 57 12

This task is for adding custom spans to track activities that happen between app start and wallet UI load. The screenshot below is an example of a trace for Wallet UI load that takes about a minute to load. During that time, we can see a large gap between app start spans and the initial http requests.

The goal here is to isolate these areas and track them with custom spans. Once implemented, we can expect to see the custom spans appearing within the gap, which would inform us of the areas to optimize

Issue: https://github.com/MetaMask/mobile-planning/issues/1940

Technical Details
* Added custom span for when the Login screen is mounted to when the login button is tapped
* Added span for when the login button is tapped to when the wallet view is mounted
* Added custom span for Engine initialization process
* Added custom span for Store creation
* Added custom span Storage rehydration
* Added custom span fro Create New Wallet to Choose Password
* Added custom span for Biometrics authentication

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Pre-merge author checklist**

- [ ] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #12374, opened 2024-11-21
perf: Remove costly reduce operation for generating Engine context

MajorLift:
## Overview

Optimizes an expensive operation from the Engine initialization process. 

A custom span for measuring this process was added in https://github.com/MetaMask/metamask-mobile/pull/11579.

## Motivation

Currently, the Engine's context object is created using a `Array.prototype.reduce()` operation on a large `controllers` array. This reduce call is also written to create a temporary object on every iteration.

Benchmarks show that this particular implementation of reduce runs significantly slower than other means of creating an object, and compared to reduce calls that are written to re-use the output object the difference is in four orders of magnitude (1555.6 Ops/sec vs. 0.2 Ops/sec).
> https://www.measurethat.net/Benchmarks/Show/21554/0/objectfromentries-vs-reduce-vs-assign

A simple fix would be to refactor the reduce operation so it doesn't rely on temporary objects and spreading:

```diff
diff --git a/app/core/Engine.ts b/app/core/Engine.ts
index decf28da51..1ba63ed106 100644
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -1738,10 +1738,10 @@ export class Engine {
       this.controllerMessenger,
     );
     this.context = controllers.reduce<Partial<typeof this.context>>(
-      (context, controller) => ({
-        ...context,
-        [controller.name]: controller,
-      }),
+      (context, controller) => {
+        context[controller.name] = controller;
+        return context;
+      },
       {},
     ) as typeof this.context;
```

However, we can also avoid the cost of running reduce altogether simply by creating a controller name-keyed object to assign to the Engine's context field. This is also the approach taken in the Extension `MetamaskController`.

## **Description**

- Replaces `controllers` array in the Engine class with a corresponding object that is directly assigned to `context`.
- Optimize `Object.values` for deriving ComposableController input with `Object.keys` and `reduce` (re-uses object).
  > https://www.measurethat.net/Benchmarks/Show/7173/0/objectvalues-vs-reduce
- Removes for-loop that sets initial state for V1 controllers now that all mobile controllers have been updated to V2.

## **Related issues**

- Contributes to https://github.com/MetaMask/mobile-planning/issues/1095
- Related https://www.notion.so/metamask-consensys/Refactor-Engine-Init-to-only-include-minimum-controllers-for-logging-in-on-startup-23ee10636dcc44d480a93a63361abb5b

### DIFF
--- a/app/core/Engine.ts
+++ b/app/core/Engine.ts
@@ -1486,10 +1486,10 @@ export class Engine {
       getMetaMetricsProps: () => Promise.resolve({}), // Return MetaMetrics props once we enable HW wallets for smart transactions.
     });
 
-    const controllers: Controllers[keyof Controllers][] = [
-      this.keyringController,
-      accountTrackerController,
-      new AddressBookController({
+    this.context = {
+      KeyringController: this.keyringController,
+      AccountTrackerController: accountTrackerController,
+      AddressBookController: new AddressBookController({
         messenger: this.controllerMessenger.getRestricted({
           name: 'AddressBookController',
           allowedActions: [],
@@ -1497,11 +1497,11 @@ export class Engine {
         }),
         state: initialState.AddressBookController,
       }),
-      assetsContractController,
-      nftController,
-      tokensController,
-      tokenListController,
-      new TokenDetectionController({
+      AssetsContractController: assetsContractController,
+      NftController: nftController,
+      TokensController: tokensController,
+      TokenListController: tokenListController,
+      TokenDetectionController: new TokenDetectionController({
         messenger: this.controllerMessenger.getRestricted({
           name: 'TokenDetectionController',
           allowedActions: [
@@ -1547,8 +1547,7 @@ export class Engine {
         useAccountsAPI: true,
         disabled: false
       }),
-
-      new NftDetectionController({
+      NftDetectionController: new NftDetectionController({
         messenger: this.controllerMessenger.getRestricted({
           name: 'NftDetectionController',
           allowedEvents: [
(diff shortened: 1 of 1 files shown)


tommasini (approved):
LOVE IT! LGTM! So cool that we now only have base controllers v2 controllers!

tommasini (approved):
LOVE IT! LGTM! So cool that we now only have base controllers v2 controllers!

app/core/Engine.ts
```diff
@@ -1710,40 +1709,27 @@ export class Engine {
...
-    for (const controller of controllers) {
-      if (
-        hasProperty(initialState, controller.name) &&
-        // Use `in` operator here because the `subscribe` function is one level up the prototype chain
-        'subscribe' in controller &&
-        controller.subscribe !== undefined
-      ) {
-        // The following type error can be addressed by passing initial state into controller constructors instead
-        // @ts-expect-error No type-level guarantee that the correct state is being applied to the correct controller here.
-        controller.update(initialState[controller.name]);
```
Cal-L, line 1731: I believe this was the old way (BaseController V1) controllers consumed initial state. Since this is removed, could you confirm that we wouldn't need this anymore (verifying all controllers are using BaseControllerV2)?

Cal-L (commented):
Left a comment

app/core/Engine.ts
Cal-L: I think @tommasini verified in a comment above. Will resolve unless found otherwise.

Cal-L (approved):
LGTM

app/core/Engine.ts
MajorLift: Yes, I can confirm this. The work for updating this has been tracked in [this epic](https://github.com/MetaMask/MetaMask-planning/issues/3234). The controllers are V2 in all of the specified target versions.

As a sanity check, I also just went and checked the type definitions for all of the properties in the engine `Controllers` interface. They all either inherit from `BaseControllerV2`/`StaticIntervalPolllingController`, or in the case of non-controllers don't inherit from `BaseControllerV1`.

PR #12375, opened 2024-11-21
fix: breaking selector due to missing controller state

### LINKED ISSUE #11909: [Sentry] TypeError: undefined is not an object (evaluating 't.isProfileSyncingEnabled')
Sentry Issue: [METAMASK-MOBILE-2XYA](https://metamask.sentry.io/issues/5991748071/?referrer=github_integration)

```
TypeError: undefined is not an object (evaluating 't.isProfileSyncingEnabled')
  at e.selectIsProfileSyncingEnabled (app/selectors/notifications/index.tsx:31:5)
  at _e.default (app/components/Views/OnboardingSuccess/DefaultSettings/index.tsx:32:35)
...
(19 additional frame(s) were not displayed)
```

Prithpal-Sooriya:
## **Description**

The selector is trying to access some engine state that is removed at build time through code-fences.
This fix ensures that the selectors are using the engine state or default to the controllers default state.

Long term - we need to investigate and ensure that the UI is not calling the selectors.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/11909

## **Manual testing steps**

1. Can you start up the application without it breaking?

## **Screenshots/Recordings**

Before / After Recording
https://www.loom.com/share/badc4b4cd3b0446486497973dcf337f6?sid=33bcfe23-c495-41bf-89f5-e57b4b72e3d2

I was not able to replicate the controller state being undefined/missing, so forced it to be undefined when testing.


app/selectors/notifications/index.tsx
```diff
@@ -17,18 +17,21 @@ type AuthenticationState =
...
 const selectNotificationServicesControllerState = (state: RootState) =>
-  state?.engine?.backgroundState?.NotificationServicesController;
+  state?.engine?.backgroundState?.NotificationServicesController ??
+  NotificationServicesController.defaultState;
 
 export const selectIsProfileSyncingEnabled = createSelector(
   selectUserStorageControllerState,
   (userStorageControllerState: UserStorageState) =>
-    userStorageControllerState.isProfileSyncingEnabled,
+    userStorageControllerState?.isProfileSyncingEnabled,
```
Prithpal-Sooriya, line 34: This should not be needed now since we are defaulting the state. However there is no hard in keeping this.

matteoscurati (dismissed):
LGTM

sethkfman:
@Prithpal-Sooriya can you show manual confirmation this has fixed the issue?

mathieuartu (dismissed):

app/selectors/notifications/index.tsx
```diff
@@ -17,18 +17,21 @@ type AuthenticationState =
...
 
 const selectAuthenticationControllerState = (state: RootState) =>
-  state?.engine?.backgroundState?.AuthenticationController;
+  state?.engine?.backgroundState?.AuthenticationController ??
+  AuthenticationController.defaultState;
 
 const selectUserStorageControllerState = (state: RootState) =>
-  state?.engine?.backgroundState?.UserStorageController;
+  state?.engine?.backgroundState?.UserStorageController ??
+  UserStorageController.defaultState;
```
Prithpal-Sooriya, line 25: This is the main fix for the missing controller state.

It is really odd that the controller state is missing sometimes. I have a [hypothesis](https://excalidraw.com/s/4tjYOJYejWM/6GlqQBXt7wN#json=-UweJuPIvJBmn2jADcsAb,_L03t-4Mfwlw-D9ENUcajQ), but it is hard to prove.

Maybe we should have it as mobile best practice to "distrust controller state", or maybe at the type level make the controllers partial/optional (but this would be a large breaking change).

Prithpal-Sooriya: I think I prefer this selector default state approach instead of migrations, as the redux state _should_ do a second re-hydration once the engine has been initialised. We can then keep migrations only for state patches/fixes.

app/util/notifications/hooks/useCreateSession.ts
```diff
@@ -31,6 +32,10 @@ function useCreateSession(): UseCreateSessionReturn {
   const [error, setError] = useState<string | undefined>(undefined);
 
   const createSession = useCallback(async () => {
+    if (!isNotificationsFeatureEnabled()) {
+      return;
+    }
```
Prithpal-Sooriya, line 37: All these other changes to hooks is to ensure we aren't unintentionally calling these hooks when the feature is off.

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -87,16 +92,22 @@ export function useAccountSettingsProps(accounts: Account[]) {
...
+      const result =
+        await Engine.context.NotificationServicesController.checkAccountsPresence(
+          memoAccounts,
+        );
+      dispatch(updateAccountState(result));
+      return result;
+    } catch (err) {
+      Logger.log('Failed to get account settings:', err);
+    }
+  }, [dispatch, memoAccounts]);
```
Prithpal-Sooriya, line 110: Ignore this change, it was a "save on format"

app/components/Views/Settings/SecuritySettings/__snapshots__/SecuritySettings.test.tsx.snap
```diff
@@ -3583,188 +3583,8 @@ exports[`SecuritySettings should render correctly 1`] = `
       swipeDirection="down"
       swipeThreshold={100}
       transparent={true}
-      visible={true}
-    >
-      <View
```
Prithpal-Sooriya, line 3588: Snapshot change as we are now not calling any notification hooks.

TODO: I want us to clean up this effect:
https://github.com/MetaMask/metamask-mobile/blob/35c51df53fe824a11736f3da97be8b5daf94aa84/app/components/Views/Settings/SecuritySettings/SecuritySettings.tsx#L209-L216

I don't want sideEffects to cascade the updates for basic functionality (especially since this can be re-invoked from dependency changes). Instead I want this to be done during the basic functionality switch-off.

Prithpal-Sooriya:
@sethkfman I've attached a recording for a before and after the changes.

I've also sneaked in some other changes to ensure notification hooks are not called when the feature is disabled.
If this is too many changes, I'm happy to make them separate PRs.

mathieuartu (approved):

app/selectors/notifications/index.tsx
tommasini: The easiest way to proof this is creating a migration for testing purposes and delete the UserStorageController data if exists, wdyt?

Prithpal-Sooriya: haha, I actually tried this, but these selectors only fire  "after" the engine is initialised. So kinda rules out my hypothesis.

The hypothesis still remains true if a component is mounted/rendered/uses selector before the engine init - which is unlikely since the wallet is locked.

app/components/Views/Settings/SecuritySettings/__snapshots__/SecuritySettings.test.tsx.snap
Prithpal-Sooriya: I've noted this, and will re-pick this up when we go back to mobile.

app/selectors/notifications/index.tsx
tommasini: Okok, yeah I see this data should never exist because it's fenced right, this controller never get's initialised in the first place, or it shouldn't

It seems with this change being the main fix, it means that we really do not know what is happening, since this shouldn't have impact to the end user I'm okay on shipping it, to let things going, but maybe we should flag this code as TO DO, since we may not want to default to a state since it seems we listens to this events we may not want to get into a state where we are showing something that it's not accurate to the user. (Didn't deep dive into the `userStorageController`)
```
 allowedEvents: [
          'KeyringController:unlock',
          'KeyringController:lock',
          'AccountsController:accountAdded',
          'AccountsController:accountRenamed',
        ],
```

tommasini: Something is not adding up
At import secret recovery phrase we are calling [enableProfileSyncing](https://github.com/MetaMask/metamask-mobile/blob/6a9ab2ddcf5b470f890c8abb28160b7b58e10632/app/components/Views/ImportFromSecretRecoveryPhrase/index.js#L105) when we press import

[enableProfilySyncing](https://github.com/MetaMask/metamask-mobile/blob/00a91cee2b65c2b42d5b05603ae7e865a15a74c0/app/actions/notification/helpers/index.ts#L41) is trying to access userStorageController, although that controller is code fenced in [engine](https://github.com/MetaMask/metamask-mobile/blob/feb6bf8b4a158a45a290c274c99e86f8d17678e3/app/core/Engine.ts#L1233), it wouldn't be that causing the issue?

tommasini: Ah I see, that's why the flag in the `useProfileSyncing` hook, okok will approve it!

tommasini (approved):

app/selectors/notifications/index.tsx
Prithpal-Sooriya: yep yep. saw this and added the checks for all our hooks to not do anything when flag is off.

I will add TODO comments in the next PR.

vinnyhoward (approved):

PR #12396, opened 2024-11-22
feat(ramp): enable buy button in asset overview

wachunei:
## **Description**

This PR adds the buy button to the Token (AssetOverview) page, using the [Ramp Intent](https://github.com/MetaMask/metamask-mobile/pull/9390) parameters to optimistically open the Buy flow with the token selected. The Buy button will show if the network is supported.

**Happy Path / Optimistic case description:**

Given:
- The user has already started the Buy flow before and has a region selected from a previous interaction
- The asset is supported in the current region and payment method the user has selected from the previous buy flow interaction

Then:
- As a user, I want to press the Buy button from the Token page and have the token selected on the Buy flow

**Other cases**:
- If the user has not started the buy button, not selected a region
	- Then the user will land on the buy flow start page or region selector
- If the asset is not supported in the current region/payment method configuration
	- Then the user will land on the buy flow and the default token will be selected.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/pull/12396/

## **Manual testing steps**

1. Go to Token page
2. Click on the Buy button

### **Before**

No Buy button

https://github.com/user-attachments/assets/7f32aa08-7fda-4cdb-b49b-a2ea5f2a7223

### **After**

Happy Path

https://github.com/user-attachments/assets/0f52cc48-af01-45cb-bc07-503fbcf82d62

Token unavailable (default token is selected)

https://github.com/user-attachments/assets/578fbbbd-e39a-4c79-981e-6139ef7cfef4


georgeweiler (dismissed):

Cal-L (commented):
Code LGTM. However, it seems like we're missing some UX here since we're not notifying the user that the token is not available for buying in their region. Defaulting to the default token can also easily be interpreted as a bug in a sense that the wrong token may have been selected. Is the fallback condition a result of not having a region selected in the first place?

wachunei:
>However, it seems like we're missing some UX here since we're not notifying the user that the token is not available for buying in their region. Defaulting to the default token can also easily be interpreted as a bug in a sense that the wrong token may have been selected.

Yes I agree, we can improve this by adding a message to let the user the token is not available. 

>Is the fallback condition a result of not having a region selected in the first place?

No, the fallback happens when the token is not available. In the case has not selected region and payment method and they do it in this flow, the token will be selected if it was available.

AxelGes (approved):

salimtb (approved):

PR #12417, opened 2024-11-25
feat: multichain detect tokens feat

salimtb:
## **Description**
PR to add the multichain autodetection to mobile under a feature flag

## **Related issues**

Fixes:

## **Manual testing steps**

1. run `PORTFOLIO_VIEW=true yarn watch`
2. run `yarn start:ios`
3. check if the tokens autodetection is multichained

### **After**

https://github.com/user-attachments/assets/c4428c82-5fb7-4701-8625-cf7bb36ee4be


app/components/Views/AssetOptions/AssetOptions.test.tsx
```diff
@@ -0,0 +1,188 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react-native';
+import { useNavigation } from '@react-navigation/native';
+import { useSelector } from 'react-redux';
+import AssetOptions from './AssetOptions';
+// import { strings } from '../../../../locales/i18n';
```
tommasini, line 6: Can we remove this commented code?

app/components/Views/DetectedTokens/index.tsx
```diff
@@ -131,10 +158,72 @@ const DetectedTokens = () => {
 
       sheetRef.current?.onCloseBottomSheet(async () => {
         try {
-          tokensToIgnore.length > 0 &&
-            (await TokensController.ignoreTokens(tokensToIgnore));
+          if (tokensToIgnore.length > 0) {
+            const tokensToIgnoreByChainId = tokensToIgnore.reduce<
```
tommasini, line 162: We may want to retry a more performance approach here, but need some more time to think about this just flagging with this comment

app/components/Views/DetectedTokens/index.tsx
```diff
@@ -131,10 +158,72 @@ const DetectedTokens = () => {
...
+                networkInstanceId,
+              );
+            });
+
+            await Promise.all(ignorePromises);
+          }
           if (tokensToImport.length > 0) {
-            await TokensController.addTokens(tokensToImport, networkClientId);
+            if (isPortfolioViewEnabled) {
+              const tokensByChainId = tokensToImport.reduce<
```
tommasini, line 196: Same here

app/selectors/tokensController.ts
```diff
@@ -58,3 +60,55 @@ export const selectAllTokensFlat = createSelector(
...
+      }
+      return acc;
+    }, {});
+  },
+);
+
+// TODO: This isn't working fully, once a network has been selected then it
+// can detect all tokens in that network. But by default it only shows
+// detected tokens if the user has chosen it in the past
+export const selectAllDetectedTokensFlat = createSelector(
```
tommasini, line 93: It also seams that this can create a big burden performance qise if the user have a lot of tokens, we should try to get it optimized

salimtb: Thanks for pointing this out! I've refactored the code for improved performance and readability:
1. Replaced reduce with for..of loops: The original code used reduce with concat to flatten the tokens, which creates a new array on every iteration. This can be inefficient for large datasets, In the updated code, tokens are directly pushed into a single array (flattenedTokens). This avoids unnecessary array allocations and reduces overhead.

app/components/Views/DetectedTokens/index.tsx
salimtb: Thanks for flagging this! here are my suggestions on how we can improve this:

1. **Removing reduce:** Replaced it with a simple `for..of` loop for grouping tokens by chainId, which simplifies the logic and reduces potential inefficiencies.

2. **Using p-limit:** Introduced `p-limit` to control concurrency when processing network operations. This ensures we don't overwhelm the system and create bottleneck on the event loop. The concurrency limit is currently set to 5, but we can adjust it based on system capacity , this also should be done everywhere

i just implemented the first one , for the second one , we need to discuss it more globally i think

tommasini (approved):
LGTM! Still we need to have an eye with this new algos with big complexity and understand the impact of those! Maybe having a custom instrumentation with Sentry here it's not a bad idea cc: @MarioAslau

sahar-fehri:
I did QA on this branch; i'm able to see correct detected tokens when "all networks"/ "current network" is selected.
I'm able to hide and add the token again.
Without the feature flag im seeing same behavior we have on main.

sahar-fehri (approved):

PR #12427, opened 2024-11-26
feat: implement remote feature flag controller

joaoloureirop:
## **Description**

Introduction of `@metamask/remote-feature-flag-controller` library.

Remote feature flag controller manages data flow, retry policy, and cache expiry.
The controller consumer manages default values, data persistency, and data distribution.

See [ADR](https://github.com/MetaMask/decisions/blob/b3094d47a568ac1e076a44fa704c2d29d1b59f35/decisions/wallet-platform/0001-remote-rollout-feature-flags.md) for a in-depth description

### Controller init on `Engine.ts` with feature flags fetching only on cold app starts.

`@metamask/remote-feature-flag-controller` is only asked to fetch feature flags after its init in `Engine.ts`. Ensures feature flags are only fetched on cold app starts.

### Fallback values

Default values are used when remote feature flags are undefined.
The fallback mechanism is implemented by each feature flag selector `app/selectors/featureFlagsController/<featureFlagName>`

In this PR we include `minimumAppVersion` selector, which manages the LD feature flag `mobile-minimum-versions`

### One selector per each feature flag

[LD feature flags can be boolean, number, strings on JSON objects](https://docs.launchdarkly.com/sdk/concepts/flag-types#understanding-flag-types). 
We've decided to have each feature flag with its own selector

A feature flag selector contains:
- state selectors for each feature flag value.
- business logic
- defaults for when feature flags values are undefined.
- TS types.
- unit tests and mocked data.

This architecture offers a clear separation between each feature flag and the logic behind it, allowing easier manipulation.
Code owners are assigned to each feature flag.

### 

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/2054
Fixes: https://github.com/MetaMask/mobile-planning/issues/1975

## **Manual testing steps**

1. Go to this page...
2.
3.


app/core/Engine.ts
```diff
@@ -744,6 +759,48 @@ export class Engine {
...
+        default: return DistributionType.Main;
+      }
+
+    };
+
+    const featureFlagController = new RemoteFeatureFlagController({
+      messenger: this.controllerMessenger.getRestricted({
+        name: 'RemoteFeatureFlagController',
+        allowedActions: ['PreferencesController:getState'],
+        allowedEvents: ['PreferencesController:stateChange'],
```
joaoloureirop, line 785: might not be needed

joaoloureirop: addressed https://github.com/MetaMask/metamask-mobile/pull/12427/commits/27b32a3bf00c8c429c1532cf063a441eb82e9074

metro.config.js
```diff
@@ -8,17 +8,35 @@
...
+const featureFlagModuleDir = path.resolve(__dirname, "../../core/feature-flags/packages/remote-feature-flag-controller");
+
+const extraNodeModules = {
+  "@metamask/remote-feature-flag-controller": featureFlagModuleDir,
+};
+
+const watchFolders = [
+  featureFlagModuleDir,
+];
+
```
joaoloureirop, line 22: To be removed once controller has been published to NPM

Will revert 00a3bd164f163a714d4e15e921dbb300b1fc70fb

yarn.lock
```diff
@@ -5345,6 +5345,10 @@
     escape-string-regexp "^4.0.0"
     invariant "2.2.4"
 
+"@metamask/remote-feature-flag-controller@link:../../core/feature-flags/packages/remote-feature-flag-controller":
+  version "0.0.0"
+  uid ""
+
 "@metamask/rpc-errors@7.0.1", "@metamask/rpc-errors@^6.0.0", "@metamask/rpc-errors@^6.2.1", "@metamask/rpc-errors@^6.3.1", "@metamask/rpc-errors@^7.0.0", "@metamask/rpc-errors@^7.0.1":
```
joaoloureirop, line 5352: update once controller has been published to NPM

app/selectors/featureFlagController/index.ts
Cal-L, line 1: Let's bucket these new selectors into `selectors/platform` directory, which the platform team will own. It probably isn't necessary to create another directory for `minimumAppVersion` as well since it's just part of the featureFlagController. The structure that could scale nicely may look like this:

`selectors/platform/featureFlagController`
This folder would then hold files 
`index.ts` - selectors
`index.test.ts` - tests for selectors
`constants.ts` - constants (your mocked values are also constants, which can go here)
`types.ts` - types

This pattern can scale as we add more controllers and could apply to other teams well

joaoloureirop: What are you trying to achieve with directories with team names?

If the concern is feature flags ownership, then the `CODEOWNERS` file should be used instead.

joaoloureirop: The directory naming based on team names is concerning due to the high chance of tech debt increase.
Internal team names can change without product or code changes. Directory names and their imports will easily become meaningless and confusing. It will become one more refactor necessity to minimize code rot.

[A feature flag can be a boolean, string, number, or a JSON object.](https://docs.launchdarkly.com/sdk/concepts/flag-types#understanding-flag-types)  Having multiple feature flags in a single place might be tempting to avoid multiple file creation but, will quickly turn into a big ball of mud with their unique fallbacks, selectors, mocks, tests all mixed up.

Would love some more insights and feedback on having a structure like that, because I see drawbacks with no benefit.

app/core/Engine/Engine.ts
```diff
@@ -933,8 +945,7 @@ export class Engine {
       encryptor,
       getMnemonic: getPrimaryKeyringMnemonic.bind(this),
       getFeatureFlags: () => ({
-        disableSnaps:
-          store.getState().settings.basicFunctionalityEnabled === false,
+        disableSnaps: !isBasicFunctionalityEnabled,
```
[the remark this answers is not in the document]
joaoloureirop: Great catch, thank you!

addressed 559c26fe7e792c843daeb018351a25009a81e725

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  });
+  it('should return fallback values', () => {
+    const {
+      appMinimumBuild,
+      appleMinimumOS,
+      androidMinimumAPIVersion,
+    } = selectMobileMinimumVersions(mockedEmptyFlagsState as RootState);
+    expect(appMinimumBuild).toBe(1024);
+    expect(appleMinimumOS).toBe(1025);
+    expect(androidMinimumAPIVersion).toBe(1026);
```
joaoloureirop, line 35: TODO:
- Test feature flag state selectors separate
- replace hardcoded numbers with mocked values.

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
```diff
@@ -0,0 +1,54 @@
...
+} from '@metamask/remote-feature-flag-controller';
+import Logger from '../../../../util/Logger';
+import { RemoteFeatureFlagInitParamTypes } from './types';
+import {
+  getFeatureFlagAppEnvironment,
+  getFeatureFlagAppDistribution
+} from './utils';
+
+
+const init = ({
```
joaoloureirop, line 14: TODO:
- add remote feature flag controller unit tests

app/reducers/index.ts
```diff
@@ -58,7 +55,6 @@ export interface RootState {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   collectibles: any;
   engine: { backgroundState: EngineState };
-  featureFlags: FeatureFlagsState;
```
joaoloureirop, line 61: TODO:
- create migration to remove featureFlags state

app/components/hooks/MinimumVersions/useMinimumVersions.tsx
```diff
@@ -4,22 +4,20 @@ import { createUpdateNeededNavDetails } from '../../UI/UpdateNeeded/UpdateNeeded
...
 
 const useMinimumVersions = () => {
   const { automaticSecurityChecksEnabled }: SecurityState = useSelector(
     (state: RootState) => state.security,
   );
-  const { featureFlags }: FeatureFlagsState = useSelector(
-    (state: RootState) => state.featureFlags,
-  );
+
+  const appMinimumBuild = useSelector((state: RootState) => selectAppMinimumBuild(state));
```
NicolasMassart, line 16: Q: not sure if this change makes the mock in the test obsolete and how this test doesn't need any changes.

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
NicolasMassart: is this a TODO?

app/selectors/featureFlagController/index.ts
Cal-L: Completely agree with the concern here and your point makes sense. The org has been moving towards codeowner based structures for a while now so there will be other opinions on this as well. we'll bring up this topic in the MCOP/engineering sync to get more thoughts. We can the codeowner based folder structure out for now.

NicolasMassart (commented):
First pass of review

app/core/Engine/Engine.ts
```diff
@@ -266,6 +268,10 @@ export class Engine {
   ) {
     this.controllerMessenger = new ExtendedControllerMessenger();
 
+    // Basic Functionality toggle defaults to true
+    const getBasicFunctionalityToggleState = () =>
```
NicolasMassart, line 272: S: should we use a selector here?

.github/CODEOWNERS
```diff
@@ -7,19 +7,24 @@
...
+patches/                                                    @MetaMask/mobile-platform
+app/core/Engine/Engine.ts                                   @MetaMask/mobile-platform
+app/core/Engine/Engine.test.ts                              @MetaMask/mobile-platform
+app/core/Engine/index.ts                                    @MetaMask/mobile-platform
+app/core/Engine/types.ts                                    @MetaMask/mobile-platform
+app/core/Engine/controllers/RemoteFeatureFlagController/    @MetaMask/mobile-platform
+app/core/Analytics/                                         @MetaMask/mobile-platform
+app/util/metrics/                                           @MetaMask/mobile-platform
+app/components/hooks/useMetrics/                            @MetaMask/mobile-platform
+app/selectors/featureFlagController/index.ts                @MetaMask/mobile-platform
```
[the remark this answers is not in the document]
joaoloureirop: We want  `app/selectors/featureFlagController/<featureFlagName>` to be owned by other teams

I'll update the `CODEOWNERS` syntax to match files like `featureFlagController/index.ts` but no further nested files

app/reducers/index.ts
joaoloureirop: addressed on d7d384c00dc19b00e4e16b619216dca186625c20

app/core/Engine/Engine.ts
```diff
@@ -478,6 +484,13 @@ export class Engine {
         'https://gas.api.cx.metamask.io/networks/<chain_id>/suggestedGasFees',
     });
 
+    const remoteFeatureFlagController = RemoteFeatureFlagController.init({
+      initialState,
+      controllerMessenger: this.controllerMessenger,
+      fetchFunction,
+      disabled: getBasicFunctionalityToggleState() === false
```
NicolasMassart, line 491: Q: why testing the exact boolean false equality here?
If the value is true, it's true.
If the value is false, it's true.
If the value eis anything like undefined or null, it will be true because of the getBasicFunctionalityToggleState default.
So if we have a false here, it's necessarily a real false as a value in the state. So the following would be enough:
```suggestion
      disabled: !getBasicFunctionalityToggleState()
```
and same on line 949

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
NicolasMassart: Thanks. Yes unit tests please!

.github/CODEOWNERS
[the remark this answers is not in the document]
joaoloureirop: addressed on b784e89e226916a201d204ee80cdb52bc0fe7292

app/selectors/featureFlagController/index.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  init: () => mockedEngine.init(),
+}));
+
+describe('featureFlagController', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+
+  it('should return feature flag initial state', () => {
```
NicolasMassart, line 16:
```suggestion
  it('returns feature flag initial state', () => {
```

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -0,0 +1,39 @@
...
+jest.mock('../../../core/Engine', () => ({
+  init: () => mockedEngine.init(),
+}));
+
+describe('Feature flag: minimumAppVersion', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should return feature flag object', () => {
```
NicolasMassart, line 19:
```suggestion
  it('returns feature flag object', () => {
```

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const {
+      appMinimumBuild,
+      appleMinimumOS,
+      androidMinimumAPIVersion,
+    } = selectMobileMinimumVersions(mockedState as RootState);
+    expect(appleMinimumOS).toBeDefined();
+    expect(appMinimumBuild).toBeDefined();
+    expect(androidMinimumAPIVersion).toBeDefined();
+  });
+  it('should return fallback values', () => {
```
NicolasMassart, line 29:
```suggestion
  it('returns fallback values', () => {
```

app/selectors/featureFlagController/minimumAppVersion/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  FEATURE_FLAG_NAME,
+  FeatureFlagType,
+  UndefinedFeatureFlagType,
+} from './types';
+
+export const featureFlagFallback: FeatureFlagType = {
+  [FEATURE_FLAG_NAME]: {
+    appMinimumBuild: 1243,
+    appleMinimumOS: 6,
+    androidMinimumAPIVersion: 21,
```
NicolasMassart, line 13: S: should we have this in a constants file?

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
NicolasMassart, line 29: Q: we only have this test that tests the fallback values. How do we test the normal values? Could we have a mocked test for this?

app/store/migrations/061.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  const invalidStates = [
+    {
+      state: null,
+      errorMessage: "FATAL ERROR: Migration 61: Invalid state error: 'object'",
+      scenario: 'state is invalid',
+    },
+  ];
+
+  for (const { errorMessage, scenario, state } of invalidStates) {
+    it(`should capture exception if ${scenario}`, async () => {
```
NicolasMassart, line 29:
```suggestion
    it(`captures exception if ${scenario}`, async () => {
```

app/store/migrations/061.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+      expect(newState).toStrictEqual(state);
+      expect(mockedCaptureException).toHaveBeenCalledWith(expect.any(Error));
+      expect(mockedCaptureException.mock.calls[0][0].message).toBe(
+        errorMessage,
+      );
+    });
+  }
+
+  it('remove featureFlags property from redux state', async () => {
```
NicolasMassart, line 40:
```suggestion
  it('removes featureFlags property from redux state', async () => {
```

NicolasMassart (commented):
Thanks for this awesome work @joaoloureirop!
Added some guideline suggestions and a few questions on some code parts I'm not sure to fully understand.

app/components/hooks/MinimumVersions/useMinimumVersions.tsx
NicolasMassart: I'm surprised the test is passing given the changes in the hook... so maybe rewrite the test in a more robust way. A change like this should make it fail otherwise it probably means it's not testing anything but the mocks...

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
Cal-L: In this case, it's debatable whether or not we should just put all of the functions in index or utils file. IMO, creating the utils file and re-exporting via index allows us to be more flexible in the future, where index can export other things such as constants, etc.

app/store/index.ts
```diff
@@ -87,11 +87,6 @@ const createStoreAndPersistor = async () => {
       basicFunctionalityEnabled:
         store.getState().settings.basicFunctionalityEnabled,
     });
-    // Fetch feature flags only if basic functionality is enabled
```
Cal-L, line 90: Let's confirm that the code above this is also no longer needed. Rehydration should automatically populate it already

app/store/migrations/061.ts
```diff
@@ -0,0 +1,12 @@
...
+import { ensureValidState } from "./util";
+
+export default function migrate(state: unknown) {
+  if (!ensureValidState(state, 61)) {
+    return state;
+  }
+  if (hasProperty(state, 'featureFlags')) {
+    delete state.featureFlags;
+  }
+  return state;
```
Cal-L, line 11: Let's update this migration to include the default state for `RemoteFeatureFlagController` to be safe

Cal-L (commented):
Left some comments

app/store/index.ts
joaoloureirop: Let's address that in a separate PR as it is out of scope

app/store/migrations/061.ts
joaoloureirop: default state is `undefined`, as with a clean redux store

app/selectors/featureFlagController/index.ts
```diff
@@ -0,0 +1,18 @@
...
+export const selectRemoteFeatureFlagControllerState = (state: RootState) =>
+  state.engine.backgroundState.RemoteFeatureFlagController;
+
+export const selectRemoteFeatureFlags = createSelector(
+  selectRemoteFeatureFlagControllerState,
+  (remoteFeatureFlagControllerState) =>
+    remoteFeatureFlagControllerState.remoteFeatureFlags
+);
+
+export const selectRemoteFeatureFlagsCacheTimestamp = createSelector(
```
DDDDDanica, line 13: Just a question: under which circumstance will we use this selector?

joaoloureirop: Probably none, I might remove it.

joaoloureirop: removed on 7bdcc819ac62606992b2bdecde4d2eb6209f7ab2

app/core/Engine/controllers/RemoteFeatureFlagController/types.ts
```diff
@@ -0,0 +1,9 @@
+import { ControllerMessenger, EngineState } from '../../types';
+
+export interface RemoteFeatureFlagInitParamTypes {
+  initialState: Partial<EngineState>;
+  controllerMessenger: ControllerMessenger,
```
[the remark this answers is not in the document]
joaoloureirop: addressed on [f5b754f](https://github.com/MetaMask/metamask-mobile/pull/12427/commits/f5b754ffb535100874030e58c3def43858e36dca)

app/core/Engine/controllers/RemoteFeatureFlagController/types.ts
```diff
@@ -0,0 +1,9 @@
+import { ControllerMessenger, EngineState } from '../../types';
+
+export interface RemoteFeatureFlagInitParamTypes {
+  initialState: Partial<EngineState>;
```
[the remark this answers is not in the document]
joaoloureirop: addressed on [f5b754f](https://github.com/MetaMask/metamask-mobile/pull/12427/commits/f5b754ffb535100874030e58c3def43858e36dca)

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  initialState,
+  controllerMessenger,
+  fetchFunction,
+  disabled,
+}: RemoteFeatureFlagInitParamTypes) => {
+
+  const remoteFeatureFlagController = new RemoteFeatureFlagController({
+    messenger: controllerMessenger.getRestricted({
```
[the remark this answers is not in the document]
joaoloureirop: addressed on [f5b754f](https://github.com/MetaMask/metamask-mobile/pull/12427/commits/f5b754ffb535100874030e58c3def43858e36dca)

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      name: 'RemoteFeatureFlagController',
+      allowedActions: [],
+      allowedEvents: [],
+    }),
+    state: {
+      ...initialState.RemoteFeatureFlagController
```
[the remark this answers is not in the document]
joaoloureirop: addressed on [f5b754f](https://github.com/MetaMask/metamask-mobile/pull/12427/commits/f5b754ffb535100874030e58c3def43858e36dca)

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
joaoloureirop: addressed on [f5b754f](https://github.com/MetaMask/metamask-mobile/pull/12427/commits/f5b754ffb535100874030e58c3def43858e36dca)

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
joaoloureirop: addressed on [f5b754f](https://github.com/MetaMask/metamask-mobile/pull/12427/commits/f5b754ffb535100874030e58c3def43858e36dca)

app/selectors/featureFlagController/index.ts
```diff
@@ -0,0 +1,18 @@
+import { createSelector } from 'reselect';
+import { RootState } from '../../reducers';
+
+export const selectRemoteFeatureFlagControllerState = (state: RootState) =>
+  state.engine.backgroundState.RemoteFeatureFlagController;
+
+export const selectRemoteFeatureFlags = createSelector(
```
DDDDDanica, line 7: I was thinking our daily use case. A dev/team working on a feature behind a flag, might not care about the whole dataset, but just the one team is working on. How about we narrow down the selector to selectRemoteFeatureFlagByName, while we can pass in a name string and only return the value of that key when requested? what's your insight on this one?

DDDDDanica: discussed with Dan and keep a general fetch for all feature flags is simplier [here](https://github.com/MetaMask/metamask-extension/pull/28684#discussion_r1862725486). resolved the thread now 🙌🏻

app/core/Engine/types.ts
```diff
@@ -153,6 +153,10 @@ import {
 } from '@metamask/accounts-controller';
 import { BaseState } from '@metamask/base-controller';
 import { getPermissionSpecifications } from '../Permissions/specifications.js';
+import {
+  RemoteFeatureFlagController,
+  RemoteFeatureFlagControllerState
+} from '@metamask/remote-feature-flag-controller';
```
tommasini, line 159: It seems that we will need to add this to the declarations file for now

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
Gudahtt: Nit: Maybe better to name this directory `platform` instead to preserve the team grouping, so that we can include other controllers here too in the future

app/core/Engine/controllers/RemoteFeatureFlagController/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export const createRemoteFeatureFlagController = ({
+  state,
+  messenger,
+  fetchFunction,
+  disabled,
+}: RemoteFeatureFlagInitParamTypes) => {
+
+  try {
```
Gudahtt, line 39: The controller should not be constructed in a `try` block, it's not fallible. It can't throw. If it does throw we would want the app to crash, because it would be an urgent problem that we wouldn't want buried in the log for us to discover too late.

app/core/Engine/types.ts
Gudahtt: No need, it is published, we can use the real package with real types now

app/selectors/featureFlagController/minimumAppVersion/index.ts
```diff
@@ -0,0 +1,52 @@
...
+export const selectMobileMinimumVersions = createSelector(
+  selectRemoteFeatureFlags,
+  (remoteFeatureFlags) => {
+    const remoteFeatureFlag = remoteFeatureFlags[FEATURE_FLAG_NAME];
+
+    const featureFlagValues =
+      remoteFeatureFlag
+      ?? defaultValues;
+
+    return featureFlagValues as FeatureFlagType;
```
Gudahtt, line 23: Why are we casting here? By doing this we're declaring this value to be a number, but we haven't validated that. This makes the type inaccurate.

metro.config.js
joaoloureirop: resolved e54d05d551d5522926bd7309df34670be4a84736

metro.config.js
```diff
@@ -8,17 +8,35 @@
...
+    watchFolders,
     resolver: {
       assetExts: assetExts.filter((ext) => ext !== 'svg'),
       sourceExts: [...sourceExts, 'svg', 'cjs', 'mjs'],
       resolverMainFields: ['sbmodern', 'react-native', 'browser', 'main'],
+      extraNodeModules: new Proxy (extraNodeModules, {
+        get: (target, name) =>
+          name in target ? target[name] : path.join(process.cwd(), `node_modules/${name}`),
+      }),
+      unstable_enableSymlinks: true,
```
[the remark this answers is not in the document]
joaoloureirop: resolved e54d05d551d5522926bd7309df34670be4a84736

yarn.lock
joaoloureirop: resolved e54d05d551d5522926bd7309df34670be4a84736

app/selectors/featureFlagController/minimumAppVersion/index.ts
joaoloureirop: I've changed the types on the selector to improve type inference

resolved on b003b6aae56aecf877222b38e0c11e855406b06a

app/core/Engine/controllers/RemoteFeatureFlagController/utils.ts
joaoloureirop: resolved 6f709eb4ad9fa2186b4b74a84ed251644663545e

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
joaoloureirop: I'm concerned about the potential tech debt we add when the team's name-based namings enter the codebase.

[See another discussion around this topic.](https://github.com/MetaMask/metamask-mobile/pull/12427#discussion_r1860452762)

app/core/Engine/types.ts
joaoloureirop: resolved b4fb8662a16640e71490eff858bacd64899560f2

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
joaoloureirop: resolved 96ca13b364a7de2d3004d2033ea1e7462eb38bde

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
Gudahtt: Well, we could think of it as "feature areas" instead perhaps, which roughly correlate to teams but are more long-lasting.

The purpose (as I understood it) of moving controller-related code to dedicated files was that it allowed us to assign more appropriate code owners (i.e. teams) and reduce code review burden and regression risk that comes from modifying core modules like `Engine.ts`.

But the problem is that there are very few changes that affect _just_ a controller. If we co-located multiple controllers owned by the same team, it would be more effective in minimizing cross-team impact/reviews for changes to how controllers interact with each other. If each controller is separate, then their interactions would be via `Engine.ts` still so we don't get that benefit.

app/core/Engine/Engine.ts
joaoloureirop: Yes, that would be more appropriate.

joaoloureirop: addressed 50795f03f5cda3c5a4ad2eb42ffe680fa7c59e98

app/core/Engine/Engine.ts
joaoloureirop: Good point Nico. Although false equality stands out to the eyes more than negation, I agree with you about simplifying this expression.

joaoloureirop: addressed 50795f03f5cda3c5a4ad2eb42ffe680fa7c59e98

app/selectors/featureFlagController/index.test.ts
joaoloureirop: addressed 96ca13b364a7de2d3004d2033ea1e7462eb38bde

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
joaoloureirop: addressed 96ca13b364a7de2d3004d2033ea1e7462eb38bde

joaoloureirop: addressed 96ca13b364a7de2d3004d2033ea1e7462eb38bde

app/store/migrations/061.test.ts
joaoloureirop: addressed 01261192e1cf576d31e7c6131f24d2ae78bb63c9

app/store/migrations/061.test.ts
joaoloureirop: addressed 01261192e1cf576d31e7c6131f24d2ae78bb63c9

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
joaoloureirop: addressed 96ca13b364a7de2d3004d2033ea1e7462eb38bde

app/core/Engine/controllers/RemoteFeatureFlagController/index.ts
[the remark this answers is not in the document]
joaoloureirop: I recognize the benefits of organizing the codebase in a feature-based structure. It helps with ownership and better outlines the relationship between components.

Namings that carry meaning for the codebase / product are long-lasting and do not need to be revisited if there's any maintainer's restructuring.

I'll leave it as is for now and follow up with an update when there's a decision on feature area names.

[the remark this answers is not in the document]
NicolasMassart: Just a note on the feature scope source folders naming: I totally agree with the pros, making things easier and faster to review, but do not forget about the cons:
- feature is then tied to a group/team and moving responsibility requires changing the app folders structure,
- siloing features mean less need for review from other team, so less knowledge sharing
- isolating features also tends to make code practices diverge. We see that for metametrics where we have already 3 kind of wrappers. So we will have to do even bigger efforts in standardizing and work with guidelines.

app/core/Engine/controllers/RemoteFeatureFlagController/utils.test.ts
```diff
@@ -0,0 +1,182 @@
...
+  let messenger: RemoteFeatureFlagControllerMessenger;
+
+  beforeEach(() => {
+    messenger =
+      new ControllerMessenger() as unknown as RemoteFeatureFlagControllerMessenger;
+    jest.clearAllMocks();
+  });
+
+  describe('createRemoteFeatureFlagController', () => {
+    it('should create controller with correct config when enabled', () => {
```
NicolasMassart, line 18: Can we, once and for all, agree that we have to name tests according to our guidelines and not according to whatever AI outputs?
```suggestion
    it('creates controller with correct config when enabled', () => {
```

app/core/Engine/controllers/RemoteFeatureFlagController/utils.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      expect(controller).toBeDefined();
+
+      // Initializing with am empty object should return an empty obj?
+      expect(controller.state).toStrictEqual({
+        cacheTimestamp: 0,
+        remoteFeatureFlags: {},
+      });
+    });
+
+    it('should handle initial state correctly', () => {
```
NicolasMassart, line 37: Try to avoir the use of terms like handle or correctly.
Handle means do something with... so just say waht we do exactly. Here we want to check the controller actually uses the initial state to set its internal state.
Also correctly means whatever you put in it. What is correct to you may not be correct for someone else. So be more specific, what is a correct handling?
Here a correct handling, for what I can read in the test, seems to me when the controler internal exactly state matches the initial state.
So let's be it. Name it:
```suggestion
    it('internal state matches initial state', () => {
```

app/core/Engine/controllers/RemoteFeatureFlagController/utils.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      const controller = createRemoteFeatureFlagController({
+        state: initialState,
+        messenger,
+        disabled: false,
+      });
+
+      expect(controller.state).toStrictEqual(initialState);
+    });
+
+    it('should call updateRemoteFeatureFlags when enabled', () => {
```
NicolasMassart, line 54:
```suggestion
    it('calls updateRemoteFeatureFlags when enabled', () => {
```

app/core/Engine/controllers/RemoteFeatureFlagController/utils.test.ts
```diff
@@ -0,0 +1,182 @@
...
+      createRemoteFeatureFlagController({
+        state: undefined,
+        messenger,
+        disabled: false,
+      });
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('should not throw when receive corrupted data', () => {
```
NicolasMassart, line 69:
```suggestion
    it('does not throw when initial state is corrupted', () => {
```
Also as a dev looking at this code for the first time, I would love to have this test tell me more about why this is considered corrupted data, why the controller can receive corrupted data and why we don't type data with an interface or type to prevent this corruption.

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -1,39 +1,144 @@
...
+    ).toEqual(defaultValues.appMinimumBuild);
+  });
+
+  it('undefined RemoteFeatureFlagController state: returns default value', () => {
+    expect(
+      selectAppMinimumBuild(mockedUndefinedFlagsState)
+    ).toEqual(defaultValues.appMinimumBuild);
+  });
+
+  it('empty feature flag state: returns default value', () => {
```
NicolasMassart, line 78: update naming

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -1,39 +1,144 @@
...
+    const { appleMinimumOS: mockedValue } = defaultValues;
+    expect(
+      selectAppleMinimumOS(mockedEmptyFlagsState)
+    ).toEqual(mockedValue);
+  });
+
+  it('undefined RemoteFeatureFlagController state: returns default value', () => {
```
NicolasMassart, line 100: update naming

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -1,39 +1,144 @@
...
+    const { appleMinimumOS: mockedValue } = defaultValues;
+    expect(
+      selectAppleMinimumOS(mockedUndefinedFlagsState)
+    ).toEqual(mockedValue);
+  });
+
+  it('empty feature flag state: returns default value', () => {
```
NicolasMassart, line 107: update naming

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -1,39 +1,144 @@
...
+    ).toEqual(mockedValue);
+  });
+});
+
+describe('minimumAppVersion Feature flag: androidMinimumAPIVersion selector', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('empty feature flag state: returns default value', () => {
```
NicolasMassart, line 122: update naming

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -1,39 +1,144 @@
...
+    const { androidMinimumAPIVersion: mockedValue } = defaultValues;
+    expect(
+      selectAndroidMinimumAPI(mockedEmptyFlagsState)
+    ).toEqual(mockedValue);
+  });
+
+  it('undefined RemoteFeatureFlagController state: returns default value', () => {
```
NicolasMassart, line 129: update naming

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -1,39 +1,144 @@
...
+    const { androidMinimumAPIVersion: mockedValue } = defaultValues;
+    expect(
+      selectAndroidMinimumAPI(mockedUndefinedFlagsState)
+    ).toEqual(mockedValue);
+  });
+
+  it('empty feature flag state: returns default value', () => {
```
NicolasMassart, line 136: update naming

NicolasMassart (changes requested):
update tests please

joaoloureirop:
I've revisited the controller initialisation testing. 
- unit test that can't be validated without mocking the feature flag service were removed.
- unit tests names were improved.

For the selector, the unit test names were updated

joaoloureirop:
I've revisited the controller initialisation testing. 
- unit test that can't be validated without mocking the feature flag service were removed.
- unit tests names were improved.

For the selector, the unit test names were updated

Thanks for your review @NicolasMassart

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -63,19 +63,19 @@ describe('minimumAppVersion Feature flag: appMinimumBuild selector', () => {
...
 
-  it('undefined RemoteFeatureFlagController state: returns default value', () => {
+  it('returns default value when undefined RemoteFeatureFlagController state', () => {
     expect(
       selectAppMinimumBuild(mockedUndefinedFlagsState)
     ).toEqual(defaultValues.appMinimumBuild);
   });
 
-  it('empty feature flag state: returns default value', () => {
+  it('returns default value empty feature flag state', () => {
```
NicolasMassart, line 78:
```suggestion
  it('returns default value when empty feature flag state', () => {
```

app/selectors/featureFlagController/minimumAppVersion/index.test.ts
```diff
@@ -90,21 +90,21 @@ describe('minimumAppVersion Feature flag: appleMinimumOS selector', () => {
...
-  it('undefined RemoteFeatureFlagController state: returns default value', () => {
+  it('returns default value when undefined RemoteFeatureFlagController state', () => {
     const { appleMinimumOS: mockedValue } = defaultValues;
     expect(
       selectAppleMinimumOS(mockedUndefinedFlagsState)
     ).toEqual(mockedValue);
   });
 
-  it('empty feature flag state: returns default value', () => {
+  it('returns default value empty feature flag state', () => {
```
NicolasMassart, line 107:
```suggestion
  it('returns default value when empty feature flag state', () => {
```

NicolasMassart (commented):
allowing myself to fix some typos and missing 'when' directly

NicolasMassart (approved):
Looks good to me on the code side.

PR #12485, opened 2024-11-28
chore: Stop suppressing pod install failures

Gudahtt:
## **Description**

Pod install failures were suppressed originally to make cross-platform development easier, as they would fail on Windows and Linux (where you would presumably be developing with Android which does not require this step). However, since the introduction of the new setup script, this step will not run on non-macOS machines by default so this error suppression is no longer needed.

This will make install errors easier to discover, saving developer time.

## **Related issues**

N/A

## **Manual testing steps**

To test that there is no regression, simply run `yarn setup`.

I'm not entirely sure how to induce a pod install error. But if you're in a state where pod install fails, then `yarn setup` should now fail as well, where previously it would succeed.

I simulated a failure by updating `pod:install` to say `bundle exec po install` (i.e. delete the `d` from `pod` so that it's an unrecognized command). This failure showed on the CLI as expected on this branch. On `main` it silently failed.

## **Screenshots/Recordings**

N/A

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -39,6 +39,6 @@ jobs:
             --rm \
             -v "$(pwd):/app" -w /app \
             metamask-mobile-builder:latest \
-            bash -c 'yarn && yarn setup --build-ios'
+            bash -c 'yarn && yarn setup --build-ios --no-install-pods'
             # restore ownership for cleanup
             sudo chown -R "$(id -u):$(id -g)" .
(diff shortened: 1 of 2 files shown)


Gudahtt:
Depends upon #12574

.github/workflows/docker.yml
```diff
@@ -39,6 +39,6 @@ jobs:
             --rm \
             -v "$(pwd):/app" -w /app \
             metamask-mobile-builder:latest \
-            bash -c 'yarn && yarn setup --build-ios'
+            bash -c 'yarn && yarn setup --build-ios --no-install-pods'
```
Gudahtt, line 42: It seems that the `pod:install` step was failing already. I'm not sure why, or when the failure started, because previously the failure was suppressed.

I've added this flag to skip the pod install step so that this pre-existing failure doesn't block this improvement.

tommasini (approved):

PR #12599, opened 2024-12-06
feat: MultichainBalanceController

owencraston:
## **Description**
Builds off of the work done in [this PR](https://github.com/MetaMask/core/pull/4965).

- This PR adds the ability to see your non EVM token balance and the fiat token rates (Solana and Bitcoin accounts)
- This is achieved by using the RatesController and the MultichainBalanceController
- This PR also extract the balance logic from the PortfolioBalance PR into the useMultichainBalances hook
- creates the selectors for multichain non evm

This PR DOES NOT...
1. show the correct native token balance in the account picker. This will be done in a [following PR](https://github.com/MetaMask/accounts-planning/issues/697) that builds off of this work.
2. enable token details for these non evm assets. (for now it will simply show no data)

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/696

## **Manual testing steps**

1. checkout this branch
3. open .js.env
4. set export METAMASK_BUILD_TYPE="flask" // flask
5. source .js.env
6. yarn setup
7. yarn start:ios
8. create/import a wallet
9. open the account menu actions
10. you should see that you can add a BTC and BTC Testnet account
11. click Add a new Bitcoin account (Beta)
12. a popup with a suggested name should appear click confirm
13. the newly selected account should be your Bitcoin account and the network picker should show the Bitcoin logo the network should not be clickable
14. Optionally, create a testnet account and fund it with this faucet: https://bitcoinfaucet.uo1.net/
15. If you have a balance, you should see the fiat balance rendered on the wallet
16. if not you have a balance, then you should see a 0 fiat balance.
17. If you change your selected currency (settings/general) and then return to the main wallet view, the balance should update based on the new fiat currency.
18. The token list should also list your account balance in both fiat and and the native token balance.
19. the token displayed should be the BTC logo and not have any network badge above it (like the eth account does)
20. now add a bitcoin TestNet account.
21. the portfolio balance should show the native token balance not the fiat balance.
22. repeat the above steps but this time add a new Solana account. If your Solana account has a balance, it should be rendered on the home page in fiat. If not you should see a 0 balance.
23. we do not currently support Solana testnet accounts.

### **After**

https://github.com/user-attachments/assets/12d4bfeb-608f-4744-9c0a-459163eb0dba


app/reducers/multichain/index.ts
```diff
@@ -1,8 +1,8 @@
 import { MultichainSettingsState } from '../../actions/multichain/state';
 
 export const initialState: MultichainSettingsState = {
-  bitcoinSupportEnabled: false,
-  bitcoinTestnetSupportEnabled: false,
+  bitcoinSupportEnabled: true,
```
owencraston, line 4: set back to false before merging

Gedoxa (approved):

app/components/hooks/usePortfolioBalance/usePortfolioBalance.ts
```diff
@@ -0,0 +1,253 @@
...
+        totalFiatBalancesCrossChain[selectedInternalAccount?.address as string]
+          ?.totalTokenFiat ?? 0,
+      shouldShowAggregatedPercentage: !isTestNet(chainId),
+      shouldShowPortfolioButton: true,
+      isPortfolioEnabled: isPortfolioViewEnabled(),
+      aggregatedBalance: getAggregatedBalance(),
+    },
+    actions: {
+      togglePrivacyMode: (value: boolean) =>
+        PreferencesController.setPrivacyMode(value),
```
gambinish, line 247: For `privacyMode` and `togglePrivacyMode` my preference would be to keep these as props getting passed into the React component, rather than abstracting them into a hook.

Toggling/reading `privacyMode` is unrelated to differentiating between evm and non-evm balances (which it seems like this hook is meant to do). Imo having this adds unnecessary scope and complexity to this hook. Having this broken out separately will make this more straightforward to control and troubleshoot.

app/hooks/usePortfolioBalance/index.ts
```diff
@@ -0,0 +1,211 @@
...
+  shouldShowAggregatedPercentage: boolean;
+  shouldShowPortfolioButton: boolean;
+}
+
+export interface PortfolioBalanceActions {
+  togglePrivacyMode: (value: boolean) => void;
+  onOpenPortfolio: () => void;
+}
+
+export const usePortfolioBalance = (): {
```
gambinish, line 66: Is this a duplicate hook? It is very similar to `app/components/hooks/usePortfolioBalance`

What are the key differences between these? If we need both pieces of code, I wonder if we should change some of the naming conventions to make it easier to understand which hook needs to be used in which scenarios.

gambinish (changes requested):
Thanks again for reminding me about this!

I mostly reviewed the logic in the `usePortfolioBalance` hooks, and left a few comments. Overall I think the approach makes sense. I do see a couple places where we might be able to simplify the business logic a bit, and also a question about some potential duplication of code.

app/hooks/usePortfolioBalance/index.ts
owencraston: Oops, I initially created this hook in this folder but moved it into `app/components/hooks/usePortfolioBalance` but forgot to delete the old version.

app/components/hooks/usePortfolioBalance/usePortfolioBalance.ts
owencraston: fair point, I removed both actions (privacy mode and the opening of the portfolio from the hook in favour of having it inline in the component.

app/components/hooks/usePortfolioBalance/usePortfolioBalance.ts
```diff
@@ -0,0 +1,253 @@
...
+    ///: END:ONLY_INCLUDE_IF
+    return true;
+  };
+
+  return {
+    data: {
+      displayBalance: getDisplayBalance(),
+      privacyMode,
+      isEvm: getIsEvm(),
+      isTestNetwork: isTestNet(chainId),
```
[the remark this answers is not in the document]
owencraston: Initially I thought adding these would simplify the view but I realize this is adding too many concerns into this hook that should really remain balance focused. I have removed them now.

app/core/Engine/controllers/RatesController/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  initialState,
+}: {
+  messenger: RatesControllerMessenger;
+  initialState?: RatesControllerState;
+}): RatesController => {
+  try {
+    const multichainRatesController = new RatesController({
+      messenger,
+      state: initialState ?? defaultRatesControllerState,
+      includeUsdRate: true,
```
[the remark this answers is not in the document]
owencraston: I have now setup a subscription for whenever the selected fiat currency in is changed in the `CurrencyRateController`. You can find this logic in `app/core/Engine/controllers/RatesController/subscriptions.ts`. This will work for most "normal" currencies but there are a few supported currencies (like TokenList) that will not work with the new rates controller. Since this is a flask only feature for now I think this is alright.

brianacnguyen (changes requested):
Can you attach a screenshot/screencap of the AggregatedPercentage story?

app/components/UI/NonEvmTokens/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+      ticker: defaultToken.symbol,
+    },
+  ];
+
+  const onRefresh = () => {
+    setRefreshing(true);
+    setTimeout(() => {
+      setRefreshing(false);
+    }, 1000);
+  };
```
gambinish, line 71: Is this function doing anything? Seems like it should it be calling a controller of some sort.

app/components/UI/NonEvmTokens/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const showRemoveMenu = useCallback(() => {
+    // Bitcoin can't be removed as it's the native token
+  }, []);
+
+  const goToAddToken = useCallback(() => {
+    // Bitcoin doesn't support adding tokens yet
+  }, []);
```
gambinish, line 79: Are these placeholders? Maybe we can pass these callbacks in as `null` instead and use the nullish value to conditionally render the buttons/links that call these callbacks in the component.

app/components/UI/Tokens/TokenList/PortfolioBalance/index.tsx
```diff
@@ -46,88 +30,22 @@ import { BrowserTab } from '../../types';
...
-      return tokenFiatTotal + ethFiatTotal;
-    }
-    if (isPortfolioViewEnabled()) {
-      return totalTokenFiat ?? 0;
-    }
-    return balance?.tokenFiat ?? 0;
-  }, [isOriginalNativeTokenSymbol, totalFiatBalance, totalTokenFiat, balance]);
-
-  const fiatBalance = `${renderFiat(total, currentCurrency)}`;
+  const { data } = useMultichainBalances();
```
gambinish, line 48: Nit: Might make sense/be more readable to reassign data to something more descriptive.

`const { data: multchainBalances } = useMultichainBalances();`

app/components/UI/Tokens/index.tsx
```diff
@@ -129,6 +129,8 @@ const Tokens: React.FC<TokensI> = memo(({ tokens }) => {
     isPortfolioViewEnabled() ? selectAccountTokensAcrossChains(state) : {},
   );
 
+  console.log('selectedAccountTokensChains', selectedAccountTokensChains);
+
```
gambinish, line 133: Note to remove

app/components/Views/Wallet/index.tsx
```diff
@@ -663,30 +723,8 @@ const Wallet = ({
           </View>
         ) : null}
         <>
-          {accountBalanceByChainId && <PortfolioBalance />}
```
owencraston, line 666: @gambinish are there any side effects or removing this check here? We are fetching the balances inside the useMultichainBalances hook so I see no reason we should conditionally render this component here.

app/components/Views/Wallet/index.tsx
```diff
@@ -586,6 +594,58 @@ const Wallet = ({
...
+  function renderTokensContent(assets: Token[]) {
+    ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+    return isEvm ? (
+      <ScrollableTabView
+        renderTabBar={renderTabBar}
+        // eslint-disable-next-line react/jsx-no-bind
+        onChangeTab={onChangeTab}
+      >
+        <Tokens
+          // @ts-expect-error tabLabel prop is used by ScrollableTabView parent
```
gambinish, line 606: Is there any way that we can remove this expect-error?

app/components/hooks/useMultichainBalances/useMultichainBalances.ts
```diff
@@ -0,0 +1,178 @@
...
+        totalFiatBalancesCrossChain[selectedInternalAccount?.address as string]
+          ?.tokenFiatBalancesCrossChains ?? [],
+      totalFiatBalance:
+        totalFiatBalancesCrossChain[selectedInternalAccount?.address as string]
+          ?.totalFiatBalance ?? 0,
+      totalTokenFiat:
+        totalFiatBalancesCrossChain[selectedInternalAccount?.address as string]
+          ?.totalTokenFiat ?? 0,
+      shouldShowAggregatedPercentage: getShouldShowAggregatedPercentage(),
+      isPortfolioEnabled: isPortfolioViewEnabled(),
```
gambinish, line 172: Can't wait until we can simply remove this feature flag entirely 🥲

gambinish: In the meantime, can we pass this along as `isPortfolioVieEnabled` rather than `isPortfolioEnabled` so that we can have an easier time when it does come down to removing this?

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,304 @@
...
+      : undefined;
+
+    if (!nonEvmNetwork) {
+      throw new Error(
+        'Could not find non-EVM provider for the current configuration. This should never happen.',
+      );
+    }
+
+    return {
+      // TODO: Adapt this for other non-EVM networks
```
gambinish, line 133: Is this TODO still relevant?

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,304 @@
...
+);
+
+const selectNonEvmCachedBalance = createDeepEqualSelector(
+  selectSelectedInternalAccount,
+  selectMultichainBalances,
+  selectMultichainCurrentNetwork,
+  (selectedInternalAccount, multichainBalances, multichainCurrentNetwork) => {
+    if (!selectedInternalAccount) {
+      console.warn('Could not find selected internal account');
+      return 0;
```
gambinish, line 232: Nit: Is `0` what we would expect the balance to be if account is not found? It could be, I'm just curious if `undefined` or something along those lines could also work.

app/components/UI/NonEvmTokens/index.tsx
owencraston: Good point. I modified it to call `MultichainBalancesController.updateBalances()`

app/components/UI/NonEvmTokens/index.tsx
owencraston: I need to play with the types so that they accept null as an option.

app/components/UI/Tokens/TokenList/PortfolioBalance/index.tsx
owencraston: I like this. I have made this change in my latest commit.

app/components/UI/Tokens/index.tsx
owencraston: Removed!

app/selectors/multichain/multichainNonEvm.ts
owencraston: I would prefer to return 0 so that we can handle this case gracefully instead of crashing the app or not rendering anything.

app/components/Views/Wallet/index.tsx
owencraston: yeah good call.

app/components/hooks/useMultichainBalances/useMultichainBalances.ts
owencraston: I prefer this name. I will make this change.

app/selectors/multichain/multichainNonEvm.ts
owencraston: I believe it is since we are still hardcoding some EVM logic in there.

app/core/Engine/Engine.ts
```diff
@@ -377,6 +385,41 @@ export class Engine {
...
+    const multichainBalancesControllerMessenger: MultichainBalancesControllerMessenger =
+      this.controllerMessenger.getRestricted({
+        name: 'MultichainBalancesController',
+        allowedEvents: [
+          AccountsControllerAccountAddedEvent,
+          AccountsControllerAccountRemovedEvent,
+        ],
+        allowedActions: [
+          AccountsControllerListMultichainAccountsAction,
+          'SnapController:handleRequest',
```
ccharly, line 398: What about creating a `app/core/Engine/controllers/SnapController/constants.ts` for this one?

We will need other actions later on when updating the `SnapKeyring` to the latest version (which relies on "actions" rather than a instance to the `SnapController`)

app/components/UI/Tokens/TokenList/PortfolioBalance/index.tsx
```diff
@@ -46,88 +30,22 @@ import { BrowserTab } from '../../types';
...
-  );
-  const currentCurrency = useSelector(selectCurrentCurrency);
   const browserTabs = useSelector((state: RootState) => state.browser.tabs);
   const privacyMode = useSelector(selectPrivacyMode);
-
-  const isOriginalNativeTokenSymbol = useIsOriginalNativeTokenSymbol(
-    chainId,
-    ticker,
-    type,
+  const ismultchainBalancesCollectionForMarketingEnabled = useSelector(
```
ccharly, line 42: Same typo here: 

```suggestion
  const isMultichainBalancesCollectionForMarketingEnabled = useSelector(
```

app/components/hooks/useMultichainBalances/useMultichainBalances.ts
```diff
@@ -0,0 +1,180 @@
...
+    ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+    return !isTestNet(chainId) && isEvm;
+    ///: END:ONLY_INCLUDE_IF
+
+    // Note: This code marked as unreachable however when the above block gets removed after code fencing this return becomes necessary
+    return !isTestNet(chainId);
+  };
+
+  return {
+    multchainBalances: {
```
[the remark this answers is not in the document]
owencraston: oops 😢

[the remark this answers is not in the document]
owencraston: serves me right for making changes at midnight.

app/components/hooks/useMultichainBalances/useMultichainBalances.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    const { result } = renderHook(() => useMultichainBalances());
+
+    expect(result.current.multchainBalances.displayBalance).toBe('150 USD');
+    expect(result.current.multchainBalances.aggregatedBalance).toEqual({
+      ethFiat: 100,
+      tokenFiat: 50,
+      tokenFiat1dAgo: 45,
+      ethFiat1dAgo: 95,
+    });
```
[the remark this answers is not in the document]
owencraston: I like it.

app/components/hooks/useMultichainBalances/useMultichainBalances.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      totalTokenFiat: 0,
+      shouldShowAggregatedPercentage: true,
+      isPortfolioVieEnabled: false,
+      aggregatedBalance: {
+        ethFiat: 0,
+        tokenFiat: 0,
+        tokenFiat1dAgo: 0,
+        ethFiat1dAgo: 0,
+      },
+    });
```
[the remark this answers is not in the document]
owencraston: Great Nit

app/core/Engine/controllers/RatesController/subscriptions.ts
```diff
@@ -0,0 +1,37 @@
...
+  controllerMessenger: {
+    subscribe: (
+      eventName: string,
+      handler: (state: CurrencyRateState) => void,
+    ) => void;
+  },
+  ratesController: RatesController,
+): void => {
+  controllerMessenger.subscribe(
+    `${CurrencyRateController.name}:stateChange`,
```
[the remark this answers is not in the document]
owencraston: Great idea.

app/core/Engine/Engine.ts
owencraston: Okay. I went ahead and made that change (as well as extracted the remaining SnapController, RatesController and MultichainBalances type into constants files). I did not fully migrate these controller to the factory pattern in order to try and keep the changes minimal.

app/components/UI/NonEvmTokens/index.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+  },
+  settings: {
+    showTestNetworks: true,
+  },
+};
+
+const Stack = createStackNavigator();
+// TODO: Replace "any" with type
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const renderComponent = (state: any = {}) =>
```
tommasini, line 109: Can we use `RootState` instead?

brianacnguyen (dismissed):

app/components/UI/NonEvmTokens/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+      ];
+    return imageSource ? Image.resolveAssetSource(imageSource).uri : '';
+  };
+
+  // Format the token data to match TokenI interface
+  const formattedTokens: TokenI[] = [
+    {
+      address: '', // Non-EVM chains don't use EVM-style addresses for native tokens
+      aggregators: [],
+      decimals: network.chainId === MultichainNetworks.Solana ? 9 : 6, // SOL uses 9 decimals, BTC uses 6
```
wachunei, line 62: is this decimals value the actual minimal unit or a format representation for the UI? Because BTC hast 8 decimals (satoshi is 10e-8 btc)

either way, these probably should be constants,

app/components/Views/AddAccountActions/AddAccountActions.tsx
```diff
@@ -173,7 +173,7 @@ const AddAccountActions = ({ onBack }: AddAccountActionsProps) => {
               )}
               iconName={IconName.Add}
               onPress={async () => {
-                await createBitcoinAccount(MultichainNetworks.BITCOIN);
+                await createBitcoinAccount(MultichainNetworks.Bitcoin);
```
wachunei, line 176: Is there a reason why `createSolanaAccount` and `createBitcoinAccount` need te receive the chain ID as param? is it just kept as is because of legacy code?

`await createSolanaAccount(MultichainNetworks.Bitcoin);` would not make sense

wachunei: aah I saw `createBitcoinAccount` can receive the testnet chain id as scope. nevermind, resolve these comments :)

app/selectors/multichain/multichainEvm.ts
```diff
@@ -4,21 +4,17 @@ import { Token, getNativeTokenAddress } from '@metamask/assets-controllers';
 import {
```
tommasini, line 4: Let's sync on this, it seems like this file is only balances related, would make sense to rename it to `multichainEvmBalances?` Let me know your thoughts

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
+///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
```
tommasini, line 1: It seems that this selectors are not nonEvm specific! Since we are both working on this, let's sync to understand if they are all being used, and compare some similar behaviours that we have, to only have one source of truth like the `solanaEnabled`!

app/core/Multichain/utils.ts
```diff
@@ -90,4 +90,16 @@ export function isBtcTestnetAddress(address: string): boolean {
...
+
+/**
+ * Returns whether an address is a valid Solana address, specifically an account's.
+ * Derived addresses (like Program's) will return false.
+ * See: https://stackoverflow.com/questions/71200948/how-can-i-validate-a-solana-wallet-address-with-web3js
+ *
+ * @param address - The address to check.
+ * @returns `true` if the address is a valid Solana address, `false` otherwise.
+ */
+export function isSolanaAddress(address: string): boolean {
```
tommasini, line 102: Out of curiosity, you created an util function to use across the codebase around one util function because it's from a third party repo?
I liked it a lot, if we want to shift from the library it would be much easier! Cool!

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/**
+ * Get the state of the `solanaSupportEnabled` flag.
+ *
+ * @param {*} state
+ * @returns The state of the `solanaSupportEnabled` flag.
+ */
+export function selectIsSolanaSupportEnabled(state: RootState) {
+  return state.multichainSettings.solanaSupportEnabled;
```
tommasini, line 52: Not sure if we want to store this data in the user device, any strong thoughts on it? I think it would be okay to use a environment variable instead, wdyt?

app/components/Views/AddAccountActions/AddAccountActions.tsx
owencraston: Yeah the reason is so we can support testnet accounts. Solana does not have test net accounts yet but we can easily enable it in the future.

app/components/UI/NonEvmTokens/index.tsx
owencraston: good idea, I have added it as part of the network config response.

app/core/Multichain/constants.ts
```diff
@@ -1,8 +1,132 @@
...
+import { BtcAccountType, SolAccountType } from '@metamask/keyring-api';
+import BTC from '../../images/bitcoin-logo.png';
+import SOL from '../../images/solana-logo.png';
 
-  SOLANA = 'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp',
-  SOLANA_DEVNET = 'solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1',
-  SOLANA_TESTNET = 'solana:4uhcVJyU9pJkvQyS88uRDiswHXSCkY3z',
+// Image imports for React Native rendering
+export const MULTICHAIN_TOKEN_IMAGES = {
+  [MultichainNetworks.Bitcoin]: BTC,
```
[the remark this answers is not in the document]
owencraston: Are you sure? I do not see the same data available in the keyring API? Even if it did, this type is easier to work with since it is the natural response from the assets controller.

app/selectors/multichain/multichainNonEvm.ts
owencraston: It wont be persisted since I put this selector state in the [persistConfig blacklist](https://github.com/MetaMask/metamask-mobile/blob/dd957381894c94faf073b3a3d5a105a212052a1e/app/store/persistConfig.ts#L111).

app/selectors/multichain/multichainEvm.ts
owencraston: fair point, There will likely be more "types" of selectors added to this file in the future so I kept it generic.

app/selectors/multichain/multichainNonEvm.ts
owencraston: yeah I explicitly broke them up for now to try and not step on other peoples toes but you are right, eventually we will want t migrate all selectors to the multichain version.

app/selectors/multichain/multichainNonEvm.ts
owencraston: On the other hand I think returning 0 could lead users to panic so it would be better to return undefined and show a loading indicator.

app/selectors/multichain/multichainEvm.ts
owencraston: this file was also originally not mine. I simply migrated the multichain.ts file to the new name multihainEvm.ts.

app/components/UI/NonEvmTokens/index.tsx
owencraston: for now I think passing in an empty function works and keeps the changes minimal? what do you think?

app/selectors/multichain/multichainNonEvm.ts
owencraston: I made the change to return undefined here.

Prithpal-Sooriya:
@owencraston is there any additional environment variables I need?

I am unable to fetch Solana and Bitcoin assets, and the app crashes if I try clicking into these assets. See loom recording
https://www.loom.com/share/4093ad17a61c45d79b9b33ab243b3324?sid=3b113a73-1096-481a-aa7a-2a9dbc336f19

owencraston:
@Prithpal-Sooriya as long as you set the `METAMASK_BUILD_TYPE` to flask inside the .js.env you should be good to create and pull the balances. Make sure the server is running. as far as the other errors go, I will look into those.

gambinish:
Noticing the following issue on the Account Picker component:

Screenshot 2025-02-05 at 10 18 07 AM

The amount showed suggests that my Bitcoin Account should have a balance of ETH, should this be BTC? It's also not reflecting my BTC balance in my Testnet account I created:

Screenshot 2025-02-05 at 10 19 11 AM

gambinish:
Also, we may want to considering piping in the balance from the token list item into the token detail page:

Screenshot 2025-02-05 at 10 22 28 AM

However, I'm not sure if this token details view is in scope, since it seems that we are still missing market data related to BTC token. Maybe we should consider deactivating the navigation on non-EVM assets. Has product weighed in at all here?

gambinish (changes requested):
Code is looking good. I was able to pull down this branch and get it setup with iOS. I verified the testing steps you outlined, and was able to add a Testnet bitcoin account, and funded it from a faucet.

The balance updated on the token list page ✅ 
However I noticed a few places in the UI where I saw some unexpected values ♻️ 

Not sure what the scope of this feature is, we may want to pull in product to weigh in here.

gantunesr (commented):
I still need to review 10 files

gantunesr:
> Also, we may want to considering piping in the balance from the token list item into the token detail `page:`
> However, I'm not sure if this token details view is in scope, since it seems that we are still missing market data related to BTC token. Maybe we should consider deactivating the navigation on non-EVM assets. Has product weighed in at all here? 
> Sidenote that on the token details page, I'm still seeing the ETH mainnet network logo

This should be out of scope, in extension the navigation to the details view is deactivate

owencraston:
@Prithpal-Sooriya I am not experiencing the same issues as you (see video below). Try and pull the latest and see if you can get it to work?

https://github.com/user-attachments/assets/83571c29-d066-461e-8428-0ceed42f20e0

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * This function extracts the `network` field from the result of `selectMultichainNetwork(state)`,
+ * which is expected to be a `MultichainProviderConfig` object. The naming might suggest that
+ * it returns a network, but it actually returns a provider configuration specific to a multichain setup.
+ *
+ * @returns The current multichain provider configuration.
+ */
+export const selectMultichainProviderConfig = createDeepEqualSelector(
+  selectMultichainCurrentNetwork,
+  (multichainCurrerntNetwork) => multichainCurrerntNetwork.network,
```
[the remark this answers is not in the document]
owencraston: good catch.

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
owencraston: yeah I know. I am no sure how to reconcile this because the previous multichain.ts file was quite large and I did not want to edit it in order to better isolate these changes. This file could be thought of working across Evm AND Non evm where as the multichainEvm file only handles evm. I think eventually we would want to migrate all of our key selectors to the non evm. What would you suggest as the best path forward at this moment?

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+);
+
+export const selectMultichainShouldShowFiat = createDeepEqualSelector(
+  selectMultichainIsMainnet,
+  selectMultichainIsEvm,
+  selectShowFiatInTestnets,
+  (multichainIsMainnet, isEvm, shouldShowFiatOnTestnets) => {
+    const isTestNet = !multichainIsMainnet;
+    if (isEvm) {
+      return isTestNet ? shouldShowFiatOnTestnets : true; // Is it safe to assume that we default show fiat for mainnet?
```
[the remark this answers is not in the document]
owencraston: Yeah I cannot think of a good reason either but I noticed that we had an existing setting to display fiat balance in test nets so in the spirit of backwards compatibility I added it. do you think I should remove it?

app/core/Multichain/constants.ts
```diff
@@ -1,8 +1,138 @@
...
+    rpcUrl: '', // not used
+    ticker: 'BTC',
+    nickname: 'Bitcoin',
+    id: 'btc-mainnet',
+    type: 'rpc',
+    decimal: 8,
+    rpcPrefs: {
+      blockExplorerUrl:
+        MULTICHAIN_NETWORK_BLOCK_EXPLORER_URL_MAP[MultichainNetworks.Bitcoin],
+    },
```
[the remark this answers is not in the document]
owencraston: This is how it was done on the [extesnion](https://github.com/MetaMask/metamask-extension/blob/be2b439924773a9dc5995c21ac33b446572a86e4/shared/constants/multichain/networks.ts#L79).

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+  },
+);
+
+const selectNonEvmCachedBalance = createDeepEqualSelector(
+  selectSelectedInternalAccount,
+  selectMultichainBalances,
+  selectMultichainCurrentNetwork,
+  (selectedInternalAccount, multichainBalances, multichainCurrentNetwork) => {
+    if (!selectedInternalAccount) {
+      console.warn('Could not find selected internal account');
```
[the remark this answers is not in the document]
owencraston: We added them in the [extension selectors](https://github.com/MetaMask/metamask-extension/blob/01c9fbd76ca73d1012f30376cc6cfff8b4769b03/ui/selectors/multichain.ts#L391) so I added them here. I can remove them if you think that is better?

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+  (multichainCurrerntNetwork) => multichainCurrerntNetwork.network,
+);
+
+export const selectMultichainDefaultToken = createDeepEqualSelector(
+  selectMultichainIsEvm,
+  selectProviderConfig,
+  selectMultichainProviderConfig,
+  (isEvm, providerConfig, multichainProviderConfig) => {
+    const symbol = isEvm
+      ? providerConfig?.ticker ?? 'ETH'
```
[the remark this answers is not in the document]
owencraston: I was following this [extension selector](https://github.com/MetaMask/metamask-extension/blob/01c9fbd76ca73d1012f30376cc6cfff8b4769b03/ui/selectors/multichain.ts#L313). I can try and find a better way.

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+      : multichainProviderConfig?.ticker;
+    return { symbol };
+  },
+);
+
+export const selectMultichainIsBitcoin = createDeepEqualSelector(
+  selectMultichainIsEvm,
+  selectMultichainDefaultToken,
+  (isEvm, token) => !isEvm && token.symbol === 'BTC',
```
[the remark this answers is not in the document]
owencraston: I will use the ticker defined in our network config. 

```ts
export const selectMultichainIsBitcoin = createDeepEqualSelector(
  selectMultichainIsEvm,
  selectMultichainDefaultToken,
  (isEvm, token) =>
    !isEvm &&
    token.symbol ===
      MULTICHAIN_PROVIDER_CONFIGS[MultichainNetworks.Bitcoin].ticker,
);
```

app/components/Views/Wallet/index.tsx
```diff
@@ -586,6 +594,47 @@ const Wallet = ({
...
+      evmTokenContent
+    ) : (
+      <ScrollableTabView renderTabBar={renderTabBar} onChangeTab={onChangeTab}>
+        <NonEvmTokens tabLabel={strings('wallet.tokens')} key={'tokens-tab'} />
+      </ScrollableTabView>
+    );
+    ///: END:ONLY_INCLUDE_IF
+
+    // Note: This code marked as unreachable however when the above block gets removed after code fencing this return becomes necessary
+    return evmTokenContent;
```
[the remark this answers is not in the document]
owencraston: which part? In the production case, this function will only ever return `evmTokenContent`

app/core/Multichain/constants.ts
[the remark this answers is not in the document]
gantunesr: damn, ok

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
gantunesr: My personal preference is to remove logs since they make overall devex a bit more noisy for the rest. Again, this is a personal preference, not sure what is the norm here

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
gantunesr: Maybe calling this one `multichain` and the other just `evm`? Not sure if this would be better, looking at the file path it would make more sense for evm at least (`multichain/evm.ts`) while the other would be redundant (`multichain/multichain.ts`).

app/components/Views/Wallet/index.tsx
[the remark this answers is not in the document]
gantunesr: Is just the fact that you'll have an unreachable `return` in a version of the code, what do you think about making it more explicit just for the sake of removing the comment? Something like,

```ts
// ...

if (!isEvm) {
  return <>...</>
}
///: END:ONLY_INCLUDE_IF

return evmTokenContent;

```

brianacnguyen (dismissed):

app/selectors/multichain/multichainNonEvm.ts
```diff
@@ -0,0 +1,293 @@
...
+      ? nonEvmNetworks.find((provider) =>
+          provider.isAddressCompatible(selectedAccount.address),
+        )
+      : undefined;
+
+    if (!nonEvmNetwork) {
+      throw new Error(
+        'Could not find non-EVM provider for the current configuration. This should never happen.',
+      );
+    }
```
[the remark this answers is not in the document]
owencraston: fair point. I'll make that change.

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
owencraston: I like this better. I have made this change in my latest commit.

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
owencraston: I removed all of the console.warn logs from this file.

app/components/Views/Wallet/index.tsx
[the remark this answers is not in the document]
owencraston: yeah this is much better actually. I have made the change in my latest commit and removed the constant.

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
owencraston: I realized that this ternary isnt even needed since the default response from the `selectProviderConfig` will always return ETH as the ticker.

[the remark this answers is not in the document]
owencraston: the new code is much simpler and more robust since it relies on existing network selctors.

```ts
export const selectMultichainDefaultToken = createDeepEqualSelector(
  selectMultichainIsEvm,
  selectEvmProviderConfig,
  selectMultichainProviderConfig,
  (isEvm, evmProviderConfig, multichainProviderConfig) => {
    const symbol = isEvm
      ? evmProviderConfig.ticker
      : multichainProviderConfig.ticker;
    return { symbol };
  },
);
```

owencraston:
> Noticing the following issue on the Account Picker component:
> 
> Screenshot 2025-02-05 at 10 18 07 AM
> The amount showed suggests that my Bitcoin Account should have a balance of ETH, should this be BTC? It's also not reflecting my BTC balance that I see on the token list for my Testnet account I created:
> 
> Screenshot 2025-02-05 at 10 19 11 AM


@gambinish  This change is out of scope for this PR. [The next issue](https://github.com/MetaMask/accounts-planning/issues/697) that I tackle will solve this problem as it will piggy back on the changes made in this PR. I am not doing this now since the changes for this PR are already massive and this is a flask only feature so none of this flawed Ux will make it into production.

owencraston:
> Also, we may want to considering piping in the balance from the token list item into the token detail page:
> 
> Screenshot 2025-02-05 at 10 22 28 AM
> However, I'm not sure if this token details view is in scope, since it seems that we are still missing market data related to BTC token. Maybe we should consider deactivating the navigation on non-EVM assets. Has product weighed in at all here?
> 
> Sidenote that on the token details page, I'm still seeing the ETH mainnet network logo

@gambinish I would prefer to handle this in a separate PR in order to keep the changes minimal (or at least not let them grow beyond where they have now).  This is a flask only feature so it will never make it into production, therefore I feel comfortable shipping this as it. The token details page handles this lack of data fairly gracefully already so the UX is not that bad IMO but I will defer to @gantunesr.

gambinish (approved):

app/selectors/multichain/multichainNonEvm.ts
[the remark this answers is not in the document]
gantunesr: hmm sounds like a product decision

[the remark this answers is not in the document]
owencraston: discussed with @Montoya on slack and we decided to keep it as is since the pattern already exists. We also agreed this is a very weird feature that should not be used.

brianacnguyen (approved):

gantunesr (approved):

FrederikBolding (approved):

tommasini (approved):

PR #12746, opened 2024-12-18
feat: add user earning history graph and list on history button click

nickewansmith:
## **Description**

This PR adds a pooled staking user rewards history button to the asset detail page for ETH. This button leads to a bar chart graph of the user rewards over a 2 year period. A user can switch to 7D, M or Y to get 7 days, monthly or yearly data in the graph and can select different bars to see the amount of that bar's time period in the chart header. Default view with no selected bars shows the total amount of rewards earned.

To add a chart view for the user to look at their historical rewards in detail

A chart view has been added that will allow the user to look at their historical rewards in detail

## **Related issues**

[Fixes:](https://consensyssoftware.atlassian.net/browse/STAKE-705)

## **Manual testing steps**

- Go to the asset view for eth by clicking on eth or staked eth in the token list on the home page
- Note the total rewards on the asset detail page, it should be the same as in the chart
- Click the button for earnings history where staking details are to see the chart, it should load
- If there is any delay, there should be placeholders for the loading state
- If there is an error and there is no data for some reason, the loading state is currently the default
- The chart should show the total rewards and when you click on a bar that bar should become highlighted and stay highlighted until clicked again or until another bar is clicked. When highlighted the graph should show that bars info and that info should match the history list
- In the chart, we do not skip zero amount time periods
- In the list we do skip tailing zero amount time period but not any that are in between time periods with amounts
- Clicking each time period in the top of the chart should switch the data to that time period and unselect any selected bar
-  The chart should be fairly quick to load and use and there should be no lagging actions

### **Before**

No chart or button n/a

### **After**

https://github.com/user-attachments/assets/750e6dc3-7055-48a2-a579-d76242e10c3e


app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
```diff
@@ -0,0 +1,334 @@
...
+        dateStr: gapDate.toISOString().split('T')[0],
+        dailyRewards: '0',
+        dailyRewardsUsd: '0',
+        sumRewards: '0',
+      });
+    }
+    return gapFilledEarningsHistory;
+  }, [earningsHistory]);
+
+  const { earningsHistoryChartData, earningsHistoryListData } = useMemo(() => {
```
Matt561, line 161: Subjective: This function is quite large. Is there anything we can do to make it a better easier to read/understand? Imo, it might make sense to breakout chunks of related logic into separate functions and call them here.

This way this function would be easier to understand at a glance with the more complex logic out of sight until we explicitly need to view it.

You could argue for or against this of course, just sharing my 2c!

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+        setLastToggledBarIndex(-1);
+      }
+    }
+  };
+  const handleTouch = (evt: GestureResponderEvent) => {
+    updateBarHoveredBarIndex(evt.nativeEvent.locationX);
+  };
+
+  // update bar fill color on index change
+  useMemo(() => {
```
Matt561, line 152: How come we're using `useMemo` here but not outputting the memoized value to a variable?

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.tsx
```diff
@@ -0,0 +1,112 @@
...
+                        >
+                          + {earning.amount} {ticker}
+                        </Label>
+                      </View>
+                      <View>
+                        <Text
+                          variant={TextVariant.BodySM}
+                          color={colors.text.alternative}
+                        >
+                          {earning.amountUsd} USD
```
Matt561, line 84: Do we want the fiat currency to be hardcoded here if we're able to swap currencies in the app settings?

app/components/UI/Stake/utils/date/index.ts
```diff
@@ -0,0 +1,13 @@
...
+  const startDate = new Date(date);
+  const dayOfWeek = startDate.getUTCDay();
+  const startOfWeek = new Date(startDate);
+  startOfWeek.setUTCDate(startDate.getUTCDate() - ((dayOfWeek + 6) % 7));
+  const endOfWeek = new Date(startOfWeek);
+  endOfWeek.setUTCDate(startOfWeek.getUTCDate() + 6);
+  const formatDate = (d: Date) => d.toISOString().split('T')[0];
+  return {
+    start: formatDate(startOfWeek),
+    end: formatDate(endOfWeek),
```
Matt561, line 11: Minor/DX: Could we place a comment above the `formatDate` function def to preview what a sample formatted date looks like?

Matt561 (commented):
Overall, this looks good! Just left a few questions and comments 👍

app/components/UI/Stake/Views/StakeEarningsHistoryView/__snapshots__/StakeEarningsHistoryView.test.tsx.snap
```diff
@@ -0,0 +1,389 @@
...
+                }
+              >
+                <View
+                  onLayout={[Function]}
+                  style={
+                    {
+                      "flex": 1,
+                    }
+                  }
+                />
```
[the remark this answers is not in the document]
nickewansmith: Hey Matthew, yes that message was very helpful indeed. In other snapshots it looks like the graph does render, I think it is missing width and height on the layout event for this test specifically. Good catch, will make that change.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
```diff
@@ -0,0 +1,334 @@
...
+  } = useStakingEarningsHistory({
+    chainId: asset.chainId as ChainId,
+    limitDays: EARNINGS_HISTORY_DAYS_LIMIT,
+  });
+
+  const ticker = asset.ticker ?? asset.symbol;
+
+  const formatRewardsWei = useCallback(
+    (rewards: number | string | BN, raw?: boolean) => {
+      if (!raw) {
```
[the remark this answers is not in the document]
nickewansmith: This could be clearer, it's formatting wei returning either a number string or a string representation that may contain special chars. In some cases the returned wei string cannot be parsed as a float due to some invalid char i.e. `< .00001`. When we pass raw = true here, it ensures that we don't worry about that and get the safe number string with as much precision as is available. We already do this, I just wrapped it to cover the use-case in the component.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
nickewansmith: I agree, the method bothers me and it is also not easy to understand. For the sake of time, I decided to leave it and get the review started, as I have attempted breaking it down a few times now and it wasn't very straightforward. I'll give it another shot, thanks for the input.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
```diff
@@ -0,0 +1,334 @@
...
+
+  const { earningsHistoryChartData, earningsHistoryListData } = useMemo(() => {
+    const historyData: {
+      earningsHistoryChartData: {
+        earnings: StakingEarningsHistoryChartData[];
+        earningsTotal: string;
+        ticker: string;
+      };
+      earningsHistoryListData: StakingEarningsHistoryListData[];
+    } = {
```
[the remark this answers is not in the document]
nickewansmith: Will update

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.tsx
nickewansmith: Currently to keep things simple, I did not attempt to account for the conversion of any currencies outside of USD. We can adapt in the future fairly easily, wdyt?

app/components/UI/Stake/utils/date/index.ts
nickewansmith: Sure

nickewansmith: Addressed

app/components/UI/Stake/Views/StakeEarningsHistoryView/__snapshots__/StakeEarningsHistoryView.test.tsx.snap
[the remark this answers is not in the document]
nickewansmith: Fixed!

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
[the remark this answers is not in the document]
nickewansmith: I updated the param name to be clearer here

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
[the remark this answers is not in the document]
nickewansmith: Added

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
nickewansmith: I broke this down a lot, it should be a bit more readable

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
nickewansmith: Updated to useEffect

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
amitabh94: Looks much better now!

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.tsx
amitabh94: Do you think this would be a heavy lift? It would be nice for users to see their rewards history in local currency.

amitabh94 (dismissed):
LGTM

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsTimePeriod/StakingEarningsTimePeriod.tsx
```diff
@@ -0,0 +1,93 @@
...
+  TextVariant,
+} from '../../../../../../../component-library/components/Texts/Text';
+import { useStyles } from '../../../../../../../component-library/hooks';
+import styleSheet from './StakingEarningsTimePeriod.styles';
+import SkeletonPlaceholder from 'react-native-skeleton-placeholder';
+
+export enum DateRange {
+  DAILY = '7D',
+  MONTHLY = 'M',
+  YEARLY = 'Y ',
```
Matt561, line 16: Probably a typo! The additional space moves the `Y` off-center
```suggestion
  YEARLY = 'Y',
```

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.styles.ts
```diff
@@ -0,0 +1,39 @@
+import { StyleSheet } from 'react-native';
+
+const styleSheet = () =>
+  StyleSheet.create({
+    stakingEarningsHistoryListContainer: {
+      flex: 1,
+      paddingTop: 24,
```
Matt561, line 7: Super minor: Adding `paddingBottom: 24` to this container would give a bit of space for longer lists. This _should_ make the last item a bit easier to read.

#### Before Bottom Padding
image

#### After Bottom Padding
image

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.tsx
```diff
@@ -0,0 +1,111 @@
...
+}: StakingEarningsHistoryListProps) => {
+  const { colors } = useTheme();
+  const styles = styleSheet();
+  let lastGroupHeader: string | null = null;
+  return (
+    <View style={styles.stakingEarningsHistoryListContainer}>
+      {earnings ? (
+        <>
+          <Label variant={TextVariant.BodyMDBold} color={colors.text.default}>
+            Payout history
```
Matt561, line 36: Might be best to replace hardcoded strings with translation variables. Apologies for not catching this on the last review!

```suggestion
            {strings("stake.payout_history")}
```

Matt561 (commented):
I was able to test this locally successfully. Just a few minor comments but otherwise this LGTM! 👍

Matt561 (dismissed):

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
```diff
@@ -0,0 +1,428 @@
...
+import { useSelector } from 'react-redux';
+import {
+  selectCurrencyRates,
+  selectCurrentCurrency,
+} from '../../../../../../selectors/currencyRateController';
+import { selectNetworkConfigurations } from '../../../../../../selectors/networkController';
+import { Hex } from '../../../../../../util/smart-transactions/smart-publish-hook';
+import { NetworkConfiguration } from '@metamask/network-controller';
+
+interface StakingEarningsHistoryProps {
```
Cal-L, line 40: Can reduce lines of code by abstracting types into types file, constants into constants file, and helpers into utils files

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
```diff
@@ -0,0 +1,428 @@
...
+  const [selectedTimePeriod, setSelectedTimePeriod] = useState<DateRange>(
+    EARNINGS_HISTORY_TIME_PERIOD_DEFAULT,
+  );
+  const currentCurrency: string = useSelector(selectCurrentCurrency);
+  const multiChainMarketData = useSelector(selectTokenMarketData);
+  const multiChainCurrencyRates = useSelector(selectCurrencyRates);
+  const networkConfigurations: Record<Hex, NetworkConfiguration> = useSelector(
+    selectNetworkConfigurations,
+  );
+  const nativeCurrency =
```
Cal-L, line 134: Is it possible for these optional chained values to be undefined? If they are, is there a better way to gracefully/intentionally handle those scenarios?

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
```diff
@@ -0,0 +1,428 @@
...
+      }
+      // return a string without special characters
+      return asset.isETH
+        ? fromWei(rewards)
+        : fromTokenMinimalUnit(rewards, asset.decimals);
+    },
+    [asset.isETH, asset.decimals],
+  );
+
+  const formatRewardsNumber = useCallback(
```
Cal-L, line 165: The helpers here tends to clutter the component and makes testing harder as well. Let's abstract out the utils and write unit tests to ensure they're functioning properly

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+  onBandWidthChange,
+  strokeColor,
+}: HorizontalLinesProps) => {
+  useEffect(() => {
+    onBandWidthChange && onBandWidthChange(bandWidth ?? 0);
+  }, [bandWidth, onBandWidthChange]);
+  if (!x || !y || !height || !data || !bandWidth) return null;
+  return (
+    <>
+      {data.map((item, index) => (
```
Cal-L, line 55: Since it's mapping in render, let's wrap this in a useCallback

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+export function StakingEarningsHistoryChart({
+  earnings,
+  ticker,
+  earningsTotal,
+  formatValue = (value) => value.toFixed(5),
+  onSelectedEarning,
+}: StakingEarningsHistoryChartProps): React.ReactElement {
+  //hooks
+  const { colors } = useTheme();
+  const { styles } = useStyles(styleSheet, {});
```
Cal-L, line 80: useStyles also returns a theme object where you can access colors. No need for a separate `useTheme`

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+  // constants
+  const animate = false;
+  const stopColorGreen = 'rgb(228, 240, 231)';
```
Cal-L, line 84: Is this a color that we need to add to design system? If so, I'd recommend reaching out to @MetaMask/design-system

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
```diff
@@ -0,0 +1,292 @@
...
+        setSelectedBarIndex(-1);
+        setLastToggledBarIndex(-1);
+      }
+    }
+  };
+  const handleTouch = (evt: GestureResponderEvent) => {
+    updateBarHoveredBarIndex(evt.nativeEvent.locationX);
+  };
+
+  // update bar fill color on index change
```
Cal-L, line 151: This is probably too big of a change at this point. In the future, you can probably utilize react-native-reanimated for more performant animations. It seems like it might help remove some of these useEffects as well.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const styles = styleSheet();
+  let lastGroupHeader: string | null = null;
+  return (
+    <View style={styles.stakingEarningsHistoryListContainer}>
+      {earnings ? (
+        <>
+          <Label variant={TextVariant.BodyMDBold} color={colors.text.default}>
+            {strings('stake.earnings_history_list_title')}
+          </Label>
+          {earnings.map((earning, index) => {
```
Cal-L, line 39: Wrap in useCallback

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryList/StakingEarningsHistoryList.tsx
```diff
@@ -0,0 +1,112 @@
...
+                  </View>
+                </View>
+              );
+            }
+            return null;
+          })}
+        </>
+      ) : (
+        <SkeletonPlaceholder>
+          {Array.from({ length: 7 }).map((_, index) => (
```
Cal-L, line 97: Wrap in useCallback

app/util/testUtils/react-native-svg-charts.ts
Cal-L, line 1: What's this file for?

Cal-L (commented):
Left some comments

app/components/UI/Stake/Views/StakeEarningsHistoryView/StakeEarningsHistoryView.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { getStakingNavbar } from '../../../Navbar';
+import StakingEarningsHistory from '../../components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory';
+import styleSheet from './StakeEarningsHistoryView.styles';
+import { StakeEarningsHistoryViewProps } from './StakeEarningsHistoryView.types';
+
+const StakeEarningsHistoryView = ({ route }: StakeEarningsHistoryViewProps) => {
+  const navigation = useNavigation();
+  const { styles, theme } = useStyles(styleSheet, {});
+  const { asset } = route.params;
+  const ticker = asset.ticker ?? asset.symbol;
```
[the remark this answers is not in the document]
nickewansmith: During development there was a change to the TokenI interface adding asset.ticker. The interface has the prop as possibly undefined, so in the case it is undefined, I fallback to symbol.

app/util/testUtils/react-native-svg-charts.ts
nickewansmith: This is a test utility from the linked GitHub discussion in the file. It is used to render the chart during testing. It is added here as there is another chart that will end up using the same helper fairly soon.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
nickewansmith: Ok, I'll reach out. The color is a pretty one off color that was needed for the graph gradient. It's possible we may not need this color frequently

georgewrmarshall: Great catch, @Cal-L! Hey @nickewansmith, as a general rule, we should avoid using static color values in the platform to prevent inconsistencies and potential UI issues, especially if theming hasn’t been considered or if a new theme is introduced in the future. When static, one-off colors are necessary, we recommend keeping them all in a central location like [[app/styles/common.ts](https://github.com/MetaMask/metamask-mobile/blob/75b5fd61ff62659244f396f0eea48c1061f29810/app/styles/common.ts)](https://github.com/MetaMask/metamask-mobile/blob/75b5fd61ff62659244f396f0eea48c1061f29810/app/styles/common.ts) to make them easier to manage and reuse.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
Matt561: > Since it's mapping in render, let's wrap this in a useCallback

I'm interested in learning more about this. Generally speaking, do you recommend wrapping render maps in a useCallback? If yes, why?

app/components/UI/Stake/Views/StakeEarningsHistoryView/StakeEarningsHistoryView.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { View } from 'react-native';
+import { ScrollView } from 'react-native-gesture-handler';
+import { strings } from '../../../../../../locales/i18n';
+import { useStyles } from '../../../../hooks/useStyles';
+import { getStakingNavbar } from '../../../Navbar';
+import StakingEarningsHistory from '../../components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory';
+import styleSheet from './StakeEarningsHistoryView.styles';
+import { StakeEarningsHistoryViewProps } from './StakeEarningsHistoryView.types';
+
+const StakeEarningsHistoryView = ({ route }: StakeEarningsHistoryViewProps) => {
```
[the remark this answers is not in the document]
Matt561: I don't have a strong opinion either way _but_ placing route in the component's prop type would make it more explicit that this component needs `route`. If we remove `route` from the props it'll be more implicit.

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistory.tsx
nickewansmith: This is currently set to string in the TokenI interface, it probably should be Hex, but I don't want to make any assumptions there. I have followed the usage of `asset.chainId` in other parts of the codebase for now.

app/components/UI/Stake/Views/StakeEarningsHistoryView/StakeEarningsHistoryView.tsx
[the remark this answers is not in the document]
Cal-L: What's a real world example of the ticker being present?

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.tsx
Cal-L: Yep, it's usually recommended to always wrap render maps since uncontrolled re-renders will always lead to the mapping function running again. By default, react components aren't too optimized so more often than not, we will experience frequent re-renders. Memoization can lead to performance issues if abused but that's usually if a project wraps memoizes things even when it doesn't need to be. It isn't often that you run into that though since most projects aren't overly optimized.

app/components/UI/Stake/Views/StakeEarningsHistoryView/StakeEarningsHistoryView.tsx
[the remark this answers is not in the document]
nickewansmith: I don't know of one, the sole reason it's added is because of the possibly undefined ticker definition. I can instead assume it will be defined and remove the alternative value as it's not a must have.

Cal-L (approved):
LGTM

Matt561 (approved):

amitabh94 (approved):

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+      chartContainer.getByTestId('earnings-history-chart'),
+      'onTouchStart',
+      {
+        nativeEvent: { locationX: 50 },
+      },
+    );
+    // expect bar 1 to be selected and highlighted on touch
+    expect(chartContainer.getByText('Day 1')).toBeTruthy();
+    expect(chartContainer.getByText('1.00000 ETH')).toBeTruthy();
+    expect(chart.data[0].svg.fill).toBe('#1c8234');
```
georgewrmarshall, line 88: _non-blocking_: I can't quite pinpoint where this hex value is coming from, but it might be more robust to use the design token instead of the static hex value—or to avoid checking for the color value altogether. If we upgrade the design tokens with different color values, this test is likely to fail.  

```ts
import { lightTheme } from '@metamask/design-tokens';

expect(chart.data[0].svg.fill).toBe(lightTheme.colors.success.default);
```

georgewrmarshall (approved):
LGTM! One non-blocking suggestion

app/components/UI/Stake/components/StakingEarnings/StakingEarningsHistory/StakingEarningsHistoryChart/StakingEarningsHistoryChart.test.tsx
nickewansmith: Ah I missed this, I will update this in another PR. The actual code is not using the hex value, but the test would fail if the design system updated that token. Thx George!

salimtb (approved):
Assets changes LGTM

PR #12765, opened 2024-12-18
fix: Prevent Engine.destroy test intermittent failure

mcmire:
## **Description**

Some unit tests for the Engine do not properly wait for the NetworkController provider to be fully initialized after initializing the engine before attempting to destroy it. This prevents NetworkController's `lookupNetwork` from functioning because it removes an event listener that NetworkController had previously created, causing an error.

This commit resolves the issue by patching NetworkController to not throw if the controller is destroyed before `lookupNetwork` has a chance to remove the event listener.

## **Related issues**

This issue was found recently when making another, unrelated update.

## **Manual testing steps**

All unit tests should pass, particularly the test in `Engine.test.ts` called "calling Engine.destroy deletes the old instance".

### DIFF
--- a/app/core/Engine/Engine.ts
+++ b/app/core/Engine/Engine.ts
@@ -268,6 +268,8 @@ export class Engine {
 
   keyringController: KeyringController;
 
+  #promiseForNetworkProviderInitialization: Promise<void>;
+
   /**
    * Creates a CoreController instance
    */
@@ -333,6 +335,21 @@ export class Engine {
     };
     const networkController = new NetworkController(networkControllerOpts);
 
+    this.#promiseForNetworkProviderInitialization = new Promise<void>(
+      (resolve) => {
+        this.controllerMessenger.subscribe(
+          'NetworkController:stateChange',
+          (networksMetadata: NetworkState['networksMetadata']) => {
+            // `networksMetadata` starts out empty and is then updated for the
+            // currently selected network, which is Mainnet by default
+            if ('mainnet' in networksMetadata) {
+              resolve();
+            }
+          },
+          (networkControllerState) => networkControllerState.networksMetadata,
+        );
+      },
+    );
     networkController.initializeProvider();
 
     const assetsContractController = new AssetsContractController({
@@ -1948,6 +1965,7 @@ export class Engine {
   }
 
   async destroyEngineInstance() {
+    await this.#promiseForNetworkProviderInitialization;
     // TODO: Replace "any" with type
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     Object.values(this.context).forEach((controller: any) => {
--- a/app/core/Engine/Engine.ts
+++ b/app/core/Engine/Engine.ts
@@ -268,8 +268,6 @@ export class Engine {
 
   keyringController: KeyringController;
 
-  #promiseForNetworkProviderInitialization: Promise<void>;
-
   /**
    * Creates a CoreController instance
    */
@@ -335,21 +333,6 @@ export class Engine {
     };
     const networkController = new NetworkController(networkControllerOpts);
 
-    this.#promiseForNetworkProviderInitialization = new Promise<void>(
-      (resolve) => {
-        this.controllerMessenger.subscribe(
-          'NetworkController:stateChange',
-          (networksMetadata: NetworkState['networksMetadata']) => {
-            // `networksMetadata` starts out empty and is then updated for the
-            // currently selected network, which is Mainnet by default
-            if ('mainnet' in networksMetadata) {
-              resolve();
-            }
-          },
-          (networkControllerState) => networkControllerState.networksMetadata,
-        );
-      },
-    );
     networkController.initializeProvider();
 
     const assetsContractController = new AssetsContractController({
@@ -1965,7 +1948,6 @@ export class Engine {
   }
 
   async destroyEngineInstance() {
-    await this.#promiseForNetworkProviderInitialization;
     // TODO: Replace "any" with type
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     Object.values(this.context).forEach((controller: any) => {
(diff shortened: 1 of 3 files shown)


patches/@metamask+network-controller+22.1.0.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/node_modules/@metamask/network-controller/dist/NetworkController.cjs b/node_modules/@metamask/network-controller/dist/NetworkController.cjs
+index cc9793f..b94112a 100644
+--- a/node_modules/@metamask/network-controller/dist/NetworkController.cjs
++++ b/node_modules/@metamask/network-controller/dist/NetworkController.cjs
+@@ -525,7 +525,11 @@ class NetworkController extends base_controller_1.BaseController {
+         let networkChanged = false;
+         const listener = () => {
+             networkChanged = true;
+-            this.messagingSystem.unsubscribe('NetworkController:networkDidChange', listener);
```
mcmire, line 9: I plan on updating `@metamask/network-controller` with this change in another PR so we don't need this patch long-term.

patches/@metamask+network-controller+22.1.0.patch
```diff
@@ -0,0 +1,30 @@
...
++            try {
++                this.messagingSystem.unsubscribe('NetworkController:networkDidChange', listener);
++            } catch {
++                // The controller may have been destroyed, so don't worry about it
```
Gudahtt, line 13: What error is expected here? An empty catch block with silently suppress any errors.

app/core/Engine/Engine.ts
```diff
@@ -1948,6 +1965,7 @@ export class Engine {
   }
 
   async destroyEngineInstance() {
+    await this.#promiseForNetworkProviderInitialization;
```
Gudahtt, line 1968: Why would we want to wait on the network provider to initialize here? This could significantly degrade performance, especially considering that we don't have a timeout for initialization.

mcmire: I was thinking this was the more "correct" fix. We only seem to call `destroyEngineInstance` in tests so I didn't think it would impact performance. However, I guess we don't need this if we have the NetworkController patch (the unit tests still pass). I can remove this.

patches/@metamask+network-controller+22.1.0.patch
mcmire: Sure, good point. I've added a check for the error we want to ignore here: [52f4433](https://github.com/MetaMask/metamask-mobile/pull/12765/commits/52f4433e779eb48ed3cf75bb8f2e95723cb03b61)

Gudahtt:
>This prevents NetworkController's lookupNetwork from functioning because it removes an event listener that NetworkController had previously created, causing an error.

Why does it matter if `lookupNetwork` fails if the engine was destroyed? We don't need it to succeed in that case.

app/core/Engine/Engine.ts
Gudahtt: We also destroy and re-create the Engine during the vault recovery flow

mcmire: Ah, I missed searching for `Engine.destroyEngine`, thanks.

mcmire:
> Why does it matter if `lookupNetwork` fails if the engine was destroyed? We don't need it to succeed in that case.

Fair enough, and that was the conclusion I eventually came to, hence the NetworkController patch.

app/core/Engine/Engine.ts
mcmire: Removed here: [aa8b070](https://github.com/MetaMask/metamask-mobile/pull/12765/commits/aa8b0706a4f21913e8010ed3e5c7f41cae994f10)

patches/@metamask+network-controller+22.1.0.patch
Gudahtt: Could we handle this upstream instead, and fix it here by updating? It looks like we're on the latest major version, so it should be relatively easy to do

Gudahtt:
>Fair enough, and that was the conclusion I eventually came to, hence the NetworkController patch.

What I meant was "why is this resulting in an error?". That is not clear from looking at the patch.

Though even in terms of just the patch changes, it seems that it's intended to ensure the operation succeeds.

patches/@metamask+network-controller+22.1.0.patch
mcmire: That would be better for sure. I don't have a lot of time today and just wanted to get this person unblocked for the time being or at least give them a solution they can use to complete their draft PR, but I can make a change upstream later today or possibly tomorrow.

mcmire:
@Gudahtt It's resulting in an error because the messenger event that it's trying to unsubscribe has been removed (because the controller has been destroyed).

There is no real reason to continue if the controller is destroyed, you are right. We should probably just return instead.

Gudahtt:
I see, so it's causing the test to fail because it's an unhandled promise rejection? I was having trouble following how this caused the test failure. But it looks like [Jest doesn't identify unhandled Promise rejections as such](https://github.com/jestjs/jest/issues/9210), so that could explain it.

patches/@metamask+network-controller+22.1.0.patch
mathieuartu: @mcmire , don't stress yourself over this! While I really appreciate the effort to unblock me (🫶), this was only a draft PR to test a core change that I'm planning to merge later this week. I'd rather go with the classic update flow instead of adding a patch as well, so let's take our time to do this :)

mcmire: I've submitted a PR to fix this in NetworkController here: https://github.com/MetaMask/core/pull/5116

mcmire:
Considering https://github.com/MetaMask/core/pull/5116 was merged a while back, I am closing this.

PR #12811, opened 2024-12-20
feat: add detected tokens automatically

sahar-fehri:
## **Description**

Now that we have most users autodetect tokens, we can make the UX even more streamlined. This PR automatically add in tokens we detect on behalf of users.

## **Related issues**

Fixes:https://consensyssoftware.atlassian.net/browse/MMASSETS-439

## **Manual testing steps**

1. Switch to an account that has tokens
2. Notice you no longer see the link telling you that you have tokens detected
3. Detected tokens are added automatically to your state
4. Disable token detection setting
5. Switch to an account that has tokens
6. Detected tokens are not added to state

### **After**

https://github.com/user-attachments/assets/3cbf7a84-fdcb-4baf-bb86-d8efe52591ce


salimtb (dismissed):
LGTM

app/components/Views/Wallet/index.tsx
```diff
@@ -444,6 +476,74 @@ const Wallet = ({
...
+              })
+              .build(),
+          ),
+        );
+      }
+    };
+    importAllDetectedTokens();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    selectedAddress,
```
tommasini, line 539: Why we have the selectedAddress as a dependency? It's weird to me that this is not handled in the controller level and instead, we cause an extra re-rendering, maybe I'm missing something, could you explain?

sahar-fehri: We have plans to make this on controller level in future; it might require a larger change.
The selectedAddress is a dependency because we want to redetect tokens and add them automatically to state if the user adds a new address.

tommasini: Okok I was just not sure if this was a bug or was never implemented, I see the TokensController does listen the change of the account [here](https://github.com/MetaMask/core/blob/20943801f54dc70a1854b3d199ac0f031d97956c/packages/assets-controllers/src/TokensController.ts#L321), but it do not perform the detect tokens!
I wonder if it's that large of a change, it seems we just need to establish the conditions to addDetectedTokens or not at that `onSelectedAccountChange` function!

Can we leave a comment here saying why selectedAddress is needed and what is planned to remove it? 

I can see the possibility of this creating unneeded re-renders, since we are adding detected tokens each time the selectedAddress changes, the feature itself is very good, but most likely is creating a bigger performance issue when switching accounts, since this will iterate this [function](https://github.com/MetaMask/core/blob/1e23e90866949567c8cc7c0880a51e1d3bd76c0d/packages/assets-controllers/src/TokenDetectionController.ts#L971) and this function is one of our sources of heavy computation for devices at the smart contract call [here](https://github.com/MetaMask/core/blob/f5ad522451002eb2b56fdac78d2b11ad26ac471a/packages/assets-controllers/src/AssetsContractController.ts#L691).

tommasini (approved):

salimtb (approved):

PR #12813, opened 2024-12-20
feat: add Network URL non-ascii -> punycode warning

digiwand:
## **Description**

Similar to the extension, we want to warn the user if the URL has non-ASCII characters. Inline alerts are not supported on mobile, so we display the new alert as a banner.

Notes: 
- existing Network URL logic removes the path with hideKeyFromUrl
- punycode is deprecated, however, the url library does not seem to parse the url into its punycode encoded version in react-native. It is tricky as the url does parse into its punycode version in node.js, jest tests

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/2365
Related to: https://github.com/MetaMask/metamask-extension/pull/29490 (fixes isValidASCIIURL to include path check in extension) 

## **Manual testing steps**

Test switching to a custom network


app/util/url/index.ts
```diff
@@ -24,6 +26,35 @@ export function isBridgeUrl(url: string) {
...
 }
 
+export const isValidASCIIURL = (urlString?: string) => {
+  try {
+    return urlString?.includes(punycode.toASCII(new URL(urlString).host));
+  } catch (exp: unknown) {
+    console.error(exp);
+    return false;
+  }
+};
```
jpuri, line 36: @digiwand : why is the check for valid ASCII includes only the host part of the URL ?

digiwand: hey @jpuri, good question. I copied this method verbatim from the metamask-extension code. I think you created the one in the metamask-extension. 

will double-check this method with consideration of @NicholasEllul's comment above https://github.com/MetaMask/metamask-mobile/pull/12813/files#r1905686031

app/util/url/index.ts
```diff
@@ -24,6 +26,35 @@ export function isBridgeUrl(url: string) {
   }
 }
 
+export const isValidASCIIURL = (urlString?: string) => {
+  try {
+    return urlString?.includes(punycode.toASCII(new URL(urlString).host));
```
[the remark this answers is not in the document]
digiwand: hey @NicholasEllul and @jpuri, thanks for flagging this! This helper method was copied over from the extension. I updated the extension code and the code in this PR to include the path in its check. can I get another look at this please? 

related extension PR: https://github.com/MetaMask/metamask-extension/pull/29490

app/util/url/index.ts
digiwand: please see https://github.com/MetaMask/metamask-mobile/pull/12813#discussion_r1906162676

jpuri (dismissed):
Nice work

app/util/url/index.ts
```diff
@@ -24,6 +26,38 @@ export function isBridgeUrl(url: string) {
...
 }
 
+export const isValidASCIIURL = (urlString?: string) => {
+  try {
+    if (!urlString) { return false; }
+
+    const urlPunycodeString = punycode.toASCII(new URL(urlString).href);
+    return urlPunycodeString?.includes(urlString);
+  } catch (exp: unknown) {
+    console.error(exp);
```
matthewwalsh0, line 36: Minor, does this warrant an error log if it's handled?

And if we still want a log, should we add some context for clarity such as:

`console.log('URL contains non-ASCII characters', urlString, urlPunycodeString)`

app/util/url/index.ts
```diff
@@ -24,6 +26,38 @@ export function isBridgeUrl(url: string) {
...
+ */
+export const toPunycodeURL = (urlString: string) => {
+  try {
+    const url = new URL(urlString);
+    const punycodeHostname = punycode.toASCII(url.hostname);
+    const { protocol, port, search, hash } = url;
+    const pathname =
+      url.pathname === '/' && !urlString.endsWith('/') ? '' : url.pathname;
+
+    return `${protocol}//${punycodeHostname}${port}${pathname}${search}${hash}`;
```
matthewwalsh0, line 58: Why do we need to build this manually and only convert the hostname to ASCII?

Could we pass the entire `href` in one go as we do in `isValidASCIIURL`?

matthewwalsh0 (changes requested):

NEllusion (commented):
I notice the following notice in the punycode README: https://github.com/mathiasbynens/punycode.js?tab=readme-ov-file#installation

> ⚠️ Note that userland modules don't hide core modules. For example, require('punycode') still imports the deprecated core module even if you executed npm install punycode. Use require('punycode/') to import userland modules rather than core modules.


We should double check to ensure we are using the punycode built into node which is soft-deprecated. I notice other files like this one import punycode as the following: https://github.com/MetaMask/metamask-mobile/blob/e0864ad83afcaf74c535efa4006e773564e31f1c/app/util/address/index.ts#L9

app/util/url/index.ts
digiwand: oh I originally proposed href then took it out to handle the pathname === '/' logic. Taking a second look at this we can keep the href and reword the logic. Thanks for the callout! updated [4c48863](https://github.com/MetaMask/metamask-mobile/pull/12813/commits/4c4886389c3a8bc8ada30f48b6baa74e621f6682)

app/util/url/index.ts
```diff
@@ -28,7 +28,10 @@ export function isBridgeUrl(url: string) {
 
 export const isValidASCIIURL = (urlString?: string) => {
   try {
-    return urlString?.includes(punycode.toASCII(new URL(urlString).host));
+    if (!urlString) { return false; }
+
+    const urlPunycodeString = punycode.toASCII(new URL(urlString).href);
+    return urlPunycodeString?.includes(urlString);
```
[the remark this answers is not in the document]
digiwand: @NicholasEllul

yeah, the react-native runtime URL library didn't seem to support the punycode encoding

---

> E.g if the url string in this case is https://opensea.io/language=français, the toASCII function will incorrectly turn it into https://example.xn--com/lang=franais-opb.

This is the reason why I was told we are displaying the given input rather than the punycode version. We show the punycode different in the warning. If only we could mimic the behavior of the browser search bar. An attempt at this would be out-of-scope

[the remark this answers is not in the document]
digiwand: @NicholasEllul I like this idea. Will update. When we do render the punycode version in the warning, we will continue to do so on the entire URL, including the path.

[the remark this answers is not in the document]
digiwand: @NicholasEllul 

ran into a test blocker. Because URL does apply punycode encoding in the node.js tests, the checks don't test what is represented in react-native and it fails to detect the differences. We want to check the original hostname of urlString. We could add a regex to check this non-punycode encoded hostname 🤔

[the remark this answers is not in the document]
digiwand: updated using a regex @NicholasEllul [c880f36](https://github.com/MetaMask/metamask-mobile/pull/12813/commits/c880f3614c3276557f899d33c4c635f6d0e67b52). Thanks for the suggestion!

NEllusion (approved):
Nice workaround 👍

matthewwalsh0 (approved):

PR #12822, opened 2024-12-20
chore: Improve Keyring/Accounts error handling and logs

owencraston:
## **Description**
This PR does the following

1. unmasks non sensitive data from the Accounts Controller and Keyring controller to help us better debug issues in prod. This means that more data will be provided in the sentry state logs. I have ensured that all addresses have remained masked
2. adds logs when creating key components such as the engine and the accounts controller. This will help us debug issues with empty state.
3. adds validation to the migrations for key components. Currently we are validating that the accounts controller and keyring controller data are sound. This validation will not block the app from initializing but will log the errors it finds.

## **Related issues**

Progresses : https://github.com/MetaMask/metamask-mobile/issues/12408

## **Manual testing steps**

1. this change only adds logs and error handling to the app so there should be no functional changes. that being said, we should still verify that the upgrade path is working.


app/core/Engine/controllers/AccountsController/utils.ts
```diff
@@ -28,6 +28,19 @@ export const createAccountsController = ({
   initialState?: AccountsControllerState;
 }): AccountsController => {
   try {
+    if (!initialState) {
+      Logger.log('Creating AccountsController with default state', {
+        defaultState: defaultAccountsControllerState,
+      });
+    } else {
+      Logger.log('Creating AccountsController with provided initial state', {
+        hasSelectedAccount: !!initialState.internalAccounts?.selectedAccount,
```
sethkfman, line 37: I would suggest optional chaining on all of the logs. If any of these are undefined we shouldn't crash init.

app/store/validateState.ts
```diff
@@ -0,0 +1,111 @@
...
+ * - Returns an array of error strings if any validations fail or an empty array otherwise.
+ */
+type ValidationCheck = (state: RootState) => string[];
+
+const LOG_TAG = 'MIGRATION_VALIDATE_STATE_ERROR';
+
+/**
+ * Verifies that the engine is initialized
+ */
+const checkEngineInitialized: ValidationCheck = (state) => {
```
sethkfman, line 17: Will we crash if the state is undefined/null? We don't want to take down the app if the state is empty.

app/core/Engine/Engine.ts
```diff
@@ -276,6 +277,8 @@ export class Engine {
     initialState: Partial<EngineState> = {},
     initialKeyringState?: KeyringControllerState | null,
   ) {
+    logEngineCreation(initialState, initialKeyringState);
```
owencraston, line 280: logs the state we pass into the engine.

app/core/Engine/Engine.ts
```diff
@@ -2158,7 +2165,10 @@ export default {
     instance = null;
   },
 
-  init(state: Partial<EngineState> | undefined, keyringState = null) {
+  init(
```
owencraston, line 2168: improves the typing of this function.

app/core/Engine/controllers/AccountsController/logger.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function logAccountsControllerCreation(
+  initialState?: AccountsControllerState,
+) {
+  if (!initialState) {
+    Logger.log('Creating AccountsController with default state', {
+      defaultState: defaultAccountsControllerState,
+    });
+  } else {
+    Logger.log('Creating AccountsController with provided initial state', {
```
owencraston, line 13: @sethkfman will using Logger.log ensure these logs show up in sentry?

app/core/Engine/utils/logger.ts
```diff
@@ -0,0 +1,20 @@
+import { KeyringControllerState } from '@metamask/keyring-controller';
+import { EngineState } from '../types';
+import Logger from '../../../util/Logger';
+
+export function logEngineCreation(
```
owencraston, line 5: this is what I propose we log when the engine is instantiated. Thoughts? Will Logger.log show up in sentry if its not an error?

app/core/EngineService/EngineService.ts
```diff
@@ -235,17 +239,20 @@ export class EngineService {
     const keyringState = await getVaultFromBackup();
     const reduxState = ReduxService.store.getState();
     const state = reduxState?.engine?.backgroundState || {};
-    // TODO: Replace "any" with type
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const Engine = UntypedEngine as any;
```
owencraston, line 240: the any typing is no longer needed since the Engine has been fully typed for a while now.

app/store/validateState.ts
owencraston: I just modified the code so that even when rootState.engine is empty we do not throw an error. Each validation step has a specific test to verify that it never throws.

app/store/validateMigration/accountsController.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  });
+
+  it('handles undefined backgroundState', () => {
+    const errors = validateAccountsController({ engine: {} } as RootState);
+    expect(errors).toEqual([
+      `${LOG_TAG}: AccountsController state is missing in engine backgroundState.`,
+    ]);
+  });
+
+  it('does not throw with malformed state', () => {
```
owencraston, line 155: @sethkfman this test helps us verify that these validation steps will not throw but will collect the logs for us.

app/store/validateMigration/validateMigration.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function validatePostMigrationState(state: RootState): void {
+  const allErrors: string[] = [];
+
+  for (const check of checks) {
+    const errors = check(state);
+    if (errors.length > 0) {
+      allErrors.push(...errors);
+    }
+  }
```
gantunesr, line 28: You can simplify it like this

```suggestion
  const allErrors = checks.flatMap(check => check(state));
```

app/store/validateMigration/validateMigration.test.ts
```diff
@@ -0,0 +1,87 @@
+import { validatePostMigrationState } from './validateMigration';
+import { RootState } from '../../reducers';
+import Logger from '../../util/Logger';
+import { validateAccountsController } from './accountsController';
+import { validateKeyringController } from './keyringController';
+import { validateEngineInitialized } from './engineBackgroundState';
```
gantunesr, line 6: The note I have about these validations is that they are really specific to certain migrations. These validations won't be true for future migration if the state of any of these components changes.

app/core/Engine/Engine.ts
mikesposito, line 280: Wouldn't this log the initial KeyringController state and thus all serialized data that should be passed to each keyring? (i.e. mnemonic and private keys)

mikesposito: Ah, sorry, nevermind. `logEngineCreation` does not log the actual state, but just booleans

app/core/Engine/controllers/AccountsController/logger.ts
Cal-L: They appear as breadcrumbs, which are attached to traces if an error occurs

Cal-L (commented):
@owencraston Could you verify the upgrade scenario that you mentioned in the description

`this change only adds logs and error handling to the app so there should be no functional changes. that being said, we should still verify that the upgrade path is working.`

app/core/Engine/Engine.ts
owencraston: yes exactly. it is only used to determine if we are using a backup or not.

app/core/Engine/controllers/AccountsController/logger.ts
owencraston: @Cal-L if we want to guarantee that a log gets sent to sentry should we use log.error? or is that a misuse of this pattern?

app/store/validateMigration/validateMigration.ts
owencraston: good call.

owencraston:
I investigated the sonar cloud quality check failures. The suggestions seem weird to me since modifying the code to their liking results in a lint error. Not sure what to do here cc @sethkfman .

owencraston:
[QA builds](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/e79df332-2e70-4d7c-8015-cbde9b45244f?tab=workflows)

app/core/Engine/utils/logger.ts
owencraston: I do see your point here but I wonder if we will be missing useful data if we only log on these cases.

owencraston:
[Full e2e regression pipeline.](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/6d62d5a6-8e23-4877-bacb-decb3684cf0e?tab=workflows)

app/core/Engine/controllers/AccountsController/logger.ts
sethkfman: @owencraston Yes, log.error is what you would want to make sure it is logged. The breadcrumbs are helpful if you log and error and wanted to see what occurred before that error.

owencraston:
Prod test build [link](https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/d5b37984-8e36-45c2-963a-c8ff1648d522?tab=workflows). I created a [new branch](https://github.com/MetaMask/metamask-mobile/pull/12869) for the test build that has the version numbers bumped but the code is the same as this branch.

owencraston:
Tested the upgrade path from version 7.35.0 (since this seems to be a problematic version) to 7.38.1 which contains the code from this PR and everything worked as expected. I had a mix EOA and imported accounts with funds and all the data was accurate post upgrade.

https://github.com/user-attachments/assets/29ee9f0d-0a8b-4abc-bf28-d51216a2af48

app/core/Engine/controllers/AccountsController/logger.ts
owencraston: After some consideration I don't think we want to report these as errors since it all add extra burden on the platform team to triage. Instead these will be added as breadcrumbs to crashes which is really what we care about.

Cal-L (approved):
Thanks for addressing comments. LGTM from platform

sethkfman (approved):
LGTM

Cal-L (approved):
LGTM

gantunesr (approved):

PR #12851, opened 2025-01-07
fix: add nativeAsset key to staked ETH asset multichain

### LINKED ISSUE #12680: [Bug]: The app unexpectedly closes when users are navigating between tokens on various networks, Error: Missing chain id
### Describe the bug

The app unexpectedly closes with the` Error: Missing chain id` when users are navigating between tokens on various networks.

### Expected behavior

-

### Screenshots/Recordings

Image

https://github.com/user-attachments/assets/9e2148ec-f575-44e2-b353-436c136d711d

https://github.com/user-attachments/assets/00c4ac81-1b6a-4450-816f-5fdcfbfc107d

### Steps to reproduce

1. Switch to 'All networks' view
2. Select token an different networks

```shell

```

### Detection stage

During release testing

### Version

v7.38.0(1514)

### Build type

None

### Device

iPhone 15

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

### LINKED ISSUE #12856: RangeError: j6.l: Malformed currency code !
Sentry Issue: [METAMASK-MOBILE-39FF](https://metamask.sentry.io/issues/6203715909/?referrer=github_integration)

```
RangeError: j6.l: Malformed currency code !
  at anonymous (app/selectors/multichain.ts:133:44)
  at Tokens (app/components/UI/Tokens/index.tsx:125:50)
  at ?anon_0_ (app/components/Nav/App/index.js:637:26)
...
(45 additional frame(s) were not displayed)
```

nickewansmith:
## **Description**

There was a crash when clicking on the Staked Ethereum asset due to it missing a key `nativeAsset` which is required in some component. This was a regression from code refactoring.

Separately, there was a currency switching bug that was introduced with the use of a different fiat conversion method. I have switched to a method that does not throw error for currency switching and this has been tested with Vince

Separately, there were some issues with the Token list asset balances being slightly off. This was noticed when the balance we use for the Asset Detail staked Ethereum asset showed up differently than in the Token List. Some of this has been addressed, but there are still issues that I've noticed where we lose precision when getting the fiat total of a token asset or native asset. I left a comment in the code about this, and maybe it could be revisited it later.

We added the `nativeAsset` key back to the staked ethereum asset

We added the weiToFiat method to replace the Intl.NumberFormat method 

We removed precision loss from multichain.ts asset fiatBalance calculations, however there is another place(s) where fiat balances are recalculated again i.e. [deriveBalanceFromAssetMarketDetails.ts](https://github.com/MetaMask/metamask-mobile/pull/12851/files#diff-283c109c8093bbb5d8e06be79bcbeac928b8353633e55dc7944594a0e6888b49)

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/12680
Fixes: https://github.com/MetaMask/metamask-mobile/issues/12856
Closes: https://consensyssoftware.atlassian.net/browse/STAKE-912

## **Manual testing steps**

- Import the staking test wallet in MM
- Click on the staked ethereum asset
- The link should not crash the application more info here on testing this https://github.com/MetaMask/metamask-mobile/issues/12680
- Currency switching should not crash the application more info here on testing this  https://github.com/MetaMask/metamask-mobile/issues/12856

### **After**

https://github.com/user-attachments/assets/c2525b70-8d04-4606-861b-3206ebd557f4

https://github.com/user-attachments/assets/a8c4010e-3589-4171-803a-4d07a962e6a6


app/components/UI/Tokens/util/deriveBalanceFromAssetMarketDetails.ts
```diff
@@ -71,5 +85,8 @@ export const deriveBalanceFromAssetMarketDetails = (
       ? addCurrencySymbol(balanceFiatCalculation, currentCurrency)
       : `< ${addCurrencySymbol('0.01', currentCurrency)}`;
 
+  // NOTE: the way we calculate the balances here loses precision for both tokens and native assets
```
amitabh94, line 88: How are we calculating in Portfolio and is the variance significant? Do the totals here match what we see on extension?

amitabh94:
This may not be in the scope of this PR but while testing and adding Holesky as network and then coming back to mainnet , the total balance includes balances on Holesky. Is that something you are able to see as well ?

image

amitabh94 (approved):
LGTM!

vinnyhoward (approved):

app/components/UI/Tokens/__snapshots__/index.test.tsx.snap
```diff
@@ -5450,7 +5450,7 @@ exports[`Tokens should show all balance tokens when hideZeroBalanceTokens settin
                                   }
                                   testID="main-balance-test-id"
                                 >
-                                  &lt; $0.01
```
gambinish, line 5453: Just making sure that this diff is intentional, and that we don't want to scope to two decimal places here.

app/selectors/multichain.ts
```diff
@@ -114,46 +161,6 @@ export const selectNativeTokensAcrossChains = createSelector(
...
-        Object.keys(tokenMarketDataByChainId).length === 0
-      ) {
-        const balanceFiatValue =
-          parseFloat(nativeBalanceFormatted) *
-          (currencyRates?.[token.nativeCurrency]?.conversionRate ?? 0);
-
-        balanceFiat = new Intl.NumberFormat('en-US', {
-          style: 'currency',
-          currency: currentCurrency,
-        }).format(balanceFiatValue);
```
gambinish, line 136: When pulling down this and running it locally, when I have USD selected as my currency I am seeing expected results. However, when I select another fiat currency (eg JPY) I get unexpected results. JPY doesn't have decimals, but in the app, it seems like I am seeing 5 decimal places. Which is why I think we reached for the `Intl` API to handle these sorts of edge cases:

Screenshot 2025-01-09 at 6 44 25 PM

I'm also getting some unexpected values for Euro. Shouldn't these also be scoped to two decimal places like the USD?

Screenshot 2025-01-09 at 6 46 18 PM

I also believe that EUR uses commas instead of decimals, which is another place that the Intl API helped.

gambinish (changes requested):
Code looks good. However, I'm noticing some unexpected values for the fiat balance of currencies other than USD.

Are the behaviors in the screenshot above expected? I wonder if there is anyway to maintain precision without losing the inherent fiat display.

app/selectors/multichain.ts
amitabh94: @gambinish Can we revisit this formatting issue in a separate PR as this is a release blocker for 7.38.0 from staking team perspective and we don't see any changes in fiat balances formatting compared to previous versions ?

sahar-fehri (approved):

app/selectors/multichain.ts
sahar-fehri: Also agree to revisit the formatting issue in a separate PR, thank you!

gambinish:
Based on our discussion in slack, we can address the `Intl` number formatting in a fast follow. It appears that it wasn't working as it should have anyway.

For context, this is what we show in extension. Would make sense to align this UX:

Screenshot 2025-01-10 at 8 14 45 AM

gambinish (approved):

PR #12976, opened 2025-01-14
fix: re-designs signatures, continue to use old designs when signing with hardware wallets

jpuri:
## **Description**

Render old designs if hardware wallet or WR aware device is used for signing.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/3870

## **Manual testing steps**

1. Enable re-designs locally
2. Try to sign signature request using hardware wallet
3. It should work in old designs

## **Screenshots/Recordings**
TODO


app/components/Views/confirmations/hooks/useQRHardwareAwareness.ts
```diff
@@ -0,0 +1,45 @@
+import { useState, useEffect } from 'react';
```
jpuri, line 1: This code is inspired from here: https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/QRHardware/withQRHardwareAwareness.tsx

app/components/Views/confirmations/hooks/useConfirmationRedesignEnabled.ts
```diff
@@ -2,26 +2,43 @@ import { useMemo } from 'react';
...
+  } = approvalRequest ?? {
     requestData: {},
   };
 
   const isRedesignedEnabled = useMemo(
     () =>
       (confirmation_redesign as Record<string, string>)?.signatures &&
       process.env.REDESIGNED_SIGNATURE_REQUEST === 'true' &&
+      // following condition will ensure that user is redirected to old designs is using QR scan aware hardware
+      !isSyncingQRHardware &&
```
OGPoyraz, line 27: Just to get the context:
Isn't checking `isExternalHardwareAccount` is enough for us to redirect user to old designs? Why do we need `isSyncingQRHardware` and `isSigningQRObject`?

jpuri: I am actually not 100% is `isSyncingQRHardware` is about hardware wallets or some other QR code scaning can be used. Thus I added this condition to be double safe.

app/components/Views/confirmations/hooks/useQRHardwareAwareness.ts
matthewwalsh0, line 1: Are we certain we this is applicable to mobile? Is there no simpler means we could redirect by? Do we need unit tests for this hook?

app/components/Views/confirmations/hooks/useConfirmationRedesignEnabled.ts
matthewwalsh0, line 27: Is it worth unit tests for this logic given the risk?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/useConfirmationRedesignEnabled.ts
jpuri: These condition should soon go away, but let me add unit tests anyways.

OGPoyraz (dismissed):

matthewwalsh0 (approved):

PR #3909, opened 2022-03-17
Upgrade controller to v27.0.0

gantunesr:
**Description**

Upgrade controllers from v26.0.0 to v27.0.0.

None of the breaking changes affects MMM.

Breaking changes related to `GasFeeController`:

> BREAKING: Further reduce load on Infura by removing non-critical data from the fallback implementation of the Gas API (https://github.com/MetaMask/controllers/pull/712)
> - In GasFeeEstimates – the type of the object returned by fetchGasEstimatesViaEthFeeHistory as well as type of gasFeeEstimates stored in GasFeeController — historicalBaseFeeRange, baseFeeTrend, latestPriorityFeeRange, historicalPriorityFeeRange, priorityFeeTrend, and networkCongestion can now be null. You should update your code to account for this.
> - The ExistingFeeHistoryBlock and NextFeeHistoryBlock types were inconvenient to use and are no longer public. You should use FeeHistoryBlock instead.
> - The BlockFeeHistoryDatasetFetcher class has been removed. There is no replacement.
> - The calculateBaseFeeRange, calculateBaseFeeTrend, calculateNetworkCongestion, calculatePriorityFeeRange, and calculatePriorityFeeTrend functions have been removed. There are no replacements.

**Test Cases to have in consideration regarding Gas Fee**

- Send flow
- Swaps
- Dapp interaction 

**References**

* [Release Notes](https://github.com/MetaMask/controllers/releases/tag/v27.0.0)

PR #13010, opened 2025-01-15
feat: add carousel component

jonybur:
## **Description**

This PR introduces a new carousel component for displaying banners in the wallet view. The implementation includes:

1. A new reusable `Carousel` component with TypeScript support
2. Integration of the carousel in the wallet view to display promotional banners
3. Banner state persistence through Redux
4. New banner images and localized content
5. With there is no balance - the "fund" banner cannot be dismissed

The main motivation is to provide a better way to showcase important announcements and features to users directly in their wallet view, improving user engagement and communication.

## **Related issues**

Fixes: [#3909](https://github.com/MetaMask/MetaMask-planning/issues/3909)

## **Manual testing steps**

1. Open the MetaMask Mobile app
2. Navigate to the wallet view
3. Verify the carousel is displayed at the top of the screen
4. Swipe left/right to navigate through different banners
5. Close the app and reopen to verify banner state persistence
6. Verify proper banner display across different device sizes
## **Screenshots/Recordings**

| Navigating Banner  | MetaMask Card  |  Buy Banner  | 
|:---:|:---:|:---:|
|banner|card|buy|

| Sell Banner  | Settings Banner  | All slides dismissible with balance  | 
|:---:|:---:|:---:|
|sell|card|all_dissmisable_with_balance|

| Cannot dismiss if no balance  | 
|:---:|
||

### **Before**

NA

### **After**

NA


app/components/Views/confirmations/components/TransactionReview/TransactionReviewInformation/index.js
```diff
@@ -763,7 +763,7 @@ const mapStateToProps = (state) => {
     ),
     shouldUseSmartTransaction: selectShouldUseSmartTransaction(state),
   };
-}
+};
```
EtherWizard33, line 766: a lot of the files have a semi colon change, it would simplify the review and probably even require less codeowners

EtherWizard33 (commented):
Nice work.

- I was able to use the Carousel on IOS.
- On android I was not able to confirm since there seems to be an crash after login, kind of at the same time the carouself show. It's most likely a crasg unrelated to the carousel (maybe main has this issue on android at the moment). This could be confirmed by updating from main and confirming if the problem is there or not on main and the feature branch. I have attached a screenshot. But the carousel code looks good to me.

Screenshot 2024-04-18 at 3 56 43 PM

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const { navigate } = useNavigation();
+  const styles = createStyles(colors);
+  const dismissedBanners = useSelector(
+    (state: RootState) => state.banners.dismissedBanners,
+  );
+
+  const isZeroBalance = multichainBalances.totalFiatBalance === 0;
+  const visibleSlides = useMemo(
+    () =>
+      PREDEFINED_SLIDES.map((slide) => {
```
Prithpal-Sooriya, line 120: I like the `PREDEFINED_SLIDES` setup!

(nit/thoughts) Is it worth encapsulating this logic into a single hook? This was we can either test in isolation, or reduce the setup steps required in this component?

```ts

/**
* Initialises/sets up the carousel slides we want to display
*/
export function useCarouselSlides() {
  const dismissedBanners = ...
  const isZeroBalance = ...
  
  const slides = useMemo(() => {
    ...
  }, [])
  
  return slides;
}

```

app/components/UI/Carousel/carousel.types.ts
```diff
@@ -0,0 +1,40 @@
+import { ViewStyle } from 'react-native';
+
+export type SlideId = 'card' | 'fund' | 'cashout' | 'aggregated';
+
+export type NavigationAction =
+  | { type: 'url'; href: string }
+  | {
+      type: 'function';
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      navigate: () => any;
```
Prithpal-Sooriya, line 10: nit - we could be stricter with the types here.
TBH it is only scoped to the Carousel component, so fine.

app/reducers/banners/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      if ((action.payload as { banners?: BannersState })?.banners) {
+        return (action.payload as { banners: BannersState }).banners;
+      }
+      return state;
+    default:
+      return state;
+  }
+};
+
+export default bannersReducer;
```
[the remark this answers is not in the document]
Prithpal-Sooriya:
```ts
type State = {
  dismissedBanners: string[]
}

const bannersSlice = createSlice({
  name: 'banners',
  initialState: {
    dismissedBanners: [],
  } as State,
  reducers: {
    dismissBanner: (state, action: PayloadAction<string>) => {
      if (!state.dismissedBanners.includes(action.payload)) {
        state.dismissedBanners.push(action.payload);
      }
    },
    [REHYDRATE]: (_state, action) => {
      if (action.payload) {
        return action.payload as State;
      }
    }
  }
})

const { actions, reducer } = bannersSlice;

export default reducer;
export const { dismissBanner } = actions;
```

Prithpal-Sooriya (dismissed):

NidhiKJha (dismissed):

EtherWizard33 (dismissed):

app/components/Nav/Main/MainNavigator.js
```diff
@@ -848,6 +848,15 @@ const MainNavigator = () => (
       // eslint-disable-next-line react-native/no-inline-styles
       headerStyle={{ borderBottomWidth: 0 }}
     />
+    {/* TODO: This is added to support slide 4 in the carousel - once changed this can be safely removed*/}
```
Cal-L, line 851: Instead of adding a duplicate stack, could we route them to the settings tab instead or is the slide from bottom sheet the intent?

app/components/UI/Carousel/carousel.types.ts
Cal-L, line 1: rename to types.ts

app/components/UI/Carousel/carousel.types.ts
```diff
@@ -0,0 +1,52 @@
+import { ViewStyle } from 'react-native';
+
+export type SlideId = 'card' | 'fund' | 'cashout' | 'aggregated';
+
+export type NavigationAction =
+  | { type: 'url'; href: string }
```
Cal-L, line 6: Would be easier to read if we abstract each part of the union and label them

app/components/UI/Carousel/index.styles.ts
Cal-L, line 1: rename to styles.ts

app/components/UI/Carousel/index.styles.ts
```diff
@@ -0,0 +1,107 @@
+import { StyleSheet, Dimensions } from 'react-native';
+import { Colors } from '../../../util/theme/models';
+
+const SCREEN_WIDTH = Dimensions.get('window').width;
+const BANNER_WIDTH = SCREEN_WIDTH - 32;
```
Cal-L, line 5: Are the gutter values something that is available in design system tokens yet?

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+import aggregatedImage from '../../../images/banners/banner_image_aggregated.png';
+
+const BannerImages: Record<SlideId, ImageSourcePropType> = {
+  card: cardImage,
+  fund: fundImage,
+  cashout: cashoutImage,
+  aggregated: aggregatedImage,
+};
+
+const PREDEFINED_SLIDES: CarouselSlide[] = [
```
Cal-L, line 48: Abstract these consts into a constants file to clean up this file

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+export const Carousel: FC<CarouselProps> = ({ style }) => {
+  const [selectedIndex, setSelectedIndex] = useState(0);
+  const [pressedSlideId, setPressedSlideId] = useState<string | null>(null);
+  const [hasRendered, setHasRendered] = useState(false);
+  const { trackEvent, createEventBuilder } = useMetrics();
+  const { multichainBalances } = useMultichainBalances();
+  const { colors } = useTheme();
+  const dispatch = useDispatch();
+  const { navigate } = useNavigation();
+  const styles = createStyles(colors);
```
Cal-L, line 112: use the pattern `useStyles` instead so that you don't have to pass in colors

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+  const dismissedBanners = useSelector(
```
Cal-L, line 113: abstract into a selector a memoize if needed

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+      slidesConfig.filter((slide) => {
+        if (slide.id === 'fund' && isZeroBalance) {
+          return true;
+        }
+        return !dismissedBanners.includes(slide.id);
+      }),
+    [slidesConfig, isZeroBalance, dismissedBanners],
+  );
+
+  useEffect(() => {
```
Cal-L, line 150: Could you elaborate on this logic? It looks like it's used to trigger the event where the banners are displayed? If that's the case, we can probably just use a `useEffect`'s component did mount life cycle within each banner component so that we don't have to keep track of the hasRendered state. The track event will automatically trigger as each banner is mounted

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+    return null;
+  }
+
+  return (
+    <View
+      style={[styles.container, style]}
+      testID={WalletViewSelectorsIDs.CAROUSEL_CONTAINER}
+    >
+      <View style={styles.bannerContainer}>
+        <ScrollableTabView
```
Cal-L, line 237: Does this need to be a tab view? It can just be a FlatList with scroll being horizontal

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+          onChangeTab={({ i }) => setSelectedIndex(i)}
+          locked={isSingleSlide}
+          initialPage={0}
+          contentProps={{
+            style: {
+              overflow: 'visible',
+            },
+          }}
+        >
+          {visibleSlides.map((slide) => (
```
Cal-L, line 248: Abstract this into a memoized callback since it's mapping

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+              ]}
+              onPress={() => handleSlideClick(slide.id, slide.navigation)}
+              onPressIn={() => setPressedSlideId(slide.id)}
+              onPressOut={() => setPressedSlideId(null)}
+            >
+              <View style={styles.slideContent}>
+                <View style={styles.imageContainer}>
+                  <Image
+                    source={BannerImages[slide.id]}
+                    style={{
```
Cal-L, line 264: This style can be abstracted into the style sheet

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+            </Pressable>
+          ))}
+        </ScrollableTabView>
+      </View>
+      {!isSingleSlide && (
+        <View
+          testID={WalletViewSelectorsIDs.CAROUSEL_PROGRESS_DOTS}
+          style={styles.progressContainer}
+        >
+          {visibleSlides.map((_, index) => (
```
Cal-L, line 307: Would abstract this into a memoized callback since it maps

app/components/UI/Carousel/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+            <View
+              key={index}
```
Cal-L, line 309: The composition of keys should result in a unique identifier. For example, in this case, it's just a number and can be the same as keys used elsewhere. Let's prefix this with some context. Ex. `banner-slide-{index}`

app/components/Views/Wallet/index.tsx
```diff
@@ -719,6 +723,9 @@ const Wallet = ({
         ) : null}
         <>
           <PortfolioBalance />
+          <View style={styles.carouselContainer}>
+            <Carousel />
```
Cal-L, line 727: Could we provide the style prop to `Carousel` instead and remove the extra `View` wrapper?

Cal-L (commented):
Left some comments

app/components/Nav/Main/MainNavigator.js
vinnyhoward: It currently behaves like a modal if that is what you're referring to when you say bottom sheet intent. 

I've tried directly routing into the nav stack into the general settings before but the transition is jarring. This is how it would look

| Modal  | Direct Routing | 
|:---:|:---:|
|bottom|direct|

app/components/UI/Carousel/index.tsx
vinnyhoward: I am actually not 100% sure why Jony had done this, but I will look into it

app/components/Nav/Main/MainNavigator.js
vinnyhoward: But if you don't think this matters and just directly route, I can do that

app/components/UI/Carousel/index.styles.ts
brianacnguyen: @Cal-L the short answer is no. right now in DS we're moving towards the spacing of Tailwind, so when `twrnc` is implemented in mobile, we can use that. Otherwise right now we don't have anything for devs to use yet

app/components/UI/Carousel/index.tsx
vinnyhoward: I believe it can, looking into it now

app/components/UI/Carousel/index.tsx
vinnyhoward: Updated to use slide id

app/components/UI/Carousel/index.tsx
vinnyhoward: Done

app/components/UI/Carousel/index.tsx
vinnyhoward: Memoized the mapping logic

app/components/UI/Carousel/index.styles.ts
vinnyhoward: Done

app/components/UI/Carousel/carousel.types.ts
vinnyhoward: Done

app/components/UI/Carousel/carousel.types.ts
vinnyhoward: Done

app/components/Nav/Main/MainNavigator.js
Cal-L: Thanks for confirming, this makes sense

vinnyhoward:
@Cal-L to answer your question about the `handleRenderSlides` function and the `useEffect` logic. It looks like it was ported over from extension and isn't needed here

wachunei (dismissed):
Ramp changes LGTM

cortisiko (changes requested):
@vinnyhoward 
Before merging this PR can we test the following:

- The behavior when the app language is set to non-English language such as Spanish/Filipino 
- The behavior when the app is set to dark mode
- The behavior on smaller screen devices
- The behavior when there are competing banners. Test on smaller screen devices. 
    - toggle off display NFT media
        - What happens? 
    - toggle off basic functionality
        -  incoming transactions
            - What happens?

vinnyhoward:
> @vinnyhoward Before merging this PR can we test the following:
> 
> * The behavior when the app language is set to non-English language such as Spanish/Filipino
> * The behavior when the app is set to dark mode
> * The behavior on smaller screen devices
> * The behavior when there are competing banners. Test on smaller screen devices.
>   
>   * toggle off display NFT media
>     
>     * What happens?
>   * toggle off basic functionality
>     
>     * incoming transactions
>       
>       * What happens?

- [ ] Behavior when the app language is set to non-English language such as Spanish/Filipino
  - Switching languages always defaults to english which makes sense, since this PR only adds english. I believe after merging there will be translations merged soon after?
- [ ] Behavior when the app is set to dark mode
  - Having issues in general with dark mode on app - currently investigating
- [x] Behavior on smaller screen devices
  - Works normally and looks good
- [x] Behavior when there are competing banners. Test on smaller screen devices.
  - There are no competing banners - they have their own spaces
- [x]  Toggle off display NFT media
  - The NFT auto detection banner appears beneath the NFT tab
 - [x] Toggle off basic functionality
   - Basic functionality banner appears above portfolio balance
- [x]  Incoming transactions
  - Nothing happens

### Smaller devices and flows

#### Pixel 2

| No banners  | Basic Func. |  NFT Media  | 
|:---:|:---:|:---:|
|banner|basic|nft|

#### iPhone SE

| No banners  | Basic Func. |  NFT Media  | 
|:---:|:---:|:---:|
|banner|basic|nft|

NidhiKJha (dismissed):

EtherWizard33 (dismissed):
re-approving my ealer approval that was reset by a commit

cortisiko (dismissed):
✅ ✅ approving based on test artifacts provided. ty for testing!

cortisiko (approved):
✅ ✅ ✅

EtherWizard33 (commented):
re-approving

EtherWizard33 (approved):
re-approving

Issue #12873, opened 2025-01-08
Create foundation for modular controller init


Cal-L:
This task tracks updating the mobile app's engine architecture to support modular controller initialization. These changes will be heavily influenced by the work that @matthewwalsh0 has done for extension [here](https://github.com/MetaMask/metamask-extension/pull/28948).

The advantages will be the same as Matthew outlined here
- Abstract class based so:
  - Easy to extend with additional configuration and default behaviour.
  - Enforces a readable and consistent template for each controller.
  - Allows iterating over controllers during initialisation with multiple functions to model the lifecycle.
- Backwards compatible with almost no changes to legacy controller initialisation.
- Facilitates an iterative approach meaning teams can refactor individual controllers ad-hoc.
- Facilitates iterative Typescript adoption of controller initialisation code.
- Supports modular unit testing of controller / messenger event handlers.

Since the extension PR includes extension specific changes as well, a review will be done to identify only mobile changes that need to be ported over. The technical description will be filled as those points are identified

Cal-L:
Merged

PR #13065, opened 2025-01-18
feat: Refactor/12873 modular controller init

Cal-L:
## **Description**

These changes modularizes the initialization of controllers in mobile's Engine file. The changes are heavily influenced by the work done on extension by @matthewwalsh0 here - https://github.com/MetaMask/metamask-extension/pull/28948. The mobile version has been simplified to only include and expose what is currently used within the Engine file. 

Note: This version modularizes **AccountsController** creation

## **Related issues**

Fixes:
#12873 

## **Manual testing steps**

Accounts still behave as normal
https://github.com/user-attachments/assets/c2a5f5a0-b607-4e7f-bcb3-9b823ce7345b


Cal-L:
Commits after successful Bitrise E2E are type updates

.github/CODEOWNERS
```diff
@@ -24,6 +24,7 @@ bitrise.yml                                                 @MetaMask/mobile-pla
 yarn.lock                                                   @MetaMask/mobile-platform
 ios/Podfile.lock                                            @MetaMask/mobile-platform
 app/components/Views/BrowserTab/BrowserTab.tsx              @MetaMask/mobile-platform
+app/core/Engine/messengers/                                 @MetaMask/mobile-platform
```
Cal-L, line 27: PR reviews will require platform reviews

app/core/Engine/Engine.ts
```diff
@@ -357,26 +358,15 @@ export class Engine {
...
-          'KeyringController:getKeyringsByType',
-          'KeyringController:getKeyringForAccount',
-        ],
-      });
-    const accountsController = createAccountsController({
-      messenger: accountsControllerMessenger,
-      initialState: initialState.AccountsController,
+    // Modular controller initialization
+    const controllerInitFunctions = [accountsControllerInit];
+    const { controllersByName } = this.#initControllers({
```
Cal-L, line 363: Initialize accounts controller using new pattern

app/core/Engine/controllers/accounts-controller/utils.test.ts
Cal-L, line 1: Checks AccountsController state when it's initialized

app/core/Engine/controllers/accounts-controller/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  },
+};
+
+/**
+ * Initialize the AccountsController.
+ *
+ * @param request - The request object.
+ * @returns The AccountsController.
+ */
+export const accountsControllerInit: ControllerInitFunction<
```
Cal-L, line 23: AccountsController init function used for modular init

app/core/Engine/messengers/accounts-controller-messenger/types.ts
Cal-L, line 1: Define AccountsController actions and events

app/core/Engine/messengers/accounts-controller-messenger/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  AccountsControllerMessengerEvents,
+} from './types';
+
+/**
+ * Get the AccountsControllerMessenger for the AccountsController.
+ *
+ * @param baseControllerMessenger - The base controller messenger.
+ * @returns The AccountsControllerMessenger.
+ */
+export function getAccountsControllerMessenger(
```
Cal-L, line 14: Create accounts controller messenger

app/core/Engine/modular-controller.types.ts
```diff
@@ -0,0 +1,70 @@
...
+/** All controller types by name. */
+export type ControllerByName = {
+  [name in ControllerName]: Controllers[name];
+};
+
+/**
+ * Persisted state for all controllers.
+ * e.g. `{ TransactionController: { transactions: [] } }`.
+ */
+export type ControllerPersistedState = Partial<{
```
Cal-L, line 18: Persisted state only exists for controllers with state

app/core/Engine/utils/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Initialize the controllers according to the provided init objects.
+ * Each init object can be a function that returns a controller, or a `ControllerInit` instance.
+ *
+ * @param options - Options bag.
+ * @param options.initFunctions - Array of init functions.
+ * @param options.initRequest - Base request used to initialize the controllers.
+ * Excluding the properties that are generated by this function.
+ * @returns The initialized controllers and associated data.
+ */
+export function initControllers({
```
Cal-L, line 29: Maps through controller init functions to initialize controllers

app/util/smart-transactions/index.test.ts
Cal-L, line 1: Just updates type name

app/util/smart-transactions/index.ts
Cal-L, line 1: Just updates type name

babel.config.js
```diff
@@ -36,6 +36,10 @@ module.exports = {
       test: './app/core/redux/ReduxService.ts',
       plugins: [['@babel/plugin-transform-private-methods', { loose: true }]],
     },
+    {
+      test: './app/core/Engine/Engine.ts',
```
Cal-L, line 40: To support # as private prefix

Cal-L (commented):
Left comments for context

app/core/Engine/controllers/remote-feature-flag-controller/index.ts
Cal-L, line 1: RemoteFeatureFlagController folders updated but logic not changed for now since this PR focuses only on AccountsController

.github/CODEOWNERS
```diff
@@ -13,7 +13,7 @@ app/core/Engine/Engine.ts                                   @MetaMask/mobile-pla
 app/core/Engine/Engine.test.ts                              @MetaMask/mobile-platform
 app/core/Engine/index.ts                                    @MetaMask/mobile-platform
 app/core/Engine/types.ts                                    @MetaMask/mobile-platform
-app/core/Engine/controllers/RemoteFeatureFlagController/    @MetaMask/mobile-platform
+app/core/Engine/controllers/remote-feature-flag-controller/ @MetaMask/mobile-platform
```
matthewwalsh0, line 16: This raises an interesting question.

Is it better to define the controllers in flat directories to allow flexible ownership?

Or to use team directories to group the controllers and simplify the ownership and related files?

We've gone for the latter in the extension.

app/core/Engine/controllers/accounts-controller/logger.ts
```diff
@@ -3,7 +3,7 @@ import Logger from '../../../../util/Logger';
 import { defaultAccountsControllerState } from './utils';
 
 export function logAccountsControllerCreation(
```
matthewwalsh0, line 5: Why is this a dedicated function and file? Could it not be inline?

Tangent, I'm a huge fan of the `debug` package we use in `core` for nested loggers and would vote to standardise it in the clients too and remove `loglevel` and `Logger` in future if feasible.

app/core/Engine/controllers/accounts-controller/utils.ts
matthewwalsh0, line 1: Do we foresee these files growing to warrant a dedicated `utils.ts` and `constants.ts` per controller?

Would it add value to flatten to a single `x-controller-init.ts` per controller?

app/core/Engine/controllers/accounts-controller/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  AccountsController
+> = (request) => {
+  const { baseControllerMessenger, persistedState } = request;
+
+  const accountControllerState = (persistedState.AccountsController ??
+    defaultAccountsControllerState) as AccountsControllerState;
+
+  const controllerMessenger = getAccountsControllerMessenger(
```
matthewwalsh0, line 31: A benefit of the original class approach I tried was that each controller init class could declare its callback without ever being provided the base messenger.

The static assignment via constant was a way to achieve the same with functions.

The ultimate benefit being that the controller functions never see the base messenger so it's conceptually decoupled and there is zero risk of a stray controller just using the base messenger directly.

While very unlikely, if a controller is calling its own messenger callback, then a PR could be merged with no oversight from the platform team, that stops doing that and just requests the permissions it wants directly from the base messenger.

app/core/Engine/modular-controller.types.ts
matthewwalsh0, line 1: Is it conceptually correct for all the controllers and messengers to live under the `Engine` directory?

What is an `Engine` in this context? Could the `messengers` and `controllers` sit under `app/core`?

app/core/Engine/modular-controller.types.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  baseControllerMessenger: BaseControllerMessenger;
+
+  /**
+   * Retrieve a controller instance by name.
+   * Throws an error if the controller is not yet initialized.
+   *
+   * @param name - The name of the controller to retrieve.
+   */
+  getController<Name extends ControllerName>(
```
matthewwalsh0, line 43: Do we need this in mobile? Are there also initialisation-time dependencies between controllers meaning we have to initialise in a specific order?

app/core/Engine/messengers/accounts-controller-messenger/types.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export type AccountsControllerMessengerActions =
+  | KeyringControllerGetAccountsAction
+  | KeyringControllerGetKeyringsByTypeAction
+  | KeyringControllerGetKeyringForAccountAction;
+
+/**
+ * The events that the AccountsControllerMessenger can handle.
+ */
+export type AccountsControllerMessengerEvents =
```
matthewwalsh0, line 21: Do we get much value defining these in a dedicated file as opposed to the `utils.ts` which isn't likely to develop more functions?

app/core/Engine/messengers/accounts-controller-messenger/utils.ts
matthewwalsh0, line 14: Do we have any instances in mobile where a controller subscribes to a messenger event within `Engine.ts` at the client level?

This was the (discouraged) practice that warranted the init messengers in the extension solution.

matthewwalsh0 (dismissed):
Thanks Cal, this looks great! Added some comments for discussion.

.github/CODEOWNERS
Cal-L: IMO, it's cleaner and more scalable long term to keep the concern of team based directories isolated to the code ownership file. As we've seen in the past and recently, teams are dynamic in a sense that teams can be renamed, shuffled, and even dissolved. It doesn't seem right that those factors should affect the code. We started aggregating some thoughts on why it makes sense keep the concerns of teams separate - https://www.notion.so/metamask-consensys/Project-directory-structure-151f86d67d68809caad6fef0409c5c91?pvs=4

Cal-L: We've also seen confusion where some individuals make folders based on teams and others based on domain, which was the initial intent. More often than not, the lines are blurred between what is considered a domain and team especially when some teams are named after domains.

app/core/Engine/controllers/accounts-controller/logger.ts
Cal-L: This code was implemented by @owencraston in previous implementations so will leave the refactor up to the accounts team. I agree though that this doesn't necessarily need to be its own file

app/core/Engine/controllers/accounts-controller/utils.ts
Cal-L: I think we should keep constants to store things like default state or other things in the future. I realized that redefining the actions and events aren't necessary since the strings are typed checks within the controller constructors so I removed those for one less thing to handle. Logging is a util so I removed the logging file and moved it into utils. And created a `init` file for the init function. I tend to like simple naming within the context of their directories. Any concerns with this approach?

app/core/Engine/modular-controller.types.ts
Cal-L: At the moment, my interpretation is that the Engine is the singleton for direct controller access. I don't have a strong opinion on the nesting of folders at the moment since this is a moving target with the ongoing refactor. I think as long as the controllers are modularized, which it's going to be, it shouldn't be too hard to change the folder structure later.

app/core/Engine/modular-controller.types.ts
Cal-L: I believe we have one or two instances (SmartTransactionController) of this so it will still be useful.

app/core/Engine/messengers/accounts-controller-messenger/types.ts
Cal-L: Removed this since the allowedActions and allowedEvents are already typed

app/core/Engine/controllers/accounts-controller/logger.ts
Cal-L: I'm not entirely sure how accounts team is using those logs at the moment so I won't replace Logger for now. But I did move this function into a utils file instead

app/core/Engine/controllers/accounts-controller/utils.ts
Cal-L: Updated code to use a messenger callback per your suggestion

app/core/Engine/controllers/accounts-controller/utils.ts
Cal-L: Just moved the function into index

app/core/Engine/modular-controller.types.ts
Cal-L: We can tackle updating folder structure in a future refactor

app/core/Engine/controllers/accounts-controller/logger.ts
Cal-L: Won't touch for now in case it breaks logs for accounts team

.github/CODEOWNERS
Cal-L: Will leave for now. I don't suspect we'll run into issues and teams can bring this up if it does

Cal-L:
Test threshold is not related to the changes made in this PR, just from moving files. Disabling sonar cloud to bypass for now.

app/core/Engine/controllers/accounts-controller/index.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  AccountsController,
+  type AccountsControllerMessenger,
+  type AccountsControllerState,
+} from '@metamask/accounts-controller';
+import type { ControllerInitFunction } from '../../types';
+import { logAccountsControllerCreation } from './utils';
+import { defaultAccountsControllerState } from './constants';
+
+// Export constants
```
tommasini, line 10: NIT: comment self explanatory? 
```suggestion

```

tommasini (approved):
This looks amazing!!!!!!

app/core/Engine/Engine.ts
```diff
@@ -410,41 +385,28 @@ export class Engine {
...
-          'KeyringController:getKeyringsByType',
-          'KeyringController:getKeyringForAccount',
-        ],
-      });
-    const accountsController = createAccountsController({
-      messenger: accountsControllerMessenger,
-      initialState: initialState.AccountsController,
+    const { controllersByName } = initModularizedControllers({
+      controllerInitFunctions: {
+        AccountsController: accountsControllerInit,
```
tommasini, line 390: So cool, now the main file of the application, will be way more readable and the coverage of each controller initialization will be fragmented on their own files, catching easily issues on the initialization of our app! Amazing!

gantunesr (approved):
looks good. Reviewed accounts code

FrederikBolding (approved):
Snaps changes lgtm

PR #13111, opened 2025-01-22
feat: add tokens search discovery controller

Bigshmow:
## **Description**

This PR introduces the `@metamask/token-search-discovery-controller` to mobile.  The package provides token search functionality via the Portfolio API & Moralis, which will enable the front end to implement a token search view.

We have created a factory function to introduce the class into Engine context, and redux selectors and a custom hook were added as well as some tests.

See [MMPD-1504](https://consensyssoftware.atlassian.net/browse/MMPD-1504) [MMPD-1526](https://consensyssoftware.atlassian.net/browse/MMPD-1526) in Jira for added context.

## **Related issues**

Fixes:

## **Manual testing steps**

1. This change introduces a dependency and the functionality to implement the search tokens feature but does not implement it herein, so I don't think there is a need for manual testing steps.

[MMPD-1504]: https://consensyssoftware.atlassian.net/browse/MMPD-1504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[MMPD-1526]: https://consensyssoftware.atlassian.net/browse/MMPD-1526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Bigshmow:
@SocketSecurity ignore npm/@metamask/token-search-discovery-controller@1.0.0

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
```diff
@@ -0,0 +1,21 @@
+import { useCallback } from 'react';
+import { useSelector } from 'react-redux';
+import Engine from '../../../core/Engine';
+import { selectRecentTokenSearches } from '../../../selectors/tokenSearchDiscoveryController';
+import { TokenSearchParams } from '@metamask/token-search-discovery-controller/dist/types.cjs';
```
salimtb, line 5: can we export this using the `index.ts` file we have on the controller to avoid using `cjs`

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const useTokenSearchDiscovery = () => {
```
salimtb, line 7: **One important consideration:** If the searchTokens hook is triggered with every change to the search parameters (e.g., when TokenSearchParams contains a search string and the user rapidly types "testToken"), would this execute the search logic individually for each character input?

If this is the intended behavior, I recommend implementing debouncing. However, let's first confirm whether each keystroke indeed triggers a separated call, as this would help determine if the debounce optimization is necessary to prevent excessive execution during rapid input.

salimtb (dismissed):

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
Bigshmow: Interesting, I didn't notice this must've been on auto import.  I must not have explicitly exported this type, but I can update it easily.  Nice catch.

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
Bigshmow: I hope to keep the controller itself as close to single responsibility as we can so as to not confuse callers let's agree to introduce that debounce here in the consuming app.  It's a good idea, thanks; however I'm not sure on exact implementation yet (re: each key press).  Either way I see a `useDebouncedValue` hook that might help here, is that what you suggest?
Something like:
```javascript
const SEARCH_DEBOUNCE_DELAY = 300;

const searchTokens = useCallback(async (params: TokenSearchParams) => {
    const debouncedParams = useDebouncedValue(params, SEARCH_DEBOUNCE_DELAY);
    const { TokenSearchDiscoveryController } = Engine.context;
    return await TokenSearchDiscoveryController.searchTokens(debouncedParams);
  }, []);
```

salimtb (dismissed):

app/core/EngineService/EngineService.test.ts
Cal-L, line 1: It looks like your changes are missing setting up the subscription in EngineService.ts

app/core/Engine/Engine.ts
Cal-L, line 1: Missing adding the controller to `get state()`

Cal-L (commented):
Also missing adding the controller `TokenSearchDiscoveryController` to `Engine/constants`. This allows state change events to to emit data to the subscriptions in EngineService, which then updates Redux. Please test to ensure Redux receives the `TokenSearchDiscoveryController` data once this is set up

app/core/Engine/controllers/TokenSearchDiscoveryController/types.ts
```diff
@@ -0,0 +1,7 @@
+import { TokenSearchDiscoveryControllerMessenger } from '@metamask/token-search-discovery-controller/dist/token-search-discovery-controller.cjs';
```
NicolasMassart, line 1: Q: why is this imported this way?

package.json
```diff
@@ -606,7 +607,8 @@
       "@react-native-firebase/app>firebase>@firebase/firestore>@grpc/proto-loader>protobufjs": false,
       "appium": false,
       "appium>@appium/support>sharp": false,
-      "@metamask/eth-ledger-bridge-keyring>@ledgerhq/hw-app-eth>@ledgerhq/domain-service>eip55>keccak": false
+      "@metamask/eth-ledger-bridge-keyring>@ledgerhq/hw-app-eth>@ledgerhq/domain-service>eip55>keccak": false,
+      "@storybook/builder-webpack5>@storybook/core-common>esbuild": false
```
NicolasMassart, line 611: Q: why this addition?

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  it('should update states correctly when searching tokens', async () => {
```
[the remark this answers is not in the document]
Bigshmow: I had not seen these before, thanks for the tip.

package.json
Bigshmow: Interesting, I'm not so sure.  Will remove.

app/core/Engine/controllers/TokenSearchDiscoveryController/types.ts
Bigshmow: Looks like the messenger type was not exported from the package.index and for now we can only grab it from the common js.  I have an [x-small PR](https://github.com/MetaMask/core/pull/5296) in core that helps to export the type.

Bigshmow:
@SocketSecurity ignore npm/@metamask/token-search-discovery-controller@2.0.0

ziad-saab (dismissed):
Thanks for making the requested changes! Looks good :-)

Bigshmow:
https://github.com/SocketSecurity ignore npm/@metamask/token-search-discovery-controller@2.1.0

Bigshmow:
@SocketSecurity ignore npm/@metamask/token-search-discovery-controller@2.1.0

Cal-L (approved):
LGTM

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
```diff
@@ -15,27 +15,35 @@ export const useTokenSearchDiscovery = () => {
...
-        setResults(result);
-      } catch (err) {
-        setError(err as Error);
-      } finally {
-        setIsLoading(false);
-      }
-    }, SEARCH_DEBOUNCE_DELAY);
+  const searchTokens = useMemo(
+    () =>
+      debounce(async (params: TokenSearchParams) => {
```
NicolasMassart, line 22: C: I see a potential race conditions and inconsistent statebecause of a new debounce function each time the searchTokens function is called. Do you see any risk of multiple concurrent calls and the possibility of earlier search results overwriting later ones? Or is it the expected behaviour to have the older requests lost?

PR #12294, opened 2024-11-14
feat: Add Bitcoin accounts (Flask Only)

owencraston:
## **Description**
This PR does enables the creation of Bitcoin (Mainnet and Testnet) accounts on MetaMask mobile. This is done via the [snap-bitcoin-wallet](https://github.com/MetaMask/snap-bitcoin-wallet) which is included as a preinstalled snap. The added UI buttons to the account actions communicate with the snap, then the snap is routed through our SnapKeyring handlers to enable custom account names. 

This is just the first step and does not include all of the necessary logic/UI to launch this feature. This PR simply enables Bitcoin account creation and renders the proper BTC address and network in the wallet view.

### How
- adds the [Snap Bitcoin Wallet](https://github.com/MetaMask/snap-bitcoin-wallet) as a pre installed snap
- added the `BitcoinWalletSnapSender` which allows us to communicate with the bitcoin snap
- added the Bitcoin Logo to the assets and conditionally render them on the wallet view if the selected account is BTC. This is just visual and is a disabled button. This also does not change any other network related logic elsewhere in the app.
- renamed `selectSelectedInternalAccountChecksummedAddress` to `selectSelectedInternalAccountFormattedAddress` since bitcoin addresses are not checksummed.
- instead of eagerly checksumming addresses, I conditionally checksum them based on wether the address is EVM or not. this is done with the `toFormattedAddress` in `app/util/address/index.ts` and `getFormattedAddressFromInternalAccount` in `app/core/MultiChain/utils.ts`
- added selector `hasCreatedBtcMainnetAccount` and `hasCreatedBtcTestnetAccount`
- added new call to actions in the account actions menu for adding Bitcoin accounts.
- added MultichainSettings to the root state to create a feature flag for Bitcoin. This state can easily be expanded to support other chains in the future. For now we have no way to setting this state to true as this is feature is not production ready. Right now this state is is used to hide the bitcoin feature in the wallet.

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/698

### Testing the multichain feature flag
By default all the bitcoin related features should be hidden behind the `bitcoinSupportEnabled` and `bitcoinTestnetSupportEnabled` which are set to false by default and have no way of modifying them inside the UI.

To verify that these feature flags are working....

1. open `.js.env` and ensure that METAMASK_BUILD_TYPE is set to `flask`
2. run the app on this branch
3. complete onboarding 
4. click on your account
5. the account list menu should open with a CTA to add a new account
[...]

PR #13160, opened 2025-01-23
feat: Add Solana account (Flask only)

owencraston:
## **Description**
- pre installed the [Solana Wallet Snap](https://github.com/MetaMask/snap-solana-wallet?tab=readme-ov-file) on mobile
- include it in the list of pre installed snaps
- add method to trigger the Solana snap
- added copy for adding a solana account
- added the account actions to add a solana account
- added selectors/functions to determine if the user has solana enabled (default to true on flask) and if they have already created a solana account
- allow user to add a solana account under the following conditions
    -  the client is in flask mode
    - solana is enabled (default to true on flask)
    - the user has not already created a solana account
- When the current selected account is Solana, we render the network as Solana and make it non selectable (same as we did for Bitcoin until we deal with the MultichainNetworks).

This PR does not...
- render balances properly
- disable token list on non evm chains
- disable dapp connections on non evm chains

Extension references:

Screenshot 2025-01-23 at 1 36 26 PM

Screenshot 2025-01-23 at 2 33 51 PM

## **Related issues**

Similar to the Bitcoin work done in [this PR](https://github.com/MetaMask/metamask-mobile/pull/12294)

Fixes: https://github.com/MetaMask/accounts-planning/issues/807

## **Manual testing steps**

- checkout this branch
- open .js.env
- set `export METAMASK_BUILD_TYPE="flask"` // flask
- `source .js.env`
- yarn setup
- yarn start:ios
- create/import a wallet 
- open the account menu actions
- you should see that you can add a solana account along with a BTC and BTC testnet account
- click `Add a new Solana account (Beta)`
- a popup with a suggested name should appear
- click confirm
- the newly selected account should be your Solana account and the network picker should show the Solana logo 
- the network should not be clickable 
- open up the account actions again and notice that you can add another Sol account. This is different from Bitcoin which can only support one account. 
- repeat these steps with bitcoin

### Testing Production
- reset your wallet
- reset your `.js.env` to have `METAMASK_BUILD_TYPE = 'main'`
- rebuild the app
- create/import wallet
- open the account actions
- there should be no mention of Bitcoin or Solana
- there should be no snaps section in the settings 

### **After**

https://github.com/user-attachments/assets/6d3739cb-019f-435f-a326-bf20ff82422a


app/reducers/multichain/index.ts
```diff
@@ -1,10 +1,13 @@
+///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
 import { MultichainSettingsState } from '../../actions/multichain/state';
 
 export const initialState: MultichainSettingsState = {
-  bitcoinSupportEnabled: false,
-  bitcoinTestnetSupportEnabled: false,
+  bitcoinSupportEnabled: true,
+  bitcoinTestnetSupportEnabled: true,
+  solanaSupportEnabled: true,
```
owencraston, line 7: Setting these to true on flask. These variable should only exist in flask anyway so having them set to true should not interfere with production.

app/components/UI/Navbar/index.js
```diff
@@ -917,10 +921,9 @@ export function getWalletNavbarOptions(
   let formattedAddress = toChecksumHexAddress(selectedInternalAccount.address);
 
   ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
-  if (isBtcAccount(selectedInternalAccount)) {
-    // BTC addresses are not checksummed
-    formattedAddress = selectedInternalAccount.address;
-  }
+  formattedAddress = getFormattedAddressFromInternalAccount(
```
tommasini, line 924: Curious, this change doens't bring anything differen, any reason why implemented it??

tommasini (dismissed):
LGTM

app/components/UI/Navbar/index.js
owencraston: Since we now have BTC AND SOLANA accounts, the `(isBtcAccount(selectedInternalAccount))`  no longer applies here. The `getFormattedAddressFromInternalAccount` does handle eth accounts however in order to have all the multichain code fenced (in between ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps) and ///: END:ONLY_INCLUDE_IF) I had import it and call it inside this code fence. The only reason I did not remove the `let formattedAddress = toChecksumHexAddress(selectedInternalAccount.address);` is for the code fence.

owencraston:
@SocketSecurity ignore npm/@metamask/solana-wallet-snap@1.2.0
@SocketSecurity ignore npm/@metamask/bitcoin-wallet-snap@0.9.0

These are both internal metamask packages.

.nvmrc
```diff
@@ -1 +1 @@
-20.17.0
+20.18.0
```
owencraston, line 1: The @solana package required a minimum version of 20.18.0. I looked at the [changelog](https://github.com/nodejs/node/releases/tag/v20.18.0) and there were no significant changes.

tommasini (dismissed):

tommasini (dismissed):

FrederikBolding (dismissed):

aganglada (dismissed):
GREAT work! 🎉 
QQ, how is the flask build code fence called, how come we are using `keyring-snaps`

.nvmrc
ccharly, line 1: Nit: But I wonder if this change was intentional or required (for this PR)?

Not a big problem, but I would have expect a dedicated PR for this kind of thing 😅

app/components/Views/AddAccountActions/AddAccountActions.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+      );
+
+      const solButton = screen.getByTestId(
+        AddAccountBottomSheetSelectorsIDs.ADD_SOLANA_ACCOUNT_BUTTON,
+      );
+      expect(solButton.findByType(Text).props.children).toBe(
+        'Add a new Solana Account (Beta)',
+      );
+      expect(solButton.props.disabled).toBe(true);
+    });
```
ccharly, line 216: Actually Solana allows for multiple account creations on the extension. We could keep it like this for now, but we would have to re-enable this somehow to align with the other client!

app/core/Multichain/test/utils.test.ts
```diff
@@ -87,6 +91,22 @@ const mockBTCAccount: InternalAccount = {
   type: BtcAccountType.P2wpkh,
 };
 
+const mockSolAccount: InternalAccount = {
+  address: SOL_ADDRESSES,
```
ccharly, line 95: Nit: Should this be singular instead? 😅 (we would need to change the declaration too)
Nit: Also, other mock addresses are prefixed by `MOCK_`, we could update this to align everything.

```suggestion
  address: SOL_ADDRESS,
```

app/core/Multichain/test/utils.test.ts
```diff
@@ -87,6 +91,22 @@ const mockBTCAccount: InternalAccount = {
...
+  id: '1',
+  type: SolAccountType.DataAccount,
+  methods: [SolMethod.SendAndConfirmTransaction],
+  options: {},
+  metadata: {
+    name: 'Solana Account',
+    importTime: 1684232000456,
+    keyring: {
+      type: KeyringTypes.hd,
```
ccharly, line 104: This should be attached to the Snap keyring:

```suggestion
      type: KeyringTypes.snap,
```

But I realized we also made the same mistake for BTC 🙈 If we change that, we might wanna add a dummy `snap` metadata as well (but looks like it's not truly needed for now and we don't have it either on the extension 🙃).

app/core/Multichain/test/utils.test.ts
```diff
@@ -87,6 +91,22 @@ const mockBTCAccount: InternalAccount = {
...
+    },
+  },
+  scopes: [SolScopes.Namespace],
```
ccharly, line 107: On the extension side, we usually use a full list of scopes with only CAIP-2 chain IDs here. We're about to enforce this for some account types, so would be great to fix it now:

```suggestion
  scopes: [SolScopes.Mainnet, SolScopes.Testnet, SolScopes.Devnet],
```

app/selectors/accountsController.ts
```diff
@@ -119,4 +120,13 @@ export function hasCreatedBtcTestnetAccount(state: RootState): boolean {
     (account) => isBtcAccount(account) && isBtcTestnetAddress(account.address),
   );
 }
+
+/**
+ * A selector that returns whether the user has already created a Solana mainnet account
+ */
+export function hasCreatedSolanaMainnetAccount(state: RootState): boolean {
```
ccharly, line 127: Technically speaking a Solana account is available for all Solana networks, so `hasCreatedSolanaAccount` could be a better name here? WDYT @aganglada?

Unless you want to keep the "symmetry" with the main bitcoin account?

locales/languages/en.json
```diff
@@ -3377,7 +3377,8 @@
     "add_hardware_wallet": "Add hardware wallet",
     "import_account": "Import account",
     "add_bitcoin_account_testnet": "Add a new Bitcoin Account (Testnet)",
-    "add_bitcoin_account_mainnet": "Add a new Bitcoin Account (Beta)"
+    "add_bitcoin_account_mainnet": "Add a new Bitcoin Account (Beta)",
+    "add_solana_account_mainnet": "Add a new Solana Account (Beta)"
```
ccharly, line 3381: Same comment here, we could maybe remove the `_mainnet` here? WDYT @aganglada?

package.json
```diff
@@ -233,6 +234,7 @@
     "@segment/sovran-react-native": "^1.0.4",
     "@sentry/integrations": "6.3.1",
     "@sentry/react-native": "^5.33.0",
+    "@solana/addresses": "2.0.0-rc.4",
```
ccharly, line 237: Could be great to use a "stable" version here, what about just using the `2.0.0`?

.nvmrc
owencraston: It is required because the Solana packages have a minimum of 20.18.0.

locales/languages/en.json
owencraston: I'll remove the mainnet, it sounds like it doesn't apply here.

package.json
owencraston: yes, I just wanted to use the same one asa extension but 2.0.0 is available.

app/core/Multichain/test/utils.test.ts
owencraston: good point.

app/selectors/accountsController.ts
owencraston: It sounds like we do not need this selector anymore.

app/components/Views/AddAccountActions/AddAccountActions.test.tsx
owencraston: I'll enable adding multiple accounts. I just figured it was the same as Bitcoin but I was wrong.

app/core/Multichain/test/utils.test.ts
owencraston: I will add both to this account and the BTC account.

package.json
owencraston: Since we do not need to even determine if a user has already created a sol account, we can remove this package for now. Likely it will be needed in the future but we can remove it now.

.nvmrc
owencraston: Since we are no longer checking to see if a user has created a sol account, we no longer need the solana/address package which required this bump. Reverted.

owencraston:
@aganglada It works by using this [shared custom library](https://github.com/MetaMask/core/blob/main/packages/build-utils/src/transforms/README.md#remove-fenced-code) inside the [metro.transform](https://github.com/MetaMask/metamask-mobile/blob/bb7907ad8ae38cf5786658b78ce3fce9c3b42b75/metro.transform.js#L24) file. In that file we define the flags that are supported on flask vs main.

owencraston:
https://github.com/user-attachments/assets/6fc5d28d-9710-4fbc-9e8c-2369368d117b

aganglada (approved):

tommasini (approved):

FrederikBolding (approved):

PR #13178, opened 2025-01-24
CHANGED - use Consensys Hosted Runners

waskow-consensys:
## **Description**

Please add the `metamask-public` runner group to this repo for the runners to be picked up

Consensys has shared infra that we are using to save money monthly on infra costs.
These follow Github's best practices for Github runners https://docs.github.com/en/actions/hosting-your-own-runners

Thanks!

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,8 @@
+self-hosted-runner:
+  # Labels of self-hosted runner in array of strings.
+  labels:
+    - gha-mm-scale-set-ubuntu-22.04-amd64-med
+# # Configuration variables in array of strings defined in your repository or
+# # organization. `null` means disabling configuration variables check.
+# # Empty array means no configuration variable is allowed.
+# config-variables: null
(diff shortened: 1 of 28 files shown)


waskow-consensys:

Gudahtt (commented):
Could you elaborate on why we'd want to do this, or how this would save money? It looks like this would cost more money, as it's replacing free runners with runners that we pay hosting costs for.

waskow-consensys:
> Could you elaborate on why we'd want to do this, or how this would save money? It looks like this would cost more money, as it's replacing free runners with runners that we pay hosting costs for.

Definitely, I actually just reached out to TechOps to see where we are at with our allotment of minutes per month. This repo showed up at the top of the report I received. So just being pro-active. Will report back.

waskow-consensys:
You are correct 
>GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners.
Just standard runners here
Thanks for raising!

PR #13182, opened 2025-01-24
fix: remove thrown Error for logging in Sentry on unrecognized network switch

### LINKED ISSUE #11731: Error: Stop throwing error in Sentry when user tries to add a chain with unrecognized rpc url
Sentry Issue: [METAMASK-MOBILE-2DVH](https://metamask.sentry.io/issues/5351012899/?referrer=github_integration)

```
Error: {"code":4100,"message":"Request blocked due to spam filter."}
  at ?anon_0_ (app/util/Logger/index.ts:83:32)
...
(11 additional frame(s) were not displayed)
```

Steps to reproduce:

- Go to https://bridge.coredao.org/coreBTC
- Click on "Add coreBTC to your wallet"
- MM will throw "Unrecognized chain ID "0x45c". Try adding the chain using wallet_addEthereumChain first."

Here we throw a Sentry error while we'd like to send and Error event in Segment instead.

frankvonhoven:
## **Description**

Don't throw and Error on unrecognized network, just report it to Sentry

## **Related issues**

Fixes: #11731 

## **Manual testing steps**

In in-app browser:
1. Go to https://bridge.coredao.org/coreBTC
2. Click on "Add coreBTC to your wallet"
3. MM should log in Sentry "Unrecognized chain ID "0x45c". Try adding the chain using wallet_addEthereumChain first."


smilingkylan (approved):
LGTM

app/core/RPCMethods/wallet_switchEthereumChain.js
```diff
@@ -85,10 +85,16 @@ const wallet_switchEthereumChain = async ({
     return;
   }
 
-  throw providerErrors.custom({
```
Gudahtt, line 88: It seems like there has been a misunderstanding about what this was intended to do. This error is for the _dapp_, not us. It's important that we tell the dapp that the request failed, and _why_ it failed.

Gudahtt (changes requested):
Seems like this would break the Wallet API's error handing for this case

app/core/RPCMethods/wallet_switchEthereumChain.js
```diff
@@ -85,10 +85,16 @@ const wallet_switchEthereumChain = async ({
...
-    code: 4902, // To-be-standardized "unrecognized chain ID" error
-    message: `Unrecognized chain ID "${_chainId}". Try adding the chain using wallet_addEthereumChain first.`,
-  });
+  MetaMetrics.getInstance().trackEvent(
+    MetricsEventBuilder.createEventBuilder(
+      MetaMetricsEvents.NETWORK_UNRECOGNIZED,
```
Gudahtt, line 90: Why are we tracking this in Segment? What insights are you hoping to get from this metric event?

app/core/RPCMethods/wallet_switchEthereumChain.js
NicolasMassart: 100% agree with @Gudahtt, this is the big difficulty here.
We HAVE to throw an error on the provider for the dapp to get the error feedback.

>[!warning]
> Removing this error throw is going to break all the dapps workflows !

The side effect is, as we have a Sentry error catch-all at the top of the app, every thrown error that is not caught in try/catch is also considered as an unhandled error by Sentry catch-all. And it's reported to sentry.

Here, to me, we should have the error reported to the dapp only, no Sentry, and optionally if we really want to track them, have a Segment tracking event.

But given that it's totally out of our scope and we can't do anything about it, I'm not sure we should track that. It would only tell us that dapps are generating errors. Maybe we could monitor what dapp throws the more errors and contact them, but is this really our problem?

NicolasMassart (changes requested):

app/core/RPCMethods/wallet_switchEthereumChain.js
frankvonhoven: From the conversation for this PR it would seem that the last lint in the ask in the original ticket is not valid?:
```
Here we throw a Sentry error while we'd like to send and Error event in Segment instead.
```

app/core/RPCMethods/wallet_switchEthereumChain.js
gauthierpetetin: @frankvonhoven is there any way to throw the error for the dapp without throwing it to Sentry? If yes we should go for that option.

app/core/RPCMethods/wallet_switchEthereumChain.js
gauthierpetetin: Unless there's a cost for us to save this event, I'd say it's always interesting to track its evolution over time, as its an indicator of the friction experienced by our users.
It can also help inform the decision of what networks we want to add to the list of popular networks.

app/core/RPCMethods/wallet_switchEthereumChain.js
frankvonhoven: That's the question, afaik when we throw an Error, it is caught by the error boundary, and reported to Sentry, so whenever we throw an error in the app, it will always get reported to sentry and the user will see the Error Screen. But we could check for that particular error, and perhaps prevent it from going to Sentry. But then we're going to be responsible for maintaining a block list. Perhaps we could silence that error in Sentry? or is it a frequent occurrence that it is polluting Sentry?

gauthierpetetin: It is noisy in Sentry: 820k over the past 30 days.
Do we know what are the pros/cons of maintaining a blocklist in the client VS silencing the error in Sentry?

joaoloureirop: We could add this error to the [custom error handler](https://github.com/MetaMask/metamask-mobile/blob/d01cf65b91aba194db789df0224cd029cf554ff8/app/core/ErrorHandler/ErrorHandler.ts)

To improve the maintainability of this handler, we could adopt an error message pattern that matches the discard mechanism filter.

PR #13208, opened 2025-01-27
fix: update token balances on pull tokenlist pull down

sahar-fehri:
## **Description**

PR to add updating balances as part of pulling down on the tokenList

## **Related issues**

Fixes:

## **Manual testing steps**

1. Make a swap between any 2 tokens
2. Go back to home page and notice your balance is not updated for the two tokens.
3. Pull down on the tokenList
4. Your balances should now be refreshed

### **After**

https://github.com/user-attachments/assets/585f8789-3509-443f-be69-eab8e4ef1ac5


app/components/UI/Tokens/index.tsx
```diff
@@ -376,6 +377,12 @@ const Tokens: React.FC<TokensI> = memo(({ tokens }) => {
             ? (Object.keys(networkConfigurationsByChainId) as Hex[])
             : [currentChainId],
         }),
+
+        TokenBalancesController.updateBalances({
+          chainIds: isPortfolioViewEnabled()
+            ? (Object.keys(networkConfigurationsByChainId) as Hex[])
+            : [currentChainId],
+        }),
```
sahar-fehri, line 385: We had a reported issue that the balances on tokenList are not being refreshed after swaps.
This should have been done once the transaction is confirmed [here](https://github.com/MetaMask/metamask-mobile/blob/1091432c0b48a3855c50b4fe6494388b359ebc0c/app/core/NotificationManager.js#L238) and call `updateBalancesByChainId` only for chain where the swap happened.
To mitigate this issue we are now adding `updateBalances` action to the list of actions once the user pulls down on the tokenList.
IMO this could be a temporary fix and can be removed once the update is done as part of the tx confirmations flow if we notice any slowness with the UI after adding this.

gambinish: I think this is a good stop gap solution, and is a good addition to the refresh logic. One thing we should maybe be aware of is that balance fetching is expensive, and this action can be spammed by users.

I agree that we should fix the root cause at the controller level.

gambinish (approved):
This works as expected. I left a comment on your thread about something that I think we should keep in mind.

salimtb (approved):

app/components/UI/Tokens/index.tsx
salimtb: you don't need to use `isPortfolioViewEnabled` anymore since it's true by default. Instead, consider using `networkFilterToken`. If the current network is selected, there's no need to update balances across all chains for all networks, which would improve optimization.

sahar-fehri: I think we have plans to remove `isPortfolioViewEnabled` in future and this can be done as part of it; to not chase approvals again will resolve this since it is a nit comment 🙏

PR #13328, opened 2025-02-04
feat: Token search and discovery with swaps

ziad-saab:
## **Description**
This PR adds token search and discovery features to MetaMask:
- browser URL bar now allows searching for tokens by name (prefix/autocomplete) and token address
- users can swap directly from the search results (will switch or add the network if they don't have it)
- users can see asset details for searched tokens, complete with price and market data
- users can swap the found asset from the asset details page's swap button (will switch or add the network if they don't have it)

### **After**

https://github.com/user-attachments/assets/449ff284-e127-4829-a65f-19fc9cf39a73


app/components/Views/AssetLoader/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const AssetLoader: React.FC<AssetLoaderProps> = ({ route: { params: { address, chainId } } }) => {
+    const tokenResult = useSelector((state: RootState) => selectSearchedToken(state, chainId, address));
+
+    const navigation = useNavigation();
+
+    useEffect(() => {
+        if (!tokenResult) {
+            Engine.context.TokensController.addSearchedToken(address, chainId);
```
gambinish, line 27: If a user searches for a token on a network not yet added to their account, we won't have some data needed to handle certain fiat conversion throughout the app (for instance, on `AsssetOverview`)

From a technical perspective, this may require us to add the required chain to the `NetworkController` state whenever you TokensController.addSearchedToken(), so that the `networkConfigurations.[chainId].nativeCurrency` for the fiat conversion is also available for the UI to consume.

We will likely also want to handle the network switch on behalf of the user, so that the correct network appears below the ticker on the TabView:

Screenshot 2025-02-05 at 1 41 20 PM

[Slack conversation for reference](https://consensys.slack.com/archives/C08AN1X0CSX/p1738688806325859)

app/components/UI/AssetOverview/TokenDetails/TokenDetails.tsx
```diff
@@ -72,17 +73,27 @@ const TokenDetails: React.FC<TokenDetailsProps> = ({ asset }) => {
...
   const tokenList = useSelector(selectTokenList);
 
-  const conversionRate = isPortfolioViewEnabled()
-    ? conversionRateBySymbol
-    : conversionRateLegacy;
+  let conversionRate;
+  if (isAssetFromSearch(asset)) {
+    conversionRate = 1;
+  } else if (isPortfolioViewEnabled()) {
+    conversionRate = conversionRateBySymbol;
```
gambinish, line 80: We should treat the `isPortfolioViewEnabled` flag as deprecated. No need to remove it here, but just a heads up that we should probably backlog a ticket for this removal across platforms.

app/selectors/tokenSearchDiscoveryDataController.ts
```diff
@@ -0,0 +1,28 @@
...
+export const selectSupportedSwapTokenAddressesByChainId = createSelector(
+  selectTokenSearchDiscoveryDataControllerState,
+  (state) => state?.swapsTokenAddressesByChainId,
+);
+
+export const selectSupportedSwapTokenAddressesForChainId = createSelector(
+  selectTokenSearchDiscoveryDataControllerState,
+  (_state: RootState, chainId: Hex) => chainId,
+  (state, chainId) => state?.swapsTokenAddressesByChainId[chainId]?.addresses,
+);
```
gambinish, line 28: Can we try to convert these into `createDeepEqualSelector` so that they get memoized as much as possible?

app/selectors/tokenSearchDiscoveryDataController.ts
```diff
@@ -0,0 +1,28 @@
+import { createSelector } from 'reselect';
+import { RootState } from '../reducers';
+import { Hex } from '@metamask/utils';
+import { selectCurrentCurrency } from './currencyRateController';
+
+export const isAssetFromSearch = (asset: unknown) => typeof asset === 'object' && asset !== null && 'isFromSearch' in asset && asset.isFromSearch === true;
```
gambinish, line 6: This condition is hard for me to wrap my head around, but it seems very important to your implementation. Is there any way that we can simplify this and/or make this more readable?

At the very least I think that adding some comments here to explain what these conditions realy on would be good.

yarn.lock
```diff
@@ -4425,12 +4425,12 @@
...
     nanoid "^3.3.8"
 
-"@metamask/assets-controllers@^51.0.2":
-  version "51.0.2"
-  resolved "https://registry.yarnpkg.com/@metamask/assets-controllers/-/assets-controllers-51.0.2.tgz#fc8ed8981fce62e4c7327393faffc2017874644d"
-  integrity sha512-dEJJw6qc5HMEiMJUTOrKhTYTMS8tuf3+gofZ9PHGSvPibUx/DivC1L3bouiMNgi2xzzH9ktfmkFyCN1lzFk22w==
+"@metamask/assets-controllers@^53.1.1":
+  version "53.1.1"
+  resolved "https://registry.yarnpkg.com/@metamask/assets-controllers/-/assets-controllers-53.1.1.tgz#87e00c362247e0c0a4f3df8bd124d6e5bb0bb16c"
+  integrity sha512-AmB5/MoqCr+JJXZm0M1lmO1AUrkmSRWaVqGauBszQeXcd3aVmycZHMWEIFKNgo0bf5kX5xIkFg2CVLaMlsv+Xg==
```
gambinish, line 4431: We have an [open PR](https://github.com/MetaMask/metamask-mobile/pull/14201) to upgrade AssetsController to 55.0.1.

Is this patch still necessary? Or can we use the bumped version?

app/components/hooks/useAddNetwork.tsx
```diff
@@ -0,0 +1,45 @@
...
+            onClose={onCloseModal}
+            networkConfiguration={popularNetwork}
+            showPopularNetworkModal
+            shouldNetworkSwitchPopToWallet={false}
+            autoSwitchNetwork
+            onNetworkSwitch={noop}
+            onAccept={resolveAddNetwork}
+            onReject={rejectAddNetwork}
+         />
+    );
```
gambinish, line 39: Do we need to return the network modal jsx in this hook? I've run into difficult to troubleshoot rendering bugs with this pattern in the past.

My preference would be to return a boolean value, such as `shouldShowNetworkModal`, and read the hook from a react component, and then conditionally render the JSX in the component's body accordingly.

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -1175,10 +1175,18 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
...
-    onSubmitEditing(url);
+
+    if (item.type === 'tokens') {
+      navigation.navigate(Routes.BROWSER.ASSET_LOADER, {
+        chainId: item.chainId,
+        address: item.address,
+      });
+    } else {
+      onSubmitEditing(item.url);
+    }
```
gambinish, line 1189: Let's convert this into a `useCallback` hook to prevent unnecessary re-renders.

app/components/Views/AssetLoader/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const AssetLoader: React.FC<AssetLoaderProps> = ({ route: { params: { address, chainId } } }) => {
+    const tokenResult = useSelector((state: RootState) => selectTokenDisplayData(state, chainId, address));
+    const navigation = useNavigation();
+
+    const { styles } = useStyles(styleSheet, {});
+
+    useEffect(() => {
+        Engine.context.TokenSearchDiscoveryDataController.fetchTokenDisplayData(chainId, address);
```
gambinish, line 29:
```suggestion
const fetchTokenDisplayData = Engine.context.TokenSearchDiscoveryDataController.fetchTokenDisplayData
fetchTokenDisplayData(chainId, address);
```
```suggestion
        Engine.context.TokenSearchDiscoveryDataController.fetchTokenDisplayData(chainId, address);
```

gambinish (changes requested):
Nice job! I left a few comments on things that caught my eye. Feel free to push back where you feel I may have misunderstood something.

Additionally, it looks like we need to `lint:tsc` and get Bitrise to pass. Could you update this branch with `main` and then re-run Bitrise? Once it's fully green I'm happy to pull this and run it locally.

Feel free to ping me directly on Slack as well.

app/components/UI/AssetOverview/TokenDetails/TokenDetails.tsx
ziad-saab: @gambinish I had a feeling that this would be the case. Should I remove it and keep only the condition where it's true?

yarn.lock
ziad-saab: This is not necessary. I brought in the changes from the open PR so that more tests would pass. Once the open PR to upgrade packages is merged, I expect all these changes to not be in my PR anymore.

This is important: I don't want this PR to contain any updates to packages. I'd rather the package updates be done in a separate PR, so that in case this current PR needs to be reverted, it won't affect the upgrade of packages.

app/components/hooks/useAddNetwork.tsx
ziad-saab: @gambinish I'd love to hear more about the rendering bugs you experienced in the past. In this case here, note the reliance on `onAccept` and `onReject` which are defined in the hook. If we were to just use a flag, we'd have to expose `resolveAddNetwork` and `rejectAddnetwork`.

It's not impossible nor is it even a problem, but I'd like to find out more about the bugs you experienced in the past to be able to make the best decision about this. I agree that it's not a common pattern. I picked it because it solved the problem with as much encapsulation as it could.

app/components/Views/AssetLoader/index.tsx
ziad-saab: @gambinish thanks for this suggestion! I'd like to hear more about what prompted it, so that it may inform further coding decisions.

ziad-saab:
@gambinish thanks for the review! The `lint:tsc` errors are due to some package updates that should not even be part of this PR. I'm waiting for the package update PRs to be merged first, then I'll merge `main` into this PR. At that point there should be less files changed, more code coverage, and no more errors.

app/components/UI/AssetOverview/TokenDetails/TokenDetails.tsx
gambinish: Yeah, let's treat `isPortfolioViewEnabled` as true. Condition shouldn't be necessary and we will be removing this early next week.

app/components/hooks/useAddNetwork.tsx
gambinish: While this technically works, the main anti-pattern here is that hooks are meant to run inside of React component, not behave like them. This can sometimes lead to unpredictable behavior, especially when other JSX elements are calling hooks of their own. Hooks were meant to separate logic from UI, rather than co-mingling them.

What would happen if we eventually needed to add some custom hook inside of `NetworkModals`? Then any component consuming `useAddNetwork` would not only call your hook, but may result in multiple render trees that are hard to track or debug.

There are also some testing implications here since you'd now need to test rendering inside of a hook.

app/components/Views/AssetLoader/index.tsx
gambinish: This is just a readability nit, doesn't need to block

app/components/UI/AssetOverview/TokenDetails/TokenDetails.tsx
ziad-saab: removed from TokenDetails.

yarn.lock
ziad-saab: patch removed due to assets-controllers update to v55 from another PR

app/components/hooks/useAddNetwork.tsx
ziad-saab: Hey @gambinish thanks for trying to explain the issue! I'm still having trouble grasping it. I'd really love to take advantage of this issue to upgrade my React skills and understanding.

> What would happen if we eventually needed to add some custom hook inside of NetworkModals? Then any component consuming useAddNetwork would not only call your hook, but may result in multiple render trees that are hard to track or debug.

My understanding is that `useAddNetwork` is simply a function returning a piece of JSX to be rendered inside another component. Only when that piece of JSX gets rendered do its hooks take effect. I'd like to understand the difference between returning this piece of JSX (basically a call to `React.createElement` the way I understand it) from `useAddNetwork`, and creating it inside another component.

> There are also some testing implications here since you'd now need to test rendering inside of a hook.

I'd love to understand the implications. This would not only improve my React-fu, but it would help me create a better solution than the one I have here.

gambinish: If values like `resolveAddNetwork` change between renders, the JSX inside the hook can end up referencing stale versions of those values. Since that JSX is being re-created on every render, it can also trigger unnecessary re-renders of the modal, especially if multiple components are using the hook and each one is rendering its own instance. It just adds up fast and makes things harder to manage.

As far as testing goes, when the hook just returns data and functions (instead of JSX), you can test it in isolation without having to render anything. No need for snapshot tests or mounting modals — you just assert that the hook gives you the right state and behaviors.

I’d recommend returning just the state and control functions from the hook, and rendering `NetworkModals` in the parent component instead:

```
const { addPopularNetwork, showNetworkModal, onCloseModal, ... } = useAddNetwork();

return (
  <>
    {showNetworkModal && (
      <NetworkModals
        isVisible
        navigation={navigation}
        onClose={onCloseModal}
        networkConfiguration={popularNetwork}
        showPopularNetworkModal
        shouldNetworkSwitchPopToWallet={false}
        autoSwitchNetwork
        onNetworkSwitch={noop}
        onAccept={resolveAddNetwork}
        onReject={rejectAddNetwork}
      />
    )}
  </>
);
```

Added benefit is that this keeps your logic reusable. The hook can be used anywhere, but the UI doesn’t matter. So, if you want to use the hook, but render a different modal, or even the same modal with different props, you can do so without having to untangle the UI from the hook.

gambinish: I also don't want to be a blocker. If you feel like this change is not reasonable, let me know

ziad-saab: @gambinish thanks a lot for these insights! How about, instead of a hook, if this was a component with a `useImperativeHandle` and a ref, then we can call `await addNetworkRef.current?.addNetwork(network)`?

gambinish (dismissed):

app/components/hooks/useAddNetwork.tsx
ziad-saab: Following our conversation on Slack, we'll keep it as a hook. It'll be converted to a component in the next PR for token discovery.

brianacnguyen (dismissed):

darkwing (dismissed):

gambinish (dismissed):

brianacnguyen (dismissed):

GuiBibeau (dismissed):

smilingkylan (commented):
FYI I'm not seeing anything my browser search results when I build your branch. Do I need to enable / disable a feature flag or anything like that?
image

app/components/UI/UrlAutocomplete/Result.tsx
```diff
@@ -6,61 +6,116 @@ import WebsiteIcon from '../WebsiteIcon';
...
+    onSwapPress: (result: TokenSearchResult) => void;
 }
 
-export const Result: React.FC<ResultProps> = memo(({ result, onPress }) => {
+export const Result: React.FC<ResultProps> = memo(({ result, onPress, onSwapPress }) => {
     const theme = useTheme();
     const styles = stylesheet({theme});
 
-    const name = typeof result.name === 'string' ? result.name : getHost(result.url);
+    const name = typeof result.name === 'string' || result.type === 'tokens' ? result.name : getHost(result.url);
```
smilingkylan, line 35: Could you please change the property name from `type` to something that is more easily searchable? My concern here is that if someone tries to search for this property they will end up with thousands of matches in this repo

app/components/UI/AssetOverview/AssetOverview.tsx
```diff
@@ -333,15 +375,21 @@ const AssetOverview: React.FC<AssetOverviewProps> = ({
...
   let priceDiff = 0;
 
-  const tickerConversionRate =
-    conversionRateByTicker?.[nativeCurrency]?.conversionRate ?? 0;
-  currentPrice =
-    exchangeRate && tickerConversionRate
-      ? exchangeRate * tickerConversionRate
-      : 0;
+  if (isAssetFromSearch(asset) && tokenResult?.found) {
+    currentPrice = tokenResult.price?.price || 0;
```
smilingkylan, line 384: IMO having a property like `price` that is nested inside of another property named `price` will cause confusion. If you have the ability to adjust the parent or child property name then I would like it changed

app/components/UI/UrlAutocomplete/index.tsx
```diff
@@ -40,25 +55,48 @@ const UrlAutocomplete = forwardRef<
...
-    const newResultsByCategory = ORDERED_CATEGORIES.flatMap((category) => {
-      let data = results.filter((result, index, self) =>
+  useEffect(() => {
+    if (currentCurrency) {
+      Engine.context.CurrencyRateController.updateExchangeRate([currentCurrency]);
+    }
+  }, [currentCurrency]);
+
+  const resultsByCategory: {category: string, data: AutocompleteSearchResult[]}[] = useMemo(() => (
+    ORDERED_CATEGORIES.flatMap((category) => {
```
smilingkylan, line 88: `category` is a better keyname for the types of results, IMO

app/components/UI/UrlAutocomplete/index.tsx
```diff
@@ -40,25 +55,48 @@ const UrlAutocomplete = forwardRef<
...
+      if (category === 'tokens') {
```
smilingkylan, line 89: We may also want to start using a constant / enum for the types rather than a hard-coded string

app/components/UI/UrlAutocomplete/index.tsx
```diff
@@ -148,21 +204,90 @@ const UrlAutocomplete = forwardRef<
...
+
+  const { addPopularNetwork, networkModal } = useAddNetwork();
+
+  const navigation = useNavigation();
+
+  const handleSwapNavigation = useCallback((result: TokenSearchResult) => {
+    hide();
+    onDismiss();
+    navigation.navigate('Swaps', {
+      screen: 'SwapsAmountView',
```
smilingkylan, line 217: Consider using constants here for `'Swaps` and `SwapsAmountView`

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
```diff
@@ -7,8 +7,15 @@ import {
   TokenSearchResponseItem,
   TokenSearchParams,
 } from '@metamask/token-search-discovery-controller';
+import { Hex } from '@metamask/utils';
+import { selectSupportedSwapTokenAddressesByChainId } from '../../../selectors/tokenSearchDiscoveryDataController';
+import { tokenSearchDiscoveryEnabled } from '../../../selectors/featureFlagController/tokenSearchDiscovery';
 
-const SEARCH_DEBOUNCE_DELAY = 300;
+const SEARCH_DEBOUNCE_DELAY = 50;
```
smilingkylan, line 14: Why have you reduced the debounce here? 0.05 seconds is very quick

app/components/Views/Asset/index.js
```diff
@@ -521,7 +523,7 @@ class Asset extends PureComponent {
     const isAssetAllowed =
       asset.isETH ||
       asset.isNative ||
-      asset.address?.toLowerCase() in this.props.swapsTokens;
+      (isAssetFromSearch(asset) ? this.props.searchDiscoverySwapsTokens?.includes(asset.address?.toLowerCase()) : asset.address?.toLowerCase() in this.props.swapsTokens);
```
smilingkylan, line 526: Consider adding improving the readability of this conditional

smilingkylan (dismissed):
Requested some minor changes. Otherwise PR seems easy to read, so thank you for that

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
ziad-saab: To get a more reactive UI. @smilingkylan I wonder if there's a happy value between 50 and 300 that wouldn't feel too quick to you, while also showing results promptly...

app/components/UI/AssetOverview/AssetOverview.tsx
ziad-saab: These results are coming from an API and through the `assets-controller`. Changing this would entail a bunch of delays to change the `assets-controller` package, get it released, and then upgraded in mobile. Are we able to live with this?

smilingkylan: @ziad-saab Re-reviewing right now

No need to delay. I will convey this concern to the assets team(?) so they can keep it in mind going forward.

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
smilingkylan: @ziad-saab My concern is that 50ms is almost unnoticeable and at that point there is minimal reason for even having a debounce. I say set it to 200ms and if it feels good then go with that. Should still be relatively snappy.

smilingkylan:
> FYI I'm not seeing anything my browser search results when I build your branch. Do I need to enable / disable a feature flag or anything like that?

@ziad-saab Looks like it's most likely a result of the TokenSearchDiscoveryController not having been updated yet.

app/components/hooks/useTokenSearchDiscovery/useTokenSearchDiscovery.ts
ziad-saab: How do you feel about 150? :-)

app/components/UI/UrlAutocomplete/index.test.tsx
```diff
@@ -5,8 +5,105 @@ import { act, fireEvent, screen } from '@testing-library/react-native';
...
+  reset,
+  searchTokens,
+}: {
+  results: TokenSearchResponseItem[];
+  isLoading: boolean;
+  reset: () => void;
+  searchTokens: () => void;
+}) => {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+  const useTSD = require('../../hooks/useTokenSearchDiscovery/useTokenSearchDiscovery');
```
gambinish, line 57: Can we convert this to use an ESM import instead?

darkwing (dismissed):

gambinish (dismissed):
Approving with a small comment. Happy to approve this in a fast follow up PR to not block this feature.

darkwing (dismissed):

smilingkylan (dismissed):
LGTM 👍 with one caveat: I need confirmation that other devs are able to get the HTTP requests for the token search working

darkwing:
Everything loading great for me!

SCR-20250421-nzyz

darkwing (dismissed):

smilingkylan (dismissed):

GuiBibeau (approved):

darkwing (approved):

brianacnguyen (approved):

gambinish (approved):

bfullam (approved):

GustavoRSSilva (approved):

smilingkylan (approved):

PR #13331, opened 2025-02-04
feat: Implement confirm layout

OGPoyraz:
## **Description**

This PR adds flat confirmation layout in re-designed confirmations. 

## **Related issues**

Fixes: Task is currently in draft

## **Manual testing steps**

N/A


app/components/Views/confirmations/Confirm/Confirm.tsx
```diff
@@ -1,40 +1,68 @@
...
     </BottomModal>
   );
 };
 
+const Confirm = () => {
+  const { approvalRequest } = useApprovalRequest();
+  const isFlatConfirmation = FLAT_CONFIRMATIONS.includes(
+    approvalRequest?.type as TransactionType,
+  );
+  const ContentDisplay = isFlatConfirmation ? ScrollView : View;
```
jpuri, line 53: ScrollView may still be useful if content overflow, if there is no overflow - it will anyways behave as normal view.

app/components/Views/confirmations/Confirm/Confirm.tsx
```diff
@@ -1,40 +1,68 @@
...
-  const { isRedesignedEnabled } = useConfirmationRedesignEnabled();
+const FLAT_CONFIRMATIONS: TransactionType[] = [
+  // To be filled with flat confirmations
+];
+
+const ConfirmationLayout = ({ children }: { children: React.ReactNode }) => {
   const { approvalRequest } = useApprovalRequest();
+  const isFlatConfirmation = FLAT_CONFIRMATIONS.includes(
+    approvalRequest?.type as TransactionType,
+  );
```
jpuri, line 27: Rather than using an array like this here can we make it dynamic, may be we check if confirmation is not initiated by DAPP, that ways we will not need not maintain it for new confirmation that we add.

app/components/Views/confirmations/Confirm/Confirm.tsx
jpuri: Ideally this component should not know about modal / fullscreen layout.

app/components/Views/confirmations/Confirm/Confirm.tsx
```diff
@@ -1,40 +1,68 @@
...
-          <SignatureBlockaidBanner />
-          <AccountNetworkInfo />
-          <Info />
-        </View>
-        <Footer />
+      <View style={styles.container}>
+        {children}
       </View>
     </BottomModal>
   );
```
jpuri, line 45: Rather than including `< ConfirmationLayout ` in <Confirm` can we rather include `<Confirm>` in place of `{children}` here.

app/components/Views/confirmations/Confirm/Confirm.styles.ts
```diff
@@ -7,6 +7,17 @@ const styleSheet = (params: { theme: Theme }) => {
...
+      position: 'absolute',
+      top: 0,
+      left: 0,
+      right: 0,
+      bottom: 0,
+      zIndex: 9999,
+      backgroundColor: theme.colors.background.alternative,
+      justifyContent: 'space-between',
+      paddingHorizontal: 16,
+    },
```
jpuri, line 20: Not sure is `fixed` position there in react native ?

OGPoyraz: Unfortunately not 
Screenshot 2025-02-04 at 14 10 28

app/components/Views/confirmations/Confirm/Confirm.tsx
OGPoyraz: Exactly, I was also thinking to capsulate this logic in `ConfirmaionLayout`. However if we want to stick up `Footer` to the bottom (for flat confirmations), I think this is the only viable solution. I am happy to apply any suggestions if you have on this.

app/components/Views/confirmations/Confirm/Confirm.tsx
OGPoyraz: That's definitely something we can do. But my understanding is that flat confirmation support is not tied to origin. Maybe we could visit this in the upcoming implementations?

app/components/Views/confirmations/Confirm/Confirm.tsx
OGPoyraz: Good call, done

app/components/Views/confirmations/Confirm/Confirm.tsx
jpuri: I tied this change and looks like footer is sticky:

Screenshot 2025-02-05 at 6 18 44 PM

Screenshot 2025-02-05 at 6 14 40 PM

jpuri (approved):

PR #13420, opened 2025-02-11
perf: Improve Snaps execution performance

### LINKED ISSUE #2873: Support fully local execution on mobile

FrederikBolding:
## **Description**

Improve Snaps execution performance by bumping the Snaps packages to latest. The latest packages have improved caching for Snap state and changes the Snaps execution on mobile to run each Snap in a separate WebView. This PR adapts `SnapsExecutionWebView` to support this new execution architecture and also attempts to speed up key derivation by using a native implementation of `hmacSha512`. Furthermore, the idle timeout for Snaps is bumped to 5 minutes. 

## **Related issues**

Fixes: https://github.com/MetaMask/snaps/issues/2873

## **Manual testing steps**

Run any Snap and make sure it still works!


FrederikBolding:
Skipping SonarCloud since part of these changes are not possible to test with the current testing framework.

app/lib/snaps/SnapsExecutionWebView.test.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import { createWebView, removeWebView, SnapsExecutionWebView } from './SnapsExecutionWebView';
+
+describe('SnapsExecutionWebView', () => {
+  it('should render correctly', () => {
+    const wrapper = shallow(<SnapsExecutionWebView />);
```
tommasini, line 8: Can we use react-testing-library instead? We do want to deprecate enzyme across the codebase

tommasini (dismissed):
Looks good! Just a comment in the test folder to not use enzyme since we are trying to replace it by react testing library across the codebase

app/lib/snaps/SnapsExecutionWebView.test.tsx
FrederikBolding: As far as I can tell you can't render the WebView component at all with `react-testing-library`. My assertions were all failing 🤔

tommasini: We are snapshot testing, the browser tab [here](https://github.com/MetaMask/metamask-mobile/blob/e046eb636755630921e3f904dd63a546b9ade420/app/components/Views/BrowserTab/index.test.tsx#L76) but yeah you are right we are not testing the webview directly hmm, although it's just a component right?

Mrtenz (dismissed):

Mrtenz (approved):

tommasini (approved):

app/lib/snaps/SnapsExecutionWebView.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  it('should create and remove WebViews correctly', async () => {
+    const wrapper = render(<SnapsExecutionWebView />);
+    createWebView('foo');
+    createWebView('bar');
+    wrapper.rerender(<SnapsExecutionWebView />);
+    expect(await wrapper.queryByTestId('foo')).toBeTruthy();
+    expect(await wrapper.queryByTestId('bar')).toBeTruthy();
+    removeWebView('foo');
+    wrapper.rerender(<SnapsExecutionWebView />);
```
tommasini, line 36: Curious question, why it was needed to rerender the SnapsExecution webview after creation and removal?

FrederikBolding: The create and remove functions force the component to re-render but it doesn't when testing. So we have to force the re-render to check that the views were removed/added

gantunesr (approved):

FrederikBolding:
> I'll leave this as a suggestion since I don't see it as a blocker.
> 
> since you can't do integration tests between the js side and the native side on react native apps, we have developed an e2e screen exclusively for cryptographic methods to ensure their correctness. You can check `e2e/specs/accounts/aes` to follow the same pattern in a follow up PR and remove `@noble/hashes` since from what I see is just being used to mock.

AFAIK the key derivation is already tested using E2E, so that should provide confidence that adding the native code hasn't changed anything!

PR #13462, opened 2025-02-12
fix: label main and secondary balances

bergarces:
## **Description**

It's been reported that setting the primary currency to "Native" was displaying fiat as the main currency and the native currency as secondary. The opposite would happen when setting the primary currency to "Fiat".

This was due to using the wrong nomenclature as the properties of the component responsible for that behaviour (`balance` was the primary balance and `mainBalance` was the secondary balance).

The chosen solution to fix nomenclature is leaving `balance` as the primary balance changing the name of `mainBalance` to `secondaryBalance`.

This component is used in three different places:
- Token list items in the home wallet.
- Balance in the asset overview.
- Staking balance in the asset overview.

The behaviour for the token list and balance have been fixed to work as the setting suggests. The staking balance has been left as it currently is pending confirmation of whether it needs to follow the same behaviour.

## **Related issues**

Fixes: No open GitHub issue for the ticket

## **Manual testing steps**

1. Go to Settings -> General -> Primary Currency -> Set to "Native"
2. Got to Wallet Home page and check assets in the Tokens tab display the native currency as primary and fiat as secondary.
3. Click on any of the assets to open the asset overview and check that the balance is displayed correctly.
4. Repeat steps 1, 2 and 3 changing the primary currency to "Fiat"

### **Before**

image
image

### **After**

Screenshot_1739377743
Screenshot_1739377733


app/components/UI/AssetOverview/Balance/Balance.tsx
```diff
@@ -144,7 +144,7 @@ const Balance = ({ asset, mainBalance, secondaryBalance }: BalanceProps) => {
       </Text>
       <AssetElement
         asset={asset}
-        mainBalance={mainBalance}
+        secondaryBalance={mainBalance}
         balance={secondaryBalance}
```
bergarces, line 148: The ticket does not specify that we ought to change the behaviour of this component, so I haven't.

However, given that this component is @MetaMask/metamask-assets responsibility, we can (and probably should) change its behaviour.

app/components/UI/AssetElement/index.constants.ts
```diff
@@ -1,2 +1,2 @@
-export const FIAT_BALANCE_TEST_ID = 'fiat-balance-test-id';
-export const MAIN_BALANCE_TEST_ID = 'main-balance-test-id';
```
bergarces, line 2: Renaming these test ids for consistency.

app/components/UI/NonEvmTokens/index.test.tsx
```diff
@@ -101,6 +101,7 @@ const initialState = {
   },
   settings: {
     showTestNetworks: true,
+    primaryCurrency: 'ETH',
```
bergarces, line 104: Need this to maintain test descriptions and assertions accurate, as otherwise the default is fiat and the `TokenListItem` component has now been fixed.

app/components/UI/Stake/components/StakingBalance/StakingBalance.tsx
```diff
@@ -207,7 +207,7 @@ const StakingBalanceContent = ({ asset }: StakingBalanceProps) => {
       {hasEthToUnstake && !isLoadingPooledStakesData && (
         <AssetElement
           asset={asset}
-          mainBalance={stakedBalanceETH}
+          secondaryBalance={stakedBalanceETH}
           balance={stakedBalanceFiat}
```
bergarces, line 211: This component is owned by @MetaMask/metamask-staking and is currently displaying fiat balance as the primary and ETH balance as secondary without taking into account the `primaryCurrency` setting.

Since this is a bug fix for another component, I didn't want to change the behaviour of this one, but it's probably worth checking what it should be.

app/components/UI/AssetOverview/Balance/Balance.tsx
```diff
@@ -144,8 +144,8 @@ const Balance = ({ asset, mainBalance, secondaryBalance }: BalanceProps) => {
       </Text>
       <AssetElement
         asset={asset}
-        mainBalance={mainBalance}
-        balance={secondaryBalance}
```
bergarces, line 148: The ticket does not specify that we ought to change the behaviour of this component, but I have, given that the same issue applied here.

The way it determines mainBalance and secondaryBalance looks a bit odd ([here](https://github.com/MetaMask/metamask-mobile/blob/7e608986a116b97492335b798288b1b31a187878/app/components/UI/AssetOverview/AssetOverview.tsx#L378-L392))

If portfolio view is enabled, it always sets mainBalance as the chain's native currency and secondaryBalance as the fiat.
If portfolio view is NOT enabled, then it sets the mainBalance and secondaryBalance in accordance to the primaryCurrency setting.
They were however being passed the wrong way around, which meant that secondaryBalance was being displayed as the main balance and mainBalance was being displayed as the secondary balance. It should now display as intended.

Prithpal-Sooriya (dismissed):

app/components/UI/AssetOverview/Balance/Balance.tsx
bergarces: After discussing the issue. `mainBalance` and `secondaryBalance` are now being calculated taking the toggle into consideration.

app/components/UI/AssetOverview/AssetOverview.test.tsx
```diff
@@ -46,17 +46,16 @@ const mockInitialState = {
       AccountTrackerController: {
         accountsByChainId: {
           [MOCK_CHAIN_ID]: {
-            [MOCK_ADDRESS_2]: { balance: '0x1' },
+            [MOCK_ADDRESS_2]: { balance: '0xDE0B6B3A7640000' },
           },
         },
       } as const,
-    },
-    CurrencyRateController: {
```
bergarces, line 54: This was outside the object, so it wasn't being used at all as state.

app/components/UI/AssetOverview/AssetOverview.test.tsx
```diff
@@ -338,4 +346,47 @@ describe('AssetOverview', () => {
...
+        },
+      },
+    );
+
+    const balance = getByTestId('balance-test-id');
+    const secondaryBalance = getByTestId('secondary-balance-test-id');
+
+    expect(balance.props.children).toEqual('$1000');
+    expect(secondaryBalance.props.children).toEqual('1 ETH');
+  });
```
bergarces, line 391: Added tests to verify new behaviour when using the toggle.

app/components/UI/AssetOverview/AssetOverview.tsx
```diff
@@ -376,19 +376,15 @@ const AssetOverview: React.FC<AssetOverviewProps> = ({
...
+  }`;
+  if (primaryCurrency === 'ETH' || !balanceFiat) {
+    mainBalance = formattedAssetBalance;
+    secondaryBalance = balanceFiat;
   } else {
-    mainBalance = `${balance} ${asset.isETH ? asset.ticker : asset.symbol}`;
-    secondaryBalance = asset.balanceFiat || '';
+    mainBalance = balanceFiat;
+    secondaryBalance = formattedAssetBalance;
   }
```
bergarces, line 388: As per the internal conversations, `isPortfolioViewEnabled()` should not have any relevance on whether the toggle works or not. This simplifies the logic.

Prithpal-Sooriya (dismissed):

Prithpal-Sooriya (approved):

app/components/UI/Stake/components/StakingBalance/StakingBalance.tsx
amitabh94: Please change this to match the behavior of the fix.

nickewansmith: @bergarces Should the balances flip based on `primaryCurrency ` on the Asset Detail page when portfolio view is enabled? Currently they do not.

bergarces: @nickewansmith the settings toggle should be respected regardless of portfolio view being or not enabled.

However, if the `StakingBalance` component is also going to use the same logic, a more comprehensive refactor would be better suited. So I've gone ahead with your suggestion, created a PR to tackle only the bug in the tokens list and not modify any behaviour for the asset detail page, which includes both the asset and staking balance.

https://github.com/MetaMask/metamask-mobile/pull/13916

I have also created a different ticket on our backlog for the asset detail page.

PR #13495, opened 2025-02-13
feat(ramps): adds most reliable tag to ramp quotes

georgeweiler:
## **Description**

This PR adds a "most reliable" tag to quotes and reduces the quotes in the modal to a single recommended quote based on previously used status, reliability and price. Most reliable tags will only appear for USD/Debit-credit requests initially.

- adds `useSortedQuotes` hook that returns various sorted and filtered versions of the raw quote response that is fetched from the ramp API
- removes filtering and sorting logic from the Quotes.tsx view
- adds `most reliable` tag to the deserving quote

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### **Before**

Screenshot 2025-02-12 at 11 04 30 AM

### **After**

simulator_screenshot_0E7CF75F-8633-4A0B-8459-81CECBE609FB


app/components/UI/Ramp/Views/Quotes/Quotes.tsx
```diff
@@ -558,52 +530,55 @@ function Quotes() {
...
-      setProviderId(filteredQuotes[0].provider?.id);
+    if (
+      quotesByPriceWithoutError &&
+      quotesByPriceWithoutError.length > 0 &&
+      isExpanded
+    ) {
+      setProviderId(quotesByPriceWithoutError[0].provider?.id);
     }
-  }, [filteredQuotes]);
+  }, [isExpanded, quotesByPriceWithoutError]);
```
wachunei, line 581: I understand we set the selected quote to be the first when the user expands the view, but what about when the user does not? We should have something like `setProviderId(recommendedQuote.provider.id)` instead of the `highlighed` hardcoded to true in https://github.com/MetaMask/metamask-mobile/pull/13495/files#diff-e6b63f4cb10a53fd622b8afeca6c8c8523e9841c1fe6006449bf5ef97f266291R754

wachunei:
Can you share an screenshot of the expanded quotes before and after, before we would highlight 2 quotes (as this const `HIGHLIGHTED_QUOTES_COUNT` has that number set). Now that we are reordering by price I don't think we should have the 2 highlighted quotes and the title in that position:
```tsx
                    {index === HIGHLIGHTED_QUOTES_COUNT &&
                      quotesByPriceWithoutError.length > 1 && (
                        <Row>
                          <Text variant={TextVariant.BodyLGMedium}>
                            {quotesByPriceWithoutError.length === 2
                              ? strings(
                                  'fiat_on_ramp_aggregator.one_more_option',
                                )
                              : strings(
                                  'fiat_on_ramp_aggregator.more_options',
                                  {
                                    count: quotesByPriceWithoutError.length - 1,
                                  },
                                )}
                          </Text>
                        </Row>
                      )}```

File: app/components/UI/Ramp/Views/Quotes/Quotes.tsx [hunk 2]
```diff
@@ -558,52 +518,55 @@ function Quotes() {
...
-            trackEvent('OFFRAMP_QUOTE_ERROR', {
-              ...payload,
-              currency_destination: params.fiatCurrency?.symbol,
-              currency_source: params.asset?.symbol,
-              provider_offramp: quoteError.provider.name,
-              chain_id_source: selectedChainId,
-            });
-          }
-        });
+      quotesWithError.forEach((quoteError) => {
```
wachunei, line 521: We used to run this `useEffect` with this `if` condition: 
```tsx
    if (quotes && !isFetchingQuotes && pollingCyclesLeft >= 0) {
```

https://github.com/MetaMask/metamask-mobile/pull/13495/files#diff-e6b63f4cb10a53fd622b8afeca6c8c8523e9841c1fe6006449bf5ef97f266291L472, because `quotes` contained both error and non-error quotes.

My concern here is that `quotesWithError` might not be defined since the new condition is only checking `quotesWithoutError`:
```tsx
if (quotesWithoutError && !isFetchingQuotes && pollingCyclesLeft >= 0) {
```

I suggest we change it to 
```tsx
if (quotesWithoutError && quotesWithError && !isFetchingQuotes && pollingCyclesLeft >= 0) {
```
so we do the event dispatching when both are defined*.

\*Currently the hook would make both `quotesWithoutError` and `quotesWithError` defined at the same time, but this is just in case the hook changes in the future.

app/components/UI/Ramp/Views/Quotes/Quotes.tsx
```diff
@@ -766,30 +729,28 @@ function Quotes() {
           <ScrollView testID={QuoteSelectors.QUOTES}>
             {isFetchingQuotes && isInPolling ? (
               <LoadingQuotes count={2} />
```
wachunei, line 731: Loading count must be 1 if we switched over to recommended quote.

wachunei (approved):

PR #13534, opened 2025-02-14
feat: Limit Segment tracked events to only random 1% of users

smilingkylan:
## **Description**

The purpose of this task is to reduce the instances of events being logged to Segment to only 1% of the current volume. The solution is to randomly select 1% of USERS (NOT EVENTS) to be tracked, and ignore the other 99%.

## **Related issues**

Fixes: [#4119](https://github.com/MetaMask/MetaMask-planning/issues/4119)
Core PR: https://github.com/MetaMask/core/pull/5341

## **Manual testing steps**

To manually test you can opt-in to being tracked via the account creation user flow, then check MixPanel to see if the events were logged or not.

### **Before**

n/a 100% of Segment events are currently being tracked

### **After**

To test, I changed the % to **50%**, went through the onboarding flow 4x and saw that exactly 2 of the 4 (50%) were logged to Segment / MixPanel

Users: 
```
1a2edf78-0571-4144-8ed7-63e3de6631a3
cf49c571-99e2-4333-819d-46b14fc056a2
5b0d2ba1-9a2b-4933-afc6-2eeb2eba123c
daa3eae4-3e3b-4d08-adda-b5969e5d0685
```

image


app/core/Analytics/MetaMetrics.ts
```diff
@@ -37,6 +37,11 @@ import generateDeviceAnalyticsMetaData from '../../util/metrics/DeviceAnalyticsM
 import generateUserSettingsAnalyticsMetaData from '../../util/metrics/UserSettingsAnalyticsMetaData/generateUserProfileAnalyticsMetaData';
 import { isE2E } from '../../util/test/utils';
 import MetaMetricsPrivacySegmentPlugin from './MetaMetricsPrivacySegmentPlugin';
+import { generateDeterministicRandomNumber } from '@metamask/remote-feature-flag-controller/dist/utils/user-segmentation-utils.cjs';
```
smilingkylan, line 40: Is this the correct way to import from the `remote-feature-flag-controller`?

DDDDDanica (dismissed):

package.json
```diff
@@ -196,7 +196,7 @@
     "@metamask/react-native-payments": "^2.0.0",
     "@metamask/react-native-search-api": "1.0.1",
     "@metamask/react-native-webview": "^14.0.4",
-    "@metamask/remote-feature-flag-controller": "^1.3.0",
+    "@metamask/remote-feature-flag-controller": "^1.5.0",
```
tommasini, line 199: Since we bumped remotr-feature-flag-comtroller, now it's using base controller v8 in it's own dependencies. In the engine file where we initialize it, we can remove the ts-expect-error!

Awesome work 🍀

DDDDDanica (dismissed):

app/core/Analytics/MetaMetrics.ts
```diff
@@ -663,6 +671,16 @@ class MetaMetrics implements IMetaMetrics {
       return;
     }
 
+    // generate a random number based on the metaMetricsId
+    const metaMetricsIdRandomNumber = generateDeterministicRandomNumber(
```
tommasini, line 675: This should be around a try catch, if it falls under the emtpy string an error will be thrown and we are not handling it , wdyt?

tommasini: Or Instead of the try catch, do not call `generateDeterministicRandomNumber ` if metametricsId doesn+t exist and return undefined since it's a possible scenario already

DDDDDanica: Good point, there is actually empty case prevention in `generateDeterministicRandomNumber ` check [here](https://github.com/MetaMask/core/blob/main/packages/remote-feature-flag-controller/src/utils/user-segmentation-utils.ts#L34)

DDDDDanica (approved):

tommasini (approved):

PR #13598, opened 2025-02-18
fix: cp-7.41.0 Handle better errors by getting tags in pre-init

### LINKED ISSUE #12289: [Sentry] TypeError: undefined is not an object (evaluating 'n.flatMap')
Sentry Issue: [METAMASK-MOBILE-31ZW](https://metamask.sentry.io/issues/6044922213/?referrer=github_integration)

```
TypeError: undefined is not an object (evaluating 'n.flatMap')
  at None (app:///main.jsbundle:4661:386)
  at E (app:///main.jsbundle:2878:564)
  at None (app:///main.jsbundle:2877:1041)
  at E (app:///main.jsbundle:2878:564)
  at None (app:///main.jsbundle:2877:1007)
...
(16 additional frame(s) were not displayed)
```

tommasini:
## **Description**

Handle errors in a more scalable way by adding a try catch to get tags function. Also added test coverage

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/util/sentry/tags/index.ts
```diff
@@ -5,44 +5,50 @@ import { selectAllTokensFlat } from '../../../selectors/tokensController';
...
+      'wallet.nft_count': nftCount,
+      'wallet.notification_count': notificationCount,
+      'wallet.pending_approval': firstApprovalType,
+      'wallet.token_count': tokenCount,
+      'wallet.transaction_count': transactionCount,
+      'wallet.unlocked': unlocked,
+    };
+  } catch (error) {
+    captureException(`Error getting tags: ${error}`);
+    return null;
```
Gudahtt, line 52: It looks like this function returns tags if they're found, or `undefined` if they're not. Perhaps we can simplify this and not introduce a third return type? We should be able to use `undefined` here as well

app/util/sentry/tags/index.ts
```diff
@@ -5,44 +5,50 @@ import { selectAllTokensFlat } from '../../../selectors/tokensController';
...
-  )
-    return;
-  if (!state?.engine?.backgroundState?.TokensController) return;
-  if (!state?.engine?.backgroundState?.TokensController?.allTokens) return;
-  if (!state?.engine?.backgroundState?.TransactionController) return;
-  if (!state?.engine?.backgroundState?.TransactionController?.transactions)
-    return;
-  if (!state?.engine?.backgroundState?.ApprovalController) return;
+  try {
+    if (!state?.engine?.backgroundState?.AccountsController) return;
```
Gudahtt, line 12: Nit: These early exits can be removed now as well, they were intended to guard against the same crashes we're now catching more broadly.

app/util/sentry/tags/index.ts
```diff
@@ -5,44 +5,50 @@ import { selectAllTokensFlat } from '../../../selectors/tokensController';
...
+      'wallet.account_count': accountCount,
+      'wallet.nft_count': nftCount,
+      'wallet.notification_count': notificationCount,
+      'wallet.pending_approval': firstApprovalType,
+      'wallet.token_count': tokenCount,
+      'wallet.transaction_count': transactionCount,
+      'wallet.unlocked': unlocked,
+    };
+  } catch (error) {
+    captureException(`Error getting tags: ${error}`);
```
Gudahtt, line 51: Note that in this case the failure is expected and not indicative of any problem, so I'm not sure we need to log anything to Sentry or Segment or anywhere. It's normal for this to be called pre-initialization.

Gudahtt (commented):
Left a few suggestions.

Also, this broader `try..catch` will definitely solve the problem, but we could also consider splitting it up by selector or chunks of related selector. e.g. if `unlocked` is retrieved in a separate `try..catch` block, we might be able to retrieve that even before Engine initialization.

app/util/sentry/tags/index.ts
tommasini: It's true! Well noted. I assumed that it would be a problem, but it wouldn't because, at this point in time, the data couldn't just be retrieved yet

tommasini:
@Gudahtt Kept the optional chaining, but based on your comment I think this is in a much better state! Let me know your thoughts!

app/util/sentry/tags/index.ts
tommasini: Addressed this commit: https://github.com/MetaMask/metamask-mobile/pull/13598/commits/29a8a15c384bfe583066e0d03339c7fa02952440

MarioAslau (approved):

PR #13634, opened 2025-02-20
feat: Implement `GasFeeDetails`

OGPoyraz:
## **Description**

This PR aims to implement `GasFeeDetails` component into staking deposit (component will be widely used in later in redesigned transaction confirmations)

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4215

## **Manual testing steps**

N/A

### **After**

Screenshot 2025-02-20 at 13 50 40


app/components/Views/confirmations/hooks/useFeeCalculations.ts
```diff
@@ -0,0 +1,136 @@
...
+      if (decimalCurrentCurrencyFee === 0) {
+        currentCurrencyFee = `< ${fiatFormatter(new BigNumber(0.01))}`;
+        currentCurrencyFeeWith18SignificantDigits = getValueFromWeiHex({
+          value: hexFee,
+          conversionRate: nativeConversionRateInBN,
+          fromCurrency: 'GWEI',
+          toCurrency: 'ETH',
+          numberOfDecimals: 18,
+          toDenomination: 'ETH',
+        });
```
pedronfigueiredo, line 76: When decimalCurrentCurrencyFee is 0, currentCurrencyFee should be `$0.00`, not `< $0.01`. But when 0 < decimalCurrentCurrencyFee < 0.01, currentCurrencyFee should be `< $0.01`.

app/components/Views/confirmations/components/Confirm/Info/GasFeesDetails/GasFeesDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+import InfoSection from '../../../UI/InfoRow/InfoSection';
+import styleSheet from './GasFeesDetails.styles';
+
+const GasFeesDetails = () => {
+  const { styles } = useStyles(styleSheet, {});
+  const transactionMetadata = useTransactionMetadataRequest();
+  const feeCalculations = useFeeCalculations(
+    transactionMetadata as TransactionMeta,
+  );
+
```
pedronfigueiredo, line 19: We should hide the secondaryValue fiat display on testnets if this option is set in settings, or in any case if that setting doesn't exist on mobile.

See extension for reference: https://github.com/MetaMask/metamask-extension/blob/dc2249c479222a3c57d41c11a222910c2fe9dd85/ui/pages/confirmations/components/confirm/info/shared/gas-fees-row/gas-fees-row.tsx#L40-L44

app/components/Views/confirmations/hooks/useFeeCalculations.ts
OGPoyraz: I thought this was intentionally implemented like this. In theory there is no way it's `$0.00`, so because of rounding it's really low and we show it as less than a cent no?

app/components/Views/confirmations/hooks/useFeeCalculations.ts
```diff
@@ -0,0 +1,136 @@
...
+          toCurrency: 'ETH',
+          numberOfDecimals: 2,
+          toDenomination: 'ETH',
+        }),
+      );
+
+      let currentCurrencyFee, currentCurrencyFeeWith18SignificantDigits;
+      if (decimalCurrentCurrencyFee === 0) {
+        currentCurrencyFee = `< ${fiatFormatter(new BigNumber(0.01))}`;
+        currentCurrencyFeeWith18SignificantDigits = getValueFromWeiHex({
```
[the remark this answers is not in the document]
OGPoyraz: Removed

app/components/Views/confirmations/components/Confirm/Info/GasFeesDetails/GasFeesDetails.tsx
OGPoyraz: Done - also added tests

app/components/Views/confirmations/hooks/useFeeCalculations.ts
pedronfigueiredo: If decimalCurrentCurrencyFee is 0, either the fiat cost is 0 (an example of this are testnets, which we hide in any case), or we don't receive a valid fiat conversion rate from the API. In that second case it is incorrect to say the fee is `< $0.01`. I think if the conversion rate is 0 it would be preferrable to either default to $0.00 or show a different notice `N/A` (for not available). Can we double check this case with design / product?

OGPoyraz: Good point on fiat conversion rate, I will validate with Barbara and Saya

OGPoyraz: Done - added all the cases above including hidden fiat value when conversion rate is not available.

pedronfigueiredo (approved):

PR #13711, opened 2025-02-25
fix: Revert native HMACSHA512 usage

FrederikBolding:
## **Description**

Revert part of change from https://github.com/MetaMask/metamask-mobile/commit/a9cd3aff74dfb980ecb8a10205bd79c4f577b8d0 which caused certain key derivation to be wrong when using the Snaps API. I assumed this was E2E tested, but the specific case that failed was not.

## **Manual testing steps**

1. Generate a Solana account
2. Compare the address to the one on extension


ccharly (approved):
Not tested, but the code LGTM.

tommasini (approved):

PR #13712, opened 2025-02-25
feat: Implement Snaps address UI components

FrederikBolding:
## **Description**

Implement `SnapUIAvatar` and `SnapUIAddress`, also fixes issues rendering icons that come from Snaps and text formatting in nested `Text` components. Also corrects the styling for the `SnapUIButton`.

Note: I've tried to keep this PR as small as possible for reviewability. Most of the diff is snapshots that doesn't need thorough review.

## **Manual testing steps**

```ts
import type { OnRpcRequestHandler } from '@metamask/snaps-sdk';
import { Box, Text, Bold, Address, Avatar, Spinner, Button, Icon, IconName } from '@metamask/snaps-sdk/jsx';

/**
 * Handle incoming JSON-RPC requests, sent through `wallet_invokeSnap`.
 *
 * @param args - The request handler args as object.
 * @param args.origin - The origin of the request, e.g., the website that
 * invoked the snap.
 * @param args.request - A validated JSON-RPC request object.
 * @returns The result of `snap_dialog`.
 * @throws If the request method is not valid for this snap.
 */
export const onRpcRequest: OnRpcRequestHandler = async ({
  origin,
  request,
}) => {
  switch (request.method) {
    case 'hello':
      return snap.request({
        method: 'snap_dialog',
        params: {
          type: 'confirmation',
          content: (
            <Box>
              <Text>
                Hello, <Bold>{origin}</Bold>!
              </Text>
              <Address address='eip155:1:0x1234567890123456789012345678901234567890' />
              <Avatar address="eip155:1:0x1234567890123456789012345678901234567890" />
              <Address address="bip122:000000000019d6689c085ae165831e93:128Lkh3S7CkDTBZ8W7BbpsN3YYizJMp8p6" />
              <Avatar address="bip122:000000000019d6689c085ae165831e93:128Lkh3S7CkDTBZ8W7BbpsN3YYizJMp8p6" />
              <Box direction="horizontal" alignment="space-between" center={true}>
                <Text>Avatar</Text>
                <Avatar address="bip122:000000000019d6689c085ae165831e93:128Lkh3S7CkDTBZ8W7BbpsN3YYizJMp8p6" />
              </Box>
              <Button>Heya!</Button>
              <Button size="sm">Small</Button>
              <Button><Icon name="swap-vertical" color="primary" size="md" /></Button>
            </Box>
          ),
        },
      });
    default:
      throw new Error('Method not found.');
  }
};

```

## **Screenshots/Recordings**

Screenshot 2025-02-26 at 10 29 35


app/components/Snaps/SnapUIRenderer/components/footer.test.ts
```diff
@@ -65,7 +65,7 @@ describe('footer', () => {
           },
           children: [
             {
-              element: 'Text',
+              element: 'RNText',
```
GuillaumeRx, line 68: What's `RNText` ? :thinking:

app/components/Snaps/SnapUIRenderer/components/icon.ts
```diff
@@ -11,9 +11,15 @@ const ICON_NAMES = new Set(Object.values(IconName));
...
-      return element.props.name as IconName;
+    const rawName = element.props.name;
+    // The icon names are formatted differently between extension and mobile,
+    // so we attempt to map from extension to the mobile format here.
+    const mappedName = rawName
+      .split('-')
+      .map((str) => str.charAt(0).toUpperCase() + str.slice(1))
+      .join('') as IconName;
+    if (ICON_NAMES.has(mappedName)) {
+      return mappedName;
```
GuillaumeRx, line 22: :dizzy_face:

app/components/Snaps/SnapUIRenderer/components/footer.test.ts
FrederikBolding: The Text component from `react-native`, renamed with no styling at all: https://github.com/MetaMask/metamask-mobile/pull/13712/files#diff-2e9c2b5fcfb7c46cffb10380c267699b74b585c6c0bef6d0c169a07165e161f9R7

Makes it easier to inherit styling from parent Text, the DS component overrides a bunch of stuff

app/components/Snaps/SnapUIRenderer/components/icon.ts
FrederikBolding: Yeah, this is an unfortunate difference in the DS naming that we have to deal with 🫠

app/components/UI/Snaps/SnapUIAddress/SnapUIAddress.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { useMemo } from 'react';
+import {
+  CaipAccountId,
+  isHexString,
+  parseCaipAccountId,
+} from '@metamask/utils';
+import { toChecksumHexAddress } from '@metamask/controller-utils';
+import { SnapUIAvatar } from '../SnapUIAvatar/SnapUIAvatar';
+import { shortenString } from '../../../../util/notifications/methods';
```
tommasini, line 9: Can we create a util method for snaps? This will break code ownership, if Notifications change this util method, will break SnapUIAddress unexpectadly 

We can also add this util method in a general util file instead

app/components/UI/Snaps/SnapUIInput/SnapUIInput.tsx
```diff
@@ -70,7 +70,7 @@ export const SnapUIInput = ({
         onChangeText={handleChange}
       />
       {error && (
-                // eslint-disable-next-line react-native/no-inline-styles
+        // eslint-disable-next-line react-native/no-inline-styles
         <HelpText severity={HelpTextSeverity.Error} style={{ marginTop: 4 }}>
```
tommasini, line 74: NIT: Can we move this to separate file to keep the existing pattern?

app/components/UI/Snaps/SnapUIAddress/SnapUIAddress.tsx
FrederikBolding: It seems that more components are relying on this implementation. I agree it should be moved, but seems out of scope for this PR. If the underlying implementation changes our snapshots would too and would require a CODEOWNER review for that

app/components/UI/Snaps/SnapUIAddress/useDisplayName.ts
```diff
@@ -0,0 +1,59 @@
...
+  const accounts = useSelector(selectInternalAccounts);
+
+  const account = accounts.find((possibleAccount) =>
+    toLowerCaseEquals(possibleAccount.address, address),
+  );
+
+  const chainAddressBook = useSelector((state: RootState) =>
+    selectAddressBookByChain(
+      state,
+      `0x${decimalToHex(isEip155 ? reference : `0`)}`,
```
tommasini, line 48: Can we live a comment that this is only supporting evm for now? We can also do it raise this later, feel free to ignore. Just probably easier when the Address Book Controller supports other chains

tommasini (approved):

app/components/UI/Snaps/SnapUIAddress/SnapUIAddress.tsx
tommasini: Nice! I completely missed that the snapshot was preventing it! Thanks <3

david0xd (approved):

PR #13715, opened 2025-02-25
chore: Revert "feat: Limit Segment tracked events to only random 1% of users"

NicolasMassart:
Reverts MetaMask/metamask-mobile#13534 as it applies to all events instead of only error events.

## Reason
PR not doing what [the original issue](https://github.com/MetaMask/MetaMask-planning/issues/4119) expected as raised in [Slack thread](https://consensys.slack.com/archives/C087GPTR5HQ/p1740489998729699?thread_ts=1740470413.754229&cid=C087GPTR5HQ). It should only sample errors, not all events.

## Unrelated changes
The original PR made change son engine.ts and remote feature flag controller that were not related to the Metrics sampling. They should not have been in the PR so I removed these changes from this revert.

### DIFF
--- a/app/core/Engine/Engine.ts
+++ b/app/core/Engine/Engine.ts
@@ -592,7 +592,6 @@ export class Engine {
 
     const remoteFeatureFlagController = createRemoteFeatureFlagController({
       state: initialState.RemoteFeatureFlagController,
-      // @ts-expect-error TODO: Resolve mismatch between base-controller versions.
       messenger: this.controllerMessenger.getRestricted({
         name: 'RemoteFeatureFlagController',
         allowedActions: [],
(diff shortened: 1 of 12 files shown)


app/core/Engine/Engine.ts
NicolasMassart, line 1: This should not be part of the revert. It should not have been in the same PR initially.

yarn.lock
NicolasMassart, line 1: This should not be part of the revert. It should not have been in the same PR initially.

package.json
NicolasMassart, line 1: This should not be part of the revert. It should not have been in the same PR initially.

NicolasMassart (commented):
Revert metametrics sampling but not the engine and remote feature flag upgrade. It should not have been in the same PR initially.

tommasini (approved):

PR #13803, opened 2025-03-03
fix: cp-7.42.0 improve performance of notifications loading when opening the app

Prithpal-Sooriya:
## **Description**

There is serious performance impact of loading notifications on startup. It is roughly a 2 second lag, and the app freezes during this process.
This is due to us processing the encryption key to be used for decrypting entries in UserStorage.
This persists the a cached encryption key using existing storage mechanisms to securely store this hashed key.

**NOTE:** This fix is an interim solution until we add e2e content keys for encryption (CC @mathieuartu @mirceanis)
https://github.com/MetaMask/core/pull/5129

## **Related issues**

Fixes:

## **Manual testing steps**

1. Enable Notifications
2. Close the app
3. Open the app and unlock the wallet
EXPECTED - you should not see any lag or app freezing.

### **Before**

https://www.loom.com/share/8951a714b9b34f79a3e988ed50f9f656?sid=ad2ff732-c893-49d6-bd4b-e7f13a3a08b5

### **After**

https://www.loom.com/share/fe0afadcd49b46bbbafe219f95777fb6?sid=ca0df6aa-b5c5-4ffd-a5ee-bc811f0f1058


app/core/Engine/controllers/identity/calculate-scrypt-key.ts
```diff
@@ -0,0 +1,75 @@
+import { sha256 } from '@noble/hashes/sha256';
+import { bytesToHex } from '@noble/hashes/utils';
+import { scrypt } from 'react-native-fast-crypto';
+import storageWrapper from '../../../../store/storage-wrapper';
+import { SCRYPT_COMPUTED_KEY } from '../../../../constants/storage';
+import Logger from '../../../../util/Logger';
+
+export const generateKeyHash = (
```
Prithpal-Sooriya, line 8: Hashed the inputs, since I don't want anyone to reverse engineer this if the storage gets leaked.

mathieuartu (dismissed):
Thanks for all the context you added in your comments! LGTM!

Also, approving since there may be not enough time to get this directly in the controllers for this release, but the contents of this PR will be replaced soon by an implementation directly in `UserStorageController`

app/core/Engine/controllers/identity/calculate-scrypt-key.ts
tommasini: This is using `@noble/hashes/sha256` that is a javascript encryption library, in order to get the full performance, we should, use a native approach! 

I wonder if using react `react-native-quick-crypto` https://github.com/margelo/react-native-quick-crypto?tab=readme-ov-file#usage would be better?

tommasini: We are using it in snaps as well (the javascript library), although, we will benchmark it and I believe it will be replaced by this as well.
I believe we will always be more protected using native approaches than javascript ones, but that's a personal belief aligned with some past experience of issues in this app :p

Prithpal-Sooriya: Great callout, we have replaced this `react-native-quick-crypto` 👍🏾

tommasini (approved):

mirceanis (commented):
I like the idea of a proxy implementation that caches the `scrypt` output, however I don't think that using a single cache slot is useful enough.
Since we're already comparing cacheHash to ensure we're returning the right value, we might as well use that as the key to that value instead of the `SCRYPT_COMPUTED_KEY` slot.

PR #13817, opened 2025-03-04
fix: Migration of `TransactionController` to modularised controller init

OGPoyraz:
## **Description**

Migrates `TransactionController` into new modularised controller initialisation. 

In order to initialise dependent controller initialisation, controller orders changed in the `Engine.ts` file. Some more information about this on the [Slack thread here](https://consensys.slack.com/archives/C02U025CVU4/p1740168363987169).

Adds following folders to code-owner of Confirmations team
- `app/core/Engine/controllers/transaction-controller`             
- `app/core/Engine/messengers/transaction-controller-messenger` 

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4331

## **Manual testing steps**

No changes in terms of UIX.


app/core/Engine/controllers/transaction-controller/transaction-controller-init.test.ts
```diff
@@ -0,0 +1,288 @@
...
+      });
+      const requestMock = buildInitRequestMock();
+
+      expect(() => TransactionControllerInit(requestMock)).toThrow(
+        'Controller initialisation failed',
+      );
+    });
+  });
+
+  it.each([
```
matthewwalsh0, line 160: Are unit tests that assert boolean properties excessive? Are we just introducing fragility?

app/core/Engine/messengers/transaction-controller-messenger/transaction-controller-messenger.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+}
+
+export function getTransactionControllerInitMessenger(
+  messenger: Messenger<MessengerActions, MessengerEvents>,
+) {
+  return messenger.getRestricted({
+    name: 'TransactionControllerInit',
+    allowedEvents: [
+      'TransactionController:transactionApproved',
```
matthewwalsh0, line 72: Do we use all these already in mobile? I appreciate we want to align with extension but we don't want these messenger types to be as minimal as possible.

app/core/Engine/types.ts
```diff
@@ -513,22 +519,14 @@ type ControllerPersistedState = Partial<{
   >]: Partial<ControllerByName[Name]['state']>;
 }>;
 
-/**
- * Map of controller messengers by controller name.
- */
-export type ControllerMessengerByControllerName = {
```
matthewwalsh0, line 519: Are we safe to remove this? Could we just assign it to `typeof CONTROLLER_MESSENGERS` if we want it to match?

app/core/Engine/types.ts
```diff
@@ -546,6 +544,24 @@ export type ControllerInitRequest<
     name: Name,
   ): ControllerByName[Name];
 
+  /**
+   * Get the chain id set in the engine.
+   *
+   * @deprecated Will be removed in the future pending multi-chain support.
+   */
+  getCurrentChainId: () => Hex;
```
matthewwalsh0, line 552: Can we call this `getGlobalChainId` to be explicit as "current" is ambiguous?

app/core/Engine/types.ts
```diff
@@ -546,6 +544,24 @@ export type ControllerInitRequest<
...
+
+  /**
+   * Get the root state of UI.
```
matthewwalsh0, line 555: Can we include an example of the state as we do in extension? Is it flattened like in extension? Can we explain how it differs to `persistedState`? Should this be called `getUIState`?

app/core/Engine/types.ts
```diff
@@ -579,9 +600,11 @@ type ControllerInitFunctionByControllerName = {
  */
 export type InitModularizedControllersFunction = (request: {
   controllerInitFunctions: ControllerInitFunctionByControllerName;
-  persistedState: ControllerPersistedState;
-  existingControllersByName?: Partial<ControllerByName>;
   baseControllerMessenger: BaseControllerMessenger;
+  existingControllersByName?: Partial<ControllerByName>;
```
matthewwalsh0, line 604: Minor, alphabetical order?

app/core/Engine/utils/utils.test.ts
```diff
@@ -1,37 +1,46 @@
...
-          AccountsController: mockControllerInitFunction,
-        },
-        persistedState: {},
-        baseControllerMessenger: new ExtendedControllerMessenger(),
-      }),
-    ).toThrow(
-      'Controller requested before it was initialized: NetworkController',
-    );
+    expect(controllers.controllersByName.TransactionController).toBeDefined();
   });
```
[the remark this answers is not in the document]
matthewwalsh0: I'm not sure I follow, this is a function defined inside `initModularizedControllers` so do we not want to define the test only once?

Irrelevant of the type safety, the logic lives in once place so we need to check it behaves as expected?

app/core/Engine/utils/utils.ts
```diff
@@ -40,18 +60,34 @@ export const initModularizedControllers: InitModularizedControllersFunction = ({
...
     const messengerCallbacks = CONTROLLER_MESSENGERS[controllerName];
-    const controllerMessenger = messengerCallbacks.getMessenger(
+
+    const controllerMessengerCallback =
+      messengerCallbacks?.getMessenger as ControllerMessengerCallback;
+
+    const initMessengerCallback =
+      messengerCallbacks?.getInitMessenger as ControllerMessengerCallback;
+
+    const controllerMessenger = controllerMessengerCallback?.(
```
matthewwalsh0, line 76: The controller messenger isn't optional so should we remove the optional chaining?

app/core/Engine/utils/utils.test.ts
matthewwalsh0, line 1: Do we need additional tests to verify the init messenger support? Both provided and not provided?

matthewwalsh0 (changes requested):

app/core/Engine/Engine.ts
```diff
@@ -1329,16 +1134,150 @@ export class Engine {
...
+      PreferencesController: preferencesController,
+      SmartTransactionsController: this.smartTransactionsController,
+    };
+
+    const initRequest = {
+      getRootState: () => store.getState(),
+      getCurrentChainId: () => currentChainId,
+    };
+
+    const { controllersByName } = initModularizedControllers({
```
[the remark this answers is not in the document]
OGPoyraz: This is exactly what I thought and tried at the beginning. However there are controllers that need other controllers to be initialised.

Here are two examples that they use `AccountController` in their constructor. This means that these controller should come after.
https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/NftController.ts#L362
https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/MultichainBalancesController/MultichainBalancesController.ts#L163

I totally agree with you that these should be handled via callbacks or messengers but we have this kind of a bottleneck right now. 

I didn't totally check all the controllers in `this.context` to prevent chaos in this PR - but if all parties agree with that I can push this `initModularizedControllers` as much bottom as I can
Screenshot 2025-03-04 at 14 48 02

app/core/Engine/controllers/accounts-controller/index.test.ts
```diff
@@ -29,7 +29,12 @@ describe('accounts controller init', () => {
     jest.resetAllMocks();
     const baseControllerMessenger = new ExtendedControllerMessenger();
     // Create controller init request mock
-    initRequestMock = buildControllerInitRequestMock(baseControllerMessenger);
+    initRequestMock = {
+      ...buildControllerInitRequestMock(),
+      controllerMessenger:
+        baseControllerMessenger as unknown as AccountsControllerMessenger,
```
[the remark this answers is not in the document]
OGPoyraz: Done in https://github.com/MetaMask/metamask-mobile/pull/13817/commits/32139fcac82057e6241f8d3ca8b99cabd9426667

app/core/Engine/controllers/notifications/notification-services-push-controller-init.test.ts
```diff
@@ -18,20 +18,26 @@ describe('notificationServicesControllerInit', () => {
       )
       .mockReturnValue({} as NotificationServicesPushController);
     const fakeMessenger = {} as NotificationServicesPushControllerMessenger;
+
     return {
       mockLog,
       mockCreateController,
       fakeMessenger,
+      getController: jest.fn(),
```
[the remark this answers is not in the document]
OGPoyraz: Done in https://github.com/MetaMask/metamask-mobile/pull/13817/commits/eafdbdb43e97abcbb719e0414066b3a9009ab8d1

app/core/Engine/controllers/transaction-controller/transaction-controller-init.test.ts
OGPoyraz: I thought these configurations are especially important for mobile and we want to ensure have no change unintentionally. Agree on excessive but do you think it's not worth having them?

app/core/Engine/utils/utils.ts
OGPoyraz: Done

app/core/Engine/utils/utils.test.ts
OGPoyraz: Done

.github/CODEOWNERS
```diff
@@ -35,6 +35,8 @@ app/reducers/fiatOrders/             @MetaMask/ramp
 
 # Confirmation Team
 app/components/Views/confirmations   @MetaMask/confirmations
+app/core/Engine/controllers/transaction-controller             @MetaMask/confirmations
+app/core/Engine/messengers/transaction-controller-messenger    @MetaMask/confirmations
```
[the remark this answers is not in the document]
OGPoyraz: I will do the change and pass ownership of messengers to platform, regarding to your [message here](https://consensys.slack.com/archives/C02U025CVU4/p1741164242761389?thread_ts=1740168363.987169&cid=C02U025CVU4)

app/core/Engine/Engine.ts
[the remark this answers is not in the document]
matthewwalsh0: This is a big problem, we absolutely should not be calling other controllers within constructors.

In general, a constructor shouldn't even have any side effects, especially calling out to other components, as its only purpose is to initialise data on the instance.

The standard solution to this is an `init` function, that can even be `async`, to encapsulate whatever processes need to run on startup.

The controllers should support initialisation in any order, since nothing should "happen" in the wallet until they are all instantiated.

.github/CODEOWNERS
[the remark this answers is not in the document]
OGPoyraz: Done

app/core/Engine/Engine.ts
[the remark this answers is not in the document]
OGPoyraz: On separate topic - do we want to move this `initModularizedControllers` under `this.context` in this PR or separate one?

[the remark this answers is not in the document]
matthewwalsh0: We should aim to do the minimum at this stage given the size of the PR, so moving it seems the best temporary solution.

app/core/Engine/controllers/transaction-controller/transaction-controller-init.test.ts
matthewwalsh0: In general, I'd vote that tests verifying static data aren't needed since there is no "logic" to validate, rather we're just adding complexity to future maintenance.

app/core/Engine/controllers/transaction-controller/transaction-controller-init.ts
```diff
@@ -0,0 +1,166 @@
...
+        state: persistedState.TransactionController,
+      });
+    return { controller: transactionController };
+  } catch (error) {
+    Logger.error(error as Error, 'Failed to initialize TransactionController');
+    throw error;
+  }
+};
+
+function publishHook(
```
matthewwalsh0, line 108: Minor, if we exceed 2 or 3 arguments, it is cleaner and more readable to use an object?

app/core/Engine/types.ts
```diff
@@ -573,6 +575,26 @@ export type ControllerInitRequest<
     name: Name,
   ): ControllerByName[Name];
 
+  /**
+   * Get the chain id set in the engine.
```
matthewwalsh0, line 579:
```suggestion
   *  Retrieve the chain ID of the globally selected network.
```

app/core/Engine/types.ts
```diff
@@ -573,6 +575,26 @@ export type ControllerInitRequest<
...
+   * @deprecated Will be removed in the future pending multi-chain support.
+   */
+  getGlobalChainId: () => Hex;
+
+  /**
+   * Get the UI state of the app, returning the current Redux state including transient UI data,
+   * whereas `persistedState` contains only the subset of state persisted across sessions.
+   * For example: `{ settings, user, engine: { backgroundState: EngineState } }`.
+   */
+  getUIState: () => RootState;
```
matthewwalsh0, line 590: I worry this is ambiguous since the entire ReactNative app could be considered UI.

Would `getAllState` or `getState` be more explicit?

matthewwalsh0 (changes requested):

matthewwalsh0 (dismissed):

app/core/Engine/controllers/transaction-controller/transaction-controller-init.test.ts
OGPoyraz: Fair, removed those checks.

app/core/Engine/controllers/transaction-controller/transaction-controller-init.ts
OGPoyraz: Done

app/core/Engine/types.ts
OGPoyraz: Adjusted jsdoc

app/core/Engine/types.ts
OGPoyraz: Done - changed this to `getState` - sounds more expected to return `store.getState` result

Prithpal-Sooriya (approved):
Approved for Notifications

matthewwalsh0 (approved):

app/core/Engine/Engine.ts
```diff
@@ -1291,6 +1104,133 @@ export class Engine {
         ],
       }),
     });
+    const existingControllersByName = {
```
tommasini, line 1107: I'm not understanding the need of this object, I'm probably missing something, can you elaborate?

app/core/Engine/Engine.ts
```diff
@@ -163,11 +161,7 @@ import { getAllowedSmartTransactionsChainIds } from '../../../app/constants/smar
 import { selectBasicFunctionalityEnabled } from '../../selectors/settings';
 import { selectShouldUseSmartTransaction } from '../../selectors/smartTransactionsController';
 import { selectSwapsChainFeatureFlags } from '../../reducers/swaps';
-import {
-  SmartTransactionStatuses,
-  ClientId,
-} from '@metamask/smart-transactions-controller/dist/types';
-import { submitSmartTransactionHook } from '../../util/smart-transactions/smart-publish-hook';
+import { ClientId } from '@metamask/smart-transactions-controller/dist/types';
```
tommasini, line 164: Could we create an issue or a PR to start export this ClientId instead of importing it from the dist?

app/core/Engine/messengers/index.ts
```diff
@@ -10,34 +10,50 @@ import { getMultichainAssetsRatesControllerMessenger } from './multichain-assets
...
+
 
 /**
  * The messengers for the controllers that have been.
  */
-export const CONTROLLER_MESSENGERS: ControllerMessengerByControllerName = {
+export const CONTROLLER_MESSENGERS = {
   AccountsController: {
     getMessenger: getAccountsControllerMessenger,
+    getInitMessenger: noop,
```
tommasini, line 25: I'm clearing missing some context sorry! are we adding the getInitMessenger to the other controller messengers, because of typescript? 
I see it's because we define it [here](https://github.com/MetaMask/metamask-mobile/pull/13817/files#diff-8b6bc312eb604a62736d89cdeff8bc857167976578dc21072fe0492371ff5a31R657)

I'm wondering, this will be something that will be refactored soon to bring the init of the messengers to this controllers or they do not needed? 
I'm not understanding, why we need to define an initMessenger and a messenger? Would love to understand the advantages that lead towards this!

app/core/Engine/Engine.ts
OGPoyraz: These are the controllers that needs to be initialised before-hands to initialise controller that using new approach. So this is kinda guaranteeing the order of initialising controllers as we do before. 

One can get remove these controllers from this object and see which controllers are failing.

app/core/Engine/messengers/index.ts
OGPoyraz: > I'm clearing missing some context sorry! are we adding the getInitMessenger to the other controller messengers, because of typescript?
> I see it's because we define it here

I wanted to keep parity with extension since it also assigns `noop` to them for clarity. But that's also right that it type check will throw if we don't assign `noop` here.

DM'ed on other question.

Cal-L (approved):
LGTM from platform

PR #13844, opened 2025-03-04
ci: Use ubuntu-latest

FrederikBolding:
## **Description**

Use `ubuntu-latest` for CI as the current image we are using is deprecated and soon to be removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
             echo "No changes detected"
           fi
   dedupe:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
             echo "No changes detected"
           fi
   dedupe:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
owencraston, line 34: is it wise for us to always pull the latest instead of pinning a specific version that we know will work?

tommasini (approved):
We are using ubuntu-latest in some yml files already, I believe this means no harm

owencraston (approved):
It seems this is fixing CI.

.github/workflows/ci.yml
owencraston: I just saw that we are already using ubuntu-latest in plenty of places in the app so I see no harm in doing it here.

Prithpal-Sooriya (approved):

PR #13879, opened 2025-03-06
feat: revisit currencies list

salimtb:
## **Description**
The purpose of this PR is to review and align the currency list with the options available in our price API. If a user has already chosen a currency that no longer appears in the updated list, they will be automatically migrated to USD.

## **Related issues**

Fixes:

## **Manual testing steps**

1.Go to main and choose any currency not available on the new list
2.Switch the branch
3.you should see usd

### **After**

Screenshot 2025-03-06 at 13 13 57


app/core/Multichain/constants.ts
```diff
@@ -47,3 +47,50 @@ export const MULTICHAIN_NETWORK_BLOCK_EXPLORER_URL_MAP = {
   [SolScope.Devnet]: 'https://explorer.solana.com/?cluster=devnet',
   [SolScope.Testnet]: 'https://explorer.solana.com/?cluster=testnet',
 } as const;
+
+export const PRICE_API_CURRENCIES = [
```
Prithpal-Sooriya, line 51: Future - since this is duplicated in both extension and mobile, lets move this into CORE.

Prithpal-Sooriya: For visibility, here was the extension PR that added this.
https://github.com/MetaMask/metamask-extension/pull/30324

salimtb: yep good idea , i'll create a PR on core to export this list

Prithpal-Sooriya (dismissed):

app/store/migrations/069.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const DEFAULT_CURRENCY = 'usd';
+
+/**
+ * Migration to update the `currentCurrency` in `CurrencyController` to a valid available currency.
+ * If it's missing or invalid, it defaults to "USD".
+ * @param {unknown} stateAsync - Promise Redux state.
+ * @returns Migrated Redux state.
+ */
+export default async function migrate(stateAsync: unknown) {
```
tommasini, line 15: Why is this migration async? It doesn't need to be

app/store/migrations/069.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    return state;
+  }
+
+  const { currentCurrency } = currencyController;
+
+  if (!currentCurrency) {
+    captureException(
+      new Error(
+        `Migration: Missing currentCurrency in CurrencyController, defaulting to ${DEFAULT_CURRENCY}`,
```
tommasini, line 42: Not sure if we want to catch this with Sentry, would you say this is an expected error?

app/store/migrations/069.ts
```diff
@@ -0,0 +1,57 @@
...
+      ),
+    );
+    currencyController.currentCurrency = DEFAULT_CURRENCY;
+    return;
```
tommasini, line 46: This will break the app; this will make the migration return an undefined state to the app if I'm not mistaken

app/store/migrations/069.ts
```diff
@@ -0,0 +1,57 @@
...
+    captureException(
+      new Error(
+        `Migration: Invalid CurrencyController state type '${typeof currencyController}'`,
+      ),
+    );
+
+    return state;
+  }
+
+  const { currentCurrency } = currencyController;
```
tommasini, line 37: We may want to check if `currentCurrency` exist before destructuring it from currencyController! In migrations we should prevent all not expected scenario, and send to Sentry if that scenario is actually breaking the app or if it will recover gracefully. But if `currencyController` doens't have `currentCurrency` property, the app would crash right?

salimtb: In this scenario, we already check if `currencyController` is an object. Destructuring would only throw an error if `currencyController` were `null` or `undefined`, which isn't the case here since it's verified as an object beforehand.
so the app will not crash if there's no `currentCurrency` but this one will juste take `undefined` as value

tommasini: You are totally right! Sorry for the false alarm!

tommasini (approved):
LGTM!

PR #13907, opened 2025-03-06
fix: cp-7.42.0`Tokens` screen performance degradation

### LINKED ISSUE #13850: [Bug]: [Android Specific] App's performance has degraded
### Describe the bug

While testing with a wallet with multiple accounts (6 to be exact) and an imported account all with funds across various networks. Based on my testing, the app exhibits sluggish behavior in the following areas:

- When I select BSC as my default network, the tokens are slow to import to the wallet and the wallet starts moving sluggish from then on
- With the "popular networks" filter selected, when I import an ERC-20 token, for example, pepe on mainnet, the app starts to move sluggish.  

Tanya also mentioned a few other cases [here](https://github.com/MetaMask/metamask-mobile/issues/12063#issuecomment-2693290308)

I experienced this on my samsung S8 running android 9 and Xaomi redmi note 8 running android 11. 

### Expected behavior

The android app performance should match that of iOS.

### Screenshots/Recordings

Xaomi recording: You can notice the sluggish behavior while interacting with the wallet:

https://github.com/user-attachments/assets/ea314c0d-460e-49ad-aba4-af956d75b365

Samsung S8: Sluggish behavior is more prevalent than xaomi. Also note: this is not in the recording, but the keyboard pops up at random points on the wallet view. 

https://github.com/user-attachments/assets/b6d307e3-2abf-4f26-ab79-0a878d04a979

### Steps to reproduce

Launch the android app with an imported wallet with assets (tokens, NFTs)
Add BSC to your network list. Notice the sluggish behavior
Now filter by popular networks. Notice how interacting with the wallet is more sluggish
Try adding a token like "PEPE" while on BSC. Notice the slugish behavior

```shell

```

### Detection stage

During release testing

### Version

7.42.0

### Build type

None

### Device

Samsung s8 and Xaomi redmi note 8

### Operating system

Android

### Additional context

_No response_

### Severity

_No response_

gambinish:
## **Description**

Over the past couple of versions, we've seen a continued regression to Android performance. This was particularly noticeable when navigating quickly throughout the app, specifically on Samsung and Xaomi devices. We also noticed issues where the TokenList would freeze and become unresponsive.

After troubleshooting with the mobile team and QA, I think we can narrow this down to (at least) two things:

1. `Tokens` screen over-rendering
2. React context mounting/unmounting unnecessarily

The following two changes are being introduced in this PR:

Memoizing selectors on `Tokens` screen to optimize rendering with `createDeepEqualSelector`
- This helps to memoize data from state, to avoid React from de-referencing and causing unnecessary re-renders. `useSelector` only creates a shallow copy, which may often lead to undesirable behavior.

Moving `AssetsPollingProvider` lower down in the component hierarchy
- This polling provider initializes polling loops in core, for controllers that extend `StaticIntervalPollingController` (this includes the EVM assets controllers). The idea behind these polling loops is to only start them for components that need the data polled (ie be UI based)
- But currently, this context provider wraps our main nav component. This means that on each navigation, it would unmount/remount the context and re-initialize polling loops unnecessarily.

While this is a good start, I think that we should continue to expand/refine these changes to optimize performance further in future versions.

## **Related issues**

Partially fixes: https://github.com/MetaMask/metamask-mobile/issues/13850 

## **Manual testing steps**

1. Launch the android app with an imported wallet with assets (tokens, NFTs)
2. Add BSC to your network list. Notice the sluggish behavior
3. Now filter by popular networks. Notice how interacting with the wallet is more sluggish
4. Try adding a token like “PEPE” while on BSC. Notice the slugish behavior

### **Before**

https://www.loom.com/share/b5c33532e7354a48ba481814dd5fa01c?sid=dd25d6af-5239-4fe4-b113-dc0ebb6e25b7

### **After**

https://www.loom.com/share/2fb837e3607a4e9dba1239555d6b488a?sid=cf8d437d-9ab2-4272-862e-19e5b6eb17df


app/components/UI/Tokens/index.tsx
```diff
@@ -147,21 +144,8 @@ const Tokens: React.FC<TokensI> = memo(({ tokens }) => {
...
   const multiChainTokenBalance = useSelector(selectTokensBalances);
-  const debouncedMultiChainTokenBalance = useDebouncedValue(
-    multiChainTokenBalance,
-    DEBOUNCE_DELAY,
-  );
   const multiChainCurrencyRates = useSelector(selectCurrencyRates);
-  const debouncedMultiChainCurrencyRates = useDebouncedValue(
-    multiChainCurrencyRates,
-    DEBOUNCE_DELAY,
-  );
```
gambinish, line 164: Since the values being selected are now memoized, this debounce should no longer be necessary.

salimtb (dismissed):

app/components/Nav/Main/index.js
```diff
@@ -415,37 +414,35 @@ const Main = (props) => {
 
   return (
     <React.Fragment>
-      <AssetPollingProvider>
```
Prithpal-Sooriya, line 418: Is moving the Provider intentional in this PR?

Prithpal-Sooriya: (side-note) I see that this provider isn't really a context/provider, but just a wrapper component that runs a bunch of side-effects.

2 thoughts:

1. Maybe we can replace this provider component with a hook.

```ts
export function useAssetPolling() {
  // all the polling hooks
}
```

2. (might be more interesting) rather than using a Wrapper component, make this an isolated component that does not render/re-render children.

```ts
/**
*  A component that only initialises polling hooks, does not re-render.
*/
function InitAssetPolling() {
  // all the polling hooks
  
  // Does not render children! This is a ghost component.
  return null;
}

// Ensure that this component is never re-rendered
export default memo(AssetPolling)

// -------

// Now we can stick this high up in the component tree, and avoid re-renders
<InitAssetPolling />
```

app/components/UI/Tokens/index.test.tsx
```diff
@@ -567,7 +567,8 @@ describe('Tokens', () => {
       expect(selectAccountTokensAcrossChainsSpy).toHaveBeenCalled();
     });
 
-    it('should not call selectAccountTokensAcrossChains when disabled', () => {
+    // this test needs to be skipped, we should remove with PV feature flag
+    it.skip('should not call selectAccountTokensAcrossChains when disabled', () => {
```
Prithpal-Sooriya, line 571: If we are skipping this test, do we still have value keeping it? (esp if we are removing it once we remove the PV feature flag)

app/selectors/multichain/evm.test.tsx
```diff
@@ -181,3 +203,177 @@ describe('Multichain Selectors', () => {
...
+    },
+  } as unknown as RootState;
+  const mockRenderCall = jest.fn();
+  let getByText: GetByQuery<TextMatch, CommonQueryOptions & TextMatchOptions>;
+  let store: Store;
+
+  beforeEach(() => {
+    mockRenderCall.mockReset();
+    // Clear memoized selectors for each test
+    selectAccountTokensAcrossChains.memoizedResultFunc.clearCache();
```
Prithpal-Sooriya, line 321: Nice tests, I think there is value in testing a selector inside a component!

(Future) If possible can we also test the selector in isolation, and with an equality check too?

```ts
it('should return exactly same data if state does not change', () => {
  const result1 = selectAccountTokensAcrossChains(state)
  const result2 = selectAccountTokensAcrossChains(state)
  expect(result1 === result2).toBe(true)
});
```

Prithpal-Sooriya: If the selector is too complex (which it does look complex since it takes many other selectors) we can leave this till later.

Prithpal-Sooriya (dismissed):

app/components/Nav/Main/index.js
Prithpal-Sooriya: I'm also very interest to see if mobile supports React Portals.

This was we can have a this component rendered but not tied to the DOM or parents. This way it can just exist and never re-render.

app/components/UI/Tokens/index.test.tsx
gambinish: I'm cool with removing it. Will push a commit.

app/components/Nav/Main/index.js
gambinish: I like both of these ideas. Let's discuss these things further with the team. Maybe we can fit this into an upcoming release?

app/components/UI/Tokens/index.test.tsx
gambinish: https://github.com/MetaMask/metamask-mobile/pull/13907/commits/81294cc396be898aa7ceec09408a4971bc65eb76

Prithpal-Sooriya (approved):

salimtb (approved):

app/components/Nav/Main/index.js
Prithpal-Sooriya: No worries merging into an upcoming release if this is a big refactor 😛 , but lets discuss!

MarioAslau (approved):

PR #13925, opened 2025-03-07
feat: add DeFi tab

bergarces:
## **Description**

Adds a DeFi positions tab to the main page, hidden behind a feature flag enabled only in dev.

Designs: [Figma](https://www.figma.com/design/OuFNyOcZFFQXHMtmwPlLLc/DeFi?node-id=199-15265&t=iVCbAOBHEwueWq8d-0)

New features:
- New DeFi tab
- List of DeFi positions
- Detailed view of each position
- Use feature flag to disable tab

Out of scope for this ticket:
- Aggregate value of all positions added to the total

## **Related issues**

Fixes: [JIRA](https://consensyssoftware.atlassian.net/browse/MMASSETS-555)

## **Manual testing steps**

1. Open the mobile app
2. Access the DeFi tab
3. Click any position to check the details

## **Screenshots/Recordings**

https://github.com/user-attachments/assets/35e6db20-d57c-4dc2-a935-ba4c7a58c06f


app/core/Engine/controllers/defi-positions-controller/defi-positions-controller-init.ts
```diff
@@ -0,0 +1,46 @@
...
+          ?.remoteFeatureFlags?.assetsDefiPositionsEnabled,
+      );
+
+      console.log('DEFI ENABLED', {
+        isBasicFunctionalityToggleEnabled,
+        featureFlagForDeFi,
+      });
+
+      // return isBasicFunctionalityToggleEnabled && featureFlagForDeFi;
+      return true;
```
bergarces, line 41: Restore flags before merging

package.json
```diff
@@ -163,7 +163,7 @@
     "@metamask/address-book-controller": "^6.0.3",
     "@metamask/app-metadata-controller": "^1.0.0",
     "@metamask/approval-controller": "^7.1.3",
-    "@metamask/assets-controllers": "^61.0.0",
+    "@metamask/assets-controllers": "^62.0.0",
```
sahar-fehri, line 166: I think there is a patch that is missing, it probably got lost in the last upgrade 🤔

app/components/UI/DeFiPositions/DeFiPositionsControlBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <View style={styles.actionBarWrapper}>
+      <ButtonBase
+        label={
+          <Text numberOfLines={1}>
+            {isAllNetworks && isPopularNetwork
+              ? `${strings('app_settings.popular')} ${strings(
+                  'app_settings.networks',
+                )}`
+              : networkName ?? strings('wallet.current_network')}
```
Prithpal-Sooriya, line 51: There may be a tiny conflict with https://github.com/MetaMask/metamask-mobile/pull/15253
We can tidy this if the above PR is merged.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
```diff
@@ -0,0 +1,297 @@
...
+    const { findByTestId, findByText, queryByText } = renderWithProvider(
+      <DeFiPositionsList tabLabel="DeFi" />,
+      {
+        state: mockInitialState,
+      },
+    );
+
+    expect(await findByTestId(DEFI_POSITIONS_CONTAINER)).toBeDefined();
+    expect(await findByText('Protocol 1')).toBeDefined();
+    expect(queryByText('Protocol 2')).toBeNull();
```
Prithpal-Sooriya, line 194: Thank you for not using snapshots and actual assertions!

app/components/UI/DeFiPositions/DeFiPositionsList.tsx
```diff
@@ -0,0 +1,121 @@
...
+        </Text>
+      </View>
+    );
+  }
+
+  return (
+    <View testID={DEFI_POSITIONS_CONTAINER}>
+      <DeFiPositionsControlBar />
+      <View>
+        <FlatList
```
Prithpal-Sooriya, line 103: Future - lets move to FlashList.

^ not do this now since we haven't measured any performance implications on this list.

app/components/UI/DeFiPositions/DeFiPositionsListItem.tsx
```diff
@@ -0,0 +1,147 @@
...
+      if (a.marketValue === undefined && b.marketValue === undefined) {
+        return 0;
+      }
+      if (a.marketValue === undefined) {
+        return 1;
+      }
+      if (b.marketValue === undefined) {
+        return -1;
+      }
+      return b.marketValue - a.marketValue;
```
Prithpal-Sooriya, line 75: Nice sorting, I think we can simplify this.
```ts
const aMarketValue = a.marketValue ?? 0;
const bMarketValue = b.marketValue ?? 0;
return bMarketValue - aMarketValue;

// Cases:
// Both undefined - will return 0;
// Both have values - will return `b - a`
// a undefined, but b has value - will return `b - 0`, so positive number
// b undefined, but a has value - will return `0 - a`, so negative number
```

app/components/UI/DeFiPositions/DeFiPositionsListItem.tsx
```diff
@@ -0,0 +1,147 @@
...
+              symbol: sortedTokens[0].symbol,
+            })
+          : sortedTokens.length === 2
+          ? strings('defi_positions.two_tokens', {
+              symbol: sortedTokens[0].symbol,
+            })
+          : strings('defi_positions.multiple_tokens', {
+              symbol: sortedTokens[0].symbol,
+              count: sortedTokens.length - 1,
+            }),
```
Prithpal-Sooriya, line 100: Nice ternary, surprised prettier didn't go haywire lol.

I'm happy to keep a ternary,

<details><summary>I was playing with with potentially using an intermediate obj? But it's not necessary.</summary>

```ts
const getTokenStr = {
  0: () => '',
  1: () => strings('...single', {
    symbol: sortedTokens[0].symbol,
  }),
  2: () => strings('...two', {
    symbol: sortedTokens[0].symbol,
  }),
  default: () => strings('...multiple', {
    symbol: sortedTokens[0].symbol,
    count: sortedTokens.length - 1,
  })
};

const tokenStr = getTokenStr[sortedTokens.length] ?? getTokenStr.default;
tokenNames: tokenStr()
```

</details>

app/components/UI/DeFiPositions/DeFiProtocolPositionDetails.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+        state: mockInitialState,
+      },
+    );
+
+    expect(mockSetOptions).toHaveBeenCalledTimes(1);
+    expect(await findByText('Protocol 1')).toBeDefined();
+    expect(
+      (await findByTestId(DEFI_PROTOCOL_POSITION_DETAILS_BALANCE_TEST_ID)).props
+        .children,
+    ).toStrictEqual('$100.00');
```
Prithpal-Sooriya, line 84: Thoughts on using:
```
expect(...).toHaveTextContent($100.00')
```

app/components/UI/DeFiPositions/DeFiProtocolPositionGroupTokens.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        tokens={mockTokens}
+        networkIconAvatar={10}
+        privacyMode={false}
+      />,
+      {
+        state: mockInitialState,
+      },
+    );
+
+    expect(await findByText('Supplied')).toBeDefined();
```
Prithpal-Sooriya, line 56: Same as above on using new jest matchers.
```ts
expect(...).tobeOnTheScreen()
```

app/components/UI/DeFiPositions/DeFiProtocolPositionGroupTokens.tsx
```diff
@@ -0,0 +1,97 @@
...
+    return null;
+  }
+
+  return (
+    <View>
+      <Text variant={TextVariant.BodyMDMedium} color={TextColor.Alternative}>
+        {strings(`defi_positions.${positionType}`)}
+      </Text>
+      {tokens.map((token, i) => (
+        <View key={i} style={styles.underlyingBalancesWrapper}>
```
Prithpal-Sooriya, line 49: oooooh, using the index as a key 😄 
This is fine, I'm not too concerned with this.

Prithpal-Sooriya: Is this list large? If so, maybe we can use a FlatList/FlashList to virtualise these items.

app/components/UI/DeFiPositions/DeFiProtocolPositionGroups.tsx
```diff
@@ -0,0 +1,103 @@
...
+        };
+      }).filter(
+        (group): group is NonNullable<typeof group> => group !== undefined,
+      ),
+    [protocolAggregate],
+  );
+
+  return (
+    <View style={styles.protocolDetailsPositionsWrapper}>
+      {positionGroups.map((positionGroup) => (
```
Prithpal-Sooriya, line 73: This is only mapping 4 items, so we don't need to use a FlatList/FlashList.

app/core/Engine/controllers/defi-positions-controller/defi-positions-controller-init.ts
```diff
@@ -0,0 +1,40 @@
...
+  DeFiPositionsControllerInitMessenger
+> = (request) => {
+  const { initMessenger, controllerMessenger } = request;
+
+  const controller = new DeFiPositionsController({
+    messenger: controllerMessenger,
+    isEnabled: () => {
+      const isBasicFunctionalityToggleEnabled = selectBasicFunctionalityEnabled(
+        store.getState(),
+      );
```
Prithpal-Sooriya, line 28: Q (can be resolved) - it is interesting that this basicFunctionalityToggle is fetched from state, whereas the defi flag is fetched from messenger call.

If we wanted to, we could stick with 1 mechanism:
- Use messenger call to grab state for both basic functionality and remote feature flag
- Use redux store to grab state for both basic functionality and remote feature flag.

I'm tending towards 1 since I don't know the state of the redux application on startup (if it has fully been hydrated before controller init)

Prithpal-Sooriya: TBH this is fine, we can leave as is.

package.json
Prithpal-Sooriya: Mobile patches are not shown in the package.json. Instead they are applied using `patch-package` in the [`/patches` folder](https://github.com/MetaMask/metamask-mobile/blob/main/patches/%40metamask%2Bassets-controllers%2B60.0.0.patch)

You are right though, the last upgrade did not apply the patch from 60.0.0 to 61.0.0

Prithpal-Sooriya: Speaking of mobile patches, when will we update the assets controllers so we can remove these mobile patches? Unsure which PRs these patches point to.

app/components/UI/DeFiPositions/DeFiProtocolPositionGroupTokens.tsx
bergarces: It shouldn't be long. Normally one or two underlyings, in cases like GLP there are 8 underlyings, but that's the longest I've ever seen.

I missed that index, it was meant to be temporary. I'm going to create the key using the protocol token address and underlying token address, along with the index (I've seen very rare protocol tokens with the same underlying twice, that's why I need the index too).

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
bergarces: I had never seen those matchers before. Definitely good to signal the intention.

app/components/UI/DeFiPositions/DeFiPositionsListItem.tsx
bergarces: 3 ternaries is definitely pushing it. I'll go the intermediate object.

app/core/Engine/controllers/defi-positions-controller/defi-positions-controller-init.ts
Prithpal-Sooriya: yah, this is fair. Lets keep as is 😛

Prithpal-Sooriya (dismissed):

app/components/UI/DeFiPositions/DeFiPositionsControlBar.tsx
bergarces: I've just updated this bit here.

Prithpal-Sooriya (dismissed):

locales/languages/en.json
```diff
@@ -3992,5 +3993,18 @@
...
     }
+  },
+  "defi_positions": {
+    "loading_positions": "Loading DeFi positions...",
+    "no_positions": "No positions yet",
+    "error_cannot_load_page": "We could not load this page.",
+    "error_visit_again": "Try visiting again later.",
+    "single_token": "{{symbol}} only",
+    "two_tokens": "{{symbol}} +1 other",
+    "multiple_tokens": "{{symbol}} +{{count}} others",
```
jpsains, line 4004: re these template stings does something like this exist in extension?

app/components/UI/DeFiPositions/DeFiPositionsControlBar.styles.ts
NicolasMassart, line 1: Network control bar text color is not visible on iOS (simulator) when in dark mode:
image

app/core/Engine/controllers/defi-positions-controller/defi-positions-controller-init.ts
NicolasMassart: @MetaMask/mobile-platform could help confirming this is the right way to init

locales/languages/en.json
bergarces: I think so. But the syntax is slightly different.

https://github.com/MetaMask/metamask-extension/blob/a5c48a5d839eee5df5644f53f3c07966716d37f2/ui/pages/swaps/view-on-block-explorer/view-on-block-explorer.js#L39

app/components/UI/DeFiPositions/DeFiProtocolPositionDetails.tsx
NicolasMassart, line 1: Q: Not sure what's the logic of displaying two sections (supplied or staked) like on [7th screen on mobile Figma](https://www.figma.com/design/OuFNyOcZFFQXHMtmwPlLLc/DeFi?node-id=199-15265&t=jptUCgB4NscOp6hC-1).
Design show the same thing in both sections, but app only shows it for staked on my test account (screenshot) and both assets (ETH and sETH) are on same chain (ethereum):
image

What's the logic of these sections?

NicolasMassart (dismissed):
Some changes needed on unit tests and some questions and suggestions but overall the PR is a great code work! Thanks!

app/components/Views/Wallet/index.tsx
```diff
@@ -720,6 +736,12 @@ const Wallet = ({
     return (
       <ScrollableTabView renderTabBar={renderTabBar} onChangeTab={onChangeTab}>
         <Tokens {...tokensTabProps} />
+        {isEvmSelected &&
+          !isTestNet(chainId) &&
```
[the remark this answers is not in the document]
bergarces: DeFi positions API does not support any testnet, so it would always be empty.

locales/languages/en.json
NicolasMassart: No, on extension, you have to use `$1`, `$2`, ... see for instance this [message string with 2 variables](https://github.com/MetaMask/metamask-extension/blob/382936e666119f273be54231486055a921d5396d/app/_locales/en/messages.json#L520-L521)

app/components/UI/DeFiPositions/DeFiProtocolPositionDetails.tsx
bergarces: It comes defined by the API, as in some cases, when a protocol can have multiple positions with multiple underlyings and their own rewards, we were asked to make sure we could put a separation so we could do:
```
Staked
- Position 1 - Underlying 1 details
- Position 1 - Underlying 2 details
Rewards
- Position 1 - Rewards details

SEPARATOR

Staked
- Position 2 - Underlying 1 details
- Position 2 - Underlying 2 details
Rewards
- Position 2 - Rewards details
```

Personally I think it'll get reworked with time.

As for that specific protocol separating the two positions, that's because the API defaults to separate them, and we need to determine if they are going to be joined on a per-product basis (so far we've done so for Aave). In this cases, chances are we decide to group them, as I don't recall Lido positions having rewards. It'll happen at API level though, not at client level.

app/components/UI/DeFiPositions/DeFiPositionsList.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <Text variant={TextVariant.BodyMD} color={TextColor.Alternative}>
+            {strings('defi_positions.error_cannot_load_page')}
+          </Text>
+          <Text variant={TextVariant.BodyMD} color={TextColor.Alternative}>
+            {strings('defi_positions.error_visit_again')}
+          </Text>
+        </>
+      );
+    } else {
+      // No positions found for the current account
```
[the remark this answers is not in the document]
bergarces: I don't have that specific one at hand. It's copied from the extension, which has already been merged.

The error one is also copied from the extension, but appears on the figma page linked.

There's also a new design we've been recently given for loading, but that'll get implemented separately because it'll change the loading for both DeFi and Tokens tab.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+    },
+  },
+} as unknown as RootState;
+
+describe('DeFiPositionsList', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('renders protocol name and aggregated value for selected account and chain', async () => {
```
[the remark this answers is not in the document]
bergarces: Missed that.

Added value entry checks.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+        state: mockInitialState,
+      },
+    );
+
+    expect(await findByTestId(DEFI_POSITIONS_CONTAINER)).toBeOnTheScreen();
+    expect(await findByText('Protocol 1')).toBeOnTheScreen();
+    expect(queryByText('Protocol 2')).not.toBeOnTheScreen();
+  });
+
+  it('renders protocol name and aggregated value for all chains when all networks is selected', async () => {
```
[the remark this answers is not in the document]
bergarces: Added value entry checks.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+            },
+          },
+        },
+      },
+    );
+
+    expect(await findByText('Loading DeFi positions...')).toBeOnTheScreen();
+  });
+
+  it('renders the error message when the positions are null for that address', async () => {
```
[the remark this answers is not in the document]
bergarces: Done.

bergarces:
> isBasicFunctionalityToggleEnabled

This way of setting the variable at this stage has been copied directly from `snap-controller-init`. It's also present in `Engine.ts`

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
```diff
@@ -192,6 +192,8 @@ describe('DeFiPositionsList', () => {
     expect(await findByTestId(DEFI_POSITIONS_CONTAINER)).toBeOnTheScreen();
     expect(await findByText('Protocol 1')).toBeOnTheScreen();
     expect(queryByText('Protocol 2')).not.toBeOnTheScreen();
+    expect(await findByText('$100.00')).toBeOnTheScreen();
+    expect(queryByText('$10.00')).not.toBeOnTheScreen();
```
NicolasMassart, line 196: C: same issue as in the other test, see my regex suggestions, if the amount changes and is displayed, your assertions will be true. So you have to test there's really no protocol 2 value.

app/components/UI/DeFiPositions/DeFiPositionsControlBar.styles.ts
bergarces: Good catch. I've pushed the missing style.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
bergarces: The test set has two positions from two different chains.
- Position 1, with value 100
- Position 2, with value 10

I'm testing that, if the chain for position 1 is selected, it shows position 1 label and value, but not position 2 label and value.

If position 2 had the value displayed, then $10.00 will be on screen and the test would fail.

bergarces: I cannot see any regex suggestion on the comments. Am I missing something?

bergarces:
Just pushed init tests for controller initialisation logic, as I noticed that all other controllers have it.

app/components/UI/DeFiPositions/DeFiProtocolPositionGroups.tsx
```diff
@@ -0,0 +1,106 @@
...
+          protocolAggregate.positionTypes[positionType];
+
+        if (!protocolPositionsByType) {
+          return undefined;
+        }
+
+        return {
+          positionType,
+          positions: protocolPositionsByType.positions.flatMap((tokenGroup) =>
+            tokenGroup.map((protocolToken) => ({
```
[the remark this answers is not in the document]
NicolasMassart: You have to fix this to [pass Sonar](https://sonarcloud.io/project/issues?id=metamask-mobile&pullRequest=13925&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
NicolasMassart: You may have to unfold all conversation, Github hides some when there's a lot of comments:
image

NicolasMassart: Then you have to get all the values matching the amount regex, and then check that first element is visible and not the second one.
What worries me is that a change on your code and test at the same time may make the test pass. This should be avoided. Your test should only pass in the exact conditions where you expect it to pass. Otherwise we will have risk of false positive passing test.

app/components/UI/DeFiPositions/DeFiProtocolPositionGroupTokens.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      },
+    );
+
+    expect(await findByText('Supplied')).toBeOnTheScreen();
+    expect(await findByText('TKN1')).toBeOnTheScreen();
+    expect(queryByText('$50.00')).not.toBeOnTheScreen();
+    expect(queryByText('500 TKN1')).not.toBeOnTheScreen();
+    expect(await findByText('TKN2')).toBeOnTheScreen();
+    expect(queryByText('$2.00')).not.toBeOnTheScreen();
+    expect(queryByText('20 TKN2')).not.toBeOnTheScreen();
```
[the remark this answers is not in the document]
bergarces: I don't think that introducing regex in case that someone changes the test values is a good solution. The fact that it would require a comment explaining what the regex is for is a red flag.

If they change the test values, then they need to rewrite or review the tests.

app/components/UI/DeFiPositions/DeFiProtocolPositionDetails.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+              },
+            },
+          },
+        },
+      },
+    );
+
+    expect(mockSetOptions).toHaveBeenCalledTimes(1);
+    expect(await findByText('Protocol 1')).toBeOnTheScreen();
+    expect(queryByText('$100.00')).not.toBeOnTheScreen();
```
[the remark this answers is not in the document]
bergarces: As before, if they change the test value and remove privacy mode, then they need to rewrite/review this test. Introducing regex for a test like this is not something I'm prepared to do.

app/components/UI/DeFiPositions/DeFiProtocolPositionGroupTokens.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <View style={styles.balance}>
+            <SensitiveText
+              variant={TextVariant.BodyMDMedium}
+              isHidden={privacyMode}
+              length={SensitiveTextLength.Medium}
+            >
+              {token.marketValue
+                ? formatWithThreshold(token.marketValue, 0.01, I18n.locale, {
+                    style: 'currency',
+                    currency: 'USD',
```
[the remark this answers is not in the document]
bergarces: Because we do not have support for any other currency in DeFi right now, so we are hardcoding USD.

app/components/UI/DeFiPositions/DeFiProtocolPositionGroups.tsx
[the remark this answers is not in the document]
bergarces: Sonar is passing, those are suggestions.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
bergarces: Thanks. I didn't see those comments.

I'm really not worried about that, and introducing regex for these tests would simply make them more difficult to read and add virtually no value.

If someone changes the values on the test, they should review the test.

What I can do, which should make the scenario you describe more difficult, is to replace the hardcoded data so that it is generated at the beginning of every test with specific values.

app/components/UI/DeFiPositions/DeFiProtocolPositionDetails.test.tsx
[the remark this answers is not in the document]
bergarces: Changes added as suggested

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
bergarces: I'm going to backtrack here and go with your suggestion. Even though I don't think them to be necessary and that they difficult the readability of the test itself, I'd rather avoid a refactor on all my unit tests to create a custom initialisation of the state for every one of them. Delaying this feature for that is not worth it.

I have pushed the suggested changes to the other tests, in this specific one, a regex cannot be used because one of the protocol entries shows whilst the other doesn't. So instead I'm checking that the correct entry is there and that only one entry is in the list.

bergarces:
Well. It looks like we cannot merge more than 1000 lines now, so I'll have to close this PR anyway and somehow split the feature.

app/components/UI/DeFiPositions/DeFiPositionsList.test.tsx
bergarces: Looks like the PR cannot be merged now that the limit of 1000 lines has been implemented. So I'll have to work on splitting it into smaller PRs.

bergarces: @NicolasMassart I have applied all the review suggestions I could see. The PR max lines is non-blocking, so not really worth splitting an already existing and reviewed PR.

Prithpal-Sooriya:
The DeFi feature looks good. QA'ed on both IOS and Android physical device

https://www.loom.com/share/9b8d4ef238be43b1b84861db31f174eb?sid=4f4a051d-a597-4fa1-8c7b-5c730beb6fe5

https://www.loom.com/share/0bbda037c88b4f6f89ff321599252efa?sid=8ef5f01c-db12-464c-b7a6-3609781b0ae9

Prithpal-Sooriya (approved):

NicolasMassart (approved):
I'm good on @MetaMask/mobile-platform files

app/components/UI/DeFiPositions/DeFiProtocolPositionGroupTokens.test.tsx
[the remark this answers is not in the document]
Prithpal-Sooriya: @NicolasMassart I really dislike adding regex in our tests, and I want to dissuade us from doing this going forward.

1. It makes it very hard to read the tests, and grep/search.
2. We are relying on comments to explain what we are testing? This can get out of date!
3. This does not protect us if values/calculations change (e.g. unexpected currency conversions)
4. It introduces a pattern that other devs will use, worst case we would end up with very complex regex patterns!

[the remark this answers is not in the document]
Prithpal-Sooriya: If the intention was to avoid false positives in the test, surely there are better solutions?
```
const elem = getByTestId('...')
expect(elem).toHaveTextContent('...') // Here we can now explicitly check that this element either had the exact text or asterisks
```

PR #14049, opened 2025-03-17
feat: Implement claim rewards confirmation

pedronfigueiredo:
## **Description**

- Implements a confirmation screen for claiming staking rewards.
- Adds static claim rewards simulation that hardcodes simulation data into the transaction metadata that is passed in.
  - Adds the new tooltip type to the simulations component tooltip.
- Introduces a hook to dismiss the pre-added confirmation from the controller state when navigating back by swipe gesture.
- Encapsulates navigation bar into a hook.
- Introduces locale support for gas fees native token.
- Adds unit tests

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4403

## **Manual testing steps**

1. Go to this page...
2.
3.

### **After**

Screenshot 2025-03-17 at 11 40 21


app/components/Views/confirmations/components/Confirm/Info/GasFeesDetails/GasFeesDetails.styles.ts
```diff
@@ -7,7 +7,8 @@ const styleSheet = (params: { theme: Theme }) => {
 
   return StyleSheet.create({
     container:{
-      paddingVertical: 8,
+      marginVertical: 0,
```
OGPoyraz, line 10: Just for heads up, I think this is a good practice in confirmation level components to strictly not have any paddings/margins like this as default and let consumer components set if needed.

app/components/Views/confirmations/components/Confirm/Info/Info.tsx
```diff
@@ -67,9 +70,14 @@ const Info = ({ route }: InfoProps) => {
...
 
-  if (transactionType === TransactionType.stakingUnstake) {
-    const StakingWithdrawalComponent = InfoComponent as React.ComponentType<UnstakeConfirmationViewProps>;
-    return <StakingWithdrawalComponent route={route as UnstakeConfirmationViewProps['route']} />;
+  if (
+    transactionType && [
+      TransactionType.stakingUnstake,
+      TransactionType.stakingClaim,
+    ].includes(transactionType)
+  ) {
```
OGPoyraz, line 78:
```suggestion
  ) {
  // Unstake and staking claim pages need route parameters to be passed into Info components (`StakingClaim` & `StakingWithdraw`)
```
Shall we add a comment here why we don't have `stakingDeposit` for future references?

app/components/Views/confirmations/components/Confirm/Info/StakingClaim/StakingClaimRewardsStaticSimulation/StakingClaimRewardsStaticSimulation.tsx
```diff
@@ -0,0 +1,39 @@
...
+    tokenBalanceChanges: [],
+  };
+  const enrichedTransactionMeta: TransactionMeta = {
+    ...(transactionMetadata as TransactionMeta),
+    simulationData: HARDCODED_SIMULATION_DATA,
+  };
+
+  return (
+    <View style={styles.container}>
+      <SimulationDetails
```
OGPoyraz, line 31: Just to get context, which part is not working for current simulation details so we do hardcoded simulation here?

app/components/Views/confirmations/components/Confirm/Info/StakingDeposit/StakingDeposit.tsx
```diff
@@ -1,26 +1,16 @@
...
-  useEffect(() => {
-    navigation.setOptions(
-      getNavbar({
-        title: strings('stake.stake'),
-        onReject,
-      }),
-    );
-  }, [navigation, onReject]);
+  useNavbar(strings('stake.stake'));
+  useClearConfirmationOnBackSwipe();
```
OGPoyraz, line 13: 🚀

app/components/Views/confirmations/hooks/useClearConfirmationOnBackSwipe.ts
```diff
@@ -0,0 +1,20 @@
...
+const useClearConfirmationOnBackSwipe = () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const navigation = useNavigation<StackNavigationProp<any>>();
+  const { onReject } = useConfirmActions();
+
+  useEffect(() => {
+    const unsubscribe = navigation.addListener('gestureEnd', () => {
+      onReject();
+    });
+
```
OGPoyraz, line 15: I was wondering if we should also handle Android back button handler here something like this 

```
import { BackHandler } from 'react-native';

///

const backHandlerSubscription = BackHandler.addEventListener(
      'hardwareBackPress',
      () => {
        onReject();
        // Return false to allow default back behavior
        return false;
      }
    );
    
    return () => {
      unsubscribe();
      backHandlerSubscription.remove();
    };
```

app/components/Views/confirmations/hooks/useFlatConfirmation.ts
```diff
@@ -5,6 +5,7 @@ import { useTransactionMetadataRequest } from '../hooks/useTransactionMetadataRe
 const FLAT_TRANSACTION_CONFIRMATIONS: TransactionType[] = [
   TransactionType.stakingDeposit,
   TransactionType.stakingUnstake,
+  TransactionType.stakingClaim,
```
OGPoyraz, line 8: It's a design question but I was wondering if we should use a modal confirmation for this. Because if you notice - claiming is not tied to staking navigator. 


https://github.com/user-attachments/assets/84c34a1d-b9e4-4fc8-a076-ae0a74e40489

app/components/UI/Stake/Views/StakeInputView/StakeInputView.tsx
```diff
@@ -76,6 +77,13 @@ const StakeInputView = ({ route }: StakeInputViewProps) => {
     });
   };
 
+  const [isSubmittingStakeDepositTransaction, setIsSubmittingStakeDepositTransaction] = useState(false);
+  useFocusEffect(
+    useCallback(() => {
+      setIsSubmittingStakeDepositTransaction(false);
+    }, [])
+  );
+
```
Matt561, line 86: Nit: Vertical spacing

```suggestion
  const [isSubmittingStakeDepositTransaction, setIsSubmittingStakeDepositTransaction] = useState(false);
  
  useFocusEffect(
    useCallback(() => {
      setIsSubmittingStakeDepositTransaction(false);
    }, [])
  );

```

app/components/UI/Stake/Views/StakeInputView/StakeInputView.tsx
```diff
@@ -76,6 +77,13 @@ const StakeInputView = ({ route }: StakeInputViewProps) => {
     });
   };
 
+  const [isSubmittingStakeDepositTransaction, setIsSubmittingStakeDepositTransaction] = useState(false);
+  useFocusEffect(
+    useCallback(() => {
+      setIsSubmittingStakeDepositTransaction(false);
```
Matt561, line 83: What is the reasoning for this `useFocusEffect`? From what I understand, this calls `setIsSubmittingStakeDepositTransaction(false)` when the screen is focused but we set the default to `false` already when calling `useState`.

app/components/UI/Stake/Views/StakeInputView/StakeInputView.tsx
```diff
@@ -111,6 +119,10 @@ const StakeInputView = ({ route }: StakeInputViewProps) => {
     const amountWeiString = amountWei.toString();
 
     if (isStakingDepositRedesignedEnabled) {
+      // this prevents the user from adding the transaction deposit into the
+      // controller state multiple times
```
Matt561, line 123: The comment is appreciated! 😄

app/components/UI/Stake/Views/UnstakeInputView/UnstakeInputView.tsx
```diff
@@ -24,11 +26,10 @@ import useUnstakingInputHandlers from '../../hooks/useUnstakingInput';
 import { withMetaMetrics } from '../../utils/metaMetrics/withMetaMetrics';
 import UnstakeInputViewBanner from './UnstakeBanner';
 import styleSheet from './UnstakeInputView.styles';
-import { selectConfirmationRedesignFlags } from '../../../../../selectors/featureFlagController';
 
 const UnstakeInputView = () => {
   const title = strings('stake.unstake_eth');
-  const navigation = useNavigation();
+  const navigation = useNavigation<StackNavigationProp<any>>();
```
Matt561, line 32: Are we able to type these `any`s?

Matt561 (changes requested):
Some minor comments but otherwise looks good!

app/components/Views/confirmations/components/Confirm/Info/StakingClaim/StakingClaimRewardsStaticSimulation/StakingClaimRewardsStaticSimulation.tsx
pedronfigueiredo: Good point, I removed this component.

app/components/Views/confirmations/hooks/useFlatConfirmation.ts
pedronfigueiredo: This confirmation follows design but I've also noticed you can't go back the assets view with gestures. Do you know how that can be added?

app/components/Views/confirmations/hooks/useClearConfirmationOnBackSwipe.ts
pedronfigueiredo: Is this case not already accounted for in the back button code on `app/components/Views/confirmations/components/Confirm/Navbar/Navbar.tsx`? Does android follow a different test path?

app/components/UI/Stake/Views/StakeInputView/StakeInputView.tsx
pedronfigueiredo: To add to that, the loading behaviour was put in place to prevent race conditions where the user adds more than on staking deposit (or staking withdrawal) confirmation into state at once. We only support one confirmation at a time for the time being, which means it needs to get discarded when we navigate back to the input view, so that another one can be added if we navigate forward once again.

app/components/Views/confirmations/hooks/useClearConfirmationOnBackSwipe.ts
OGPoyraz: You're right, this should have been in the `Navbar` with only the back button. It just came to mind, but we don't need to address it in this PR.

app/components/Views/confirmations/hooks/useFlatConfirmation.ts
OGPoyraz: In general mobile UX - this page should be a clear modal window to me. 

Because "Claim eth" button we see above is not tied to context of this screen - so we cannot put it in their navigator. Since claiming also don't have an entry screen like "Stake input" (in deposit and withdraw), I believe claiming transaction should be handled in a modal transaction. 

I will open a thread and ask Saya about this.

app/components/Views/confirmations/hooks/useClearConfirmationOnBackSwipe.ts
pedronfigueiredo: I created a new task here https://github.com/orgs/MetaMask/projects/78/views/1?pane=issue&itemId=102479543

app/components/UI/Stake/Views/StakeInputView/StakeInputView.tsx
Matt561: Sounds good! Thank you for the explanation.

Matt561 (approved):
Approving on behalf of earn team ✅

app/components/Views/confirmations/components/Confirm/Navbar/Navbar.tsx
```diff
@@ -12,13 +12,16 @@ import { IconName } from '../../../../../../component-library/components/Icons/I
 export function getNavbar({
```
OGPoyraz, line 12: Do we want to add `headerRight` component and make "Cancel" button for claim screen as we discussed? 
We can copy paste from here: https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/Navbar/index.js#L1982

OGPoyraz (approved):

PR #14175, opened 2025-03-21
feat: Enforce 3-minute idle time for BrowserTabs before unmounting them

smilingkylan:
## **Description**

The purpose of this task is to enforce a 3-minute idle time on on BrowserTabs before unmounting them. This is to save memory, among other things. 

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/2159

## **Manual testing steps**

1. Go to the in-app browser and open more than one tab
2. Keep one tab active for over 3 minutes
3. Switch to any other (non-active) tab and notice that the browser starts from scratch (and goes to correct URL)
4. Now switch to any other tab that has been active within the last 3 minutes. The browser should *not* start from scratch... the webpage should still be loaded with webpage state preserved.

### **Before**

https://github.com/user-attachments/assets/877bab7f-ae81-4bbe-a0fa-9b731cf41ebf

### **After**

**WITH 10-SECOND IDLE TIME MAX**

https://github.com/user-attachments/assets/7a0ba625-1f3e-4955-9b30-c61287ab81c5


app/components/Views/Browser/index.js
```diff
@@ -105,11 +105,50 @@ export const Browser = (props) => {
...
+    updateTabInfo(tab.id, {
+      url: tab.url,
+      isArchived: false,
+    });
   };
 
   const hasAccounts = useRef(Boolean(accounts.length));
 
+  useEffect(() => {
+    const interval = setInterval(() => {
```
Cal-L, line 117: Just wanted to note - Since this runs on fixed intervals that are not necessarily in sync with browser tab switching activity, the inactivity threshold isn't 100% accurate. But I think the trade off here is that we only have one process checking instead of an interval per tab, which is better for performance.

app/components/Views/Browser/index.js
```diff
@@ -319,8 +358,8 @@ export const Browser = (props) => {
     return null;
   };
 
-  const renderBrowserTabs = () =>
-    tabs.map((tab) => (
+  const renderBrowserTabWindows = () =>
```
Cal-L, line 361: Let's wrap this in a useCallback to save on performance

Cal-L (commented):
Left a comment

app/components/Views/Browser/index.js
smilingkylan: Yes, as far as I am concerned the 3-min idle time is a heuristic more than anything. Not necessary to hit 3-min right "on the dot".

Cal-L (approved):
Code lgtm!

PR #14088, opened 2025-03-18
fix: cp-7.44.0 Update traits when tracked settings change 2/2

NicolasMassart:
## **Description**

This PR allows to instantly update user traits used in `app/util/metrics/UserSettingsAnalyticsMetaData/generateUserProfileAnalyticsMetaData.ts` when a property used in the traits is changed.
This prevents having to wait for user to background/foreground the app to have the metrics profile updated.

Updated traits are:

- [x] - ENABLE_OPENSEA_API: preferencesController.displayNftMedia
- [x] - NFT_AUTODETECTION: preferencesController.useNftDetection
- [x] - THEME: from app style
- [x] - TOKEN_DETECTION: preferencesController.useTokenDetection
- [x] - MULTI_ACCOUNT_BALANCE: preferencesController.isMultiAccountBalancesEnabled
- [x] - SECURITY_PROVIDERS: if preferencesController.securityAlertsEnabled it return 'blockaid' otherwise empty string

>[!note]
> THEME can be updated only on app foregrounding as it requires to go on system settings to change, forcing app background then foreground. We already have this on app state listener. So no code change, but we still have manual test process in this PR.

Out of scope:
- [x] - HAS_MARKETING_CONSENT: reduxState?.security?.dataCollectionForMarketing (already handled in https://github.com/MetaMask/mobile-planning/issues/1459)

### What was updated
- addUserTraits calls with correct properties
- some logic for associated track events
- existing unit tests when track events values were fixed
- some code cleanup/simplification when possible and directly related to the changes
- some more new unit tests for coverage to pass

## **Related issues**

Fixes https://github.com/MetaMask/mobile-planning/issues/2150

## **Manual testing steps**

Start testing with a fully onboarded app with user optin for metrics.

### Testing ENABLE_OPENSEA_API and NFT_AUTODETECTION
- Go in security and privacy settings and change settings
- Check that IDENTIFY events with these props are sent with proper ON/OFF values
Screenshot 2025-03-19 at 11 08 02

#### Display NFT Media

>[!note]
> There's no track event here, no event defined for this and this may be something product could be interested in adding in the future, but for now, no track.

>[!important]
> Initial state should be both switches to ON for the test (change if needed).
> Both ON is the expected state after a regular onboarding

##### Switch Display NFT Media to OFF
- Switch Display NFT Media to OFF
- IMPORTANT: When it's your first time switching one of these two options OFF, you will have a modal appear:
- image
- Press "Allow"
- Expected event log:
  ```
[...]

PR #14189, opened 2025-03-21
feat: add profile id metametrics trait

mathieuartu:
## **Description**

This PR adds `profile_id` as a MetaMetrics trait.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/IDENTITY-58

## **Manual testing steps**

1. No manual testing steps

### DIFF
--- a/app/util/metrics/UserSettingsAnalyticsMetaData/UserProfileAnalyticsMetaData.types.ts
+++ b/app/util/metrics/UserSettingsAnalyticsMetaData/UserProfileAnalyticsMetaData.types.ts
@@ -11,6 +11,7 @@ export enum UserProfileProperty {
   PRIMARY_CURRENCY = 'primary_currency',
   CURRENT_CURRENCY = 'current_currency',
   HAS_MARKETING_CONSENT = 'has_marketing_consent',
+  PROFILE_ID = 'profile_id',
 }
 
 export interface UserProfileMetaData {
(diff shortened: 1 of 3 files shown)


NicolasMassart (commented):
Where is this profile_id generated and set for the first time?
We should have a call to addUserTrait when it's set or updated.
Eventually if you don't have this updated in this PR, it will have to be added in https://github.com/MetaMask/metamask-mobile/pull/14088

mathieuartu:
> Where is this profile_id generated and set for the first time? We should have a call to addUserTrait when it's set or updated. Eventually if you don't have this updated in this PR, it will have to be added in #14088

@NicolasMassart I was under the impression that it would eventually happen at [this point](https://github.com/MetaMask/metamask-mobile/blob/711a583de3319a7f7773e31150f06b99361c20c9/app/core/AppStateEventListener.ts#L48-L82)

Additionally, searching the project for `addUserTrait` does not return any single occurrence. 
I'm not sure what you're asking here, should I wait for your mentioned PR to be merged before doing anything?

NicolasMassart:
> > Where is this profile_id generated and set for the first time? We should have a call to addUserTrait when it's set or updated. Eventually if you don't have this updated in this PR, it will have to be added in #14088
> 
> @NicolasMassart I was under the impression that it would eventually happen at [this point](https://github.com/MetaMask/metamask-mobile/blob/711a583de3319a7f7773e31150f06b99361c20c9/app/core/AppStateEventListener.ts#L48-L82)

It will indeed happen here but only if user comes back to the app or opens it. So ideally the user trait is updated when you actually update the value. The app event listener entrypoint is a workaround that makes sure the value is updated if user crashes the app or never returns for values that are not updated at the moment they are changed. But with #14088 this should not be required, so ideall all new values should update when they are set.

> Additionally, searching the project for `addUserTrait` does not return any single occurrence. I'm not sure what you're asking here, should I wait for your mentioned PR to be merged before doing anything?

Yes, sorry, I typed it by memory, it's `addTraitsToUser`

PR #14201, opened 2025-03-23
chore: bump assets controllers v55.0.1

sahar-fehri:
## **Description**
This PR bumps assets-controllers to v55.0.1 and introduces the network filter to the NFT tab.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to NFT tab and notice the network filter bar.
2. You should be able to see your NFTs detected on Ethereum and linea. Make sure you see the correct NFTs and the right 
network icon.
3. Switch to polygon and import the NFTs you have on polygon, make sure they are imported properly.
4. Click on the NFTs and see the NFT  details page


app/components/hooks/useGetChainIdsToDetectNfts.ts
```diff
@@ -0,0 +1,33 @@
...
+import { useSelector } from 'react-redux';
+import {
+  selectAllPopularNetworkConfigurations,
+  selectChainId,
+  selectIsAllNetworks,
+  selectIsPopularNetwork,
+} from '../../selectors/networkController';
+import { Hex } from '@metamask/utils';
+
+export const useGetChainIdsToDetectNfts = (): Hex[] => {
```
Prithpal-Sooriya, line 11: Nice reusable hook!

app/components/Base/RemoteImage/index.js
```diff
@@ -1,4 +1,5 @@
-import React, { useEffect, useState } from 'react';
+/* eslint-disable no-console */
```
Prithpal-Sooriya, line 1: Do we still need this disable comment?

app/components/hooks/useGetChainIdsToDetectNfts.test.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useSelector: jest.fn(),
+}));
+
+describe('useGetChainIdsToDetectNfts', () => {
+  it('should return current chain id when isAllNetworks is false', async () => {
+    (useSelector as jest.Mock).mockImplementation((selector) => {
```
Prithpal-Sooriya, line 32: (nit, can ignore) I think we could replace this with mocking the selectors directly instead of mocking the `useSelector`, this way we own't need to have all these if checks?

Prithpal-Sooriya (dismissed):
LGTM

app/components/Base/RemoteImage/index.js
sahar-fehri: removed

patches/@metamask+assets-controllers+55.0.1.patch
Prithpal-Sooriya, line 1: This patch is still required as we need to validate (on extension) and merge the fix into CORE.
https://github.com/MetaMask/core/pull/4774

Prithpal-Sooriya (dismissed):

app/components/hooks/useGetChainIdsToDetectNfts.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  },
+};
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useSelector: jest.fn(),
+}));
+
+describe('useGetChainIdsToDetectNfts', () => {
+  it('should return current chain id when isAllNetworks is false', async () => {
```
NicolasMassart, line 31: S: "isAllNetworks is false" is the implementation, explain the goal of this test instead. When is isAllNetworks false ? Is this when token filter is showing all networks? If so, isAllNetworks is false means only the current network?
If that's right, try:
```suggestion
  it('returns current chain id when filtered on current network', async () => {
```
this allows others to understand the meaning, the goal of you test.  The implementation and how to get this info is in the test itself and it's selector mock.

app/components/Base/RemoteImage/index.js
```diff
@@ -65,8 +69,9 @@ const RemoteImage = (props) => {
   const isImageUrl = isUrl(props?.source?.uri);
   const ipfsGateway = useIpfsGateway();
   const styles = createStyles();
-  const chainId = useSelector(selectChainId);
-  const ticker = useSelector(selectEvmTicker);
+  const currentChainId = useSelector(selectChainId);
+  // The chainId would be passed in props from parent for collectible media
+  const chainId = props.chainId ? toHex(props.chainId) : currentChainId;
```
NicolasMassart, line 74: S: Why not typing the prop to be hex format directly so the conversion would happen when we pass it, making it more consistent on all components without the need to check and convert in the component?

NicolasMassart (changes requested):
Some suggestions for matching our guidelines and make code more simple to understand

app/components/Base/RemoteImage/index.js
sahar-fehri: Correct, i think i initially wanted to avoid this because making this change would require more refactoring to make the file ts instead of Js;
and to update the current declaration of props to an interface and declare `chainId: Hex`

```
RemoteImage.propTypes = {
  /**
   * Flag that determines the fade in behavior
   */
  fadeIn: PropTypes.bool,
  /**
   * Source of the image
   */
  source: PropTypes.any,
  /**
   * Style for the image
   */
  style: ViewPropTypes.style,
  /**
   * Style for the placeholder (used for fadeIn)
   */
  placeholderStyle: ViewPropTypes.style,
  /**
   * Called when there is an error
   */
  onError: PropTypes.func,
  /**
   * This is set if we know that an image is remote
   */
  isUrl: PropTypes.bool,
  /**
   * Token address
   */
  address: PropTypes.string,

  isTokenImage: PropTypes.bool,

  isFullRatio: PropTypes.bool,
  chainId: PropTypes.string,
};
```

patches/@metamask+assets-controllers+55.0.1.patch
```diff
@@ -135,7 +135,7 @@ index a390e86..4d0346d 100644
 +    if (blockchainMetadata?.error && nftApiMetadata?.error) {
 +        return {
 +            image: null,
-+            name: null,
++           name: null,
```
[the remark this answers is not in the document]
sahar-fehri: Indent was not intentional here

app/components/Base/RemoteImage/index.js
NicolasMassart: To be honest, I did not notice it was not TS 😅 

Ideally we should never add new code to a JS file. If needed, making it a TS file first is the perfect move.
Adding some code logic to avoid refactor to TS is probably a dangerous thing to do (more logic = more bugs)

app/components/UI/CollectibleContractElement/index.test.tsx
```diff
@@ -147,4 +147,29 @@ describe('CollectibleContractElement Snapshot', () => {
 
     expect(onPressMock).toHaveBeenCalled();
   });
+
+  it('should setCollectiblesVisible to false', async () => {
```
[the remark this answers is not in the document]
NicolasMassart: For what I read on this onPress, if calls onItemPress in app/components/UI/CollectibleContracts/index.js that navigates to collectible detail:
https://github.com/MetaMask/metamask-mobile/blob/af152f91346c4f757b15a089041491284239cf76/app/components/UI/CollectibleContracts/index.js#L127-L133
then
https://github.com/MetaMask/metamask-mobile/blob/af152f91346c4f757b15a089041491284239cf76/app/components/UI/CollectibleContracts/index.js#L93-L95

If this is the case, it's not hiding, but just moving to something else so yes, the original component is not visible anymore.
Should this test look for the navigation call instead?
Am I not understanding this properly?

NicolasMassart (commented):
resolved most convos

app/components/UI/CollectibleContractElement/index.test.tsx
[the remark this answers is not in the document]
sahar-fehri: The testId we are pressing in the test would call this fct
```
  const toggleCollectibles = useCallback(() => {
    setCollectiblesVisible(!collectiblesVisible);
  }, [collectiblesVisible, setCollectiblesVisible]);
```
Which should only hide the NFT and not navigate to NFT details, I think the one that you are referring to is a different press event which indeed navigates to NFT details page.

app/components/Base/RemoteImage/index.js
sahar-fehri: Agree with that, we have been refactoring tasks to update code/convert js to ts; i think it would be better to not make this change as part of this PR yet 🙏

app/components/Base/RemoteImage/index.js
```diff
@@ -132,17 +137,33 @@ const RemoteImage = (props) => {
     );
   }, [uri]);
 
-  const NetworkBadgeSource = () => {
+  const NetworkBadgeSource = useCallback(() => {
```
gambinish, line 140: 👍

gambinish (dismissed):

Prithpal-Sooriya (dismissed):

app/components/UI/CollectibleContractElement/index.test.tsx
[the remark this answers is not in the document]
NicolasMassart: The component structure is a bit too complex IMO, it requires very intense reading to understand.
Ideally it should be simplified or have more obvious naming on its parts.
We have a mix of "collectible", "contract" and "element" and it's not clear to me what should be the parent of the other (list of elements and elements).

I see the toggleCollectibles now and it seems to hide a list of collectible elements, so maybe we should have the term "list" in the parent element that contains all the other elements.

So for what I understand now, I would suggest something like "hides collectibles list when pressed"

app/components/Base/RemoteImage/index.js
NicolasMassart: Ok, but don't include artificial and cluttering code check only because it's JS either then.
Check if the component can be called with a prop that is not hex, and fix the value at caller level then take it for granted in your component that it's hex. Will ease migration to TS later. Otherwise the risk is that the useless check logic is also moved to TS.
Last resort, add code comments like a `//TODO remove once migrated to TS and chainID is properly typed to hex`.

Prithpal-Sooriya (approved):

salimtb (approved):

NicolasMassart (approved):
Looks good to me.
Thanks for this great contribution!

PR #12180, opened 2024-11-05
fix: trackevent enabled is undefined

NicolasMassart:
## **Description**

> [!IMPORTANT]
> The issue is due to the fact that `MetaMetrics.trackEvent` method has multiple signatures to handle the backward compatibility with old ways to call it.
> After discussion (see the change history of this description), we decided to remove backward compatibility and the multiple signature system. We are going to simplify and then fix all the `trackEvent` calls.
Asside from the changes in `MetaMetrics` and `useMetrics` hook (and tests) all the changes should be on the calls of `trackEvent`.

### Current changes in this PR
- removes multiple signature `MetaMetrics.trackEvent`
- updates `MetaMetrics` unit tests
- updates `useMetrics` hook
- updates `useMetrics` hook unit tests
- deletes now useless legacy compatibility utils
- updates all `trackEvent` calls
- updates all unit tests that test `trackEvent` calls

## **Related issues**

Fixes #12117 

## **Manual testing steps**

1. navigate the app
2. check trackEvent is called (check app logs)

## **Screenshots/Recordings**

https://github.com/user-attachments/assets/058f6607-eda1-4eb5-bcae-a774deb13648

https://github.com/user-attachments/assets/fa0816cf-5459-4825-a162-a9a2e87c86ac

### **Before**

N/A

### **After**

N/A

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #14205, opened 2025-03-24
fix: return the same track event reference

bergarces:
## **Description**

Calling `trackEvent.bind` inside the `useMetrics` hook generates a new reference to the function every time the hook is called. This causes unnecessary re-renders when `trackEvents` is used as a dependency.

Calling `bind` was necessary because the method was not defined as an arrow function, so when called as a standalone function, `this` becomes `undefined`. Having it implemented as an arrow function like the rest of the methods of that class fixes the issue.

## **Related issues**

Fixes:

## **Manual testing steps**

Should not create any functionality changes


app/core/Analytics/MetaMetrics.ts
```diff
@@ -655,11 +655,11 @@ class MetaMetrics implements IMetaMetrics {
    * @param event - Analytics event built with {@link MetricsEventBuilder}
    * @param saveDataRecording - param to skip saving the data recording flag (optional)
    */
-  trackEvent(
+  trackEvent = (
```
Prithpal-Sooriya, line 658: Ok nice, so this is to help preventing the binding approach we did before?

Prithpal-Sooriya (dismissed):

app/components/hooks/useMetrics/useMetrics.test.tsx
```diff
@@ -153,4 +153,35 @@ describe('useMetrics', () => {
...
+  it('keeps the same reference to the functions after every call', async () => {
+    const {
+      result: { current: firstResult },
+    } = renderHook(() => useMetrics());
+    const {
+      result: { current: secondResult },
+    } = renderHook(() => useMetrics());
+
+    expect(firstResult.trackEvent).toBe(secondResult.trackEvent);
+    expect(firstResult.enable).toBe(secondResult.enable);
```
Prithpal-Sooriya, line 166: Nice equality checks!

Prithpal-Sooriya: Could we clean up the tests to instead do this?
```ts
  Object.keys(firstResult).forEach((key) => {
    expect(firstResult[key]).toBe(secondResult[key]);
  });
```

bergarces: Good call. It'll also work out of the box if new entries are added to the hook.

Prithpal-Sooriya (approved):

app/core/Analytics/MetaMetrics.ts
bergarces: Yes.

Arrow function methods inherit `this` from the context in which they were defined, regardless of whether they are called from the object or as standalone.

Non-arrow function methods will set this to undefined if they are called as standalone functions, thus why this method needed binding.

NicolasMassart (approved):
Very nice catch!
This non arrow function definition was a remain from the time I tired to have multiple signatures to keep backward compatibilty, but that was a [mistake we fixed](https://github.com/MetaMask/metamask-mobile/pull/12180). However, we kept the function syntax as is. Thanks for this improvement.

NOTE: one next nice thing could be to move to pure functional code (no class) but we will build a controller soon so this will be part of the job.

PR #14243, opened 2025-03-26
fix: alignment of Snaps icons within ButtonLink/SnapUIButton component

### LINKED ISSUE #3244: Alignment is off on some SnapUIButtons
Image

Notice alignment is off on the X and the "swap" icon in this screenshot.

### LINKED ISSUE #3245: Alignment is off on SnapUILink icons
Image

Alignment might also be off inside the row?

Daniel-Cross:
## **Description**

Snaps sometimes passes in Icons as part of a button, such as a close or swap button. They are currently wrapped in multiple `<Text/>` components which pushes the alignment out. The ways to address this is to remove the text component wrappers but that would also involve making changes to the DS component `ButtonLink`, which would affect the rest of the app.

`ButtonLink` wraps any child passed in with a text component if `ButtonSize` is set to `Auto`. When set to `Sm`, `Md`, or `Lg`, and the text component wrapper is removed from the `SnapUIButton` component the Icon is correctly aligned, however `paddingHorizonal: 16` is then applied to the Icon which forces the Icon button to take up more space in text inputs and the layout in general.

It currently seems the only way to address this issue is to make changes to the DS component or build a custom button component within `SnapUIButton` to fit the needs of Snaps specifically. 

## **Related issues**

Fixes: https://github.com/MetaMask/snaps/issues/3244

## **Manual testing steps**

1. Add a Solana account
2. Press `Send`
3. Add a wallet and see the icons appear within the text inputs

### **Before**

Screenshot 2025-03-25 at 14 27 14

### **After**

Simulator Screenshot - iPhone 16 Pro - 2025-03-26 at 11 34 15


app/components/Snaps/SnapUIButton/SnapUIButton.tsx
```diff
@@ -55,33 +70,80 @@ export const SnapUIButton: FunctionComponent<
...
+
+    if (typeof children === 'string') {
+      return (
+        <Text color={color} variant={textVariant}>
+          {children}
+        </Text>
+      );
+    }
+
+    return children;
```
FrederikBolding, line 132: One of the problems with this is that we don't inherit the text color when complex children are used:
Screenshot 2025-03-27 at 17 06 30

Also for some reason this text is now center aligned?

Can be reproed by rendering
```jsx
<Box>
  <Button>Foo <Icon name="arrow-right" /> Bar</Button>
</Box>
```

Daniel-Cross: Thanks! Let me take a look and make some changes.

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
```diff
@@ -55,33 +70,80 @@ export const SnapUIButton: FunctionComponent<
   };
 
   const overriddenVariant = disabled ? 'disabled' : variant;
-
   const color = COLORS[overriddenVariant as keyof typeof COLORS];
 
+  const isIcon =
+    React.isValidElement(children) &&
+    (children.props?.type === 'Icon' ||
+      (children.type as React.ComponentType<unknown> & { displayName?: string })
```
FrederikBolding, line 78: What is going on here? What even is `.displayName`?

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
```diff
@@ -24,28 +36,31 @@ const COLORS = {
...
 
   const handlePress = () => {
+    onPress?.();
+
     handleEvent({
       event: UserInputEventType.ButtonClickEvent,
       name,
     });
 
-    // Since we don't have onSubmit on mobile, the button submits the form.
```
FrederikBolding, line 48: Why are we removing this documentation?

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
```diff
@@ -129,6 +124,36 @@ export const SnapUIButton: FunctionComponent<SnapUIButtonProps> = ({
...
+          }
+          return section;
+        });
+
+        return React.cloneElement(
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          children as React.ReactElement<{ sections: any }>,
+          {
+            sections: modifiedSections,
+          },
```
FrederikBolding, line 150: Seems simpler to try and solve this at the mapper level before we turn it into React components to prevent having to do this

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
Daniel-Cross: This seems to be resolved now

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
Daniel-Cross: Re-added this comment

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
Daniel-Cross: Yeah this has been removed, I was experimenting with something that didn't work.

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
Daniel-Cross: I took a look at moving this over to the `button.ts` file and it's possible to map the children there correctly instead of in the `SnapUIButton.tsx` component, however we still have to map the children again in this component to apply the correct colour to the text as we don't have access to the Theme colours, which is a React hook, and not accessible in the `button.ts` file.

Not really sure how to solve both the alignment issue (multiple Text component wraps) and the text colour inheritance issue at this point if we need this to be in the `button.ts` mapper component.

I can perhaps create some sort of theme utility that doesn't rely on react, or I can pass them through via params from the `utils.ts` file via the `mapToTemplate` function?

FrederikBolding: > however we still have to map the children again in this component to apply the correct colour to the text as we don't have access to the Theme colours, which is a React hook, and not accessible in the button.ts file.

We do! https://github.com/MetaMask/metamask-mobile/blob/a9ccd4949f616f5f6ba84234c75062cab494272e/app/components/Snaps/SnapUIRenderer/components/section.ts#L7

Daniel-Cross: Awesome, this might work then

app/components/Snaps/SnapUIRenderer/components/button.ts
```diff
@@ -6,36 +6,215 @@ import {
...
+    enumColor === TextColor.Info
+      ? theme.colors.primary.default
+      : enumColor === TextColor.Error
+      ? theme.colors.error.default
+      : theme.colors.text.muted;
+
+  /**
+   * Apply styles to a component based on its type
+   */
+  function styleComponent(component: UIComponent): UIComponent {
```
FrederikBolding, line 98: I think we can avoid all of this with https://github.com/MetaMask/metamask-mobile/pull/14355

PR #14271, opened 2025-03-27
chore: added earn team as codeowners for earn directories

Matt561:
Codeowners file has been updated so that earn directories are owned by @MetaMask/metamask-staking


Cal-L (approved):
Lgtm!

PR #14278, opened 2025-03-27
feat: Multichain account list balance

owencraston:
## **Description**

This PR enables the account balances for non evm accounts (Solana) to be renders in the Account Selector List. This change was achieved by leveraging the existing logic in the `useMultichainBalances` hook which is already used in production to display both evm and non evm account balances for the selected account on the main wallet page. I had to expand this logic to not only derive the balance for the selected account on the selected chain but instead to derive the balances for all accounts. This way we can lookup the `MultichainBalancesData` (defined in useMultichainBalances.types.ts) in a map where the key is the account ID. The key changes that enabled this change were the following...

1. multichain.ts
    - refactor some of the balance selector to use util functions so that they could be reused. 
    - created a new selector called `selectMultichainNetworkAggregatedBalanceForAllAccounts` which, uses the same logic as `selectSelectedAccountMultichainNetworkAggregatedBalance` except for all accounts and returns a map of account ids and `MultichainNetworkAggregatedBalance`

2. useMultichainBalances.ts
    - modified the return type of this hook so that it returns a map of account balance data and account ids as well as one single object for the selected account. The return type looks like this...
    
```ts
export interface UseMultichainBalancesHook {
  multichainBalancesForAllAccounts: Record<
    InternalAccount['id'],
    MultichainBalancesData
  >;
  selectedAccountMultichainBalance?: MultichainBalancesData;
}
```
   - I also had to decouple the the hook from the currently selected chainId. This means that we must select the evmChainId (either currently selected or previously selected chainId), and the nonEvmChainId (either currently selected or previously selected)

3. useAccounts.ts
    - removed all of the balance computation inside the useAccounts hook (mostly through the `getAccountBalances` util that I removed) in favour of the results from the useMultichainBalances hook. 
    - Memoized many of the functions in this hook.

4. Engine.ts
    - renamed `getTotalFiatAccountBalance` to `getTotalEvmFiatAccountBalance`
    - Modified `getTotalEvmFiatAccountBalance`  to take an optional InternalAccount param that allows this function to derive the evm balance for a specified account not just the currently selected account
    - Modified the `getTotalEvmFiatAccountBalance` to use `TokenController.allTokens` instead of `TokenController.tokens` which is tied to the currently selected account

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/697

## **Manual testing steps**

1. ensure you are building flask by verifying that your .js.env file contains export METAMASK_BUILD_TYPE="flask"
2. yarn setup && yarn start:ios
3. Import an account that has a balance, I can provide you with an SRP from 1Password if needed.
4. click on "add account or hardware wallet"
5. click on Solana account
6. Once the account is created, open the account selector list
7. The balance that is rendered on the right side of the the cell should match the balance rendered on the main wallet view. It should show your total fiat balance which is the sum of all of your native tokens (ETH/SOL) plus the sum of all of your token balances (ERC20/SPL tokens). This fiat balance should be in your currently selected currency and it should also show your native token balance.
8. Changing your selected account between non evm and evm should not change the balances you see in the account list. 
9. Both EVM and Solana balances should be correct
10. changing your selected currency should be reflected in balance in the accounts list. 
11. If you switch the network to a test net, the fiat balance should not be rendered and instead you should see your token balance.

### **Before**

<image src="https://github.com/user-attachments/assets/e1952e3e-4547-415d-bd2e-90843565a204" height="600" width="300" />

https://github.com/user-attachments/assets/762f9946-f19f-499d-b140-50226c694c9a

### **After**

<image src="https://github.com/user-attachments/assets/13b073f6-8708-4c0f-9afa-f6f2eadd9ebb" height="600" width="300" />

<image src="https://github.com/user-attachments/assets/89ece564-2535-419c-82e4-09d1667018ed" height="600" width="300" />

<image src="https://github.com/user-attachments/assets/60db4406-9137-4084-b7ad-39b7c06672da" height="600" width="300" />


app/components/hooks/useAccounts/useAccounts.ts
```diff
@@ -145,90 +108,97 @@ const useAccounts = ({
...
+      };
+    });
+
+    return balances;
+  }, [internalAccounts, multichainBalancesForAllAccounts, checkBalanceError]);
+
+  const getAccounts = useCallback(() => {
+    if (!isMountedRef.current) return;
+    // Keep track of the Y position of account item. Used for scrolling purposes.
+    let yOffset = 0;
```
gambinish, line 142: Are we implementing infinite scroll here?

app/selectors/multichain/multichain.ts
```diff
@@ -289,50 +309,128 @@ export const selectMultichainTokenList = createDeepEqualSelector(
...
+    const balanceInFiat =
+      balance.amount && rate
+        ? new BigNumber(balance.amount).times(rate)
+        : new BigNumber(0);
+
+    // Only update native token balance if this is the native asset
+    if (assetId === nativeAsset) {
+      totalNativeTokenBalance = {
+        amount: balance.amount || '0',
+        unit: balance.unit || '',
```
gambinish, line 355: Shouldn't we fallback to `undefined` here rather than `'0'`? Unless the zero value is expected. I think having this `undefined` may retain our skeleton loader if that's still the intended behavior

app/selectors/multichain/multichain.ts
```diff
@@ -289,50 +309,128 @@ export const selectMultichainTokenList = createDeepEqualSelector(
...
+      selectedAccount,
+      multichainBalances,
+      assets,
+      assetsRates,
+      nonEvmNetworkChainId,
+    ): MultichainNetworkAggregatedBalance => {
+      if (!selectedAccount) {
+        return {
+          totalNativeTokenBalance: { amount: '0', unit: '' },
+          totalBalanceFiat: 0,
```
gambinish, line 387: Same note about 0 fallback here vs `undefined` fallback

app/components/hooks/useGetFormattedTokensPerChain.tsx
```diff
@@ -79,95 +80,115 @@ export const useGetFormattedTokensPerChain = (
   const currencyRates = useSelector(selectCurrencyRates);
   const showFiatOnTestnets = useSelector(selectShowFiatInTestnets);
 
-  //If the current network is a testnet, UI should display 0 unless conversions are enabled
-  const validAccounts =
-    accounts.length > 0 && accounts.every((item) => item !== undefined);
-  if (!validAccounts || (isTestNet(currentChainId) && !showFiatOnTestnets)) {
-    return {};
-  }
+  return useMemo(() => {
```
gambinish, line 83: What was the reasoning for not making `useGetFormattedTokensPerChain` a hook rather than a memoized selector? We're not really dealing with any internal state here, outside of `useSelector`

I wonder if it would be best to consolidate this into a `createDeepEqual` selector instead? This comment doesn't need to block

gambinish (commented):
I've pulled this down and tested. Feature looks to be functioning as intended. I will do one more code review before approving.

app/components/hooks/useMultichainBalances/useMultichainBalances.ts
```diff
@@ -50,140 +62,242 @@ const useMultichainBalances = (): UseMultichainBalancesHook => {
...
-    if (isOriginalNativeTokenSymbol) {
-      if (isPortfolioViewEnabled()) {
+      if (isOriginalNativeEvmTokenSymbol) {
+        if (isPortfolioEnabled) {
+          total =
+            totalFiatBalancesCrossEvmChain[account?.address as string]
+              ?.totalFiatBalance ?? 0;
+        } else {
+          const tokenFiatTotal = balance?.tokenFiat ?? 0;
+          const ethFiatTotal = balance?.ethFiat ?? 0;
```
sahar-fehri, line 104: For exp i think `balance.tokenFiat `should match `totalTokenFiat` from `totalFiatBalancesCrossEvmChain`

sahar-fehri: The native `ethFiat` should also be available in `useGetTotalFiatBalanceCrossChains` and we use it to compute the totalValue but maybe its not being explicitly returned;
I understand this is out of scope of this PR 🙏 , we can keep this in mind to address in future 🙏

app/selectors/multichain/multichain.ts
owencraston: Good point, I modified it the function to return undefined.

app/selectors/multichain/multichain.ts
owencraston: Modified the function to default to undefined.

app/components/hooks/useAccounts/useAccounts.ts
owencraston: Honestly not super sure whats going on with the yOffset. I think its used in the UI to properly space the accounts list but I am not certain. This was already part of [the hook](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/hooks/useAccounts/useAccounts.ts#L152) so I added it here for parity. Perhaps @Cal-L can shed some light on this.

app/components/hooks/useGetFormattedTokensPerChain.tsx
owencraston: I hadn't really thought of it. I was really just trying to change as little as possible but I had to memoize this function due to rerenders. What benefit would we get from converting it to a hook? I am not really seeing what changes we would make except perhaps adding a loading state.

app/components/hooks/useMultichainBalances/useMultichainBalances.ts
```diff
@@ -50,140 +62,242 @@ const useMultichainBalances = (): UseMultichainBalancesHook => {
...
-  // Production balance calculation (EVM)
-  const getEvmDisplayBalance = () => {
-    const balance = Engine.getTotalFiatAccountBalance();
-    let total;
+  const isPortfolioEnabled = isPortfolioViewEnabled();
+
+  // Production balance calculartion (EVM)
+  const getEvmBalance = useCallback(
+    (account: InternalAccount) => {
+      const balance = Engine.getTotalEvmFiatAccountBalance(account);
```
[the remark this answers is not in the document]
owencraston: Hey @sahar-fehri, This is a good point so did some research on this. It seems like the basic data we get from the `getTotalEvmFiatAccountBalance` is also available in the `totalFiatBalancesCrossEvmChain`. Things like the `totalTokenFiat` and `balance.tokenFiat` as well as the cross evm chain data. The issue I see is that the `totalFiatBalancesCrossEvmChain` does not contain historical data. Here are some logs...

```JSON
 LOG  getEvmBalance balance {
  "ethFiat": 45.31,
  "ethFiat1dAgo": 39.454098314085186,
  "tokenFiat": 1.56,
  "tokenFiat1dAgo": 1.559872973153214,
  "totalNativeTokenBalance": "0.02703",
  "ticker": "ETH"
}
 LOG  getEvmBalance totalFiatBalancesCrossEvmChain {
  "totalFiatBalance": 46.870000000000005,
  "totalTokenFiat": 1.56,
  "tokenFiatBalancesCrossChains": [
    {
      "chainId": "0x1",
      "tokensWithBalances": [
        {
          "address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
          "symbol": "USDC",
          "decimals": 6,
          "balance": "1.5643",
          "tokenBalanceFiat": 1.56
        }
      ],
      "tokenFiatBalances": [
        1.56
      ],
      "nativeFiatValue": 45.31
    }
  ]
}
```

We are missing a few things from the `totalFiatBalancesCrossEvmChain`, namely historical data (`tokenFiat1dAgo` and `ethFiat1dAgo`) that are used to render the percentage changed values on the home page. The other data is the `totalNativeTokenBalance` which I am sure we could easily add to the `totalFiatBalancesCrossEvmChain`. My main concern is really the historical data and the fact that the `Engine.getTotalEvmFiatAccountBalance` is heavily relied upon in the app. The real purpose of this PR is to include multichain balances in our response of `useAccounts` and not refactor the entire "backend" for balances. I only made the minimum changes I needed to enable this. I do agree that the balance logic is a huge mess however I believe this warrants more attention than I can possible deliver in this PR. If you feel strongly that we should deprecate the `Engine.getTotalEvmFiatAccountBalance` then I think we should create a new issue/epic for this and discuss with all the stakeholders what the best pattern would be. What do you think?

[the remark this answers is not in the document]
owencraston: For what its worth, I did try to migrate away from `Engine.getTotalEvmFiatAccountBalance` but it opened up a huge can of worms.

app/components/hooks/useMultichainBalances/useMultichainBalances.ts
owencraston: responded in the above [comment](https://github.com/MetaMask/metamask-mobile/pull/14278#discussion_r2036236443).

app/components/UI/AccountSelectorList/AccountSelector.test.tsx
```diff
@@ -193,38 +244,6 @@ describe('AccountSelectorList', () => {
     });
   });
 
-  it('should render all accounts but only the balance for selected account', async () => {
```
owencraston, line 196: I think this test is invalid as we render all balances. This is how it was before my PR as well.

app/components/UI/AccountSelectorList/AccountSelector.test.tsx
```diff
@@ -193,38 +244,6 @@ describe('AccountSelectorList', () => {
...
-      const businessAccountItem = await queryByTestId(
-        `${AccountListBottomSheetSelectorsIDs.ACCOUNT_BALANCE_BY_ADDRESS_TEST_ID}-${BUSINESS_ACCOUNT}`,
-      );
-
-      expect(within(businessAccountItem).getByText(regex.eth(1))).toBeDefined();
-      expect(
-        within(businessAccountItem).getByText(regex.usd(3200)),
-      ).toBeDefined();
-
-      expect(toJSON()).toMatchSnapshot();
```
owencraston, line 224: Removing this line also removed the snapshot which is one of the reasons the number of lines changed is so large.

app/components/hooks/useAccounts/useAccounts.ts
```diff
@@ -145,90 +108,99 @@ const useAccounts = ({
...
-            getFormattedAddressFromInternalAccount(internalAccount);
-          const isSelected =
-            selectedInternalAccount?.address === internalAccount.address;
-          if (isSelected) {
-            selectedIndex = index;
-          }
-
-          // TODO - Improve UI to either include loading and/or balance load failures.
-          // TODO - Non EVM accounts like BTC do not use hex formatted balances. We will need to modify this to support multiple chains in the future.
-          const { balanceETH, balanceFiat, balanceWeiHex } = getAccountBalances(
```
owencraston, line 166: The main difference in this component is the removal of the `getAccountBalances` util in favour or the data from `useMultichainBalances` This hook now has a memoized `accountBalances` map (starting on line 111) that return the following structure...
```
 const balances: Record<
      string,
      {
        displayBalance: string;
        balanceError: string | undefined;
      }
    >
```

The data for the account balances come fro the `useMultichainBalances.multichainBalancesForAllAccounts`. The benefit of this is that this hook is the same hook that is used to display the balances on the home page so we are relying on the exact same logic/data for both the account list and the home page. This also reduces the number of balances related utils we need to maintain (although admittedly still not great).

app/components/hooks/useAccounts/useAccounts.ts
owencraston: See lines https://github.com/MetaMask/metamask-mobile/pull/14278/files#diff-ea9ffa6b4c13e33bbcf27b1b8ce695299ea11dd6fbdcaefb55310eb29c99ecc5L152-L153

app/components/hooks/useGetTotalFiatBalanceCrossChains.tsx
```diff
@@ -46,100 +47,124 @@ export const useGetTotalFiatBalanceCrossChains = (
...
-        );
-        const totalAccountBalance = balanceBN
-          .add(stakedBalanceBN)
-          .toString('hex');
-        ethFiat = weiToFiatNumber(
-          totalAccountBalance,
-          conversionRate,
-          decimalsToShow,
-        );
+  return useMemo(() => {
```
owencraston, line 50: I changed nothing in this util except memoizing it. I was experiencing a lot of re renders and memoizing this, along with some other selectors solved this issue.

app/core/Engine/Engine.ts
```diff
@@ -1620,11 +1623,15 @@ export class Engine {
     AccountTrackerController.refresh();
   }
 
-  getTotalFiatAccountBalance = (): {
+  getTotalEvmFiatAccountBalance = (
+    account?: InternalAccount,
```
owencraston, line 1627: The key differences here are the following...

1. more specific function name to indicate that this function only works for evm.
2. allow us to pass in a specific account we want to fetch the balance for. This is an optional param that defaults to the currently selected account if no param is passed. 
3. return `totalNativeTokenBalance` and the `ticker`

app/components/hooks/useGetFormattedTokensPerChain.tsx
gambinish: Not a huge deal, I've run into performance issues with overusing hooks in the past that I've gotten out of with `createDeepEqualSelector` since it's using `cloneDeep` under the hood.

Maybe we can just keep this in mind if anything gets flagged re: performance from QA. Will resolve this comment for now.

gambinish (dismissed):
lgtm, thanks for the code comments

pedronfigueiredo (dismissed):

bfullam (dismissed):

app/components/hooks/useAccounts/useAccounts.ts
```diff
@@ -145,90 +108,99 @@ const useAccounts = ({
...
+        };
+
+        const isBalanceAvailable = isMultiAccountBalancesEnabled || isSelected;
+        const mappedAccount: Account = {
+          name: internalAccount.metadata.name,
+          address: formattedAddress,
+          type: internalAccount.metadata.keyring.type as KeyringTypes,
+          yOffset,
+          isSelected,
+          // TODO - Also fetch assets. Reference AccountList component.
```
tommasini, line 168: Is this todo forgotten or will be addressed later?

app/selectors/multichainNetworkController/index.ts
```diff
@@ -15,7 +16,7 @@ export const selectIsEvmNetworkSelected = createSelector(
     multichainNetworkControllerState.isEvmSelected,
 );
 
-export const selectSelectedNonEvmNetworkChainId = createSelector(
+export const selectSelectedNonEvmNetworkChainId = createDeepEqualSelector(
```
tommasini, line 19: Why deepEqual was necessary here?

app/selectors/networkController.ts
```diff
@@ -129,7 +129,7 @@ export const selectProviderConfig = createDeepEqualSelector(
   },
 );
 
-export const selectEvmTicker = createSelector(
+export const selectEvmTicker = createDeepEqualSelector(
```
tommasini, line 132: Why deepEqual was necessary here?

app/selectors/networkController.ts
```diff
@@ -147,7 +147,7 @@ export const selectEvmChainId = createSelector(
   (providerConfig) => providerConfig.chainId,
 );
 
-export const selectChainId = createSelector(
+export const selectChainId = createDeepEqualSelector(
```
tommasini, line 150: Why deepEqual was necessary here?

app/selectors/networkController.ts
```diff
@@ -247,7 +247,7 @@ export const selectAllPopularNetworkConfigurations = createSelector(
   },
 );
 
-export const selectIsPopularNetwork = createSelector(
+export const selectIsPopularNetwork = createDeepEqualSelector(
```
tommasini, line 250: Why deepEqual was necessary here?

app/selectors/multichainNetworkController/index.ts
owencraston: Necessary is a strong word. The reason I added these was to memoize selectors that have large state objects as inputs. So in this specific case, the input to this selector is the entire `MultichainNetworkControllerState` object. This object reference can change without any values actually changing. In this case we should not re render a component.

app/selectors/networkController.ts
owencraston: same response as above, large objects as inputs. I think this is a best practice for more complex selectors.

app/selectors/networkController.ts
owencraston: yeah this one is not needed since the input is just a string. I'll revert it.

app/components/hooks/useAccounts/useAccounts.ts
owencraston: This TODO was not mine and I do not believe I addressed it. I believe this is related to an effort to show assets in the accounts list. This screenshot is from extension and what we are moving towards. 

Screenshot 2025-04-10 at 11 42 21 AM

gambinish (dismissed):

bfullam (dismissed):

pedronfigueiredo (dismissed):

tommasini (dismissed):
After talking with @owencraston it was decided to remove the deep equal selectors on a  follow up PR! LGTM

vinnyhoward (dismissed):
Code changes look good. Approved ✅

gambinish (approved):

bfullam (approved):

pedronfigueiredo (approved):

tommasini (approved):

zone-live (approved):

vinnyhoward (approved):

PR #14326, opened 2025-03-31
chore: bump transaction-controller to 52.3.0

mathieuartu:
## **Description**

This PR bumps `@metamask/transaction-controller` to `^52.3.0`

## **Related issues**

Fixes:
- https://github.com/MetaMask/metamask-mobile/issues/13842
- https://github.com/MetaMask/metamask-mobile/issues/13894
- https://github.com/MetaMask/metamask-mobile/issues/14005
- https://github.com/MetaMask/metamask-mobile/issues/14101
- https://github.com/MetaMask/metamask-mobile/issues/14180

## **Manual testing steps**

1. No manual testing steps

### DIFF
--- a/app/util/transaction-controller/index.ts
+++ b/app/util/transaction-controller/index.ts
@@ -6,6 +6,8 @@ import { Hex } from '@metamask/utils';
 
 import Engine from '../../core/Engine';
 import { NetworkClientId } from '@metamask/network-controller';
+import { selectBasicFunctionalityEnabled } from '../../selectors/settings';
+import { store } from '../../store';
 
 export async function addTransaction(
   transaction: TransactionParams,
@@ -62,29 +64,31 @@ export function speedUpTransaction(
   return TransactionController.speedUpTransaction(...args);
 }
 
-export function startIncomingTransactionPolling(
-  ...args: Parameters<
-    BaseTransactionController['startIncomingTransactionPolling']
-  >
-) {
-  const { TransactionController } = Engine.context;
-  return TransactionController.startIncomingTransactionPolling(...args);
+export function startIncomingTransactionPolling() {
+  const isBasicFunctionalityToggleEnabled = selectBasicFunctionalityEnabled(
+    store.getState(),
+  );
+
+  if (isBasicFunctionalityToggleEnabled) {
+    const { TransactionController } = Engine.context;
+    return TransactionController.startIncomingTransactionPolling();
+  }
 }
 
-export function stopIncomingTransactionPolling(
-  ...args: Parameters<
-    BaseTransactionController['stopIncomingTransactionPolling']
-  >
-) {
+export function stopIncomingTransactionPolling() {
   const { TransactionController } = Engine.context;
-  return TransactionController.stopIncomingTransactionPolling(...args);
+  return TransactionController.stopIncomingTransactionPolling();
 }
 
-export function updateIncomingTransactions(
-  ...args: Parameters<BaseTransactionController['updateIncomingTransactions']>
-) {
-  const { TransactionController } = Engine.context;
-  return TransactionController.updateIncomingTransactions(...args);
+export function updateIncomingTransactions() {
+  const isBasicFunctionalityToggleEnabled = selectBasicFunctionalityEnabled(
+    store.getState(),
+  );
+
+  if (isBasicFunctionalityToggleEnabled) {
+    const { TransactionController } = Engine.context;
+    return TransactionController.updateIncomingTransactions();
+  }
 }
 
 export function updateSecurityAlertResponse(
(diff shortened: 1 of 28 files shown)


Prithpal-Sooriya (dismissed):
Approved from assets perspective.
It seems that the only change impacting us is the new interface for `updateIncomingTransactions` which has been updated.

OGPoyraz:
I will add `skip-sonar-cloud` since most of the complains about just removing parameters from existing functions.
I believe @jpuri agree on this with me since she made the core changes.

OGPoyraz (dismissed):

Prithpal-Sooriya (dismissed):

darkwing (dismissed):

digiwand (dismissed):

vinistevam (dismissed):

package.json
```diff
@@ -411,7 +411,7 @@
     "@metamask/mobile-provider": "^3.0.0",
     "@metamask/object-multiplex": "^1.1.0",
     "@metamask/providers": "^18.3.1",
-    "@metamask/test-dapp": "^9.0.0",
+    "@metamask/test-dapp": "^9.2.0",
```
Cal-L, line 414: What was the reason for bumping the test-dapp?

app/util/transaction-controller/index.ts
```diff
@@ -62,29 +64,31 @@ export function speedUpTransaction(
...
-  ...args: Parameters<BaseTransactionController['updateIncomingTransactions']>
-) {
-  const { TransactionController } = Engine.context;
-  return TransactionController.updateIncomingTransactions(...args);
+export function updateIncomingTransactions() {
+  const isBasicFunctionalityToggleEnabled = selectBasicFunctionalityEnabled(
+    store.getState(),
+  );
+
+  if (isBasicFunctionalityToggleEnabled) {
```
Cal-L, line 88: Since this is now conditional with the basic toggle functionality, what is the expected behavior when the toggle is off? For example, what would the user experience and would it lead to a potentially broken UX?

Cal-L (commented):
Left a comment

package.json
OGPoyraz: New `TransactionController` updates taking dapp suggested gas limit into account if it's settled.

The reason behind bumping test-dapp is to pass e2e. Because there was a typo in some of the transaction examples on their hardcoded gas limits in the old versions of test-dapp. Instead of `0x5208` it was `0x5028`. 
This is fixed in : https://github.com/MetaMask/test-dapp/pull/397/files

So this has to be bumped to e2e to pass. Specifically for e2e/specs/confirmations/send-failing-contract.spec.js

app/util/transaction-controller/index.ts
OGPoyraz: Great callout Cal, I think the idea putting the incoming transactions behind basic functionality toggle is because using accounts API - a service so it shares user IP. 

Screenshot 2025-04-08 at 07 06 18

I agree with you it could lead potential broken UX when there is a incoming transaction but if user toggled off "Basic functionality" we cannot use the service anymore. 

This changes done by @jpuri she might have more opinion on this.

jpuri: @bschorchit : will have more context on the product side of the decision and user experience.

OGPoyraz (dismissed):

OGPoyraz (approved):

yarn.lock
```diff
@@ -2050,41 +2156,95 @@
...
+  resolved "https://registry.yarnpkg.com/@ethersproject/json-wallets/-/json-wallets-5.8.0.tgz#d18de0a4cf0f185f232eb3c17d5e0744d97eb8c9"
+  integrity sha512-HxblNck8FVUtNxS3VTEYJAcwiKYsBIF77W15HufqlBF9gGfhmYOJtYZp8fSDZtn9y5EaXTE87zDwzxRoTFk11w==
+  dependencies:
+    "@ethersproject/abstract-signer" "^5.8.0"
+    "@ethersproject/address" "^5.8.0"
+    "@ethersproject/bytes" "^5.8.0"
+    "@ethersproject/hdnode" "^5.8.0"
+    "@ethersproject/keccak256" "^5.8.0"
+    "@ethersproject/logger" "^5.8.0"
+    "@ethersproject/pbkdf2" "^5.8.0"
```
tommasini, line 2170: Interesting, curious question, this most likely not used, but wondering here if it is!
Do we experience that the speed of a transaction using the @ethersproject could be improved?

This in an dept that I don't think it uses native code, and if this code is running, there is room for a good optimization here!

tommasini (approved):

yarn.lock
OGPoyraz: Great question, @tommasini!
I believe there’s significant room for improvement in the transaction lifecycle, especially since we haven’t had a chance to thoroughly review it yet.

As for your specific question - I can't say anything definitive at this point, but it’s definitely worth measuring as you suggested.

We do have a backlog focused on improving transaction performance, and I think this type of query would fit well there, noted.

Prithpal-Sooriya (approved):
Approved on assets side

salimtb (approved):

NidhiKJha (approved):

PR #14355, opened 2025-04-01
feat: Snaps UI text styling inheritance

### LINKED ISSUE #3244: Alignment is off on some SnapUIButtons

### LINKED ISSUE #3271: Default Cancel button in SnapUI footer on mobile is not displayed correctly
When footer is omitted from a Snap UI code, we display default Cancel button. This button should look properly and have some expected styling like on the extension.

Current look on mobile:
Image

Current look on extension:
Image

Note: Investigate if this is UI issue related to the DS button or problem with Snap button customization.

FrederikBolding:
## **Description**

Add generalized pattern for text style inheritance. Parents are now expected to pass along `textVariant`, `textColor`, `textFontWeight`, `textAlign` or `textSize` in the component mappers to facilitate inheritance. This change is required because `color: 'inherit'` is quite inconsistent in React Native and generally we cannot rely on the standard Text inheritance for all cases. 

This PR fixes some of these props that didn't work at all before and also brings in some changes proposed by Dan to modify `SnapUIButton` to use our own component in an effort to fix alignment. I brought this in as some of my changes broke the existing implementation.

## **Related issues**

Fixes: https://github.com/MetaMask/snaps/issues/3271
Fixes https://github.com/MetaMask/snaps/issues/3244


david0xd (approved):

app/components/Snaps/SnapUIAddress/SnapUIAddress.tsx
```diff
@@ -22,6 +22,7 @@ export interface SnapUIAddressProps {
   truncate?: boolean;
   displayName?: boolean;
   avatar?: boolean;
+  color?: string;
```
Daniel-Cross, line 25: As this is a conditional prop, what happens if `color` isn't passed down?

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
```diff
@@ -54,34 +37,22 @@ export const SnapUIButton: FunctionComponent<
...
+          autoPlay
+          loop
+          // eslint-disable-next-line react-native/no-inline-styles
+          style={{
+            width: 24,
+            height: 24,
+          }}
+        />
+      ) : (
+        children
```
Daniel-Cross, line 54: There was an issue with this previously where the `children` was a string value and not a `<Text />` component, which caused the app to crash. Is this still the case or have we factored that in the mapping component?

app/components/Snaps/SnapUIRenderer/utils.ts
```diff
@@ -130,33 +137,37 @@ export const mapTextToTemplate = (
...
+    | 'textFontWeight'
   >,
 ): NonEmptyArray<UIComponent | string> =>
   elements.map((e) => {
     if (typeof e === 'string') {
       const text = unescapeFn(e);
       const key = generateKey(params.map, Text({ children: text }));
       return {
-        element: 'RNText',
+        element: 'Text',
```
Daniel-Cross, line 156: Is this `Text` a new custom component or are we just leveraging the native `Text` component from RN?

FrederikBolding: We are using the DS `Text` component, since otherwise we can't support the props we need. It boils down to a styled native Text component though

app/components/Snaps/SnapUIAddress/SnapUIAddress.tsx
FrederikBolding: The `Text` component uses the default color for the text variant when `color` is undefined

app/components/Snaps/SnapUIButton/SnapUIButton.tsx
FrederikBolding: Since `mapTextTemplate` converts strings to `Text` components, I believe we should be good!

app/components/Snaps/SnapUIRenderer/utils.ts
Daniel-Cross: Got it, thanks

Daniel-Cross (approved):
Looks good to me but we'll have to test on the current Solana flow to see if there any unexpected issues.

PR #14515, opened 2025-04-08
feat: real time dapp scanning BrowserTab

imblue-dabadee:
## **Description**
This feature will implement real time dapp scanning for phishing protection within BrowserTab.

This is intended to be the alternative solution to EPD and part of the plan to phase out EPD from within Mobile. It will reduce the amount of memory (EPD) that users need as well as provide real time phishing protection.

## **Related issues**

Fixes:

### Test 1: When on a website deemed as safe, it does not have the `PhishingModal` appear
**Variation 1: Happy case:**
1. Go to `portfolio.metamask.io`
2. Wait a couple seconds and confirm that the `PhishingModal` hasn't appeared.

**Variation 2: Begun navigating away the malicious page before the phishing scan has returned AND web content from a valid website has loaded (either partially or completely) into the screen**
1. Go to `holichain.eu` (be cautious as this is a currently live malicious website).
2. Press the home button to navigate to `opensea.io`. It is easier to test this by adding a 2-3 second delay within `isAllowedOrigin`.

### Test 2: When on a malicious website, the `PhishingModal` appears
1. Go to a malicious website: e.g. `dcxminess.live`. 
2. Confirm the `PhishingModal` to appears after a few seconds.

### Test 3: Edge cases
**Variation 1: A website is malicious but fails to resolve due to DNS resolution failures.**
**Expected behavior: The `PhishingModal` does not appear.**
1. Go to `scamsite.xyz`
2. Wait for a few seconds to see that the `PhishingModal` does not appear.

**Variation 2: A website is malicious but Chrome's security blocks the site before the phishing scan finishes.**
**Expected behavior: The `PhishingModal` does not appear.**
1. Go to `cryptcurency.com`
2. Wait for a few seconds on Chrome's security block to see that the `PhishingModal` does appear.


app/components/UI/AccountApproval/index.js
```diff
@@ -287,12 +289,21 @@ class AccountApproval extends PureComponent {
 
   checkUrlFlaggedAsPhishing = (hostname) => {
     const { PhishingController } = Engine.context;
+    const productSafetyDappScanningEnabled = selectProductSafetyDappScanningEnabled(store.getState());
+    if (productSafetyDappScanningEnabled) {
+      // eslint-disable-next-line no-console
+      console.log('Real time dapp scanning enabled');
```
mindofmar, line 295: should remove this

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -293,34 +298,69 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
...
+        }
+        // eslint-disable-next-line no-console
+        console.log(
+          `${new Date().toISOString()} [BrowserTab][isAllowedOrigin] finished scanning: ${urlOrigin}: ${JSON.stringify(
+            scanResult,
+          )}`,
+        );
+        return !(
+          scanResult.recommendedAction === RecommendedAction.Block ||
+          scanResult.recommendedAction === RecommendedAction.Warn
```
mindofmar, line 332: `WARN` should not result in a block page being shown

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -293,34 +298,69 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
...
-      // Update phishing configuration if it is out-of-date
-      // This is async but we are not `await`-ing it here intentionally, so that we don't slow
-      // down network requests. The configuration is updated for the next request.
-      PhishingController.maybeUpdateState();
+      const { PhishingController } = Engine.context;
 
-      const phishingControllerTestResult = PhishingController.test(urlOrigin);
+      if (productSafetyDappScanningEnabled) {
+        // eslint-disable-next-line no-console
+        console.log(
```
mindofmar, line 311: should remove

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -293,34 +298,69 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
...
    */
   const handleNotAllowedUrl = useCallback((urlOrigin: string) => {
+    // eslint-disable-next-line no-console
+    console.log(
+      `${new Date().toISOString()} [BrowserTab][handleNotAllowedUrl] urlOrigin: ${urlOrigin}`,
+    );
+    // Ignore showing the phishing modal if the user is no longer on the page.
+    const currentUrlOrigin = new URLParse(resolvedUrlRef.current).origin;
+    if (currentUrlOrigin !== urlOrigin) {
+      // eslint-disable-next-line no-console
```
mindofmar, line 358: you already know what im gonna say and directly above this too 🤣

.js.env.example
```diff
@@ -112,6 +112,10 @@ export MM_PERMISSIONS_SETTINGS_V1_ENABLED=""
 # Feature flag for Stablecoin Lending UI
 export MM_STABLECOIN_LENDING_UI_ENABLED="true"
 
+export FEATURE_FLAG_PRODUCT_SAFETY_DAPP_SCANNING="false"
+
+# Feature flag for Product Safety Dapp Scanning
+export PRODUCT_SAFETY_DAPP_SCANNING="true"
```
AugmentedMode, line 116: remove this . We only want to keep the top one

app/components/Views/BrowserTab/components/PhishingModal/index.tsx
```diff
@@ -113,16 +111,13 @@ const PhishingModal = ({
    * Go back from phishing website alert
    */
   const goBackToSafety = () => {
-    urlBarRef.current?.setNativeProps({ text: activeUrl });
-
     setTimeout(() => {
       setShowPhishingModal(false);
       setBlockedUrl(undefined);
+      goToUrl(HOMEPAGE_URL);
```
AugmentedMode, line 117: Should this be moved to line 115 above `setShowPhishingModal`

app/components/Views/BrowserTab/components/PhishingModal/index.tsx
```diff
@@ -113,16 +111,13 @@ const PhishingModal = ({
    * Go back from phishing website alert
    */
   const goBackToSafety = () => {
-    urlBarRef.current?.setNativeProps({ text: activeUrl });
```
AugmentedMode, line 116: What did this do ?

app/core/Engine/Engine.test.ts
```diff
@@ -46,13 +50,37 @@ describe('Engine', () => {
   const validAddress = MOCK_ADDRESS_1;
   const mockAccount = createMockInternalAccount(validAddress, 'Test Account');
 
+  beforeEach(() => {
```
AugmentedMode, line 53: do you need this ?

app/core/Engine/Engine.test.ts
```diff
@@ -46,13 +50,37 @@ describe('Engine', () => {
...
+  });
+
   afterEach(() => {
     jest.restoreAllMocks();
     (backupVault as jest.Mock).mockReset();
   });
 
   it('should expose an API', () => {
-    const engine = Engine.init({});
+    const engine = Engine.init({
```
AugmentedMode, line 78: do you need this ?

app/core/Engine/Engine.test.ts
```diff
@@ -28,6 +28,10 @@ jest.mock('../../selectors/smartTransactionsController', () => ({
 jest.mock('../../selectors/settings', () => ({
   selectBasicFunctionalityEnabled: jest.fn().mockReturnValue(true),
 }));
+jest.mock('../../util/phishingDetection', () => ({
```
AugmentedMode, line 31: This is the only thing i added. so if you dont need any of the below you should remove

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -294,18 +293,19 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
...
+      if (whitelisted) {
         return true;
       }
-      const phishingResult = getPhishingTestResult(urlOrigin);
-      // Is safe if no phishing result or if phishing result is false
-      const isSafe = !phishingResult?.result;
-      if (phishingResult && !isSafe && phishingResult.name) {
-        blockListType.current = phishingResult.name;
+
+      const testResult = await getPhishingTestResultAsync(urlOrigin);
```
AugmentedMode, line 302: Doesnt this await cause the browser the block the page load ?

AugmentedMode:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=metamask-mobile&pullRequest=14515) **Quality Gate failed**
> Failed conditions  [30.5% Coverage on New Code](https://sonarcloud.io/component_measures?id=metamask-mobile&pullRequest=14515&metric=new_coverage&view=list) (required ≥ 80%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=metamask-mobile&pullRequest=14515)

You will need to hit test coverage in order to merge

app/util/phishingDetection.ts
```diff
@@ -1,14 +1,64 @@
...
   const { PhishingController } = Engine.context as {
     PhishingController: PhishingControllerClass;
   };
+
+  if (isProductSafetyDappScanningEnabled()) {
+    const scanResult = await PhishingController.scanUrl(origin);
+    return {
+      result:
+        scanResult.recommendedAction === RecommendedAction.None ||
+        scanResult.recommendedAction === RecommendedAction.Warn,
```
AugmentedMode, line 57: The result should only return true when the reccomendedAction is block right ?
```suggestion
scanResult.recommendedAction === RecommendedAction.Block,
```

app/components/Views/BrowserTab/BrowserTab.tsx
imblue-dabadee: It does not

app/util/phishingDetection.ts
imblue-dabadee:
```typescript
/**
 * Represents the result of checking a domain.
 */
export type PhishingDetectorResult = {
    /**
     * Whether the domain is regarded as allowed (true) or not (false).
     */
    result: boolean;
    // other variables
 }
 ```
 Hence, we want result to be `true` if the website has no confident flags of being malicious. This is true when the scan result's `recommendedAction` is either `NONE` or `WARN`.

**Comment by AugmentedMode** [comment]
i found a bug where if you navigate to a malicious site such as `j-node.net` it will correctly show the phishing screen. But if you click `back to safety` then navigate to the malicious site again i get this error and i do not get the phishing screen again 

```
'scanResult', { domainName: '',
  recommendedAction: 'NONE',
  fetchError: 'timeout of 8000ms exceeded' }
  ```

app/util/phishingDetection.ts
```diff
@@ -7,20 +12,52 @@ import { selectProductSafetyDappScanningEnabled } from '../selectors/featureFlag
...
+  };
+
+  if (isProductSafetyDappScanningEnabled()) {
+    const scanResult = await PhishingController.scanUrl(origin);
+    return {
+      result:
+        scanResult.recommendedAction === RecommendedAction.None ||
+        scanResult.recommendedAction === RecommendedAction.Warn,
+      name: 'Product safety dapp scanning is enabled',
+      type: 'DAPP_SCANNING' as PhishingDetectorResultType,
```
cryptotavares, line 53: Can we align these with the current meaning in the PhishinController?
For example, current there's only one phishing (list) name: MetaMask. And the current type depends on the list where we find a match (valid values are here - https://github.com/MetaMask/core/blob/main/packages/phishing-controller/src/types.ts#L42-L73).

We can hardcode aligned values here, but then we should open up a PR to get this details directly from the phishingController.

For the actual values I would suggest something like this:
```
import { ListNames } from '@metamask/phishing-controller';

...
return {
    ...
    name: ListNames.MetaMask,
    type: PhishingDetectorResultType.Blocklist
};
...
```

app/util/phishingDetection.ts
```diff
@@ -7,20 +12,52 @@ import { selectProductSafetyDappScanningEnabled } from '../selectors/featureFlag
...
+    };
+  }
+  PhishingController.maybeUpdateState();
+  const result = PhishingController.test(origin);
+  // PhishingDetectorResult.result docs says it is true if the domain is allowed.
+  // This is the opposite of what happens and instead it returns false if it is allowed.
+  // Hence, we negate the result to assume consistency with dev impression of the doc.
+  result.result = !result?.result;
```
cryptotavares, line 61: Why are we reversing this? Wouldn't it be easier to just align the result in the real-time dapp scanning branch? Something like this:
```js
...
    return {
      result:
        scanResult.recommendedAction !== RecommendedAction.None &&
        scanResult.recommendedAction !== RecommendedAction.Warn,
...
```
This would keep the response from `scanUrl` and `test` methods aling. And thus requiring less changes to start using the real-time dapp scanning.

imblue-dabadee:
```ts
/**
 * Represents the result of checking a domain.
 */
export type PhishingDetectorResult = {
    // ... other variables
    
    /**
     * Whether the domain is regarded as allowed (true) or not (false).
     */
    result: boolean;
    
    // ... other variables
};
```

It's to align better with what the docs say. It feels very off to say that if `scanResult.recommendedAction` is not `None` and `Warn`, that result is true i.e. the domain is regarded as allowed.

It won't be impacting the other components until they also use `getPhishingTestResultAsync`. Currently, they use `getPhishingTestResult` which means their logic isn't impacted.

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -314,6 +316,19 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
...
    */
   const handleNotAllowedUrl = useCallback((urlOrigin: string) => {
+    const resolvedUrlOrigin = new URLParse(resolvedUrlRef.current).origin;
+    const loadingUrlOrigin = new URLParse(loadingUrlRef.current).origin;
+    // If the resolved URL and the loading URL are different from the phishing URL
+    // and dapp scanning is enabled, don't show the phishing modal
+    if (
+      resolvedUrlOrigin !== urlOrigin &&
+      loadingUrlOrigin !== urlOrigin &&
+      loadingUrlOrigin != null &&
```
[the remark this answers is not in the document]
imblue-dabadee: Since it's async now, there's a delay from when the user navigates to a page and the dapp scanning API returns a response. These checks prevent the phishing modal appearing when:
- The user has completely navigated and loaded web content for another website
- The user has started navigating and loading web content for another website
This is important because if we've navigated to something like `metamask.io` after being on a phishing website and the response from the dapp scanning API returns after we've already navigated away on the phishing website, then the user thinks the modal is for `metamask.io`.

Good pick up on the `loadingUrlOrigin`, didn't realise it was returned as a string.

app/util/phishingDetection.ts
imblue-dabadee: Yeah, we should create a card for this to complete in future - @mindofmar

app/components/Views/BrowserTab/BrowserTab.tsx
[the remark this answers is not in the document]
cryptotavares: Ok that makes sense. I think that you could update the code comment o reflect what you just wrote above.. or maybe even move this code into its own validation method? Maybe something like:
```js
  /**
   * Check if user is still on the same url origin and has not started to navigate to a different one.
   */
  const isUserOnSameOrigin = (urlOrigin) => {
    const resolvedUrlOrigin = new URLParse(resolvedUrlRef.current).origin;
    const loadingUrlOrigin = new URLParse(loadingUrlRef.current).origin;

    return resolvedUrlOrigin !== urlOrigin && loadingUrlOrigin !== urlOrigin && loadingUrlOrigin !== 'null';
  }

...

  /**
   * Show a phishing modal when a url is not allowed
   */
  const handleNotAllowedUrl = useCallback((urlOrigin: string) => {
    // don't show the phishing modal if user has started to navigate to a different url origin
    if (isUserOnSameOrigin(urlOrigin) && isProductSafetyDappScanningEnabled()) {
      return;
    }
    setBlockedUrl(urlOrigin);
    setTimeout(() => setShowPhishingModal(true), 1000);
  }, []);
```

app/util/phishingDetection.ts
cryptotavares: I don't think that that type description is explicit.. as the PhishingDetector result is in fact just checking if the domain is part of any of the lists (blocklist, whitelist, etc etc). And if the description is indeed wrong (i.e. not matches the code implementation that is used across our codebase), then we update the description.
It is way easier to update the that code comment, rather than updating the implementation everywhere (to keep it consistent).
Otherwise you are creating another layer to invert the result.. but you are only adding that to a specific case in Mobile. And we will end up with different meaning for results across our codebase:
* a version for mobile (if you read through the helper that we have added)
* a version for when you read from the controller directly
* a version for what is implemented in the Extension (might match the one from the controller, for now)

imblue-dabadee: I get the concern so I'll have `scanUrl` conform to how EPD returns `PhishingDetectorResult`. I don't think the concern is that important since in future I imagine we would deprecate this type as I can't see why we would need most of the variables in the type since `scanUrl` doesn't return this values.

AugmentedMode (dismissed):
lgtm

smilingkylan (dismissed):
LGTM ✅

mindofmar (approved):

smilingkylan (approved):

AugmentedMode (approved):

PR #14659, opened 2025-04-14
fix: Add NFT detection event

gambinish:
## **Description**

Adds and distinguishes between custom NFT import and auto-detect import metrics in the mobile app.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MMASSETS-659

## **Manual testing steps**

1. Install fresh app, and import a new account with known NFTs
2. Ensure that events from auto-detect flows are captured with a source of "detected" while custom import flow is "custom"

## **Screenshots/Recordings**

N/A


app/components/UI/CollectibleContracts/index.js
```diff
@@ -370,18 +372,64 @@ const CollectibleContracts = ({
...
+      return {
+        chain_id: getDecimalChainId(nft.chainId),
+        source: 'detected',
+      };
+    } catch (error) {
+      Logger.error(
+        error,
+        'CollectibleContracts.getNftDetectionAnalyticsParams',
+      );
+      return undefined;
```
Prithpal-Sooriya, line 387: Aha nice, so we inject this so we can then log different error messages on failure?

app/components/UI/CollectibleContracts/index.js
```diff
@@ -370,18 +372,64 @@ const CollectibleContracts = ({
...
+        previousNfts,
+        newNfts,
+        getNftDetectionAnalyticsParams,
+      );
+      eventParams.forEach((params) => {
+        trackEvent(
+          createEventBuilder(MetaMetricsEvents.COLLECTIBLE_ADDED)
+            .addProperties(params)
+            .build(),
+        );
```
Prithpal-Sooriya, line 423: Pretty sure that `trackEvent` under the hood will batch all these requests.
I'm not being paranoid right? We won't be spamming the user on many detections?

Prithpal-Sooriya: (out of scope) wondering if we could send a single event per chain with a `count` property to indicate how many NFTs were detected on this chain?

^ But I don't think this will play well with MXP analytics.

Prithpal-Sooriya (approved):

app/components/UI/CollectibleContracts/index.js
gambinish: I was thinking about this too. I know that we are doing something similar to this with the `TOKEN_ADDED` event, so I figured that this would probably be okay here as well. Fwiw, we're only sending these event when the NFT list changes.

So, in the case where we have an account with a ton of NFTs, they would only get sent on the first import after a fresh install.

gambinish: But yeah, if we start over-sending events, maybe we can revisit this.

salimtb (approved):

PR #14539, opened 2025-04-08
feat(4144): per-dapp-selected-networks

EtherWizard33:
## **Description**

**Issue**: All dapps are connected on the same chain, if one switches, others as well. Most likely the user would prefer to have his selected networks to be specific to each dapp (aka per-dapp).

**Solution**: Add per-dapp-selected-network, Add a touchable favicon to the permission summary header that enables network switching for dapps. 

# Per-DApp Selected Network Implementation

### Features
- Implemented per-dapp network selection functionality with interactive favicon
- Added network badge and switching capability in permission summary header
- Enhanced chain fallback logic for per-dapp-selected-network feature
- Added origin-specific network info for transaction reviews
- Removed non-permitted network flow checks for dapps (TODO: keep behind feature flag, enabled when !perDapp)

### Fixes
- Updated transaction notifications to use chain-specific ticker symbols
- Fixed network client ID handling for smart transactions
- Improved chain comparison logic for per-dapp transactions
- Corrected domain logo container placement behind feature flag
- Fixed favicon display logic in permission summary for unconnected dapps

### Refactoring
- Centralized per-DApp network logic using `isPerDappSelectedNetworkEnabled` helper
- Cleaned up network selection implementation
- Removed temporary network picker components
- Updated network name and image source to be origin-aware
- Standardized feature flag usage across components

### Code Quality
- Removed debug logging statements
- Code cleanup and linting improvements
- Updated test snapshots for network name changes

## **Related issues**

Contributes to: https://github.com/MetaMask/MetaMask-planning/issues/4104

## **Manual testing steps**

Testing:
1. Enable feature flags below if they are not enabled already
2. Connect to a dapp and verify network picker appears in permissions view
3. Switch networks and verify correct network displays for that specific dapp
4. Verify network switching works properly and modal dismisses correctly
5. Verify network info displays correctly in permissions view

Make sure to include all env vars:
```
export MM_MULTICHAIN_V1_ENABLED="true"
export MM_CHAIN_PERMISSIONS="true"
export MM_PER_DAPP_SELECTED_NETWORK="true"
export MULTICHAIN_V1="true"
```

## **Screenshots/Recordings**

| Before       | After        |
|--------------|--------------|
| Screenshot 2024-04-18 at 3 56 43 PM |Screenshot 2024-04-18 at 3 56 43 PM |

## **Pre-merge author checklist**

[...]

PR #14756, opened 2025-04-21
feat: Add Multichain API to Flask

ffmcgee725:
## **Description**

This branch adds support for the [Multichain API](https://github.com/MetaMask/metamask-improvement-proposals/blob/main/MIPs/mip-5.md) to the Flask build of the Extension.

The existing API (via injected provider) should be completely unchanged.

# (Very Briefly) What is the MetaMask Multichain API
- Concurrent connection to any number of chains (no network switching)
- Unified entry point for all chain ecosystems (EVM, BTC, Solana, Cosmos, Polkadot etc)
- Accessible via [`window.postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage). Not accessible via an injected global like `window.ethereum`

### Key Documents/Standards
_mip = MetaMask Improvement Proposal_
- [MIP-5](https://github.com/MetaMask/metamask-improvement-proposals/blob/main/MIPs/mip-5.md) (Overview of the Multichain API)
  - [CAIP-25](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-25.md) (new connection request API)
  - [CAIP-27](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-27.md) (new request API, envelope with target scope/chainId included)
- [MIP-6](https://github.com/MetaMask/metamask-improvement-proposals/blob/main/MIPs/mip-6.md) (Overview of how the Multichain API’s EVM support diverges from the 1193 injected provider)

## **Manual testing steps**
**(RECOMMENDED)** Use the [Multichain Test Dapp](https://metamask.github.io/test-dapp-multichain/latest/)
NOTE: `window.postMessage` should be used as extensionId for mobile.
- [Video Demo for Interacting w/ Multichain Test Dapp](https://drive.google.com/file/d/113iM80dZItEAGkobJkBh17h68Y5dgTS-/view)

**OR**
```
yarn setup:flask
```
Then

```
yarn start:android:flask
```

or

```
yarn start:ios:flask
```

Setup event listener via dev tools

```
window.addEventListener('message', (event) => {
  const {
    target,
    data
  } = event.data;
  if (
    target !== "metamask-inpage" ||
    data?.name !== 'metamask-multichain-provider'
  ) {
    return;
  }
  console.log(data.data)
})
```

Send a multichain API request via dev tools
```
const caipPostMessage = (data) => {
  window.postMessage({
    target: 'metamask-contentscript',
    data: {
      name: 'metamask-multichain-provider',
      data
    }
  }, location.origin)
}

// create session
caipPostMessage({
    jsonrpc: '2.0',
    method: 'wallet_createSession',
    params: {
      optionalScopes: {}
    }
})

// get session
caipPostMessage({
    jsonrpc: '2.0',
    method: 'wallet_getSession',
    params: {}
})

// revoke session
caipPostMessage({
    jsonrpc: '2.0',
    method: 'wallet_revokeSession',
    params: {}
})

// invoke method
caipPostMessage({
    jsonrpc: '2.0',
    method: 'wallet_invokeMethod',
    params: {
            scope: 'eip155:1',
            request: {
                "method": "eth_blockNumber",
                "params": [],
            }
        }
})
```

[More detailed example of manual requests](https://docs.google.com/document/d/1uq7N8Pt5ZrBQiPa0QFoGYGCnvDeTwhNK3ilOqUGX9JQ/edit?usp=sharing)

taken from [extension equivalent Add Multichain API to Flask work](https://github.com/MetaMask/metamask-extension/pull/27782)

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/CODING_GUIDELINES.md).
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-extension/blob/develop/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.


app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +353,369 @@ export class BackgroundBridge extends EventEmitter {
...
+              //   currentSelectedSolanaAccountAddress
+              //     ? [currentSelectedSolanaAccountAddress]
+              //     : [],
+              // );
+            }
+          });
+        },
+        getAuthorizedScopesByOrigin,
+      );
+
```
adonesky1, line 603:
```suggestion
```

This should be added in a separate PR later on, per our sequencing/breakdown: https://docs.google.com/document/d/1uUK2L6KBSHnz5yzyW54MVYhsSXcaZajNKVIUczljL8g/edit?tab=t.0

ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/ba1896c71e7dd890ea94c432673bed4f7d2c42e3

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +351,222 @@ export class BackgroundBridge extends EventEmitter {
...
+   * changes.
+   *
+   * Some of the subscriptions in this method are Messenger selector
+   * event subscriptions. See the relevant documentation for
+   * `@metamask/base-controller` for more information.
+   *
+   * Note that account-related notifications emitted when the extension
+   * becomes unlocked are handled in MetaMaskController._onUnlock.
+   */
+  setupControllerEventSubscriptions() {
```
adonesky1, line 366: Have we verified whether some of these subscriptions are being handled elsewhere?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +351,222 @@ export class BackgroundBridge extends EventEmitter {
...
+
+    controllerMessenger.subscribe(
+      'NetworkController:networkDidChange',
+      async () => {
+        if (PreferencesController.state.useExternalServices) {
+          TransactionController.stopIncomingTransactionPolling();
+          await TransactionController.updateIncomingTransactions();
+          TransactionController.startIncomingTransactionPolling();
+        }
+      },
```
adonesky1, line 566: I imagine this must exist somewhere else already...

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: The `smartTransactionPoller` logic seems to be handled in the `Engine` https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Engine/Engine.ts#L1507C10-L1507C22, but the starting condition doesn't seem to be the same.

and `_checkTokenListPolling` doesn't seem to be handled elsewhere.

so all in all, if we are talking about controllerMessenger subscription to `PreferencesController:stateChange`, none existed prior to this implementation.

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: searching up `NetworkController:networkDidChange`, it doesn't seem to be the case.

I only see stuff such as [this](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/Engine/Engine.ts#L333), and not an actual subscription by the `controllerMessenger` to this action.

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -396,10 +686,30 @@ export class BackgroundBridge extends EventEmitter {
...
+
+    // setup connection
+    const providerStream = createEngineStream({
+      engine: this.multichainEngine,
+    });
+
+    pump(outStream, providerStream, outStream, (err) => {
+      // handle any middleware cleanup
+      this.multichainEngine.destroy();
+      if (err) Logger.log('Error with provider stream conn', err);
```
darkwing, line 705: Do we know what this error could look like?

ffmcgee725: I personally don't have enough domain knowledge atm to know this and give you a proper answer. I just followed the same pattern from [the EIP1193 equivalent provider engine setup](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/BackgroundBridge/BackgroundBridge.js#L395)

perhaps @jiexi or @adonesky1 can chime in here ?

jiexi: likely something like `StreamXYZ has closed.`

app/core/RPCMethods/wallet_addEthereumChain.js
```diff
@@ -286,19 +286,3 @@ export const wallet_addEthereumChain = async ({
...
-    addNetwork: true,
-    updateNetwork: true,
-    getNetworkConfigurationByChainId: true,
-    setActiveNetwork: true,
-    requestUserApproval: true,
-    getCurrentChainIdForDomain: true,
-    getCaveat: true,
-    requestPermittedChainsPermissionIncrementalForOrigin: true,
-    rejectApprovalRequestsForOrigin: true,
-  },
```
abretonc7s, line 303: I am assuming we dont need this anymore?

app/core/RPCMethods/wallet_addEthereumChain.js
```diff
@@ -286,19 +286,3 @@ export const wallet_addEthereumChain = async ({
...
-  methodNames: [MESSAGE_TYPE.ADD_ETHEREUM_CHAIN],
-  implementation: wallet_addEthereumChain,
-  hookNames: {
-    addNetwork: true,
-    updateNetwork: true,
-    getNetworkConfigurationByChainId: true,
-    setActiveNetwork: true,
-    requestUserApproval: true,
-    getCurrentChainIdForDomain: true,
-    getCaveat: true,
```
abretonc7s, line 300: do we need to keep this on 1193 path?

abretonc7s (dismissed):

app/core/RPCMethods/wallet_addEthereumChain.js
ffmcgee725: No handler needed for either side here because `wallet_addEthereumChain` handler is added to the pipeline via [this](https://github.com/MetaMask/metamask-mobile/blob/main/app/core/BackgroundBridge/BackgroundBridge.js#L543).

abretonc7s (dismissed):

.js.env.example
```diff
@@ -58,6 +58,9 @@ export METAMASK_ENVIRONMENT="local"
 # Build type: "main" or "flask" or "beta"
 export METAMASK_BUILD_TYPE="main"
 
+# Activates MultiChainAPI pipeline in the in-app browser.
+export MULTICHAIN_API="true"
```
tommasini, line 62: Should we add to the bitrise.yrml file as well this env variable?

tommasini:
It would be nice to add some more coverage to the background bridge file, any reason why we are not doing it?

tommasini:
Can we also update with main and re run e2e?

.js.env.example
ffmcgee725: We still have no end to end tests for Multichain atm, I believe @abretonc7s will be working on this. So it might make sense to add it by that time.

chrisleewilcox:
Scenario: Send Legacy Eth from [Test Dapp](https://metamask.github.io/test-dapp/) using in-app browser
https://www.loom.com/share/f2673afc24b743259a64e6b21f556c2d?sid=9e231244-173e-4b9d-9cb3-055cc4dbcebf

chrisleewilcox:
Not able to send EIP 1559, has silent failure.

Scenario: Send EIP 1559 from Test Dapp using in-app browser
https://www.loom.com/share/6a41ffda7c244e7b83517a2644dae88f?sid=e0577508-138e-42f2-9873-6958bdd1fea3

Not an issue with current main testing on browserstack:
https://www.loom.com/share/74f2fb034b324b07a133bb84cbb45066?sid=760c5e05-33bc-425c-914a-95c8d9126e49

chrisleewilcox:
App crash on malformed signatures INVALID TYPE and INVALID PRIMARY TYPE.

Scenario: Malformed signatures on Test Dapp INVALID TYPE and INVALID PRIMARY TYPE
https://www.loom.com/share/790adcac25f24432afd6a5bdc75da12e?sid=18a1ddc2-3526-491d-b0d9-b47537fcca94

App is not crashing on Pixel device, different type of crash on Samsung testing on browserstack:
https://www.loom.com/share/6a095577222d44279e111c30ca066bc8?sid=ee2e083d-9410-4769-b7a5-1d6740c0e570

ffmcgee725:
> App crash on malformed signatures INVALID TYPE and INVALID PRIMARY TYPE.

fixed by https://github.com/MetaMask/metamask-mobile/pull/14756/commits/c3646db081beeb691b96230da35b492f36b966b0

abretonc7s:
> > App crash on malformed signatures INVALID TYPE and INVALID PRIMARY TYPE.
> 
> fixed by [c3646db](https://github.com/MetaMask/metamask-mobile/commit/c3646db081beeb691b96230da35b492f36b966b0)

A type-checking improvement inadvertently removed early failure on bad inputs. We added explicit validation that now properly rejects bad signatures before they can crash the UI. Tests needed proper EIP-712 data to pass validation.

tommasini (dismissed):

app/components/Views/BrowserTab/BrowserTab.tsx
```diff
@@ -901,6 +901,7 @@ export const BrowserTab: React.FC<BrowserTabProps> = ({
 
       //@ts-expect-error - We should type bacgkround bridge js file
       const newBridge = new BackgroundBridge({
+        tabId,
```
jiexi, line 904: I'm not sure tabId is necessary because each Browser tab gets its own BackgroundBridge instance? I.e. each BackgroundBridge serves exactly one tab, so there are no other tabId to differentiate against?

app/components/Views/confirmations/legacy/components/TypedSign/index.js
jiexi, line 1: why are these changes necessary?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -21,11 +21,24 @@ import WalletConnectPort from './WalletConnectPort';
...
 import { store } from '../../store';
 ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps)
-import { rpcErrors } from '@metamask/rpc-errors';
+import { providerErrors, rpcErrors } from '@metamask/rpc-errors';
 import snapMethodMiddlewareBuilder from '../Snaps/SnapsMethodMiddleware';
-import { SubjectType } from '@metamask/permission-controller';
+import {
+  PermissionDoesNotExistError,
+  SubjectType,
+} from '@metamask/permission-controller';
```
jiexi, line 29: does this need to be code fenced by `(preinstalled-snaps,external-snaps)`?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +347,223 @@ export class BackgroundBridge extends EventEmitter {
...
+
+    controllerMessenger.subscribe(
+      `${AccountsController.name}:selectedAccountChange`,
+      async (account) => {
+        if (account.address && account.address !== lastSelectedAddress) {
+          lastSelectedAddress = account.address;
+          await this._onAccountChange(account.address);
+        }
+      },
+    );
```
jiexi, line 403: do we need to verify these events aren't fired elsewhere? Mobile hasn't been making good use of event listeners to fire off side effects and instead inline side effects next to the main code triggering them

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -555,6 +867,176 @@ export class BackgroundBridge extends EventEmitter {
...
+          MESSAGE_TYPE.WALLET_GET_SESSION,
+          MESSAGE_TYPE.WALLET_REVOKE_SESSION,
+        ].includes(req.method)
+      ) {
+        return end(rpcErrors.methodNotFound({ data: { method: req.method } }));
+      }
+      return next();
+    });
+
+    engine.push(multichainMethodCallValidatorMiddleware);
```
jiexi, line 908: missing the rpc tracking middleware around here

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -555,6 +867,176 @@ export class BackgroundBridge extends EventEmitter {
...
+          ...UNSUPPORTED_RPC_METHODS,
+          'eth_requestAccounts',
+          'eth_accounts',
+        ]),
+      ),
+    );
+
+    engine.push(
+      createMultichainMethodMiddleware({
+        // getProviderState handler related
```
jiexi, line 970: ?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -58,6 +109,7 @@ const legacyNetworkId = () => {
 
 export class BackgroundBridge extends EventEmitter {
   constructor({
+    tabId,
```
jiexi, line 112: correct me if i'm wrong. BackgroundBridge can only serve one tabId and origin, so we don't need to do all this looping over all the origins/tabId. For multichainSubscriptionManager and multichainMiddlewareManager, i think the `tabId` value is optional and can be omitted

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: What's the question here ? I don't understand

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: @abretonc7s are you able to provide context here ?

app/components/Views/confirmations/legacy/components/TypedSign/index.js
ffmcgee725: @abretonc7s can provide more context here, I believe it has to do with the bug reported above related to `signTypedData` calls, but not sure.

app/core/RPCMethods/RPCMethodMiddleware.ts
jiexi, line 1: i feel like these changes belong in a separate PR 😬

app/core/RPCMethods/createMultichainMethodMiddleware/index.js
jiexi, line 1: does this need a test?

app/selectors/permissions/index.js
jiexi, line 1: the selectors in here may not need to iterate over every origin

app/util/validators/index.test.ts
jiexi, line 1: feels like this belongs in a separate PR

app/core/BackgroundBridge/BackgroundBridge.js
jiexi: ohh... you're saying this is `metamask_getProviderState` related. Right, okay 👍

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/files#r2069998692

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +347,223 @@ export class BackgroundBridge extends EventEmitter {
...
+          const { chainId: currentChainIdForOrigin } =
+            NetworkController.getNetworkConfigurationByNetworkClientId(
+              currentNetworkClientIdForOrigin,
+            );
+          // if(chains.length === 0) {
+          // TODO: [migrated from extension?] This particular case should also occur at the same time
+          // that eth_accounts is revoked. When eth_accounts is revoked,
+          // the networkClientId for that origin should be reset to track
+          // the globally selected network.
+          // }
```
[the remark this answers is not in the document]
ffmcgee725: how should we handle then ?
Do we reset networkClientId to the globally selected network, or do we just leave it be ?

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/20578f7611f670c3e2d331ebc6b02369d9d5788d

app/core/RPCMethods/createMultichainMethodMiddleware/index.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/20578f7611f670c3e2d331ebc6b02369d9d5788d

app/components/Views/BrowserTab/BrowserTab.tsx
ffmcgee725: @abretonc7s

app/components/Views/confirmations/legacy/components/TypedSign/index.js
abretonc7s: The issue is that the UI component and RPC middleware run in separate contexts - the UI can mount with bad props before validation happens. 

I found that fixing just the middleware wasn't enough in testing - the app still crashed in some scenarios. The empty Views are just a safety net that cost almost nothing but prevent app crashes.

We can revisit the UI safeguards later if we're sure all paths are covered by the middleware.

abretonc7s (dismissed):

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/d9ec93447bd8b978f38f8c5fe4783b35ab327dec

app/components/Views/BrowserTab/BrowserTab.tsx
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/d9ec93447bd8b978f38f8c5fe4783b35ab327dec

app/selectors/permissions/index.js
ffmcgee725: For the sake of keeping shared code, are you open to keeping this as it is so that we can later port these functions over to monorepo as per [this ticket](https://github.com/MetaMask/MetaMask-planning/issues/4899) ?
That way functionality can be shared across extension and mobile, regardless of us just iterating through one origin in `BackgroundBridge.js`

abretonc7s:
# Typed Signature Crash Fix

## The Bug

We had an app crash with the error: `TypeError: message.map is not a function (it is undefined)` when handling certain typed signature requests. Classic undefined error when trying to map over nothing.

## What Happened

Check out this subtle change that broke things:

```diff
- const data = JSON.parse(req.params[1]);
+ const data = typeof req.params[1] === 'string'
+   ? JSON.parse(req.params[1])
+   : req.params[1];
```

The old code would throw immediately if params[1] was undefined. The new code silently passes undefined through the pipeline until it hits the UI, where it crashes trying to call `.map()` on undefined.

## Root Cause of the Crash

The crash happened in `sanitizeParsedMessage` function in `signature.ts`. When we received malformed signature data, the function would try to call `.map()` on a message that wasn't actually an array:

```javascript
// This would crash when message wasn't an array but primaryType indicated it should be
return {
  value: (message as string[]).map(
    (value: string): ValueType =>
      sanitizeParsedMessage(value, stripOneLayerofNesting(primaryType), types),
  ),
  type: primaryType,
};
```

Our fix adds validation to catch this type mismatch before attempting to use map.

## Our Fix

We added validation in the middleware to catch this early:

```javascript
if (!req.params?.[1]) {
  throw rpcErrors.invalidParams('Invalid parameters: missing message data');
}

// Parse if needed
data = typeof req.params[1] === 'string'
  ? JSON.parse(req.params[1])
  : req.params[1];

// Validate required fields
if (!data?.domain || !data?.types || !data?.message) {
  throw rpcErrors.invalidParams('Invalid parameters: missing required fields');
}

// Pre-validate with the same function the UI would use
parseAndSanitizeSignTypedData(/* ... */);
```

Now invalid requests fail fast with proper error messages instead of crashing the UI.

## Tests Broke Because...

Our test mock was too minimal:

```javascript
const dataJsonMock = JSON.stringify({
  test: 'data',
  domain: { chainId: '0x1' },
});
```

It was missing the required `types` and `message` fields that our validation now checks for. Had to update to a fully compliant EIP-712 object.

## E2E Test Failing

There's a problem with the `error-boundary-srp-backup.spec.js` test now. This test was designed to verify users can access their Secret Recovery Phrase through the error boundary screen if the app crashes.

### The issue:

1. This test was intentionally causing the app to crash using invalid signature data
2. Our new validation prevents this crash by properly handling invalid data
3. Now the test can't reach the error boundary screen because the app doesn't crash

### Timeline of events:

1. We had a bug where bad signature data would crash the app
2. We fixed the validation to properly handle bad data (good!)
3. The e2e test that relied on crashing now fails (bad, but expected)

## Future Considerations: How Should We Handle Malformed Signatures?

Our current approach prevents crashes by sanitizing malformed signature data, but we should consider if this is the best approach. There are several alternatives we could explore in a future PR:

### Option 1: Strict Rejection
- Return clear RPC errors to dApps when signature data is malformed
- Most secure approach as it prevents users from signing potentially misleading data
- Could break compatibility with some dApps that send slightly malformed data

### Option 2: Validate and Warn
- Allow signing of malformed data but show clear warning screens to users
- Balances security with compatibility
- Gives users more information to make informed decisions

### Option 3: Current Sanitizing Approach
- Continue sanitizing malformed data to prevent crashes
- Most compatible with existing dApps
- Risk that users sign data they don't fully understand

We should test these approaches using the Multichain Test Dapp to:
- Compare the user experience with different types of malformed data
- Evaluate how clear the resulting data appears to users
- Determine whether users can make informed decisions about what they're signing

Given that signatures authorize blockchain actions, we should carefully consider the security implications of how we handle malformed data.

## Recommendation

Since this PR is focused on the Multichain API work, I recommend:

1. Temporarily disable the failing e2e test by adding `.skip` to the test:
   ```js
   it.skip('should trigger error boundary screen to reveal SRP', async () => {
   ```

2. Create a separate follow-up PR that:
   - Re-enables the test
   - Implements a proper way to trigger the error boundary for testing (e.g., a dedicated test method that throws errors)
   - Updates the documentation

3. The minimal improvements to `signature.ts` are included here to prevent crashes, but more comprehensive error handling should be in a separate PR. These additional changes would:
   - Add better validation and user-friendly error messages
   - Improve the robustness of signature handling code
   - Be properly reviewed on their own since they touch critical security functionality

This approach keeps our current PR focused while ensuring we have a plan to properly fix the test in a dedicated PR.

abretonc7s (dismissed):

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +345,223 @@ export class BackgroundBridge extends EventEmitter {
...
+
+            // if the eth_subscription notification is in the scope and eth_subscribe is in the methods
+            // then get the subscriptionManager going for that scope
+            Object.entries(sessionScopes).forEach(([scope, scopeObject]) => {
+              if (
+                scopeObject.notifications.includes('eth_subscription') &&
+                scopeObject.methods.includes('eth_subscribe')
+              ) {
+                // for each tabId
+                Object.values(this.connections[origin]).forEach(({ tabId }) => {
```
jiexi, line 476: this.connections doesn't exist

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +345,223 @@ export class BackgroundBridge extends EventEmitter {
...
+                  });
+                });
+              } else {
+                this.removeMultichainApiEthSubscriptionMiddleware({
+                  scope,
+                  origin,
+                });
+              }
+            });
+            this._notifyAuthorizationChange(origin, authorization);
```
jiexi, line 490: i don't believe this exists?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +347,223 @@ export class BackgroundBridge extends EventEmitter {
...
+                scopeObject.notifications.includes('eth_subscription') &&
+                scopeObject.methods.includes('eth_subscribe')
+              ) {
+                this.removeMultichainApiEthSubscriptionMiddleware({
+                  scope,
+                  origin,
+                });
+              }
+            });
+          }
```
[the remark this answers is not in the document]
jiexi: it's one auth per origin. each BrowserTab only needs to be aware of one origin

app/core/BackgroundBridge/BackgroundBridge.js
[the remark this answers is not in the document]
jiexi: i think it needs to point back to the currently selected network if no networks are currently permitted

app/selectors/permissions/index.js
jiexi: i'm worried about performance. @adonesky1

adonesky1: agreed we need to validate that the backgroundBridge instantiations are 1:1 for tabs and then adapt accordingly

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: fixed: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/bfe98a0eba9f90abd3bfe1252c3ba39e523e23fe

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: fixed: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/95fce88de36b1a5f34a0ad9d0bc054e7f3479430

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -35,14 +48,52 @@ const pump = require('pump');
...
+import { previousValueComparator } from '../../util/validators';
+import { hexToBigInt, toCaipChainId } from '@metamask/utils';
+import { TransactionController } from '@metamask/transaction-controller';
+import { createMultichainMethodMiddleware } from '../RPCMethods/createMultichainMethodMiddleware';
+
+// Types of APIs
+const API_TYPE = {
+  EIP1193: 'eip-1193',
+  CAIP_MULTICHAIN: 'caip-multichain',
+};
```
adonesky1, line 96: [nit] - I know this was copied from extension, but ideally we'd provide more context about what these two mean

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: I don't believe we should add any of these subscriptions that aren't explicitly tied to our Multichain API implementation

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +345,222 @@ export class BackgroundBridge extends EventEmitter {
...
+      PermissionController,
+    } = Engine.context;
+
+    controllerMessenger.subscribe(
+      'PreferencesController:stateChange',
+      previousValueComparator(async (prevState, currState) => {
+        this._restartSmartTransactionPoller();
+        this._checkTokenListPolling(currState, prevState);
+      }, PreferencesController.state),
+    );
```
adonesky1, line 391: we should remove this

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: We shouldn't be adding subscriptions/side-effects for the legacy API in this PR.

It would be very strange to me if these aren't somehow implemented elsewhere that we're not seeing. But in any case the job of this PR is to implement the Multichain API

jiexi: yeah... probably should remove this listener

app/core/BackgroundBridge/BackgroundBridge.js
jiexi: yeah makes sense

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -272,6 +351,222 @@ export class BackgroundBridge extends EventEmitter {
...
+            // setActiveNetwork should be called before setNetworkClientIdForDomain
+            // to ensure that the isConnected value can be accurately inferred from
+            // NetworkController.state.networksMetadata in return value of
+            // `metamask_getProviderState` requests and `metamask_chainChanged` events.
+            NetworkController.setActiveNetwork(networkClientId);
+            SelectedNetworkController.setNetworkClientIdForDomain(
+              origin,
+              networkClientId,
+            );
+          }
```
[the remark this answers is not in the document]
adonesky1: Have you verified whether `setNetworkClientIdForDomain` is actually executing here?

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: Another case where we shouldn't be adding this. Really we should be asking why it is that this wasn't here before. But regardless of the answer we shouldn't be adding it here as part of this PR

adonesky1: Here for example is one other subscription/side-effect listening to this event: https://github.com/MetaMask/metamask-mobile/blob/7819a6bdb362c41489c873ff9a774a1f0f3d13f1/app/core/Engine/Engine.ts#L1438

adonesky1: and here is where starting `incomingTransactionPolling` is handled in mobile: https://github.com/MetaMask/metamask-mobile/blob/7819a6bdb362c41489c873ff9a774a1f0f3d13f1/app/core/Engine/Engine.ts#L1526

Its still not clear why we don't need to stop/restart on this event in mobile, but again not in scope for this PR

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: Asking about this in Mobile Dev's channel: https://consensys.slack.com/archives/C02U025CVU4/p1747423771174649

app/util/validators/index.ts
```diff
@@ -2,6 +2,37 @@ import { ethers } from 'ethers';
...
+ * value and the current value. If specified, the initialValue will be passed
+ * in as the previous value on the first invocation of the returned method.
+ *
+ * @template A - The type of the compared value.
+ * @param comparator - A method to compare
+ * the previous and next values.
+ * @param [initialValue] - The initial value to supply to prevValue
+ * on first call of the method.
+ */
+export function previousValueComparator<A>(
```
adonesky1, line 17: we can get rid of this and all the associated tests, because its used for a subscription we shouldn't be adding in this PR

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: Agreed. I still need to re-evaluate how we will use these though, since `BackgroundBridge` does not hold a list of connections which know the `API TYPE`. Only one connection per `BackgroundBridge`.

app/core/BackgroundBridge/BackgroundBridge.js
adonesky1: Yeah so the pattern is frustratingly different but we should basically assume all of these subscriptions that appear to be missing are handled differently in mobile: https://consensys.slack.com/archives/C02U025CVU4/p1747424652300869?thread_ts=1747423771.174649&cid=C02U025CVU4

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -579,6 +1056,237 @@ export class BackgroundBridge extends EventEmitter {
...
+
+  /**
+   * Sorts a list of addresses by most recently selected by using the lastSelected value for
+   * the matching InternalAccount object from the list of internalAccounts provided.
+   *
+   * @param {string[]} [addresses] - The list of caip accounts addresses to sort.
+   * @param {InternalAccount[]} [internalAccounts] - The list of InternalAccounts to determine lastSelected from.
+   * @returns {string[]} The sorted accounts addresses.
+   */
+  sortAddressesWithInternalAccounts(addresses, internalAccounts) {
```
adonesky1, line 1081: these don't appear to be used anywhere?

adonesky1:
Finally have this up and running locally 🙌 

Seeing that some state selectors seem to be stuck on globally selected network state:
Screenshot 2025-05-16 at 3 35 45 PM


I wonder if the per dapp selected network PR (currently in review) will resolve this: https://github.com/MetaMask/metamask-mobile/pull/14539 🙏

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: Think these were used back in the Solana accounts changed event code we removed early in this branch's life. I could be wrong, but yeah not referenced anymore, removed here https://github.com/MetaMask/metamask-mobile/pull/14756/commits/0cbdc6d600225fb05a2c554047f2a05eab029302

app/util/validators/index.ts
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/0cbdc6d600225fb05a2c554047f2a05eab029302

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/0cbdc6d600225fb05a2c554047f2a05eab029302

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: seems like we agreed on assuming subscriptions are handled elsewhere in mobile, so after removing the event subscription method, this was not referenced anymore, so removed in altogether in https://github.com/MetaMask/metamask-mobile/pull/14756/commits/0cbdc6d600225fb05a2c554047f2a05eab029302

app/selectors/permissions/index.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/5a5b8086fa071e504db233d6eee18df1b3144480

app/core/BackgroundBridge/BackgroundBridge.js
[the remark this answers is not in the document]
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/5a5b8086fa071e504db233d6eee18df1b3144480

app/core/BackgroundBridge/BackgroundBridge.js
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/0cbdc6d600225fb05a2c554047f2a05eab029302

app/core/BackgroundBridge/BackgroundBridge.js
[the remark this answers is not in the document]
ffmcgee725: outdated after removing most of `setupControllerEventSubscriptions` : https://github.com/MetaMask/metamask-mobile/pull/14756/commits/0cbdc6d600225fb05a2c554047f2a05eab029302

abretonc7s (dismissed):

app/selectors/permissions/index.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Get the authorized CAIP-25 scopes for the subject.
+ * The returned value is an immutable value from the PermissionController state,
+ * or undefined if no authorization exists.
+ *
+ * @returns {Caip25CaveatValue | undefined} The current authorization or undefined if no authorization exists.
+ */
+export const getAuthorizedScopes = createSelector(getSubjects, (subjects) => {
+  const subject = Object.values(subjects)[0];
```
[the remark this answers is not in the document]
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/665de09b687652d0eac308906c726dd7427fcd02

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -557,6 +646,246 @@ export class BackgroundBridge extends EventEmitter {
...
+  /**
+   * This handles CAIP-25 authorization changes every time relevant permission state changes, for any reason.
+   */
+  setupCaipEventSubscriptions() {
+    const { controllerMessenger, context } = Engine;
+
+    // wallet_sessionChanged and eth_subscription setup/teardown
+    controllerMessenger.subscribe(
+      `${context.PermissionController.name}:stateChange`,
+      this.handleCaipSessionScopeChanges(),
```
jiexi, line 822:
```suggestion
      this.handleCaipSessionScopeChanges,
```

maybe even `.bind(this)`

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -557,6 +646,246 @@ export class BackgroundBridge extends EventEmitter {
...
+      getAuthorizedScopes(this.origin),
+    );
+  }
+
+  /**
+   * Handler for CAIP permission state changes.
+   * @returns function that handlers session scope changes.
+   */
+  handleCaipSessionScopeChanges() {
```
jiexi, line 831:
```suggestion
  handleCaipSessionScopeChanges(currentValue, previousValue) {
```

ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/52a953cb707644cf2b0778ffb7a0d27c0f6f0943

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -557,6 +646,247 @@ export class BackgroundBridge extends EventEmitter {
...
+          });
+        }
+      });
+    } catch (err) {
+      // noop
+    }
+
+    this.multichainSubscriptionManager.on(
+      'notification',
+      (targetOrigin,_, message) => {
```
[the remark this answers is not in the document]
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/14756/commits/7f63d6cfee7d2c0dc6df04e12a287f42f5748575

app/util/permissions/index.test.ts
```diff
@@ -602,4 +608,188 @@ describe('Permission Utils', () => {
       );
     });
   });
+
+  describe('getChangedAuthorization', () => {
```
jiexi, line 612: is it okay if we add a scenario testing two accounts and then one account removed?

jiexi (approved):
🚀 🚀 🚢 🎉

app/core/RPCMethods/createMultichainMethodMiddleware/index.js
```diff
@@ -0,0 +1,7 @@
+import { multichainLocalHandlers } from '../handlers';
+import { makeMethodMiddlewareMaker } from '../utils';
+
+// The primary home of RPC method implementations for the MultiChain API.
```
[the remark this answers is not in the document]
adonesky1: not a blocker ☝️

chrisleewilcox:
@adonesky1 @ffmcgee725 @jiexi 
Was wondering how I can run the test added with this PR.  I see page objects but I don't see a spec file for running the test and was wondering how it's done.  Is it possible to have a recording of the tests running on this PR that I can review?

adonesky1:
Tiny bug w/ `wallet_sessionChanged`:
https://github.com/user-attachments/assets/62a7a8c5-28db-4e66-916e-77bf396a151b

At the end I misspoke and said 2 accounts connected, but actually the connection state is correct with only one account, but in the `wallet_sessionChanged` event 2 accounts are shown

adonesky1:
@chrisleewilcox no e2e tests were added in this PR. @abretonc7s is starting to add them in a separate PR which should be ready soon!

app/core/RPCMethods/createMultichainMethodMiddleware/index.js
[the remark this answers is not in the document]
jiexi: perhaps this comment should mention EIP1193 methods that are served by the Multichain API

adonesky1:
Seems like we're always returning `0x1` for `eth_chainId`?
https://github.com/user-attachments/assets/e479145b-2af2-4c04-82d6-72d0caed61f9

chrisleewilcox:
Regression tests: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/99a51c7a-d1de-4680-aac9-dfd64f615b53

The failures are related to the Solana slides that have been fixed in another PR: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/a9b9cd77-f914-4c54-80c1-b7b4667340a7

adonesky1:
> Regression tests: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/99a51c7a-d1de-4680-aac9-dfd64f615b53
> 
> The failures are related to the Solana slides that have been fixed in another PR: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/a9b9cd77-f914-4c54-80c1-b7b4667340a7

@chrisleewilcox got it, just so we're clear nothing that isn't fixed in that other PR is showing up. Looks like thats the case in bitrise

ffmcgee725:
@adonesky1 

https://github.com/user-attachments/assets/3c602b24-066d-4dc4-b7b8-c4c05f36ce6e

Actually this seems to be an issue with the globally selected chain apparently... notice how when we only have a single permitted scope, `eth_chainId` returns the proper hex chain, but when we have multiple permitted scopes, `eth_chainId` returns the hex chain for the first one (which I assume is the globally selected one)

tommasini (approved):

Prithpal-Sooriya:
Smoked skipped due to a flakey test with Notifications.
Validated that this should not block this PR

CC @MetaMask/notifications

PR #14784, opened 2025-04-23
fix: Do not initialize RemoteFeatureFlagController with state

mcmire:
## **Description**

Currently, the state kept by RemoteFeatureFlagController is persisted and passed back into the controller on a reload. This creates a problem if we want to place another controller in a certain mode based on a feature flag. For instance, if we want to call `NetworkController.enableRpcFailover` when the `walletFrameworkRpcFailoverEnabled` flag is true, we might think to use a `RemoteFeatureFlagController:stateChange` subscription handler to wait until RemoteFeatureFlagController knows about this flag. However, if this flag is already true then the RemoteFeatureFlagController state will never change, and so `NetworkController.enableRpcFailover` will never be called.

To fix this, this commit effectively ignores the RemoteFeatureFlagController persisted state, allowing the controller to start fresh each time the app is loaded. This behavior solves the problem mentioned above, and it is also consistent with the way extension works.

## **Related issues**

This bug was discovered while working on another feature and is a prerequisite to https://github.com/MetaMask/metamask-mobile/issues/14120.

## **Manual testing steps**

The changes in this PR should not be user-visible.

## **Screenshots/Recordings**

(N/A)

### DIFF
--- a/app/core/Engine/Engine.test.ts
+++ b/app/core/Engine/Engine.test.ts
@@ -277,6 +277,20 @@ describe('Engine', () => {
     });
   });
 
+  it('does not pass initial RemoteFeatureFlagController state to the controller', () => {
+    const state = {
+      RemoteFeatureFlagController: {
+        remoteFeatureFlags: {},
+        cacheTimestamp: 20000000000000,
+      },
+    };
+    const engine = Engine.init(state);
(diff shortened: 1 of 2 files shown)


Cal-L (approved):
LGTM

PR #14839, opened 2025-04-24
fix: add percent change for 24h on asset list

sahar-fehri:
## **Description**

Adds asset list 24h price change

## **Related issues**

Fixes:

## **Manual testing steps**

1. Switch to a Solana account
2. You should be able to see the 24h percent change on the asset list and aggregated balance

### **After**

Screenshot 2025-04-24 at 15 44 55


gambinish:
Noticing a couple things:

1. The percent change on the main token list is different than the AssetOverview screen. Is this expected?

https://github.com/user-attachments/assets/63b2cc9f-5f3d-4bb7-be37-936e9198f480

2. For USDC, I'm noticing that it's `-0.00%`. This seems like an odd display value. On AssetOverview we are showing a different value of +0.00%. Which is correct? Neither seems logical, and I wonder what the intended behavior should be. Maybe we can use `formatWithThreshold` to display `-<0.01%` in red when the difference is below the smallest display value (say `-0.00001%`). We could also do something similar if the trend is upwards, and display `+<0.01%` in green.

gambinish (dismissed):
Approving since my comments aren't technically regressions, and this is the existing behavior.

We should probably revisit this in the future.

sahar-fehri:
TODO: update title with cp-7.46.0

owencraston:
In testing this seemed to work fairly well but I am noticing a few things. Namely the fact that the percentage change is different on the token list vs the asset overview. I tested with eth accounts and noticed this is the case as well.

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  const nonEvmAccountBalance = getMultichainNetworkAggregatedBalance(
+    account as unknown as InternalAccount,
+    multichainBalances,
+    multichainAssets,
+    multichainAssetsRates,
+    nonEvmChainId,
+  );
+
+  const nonEvmFiatBalancesArray = Object.entries(
```
owencraston, line 56: Both this and nonEvmAccountBalance should be memoized. The `Object.entries` will eagerly create a new object on every render which can trigger more re renders. Instead we can warp these two in a useMemo.

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const currentCurrency = useSelector(selectCurrentCurrency);
+
+  const account = useSelector(selectSelectedInternalAccount);
+  const multichainBalances = useSelector(selectMultichainBalances);
+  const multichainAssets = useSelector(selectMultichainAssets);
+  const multichainAssetsRates = useSelector(selectMultichainAssetsRates);
+  const nonEvmChainId = useSelector(selectSelectedNonEvmNetworkChainId);
+
+  const nonEvmAccountBalance = getMultichainNetworkAggregatedBalance(
+    account as unknown as InternalAccount,
```
owencraston, line 49: I think the casting is unnecessary here. If not we should explicit mark the return type of this selector to be InternalAccount.

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
```diff
@@ -0,0 +1,146 @@
...
+        variant={TextVariant.BodyMDMedium}
+        testID={FORMATTED_PERCENTAGE_TEST_ID}
+      >
+        {formattedPercentage}
+      </SensitiveText>
+    </View>
+  );
+};
+
+export default NonEvmAggregatedPercentage;
```
owencraston, line 146: lets wrap this whole thing in a React.memo

app/components/UI/AssetOverview/AssetOverview.tsx
```diff
@@ -255,7 +278,10 @@ const AssetOverview: React.FC<AssetOverviewProps> = ({
       )),
     [handleSelectTimePeriod, timePeriod],
   );
-  const itemAddress = safeToChecksumAddress(asset.address);
+
+  const itemAddress = isEvmSelected
```
owencraston, line 282: In see we are doing lots of this across the ap. Perhaps we can turn this into a util.

app/components/UI/AssetOverview/TokenDetails/TokenDetails.tsx
```diff
@@ -48,103 +59,125 @@ interface TokenDetailsProps {
...
+    : asset.address;
 
-  let tokenMetadata;
-  let marketData;
+  const tokenSearchResult = useSelector((state: RootState) =>
+    selectTokenDisplayData(state, asset.chainId as Hex, asset.address as Hex),
+  );
+
+  ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+  const allMultichainAssetsRates = useSelector(selectMultichainAssetsRates);
```
owencraston, line 88: Lets make sure this selector is a deep equal selector. I did not verify that it is/is not. Just wanted to mention it.

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  if (!totalNonEvmBalance) {
+    return null;
+  }
+
+  const totalNonEvmBalance1dAgo = getNonEvmTotalFiat1dAgo();
+  const amountChange = totalNonEvmBalance - totalNonEvmBalance1dAgo;
+
+  const percentageChange =
+    ((totalNonEvmBalance - totalNonEvmBalance1dAgo) / totalNonEvmBalance1dAgo) *
```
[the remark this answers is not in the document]
owencraston: Also can we just reuse the `amountChange` var from above?

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
```diff
@@ -0,0 +1,146 @@
...
+  );
+
+  const nonEvmFiatBalancesArray = Object.entries(
+    nonEvmAccountBalance?.fiatBalances ?? {},
+  ).map(([id, amount]) => ({
+    id: id as CaipAssetType,
+    fiatAmount: Number(amount),
+  }));
+
+  const getNonEvmTotalFiat1dAgo = () => {
```
owencraston, line 63: Can we also memoize this?

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
```diff
@@ -0,0 +1,146 @@
...
+        currentCurrency,
+        DECIMALS_TO_SHOW,
+      )} `
+    : '';
+
+  return (
+    <View style={styles.wrapper}>
+      <SensitiveText
+        isHidden={privacyMode}
+        length="10"
```
owencraston, line 126: Is the length always going to be 10?

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
sahar-fehri: I can make the return type explicit in the selector;
Although i think i might need to still cast because the return type is InternalAccount | undefined

app/component-library/components-temp/Price/AggregatedPercentage/NonEvmAggregatedPercentage.tsx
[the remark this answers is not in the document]
sahar-fehri: Good point; i can add a check but i think that to be in that case everything would have to be zero
if 
const totalNonEvmBalance1dAgo = getNonEvmTotalFiat1dAgo();
is zero then totalNonEvmBalance is also zero in which case we return null
```
  if (!totalNonEvmBalance) {
    return null;
  }
```

app/components/UI/AssetOverview/AssetOverview.tsx
sahar-fehri: I haven't touched this component i think you see this change now because the branch i was branched off of was merged; good point though can add a TODO seems like it would not be a small refactor ^^

app/components/UI/AssetOverview/TokenDetails/TokenDetails.tsx
sahar-fehri: Will update this, i also think this was not originally modifed in this PR but worth improving 🙏

sahar-fehri:
closing in favor of https://github.com/MetaMask/metamask-mobile/pull/14892

sahar-fehri:
> hings. Namely the fact that the p

This is happening also in extension; for evms too; both values are computed differently;
One mainly from price-api (historical prices endpoint) (the one on the asset details)
While the percentage on the aggregated balance view is computed based on the user's balance and the market data we get from price api.

PR #14927, opened 2025-04-28
perf: Debounce and group state updates being flushed to Redux

FrederikBolding:
## **Description**

Debounce and group state updates being flushed to Redux since dispatching the Redux actions has proved to be slow and often tied directly to the UI being rendered at the time. It is less than ideal for a controller state update to be waiting for this to happen. By grouping and debouncing, the Redux dispatch is deferred and simultaneous state updates get grouped into one dispatch.

### DIFF
--- a/app/core/EngineService/EngineService.ts
+++ b/app/core/EngineService/EngineService.ts
@@ -106,11 +106,6 @@ export class EngineService {
 
     const updatedControllers = new Set<string>()
 
-    const updateReduxStateDebounced = (controllerName: string) => {
-      updatedControllers.add(controllerName);
-      flushStateDebounced();
-    }
-
     const flushStateDebounced = debounce(() => {
       if (!engine.context.KeyringController.metadata.vault) {
         Logger.log('keyringController vault missing for UPDATE_BG_STATE_KEY');
@@ -121,6 +116,11 @@ export class EngineService {
       });
       updatedControllers.clear();
     }, 200);
+
+    const updateReduxStateDebounced = (controllerName: string) => {
+      updatedControllers.add(controllerName);
+      flushStateDebounced();
+    }
 
     BACKGROUND_STATE_CHANGE_EVENT_NAMES.forEach((eventName) => {
       engine.controllerMessenger.subscribe(eventName, () =>
--- a/app/core/EngineService/EngineService.ts
+++ b/app/core/EngineService/EngineService.ts
@@ -9,12 +9,12 @@ import {
 import { getTraceTags } from '../../util/sentry/tags';
 import { trace, endTrace, TraceName, TraceOperation } from '../../util/trace';
 import getUIStartupSpan from '../Performance/UIStartup';
-import { BACKGROUND_STATE_CHANGE_EVENT_NAMES } from '../Engine/constants';
 import ReduxService from '../redux';
 import NavigationService from '../NavigationService';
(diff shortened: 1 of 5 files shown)


app/core/EngineService/EngineService.ts
```diff
@@ -103,19 +104,29 @@ export class EngineService {
       () => !this.engineInitialized,
     );
 
-    const update_bg_state_cb = (controllerName: string) => {
+    const updatedControllers = new Set<string>();
+
+    // Reduce the amount of state updates being flushed by using debouncing
+    // and a running list of controllers to update.
+    const flushStateDebounced = debounce(() => {
```
Prithpal-Sooriya, line 111: Do we want to debounce state updates or throttle them?

FrederikBolding: I think debouncing makes the most sense because we don't want controllers to be waiting for the state to be flushed to the UI. But there is probably an argument for doing something more complex than a simple debounce.

app/core/EngineService/EngineService.ts
```diff
@@ -103,19 +104,29 @@ export class EngineService {
...
       if (!engine.context.KeyringController.metadata.vault) {
         Logger.log('keyringController vault missing for UPDATE_BG_STATE_KEY');
       }
       ReduxService.store.dispatch({
         type: UPDATE_BG_STATE_KEY,
-        payload: { key: controllerName },
+        payload: { updatedControllers: updatedControllers.values() },
       });
+      updatedControllers.clear();
+    }, 200);
```
tommasini, line 120: We could specify the configuration of trailing and leading, just for vidibility purposes,  we are doing the same where we write to the storage in persistConfig

tommasini (approved):

app/core/EngineService/EngineService.ts
FrederikBolding: I guess it's not ideal if we delay interface updates too much either, so it may require more thought 🤔 

But certainly we shouldn't be blocking controllers from continuing execution to sync the UI..

Prithpal-Sooriya: My only concern with debouncing is that if there are a lot of updates here, then the redux dispatch could be blocked. Whereas with throttling we will at least have consistent updates dispatched to our redux store.

Example demo:
https://web.archive.org/web/20220117092326/http://demo.nimius.net/debounce_throttle/

Prithpal-Sooriya: TBH I'm open to both.

Since our persistence is in our redux layer (redux persist), I think this change will not only help our UI updates but also reduce the persistence spam.

FrederikBolding: Yeah, that's a fair point. But we would want to configure throttle to apply to the initial call too I think

Cal-L (approved):
Would be good to make debounce config explicit - https://github.com/MetaMask/metamask-mobile/pull/14927/files#r2071433223 so readers know the behavior. Other than that, LGTM!

app/core/EngineService/EngineService.ts
FrederikBolding: Yeah, I think I'm convinced that debounce is not the way. Moved the PR to draft for now.

FrederikBolding:
Closing in favor of other batching changes

PR #14958, opened 2025-04-29
feat: add asset selector

### LINKED ISSUE #3285: Port `AssetSelector` component on mobile

hmalik88:
## **Description**

Write a short description of the changes included in this pull request, also include relevant motivation and context. Have in mind the following questions:
1. What is the reason for the change? Porting over `AssetSelector` from extension.

## **Related issues**

Fixes: https://github.com/MetaMask/snaps/issues/3285

## **Manual testing steps**

1. Pull down the branch and build
2. Edit the send-flow snap (and test-snaps) with the following snippet:
```ts
            <Box direction="vertical">
              <Field label="Asset">
                <AssetSelector
                  name="asset1"
                  addresses={[
                    'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:4puok8BCAGTbEEX4qgtxYkjk4yNi7xmnf4R8kq68xyPn',
                  ]}
                />
              </Field>
              <Box direction="horizontal">
                <Field label="Asset">
                  <AssetSelector
                    name="asset"
                    addresses={[
                      'solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp:4puok8BCAGTbEEX4qgtxYkjk4yNi7xmnf4R8kq68xyPn',
                    ]}
                  />
                </Field>
                <Field label="Send amount">
                  <Input
                    name="amount"
                    type="number"
                    placeholder="Enter amount to send"
                  />
                </Field>
              </Box>
            </Box>
```
3. Run test-snaps locally
4. Navigate to the locally ran test-snaps in the simulator
5. Observe the results.

### **After**

**Recording**:

https://github.com/user-attachments/assets/97845adf-bcf5-4be4-a9af-5ef2314cf32f

**Disabled State**

Screenshot 2025-05-05 at 9 12 44 PM

**Error State**

Screenshot 2025-05-05 at 9 11 48 PM


app/component-library/components/Form/TextField/TextField.tsx
```diff
@@ -39,6 +40,7 @@ const TextField = React.forwardRef<TextInput, TextFieldProps>((
 
   const { styles } = useStyles(styleSheet, {
     style,
+    inputStyle,
```
FrederikBolding, line 43: Ideally we can do this without changing DS components. Why is this needed?

app/components/Snaps/SnapUIRenderer/components/box.ts
```diff
@@ -59,7 +59,11 @@ export const box: UIComponentFactory<BoxElement> = ({
 }) => ({
   element: 'Box',
   children: getJsxChildren(e).map((children) =>
-    mapToTemplate({ ...params, element: children as JSXElement }),
+    mapToTemplate({
+      ...params,
+      element: children as JSXElement,
+      parentIsFlexRow: e.props.direction === 'horizontal',
```
FrederikBolding, line 65: ?

hmalik88: I was trying to replicate this: https://github.com/MetaMask/metamask-extension/pull/30858/files#diff-98fab621c35ee8064316025b4cb009e9d9b51fdf32207f7ff333587be4fd49df 

Not sure if this is the proper way to do it

app/component-library/components/Form/TextField/TextField.tsx
hmalik88: Trying to do this: https://github.com/MetaMask/metamask-extension/pull/30858/files#diff-2acc20b66f1b5349b6c521fedd297661091a59304a8657de648006cecce61def

Any suggestions how I can do this different?

FrederikBolding: Wouldn't it work by just applying to the `style` prop directly for `TextField`?

app/components/Snaps/SnapUIRenderer/components/box.ts
FrederikBolding: For some reason that link doesn't work for me 😄

app/core/Multichain/constants.ts
```diff
@@ -6,8 +6,98 @@ import {
   SolScope,
 } from '@metamask/keyring-api';
 import BTC from '../../images/bitcoin-logo.png';
+import BTC_SVG from '../../images/bitcoin-logo.svg';
```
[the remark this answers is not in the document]
hmalik88: yeah, I can do that. I'm not confident this is all working yet though.

app/components/Snaps/SnapUIRenderer/components/box.ts
hmalik88: From the CSS I understood that to mean, we should apply this styling to asset selector when we have a field that is inside a box with flex direction row

app/component-library/components/Form/TextField/TextField.tsx
hmalik88: No because that style prop gets used in the stylesheet to construct the base portion of the styles and not the input.

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
```diff
@@ -0,0 +1,148 @@
...
+        asset.chainId as AllowedBridgeChainIds
+      ] ?? networks[asset.chainId]?.name;
+
+    return {
+      icon: asset.image,
+      symbol: asset.symbol,
+      name: asset.title,
+      balance: formatAssetBalance(asset.primary),
+      networkName,
+      networkIcon: getImageForChainId(asset.chainId),
```
GuillaumeRx, line 119: This should use: https://github.com/MetaMask/metamask-mobile/blob/07aff153e1c768e2647b125dee7cf96f73b765be/app/util/networks/customNetworks.tsx#L153

app/core/Multichain/constants.ts
GuillaumeRx, line 1: We should drop all of these icons, we don't need them as the asset selector only supports NON-EVM networks and most of the icons are already available here: https://github.com/MetaMask/metamask-mobile/blob/main/app/util/networks/customNetworks.tsx

GuillaumeRx (changes requested):

app/component-library/components/Form/TextField/TextField.tsx
hmalik88: Was mistaken, ended up removing this.

app/constants/assets.ts
GuillaumeRx, line 1: Do we really have to bring all of these types in ?

GuillaumeRx: We should also make sure that the type matches, I can see that they use [this](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/UI/Tokens/types.ts#L6) to type the selector return [here](https://github.com/MetaMask/metamask-mobile/blob/ca72707883d536d70a7f433662b6b5b88fcb89bd/app/components/UI/Bridge/hooks/useTokensWithBalance/index.ts#L158)

hmalik88: Ah! thanks I'll use that

GuillaumeRx: Actually disregard what I said it's the wrong selector, but the actual selector is written in Typescript, does it not infers the correct type ?

hmalik88: I didn't actually try letting it infer, I'll try it out

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
```diff
@@ -40,16 +44,32 @@ const SelectorItem: React.FunctionComponent<SelectorItemProps> = ({
...
-  const { styles } = useStyles(stylesheet, {});
+  const { styles } = useStyles(stylesheet, { selected });
 
   const handlePress = () => {
     onSelect(value);
   };
 
+  // We pass the context prop to the children so that the component
+  // can decide to apply some additional UI based on the context.
+  const buttonChildren = React.isValidElement(children) ? (
```
FrederikBolding, line 57: I would really like if we could avoid cloning, what is the use here?

app/selectors/multichainBalancesController.test.ts
FrederikBolding, line 1: ?

app/constants/assets.ts
```diff
@@ -0,0 +1,60 @@
+import { CaipAssetType, CaipChainId, Hex } from '@metamask/utils';
+
+// Common mixin for primary and secondary display values
+export interface TokenDisplayValues {
```
FrederikBolding, line 4: Do these not exist already in mobile? 🤔

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,149 @@
...
+    disabled: false,
+  }));
+
+  const optionComponents = assets.map((asset, index) => (
+    <SnapUIAssetSelectorOption {...asset} key={index} isParentFlexRow={isParentFlexRow} />
+  ));
+
+  return (
+    <SnapUISelector
+      title={strings('app_settings.snaps.snap_ui.asset_selector.title')}
```
FrederikBolding, line 140: `app_settings`?

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,149 @@
...
+        ellipsizeMode="tail"
+      >
+        {networkName}
+      </Text>
+    </Box>
+    <Box
+      flexDirection={FlexDirection.Column}
+      // We hide the balance and fiat in inline mode when the asset selector has a sibling element
+      // eslint-disable-next-line react-native/no-inline-styles
+      style={{ marginLeft: 'auto', ...(context === 'inline' && isParentFlexRow ? { display: 'none' } : {}) }}
```
FrederikBolding, line 85: Can you explain what the `context` prop is supposed to be for?

app/components/Snaps/SnapUIRenderer/components/box.ts
```diff
@@ -59,7 +59,14 @@ export const box: UIComponentFactory<BoxElement> = ({
...
   element: 'Box',
   children: getJsxChildren(e).map((children) =>
-    mapToTemplate({ ...params, element: children as JSXElement }),
+    mapToTemplate({
+      ...params,
+      element: children as JSXElement,
+      // We have to pass this information down to the children so that they can
+      // hide the balance and fiat in inline mode when the asset selector has a sibling element
+      // and to reduce the padding of the snap ui selector when it has a sibling element.
+      isParentFlexRow: e.props.direction === 'horizontal',
```
FrederikBolding, line 68: Maybe at some point it would be easier to pass a `parent` instead of a direct flag 🤔

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
FrederikBolding: Even if we need to re-apply some props we shouldn't need to do that for both the options and the selected value

app/components/Snaps/SnapUIRenderer/components/box.ts
hmalik88: Wdym? Like pass the whole element?

hmalik88: Probably a good idea as we see more use cases where we need information from the parent

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
hmalik88: Yup, explainer below -- this also addresses your comment here: https://github.com/MetaMask/metamask-mobile/pull/14958#discussion_r2073359339

The `context` prop was added because we don't have information as to where the `SnapUIAssetSelectorOption` component is being rendered until it gets to the `SnapUISelector`. If we don't include this information then we get UI like this:


Screenshot 2025-05-05 at 2 48 13 PM

app/component-library/components/Badges/BadgeWrapper/BadgeWrapper.tsx
```diff
@@ -21,10 +21,20 @@ const BadgeWrapper: React.FC<BadgeWrapperProps> = ({
...
     useComponentSize();
 
+  /**
+   * The parentSize indicates the height of the badge wrapper's parent element.
+   * In some cases, the computed badge position offset is not correct because the style
+   * gets computed before the hook's size has had a chance to update.
+   * This is a workaround to ensure the offset is correct on first render when the
+   * parentSize is known.
+   */
+  const containerSize = parentSize ? { height: parentSize, width: parentSize } : size;
```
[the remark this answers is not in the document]
hmalik88: I don't have a great explanation for this, but the logs show that the `useComponentSize` hook isn't capturing the container size on first mount, it shows it to be 0. I tried using the `Badge` component directly, same issue. I'm suspecting that the hook is slowed down because of state updates or something...I checked behavior after reverting the changes I made here in the `TokenListItem` component and it doesn't look great. Perhaps the hook will be reliable after perf is better? As you can see in the clip below there's a major lag to get the badge element to be the correct size/position. Adding the `parentSize` ensures consistency between all renders (at least for our screen).

https://github.com/user-attachments/assets/6b9b6447-046a-4497-b500-4f2aa9cdb43b

app/selectors/multichainBalancesController.ts
```diff
@@ -0,0 +1,10 @@
+import { RootState } from '../reducers';
+import { createDeepEqualSelector } from './util';
+
+const selectMultichainBalancesControllerState = (state: RootState) =>
```
FrederikBolding, line 4: This already exists: https://github.com/MetaMask/metamask-mobile/blob/35587c3f268819dc7ab2008af349ebe79494b3f6/app/selectors/multichain/multichain.ts#L108-L112

app/components/Snaps/SnapUIRenderer/components/box.ts
FrederikBolding: At least passing the parent props or something, but lets not do it in this PR

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
hmalik88: P.S. I think `React.cloneElement` is pretty lightweight. The alternatives I suppose would be render props or context, but that feels like more complexity then needed. If your concern is a large list of tokens, I think at that point the larger issue would be the `ScrollView` trying to render everything at once. I don't think a user will have hundreds of tokens, but if it turns out to be problematic, we can use `FlatList` (which I believe would only clone the visible elements).

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
hmalik88: See https://github.com/MetaMask/metamask-mobile/pull/14958#discussion_r2073988198

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
FrederikBolding: Just add a new section in this PR and move the new string there, we can do a follow up to move old strings

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
```diff
@@ -136,10 +174,11 @@ export const SnapUISelector: React.FunctionComponent<SnapUISelectorProps> = ({
             >
               {optionComponents.map((component, index) => (
                 <SelectorItem
-                  key={options[index].value}
+                  key={`snap-ui-selector-${options[index].value.toString()}`}
```
FrederikBolding, line 177: This key will be duplicated for all the objects since `options[index].value.toString()` is just gonna be `[object Object]`

app/component-library/components/Badges/BadgeWrapper/BadgeWrapper.tsx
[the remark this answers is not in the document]
FrederikBolding: I can't reproduce this problem, when I remove `parentSize` this branch works as-is 🤔

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
hmalik88: Woops, meant to stringify but wondering if that might be too expensive? Should i just use a uuid?

FrederikBolding: We use the index on the extension, might be fine for now tbh

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
FrederikBolding: That is what I was trying to propose here too: https://github.com/MetaMask/metamask-mobile/pull/14958#discussion_r2073415521

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * Gets the assets from state and format them for the SnapUIAssetSelector.
+ *
+ * @param params - The parameters for the hook.
+ * @param params.addresses - The addresses to get the assets for.
+ * This is a list of the same address but for different chains.
+ * @param params.chainIds - The chainIds to filter the assets by.
+ * @returns The formatted assets.
+ */
+export const useSnapAssetSelectorData = ({
```
FrederikBolding, line 66: @GuillaumeRx should approve this before we merge

app/component-library/components/Badges/BadgeWrapper/BadgeWrapper.tsx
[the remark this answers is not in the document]
hmalik88: I wonder if the perf improvements made this a non-issue, it was definitely happening to me before. I'll try again locally and remove this if that's the case.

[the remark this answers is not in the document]
FrederikBolding: Could also be a platform specific issue, I am testing on Android right now. But I am assuming there is a way to fix this without modifying DS components since every other component using this pattern works?

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @param props - The component props.
+ * @param props.addresses - The addresses to get the assets for.
+ * @param props.chainIds - The chainIds to filter the assets by.
+ * @param props.disabled - Whether the selector is disabled.
+ * @returns The AssetSelector component.
+ */
+export const SnapUIAssetSelector: FunctionComponent<
+  SnapUIAssetSelectorProps
+> = ({ addresses, chainIds, disabled, style, isParentFlexRow, ...props }) => {
```
FrederikBolding, line 147:
```suggestion
> = ({ addresses, chainIds, disabled, style, compact, ...props }) => {
```

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,171 @@
...
+  }
+> = ({
+  icon,
+  symbol,
+  name,
+  balance,
+  fiat,
+  networkName,
+  networkIcon,
+  isParentFlexRow = false,
```
FrederikBolding, line 44:
```suggestion
  compact= false,
```

app/components/Snaps/SnapUIRenderer/components/field.ts
```diff
@@ -157,6 +159,28 @@ export const field: UIComponentFactory<FieldElement> = ({
...
+
+      return {
+        ...assetSelectorMapped,
+        element: 'SnapUIAssetSelector',
+        props: {
+          ...assetSelectorMapped.props,
+          label: e.props.label,
+          form,
+          error: e.props.error,
+          isParentFlexRow: params.isParentFlexRow,
```
FrederikBolding, line 178:
```suggestion
          compact: params.isParentFlexRow,
```

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
```diff
@@ -89,29 +111,45 @@ export const SnapUISelector: React.FunctionComponent<SnapUISelectorProps> = ({
...
+      ? option.value[option.key as keyof typeof option.value] ===
+        selectedOptionValue?.[option.key as keyof typeof selectedOptionValue]
+      : option.value === selectedOptionValue,
   );
 
-  const selectedOption = optionComponents[selectedOptionIndex];
+  const selectedOptionComponent = optionComponents[selectedOptionIndex];
+
+  // We clone the selected option component so that we can pass the render context to it.
+  const inlineButtonLabel = React.isValidElement(selectedOptionComponent) ? (
```
FrederikBolding, line 135: When would it not be a valid element?

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
```diff
@@ -0,0 +1,157 @@
...
+}: UseSnapAssetSelectorDataParams) => {
+  const currentCurrency = useSelector(selectCurrentCurrency);
+  const locale = I18n.locale;
+
+  const parsedAccounts = addresses.map(parseCaipAccountId);
+
+  const account = useSelector((state) =>
+    getMemoizedInternalAccountByAddress(state, parsedAccounts[0].address),
+  );
+  const networks = useSelector(selectNetworkConfigurations);
```
GuillaumeRx, line 78: This selector should probably be memoized as it is not right now: https://github.com/MetaMask/metamask-mobile/blob/3047b520e6667da51c745f2dc0c5d5e4af0949ac/app/selectors/networkController.ts#L185C44-L185C58

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const assets = useSelector((state: RootState) =>
+    selectMultichainTokenListForAccountId(state, account?.id)
+  ) as unknown as TokenWithFiatAmount[];
```
GuillaumeRx, line 82: Do we really have to cast here ? I expect that the selector infers a type if you have to cast it as `unknown`

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { selectNetworkConfigurations } from '../../../selectors/networkController';
+import { AllowedBridgeChainIds, NETWORK_TO_SHORT_NETWORK_NAME_MAP } from '../../../constants/bridge';
+import { selectCurrentCurrency } from '../../../selectors/currencyRateController';
+import { RootState } from '../../../reducers';
+import { getNonEvmNetworkImageSourceByChainId } from '../../../util/networks/customNetworks';
+
+/**
+ * An asset for the SnapUIAssetSelector.
+ */
+export interface SnapUIAsset {
```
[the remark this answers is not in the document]
hmalik88: it's a mobile thing, linter starts screaming -- not sure why they opt for interface over type in this codebase 🤷‍♂️

GuillaumeRx (changes requested):

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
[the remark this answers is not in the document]
GuillaumeRx: Sounds good :+1:

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
hmalik88: Yeah you're right, I was thinking plain text, but these are components regardless.

app/component-library/components/Badges/BadgeWrapper/BadgeWrapper.tsx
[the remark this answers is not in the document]
hmalik88: Yeah, it might be platform specific. Removed this logic and this is the result:


https://github.com/user-attachments/assets/d2990009-066e-41aa-bedd-3bd32490d197

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
hmalik88: The casting seems to be necessary, `formatAsset` starts to complain about the types. I did remove the `unknown` though.

app/components/Snaps/SnapUIAssetSelector/useSnapAssetDisplay.tsx
hmalik88: Good catch, updated.

app/selectors/multichainBalancesController.ts
FrederikBolding: Why was this re-added?

hmalik88: That's weird, not sure how those got added in again. Git is being weird, removed them again.

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,172 @@
...
+      flexDirection={FlexDirection.Row}
+      gap={16}
+      // eslint-disable-next-line react-native/no-inline-styles
+      style={{ overflow: 'hidden', flex: 1 }}
+    >
+      {/* eslint-disable-next-line react-native/no-inline-styles */}
+      <Box alignItems={AlignItems.center}>
+        <BadgeWrapper
+          badgeElement={
+            <AvatarNetwork
```
FrederikBolding, line 69: Let's use `Badge` here as most other instances in the codebase does

app/component-library/components/Badges/BadgeWrapper/BadgeWrapper.tsx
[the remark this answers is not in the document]
FrederikBolding: I can't reproduce this locally, IMO we just take it out for now and mirror all the other uses with `Badge`?

locales/languages/en.json
```diff
@@ -1145,7 +1145,7 @@
       "snap_ui": {
         "link": {
           "accessibilityHint": "Opens in a new tab"
-        }
+        },
```
FrederikBolding, line 1148:
```suggestion
        }
```

app/component-library/components/Badges/BadgeWrapper/BadgeWrapper.tsx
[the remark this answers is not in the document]
hmalik88: That is so weird...can we get one other person to test this locally? @david0xd perhaps? I tried it out with `Badge` and it gets a bit more complicated because the `BadgeNetwork` component handles the layout sizing themselves so the styles that I'm applying at the `BadgeWrapper` get overridden iirc. I'll try to swap out again and report back if it gets complicated.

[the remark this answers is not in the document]
FrederikBolding: And I really wanna avoid modifying DS components so we can get this merged right away

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,163 @@
...
+        }
+        badgePosition={BadgePosition.BottomRight}
+      >
+        <AvatarToken size={AvatarSize.Sm} imageSource={{ uri: icon }} />
+      </BadgeWrapper>
+    </Box>
+    <Box
+      flexDirection={FlexDirection.Column}
+      // eslint-disable-next-line react-native/no-inline-styles
+      style={{ overflow: 'hidden' }}
```
FrederikBolding, line 78:
```suggestion
      style={{ overflow: 'hidden', flex: 1 }}
```

app/components/Snaps/SnapUIAssetSelector/SnapUIAssetSelector.tsx
```diff
@@ -0,0 +1,163 @@
...
+    >
+      <Text variant={TextVariant.BodyMDMedium} ellipsizeMode="tail">
+        {name}
+      </Text>
+      <Text
+        color={TextColor.Alternative}
+        variant={TextVariant.BodySM}
```
FrederikBolding, line 85:
```suggestion
        variant={TextVariant.BodySM}
        numberOfLines={1}
```

FrederikBolding (dismissed):
LGTM - please wait for @GuillaumeRx to sign off too

FrederikBolding (approved):

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
```diff
@@ -89,29 +103,44 @@ export const SnapUISelector: React.FunctionComponent<SnapUISelectorProps> = ({
...
   );
 
-  const selectedOption = optionComponents[selectedOptionIndex];
+  const selectedOptionComponent = optionComponents[selectedOptionIndex];
+
+  // We clone the selected option component so that we can pass the render context to it.
+  const inlineButtonLabel = React.cloneElement(
+    selectedOptionComponent as React.ReactElement,
+    { context: 'inline' },
+  );
```
FrederikBolding, line 130: Hold on, seems like `selectedOptionComponent` may be undefined, we need to deal with that.

The Solana send flow is crashing as-is.

FrederikBolding (changes requested):
See latest comment

app/components/Snaps/SnapUISelector/SnapUISelector.tsx
FrederikBolding:
```suggestion
  const inlineButtonLabel = selectedOptionComponent && React.cloneElement(
    selectedOptionComponent as React.ReactElement,
    { context: 'inline' },
  );
```

FrederikBolding (approved):

GuillaumeRx (approved):
Code LGTM

PR #15020, opened 2025-04-30
chore: add controllers metadata

tommasini:
## **Description**

This PR brings the functionality to respect the controllers metadata `persist` property, using the transform feature of redux-persist. So every controller data this is assigned as with `persist` metadata as `false` will not be kept in storage.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/core/Engine/Engine.ts
```diff
@@ -2094,6 +2095,26 @@ export default {
     return instance.datamodel;
   },
 
+  get stateWithMetadata() {
```
Gudahtt, line 2098: Nit: Why use a getter here rather than a method?

This behaves more like a method than a property (e.g. it can throw an error), so using a getter seems more confusing than helpful.

I typically only see getters used when we want to comply with some pre-existing interface where a property was already declared, or for cases where the getter behaves like a property (e.g. never throws, doesn't perform any significant operations, like if you wanted to create a property alias for example)

app/core/redux/slices/engine/index.ts
```diff
@@ -27,14 +27,16 @@ const engineReducer = (
...
-      return { backgroundState: Engine.state };
+      return { backgroundState: Engine.stateWithMetadata };
     }
     case updateBgState.type: {
       const newState = { ...state };
 
       if (action.payload) {
         const newControllerState =
-          Engine.state[action.payload.key as keyof typeof Engine.state];
+          Engine.stateWithMetadata[
```
Gudahtt, line 37: What is the purpose of putting controller metadata in Redux? We don't use it anywhere in the UI or in reducers, so it doesn't seem helpful to include here

app/util/logs/index.ts
Gudahtt, line 1: We may want to consider keeping some KeyringController state in state logs. It can be very useful for diagnosing problems.

tommasini: I kept the keyrings and the isUnlocked, any other state property that do you think it would be useful?

app/store/getPersistentState/getPersistentState.ts
```diff
@@ -0,0 +1,61 @@
...
+): Record<keyof ControllerState, Json> {
+  return (Object.keys(state) as (keyof ControllerState)[]).reduce<
+    Record<keyof ControllerState, Json>
+  >((derivedState, key) => {
+    try {
+      const stateMetadata = metadata[key];
+      if (!stateMetadata) {
+        // If there is no metadata for this key, include the state property (THIS IS CHANGED FROM BASE CONTROLLER)
+        derivedState[key] = state[key];
+        return derivedState;
```
tommasini, line 45: What approach do we want to take, do we want this one? or we want to use the old logic, to capture the error with sentry instead of throwing it, since it was causing a crash. https://github.com/MetaMask/core/blob/c580f85e42eb549ce767fb377eb0599a9900ab9f/packages/base-controller/src/BaseControllerV2.ts#L383

app/util/logs/index.ts
Gudahtt: keyringMetadata tells us what's in the vault without including any sensitive data, that would be most helpful here

tommasini: Thanks Mark! Here: https://github.com/MetaMask/metamask-mobile/pull/15020/commits/60948d4d19f49a36ee8121ebdc0a4b3298224103

app/store/persistConfig.ts
```diff
@@ -70,33 +81,46 @@ const MigratedStorage = {
...
-    const {
-      aggregatorMetadata,
-      aggregatorMetadataLastFetched,
-      chainCache,
-      tokens,
-      tokensLastFetched,
-      topAssets,
-      ...persistedSwapsController
-    } = SwapsController;
+    const filteredControllers = Object.entries(controllers).reduce<
```
Cal-L, line 104: This is a place to lookout for in terms of performance since transform will trigger frequently. On top of that, `getPersistentState` is already calling reduce under the hood. Let's change this reduce to a for loop to optimize performance as best as we can - https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/#arrayreduce-vs-for-and-forof

tommasini: Solved! Thank youuu! Using a for of now!

tommasini:
Skipping sonar cloud because we do not want to handle the error, just want to caught it at the transform function

e2e/fixtures/fixture-builder.js
```diff
@@ -298,70 +295,21 @@ class FixtureBuilder {
...
-                  '0x504': true,
-                  '0x507': true,
-                  '0x505': true,
-                  '0x64': true,
-                },
-              },
-              _X: null,
             },
             TokenBalancesController: {
-              contractBalances: {},
```
Gudahtt, line 348: For each of these updated properties, do we know if the old keys were actually in state in production, or was this a fixture error?

If these keys were in production, were they removed in a migration? If not, users will encounter errors upon update that we'll find out about in Sentry. In that case it would be better to pre-emptively delete these properties with a migration.

app/util/logs/index.ts
```diff
@@ -25,24 +25,24 @@ export const generateStateLogs = (state: any, loggedIn = true): string => {
   delete fullState.engine.backgroundState.PhishingController;
   delete fullState.engine.backgroundState.AssetsContractController;
 
-  // Remove encrypted vault from logs
-  delete fullState.engine.backgroundState.KeyringController.vault;
+  // Remove Keyring controller data
```
Gudahtt, line 28: Nit: The previous comment did a better job of explaining _why_ we're removing state. Comments that explain what we're doing are less helpful (it's clear already from reading the code that we're removing keyring controller state).

Here's a suggested improvement:
```suggestion
  // Remove Keyring controller data so that encrypted vault is not included in logs
```

e2e/fixtures/fixture-builder.js
tommasini: This issue were just appearing in e2e, but I didn't validate of they were removed by a migration indeed! I will check that!

app/util/logs/index.ts
tommasini: Love it! Thanks!

e2e/fixtures/fixture-builder.js
```diff
@@ -298,70 +295,21 @@ class FixtureBuilder {
...
+              tokenBalances: {},
             },
             TokenRatesController: {
               marketData: {},
             },
             TokensController: {
-              tokens: [],
```
tommasini, line 354: Removed by this migration: https://github.com/MetaMask/metamask-mobile/blob/5ea7189733bc5984408d5472b1e75c1df50e8958/app/store/migrations/074.ts#L27

e2e/fixtures/fixture-builder.js
```diff
@@ -749,14 +697,6 @@ class FixtureBuilder {
     // Update selectedNetworkClientId to the new network client ID
     networkController.selectedNetworkClientId = newNetworkClientId;
 
-    // Merge the rest of the data
-    merge(networkController, data);
-
-    if (data.providerConfig.ticker !== 'ETH') {
-      this.fixture.state.engine.backgroundState.CurrencyRateController.pendingNativeCurrency =
```
tommasini, line 756: removed by this migration: https://github.com/MetaMask/metamask-mobile/blob/d75ae4f7ddfc727f13a95189b05ec8006623a865/app/store/migrations/038.ts#L32

e2e/fixtures/fixture-builder.js
```diff
@@ -749,14 +697,6 @@ class FixtureBuilder {
     // Update selectedNetworkClientId to the new network client ID
     networkController.selectedNetworkClientId = newNetworkClientId;
 
-    // Merge the rest of the data
-    merge(networkController, data);
```
tommasini, line 753: providerConfig deleted here: https://github.com/MetaMask/metamask-mobile/blob/1d031703bf017a493b5f8eb0191f71e93e2ce207/app/store/migrations/053.ts#L23

e2e/fixtures/fixture-builder.js
tommasini: Added at these commits:
* https://github.com/MetaMask/metamask-mobile/pull/15020/commits/1c77ccd0a150793a70ada8b9631b3f6a32838f4e
* https://github.com/MetaMask/metamask-mobile/pull/15020/commits/9cce0076ab32572108b420798490057201f0dd03

Cal-L (approved):
LGTM

PR #15045, opened 2025-04-30
fix: Fix iOS build issue on Xcode 16.3

Cal-L:
## **Description**

Fix iOS build issues on Xcode 16.3 by bumping Sentry and Expo deps. RN sentry package is bumped a major version from v5 -> v6, which introduced some breaking changes under the hood. This also updates pods repo so a few other minor version bumps will exist in Podfile.lock as well. 

## **Related issues**

Fixes:

## **Manual testing steps**

How it was tested: Both Android and iOS apps should run after running `yarn start:ios` or `yarn start:android`

### **Before**

Individuals began experiencing iOS build issues with Xcode 16.3 as reported here - https://consensys.slack.com/archives/C02U025CVU4/p1746041871082239. The fix was to bump Expo SDK deps as well as Sentry.

Expo SDK deps - https://expo.dev/changelog/xcode-16-3-patches
Sentry issue - https://github.com/getsentry/sentry-cocoa/releases/tag/8.46.0 (we bumped to stable 8.48.0 with pod repo update)

Sentry spans were not associated correctly in v5
image

### **After**

Spans are now correctly associated with the parent span with v6
image


chrisleewilcox (dismissed):
LGTM

Cal-L:
@socket-security ignore-all

chrisleewilcox (dismissed):
Nice!

tommasini (dismissed):

cortisiko (dismissed):

owencraston (dismissed):
This fixed my iOS builds (using xcode 16.3)

app/util/trace.test.ts
```diff
@@ -5,8 +5,7 @@ import {
   startSpanManual,
   withScope,
 } from '@sentry/react-native';
-
-import { Span } from '@sentry/types';
+import type { Span } from '@sentry/core';
```
Cal-L, line 8: @sentry/types is deprecated. Types have been moved to @sentry/core

app/components/Views/confirmations/legacy/components/BlockaidBanner/BlockaidBanner.tsx
```diff
@@ -124,14 +124,19 @@ const BlockaidBanner = (bannerProps: BlockaidBannerProps) => {
...
+            ConfirmationTopSheetSelectorsIDs.SECURITY_ALERT_RESPONSE_FAILED_BANNER
+          }
         />
       </View>
     );
   }
 
   if (!REASON_DESCRIPTION_I18N_KEY_MAP[reason]) {
-    captureException(`BlockaidBannerAlert: Unidentified reason '${reason}'`);
+    const unidentifiedReasonError = new Error(
```
Cal-L, line 136: We use types to ensure that we're passing Errors into captureException, which provides better stack tracing

app/util/trace.test.ts
```diff
@@ -19,8 +18,10 @@ jest.mock('@sentry/react-native', () => ({
 const NAME_MOCK = TraceName.Middleware;
 const ID_MOCK = 'testId';
 const PARENT_CONTEXT_MOCK = {
-  spanId: 'parentSpanId',
-} as Span;
+  spanContext: () => ({
```
Cal-L, line 21: spanId is now accessed via spanContext()

app/util/trace.ts
```diff
@@ -285,7 +285,7 @@ function startSpan<T>(
     op: op || OP_DEFAULT,
     // This needs to be parentSpan once we have the withIsolatedScope implementation in place in the Sentry SDK for React Native
     // Reference PR that updates @sentry/react-native: https://github.com/getsentry/sentry-react-native/pull/3895
-    parentSpanId: parentSpan?.spanId,
+    parentSpan,
```
Cal-L, line 288: Takes a whole span now

app/util/trace.ts
```diff
@@ -373,11 +373,3 @@ function tryCatchMaybePromise<T>(
 
   return undefined;
 }
-
-function sentrySetMeasurement(
```
Cal-L, line 377: Don't need util. Can just call setMeasurement directly

patches/@sentry+core+8.54.0.patch
```diff
@@ -0,0 +1,16 @@
...
++++ b/node_modules/@sentry/core/build/types/exports.d.ts
+@@ -6,8 +6,10 @@ import type { ExclusiveEventHintOrCaptureContext } from './utils/prepareEvent';
+  * @param exception The exception to capture.
+  * @param hint Optional additional data to attach to the Sentry event.
+  * @returns the id of the captured Sentry event.
++ * 
++ * Patch to enforce better stack tracing using Error - https://docs.sentry.io/platforms/javascript/usage/#capturing-errors
+  */
+-export declare function captureException(exception: unknown, hint?: ExclusiveEventHintOrCaptureContext): string;
++export declare function captureException(exception: Error, hint?: ExclusiveEventHintOrCaptureContext): string;
```
Cal-L, line 13: Ensure better stack trace with Error

Cal-L:
@SocketSecurity ignore-all

OGPoyraz (dismissed):
Confirmation changes LGTM

tommasini (dismissed):

patches/@sentry+core+8.54.0.patch
Gudahtt: Agreed that we should be passing errors to `captureException`, but why do we need this patch?

If we really want this type checked more strictly, we can define our own `captureException` that calls Sentry's. Typically we avoid patches except as a last resort, as they make updates significantly more difficult.

package.json
```diff
@@ -254,7 +254,7 @@
     "@reown/walletkit": "^1.2.3",
     "@segment/sovran-react-native": "^1.0.4",
     "@sentry/integrations": "6.3.1",
-    "@sentry/react-native": "^5.33.0",
+    "@sentry/react-native": "~6.10.0",
```
Gudahtt, line 257: Why limit the range just to this minor version? Is there a breaking change in version 6.11?

Gudahtt, line 257: This breaking change made in v6 seems to impact us: https://docs.sentry.io/platforms/react-native/migration/v5-to-v6/#updated-behavior-of-tracepropagationtargets-in-react-native

Effectively it means tracing headers are now added on _all_ network requests made outside of the webview, whereas previously they were not. We may want to explicitly set `tracePropagationTargets` to the v5 default value to avoid any unintended changes here

package.json
```diff
@@ -427,7 +427,7 @@
     "@open-rpc/test-coverage": "^2.2.2",
     "@react-native/metro-config": "0.76.9",
     "@rpii/wdio-html-reporter": "^7.7.1",
-    "@sentry/types": "^7.117.0",
+    "@sentry/core": "8.54.0",
```
Gudahtt, line 430: The update to v8 includes some breaking changes that haven't been dealt with yet. I'm surprised it isn't throwing an error 🤔 

See here: https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/#removal-of-class-based-integrations
We're still using two class-based integrations

app/util/trace.ts
Gudahtt: Do you know where this change is documented? I'm having a bit of trouble following.

I see something about Span mentioned here: https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/#removal-of-span-class-export-from-sdk-packages

Gudahtt (dismissed):
Seems like some breaking changes are not yet handled

package.json
Cal-L: This specific version was auto updated using Expo CLI via `npx expo install --fix`. It's part of Expo's workflow for version compatibility. My guess is that they have a manifest somewhere that keeps track of this. This was just their suggested version but I don't see why we need to limit to the minor version as well. I can update this in a bit

package.json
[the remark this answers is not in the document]
Gudahtt: I see that `@sentry/react-native` pins to a specific version, and that they recommend pinning the version of core used if we need to import it directly.

Unfortunate, but it seems this has to be the approach given how the library works.

Gudahtt: Looks like the integrations aren't throwing an error because we're importing them from the integrations package, not from core. And the integrations package was not updated in this PR.

Still seems advisable to update it though, at least according to the migration doc. It's unclear whether there are any consequences to using outdated integrations.

app/util/trace.ts
Gudahtt: I am having trouble finding the `Span` type being imported here. Looks like it was removed, and isn't intended for external use in general.

Gudahtt: Oh I see, the removed the `Span` class from exports, but they are still exporting the type.

I found a dedicated page about the trace API changes. There is a section about how to set a parent span: https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/v8-new-performance-api/#creating-a-child-span-of-a-specific-span

They suggest a different method (`Sentry.withActiveSpan`). It's unclear how this differs from setting the `parentSpan` option 🤔.

Gudahtt: Not sure why this option wasn't referenced in the migration docs, but I found it elsewhere in their docs here: https://docs.sentry.io/platforms/javascript/tracing/instrumentation/#starting-spans-as-children-of-a-specific-span

This suggests what we're doing here is correct, no need to use `withActiveSpan`

app/util/trace.ts
```diff
@@ -285,7 +285,7 @@ function startSpan<T>(
     op: op || OP_DEFAULT,
     // This needs to be parentSpan once we have the withIsolatedScope implementation in place in the Sentry SDK for React Native
     // Reference PR that updates @sentry/react-native: https://github.com/getsentry/sentry-react-native/pull/3895
```
Gudahtt, line 287: This comment is now obsolete
```suggestion
```

Cal-L:
Skipping sonar cloud since function was not changed on files complaining about coverage

app/util/trace.ts
Cal-L: Removed

package.json
Cal-L: It is a bit confusing since the RN SDK doesn't indicate that anything is broken but the browser docs implies that it should be updated. Will check in with the Sentry team about this today. In any case, updated per your suggestion here - https://github.com/MetaMask/metamask-mobile/pull/15045/commits/5d09c13425a0f6270b60d6904c18a554272c53b1

package.json
Cal-L: Updated to use the v5 defaults - https://github.com/MetaMask/metamask-mobile/pull/15045/commits/5d09c13425a0f6270b60d6904c18a554272c53b1

patches/@sentry+core+8.54.0.patch
Cal-L: I experimented with your suggestion but since we're using captureException is so many places (~175), I didn't want to have to replace all instances and bloat the PR. The approach that I took instead was to redeclare the @sentry/react-native module and replace `captureException` type instead. LMK if you know of a simpler solution that doesn't require a large refactor.

Gudahtt: To clarify, mainly I was hoping to gain insight into the problem this was solving, as it wasn't clear to me.

I understand preferring Error here, but what changed in this update related to this? We've always preferred error for exceptions

tommasini (approved):

OGPoyraz (approved):

patches/@sentry+core+8.54.0.patch
Cal-L: For @sentry/core v7, which we were on, the exception type was `any` - https://github.com/getsentry/sentry-javascript/blob/7.111.0/packages/core/src/exports.ts#L43. In v8, which we've upgraded to, that type changed to `unknown` - https://github.com/getsentry/sentry-javascript/blob/8.54.0/packages/core/src/exports.ts#L37

owencraston (approved):

PR #15048, opened 2025-05-01
perf: patch tokenBalances state update modifications

sahar-fehri:
## **Description**

Patches [changes](https://github.com/MetaMask/core/pull/5726) on tokenBalancesController state update.

PR does not introduce any new functionalities.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Import new accounts
2. You should be able to see tokens and balances correctly
3. Import new tokens and you should be able to see balances correctly
4. Remove an account
5. Check state in tokensController and tokenBalancesController, you should not see data for the deleted account


Prithpal-Sooriya:
Tested the performance of this PR. It definitely cuts away some rendering commits and performance issues. Below are profile screenshots from switching to PV (current network to popular networks), with 30 accounts and 1 imported account with many tokens.

<details><summary>IOS Simulator</summary>

| Before | After |
|--------|--------|
| 44 <0.1ms commits taking 2.5s | 35 <0.1ms commits taking 1.9s |
| Screenshot 2025-05-01 at 10 57 54 | Screenshot 2025-05-01 at 11 01 25 | 

</details> 

<details><summary>Low end Android Physical Device</summary>

I want to emphasise this was a development build (which has logging, tracing, etc) which contributed to the much slower numbers.

| Before | After |
|--------|--------|
| 37 <0.1ms commits taking 27.5s | 36 <0.1ms commits taking 19.2s |
| 
Screenshot 2025-05-01 at 11 12 41 | Screenshot 2025-05-01 at 11 06 23 | 

</details>

Prithpal-Sooriya (dismissed):

gambinish (dismissed):

app/store/migrations/079.ts
```diff
@@ -0,0 +1,158 @@
...
+  [chainId: Hex]: { [key: string]: Token[] };
+}
+
+/**
+ * Migration 79: removes from the TokensController state all tokens that belong to an EVM account that has been removed.
+ * Also removes from TokenBalancesController all balances that belong to an EVM account that has been removed.
+ */
+
+const migration = (state: unknown): unknown => {
+  if (!ensureValidState(state, 78)) {
```
Prithpal-Sooriya, line 24: Is `78` right, I thought it should point to the existing version?

app/store/migrations/079.ts
```diff
@@ -0,0 +1,158 @@
...
+    !isObject(tokenBalancesControllerState) ||
+    !hasProperty(tokenBalancesControllerState, 'tokenBalances') ||
+    !isObject(tokenBalancesControllerState.tokenBalances)
+  ) {
+    captureException(
+      new Error(
+        `FATAL ERROR: Migration 79: Invalid TokenBalancesController state error: '${typeof tokenBalancesControllerState}'`,
+      ),
+    );
+    return state;
```
Prithpal-Sooriya, line 46: IDK if this is a fatal error since we do return state?

Prithpal-Sooriya: Same for the ones below, I don't think they are fatal.

tommasini: We need to always return the state anyway, if not, we will wipe out all the data persisted (if we return null or undefined)

app/store/migrations/079.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  if (
+    !isObject(tokensControllerState) ||
+    !hasProperty(tokensControllerState, 'allTokens') ||
```
tommasini, line 51: If this properties are all indeed defined in a default state, and there is no way of having those undefined when we reach this migration in an update, we should split this in different conditions and accomodate a proper error message to each one, so if we have this error log, we can act accordingly, this way we will not know specifically what the issue is, wdyt?

app/store/migrations/079.ts
```diff
@@ -0,0 +1,158 @@
...
+      ] as Record<`0x${string}`, Token[]>;
+      for (const account of Object.keys(tokensByAccounts)) {
+        if (!evmAccountAddresses.includes(account)) {
+          delete (tokensControllerState.allTokens as AllTokens)[chainId as Hex][
+            account
+          ];
+        }
+      }
+    }
+  }
```
Prithpal-Sooriya, line 105: Since we using both the key and the value, could we use Object.entries?
```
  if (hasProperty(tokensControllerState, 'allTokens')) {
    Object.entries(tokensControllerState.allTokens).forEach(([chainId, accountsTokens]) => {
      Object.keys(accountsTokens).forEach(account => {
        if (!evmAccountAddresses.includes(account)) {
          delete tokensControllerState.allTokens[chainId as Hex][account];
        }
      });
    });
  }
```

Prithpal-Sooriya: Same for below with `allDetectedTokens`

patches/@metamask+assets-controllers+62.0.0.patch
```diff
@@ -161,3 +161,566 @@ index 4384784..e35e71c 100644
      collection?: Collection;
      address?: string;
      attributes?: Attributes[];
+diff --git a/node_modules/@metamask/assets-controllers/dist/TokenBalancesController.cjs b/node_modules/@metamask/assets-controllers/dist/TokenBalancesController.cjs
+index a70d3f4..05ed007 100644
+--- a/node_modules/@metamask/assets-controllers/dist/TokenBalancesController.cjs
++++ b/node_modules/@metamask/assets-controllers/dist/TokenBalancesController.cjs
```
Prithpal-Sooriya, line 167: Nice patch!

Small nit - can we add a code comment to point to the CORE change this patch is for? It makes it a little easier for devs to clean up this patch code once the CORE changes are merged 😄 

```diff
+ // Patch can be removed once https://github.com/MetaMask/core/pull/5726 is released
```

app/store/migrations/079.ts
Prithpal-Sooriya: Agreed with this, should the error we send to sentry logged with "FATAL ERROR"? Especially since this won't be wiping out the state.

app/store/migrations/079.ts
Prithpal-Sooriya: Ooops, it looks like we might need to update the prev patch too 😅 
https://github.com/MetaMask/metamask-mobile/blob/c62b04237d074a9511d2e25c6db7d3d322ce0e0f/app/store/migrations/078.ts#L15

patches/@metamask+assets-controllers+62.0.0.patch
sahar-fehri: We will definitely remove this patch once assets is released; this is just for devs who review and test this; do you still think we should add the link?

app/store/migrations/079.ts
sahar-fehri: allTokens, allDetectedTokens and allIgnoredTokens are all defined in the default state and they all cannot be undefined which is basically what the whole block is checking
`  if (
    !isObject(tokensControllerState) ||
    !hasProperty(tokensControllerState, 'allTokens') ||
    !isObject(tokensControllerState.allTokens) ||
    !hasProperty(tokensControllerState, 'allDetectedTokens') ||
    !isObject(tokensControllerState.allDetectedTokens) ||
    !hasProperty(tokensControllerState, 'allIgnoredTokens') ||
    !isObject(tokensControllerState.allIgnoredTokens)
  )`
so if we get the error it means one of these values is not correct; 

Were you suggesting to have a separate block for each and a different error msg for each?
I grouped them all since its all about tokensController default state

app/store/migrations/079.ts
sahar-fehri: [340091f](https://github.com/MetaMask/metamask-mobile/pull/15048/commits/340091f6c79a89580e24d29219132421af946a64)

patches/@metamask+assets-controllers+62.0.0.patch
Prithpal-Sooriya: Happy to leave it otherwise. We now have a GH comment, so devs who want to search for this through git-blame can find this PR/comment.

Prithpal-Sooriya (dismissed):

Prithpal-Sooriya (approved):

tommasini (approved):

PR #15051, opened 2025-05-01
fix: Auto reject terminated Snap dialogs

FrederikBolding:
## **Description**

If a Snap crashes we want to reject it's dialog so that it closes since the dialog will be unresponsive anyway. This mirrors the implementation in the extension: https://github.com/MetaMask/metamask-extension/blob/7a22dbc2ea9d47efd4901c7326ccf3ab4218b1c3/app/scripts/metamask-controller.js#L3113-L3130


Mrtenz (approved):

Daniel-Cross (approved):
LGTM

FrederikBolding:
Skipping E2E since I am high confidence that the tests failing are due to flake.

PR #15060, opened 2025-05-01
fix: improve tokenListItem rerenders

sahar-fehri:
## **Description**

`<TokenListItem>` has been using
```
    const nonEvmAsset = useSelector((state: RootState) =>
      selectNonEvmAssetById(state, {
        accountId: selectedAccount?.id,
        assetId: assetKey.address,
      }),
    );
```
To get non EVM asset metadata; even though the selector is a deepEqualSelector, each instance of the `<TokenListItem>` were passing a different assetId to this selector which breaks reselect memoization and the selector had to recompute each time. 
Instead we are using a create selector factory pattern, where each instance of `<TokenListItem>` makes their own selector to then be used to get the non-evm asset metadata.

Resources:
[Reselect Docs](https://reselect.js.org/introduction/how-does-reselect-work#memoization-with-reselect)

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/components/UI/Tokens/TokenList/TokenListItem/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useCallback, useMemo } from 'react';
 import { View } from 'react-native';
 import {
+  CaipAssetId,
```
Prithpal-Sooriya, line 4: Should we move this under the codefence?

It is a type, so it should be fine.

app/selectors/multichain/multichain.ts
```diff
@@ -513,4 +514,73 @@ export const selectNonEvmAssetById = createDeepEqualSelector(
   },
 );
 
+export const makeSelectNonEvmAssetById = () =>
```
Prithpal-Sooriya, line 517: We can delete the above selector for the fix in `main`, I don't think it is used anymore.

sahar-fehri: Removed here [e979ded](https://github.com/MetaMask/metamask-mobile/pull/15060/commits/e979ded130bcf0c37d0cff10147940305c8c5339)

sahar-fehri:
Will use this for the CP instead https://github.com/MetaMask/metamask-mobile/pull/15063

Issue #13117, opened 2025-01-22
[feat]: Move AnimatedPulse into components-temp for reusability


digiwand:
### What is this about?

Following https://github.com/MetaMask/metamask-mobile/pull/13087, we will move AnimatedPulse to components-temp and mention to the DS in the effort to align with design. We will also create a storybook page for this.

Path to be updated in `Confirm/Info/TypedSignV3V4/Simulation/components/ValueDisplay/ValueDisplay.tsx`: 
```
import AnimatedPulse from '../../../../../../../../../../component-library/components-temp/AnimatedPulse/AnimatedPulse
```

### Stakeholder review needed before the work gets merged

- [ ] Engineering (needed in most cases)
- [ ] Design
- [ ] Product
- [ ] QA (automation tests are required to pass before merging PRs but not all changes are covered by automation tests - please review if QA is needed beyond automation tests)
- [ ] Security
- [ ] Legal
- [ ] Marketing
- [ ] Management (please specify)
- [ ] Other (please specify)

OGPoyraz:
This will be resolved in https://github.com/MetaMask/metamask-mobile/pull/15074

PR #13087, opened 2025-01-21
fix: Flickering permit and decoding simulation values

digiwand:
## **Description**

Permit Simulation displays value without decimals applied before displaying value with decimals applied.

This PR fixes the issue but introducing an isPending state to display a loader while the token details are being fetched 

Created a new UI to support the loading behavior called AnimatedPulse. This was approved by @SayaGT to be used here. 
Created a follow-up issue to move AnimatedPulse to components-temp later https://github.com/MetaMask/metamask-mobile/issues/13117

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/12993

## **Manual testing steps**

1. Request a redesign Permit signature
2. Observe loader and value

### **Before**

https://github.com/user-attachments/assets/dc7c5015-811c-4f53-a10b-eb698644c247

### **After**

https://github.com/user-attachments/assets/e0870c75-b9d6-47c0-96fc-615b23cd62f8

#### Not being used, but if we used react-native Loader component. Instead of this, we are utilizing AnimatedPulse above
https://github.com/user-attachments/assets/fff5d869-19ce-465b-9414-cbb35331763a

## **Pre-merge author checklist**

- [ ] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #15074, opened 2025-05-02
feat: Add max value refresher for `transfer` redesigned confirmations

OGPoyraz:
## **Description**

This PR aims to add max value refresher hook into `transfer` confirmations.

Essentially hook is updating transaction value via `updateEditableParams`. After design discussions - every update should pulse the `TokenHero` component. Hence this PR also moves `animated-pulse` component in to UI components and re-use that.

Regarding changes in `animated-pulse`: it appears if `isPulsing` changing too fast it stops animation. Hence there are modifications in `animated-pulse` component but it's basically making sure pulse animation at least done twice.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4797

## **Manual testing steps**

Only possible to test it out via setting `FEATURE_FLAG_REDESIGNED_TRANSFER` to `true`

### **After**

https://github.com/user-attachments/assets/581df848-b6c9-40b6-bb12-aaf673dc4dfd


app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
OGPoyraz, line 1: Not going to add any test for this file since it just does the animation - we may ask for some reusable component from DS as described here https://github.com/MetaMask/metamask-mobile/issues/13117

app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }, [isPulsing, opacity, runSinglePulseCycle]);
+
+  return (
+    <Animated.View style={{ opacity }} {...props}>
+      {children}
+    </Animated.View>
+  );
+};
+
+export default AnimatedPulse;
```
digiwand, line 125: [comment] I think it would have been helpful to separate the filepath change from the logical changes. It is hard to compare the logic changes when they've been combined into a single comment

app/components/Views/confirmations/hooks/useMaxValueRefresher.ts
```diff
@@ -0,0 +1,67 @@
...
+      // Not compatible with transaction value refresh logic
+      return;
+    }
+
+    const balance = new BigNumber(balanceWeiInHex);
+    const fee = new BigNumber(preciseNativeFeeInHex);
+    const maxValue = balance.minus(fee);
+    const maxValueHex = add0x(maxValue.toString(16));
+
+    if (maxValueHex !== txParams.value) {
```
digiwand, line 43: [optional]
```suggestion
    const shouldUpdate = maxValueHex !== txParams.value;

    if (shouldUpdate) {
```

app/components/Views/confirmations/hooks/useMaxValueRefresher.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    if (!maxValueMode || type !== TransactionType.simpleSend) {
+      // Not compatible with transaction value refresh logic
+      return;
+    }
+
+    const balance = new BigNumber(balanceWeiInHex);
+    const fee = new BigNumber(preciseNativeFeeInHex);
+    const maxValue = balance.minus(fee);
```
digiwand, line 40: Q: is it possible for the fee to be greater than the balance -> negative maxValue

digiwand (changes requested):

app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
OGPoyraz: Noted

app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
```diff
@@ -0,0 +1,125 @@
...
+}: AnimatedPulseProps) => {
+  const opacity = useRef(new Animated.Value(0.3)).current;
+  const currentAnimationRef = useRef<Animated.CompositeAnimation | null>(null);
+  const isCurrentlyPulsingRef = useRef(isPulsing);
+  const cyclesNeededRef = useRef(0);
+  const cyclesCompletedRef = useRef(0);
+
+  const durationFadeIn = 750;
+  const durationFadeOut = 750;
+  const durationFadeInFinal = 300;
```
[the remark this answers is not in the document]
OGPoyraz: Done

app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
```diff
@@ -0,0 +1,125 @@
...
+    };
+  }, [isPulsing, opacity, runSinglePulseCycle]);
+
+  return (
+    <Animated.View style={{ opacity }} {...props}>
+      {children}
+    </Animated.View>
+  );
+};
+
```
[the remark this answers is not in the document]
OGPoyraz: So for `TokenHero` the animation was not existed before - but after recording is already in the description. For other use-cases changes are not affecting component will work as before.

app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
```diff
@@ -0,0 +1,125 @@
...
+      }).start();
+    }
+
+    // Cleanup
+    return () => {
+      if (currentAnimationRef.current) {
+        currentAnimationRef.current.stop();
+        currentAnimationRef.current = null;
+      }
+    };
```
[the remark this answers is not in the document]
OGPoyraz: Done

app/components/Views/confirmations/hooks/useMaxValueRefresher.ts
OGPoyraz: Done

app/components/Views/confirmations/hooks/useMaxValueRefresher.ts
OGPoyraz: Good call - added a positive check in order to do update or not

pedronfigueiredo (dismissed):

app/components/Views/confirmations/components/UI/animated-pulse/animated-pulse.tsx
```diff
@@ -0,0 +1,125 @@
...
+interface AnimatedPulseProps extends ViewProps {
+  children: React.ReactNode;
+  isPulsing?: boolean;
+  minCycles?: number;
+}
+
+const AnimatedPulse = ({
+  children,
+  isPulsing = true,
+  minCycles = 2, // Default to 2 cycles minimum for a better visual effect
```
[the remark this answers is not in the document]
digiwand: I added AnimatedPulse here a while back 
https://github.com/MetaMask/metamask-mobile/pull/13087 

The after screen recording shows the pulse on an empty placeholder element. When the data loads, the pulsing stops. 

This is why I suggested we default to 0 pulsing. I see now that the pulsing here is used for polling data. Thanks for the recordings! I'm on board with keeping the default minPulse set to 2 now. Following this, I think we should set minPulse to 0 for previous instances used. 

re: new preventPulse
I think rather than adding this, we can utilize the isPulsing param

e.g.
```
      isPulsing={isTransactionValueUpdating}
      preventPulse={!maxValueMode}
```
→
```
      isPulsing={maxValueMode && isTransactionValueUpdating}
```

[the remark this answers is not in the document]
OGPoyraz: It make sense - settled minCycle as `0` for the signature usage, done here https://github.com/MetaMask/metamask-mobile/pull/15074/commits/a244d6dfb04d0dc678866b48f6d9976bfacea098

I still think its beneficial to have separate `preventPulse` property to make it explicitly render within `View` instead of `AnimatedView`. 

Imagine if we try to derive that value from `isPulsing`, it would work for prevented examples but it could lead potential issues for animated ones - `isPulsing` may get a value change while pulsing and it could interrupt animation if we depend on that value for container.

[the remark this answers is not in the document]
digiwand: that's a good point avoiding the unnecessary AnimatedView 👍🏼

digiwand (approved):

PR #15063, opened 2025-05-01
fix: cp-7.46.0 improve token list item re renders

sahar-fehri:
## **Description**

`<TokenListItem>` has been using
```
    const nonEvmAsset = useSelector((state: RootState) =>
      selectNonEvmAssetById(state, {
        accountId: selectedAccount?.id,
        assetId: assetKey.address,
      }),
    );
```
To get non EVM asset metadata; even though the selector is a deepEqualSelector, each instance of the `<TokenListItem>` were passing a different assetId to this selector which breaks reselect memoization and the selector had to recompute each time. 
Instead we are using a create selector factory pattern, where each instance of `<TokenListItem>` makes their own selector to then be used to get the non-evm asset metadata.

Resources:
[Reselect Docs](https://reselect.js.org/introduction/how-does-reselect-work#memoization-with-reselect)

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Pre-merge author checklist**

- [ ] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #15075, opened 2025-05-02
perf: Fix a couple of selectors used for tokens

FrederikBolding:
## **Description**

Updates a couple of selectors to prevent re-rendering of `TokenListItem`.


app/components/UI/Tokens/TokenList/TokenListItem/index.tsx
```diff
@@ -119,6 +120,7 @@ export const TokenListItem = React.memo(
         accountId: selectedAccount?.id,
         assetId: assetKey.address,
       }),
+      isEqual,
```
FrederikBolding, line 123: It's not clear to me exactly why but `selectNonEvmAssetById` returns a new reference every time and as such the entire token component re-renders.

app/components/UI/Earn/hooks/useEarnTokens.ts
```diff
@@ -14,9 +14,10 @@ import {
 
 // Filters user's tokens to only return the supported and enabled earn tokens.
 const useEarnTokens = () => {
-  const tokens = useSelector((state: RootState) =>
-    isPortfolioViewEnabled() ? selectAccountTokensAcrossChains(state) : {},
```
FrederikBolding, line 18: Returning an empty object in the selector causes re-renders since it is a new reference every time.

app/components/UI/Tokens/TokenList/TokenListItem/index.tsx
Prithpal-Sooriya: Yes we have a similar fix:
https://github.com/MetaMask/metamask-mobile/pull/15063

app/components/UI/Earn/hooks/useEarnTokens.ts
sahar-fehri: I noticed this too; removing the check for `isPortfolioViewEnabled` and returning only `selectAccountTokensAcrossChains` would fix the infinite re-renders for earn logic

FrederikBolding: I just pushed a commit that simply moves this check into the memoized function, should have the same behavior but fix re-renders.

sahar-fehri: Should we push the other fix https://github.com/MetaMask/metamask-mobile/pull/15063 first then we merge this one?

FrederikBolding: Is my change to `TokenListItem` needed after that?

sahar-fehri: I dnt think so; just the earn logic change would be needed 🙏

sahar-fehri (approved):

amitabh94 (approved):

PR #15244, opened 2025-05-08
bug: clear MMKV on fresh install

### LINKED ISSUE #9605: [Sentry] Error: Vault Corruption Error
Sentry Issue: https://metamask.sentry.io/issues/4614917373/?environment=production&project=2299799&query=is%3Aunresolved%20vault&referrer=issue-stream&sort=date&stream_index=0

```
Error: Vault Corruption Error
  at Login#render (app/components/Views/Login/index.js:622:20)
  at <global> (app/component-library/components/Texts/Text/Text.types.ts:25:22)
  at flushedQueue (app:///index.android.bundle:183:11)
...
(20 additional frame(s) were not displayed)
```

### LINKED ISSUE #13903: Feature to better handle phone migrations
### What is this about?

When a user migrates to a new phone the MM app will automatically install and prompt the user to login. This is misleading because no app data is transferred to the new phone (for security reasons). We should handle this case by presenting the user with the onboarding flow instead of the login screen.

The current resolution for this is to have the user to `Reset Wallet` or uninstall/reinstall.

Below are a list of considerations we should review when starting this ticket:

---

#### 🚫 Disable iCloud backup for app data (this is already done):
Use the `NSURLIsExcludedFromBackupKey` flag to **exclude files from iCloud backup**.

```swift
var url = URL(fileURLWithPath: filePath)
do {
    var resourceValues = URLResourceValues()
    resourceValues.isExcludedFromBackup = true
    try url.setResourceValues(resourceValues)
} catch {
    print("Failed to exclude \(url.lastPathComponent) from backup")
}
```

#### 🧼 Store minimal persistent data in `UserDefaults`, Keychain, or documents — and **clear everything** on reinstall or reinstall+restore if necessary.

- Be cautious with **Keychain** — it's persisted even after uninstall.
- Consider storing a flag in a **non-persisted location** (e.g., temporary directory) that indicates first install.

---

#### 🧼 Key considerations:
- Like iOS, **EncryptedSharedPreferences**, **Room**, or **Jetpack DataStore** will persist across reinstalls if backups are enabled.
- With backups disabled, everything starts fresh unless you're syncing from your own backend.

---

### 2. **Detect a Restored State (Optional Advanced Pattern)**
If disabling cloud backup isn’t sufficient (e.g., you suspect a system restore is restoring data anyway), you could:

- Store a **device identifier or first install token** in local storage or a secure backend.
- When a user logs in, check if the device is the same one — if not, prompt them to re-authenticate and wipe local data.

---

### 3. **Avoid Keychain (iOS) long-lived sessions**
These survive uninstalls, so if you use them, your app may “magically” restore sessions on a new device. You may need to explicitly clear them or gate access with device binding.

---

## 🧪 Test It:
1. Install your app on one device.
2. Log in, set some local state.
3. Backup or don't
  a. Option 1: Backup that device (iCloud or Google Drive).
  b. Option 2: Do not back up. 
6. Restore to a new phone.
7. Download app again. Check:
- Is the user still logged in?
- Are previous settings still there?
- If yes, identify where they were saved (Keychain? Cloud backup? Secure preferences?).

---

### Scenario

_No response_

### Design

_No response_

### Technical Details

_No response_

### Threat Modeling Framework

_No response_

### Acceptance Criteria

- When upgrading to a new phone users are not prompted to login

- [x] Engineering (needed in most cases)
- [ ] Design
- [x] Product
- [ ] Security
- [ ] Legal
- [ ] Marketing
- [ ] Other (please specify)

### References

_No response_

MarioAslau:
## **Description**

When digging migration issue [more](https://github.com/MetaMask/metamask-mobile/issues/13903).  we've come to realise that MMKV causing the issue with the EXISTING_USER flag.

- Right now, when the app starts, in App.tsx,  in the second useEffect the app checks for existing user: const existingUser = await StorageWrapper.getItem(EXISTING_USER);
- If existing user exists, it calls Authentication.appTriggeredAuth(), inside it, it tries to get credentials form the keychain. If there are none, it will thrown an error. Which will be our case, on a new device.
- Error is caught and lockApp is called. This method navigates to the login screen.
- The MMKV Storage gets restored via iCloud Backup-> app sees EXISTING_USER and tries to auth -> Auth fails due to lack of actual vault/credentials -> app shows login screen instead of onboarding.

**Solution:**

 Creating a new migration that will move this flag from MMKV to FileSystem storage, enabling it to not be backed up anymore.
 
 Solution Progress Notion Page: https://www.notion.so/metamask-consensys/Feature-to-better-handle-phone-migrations-1d5f86d67d688075b818fd2c8a291b14?source=copy_link
 
 PR Review Recording: https://consensys.zoom.us/rec/share/JKX_WCkihB9LyXo8USn1dtELH8PGFth31vJug16-gAePqqvHLb4c7prNfBxvWIqP.8UHtS5VF3yr-5kei
Passcode: LE8m.K*C

**Additonally: Optin Metrics Changes**

After moving EXISTING_USER flag from MMKV to Redux, it created a bug with the OptinMetrics. Every time a user would unlock the app, the OptinMetrics would pop up again and again.

I idetified two issues:
* Dependency array issue: Including isMetaMetricsUISeen in the dependency array caused unnecessary re-runs of the authentication flow (in App.tsx)
* Flag setting issue: The isMetaMetricsUISeen flag wasn't being set in all completion paths, causing the screen to reappear

Fix:
* In the auth flow, isMetaMetricsUISeen is part of the dependancy array. Before when it was part of MMKV, components do not automatically re-render when values change. Even though it was part of the array, the use effect would not run again unless the component re-renders. Now due to it being in redux it creates a circular dependancy. I removed it from the dependancy array 
* In OptinMetrics/index.js moved this.props.setMetaMetricsUISeen(true); outside if. The isMetaMetricsUISeen flag must always be set to true when the user completes the OptinMetrics flow, regardless of how they complete it.

After these fixes:
* The OptinMetrics screen only appears once for new users
* The authentication flow doesn't restart when isMetaMetricsUISeen changes
* The isMetaMetricsUISeen flag is properly set in all completion scenarios
* this stops the repeated screens

## **Related issues**

Fixes:

## **Manual testing steps**

Physical Devices used were Samsung Glaxy S22 and iPhone 12 Pro Max

Testing Scenarios:

Bitrise build release_builds_to_store_pipeline on PR branch: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/99183957-20a4-4927-9904-ae539a5d696d

1. Upgrade Path
* Android: Tested from v7.44 (via Github releases), upgrading to PR branch (has latest main). Upgrade successful

https://github.com/user-attachments/assets/9b690d33-cb0f-4236-8c6c-c700b5cfd7c3

* iPhone: tested via Testflight from v7.44  to 7.48.99 (2040 - PR Branch via bitrise)

https://github.com/user-attachments/assets/b3f59440-ae14-4946-8e16-5b9ce69ccb83

2. Recover via iCloud
* iPhone: tested via Testflight from v7.44  to 7.48.99 (2040 - PR Branch via bitrise)

https://github.com/user-attachments/assets/16339179-90ba-4053-ad5f-f2bc0ea6aa68

https://github.com/user-attachments/assets/0885250e-2569-472c-ba2c-6f65e8116ef3

https://github.com/user-attachments/assets/0bf03bf6-4cb1-45d3-abaa-a9016d406d4d


sethkfman:
@MarioAslau have you tested the upgrade path to make sure we are not flushing the store?

tommasini (dismissed):

app/components/Nav/App/App.tsx
```diff
@@ -922,6 +922,15 @@ const App: React.FC = () => {
 
   useEffect(() => {
     async function startApp() {
+      const firstInstallTime = await getFirstInstallTime();
+      const lastUpdateTime = await getLastUpdateTime();
+      const isFreshInstall = firstInstallTime === lastUpdateTime;
+      
+      if (isFreshInstall) {
+        // Clear all MMKV storage on fresh install to avoid leftovers of EXISTING_USER in case of phone migration
+        await StorageWrapper.clearAll();
```
Cal-L, line 931: Let's verify that this clears all keys. Ex. default as well as the unique keys used to create MMKV instances

app/components/Nav/App/App.tsx
```diff
@@ -922,6 +922,15 @@ const App: React.FC = () => {
 
   useEffect(() => {
     async function startApp() {
+      const firstInstallTime = await getFirstInstallTime();
```
Cal-L, line 925: Let's move this logic into a function in the sagas file and then call it at the beginning of `startAppServices` saga since this isn't component dependent

MarioAslau:
> @MarioAslau have you tested the upgrade path to make sure we are not flushing the store?
Primary solution:

Leveraging react-native-device-info, which has the getFirstInstallTime method and lastUpdateTime to check if it is a fresh install in App, if so, reset the flag.

The flag FRESH_INSTALL_CHECK_DONE is psecifcally used for the fresh install check. This flag was implemented for two reasons:

getFirstInstallTime (returns the timestamp of when the app was first installed on the device
), getLastUpdateTime (returns the timestamp of when the app was last updated). Due to this, in the situation where the app is freshly migrated to a new phone and it is also on the latest version, the two values will be identical. Even thought the app might no longer be freshly installed, until a new version comes out to update it, during that time the two methods will give the same number. To avoid this, the FRESH_INSTALL_CHECK_DONE will be used.

FRESH_INSTALL_CHECK_DONE is stored in AsyncStorage not MMKV due to:

The handleFreshInstallWithRestoredData method is specifically designed to detect when a user has a fresh install but with restored MMKV data from a backup. This is a critical security check.
If this flag would be stored in MMKV, when a user restores their MMKV data from a backup, these flags would be restored too, which can lead to a circular problem. The freshInstallCheckDone flag would be restored as 'true'. This would prevent the fresh install check from running. The method would miss detecting that this is actually a fresh install with restored data. By keeping separate they are not part of the MMKV backup/restore process, providing a reliable way to detect true fresh installs



There are 3 conditions:

1. To determine if the app has been freshly installed, I use getFirstInstallTime() and getLastUpdateTime() to compare the install date with last update 
- On Android, it uses PackageManager.getPackageInfo().firstInstallTime and lastUpdateTime
- On iOS, it uses NSFileManager.attributesOfItemAtPath to get the creation and modification dates of the app bundle

getFirstInstallTime -> will show the timestamp of when the app was first installed
getFirstInstallTime -> will show the timestamp of when the app was last updated. In the case of a fresh install it will coincide

For a fresh install, both timestamps will be identical since the app bundle is created and modified at the same time.

2. Checking for EXISTING_USER - this is the flag that can carry over migration due to it being stored in MMKV

To make sure that this condition is not true indefinitely until the user updates the app, we have the 3rd check.

3. If both of the conditions above are met + checking for the flag in Async Storage FRESH_INSTALL_CHECK_DONE. If this flag is not present, then it means that we have  Fresh install with an app stored from a backup/migration.

app/components/Nav/App/App.tsx
MarioAslau: I moved it all to the Performance class

app/core/Performance/Performance.ts
```diff
@@ -15,6 +25,45 @@ async function setPerformanceValues(appStartTime: number) {
 
 class Performance {
   static appLaunchTime: number;
+
+  /**
+   * Checks if this is a fresh install with restored MMKV data from a backup.
+   * If so, clears the MMKV storage to ensure a clean state.
+   */
+  static handleFreshInstallWithRestoredData = () => {
```
sethkfman, line 33: It would be good to see tests for this function.

MarioAslau:
After testing the  solution on the upgrade Path on iOS, it would fail every time, opening a brand new wallet.

After investigating it appears that on iOS, the app uses Branch SDK to track installs and updates. The Branch SDK:

- Tracks first install date and current install date
- Has specific logic for handling updates vs fresh installs
- Uses the app's library directory creation date to determine install dates

The key difference is that on iOS, the app uses the library directory creation date to determine install dates, while on Android it uses a simpler comparison. This can lead to different behavior when:

- The app is updated / reinstalled / updated / installed fresh

The Root Issue: Different Behavior of `getFirstInstallTime()` and `getLastUpdateTime()`

The key problem is in how these two methods work differently across platforms:

### Android Behavior:

- `getFirstInstallTime()`: Returns the timestamp when the app was **first installed** on the device (persists across updates)
- `getLastUpdateTime()`: Returns the timestamp of the **most recent update** (changes with each app update)
- On a fresh install: Both values are identical
- On an app update: `getFirstInstallTime()` stays the same, `getLastUpdateTime()` updates

### iOS Behavior:

- `getFirstInstallTime()`: May not persist reliably across app updates, especially with certain iOS backup/restore scenarios
- `getLastUpdateTime()`: Can behave inconsistently on iOS, sometimes matching `getFirstInstallTime()` even on updates
- iOS app sandboxing and the way app updates are handled can cause these timestamps to be reset or modified

The main issue is that iOS handles app installation/update metadata differently than Android, making timestamp-based detection unreliable. 

A data-presence-based approach or version tracking would be more robust across both platforms.

MarioAslau:
After testing the data approach solution, the upgrade Path would consistently fail due to the previous version not having the Async Storage flag that separated an Upgrade scenario from a Restore/phone migration, which would end up clearing the wallet on upgrade. No way was found around this, which resulted in a new approach

Due to the EXISTING_USER flag being stored in MMKV being the one responsible for determining if a user will land on Log IN screen or onboarding, a new approach is being taken, where a new migration will move this flag from MMKV to FileSystem storage, enabling it to not be backed up anymore.

MarioAslau:
The new Solution was branched off here https://github.com/MetaMask/metamask-mobile/pull/15874

MarioAslau:
Merged solution branch: https://github.com/MetaMask/metamask-mobile/pull/15874 into this one

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -1496,7 +1496,7 @@
 					"\"$(SRCROOT)/MetaMask/System/Library/Frameworks\"",
 				);
 				LLVM_LTO = YES;
-				MARKETING_VERSION = 7.45.2;
+				MARKETING_VERSION = 7.47.0;
```
sethkfman: Remove all version updates from the PR.

ios/MetaMask/AppDelegate.m
```diff
@@ -29,7 +29,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   // You can add your custom initial props in the dictionary below.
   // They will be passed down to the ViewController used by React Native.
   self.initialProps = @{@"foxCode": foxCode};
-
+
```
sethkfman: Remove these minor line changes.

app/components/Nav/App/App.tsx
MarioAslau: Resolving this as a new solution was implemented

app/core/Performance/Performance.ts
MarioAslau: Resolving this as new solution was implemented

sethkfman:
@MarioAslau Can you resolve the CI issues, add video evidence?

Cal-L:
Kicking off e2e again

app/components/Nav/App/App.test.tsx
```diff
@@ -80,10 +80,18 @@ describe('App', () => {
         });
       });
     });
-    it('navigates to login when user exists and logs in', async () => {
+    it('navigates to home when user exists and logs in', async () => {
       jest.spyOn(StorageWrapper, 'getItem').mockResolvedValue(true);
       jest.spyOn(Authentication, 'appTriggeredAuth').mockResolvedValue();
-      renderScreen(App, { name: 'App' }, { state: initialState });
+      renderScreen(App, { name: 'App' }, {
```
tommasini: NIT: Could we solve this lint warnings?

tommasini:
Can we update the description? Doesn't seem that it is the solution applied is the same as described there

app/store/migrations/086.ts
```diff
@@ -0,0 +1,55 @@
...
+    [key: string]: unknown;
+  };
+}
+
+/**
+ * Migration 83: Move EXISTING_USER flag from MMKV to Redux state
+ * This unifies user state management and fixes iCloud backup inconsistencies
+ */
+const migration = async (state: unknown): Promise<unknown> => {
+  if (!ensureValidState(state, 85)) {
```
tommasini: migration is number 86 and there is no test file for it yet, can we create a unit test file?

app/store/migrations/086.ts
```diff
@@ -0,0 +1,55 @@
...
+    return state;
+  }
+
+  const newState = cloneDeep(state) as ValidStateWithUser;
```
tommasini: Interesting, just curious, why are you using cloneDeep here?

app/store/migrations/086.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  try {
+    // Get existing user value from MMKV
+    const existingUser = await StorageWrapper.getItem(EXISTING_USER);
+
+    // Ensure user state exists
+    if (!isObject(newState.user)) {
+      newState.user = {};
```
tommasini: User reducer have an[ initial state defined](https://github.com/MetaMask/metamask-mobile/blob/29568c66bd1c16104d10d5740127fbeb34fe615d/app/reducers/user/index.ts#L12), is not expected to be an empty object. I believe if the user state is empty it was cleared wrongly and we should fix the bug.

I believe this is a situation where we should capture the exception with sentry instead

app/store/migrations/086.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // Clear from MMKV
+    if (existingUser !== null) {
+      await StorageWrapper.removeItem(EXISTING_USER);
+    }
+  } catch (error) {
+    captureException(error as Error);
+    // If we can't migrate the data, default to false for safety
+    if (!isObject(newState.user)) {
+      newState.user = {};
```
tommasini: Same here I don't believe this is the proper way to handle missing the user state since there is a default value for it 
https://github.com/MetaMask/metamask-mobile/blob/29568c66bd1c16104d10d5740127fbeb34fe615d/app/reducers/user/index.ts#L12

app/store/storage-wrapper.ts
```diff
@@ -155,4 +155,7 @@ class StorageWrapper {
   }
 }
 
-export default StorageWrapper.getInstance();
+// Get the singleton instance
+const storageWrapperInstance = StorageWrapper.getInstance();
```
tommasini: I think that this change is not necessary and the end result is the same, let me know if I'm missing something

index.js
```diff
@@ -14,13 +14,16 @@ import * as Sentry from '@sentry/react-native'; // eslint-disable-line import/no
 import { setupSentry } from './app/util/sentry/utils';
 setupSentry();
 
-import { AppRegistry, LogBox, ErrorUtils } from 'react-native';
```
tommasini: The changes to this file seem unrelated but not harmful

android/app/build.gradle
```diff
@@ -187,8 +187,8 @@ android {
         applicationId "io.metamask"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionName "7.46.2"
```
tommasini: Should we revert this changes?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "metamask",
-  "version": "7.46.2",
+  "version": "7.48.99",
```
tommasini: Should we revert this as well?

ios/MetaMask/AppDelegate.m
tommasini: This doesn't seem resolved

android/app/build.gradle
MarioAslau: as mentioned in Slack, I will revert the version bumps once testing is done

app/store/migrations/086.ts
MarioAslau: @tommasini  The migration needs to create a deep copy of the entire Redux state object. Our state is deeply nested, a shallow copy would not suffice, copying only top level, leaving nested objects as references. Using cloneDeep is the established pattern across different migrations

index.js
MarioAslau: One solution involved this file, when removing it I also cleaned it up with the formatter and removed unused imported ErrorUtils

package.json
MarioAslau: once testing is finalised and I don't need TestFlight I will revert the version changes

android/app/build.gradle
```diff
@@ -187,8 +187,8 @@ android {
         applicationId "io.metamask"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionName "7.47.1"
-        versionCode 2004
```
Cal-L: revert version changes

app/components/Nav/App/App.tsx
```diff
@@ -850,7 +855,10 @@ const App: React.FC = () => {
         const locked =
           errorMessage === AUTHENTICATION_APP_TRIGGERED_AUTH_NO_CREDENTIALS;
 
-        await Authentication.lockApp({ reset: false, locked });
+        // Only call lockApp if there is an existing user to prevent unnecessary calls
+        if (existingUser) {
+          await Authentication.lockApp({ reset: false, locked });
```
Cal-L: Not sure if this scenario where `existingUser` is false is possible but if it is, might be cleaner to throw here and navigate to onboarding in the `catch` block where `appTriggeredAuth()` is called instead. That can be treated as the worst case scenario.

bitrise.yml
```diff
@@ -2515,16 +2515,16 @@ app:
       PROJECT_LOCATION_IOS: ios
     - opts:
         is_expand: false
-      VERSION_NAME: 7.47.1
+      VERSION_NAME: 7.48.99
     - opts:
         is_expand: false
-      VERSION_NUMBER: 2004
+      VERSION_NUMBER: 2066
```
Cal-L: Remove version changes

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -1311,7 +1311,7 @@
 					"${inherited}",
 				);
 				LLVM_LTO = YES;
-				MARKETING_VERSION = 7.47.1;
+				MARKETING_VERSION = 7.48.99;
```
Cal-L: Revert version changes

app/components/Views/Onboarding/index.js
```diff
@@ -546,13 +551,40 @@ class Onboarding extends PureComponent {
       </View>
     );
   }
+
+  /**
+   * Fallback check for existing user flag
+   * Checks MMKV storage if Redux state indicates no existing user
+   * This handles edge cases where migration might have failed
+   */
+  async checkIfExistingUser() {
```
Cal-L: Have we seen cases where migration 89 fails? It would be worth faking a failure on migration 88 and then see if 89 fails as well. Since migrations should move to the next even if it prematurely returns, I'd expect 89 to at least run once. The question is if the `user` object is still intact. If it is, then we can probably remove `checkIfExistingUser`

Cal-L (commented):
Left some comments

app/components/Views/Onboarding/index.js
MarioAslau: I can confirm migrations keep running from my tests.

app/components/Nav/App/App.tsx
Cal-L: Discussed with @MarioAslau to revert this since it shouldn't be possible to enter this scenario

Cal-L (commented):
Left last comments

app/components/Views/Onboarding/index.js
MarioAslau: Removed this method after discussing with @Cal-L, if the migrations fails the problem would be due to a previous migration

app/components/Nav/App/App.tsx
```diff
@@ -676,147 +678,149 @@ const AppFlow = () => {
...
-      />
-      <Stack.Screen
-        name={Routes.ADD_NETWORK}
-        component={AddNetworkFlow}
-        options={{ animationEnabled: true }}
-      />
-      {isNetworkUiRedesignEnabled() ? (
         <Stack.Screen
-          name={Routes.EDIT_NETWORK}
+          name={Routes.ADD_NETWORK}
```
tommasini, line 794: This doesn't seem related to this PR

app/components/Nav/App/App.tsx
```diff
@@ -997,15 +1005,17 @@ const App: React.FC = () => {
...
+            err as Error,
+            'Cannot initialize WalletConnect Manager.',
+          );
         },
       );
     }
   }, []);
 
   useEffect(() => {
     async function startApp() {
```
tommasini, line 1102: We can create a separate useCallback for this function and remove the react eslint rules disablement 

Let's address it on a follow up PR

app/store/migrations/89.test.ts
```diff
@@ -0,0 +1,311 @@
...
+jest.mock('../storage-wrapper', () => ({
+  getItem: jest.fn(),
+  removeItem: jest.fn(),
+}));
+
+const mockedCaptureException = jest.mocked(captureException);
+const mockedEnsureValidState = jest.mocked(ensureValidState);
+const mockedStorageWrapper = jest.mocked(StorageWrapper);
+
+describe('Migration 086', () => {
```
tommasini:
```suggestion
describe('Migration 089', () => {
```

app/store/persistConfig.ts
```diff
@@ -143,7 +143,9 @@ const persistConfig = {
     persistOnboardingTransform,
   ],
   stateReconciler: autoMergeLevel2, // see "Merge Process" section for details.
-  migrate: createMigrate(migrations, { debug: false }),
+  migrate: createMigrate(migrations, { 
+    debug: true, // Enable debug logging for migrations
```
tommasini:
```suggestion
    debug: false, 
```

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -202,24 +202,24 @@
...
 		650F2B9724DC5FEB00C3B9C4 /* RCTAesForked.xcodeproj */ = {isa = PBXFileReference; lastKnownFileType = "wrapper.pb-project"; name = RCTAesForked.xcodeproj; path = "../node_modules/react-native-aes-crypto-forked/ios/RCTAesForked.xcodeproj"; sourceTree = "<group>"; };
 		654378AE243E2ADB00571B9C /* MetaMask-Bridging-Header.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "MetaMask-Bridging-Header.h"; sourceTree = "<group>"; };
 		654378AF243E2ADC00571B9C /* File.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = File.swift; sourceTree = "<group>"; };
 		683865D794CE6007E46CAD3A /* ExpoModulesProvider.swift */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.swift; name = ExpoModulesProvider.swift; path = "Pods/Target Support Files/Pods-MetaMask/ExpoModulesProvider.swift"; sourceTree = "<group>"; };
-		6E7939D848B5467AA6602966 /* Geist Medium Italic.otf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = undefined; includeInIndex = 0; lastKnownFileType = unknown; name = "Geist Medium Italic.otf"; path = "../app/fonts/Geist Medium Italic.otf"; sourceTree = "<group>"; };
```
tommasini: This doens't seem necessary changes

tommasini (dismissed):
Let's just revert projcet.pbxproj changes they seem unrelated, and it's good on my sideeeee <3

tommasini (dismissed):

chaitanyapotti (dismissed):

vinnyhoward (dismissed):

EtherWizard33 (dismissed):
LGTM, approving for file app/components/Views/Settings/SecuritySettings/SecuritySettings.tsx
codeowned by wallet-ux, the one line change appears to be the expected one described in the PR description, which is storing in redux rather than localStorage.

sethkfman:
@MarioAslau Did you see [this](https://github.com/MetaMask/metamask-mobile/pull/15244#pullrequestreview-3017690527) feedback from Cursor Bot?

I was looking at the code and it looks like if we default to false users would not be routed to Vault Recovery.

MarioAslau:
> @MarioAslau Did you see [this](https://github.com/MetaMask/metamask-mobile/pull/15244#pullrequestreview-3017690527) feedback from Cursor Bot?
> 
> I was looking at the code and it looks like if we default to false users would not be routed to Vault Recovery.

addressed

sethkfman (dismissed):
LGTM

Cal-L (dismissed):
I've briefly reviewed Cursor bot comments, none of which we need to address now. ChoosePassword does eventually set existing user properly. The first comment is a TODO though since I'm not sure why `useDeleteWallet` hook separates out `resetWalletState` and `deleteUser`. Probably should consolidate in the future since it risks states going out of sync. Other than that, changes LGTM

infiniteflower (dismissed):

vinnyhoward (dismissed):

EtherWizard33 (approved):
re-approving previous approval for
> app/components/Views/Settings/SecuritySettings/SecuritySettings.tsx

chaitanyapotti (approved):

infiniteflower (approved):

app/components/Nav/App/App.tsx
```diff
@@ -972,6 +977,8 @@ const App: React.FC = () => {
     appTriggeredAuth().catch((error) => {
       Logger.error(error, 'App: Error in appTriggeredAuth');
     });
+    // existingUser and isMetaMetricsUISeen are not present in the dependency array because they are not needed to re-run the effect when they change and it will cause a bug.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tommasini, line 981: We shouldn't skip react-hooks because that disables react compiler, we need to make an effort to try to keep the app clean

We currently have 76 disablements at my current knowledge and we should work to decrease those 
thread: https://consensys.slack.com/archives/C087GPTR5HQ/p1752245049580249

app/components/Nav/App/App.tsx
```diff
@@ -1159,6 +1168,8 @@ const App: React.FC = () => {
     startApp().catch((error) => {
       Logger.error(error, 'Error starting app');
     });
+    // existingUser is not present in the dependency array because it is not needed to re-run the effect when it changes and it will cause a bug.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tommasini, line 1172: Same: https://github.com/MetaMask/metamask-mobile/pull/15244/files#r2210625184

app/components/Nav/App/App.tsx
MarioAslau: Happy to create a follow-up ticket :D

tommasini (approved):
Let's address those comments at nav/app at a follow up PR <3

app/components/Nav/App/App.tsx
MarioAslau: Resolving to be able to merge this PR - follow up ticket..will follow

app/components/Nav/App/App.tsx
MarioAslau: Resolving to be able to merge this PR - follow up ticket..will follow

Cal-L (approved):
Lgtm

chrisleewilcox (approved):
Removed run e2e label because failure is not related and is also occurring on main branch.

PR #15259, opened 2025-05-09
feat: extend TokenHero support for all ERC20 tokens - transfer transactions

digiwand:
## **Description**

This extends the support of the TokenHero component to all ERC20 tokens. 

Additionally:
- Updates the amount precision display logic by running the value through formatAmountMaxPrecision.
- Hides fiat values when hideFiatForTestnet is true

Helpers:
- adds useTokenAsset hook
- adds useTokenAmount hook
- deprecates useTokenValue in favor of useTokenAmount

Disclaimers: 
1. The fiat value in the simulation can sometimes mismatch the fiat value displayed in the rest of the application. There is a ticket to address this here: https://github.com/MetaMask/metamask-mobile/issues/15381
2. Notice Sepolia token does not have the purple background like the one in "Estimated changes" seciton. The white background matches the behavior in the extension.
3. Holding off on creating the AvatarTokenWithNetworkBadge test. Will add when adding support for NFT TokenHero
4. "No QA Needed" label as this is behind a feature flag and will be tested later as a whole

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4759

## **Manual testing steps**

Send transactions in-app or through dapp

### **After**

CleanShot 2025-05-19 at 22 10 24
CleanShot 2025-05-20 at 01 15 09
CleanShot 2025-05-19 at 22 05 22
CleanShot 2025-05-21 at 08 29 12

## **Pre-merge author checklist**

- [ ] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #15262, opened 2025-05-09
Temp `TokenHero` hook

OGPoyraz:
## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### **After**

ERC 20 transfer

https://github.com/user-attachments/assets/2cea90b7-c629-44ef-ad9a-489c8421d980

ERC 721 transfer

https://github.com/user-attachments/assets/edc6232a-80c9-4ab9-a3c0-8fc60ffed66a

Native transfer

https://github.com/user-attachments/assets/98f046e4-d18c-46a4-bf87-5506efea1168

### DIFF
--- a/app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
+++ b/app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
@@ -1,9 +1,16 @@
-import React, { useState } from 'react';
-import { StyleSheet, TouchableOpacity, View } from 'react-native';
-import { strings } from '../../../../../../../../locales/i18n';
-import Badge, {
-  BadgeVariant,
-} from '../../../../../../../component-library/components/Badges/Badge';
+import React from 'react';
+import {
+  StyleSheet,
+  TouchableOpacity,
+  View,
+  ImageSourcePropType,
+} from 'react-native';
+import { BigNumber } from 'bignumber.js';
+import { Hex } from '@metamask/utils';
+import { Collection } from '@metamask/assets-controllers';
+import { TransactionType } from '@metamask/transaction-controller';
+import { toChecksumAddress } from 'ethereumjs-util';
+
 import BadgeWrapper, {
   BadgePosition,
 } from '../../../../../../../component-library/components/Badges/BadgeWrapper';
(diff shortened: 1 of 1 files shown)


app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
```diff
@@ -75,44 +103,195 @@ const AssetFiatConversion = ({
...
+      const preciseFiatValue = amountBN.times(nativeConversionRateInBN);
+
+      amount = decimalAmount.toString();
+      tokenFormattedFiatValue = fiatFormatter(preciseFiatValue);
+      tokenName = nativeCurrency;
+      break;
+    }
+    case TransactionType.tokenMethodTransfer: {
+      // ERC20
+      const amountBN = new BigNumber(transferData?.args[1].toString());
```
digiwand, line 180: I think transferData?.args[1] should equate to `transactionData?.args?._value`

app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
```diff
@@ -75,44 +103,195 @@ const AssetFiatConversion = ({
...
+
+      const { name: erc20TokenName, iconUrl: image } =
+        erc20TokensByChain[chainId as Hex]?.data?.[txParams?.to as string] ??
+        {};
+
+      const { decimals } = tokenDetails ?? {};
```
digiwand, line 186: rather than using tokenDetails to fetch decimals, I think fetchErc20Decimals is a nice option since it will return the default of 18 decimals if not found. This keeps the type consistent and avoids `decimals ?? 18` below. 

tokenDetails may only be used for nft logic

digiwand (commented):
hi @OGPoyraz, thanks for creating this logic! It's helpful

From my QA, the components and hooks [I created the last couple of days](https://github.com/MetaMask/metamask-mobile/pull/15259) mostly resulted in the same display values as this PR while also supporting staking.

There was a bug supporting simpleSend network image and fiat values that I was able to fix in my PR after analyzing this logic. 

I am leaning towards keeping NFT logic separate from ERC20 logic to apply the separation of concerns. 

The extension code has a more simplified display value aggregation, similar to the PR I have open. It also utilizes SendHeading and NFTSendHeading to separate logic, which I think also makes sense in mobile. In our case, we can have something like TokenHero/HeroToken and HeroTokenNFT.

app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
```diff
@@ -75,44 +103,195 @@ const AssetFiatConversion = ({
...
+      const amountBN = new BigNumber(decimalAmount);
+      const nativeConversionRateInBN = new BigNumber(
+        nativeConversionRate as number,
+      );
+      const preciseFiatValue = amountBN.times(nativeConversionRateInBN);
+
+      amount = decimalAmount.toString();
+      tokenFormattedFiatValue = fiatFormatter(preciseFiatValue);
+      tokenName = nativeCurrency;
+      break;
```
digiwand, line 176: somewhere there is a bug here.

CleanShot 2025-05-09 at 17 12 44

https://github.com/MetaMask/metamask-mobile/pull/15259 appears to cover this

digiwand: merging fiat logic

app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
OGPoyraz: Sure thing, that make sense.

OGPoyraz: It's a bit of tradeoff I would say - to use `TokenListController` or make component to fetch the details itself. In ideal world we should use the same token state which is from `TokenListController` so that we don't have to make all components to fetch their own details.

digiwand:
@OGPoyraz 

> separate 3 components like NativeInfo, ERC20Info, NFTInfo (Maybe even ERC721 ERC1155 UnknownTokenInfo) and have those components under token-hero folder, not outside like we did it in the extension. 

I will consider this. I'm a fan of separating the logic by transaction type like you've suggested with the hook! It would be beneficial in that it would help us understand the code more easily and aid debugging the logic in the future. Following this, I think we'll want to update the logic in the extension as well

> It's reasonable to use utilities like fetchErc20Decimals from some other files but if we want to do that please move them into utilities if we need to. 
👍🏼 

thanks for the reminder with useDisplayHook too! I was planning to move this as well when we get closer to finalizing the logic in case we go with different logic

app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
digiwand: I agree it would be better to resolve this issue with a single source of truth.

I'll look into this more. Planning to examine if AssetsContractController is a candidate as well

PR #15280, opened 2025-05-12
fix: iOS `react-native-quick-crypto` crash

gantunesr:
## **Description**

This change conditionally sets the cryptography module in the following way,

- iOS -> `react-native-aes`
- Android -> `react-native-quick-crypto`

## **Related issues**

Fixes: https://github.com/MetaMask/accounts-planning/issues/929

## **Manual testing steps**

1. Create a new wallet from the previous version
2. Switch to this branch
3. Execute actions like lock and unlock
4. Reveal SRP
5. Reset wallet
  a. Import new SRP
  b. Create new SRP 

## **Screenshots/Recordings**

Not applicable


Cal-L (commented):
- The original issue is leading to a 404
- Code looks fine overall
- Do we know what the reason for the crash is on iOS when using quick crypto? (ideally, we can have one solution for both platforms)
- Let's also have @ccharly do a review here

gantunesr:
> The original issue is leading to a 404
I think you don't have access to the issue

> Do we know what the reason for the crash is on iOS when using quick crypto? (ideally, we can have one solution for both platforms)
I can't reproduce it nor find any information about it on the quick-crypto repository. One solution is the ideal path but it seems quite challenging given the release happening next week

Cal-L (approved):
This PR fixes the crash for when I create a wallet on a fresh install for iOS

app/core/Encryptor/Encryptor.test.ts
```diff
@@ -83,46 +88,65 @@ describe('Encryptor', () => {
...
         {
-          lib: 'random-lib',
+          lib: ENCRYPTION_LIBRARY.original,
           exportable: true,
           keyMetadata: LEGACY_DERIVATION_OPTIONS,
         },
       ],
       [
-        'non-exportable with random lib',
+        'non-exportable with quick lib iOS',
```
ccharly, line 133:
```suggestion
        'non-exportable with original lib (iOS)',
```

app/core/Encryptor/Encryptor.test.ts
```diff
@@ -83,46 +88,65 @@ describe('Encryptor', () => {
...
+          lib: ENCRYPTION_LIBRARY.original,
           exportable: false,
           keyMetadata: LEGACY_DERIVATION_OPTIONS,
         },
       ],
     ])(
       'generates a key with the right attributes: %s',
       async (_, { lib, exportable, keyMetadata }) => {
+        // @ts-expect-error - Mocking the getSystemName to return 'iOS'
+        jest.mocked(getSystemName).mockResolvedValueOnce('iOS');
```
ccharly, line 144: Not sure why we need that `ts-expect-error`, but another option could be to do (not tested):

```ts
const mockGetSystemName = jest.fn().mockResolvedValue('Android');

jest.mock('react-native-device-info', () => ({
  getSystemName: mockGetSystemName,
}));

...

      async (_, { lib, exportable, keyMetadata }) => {
        mockGetSystemName.mockResolvedValueOnce('iOS');
```

app/core/Encryptor/Encryptor.test.ts
```diff
@@ -294,7 +318,7 @@ describe('Encryptor', () => {
 
       const importedKey = await encryptor.importKey(result.exportedKeyString);
       expect(importedKey).toHaveProperty('exportable', true);
-      expect(importedKey).toHaveProperty('lib', 'original');
+      expect(importedKey).toHaveProperty('lib', ENCRYPTION_LIBRARY.quickCrypto);
```
ccharly, line 321: Very nit: But we do use raw literal in all other tests (with `.toHaveProperty`), so to be consistent, we could do the same here, or re-use the `enum` everywhere?

```suggestion
      expect(importedKey).toHaveProperty('lib', 'quick-crypto');
```

app/core/Encryptor/Encryptor.test.ts
```diff
@@ -83,46 +88,65 @@ describe('Encryptor', () => {
   describe('keyFromPassword', () => {
     it.each([
       [
-        'exportable with original lib',
+        'exportable with quick crypto lib (android)',
```
ccharly, line 91:
```suggestion
        'exportable with quick-crypto lib (android)',
```

app/core/Encryptor/Encryptor.test.ts
```diff
@@ -83,46 +88,65 @@ describe('Encryptor', () => {
...
         {
-          lib: ENCRYPTION_LIBRARY.original,
+          lib: ENCRYPTION_LIBRARY.quickCrypto,
           exportable: true,
           keyMetadata: LEGACY_DERIVATION_OPTIONS,
         },
       ],
       [
-        'non-exportable with original lib',
+        'non-exportable with quick crypto lib (android)',
```
ccharly, line 99:
```suggestion
        'non-exportable with quick-crypto lib (android)',
```

app/core/Encryptor/Encryptor.ts
```diff
@@ -335,6 +336,11 @@ class Encryptor implements WithKeyEncryptor<EncryptionKey, Json> {
       exportedKeyString,
     };
   };
+
+  #getCryptoLib = async (): Promise<EncryptionLibrary> => {
+    const os = await getSystemName();
+    return os === 'iOS' ? AesLib : QuickCryptoLib;
+  };
```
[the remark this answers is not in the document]
ccharly: Other suggestion: also, we could add a test for this new function to check that we do get the expected library based on the `getSystemName` value?

Cal-L:
I figured out why my builds were crashing on iOS when creating a new vault. It was because `FOX_CODE`, which is used as the salt for initializing `SecureKeychain` was not defined in my env file. And because SecureKeychain uses the encryptor under the hood, it was essentially passing it an empty salt, causing the app to crash. Apologies for the extra work, will close this PR.

ccharly:
IDK how easy it would be to prevent this behavior in the future (in case this happens to someone else), hopefully we could get a clear error message instead?

Nevertheless, I think that some of those changes you made @gantunesr are good, so we could just re-open this and just re-use QuickCrypto as the main `this.{lib,crypto}` (that's a nice to have, not very high-prio though)

gantunesr:
Thanks @Cal-L!

Agreed @ccharly, will open a new PR with some of the changes we had here

PR #15442, opened 2025-05-19
feat: Implementing features in simulation section for batched transactions

jpuri:
## **Description**

Implementing approve row for batched transaction in simulation section.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4955

## **Manual testing steps**

1. Enable 7702 locally
2. Submit a batched request needing approval
3. Check simulation section

## **Screenshots/Recordings**
Screenshot 2025-05-19 at 5 27 18 PM


app/components/UI/SimulationDetails/BatchApprovalRow/BatchApprovalRow.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return null;
+  }
+
+  return (
+    <>
+      {approveBalanceChanges.map((balanceChange) => (
+        <BalanceChangeRow
+          label={strings('confirm.simulation.label_change_type_approve')}
+          balanceChange={balanceChange as BalanceChange}
+        />
```
OGPoyraz, line 22: Can we pass a `key` property as it's best practice?

app/components/UI/SimulationDetails/BatchApprovalRow/BatchApprovalRow.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import { strings } from '../../../../../locales/i18n';
+import { useBatchApproveBalanceChanges } from '../../../Views/confirmations/hooks/7702/useBatchApproveBalanceChanges';
+import BalanceChangeRow from '../BalanceChangeRow/BalanceChangeRow';
+import { BalanceChange } from '../types';
+
+const BatchApprovalRow = () => {
+  const { value: approveBalanceChanges } =
+    useBatchApproveBalanceChanges() ?? {};
```
OGPoyraz, line 10: Isn't `useBatchApproveBalanceChanges` always returning an object, do we need that nullish operator?

app/components/UI/SimulationDetails/AmountPill/AmountPill.tsx
```diff
@@ -15,6 +15,9 @@ import { hexToDecimal } from '../../../../util/conversions';
 interface AmountPillProperties extends ViewProps {
   asset: AssetIdentifier;
   amount: BigNumber;
+  isApproval?: boolean;
+  isAllApproval?: boolean;
```
OGPoyraz, line 19: Just to get context, what is `isAllApproval`?

app/components/Views/confirmations/hooks/7702/useBatchApproveBalanceChanges.ts
```diff
@@ -0,0 +1,181 @@
...
+  isAll: boolean;
+  isUnlimited: boolean;
+  nestedTransactionIndex: number;
+};
+
+export type ApprovalBalanceChange = BalanceChange & {
+  nestedTransactionIndex: number;
+};
+
+export function useBatchApproveBalanceChanges() {
```
OGPoyraz, line 28: Can we type the return value of this hook as it's probably something like `{pending: boolean, value: ApprovalSimulationBalanceChange[]}`

app/components/UI/SimulationDetails/BatchApprovalRow/BatchApprovalRow.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+import * as BatchApprovalUtils from '../../../Views/confirmations/hooks/7702/useBatchApproveBalanceChanges';
+import { AssetType } from '../types';
+import BatchApprovalRow from './BatchApprovalRow';
+
+const approvalData = [
+  {
+    asset: {
+      type: 'ERC20' as AssetType.ERC20 | AssetType.ERC721 | AssetType.ERC1155,
+      address: '0x6b175474e89094c44da98b954eedeac495271d0f' as Hex,
+      chainId: '0x1' as Hex,
```
OGPoyraz, line 20: Minor, rather than type conversions here, wouldn't it be cleaner to tell `approvalData` is `ApprovalSimulationBalanceChange[]`? Otherwise this type conversion could fail if new type gets in

app/components/UI/SimulationDetails/useBalanceChanges.ts
```diff
@@ -188,10 +199,17 @@ export default function useBalanceChanges({
 }: {
   chainId: Hex;
   simulationData?: SimulationData;
-  networkClientId: string;
+  networkClientId?: string;
```
OGPoyraz, line 202: Why did we need to make it optional here? Can we let the consumer to pass `networkClientId` rather than using `Engine.context.NetworkController.findNetworkClientIdByChainId` here?

app/components/Views/confirmations/hooks/7702/useBatchApproveBalanceChanges.ts
```diff
@@ -0,0 +1,181 @@
...
+    const transaction = nestedTransactions[i];
+    const { data, to } = transaction;
+
+    if (!data || !to) {
+      continue;
+    }
+
+    const tokenData = await memoizedGetTokenStandardAndDetails({
+      tokenAddress: to,
+    });
```
OGPoyraz, line 136: Wouldn't it be faster to call this async function with `Promise.all` above and use the `tokenData` array then, because this is fetching one by one now.

OGPoyraz (changes requested):

app/components/UI/SimulationDetails/BatchApprovalRow/BatchApprovalRow.test.tsx
jpuri: using `ApprovalSimulationBalanceChange[]` does not works here

app/components/Views/confirmations/hooks/7702/useBatchApproveBalanceChanges.ts
jpuri: I do not expect it to be major performance issue as fn is already memoized and there will not be many different tokens in nested transactions.

OGPoyraz: Imo it does count every small performance improvement in mobile, especially thinking low-end android devices. Also I wouldn't trust assets memoization that much since approval balance changes on top and this will work first and then "simulations" will execute.

I probably overlooked at this in the extension PR - would definitely raise the same concern there as well. But since this is not a blocker and we want to keep parity with extension I will resolve this.

OGPoyraz (approved):

PR #15477, opened 2025-05-20
chore: Bump Snaps packages

FrederikBolding:
## **Description**

This bumps Snaps packages to the latest versions. Notable changes are:

- Support for `wallet_switchEthereumChain` inside Snaps.
   - Like with websites, we want to support a custom network per Snap rather than relying on the globally selected network. For Snaps, the permission for the network to switch to is automatically granted.
- Support for SIP-31 `onClientRequest` handler.
- New `showTestnets` property in `snap_getPreferences`.
- Added `isMinimumPlatformVersion` action.

For `wallet_switchEthereumChain` we have applied the same changes as on the extension where `isAddNetworkFlow` was changed to be `autoApprove`. It seems like `autoApprove` was already attempted to be used.

### `wallet_switchEthereumChain`

1. Go to [`test-snaps`](https://metamask.github.io/snaps/test-snaps/latest/)
2. Install the Ethereum Provider Example Snap.
3. Switch the chain ID using the `test-snaps` selector for the Ethereum Provider example.
4. Click "Get version".
5. The Snap should now use the new network, without any permission requests, and the appropriate chain ID should be shown.


FrederikBolding:
@SocketSecurity ignore npm/@metamask/snaps-execution-environments@8.1.0

Network access is expected.

app/core/RPCMethods/lib/ethereum-chain-utils.js
```diff
@@ -215,7 +215,7 @@ export function findExistingNetwork(chainId, networkConfigurations) {
  * @param {Function} params.requestUserApproval - The callback to trigger user approval flow.
  * @param {object} params.analytics - Analytics parameters to be passed when tracking event via `MetaMetrics`.
  * @param {string} params.origin - The origin sending this request.
- * @param {boolean} params.isAddNetworkFlow - Variable to check if its add flow.
+ * @param {boolean} params.autoApprove - Variable to check if the switc should be auto approved.
```
Mrtenz, line 218:
```suggestion
 * @param {boolean} params.autoApprove - Variable to check if the switch should be auto approved.
```

Mrtenz (changes requested):

Mrtenz (approved):

GuillaumeRx (approved):

app/core/RPCMethods/wallet_addEthereumChain.js
```diff
@@ -107,7 +107,6 @@ export const wallet_addEthereumChain = async ({
       requestUserApproval,
       analytics,
       origin,
-      isAddNetworkFlow,
       autoApprove: shouldAddOrUpdateNetwork,
```
ffmcgee725, line 110:
```
    const shouldAddOrUpdateNetwork =
      !existingNetwork ||
      rpcIndex !== existingNetwork.defaultRpcEndpointIndex ||
      (firstValidBlockExplorerUrl &&
        blockExplorerIndex !== existingNetwork.defaultBlockExplorerUrlIndex);
```
can we confirm no regressions where introduced in `switchToNetwork` call on `wallet_addEthereumChain`,  with changing `isAddNetworkFlow` from always being true, to now falling back to `autoApprove` (which computes to the `shouldAddOrUpdateNetwork` value above) ?

E2E Smoke tests seem to be passing but better safe than sorry since this handler is somewhat complex logic-wise on mobile codebase.

FrederikBolding: You tell me 😄 I'm not familiar enough with this code to check for regressions beyond what I've already done and all tests seem to pass.

app/util/permissions/index.ts
```diff
@@ -28,12 +27,12 @@ const approvalLog = createProjectLogger('approval-utils');
...
  * and returns a granted permissions object.
  *
- * @param {string} origin - The origin to request approval for.
+ * @param {string} _origin - The origin to request approval for.
  * @param requestedPermissions - The legacy permissions to request approval for.
  * @returns the approved permissions object.
  */
 export const getCaip25PermissionFromLegacyPermissions = (
-  origin: string,
+  _origin: string,
```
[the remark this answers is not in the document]
FrederikBolding: Simply mirroring the changes in https://github.com/MetaMask/metamask-extension/pull/33018 to reduce the diff here. We can always remove this in a follow-up refactor.

Daniel-Cross (approved):

app/util/permissions/index.ts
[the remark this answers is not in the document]
ffmcgee725: If we have a ticket to keep track of this, or make sure we address it in follow up refactor right after so it doesn't get lost in time, I'm okay with proceeding as such. Keeping diff to a minimum seems to make sense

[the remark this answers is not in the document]
FrederikBolding: https://github.com/MetaMask/metamask-mobile/issues/15520

app/core/RPCMethods/wallet_addEthereumChain.js
ffmcgee725: https://github.com/user-attachments/assets/2445efee-2522-440d-b8a3-9397ade269fb

I mean something like this, changing chain via the dapp to make sure the confirmation pop up to use the recently enabled network shows

FrederikBolding: I've done basic testing on the existing functionality before posting the PR for review, including existing EVM functionality as described in the video 👍

FrederikBolding: If you are asking for anything beyond that I would appreciate the help sanity checking the edge-cases.

ffmcgee725 (approved):

elribonazo (approved):

tommasini (approved):

PR #15538, opened 2025-05-22
fix: cp-7.46.2 Sanitize gas values before transaction updater functions

### LINKED ISSUE #15496: [Bug]: Json rpc and invalid transaction type envelope errors
### Describe the bug

We received reports of errors when transacting on some custom networks on mobile: `Invalid transaction envelope type: specified type "0x2" but included gasPrice instead of maxFeePerGas and maxPriorityFeePerGas`.

Such error was addressed in v7.46 by this PR https://github.com/MetaMask/metamask-mobile/pull/14935.

But it seems we're now hitting a new error "Internal Json rpc error" for the same transactions.

Update: the invalid transaction envelope type seems to still be happening on v7.46 as well

### Expected behavior

No errors and transaction is successful. 

### Screenshots/Recordings

https://github.com/user-attachments/assets/3af242f5-1598-4004-8df8-c50aa949a0ec

### Steps to reproduce

1. Have $5 in ETH on Abstract
2. Go to https://opensea.io/collection/dyli-abstract
3. Purchase a NFT
4. Confirm the transaction 
5. See the "Internal Json RPC error"

---
https://github.com/user-attachments/assets/ac375773-ece7-41ea-a837-7435c5f11f09

1. Go to https://app.pulsex.com/
2. Click pulsex.mypinata.cloud link
3. Swap PLSX token for DAI
4. See "invalid transaction envelope type" error

```shell

```

### Detection stage

In production (default)

### Version

7.46.1

### Build type

None

### Device

iphone 11

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

OGPoyraz:
## **Description**

This PR aims to sanitize gas properties before legacy transactions using updater functions.

Hot fix branch will be defined when it's created hence the title `cp-7.46.X`.
Change log will be added once version is defined.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/15496

## **Manual testing steps**

- import Abstract chain https://chainlist.org/chain/2741
- Have ETH on Abstract chain (Confirmations account have that ETH - do your test with `0xDc47789de4ceFF0e8Fe9D15D728Af7F17550c164`)
- Login opensea.io
- Navigate https://opensea.io/token/abstract/0xcefcffaa272c51a1962f0d73ea6ad9a9a8c64c1d
- In this page try swap from `ETH` to `ABBY` token ( you are on the correct page - you shouldn't need to change token selector in opensea.io)
- In `main` branch you will have envelope issue
- In this branch `ogp/15496` errors should not appear and swap should happen as it is.

### **Before**

https://github.com/user-attachments/assets/e0aa8154-eb94-4204-b81e-fd3ee599faa6

### **After**

https://github.com/user-attachments/assets/650b7051-b384-4747-b67e-02161b350690

### DIFF
--- a/app/util/transaction-controller/index.ts
+++ b/app/util/transaction-controller/index.ts
@@ -1,5 +1,6 @@
 import {
   TransactionParams,
+  TransactionEnvelopeType,
   TransactionController as BaseTransactionController,
 } from '@metamask/transaction-controller';
 import { Hex } from '@metamask/utils';
@@ -8,32 +9,6 @@ import Engine from '../../core/Engine';
 import { NetworkClientId } from '@metamask/network-controller';
 import { selectBasicFunctionalityEnabled } from '../../selectors/settings';
 import { store } from '../../store';
-
-function sanitizeTransactionParamsGasValues(
-  transactionId: string,
-  requestedTransactionParamsToUpdate: Partial<TransactionParams>,
-) {
-  const { TransactionController } = Engine.context;
-
-  const transactionMeta = TransactionController.getTransactions({
-    searchCriteria: { id: transactionId },
-  })?.[0];
-
-  if (!transactionMeta || !requestedTransactionParamsToUpdate) {
-    return;
-  }
-
-  const envelopeType = transactionMeta.txParams.type;
-
-  if (envelopeType === '0x0') {
-    // legacy transaction
-    delete requestedTransactionParamsToUpdate.maxFeePerGas;
-    delete requestedTransactionParamsToUpdate.maxPriorityFeePerGas;
-  } else {
-    // EIP-1559 compatible transaction
-    delete requestedTransactionParamsToUpdate.gasPrice;
-  }
-}
 
 export async function addTransaction(
   transaction: TransactionParams,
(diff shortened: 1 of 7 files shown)


app/util/transaction-controller/index.ts
```diff
@@ -9,6 +9,32 @@ import { NetworkClientId } from '@metamask/network-controller';
...
+    searchCriteria: { id: transactionId },
+  })?.[0];
+
+  if (!transactionMeta || !requestedTransactionParamsToUpdate) {
+    return;
+  }
+
+  const envelopeType = transactionMeta.txParams.type;
+
+  if (envelopeType === '0x0') {
```
matthewwalsh0, line 28: Minor, can we use `TransactionEnvelopeType.legacy`?

app/util/transaction-controller/index.ts
```diff
@@ -9,6 +9,32 @@ import { NetworkClientId } from '@metamask/network-controller';
...
+    // legacy transaction
+    delete requestedTransactionParamsToUpdate.maxFeePerGas;
+    delete requestedTransactionParamsToUpdate.maxPriorityFeePerGas;
+  } else {
```
matthewwalsh0, line 32: Should this be an `else if` as there could be no envelope type at all or a type 4?

Should we only delete `gasPrice` if we can see the 1559 parameters?

Should we set the legacy type here if there is only `gasPrice`?

app/util/transaction-controller/index.ts
```diff
@@ -9,6 +9,32 @@ import { NetworkClientId } from '@metamask/network-controller';
 import { selectBasicFunctionalityEnabled } from '../../selectors/settings';
 import { store } from '../../store';
 
+function sanitizeTransactionParamsGasValues(
```
matthewwalsh0, line 12: Minor, should we keep exported functions at the top?

matthewwalsh0 (changes requested):

app/util/transaction-controller/index.ts
OGPoyraz: > Should this be an else if as there could be no envelope type at all or a type 4?

We can add and `else if` for extra safety, done.

> Should we only delete gasPrice if we can see the 1559 parameters?

Wouldn't if we can't see 1559 parameters transaction will fail anyway? Isn't that overcomplicate logic here?

> Should we set the legacy type here if there is only gasPrice?

I think envelope type is not an object to change anywhere in client but for extra safety we can do it, done.

app/util/transaction-controller/index.ts
OGPoyraz: Done

app/util/transaction-controller/index.ts
OGPoyraz: Done

matthewwalsh0 (dismissed):

vinistevam (dismissed):

vinistevam (dismissed):

matthewwalsh0 (dismissed):

vinistevam (approved):

PR #15549, opened 2025-05-22
fix: update account list items design

PatrykLucka:
## **Description**

This PR updates account list item designs according to new designs for accounts hierarchy

Figma: https://www.figma.com/design/nmBs86I42Cp0T39BlRdX7x/Network-Expansion-Specs?node-id=4107-45412&t=5V9pe1PHnRCxIcXk-4

SRP pills should be removed after introducing accounts hierarchy.
## **Related issues**

Fixes:

## **Manual testing steps**

1. Open accounts list

### **Before**

Screenshot 2025-05-22 at 16 53 18

### **After**

Screenshot 2025-05-22 at 16 52 04


app/components/hooks/useMultichainBalances/useSelectedAccountMultichainBalances.test.ts
```diff
@@ -268,4 +268,84 @@ describe('useSelectedAccountMultichainBalances', () => {
...
+    const stateWithBalance = {
+      ...nonEvmState,
+      engine: {
+        ...nonEvmState.engine,
+        backgroundState: {
+          ...nonEvmState.engine.backgroundState,
+          MultichainBalancesController: {
+            balances: {
+              [MOCK_ACCOUNT_BIP122_P2WPKH.id]: {
+                'bip122:000000000019d6689c085ae165831e93/slip44:0': { amount: '0.5', unit: 'BTC' },
```
owencraston, line 311: Is this not balance data? Maybe I am confused but the balance seems to be 0.5 BTC.

app/util/networks/index.js
```diff
@@ -559,7 +559,7 @@ export const getNetworkImageSource = ({ networkType, chainId }) => {
     if (namespace !== KnownCaipNamespace.Eip155) {
       return getNonEvmNetworkImageSourceByChainId(chainId);
     }
-    hexChainId = toHex(reference);
+    hexChainId = toHex(reference === '0' ? '1' : reference); // default to mainnet if chainId is 0
```
owencraston, line 562: Can you explain this a bit more? I am a bit confused as to what this does?

app/component-library/components/Cells/Cell/foundation/CellBase/CellBase.types.ts
```diff
@@ -17,6 +17,10 @@ export interface CellBaseProps {
    * Title of the Cell Account, 1 line truncation.
    */
   title: React.ReactNode;
+  /**
+   * Optional props for title.
+   */
+  titleProps?: Partial<TextProps>;
```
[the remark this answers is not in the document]
PatrykLucka: I think it's better to have flexibility to use all TextProps, do we want to limit it for any reason?

app/components/hooks/useMultichainBalances/utils.ts
```diff
@@ -79,7 +79,7 @@ const getNonEvmDisplayBalance = (
   shouldShowFiat: boolean,
   currentCurrency: string,
 ) => {
-  if (!shouldShowFiat || !nonEvmAccountBalance.totalBalanceFiat) {
+  if (!shouldShowFiat || nonEvmAccountBalance.totalBalanceFiat === undefined) {
```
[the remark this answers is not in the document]
PatrykLucka: I'm not sure if it's still needed with the change I proposed below - to set `totalBalanceFiat` to `undefined` whenever there is issue with getting rate. But before that I encountered an issue where `totalBalanceFiat` was 0 even when the native token balance was not and because of that 0 was returned instead of native token balance.

app/selectors/multichain/multichain.ts
```diff
@@ -346,7 +346,8 @@ export const getMultichainNetworkAggregatedBalance = (
     if (totalBalanceFiat) {
       totalBalanceFiat = totalBalanceFiat.plus(balanceInFiat);
     } else {
-      totalBalanceFiat = balanceInFiat;
+      // If the rate is undefined, we don't want to set the total balance fiat to 0
+      totalBalanceFiat = rate !== undefined ? balanceInFiat : undefined;
```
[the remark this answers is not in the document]
PatrykLucka: I haven't noticed this happening anywhere

app/util/networks/index.js
PatrykLucka: Currently, for EVM addresses, we're using the format eip155:0:*address*, where 0 represents the network. This is fine because we want to indicate that it's compatible with all EVM networks. However, in this case, we're trying to retrieve a network image, and one doesn't exist for the non-existent '0' network. So I had to change it to 'mainnet' in order to get the Ethereum network image.
Let me know if it makes sense, I'll expand the comment there to make it more clear.

app/components/hooks/useMultichainBalances/useSelectedAccountMultichainBalances.test.ts
PatrykLucka: It is, it's 2nd condition in this test, which tests how it behaves with balance, but with no rates. It should just show "0.5 BTC" instead of fiat balance.

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -93,10 +95,9 @@ const EvmAccountSelectorList = ({
   }, [selectedAddresses]);
 
   const renderAccountBalances = useCallback(
```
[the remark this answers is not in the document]
PatrykLucka: Good point, from what I understand it comes from the distinction between EVMAccountSelectorList and CaipAccountSelectorList. The first one is deprecated, but still used in few places. I didn't want to make a bigger change here, but the first one was the one that was needed to be changed here, however we should probably also update the 2nd one? I'm not sure if it's something we should do now or whenever there will be a proper refactor/cleanup to use just the 2nd component. Wdyt?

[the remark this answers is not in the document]
owencraston: I agree. I was just asking to see if you knew.

app/selectors/multichain/multichain.ts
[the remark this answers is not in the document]
owencraston: its probably caught in the useMultichainBalances hook. This is good. Better to be more explicit.

app/util/networks/index.js
owencraston: Fair enough.

app/component-library/components/Cells/Cell/foundation/CellBase/CellBase.types.ts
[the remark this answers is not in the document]
owencraston: Valid point. I was mostly just thinking out loud in the case that we wanted some specific props.

owencraston (commented):
Simulator Screenshot - iPhone 16 Plus - 2025-06-02 at 21 19 31


Looks great. The only remaining "issue" I have are these inline props ([inline props with a .map creation](https://github.com/MetaMask/metamask-mobile/pull/15549/files#r2116201032), [inline styles](https://github.com/MetaMask/metamask-mobile/pull/15549/files#r2116202300)). Once that is addressed I will approve.

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -94,10 +98,44 @@ const EvmAccountSelectorList = ({
     return lookupSet;
   }, [selectedAddresses]);
 
+  // Memoized avatar props cache to avoid recreating arrays/objects for identical token sets
+  const avatarPropsCache = useMemo(() => {
```
owencraston: this seems overly complex. I think we could just do something like this...

```ts
const avatarProps = useMemo( => {
return tokens ? tokens.map((tokenObj) => ({
              ...tokenObj,
              variant: AvatarVariant.Token,
            })) : [
              {
                variant: AvatarVariant.Network,
                imageSource: networkImage,
              },
            ]}
, [deps] )
```

(pseudocode code so not exact)

PatrykLucka: The problem is that tokens array will be in the deps and it's something that will change on each `renderAccountBalances`, which is used in every `renderAccountItem`. So I think we need to have some kind of cache with each option memoized in order to make it memoize properly

infiniteflower (dismissed):

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
owencraston: okay fair enough. I still think that even though it might change a lot, it is still better to break it out of the prop definition since that is a better pattern and makes the JSX easier to read. Lets omit the custom cashing logic. Also in the future, when we add the react native compiler, having it separated will be beneficial and likely improve memoization.

owencraston (approved):

bfullam (approved):

brianacnguyen (approved):

PR #9203, opened 2024-04-11
refactor(encryptor): align Encryptor methods to match @metamask/browser-passworder

ccharly:
## **Description**

Align `Encryptor` methods to match `@metamask/browser-passworder` as much as possible.

## **Related issues**

Fixes https://github.com/MetaMask/accounts-planning/issues/379
 
### Update app version and verify secrets

1. Create a new wallet using the last version of the app
2. Lock and unlock the app
3. Update the wallet to the code in this branch
4. Lock and unlock the app
5. Add 2-3 accounts
6. Lock and unlock the app
8. Verify the private for a random account is still accessible
9. Verify the SRP for the wallet is still accessible
10. Connect a hardware wallet and import accounts
11. Lock and unlock the app
12. Verify the state of the app is consistent

## **Screenshots/Recordings**

Not applicable

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [x] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [x] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #1755, opened 2020-08-06
Improvement/migrate aes crypto lib

andrepimenta:
**Description**

Migration to the current version of aes-crypto lib from the forked one.

PR #15655, opened 2025-05-26
refactor: Encryptor class

gantunesr:
## **Description**

Low risk refactor to the `Encryptor` class

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUL-223

## **Manual testing steps**

1. Onboard into MetaMask
2. Lock and unlock the wallet

## **Screenshots/Recordings**

Not applicable


app/core/Encryptor/lib/aes-native.ts
```diff
@@ -45,6 +46,8 @@ export function assertIsKdfAlgorithm(algorithm: string): asserts algorithm is ty
  * Please use `QuickCryptoEncryptionLibrary` instead.
  */
 class AesEncryptionLibrary implements EncryptionLibrary {
+  type: string =  ENCRYPTION_LIBRARY.original;
```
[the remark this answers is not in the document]
gantunesr: weird...

fixed https://github.com/MetaMask/metamask-mobile/pull/15655/commits/a1710b5d6cb48a816035c96b91b26cbb6621a192

app/core/Encryptor/lib/quick-crypto.ts
```diff
@@ -22,6 +22,12 @@ export enum HashAlgorithmQuickCrypto {
  * Class representing the QuickCrypto encryption library.
  */
 class QuickCryptoEncryptionLibrary implements EncryptionLibrary {
+  /**
+   * Encryption library type. This must be unique identifier to be able to
+   * differentiate all encryption library implementations.
+   */
```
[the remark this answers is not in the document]
gantunesr: remove all comments since the interface already has jsdocs

fixed https://github.com/MetaMask/metamask-mobile/commit/a1710b5d6cb48a816035c96b91b26cbb6621a192

app/components/Views/AesCryptoTestForm/AesCryptoTestForm.tsx
```diff
@@ -75,7 +75,7 @@ const AesCryptoTestForm = () => {
     // TODO: Replace "any" with type
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     async (args: any[]) => {
-      const response = await encryptor?.keyFromPassword(args[0], args[1]);
+      const response = await encryptor?.keyFromPassword(args[0], args[1], ENCRYPTION_LIBRARY.quickCrypto);
```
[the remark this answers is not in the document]
gantunesr: fixed https://github.com/MetaMask/metamask-mobile/commit/a1710b5d6cb48a816035c96b91b26cbb6621a192

app/core/Encryptor/Encryptor.ts
```diff
@@ -112,14 +130,14 @@ class Encryptor implements WithKeyEncryptor<EncryptionKey, Json> {
     salt: string,
     exportable = true,
     opts: KeyDerivationOptions = this.keyDerivationOptions,
-    lib = ENCRYPTION_LIBRARY.quickCrypto,
+    encryptionLibType: string = this.encryptionLib.type,
   ): Promise<EncryptionKey> => {
-    const derivedKey = await QuickCryptoLib.deriveKey(password, salt, opts);
+    const derivedKey = await this.encryptionLib.deriveKey(password, salt, opts);
```
ccharly, line 135: Actually, after re-reviewing this, I don't think the `encryptionLibType` makes any more sense (not with the current implementation at least..)

It used to be (during my first refactor):
```ts
    const key = await getEncryptionLibrary(lib).deriveKey(password, salt, opts);
```

So this was working fine because we were getting the underlying crypto lib associated with the `lib` parameter.

But we're not doing this anymore here... I believe we were doing this initially to automatically re-encrypt with the most secure crypto lib we had initially. Going from `aes-forked` to `react-native-aes`, and since this one was defaulting to the "original" lib (== `react-native-aes), we were automatically re-creating a new key with the new "crypto parameters and lib".

Also, I think the current implementation is potentially broken (since the introduction of QuickCrypto, unless this one is 100% compatible with the "original" lib)

But in `decrypt` we use `keyFromPassword` and pass the `lib` that was initially used to re-generate the same key to decrypt. And when we re-`encrypt`, we automatically start using QuickCrypto.

We should probably restore the initial behavior and re-use `getEncryptionLib` here to allow "smooth migration" like in this PR:
- https://github.com/MetaMask/metamask-mobile/pull/9203

LMK if that makes sense 😄

gantunesr: The "original" lib and quick crypto are 100% compatible for the use cases that are relevant for the `Encryptor` class, therefore the implementation is not broken and a migration will not be necessary. Also, we don't have to maintain support for backwards compatibility with the forked library since we don't have any users on those versions anymore, here is the evidence,

- Relevant PR with the migration included in v1.0.0: https://github.com/MetaMask/metamask-mobile/pull/1755
- Active app versions: https://mixpanel.com/s/2we26K
- From Mobile Platform team: "_Everyone should also be getting minimum update versions to 7.45.0 when they open the app._"

Given this, the use of `lib` as a factor in these operations have become unnecessary, I'll keep the type property on the current classes but remove the use it of in the `Encryptor` class.

gantunesr: https://github.com/MetaMask/metamask-mobile/pull/15655/commits/58227c7f4dddbace231fe4aa12827cd7dc33adc4

ccharly: I'm ok to remove it if it does not make sense anymore indeed 👍

We might need to re-introduce this "auto-migration" one day if we need another crypto implementation, but for the time being, we can probably remove it yes 😄

Though, about "Everyone should also be getting minimum update versions to 7.45.0 when they open the app.", if it turns out that some old users (that do not have metrics enabled) get upgraded to 7.45.0, that might eventually break their vault I guess? 🤔

Having the auto-migration "feature" was a nice thing IMO, but if you prefer to get rid of it now, I won't fight back either 😄

PR #15704, opened 2025-05-26
fix: Do not lowercase account addresses cp-7.47.0

### LINKED ISSUE #15404: [Bug]: Do not `.toLowerCase` non-EVM addresses
### Describe the bug

- https://github.com/MetaMask/metamask-mobile/pull/14957

This PR mention using .toLowerCase, but we should not re-format non-EVM addresses, given that
Solana addresses are case-sensitive. This should not break anything, but we should avoid this in
the future.

### Expected behavior

Solana addresses are case sensitive so we should not format them at all, even when doing string comparison. 

### Screenshots/Recordings

_No response_

### Steps to reproduce

this is not a user facing bug

```shell

```

### Detection stage

In production (default)

### Version

7.47.0

### Build type

None

### Device

iPhone and Android

### Operating system

Android, iOS

### Additional context

_No response_

### Severity

_No response_

owencraston:
## **Description**
Given the introduction of Solana accounts in 7.47.0, we must now be diligent about address comparison. Solana addresses are case sensitive so we must remove all cases of `.toLowerCase` AND `.toLowerCaseEquals`  when performing address comparison/lookup. Instead I created a new function  called `areAddressesEqual`  (code pasted below) which we can use to handle this case sensitive comparison use case. I maintained the lowercase comparison for eth addresses since there are many assumptions about lower case addresses already in the codebase. In places where comparison was not needed I opted for the `toFormattedAddress` which handles the account type case for us. Instead of calling `safeToChecksumAddress` we should call  `toFormattedAddress` which conditionally checksums the address if the account is EVM. Both of these function definitions are pasted below.

```js
export function areAddressesEqual(address1: string, address2: string) {
  if (!address1 || !address2) {
    return false;
  }

  const isAddress1Eth = isEthAddress(address1);
  const isAddress2Eth = isEthAddress(address2);

  // If one is an ETH address and the other is not, return false
  if (isAddress1Eth !== isAddress2Eth) {
    return false;
  }

  // If both are ETH addresses, do a lowercase comparison
  if (isAddress1Eth && isAddress2Eth) {
    return address1.toLowerCase() === address2.toLowerCase();
  }

  // If both are not ETH addresses, do a raw comparison.
  // This is important for non-EVM addresses since they are case sensitive.
  return address1 === address2;
}


export function toFormattedAddress(address: string) {
  return isEthAddress(address) ? toChecksumAddress(address) : address;
}
```

We want to avoid any potential issues that could occurs from comparing two unique addresses that are equal when lowercased. This could cause issues such as ...
- incorrect assets being displayed
- sending to the wrong address
- Incorrect transaction data

#### Note
- There were several changes to some of these files that were auto format changes from prettier. The only real changes in this PR are swapping out `.toLowerCase` for `toFormattedAddress(address)` and the equivalent unit test fixes. 

### For reviewers
- Ensure that your owned file changes make sense. If a component is EVM only then we may not need to perform this change.
- Look for other cases of `.toLowerCase` AND `.toLowerCaseEquals` and ensure that the ones I left in the code make sense. 
    - I left some that were related to ENS, transactions and asset data.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/15404
Resolves: https://consensyssoftware.atlassian.net/browse/MUL-239

## **Manual testing steps**

1. Create a wallet
2. create a sol account
3. import an new SRP
4. create a new Sol account
5. perform a send between your accounts
6. ensure that all the data is correct
7. ensure that all the assets being displayed are as expected
8. ensure that the transaction data being shown on the transaction details screen is correct for each account.

### DIFF
--- a/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
+++ b/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
@@ -37,6 +37,24 @@ jest.mock('../../../../../core/Engine', () => ({
         ],
       },
     },
+    AccountsController: {
+      state: {
+        internalAccounts: {
+          selectedAccount: '30786334-3935-4563-b064-363339643939',
+          accounts: {
+            '30786334-3935-4563-b064-363339643939': {
+              id: '30786334-3935-4563-b064-363339643939',
+              address: '0x1234567890123456789012345678901234567890',
+              name: 'Account 1',
+              type: 'eip155:eoa',
+              metadata: {
+                lastSelected: 0,
+              },
+            },
+          },
+        },
+      },
+    },
     GasFeeController: {
       startPolling: jest.fn(),
       stopPollingByPollingToken: jest.fn(),
@@ -129,10 +147,8 @@ jest.mock('../../hooks/useBridgeQuoteData', () => ({
 }));
 
 jest.mock('../../../../../util/address', () => ({
-  isHardwareAccount: jest.fn().mockReturnValue(false),
-  formatAddress: jest.fn().mockImplementation((address) => address),
-  getLabelTextByAddress: jest.fn().mockReturnValue(''),
-  toFormattedAddress: jest.fn().mockImplementation((address) => address),
+  ...jest.requireActual('../../../../../util/address'),
+  isHardwareAccount: jest.fn(),
 }));
 
 describe('BridgeView', () => {
--- a/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
+++ b/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
@@ -132,6 +132,7 @@ jest.mock('../../../../../util/address', () => ({
   isHardwareAccount: jest.fn().mockReturnValue(false),
   formatAddress: jest.fn().mockImplementation((address) => address),
   getLabelTextByAddress: jest.fn().mockReturnValue(''),
+  toFormattedAddress: jest.fn().mockImplementation((address) => address),
 }));
 
 describe('BridgeView', () => {
@@ -679,7 +680,9 @@ describe('BridgeView', () => {
 
       // Wait for the banner text to appear
       await waitFor(() => {
-        expect(getByText(strings('bridge.hardware_wallet_not_supported'))).toBeTruthy();
+        expect(
+          getByText(strings('bridge.hardware_wallet_not_supported')),
+        ).toBeTruthy();
       });
     });
   });
@@ -780,4 +783,3 @@ describe('BridgeView', () => {
     });
   });
 });
-
--- a/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
+++ b/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
@@ -47,6 +47,7 @@ jest.mock('../../../../../core/Engine', () => ({
               address: '0x1234567890123456789012345678901234567890',
               name: 'Account 1',
               type: 'eip155:eoa',
+              scopes: 'eip155:0',
               metadata: {
                 lastSelected: 0,
               },
--- a/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
+++ b/app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
@@ -47,7 +47,7 @@ jest.mock('../../../../../core/Engine', () => ({
               address: '0x1234567890123456789012345678901234567890',
               name: 'Account 1',
               type: 'eip155:eoa',
-              scopes: 'eip155:0',
+              scopes: ['eip155:0'],
               metadata: {
                 lastSelected: 0,
               },
--- a/app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
+++ b/app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
@@ -13,7 +13,6 @@ import { useNavigation } from '@react-navigation/native';
 import { KeyringTypes } from '@metamask/keyring-controller';
 import { isAddress as isSolanaAddress } from '@solana/addresses';
 
-
 // External dependencies.
 import Cell, {
   CellVariant,
@@ -253,7 +252,8 @@ const EvmAccountSelectorList = ({
         onLongPress({
           address,
           isAccountRemoveable:
-            type === KeyringTypes.simple || (type === KeyringTypes.snap && !isSolanaAddress(address)),
+            type === KeyringTypes.simple ||
+            (type === KeyringTypes.snap && !isSolanaAddress(address)),
           isSelected: isSelectedAccount,
           index,
         });
@@ -323,9 +323,11 @@ const EvmAccountSelectorList = ({
       let selectedAccount: Account | undefined;
 
       if (selectedAddresses?.length) {
-        const selectedAddressLower = toFormattedAddress(selectedAddresses[0]);
+        const selectedFormattedAddress = toFormattedAddress(
+          selectedAddresses[0],
+        );
         selectedAccount = accounts.find(
-          (acc) => toFormattedAddress(acc.address) === selectedAddressLower,
+          (acc) => toFormattedAddress(acc.address) === selectedFormattedAddress,
         );
       }
       // Fall back to the account with isSelected flag if no override or match found
--- a/app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
+++ b/app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
@@ -24,6 +24,7 @@ import SensitiveText, {
 } from '../../../component-library/components/Texts/SensitiveText';
 import AvatarGroup from '../../../component-library/components/Avatars/AvatarGroup';
 import {
+  areAddressesEqual,
   formatAddress,
   getLabelTextByAddress,
   toFormattedAddress,
@@ -323,11 +324,9 @@ const EvmAccountSelectorList = ({
       let selectedAccount: Account | undefined;
 
       if (selectedAddresses?.length) {
-        const selectedFormattedAddress = toFormattedAddress(
-          selectedAddresses[0],
-        );
-        selectedAccount = accounts.find(
-          (acc) => toFormattedAddress(acc.address) === selectedFormattedAddress,
+        const selectedAddress = selectedAddresses[0];
+        selectedAccount = accounts.find((acc) =>
+          areAddressesEqual(acc.address, selectedAddress),
         );
       }
       // Fall back to the account with isSelected flag if no override or match found
--- a/app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
+++ b/app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
@@ -24,7 +24,11 @@ import SensitiveText, {
   SensitiveTextLength,
 } from '../../../component-library/components/Texts/SensitiveText';
 import AvatarGroup from '../../../component-library/components/Avatars/AvatarGroup';
-import { formatAddress, getLabelTextByAddress } from '../../../util/address';
+import {
+  formatAddress,
+  getLabelTextByAddress,
+  toFormattedAddress,
+} from '../../../util/address';
 import { AvatarAccountType } from '../../../component-library/components/Avatars/Avatar/variants/AvatarAccount';
 import { isDefaultAccountName } from '../../../util/ENSUtils';
 import { strings } from '../../../../locales/i18n';
@@ -87,7 +91,7 @@ const EvmAccountSelectorList = ({
     if (!selectedAddresses?.length) return null;
     const lookupSet = new Set<string>();
     selectedAddresses.forEach((addr) => {
-      if (addr) lookupSet.add(addr.toLowerCase());
+      if (addr) lookupSet.add(toFormattedAddress(addr));
     });
     return lookupSet;
   }, [selectedAddresses]);
@@ -233,7 +237,9 @@ const EvmAccountSelectorList = ({
       }
       let isSelectedAccount = isSelected;
       if (selectedAddressesLookup) {
-        isSelectedAccount = selectedAddressesLookup.has(address.toLowerCase());
+        isSelectedAccount = selectedAddressesLookup.has(
+          toFormattedAddress(address),
+        );
       }
 
       const cellStyle: ViewStyle = {
@@ -317,9 +323,9 @@ const EvmAccountSelectorList = ({
       let selectedAccount: Account | undefined;
 
       if (selectedAddresses?.length) {
-        const selectedAddressLower = selectedAddresses[0].toLowerCase();
+        const selectedAddressLower = toFormattedAddress(selectedAddresses[0]);
         selectedAccount = accounts.find(
-          (acc) => acc.address.toLowerCase() === selectedAddressLower,
+          (acc) => toFormattedAddress(acc.address) === selectedAddressLower,
         );
       }
       // Fall back to the account with isSelected flag if no override or match found
--- a/app/components/UI/Ramp/components/Account.tsx
+++ b/app/components/UI/Ramp/components/Account.tsx
@@ -11,7 +11,7 @@ import {
   selectSelectedInternalAccountFormattedAddress,
   selectInternalAccounts,
 } from '../../../../selectors/accountsController';
-import { toFormattedAddress } from '../../../../util/address';
+import { areAddressesEqual } from '../../../../util/address';
 
 // TODO: Convert into typescript and correctly type
 // TODO: Replace "any" with type
@@ -58,15 +58,12 @@ const Account = ({
   const internalAccounts = useSelector(selectInternalAccounts);
 
   const selectedInternalAccount = address
-    ? internalAccounts.find(
-        (account) =>
-          toFormattedAddress(account.address) === toFormattedAddress(address),
+    ? internalAccounts.find((account) =>
+        areAddressesEqual(account.address, address),
       )
     : selectedAddress
-    ? internalAccounts.find(
-        (account) =>
-          toFormattedAddress(account.address) ===
-          toFormattedAddress(selectedAddress),
+    ? internalAccounts.find((account) =>
+        areAddressesEqual(account.address, selectedAddress),
       )
     : undefined;
 
--- a/app/components/UI/Ramp/components/Account.tsx
+++ b/app/components/UI/Ramp/components/Account.tsx
@@ -11,7 +11,7 @@ import {
   selectSelectedInternalAccountFormattedAddress,
   selectInternalAccounts,
 } from '../../../../selectors/accountsController';
-import { toLowerCaseEquals } from '../../../../util/general';
+import { toFormattedAddress } from '../../../../util/address';
 
 // TODO: Convert into typescript and correctly type
 // TODO: Replace "any" with type
@@ -58,12 +58,17 @@ const Account = ({
   const internalAccounts = useSelector(selectInternalAccounts);
 
   const selectedInternalAccount = address
-    ? internalAccounts.find((account) =>
-        toLowerCaseEquals(account.address, address),
+    ? internalAccounts.find(
+        (account) =>
+          toFormattedAddress(account.address) === toFormattedAddress(address),
       )
-    : internalAccounts.find((account) =>
-        toLowerCaseEquals(account.address, selectedAddress),
-      );
+    : selectedAddress
+    ? internalAccounts.find(
+        (account) =>
+          toFormattedAddress(account.address) ===
+          toFormattedAddress(selectedAddress),
+      )
+    : undefined;
 
   const accountName = selectedInternalAccount?.metadata?.name || '';
   return (
--- a/app/components/Views/LedgerSelectAccount/index.tsx
+++ b/app/components/Views/LedgerSelectAccount/index.tsx
@@ -123,7 +123,7 @@ const LedgerSelectAccount = () => {
 
   useEffect(() => {
     keyringController.getAccounts().then((value: string[]) => {
-      setExistingAccounts(value);
+      setExistingAccounts(value.map(toFormattedAddress));
     });
   }, [keyringController]);
 
--- a/app/components/Views/LedgerSelectAccount/index.tsx
+++ b/app/components/Views/LedgerSelectAccount/index.tsx
@@ -40,6 +40,7 @@ import {
 } from '../../../core/Ledger/constants';
 import SelectOptionSheet from '../../UI/SelectOptionSheet';
 import { AccountsController } from '@metamask/accounts-controller';
+import { toFormattedAddress } from '../../../util/address';
 
 interface OptionType {
   key: string;
@@ -193,7 +194,7 @@ const LedgerSelectAccount = () => {
     if (LEDGER_LEGACY_PATH === (await getHDPath())) {
       const ledgerAccounts = await getLedgerAccounts();
       const newAddedAccounts = ledgerAccounts.filter(
-        (account) => !existingAccounts.includes(account.toLowerCase()),
+        (account) => !existingAccounts.includes(toFormattedAddress(account)),
       );
 
       if (newAddedAccounts.length > 0) {
--- a/app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
+++ b/app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
@@ -5,7 +5,10 @@ import { useAccounts } from '../../../hooks/useAccounts';
 import { RootState } from '../../../../reducers';
 import { useFetchAccountNotifications } from '../../../../util/notifications/hooks/useSwitchNotifications';
 import { getValidNotificationAccounts } from '../../../../selectors/notifications';
-import { toFormattedAddress } from '../../../../util/address';
+import {
+  areAddressesEqual,
+  toFormattedAddress,
+} from '../../../../util/address';
 
 export function useNotificationAccountListProps(addresses: string[]) {
   const { update, initialLoading, accountsBeingUpdated, data } =
@@ -20,10 +23,14 @@ export function useNotificationAccountListProps(addresses: string[]) {
   }, [addresses, update]);
 
   const isAccountLoading = (address: string) =>
-    accountsBeingUpdated.includes(toFormattedAddress(address));
+    accountsBeingUpdated.some(
+      (addr) => toFormattedAddress(addr) === toFormattedAddress(address),
+    );
 
   const isAccountEnabled = (address: string) =>
-    data?.[toFormattedAddress(address)] ?? false;
+    data?.[toFormattedAddress(address)] ??
+    data?.[address.toLowerCase()] ?? // fallback to lowercase address lookup
+    false;
 
   return {
     isAnyAccountLoading,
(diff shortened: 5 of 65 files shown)


ffmcgee725 (dismissed):
wallet-api-platform-engineers code owned reviewed (skipped rest). LGTM 🚀 🙌🏾

app/components/Views/confirmations/hooks/useAccountNativeBalance.ts
OGPoyraz, line 1: It's arguable why we are doing the lowercased comparison here.

But I don't think we want to take the risk here as comparisons done in very isolated for this hook. As these changes already very well tested and both usages about to get rolled out in next release, I prefer to revert this changes.

app/lib/ppom/ppom-util.ts
```diff
@@ -84,9 +85,10 @@ async function validateRequest(
       ?.params?.[0] as Partial<TransactionParams>;
 
     if (
+      toAddress &&
       internalAccounts.some(
         ({ address }: { address: string }) =>
-          address?.toLowerCase() === toAddress?.toLowerCase(),
+          toFormattedAddress(address) === toFormattedAddress(toAddress),
```
OGPoyraz: Same here - can we revert this change as well? As it is more secure check for now.

OGPoyraz (changes requested):

app/components/Views/confirmations/hooks/useAccountNativeBalance.ts
owencraston: @OGPoyraz can you confirm that this hook is only used for evm addresses?

OGPoyraz: Exactly @owencraston - it's usual transaction flows - nothing related with non-evm networks here

owencraston: Done in the latest commit (3682eb40c87ae95c4bca5979fae319a964c5a548)

app/lib/ppom/ppom-util.ts
owencraston: Done in the latest commit (3682eb40c87ae95c4bca5979fae319a964c5a548)

app/components/Views/confirmations/legacy/Send/index.js
```diff
@@ -427,8 +427,10 @@ class Send extends PureComponent {
     }
 
     newTxMeta.from = selectedAddress;
-    const fromAccount = internalAccounts.find((account) =>
-      toLowerCaseEquals(account.address, selectedAddress),
+    const fromAccount = internalAccounts.find(
+      (account) =>
+        toFormattedAddress(account.address) ===
+        toFormattedAddress(selectedAddress),
     );
```
OGPoyraz, line 431: These flows are also not used in non-evm user flows - we can revert

app/components/Views/confirmations/legacy/SendFlow/SendTo/index.js
```diff
@@ -225,12 +225,12 @@ class SendFlow extends PureComponent {
     const { addressBook, globalChainId, internalAccounts } = this.props;
     const networkAddressBook = addressBook[globalChainId] || {};
     const checksummedAddress = this.safeChecksumAddress(toAccount);
-    return !!(
-      networkAddressBook[checksummedAddress] ||
-      internalAccounts.find((account) =>
-        toLowerCaseEquals(account.address, checksummedAddress),
-      )
+    const account = internalAccounts.find(
```
OGPoyraz: Same here send flow is non-evm

app/components/Views/confirmations/legacy/Send/index.js
owencraston: Done in the latest commit (daef9348fcae74fcf6b537ed42f0b5dc8a8e1ae2)

app/components/Views/confirmations/legacy/SendFlow/SendTo/index.js
owencraston: Done in the latest commit (daef9348fcae74fcf6b537ed42f0b5dc8a8e1ae2)

OGPoyraz (commented):
Lifting off RC as there are no more confirmation changes, thanks @owencraston

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
```diff
@@ -19,10 +20,10 @@ export function useNotificationAccountListProps(addresses: string[]) {
   }, [addresses, update]);
 
   const isAccountLoading = (address: string) =>
-    accountsBeingUpdated.includes(address.toLowerCase());
+    accountsBeingUpdated.includes(toFormattedAddress(address));
 
   const isAccountEnabled = (address: string) =>
-    data?.[address.toLowerCase()] ?? false;
+    data?.[toFormattedAddress(address)] ?? false;
```
elribonazo: Would maybe suggest a cleaner approach like, little cleaner and easier to read. Fine with either way, feel free to solve comment if we don't agree:

```typescript
  const isAccountEnabled = (address: string) =>
    toFormattedAddress(address) in data;
```

hmalik88 (dismissed):

app/util/activity/index.ts
```diff
@@ -14,8 +13,10 @@ export const isFromOrToSelectedAddress = (
   to: string,
   selectedAddress: string,
 ): boolean =>
-  toLowerCaseEquals(safeToChecksumAddress(from), selectedAddress) ||
-  toLowerCaseEquals(safeToChecksumAddress(to), selectedAddress);
+  (from
```
elribonazo: Would propose one of the following to make this piece easier to read. Not strongly opinionated here but its def easier to read like this 2 options.

Option 1
```typescript
export const isFromOrToSelectedAddress = (
  from: string,
  to: string,
  selectedAddress: string,
): boolean => {
  if (from) {
    return toFormattedAddress(from) === toFormattedAddress(selectedAddress)
  }
  if (to) {
    return toFormattedAddress(to) === toFormattedAddress(selectedAddress)
  }
  return false;
}
```

Option 2:
```typescript
export const isFromOrToSelectedAddress = (
  from: string,
  to: string,
  selectedAddress: string,
): boolean =>
  from ?
    toFormattedAddress(from) === toFormattedAddress(selectedAddress):
    to ?
      toFormattedAddress(to) === toFormattedAddress(selectedAddress) :
      false;
```

elribonazo (changes requested):

app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
```diff
@@ -37,6 +37,24 @@ jest.mock('../../../../../core/Engine', () => ({
...
+    AccountsController: {
+      state: {
+        internalAccounts: {
+          selectedAccount: '30786334-3935-4563-b064-363339643939',
+          accounts: {
+            '30786334-3935-4563-b064-363339643939': {
+              id: '30786334-3935-4563-b064-363339643939',
+              address: '0x1234567890123456789012345678901234567890',
+              name: 'Account 1',
+              type: 'eip155:eoa',
```
ccharly, line 49: Nit: Not sure if that's important, but this account misses a `scopes: 'eip155:0'` field

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -317,9 +323,9 @@ const EvmAccountSelectorList = ({
       let selectedAccount: Account | undefined;
 
       if (selectedAddresses?.length) {
-        const selectedAddressLower = selectedAddresses[0].toLowerCase();
+        const selectedAddressLower = toFormattedAddress(selectedAddresses[0]);
```
ccharly: Nit: But we might want to also change the name, `Lower` does not make sense anymore now

app/components/Views/LedgerSelectAccount/index.tsx
```diff
@@ -193,7 +194,7 @@ const LedgerSelectAccount = () => {
     if (LEDGER_LEGACY_PATH === (await getHDPath())) {
       const ledgerAccounts = await getLedgerAccounts();
       const newAddedAccounts = ledgerAccounts.filter(
-        (account) => !existingAccounts.includes(account.toLowerCase()),
+        (account) => !existingAccounts.includes(toFormattedAddress(account)),
```
ccharly, line 197: Might be not needed, but we could also `.map` all `existingAccounts` (in the `useEffect` above) and format each of their addresses too.

IIRC, the `KeyringController` will return lower-cased addresses for EVM, meaning this one would fail (if we use `toChecksumAddress` in `toFormattedAddress`.

app/components/Views/Settings/NotificationsSettings/NotificationOptionToggle/index.tsx
```diff
@@ -123,7 +123,7 @@ const NotificationOptionToggle = ({
         </Text>
         {Boolean(address) && (
           <Text variant={TextVariant.BodyMD} color={TextColor.Alternative}>
-            {formatAddress(address, 'short').toLowerCase()}
+            {formatAddress(address, 'short')}
```
ccharly, line 126: Question: Do we want to "sanitize" the `address` still? We could use:
```suggestion
            {formatAddress(toFormattedAddress(address), 'short')}
```

This way this would show a checksummed address for EVM and just the `address` as-is for non-EVM?

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
```diff
@@ -87,9 +88,14 @@ const SnapSettings = () => {
...
         const addresses = await getAccountsBySnapId(snap.id);
+        const formattedAddresses = addresses.map((address) =>
+          toFormattedAddress(address),
+        );
         const snapIdentities = Object.values(internalAccounts).filter(
           (internalAccount) =>
-            addresses.includes(internalAccount.address.toLowerCase()),
+            formattedAddresses.includes(
+              toFormattedAddress(internalAccount.address),
+            ),
```
ccharly: Nit: We could omit having the `.map` above and rely on `.some` instead? That's very nit, but that could avoid mapping all addresses, when like the first one will be the one that match?

```suggestion
            const accountAddress = toFormattedAddress(internalAccount.address);
            addresses.some((address) =>
              toFormattedAddress(address) === accountAddress,
            ),
```

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
owencraston: works for me.

app/util/activity/index.ts
owencraston: Option 1 is cleaner to me.

app/components/Views/TransactionsView/index.js
```diff
@@ -125,16 +127,15 @@ const TransactionsView = ({
...
           return false;
         }
         const alreadySubmitted = submittedNonces.includes(nonce);
         const alreadyConfirmed = confirmedTxs.find(
           (tx) =>
-            toLowerCaseEquals(
-              safeToChecksumAddress(tx.txParams.from),
-              selectedAddress,
-            ) && tx.txParams.nonce === nonce,
+            toFormattedAddress(safeToChecksumAddress(tx.txParams.from)) ===
```
ccharly: Not sure if `safeToChecksumAddress` is needed anymore here 🤔 given that we only care about address equality here, we could omit it and use `toFormattedAddress` only IMO

ccharly: Moreover, `safeToChecksumAddress` could return `undefined` for non-EVM addresses! (Not sure we use this component for non-EVM, but it's probably best to follow the same pattern everywhere IMO)

app/components/Views/Wallet/index.tsx
```diff
@@ -680,7 +681,7 @@ const Wallet = ({
         // Call detect nfts
         const { NftDetectionController, NftController } = Engine.context;
         const previousNfts = cloneDeep(
-          NftController.state.allNfts[selectedAddress.toLowerCase()],
+          NftController.state.allNfts[toFormattedAddress(selectedAddress)],
```
ccharly: Nit: But we could save the result of `toFormattedAddress(selectedAddress)` since we use twice here and below (avoid computing the formatted address twice).

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
owencraston: Actually this resulted in a test failure so I kept my old approach.

app/util/activity/index.ts
owencraston: Actually option 1 resulted in a unit test failure so I had to modify it slightly but I did make it easier t read. commit is df6ce83be48898d4e7cb02dd92509ed73aec34f9

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -348,15 +359,16 @@ export class BackgroundBridge extends EventEmitter {
       }
       // Check if selectedAddress is approved
       const found = approvedAccounts
-        .map((addr) => addr.toLowerCase())
-        .includes(selectedAddress.toLowerCase());
+        .map((addr) => toFormattedAddress(addr))
+        .includes(toFormattedAddress(selectedAddress));
```
ccharly: Nit: We could also use `.some` here instead. But I can understand that `.map.includes` is pretty elegant too.

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -397,8 +409,8 @@ export class BackgroundBridge extends EventEmitter {
     // ONLY NEEDED FOR WC FOR NOW, THE BROWSER HANDLES THIS NOTIFICATION BY ITSELF
     if (this.isWalletConnect || this.isRemoteConn) {
       if (
-        this.addressSent?.toLowerCase() !==
-        memState.selectedAddress?.toLowerCase()
+        toFormattedAddress(this.addressSent) !==
+        toFormattedAddress(memState.selectedAddress)
```
ccharly: Nit: Type-wise this could be seen as incorrect given that both `this.addressSent` and `memState.selectedAddress` could be `undefined`.

Not sure if we want to allow `undefined` in `toFormattedAddress` or we want to check if they are defined here.. We might have this pattern in other places too.

app/core/Permissions/index.test.ts
```diff
@@ -908,32 +920,44 @@ describe('Permission Utility Functions', () => {
...
-      const accounts: Hex[] = ['0x1', '0x2', '0x3'];
+      const accounts: Hex[] = [
+        '0xc4955c0d639d99699bfd7ec54d9fafee40e4d272',
+        '0x742d35Cc6634C0532925a3b844Bc454e4438f44e',
+        '0x49b6FFd1BD9d1c64EEf400a64a1e4bBC33E2CAB2',
+      ];
       const internalAccounts = [
         {
-          address: '0X1', // Uppercase
+          address: '0xC4955C0D639D99699BFD7EC54D9FAFEE40E4D272', // Uppercase
```
ccharly, line 930: Watchout, the `0x` is not upper-cased here, but it is in the original code.

Also, small suggestions, we could re-use `accounts[0].toUpperCase()` here (the same is true for the other `address` below, but without the `.toUpperCase()`).

owencraston: This was failing because 0X is not a valid eth address so the toFormattedAddress was failing. I don't think this is a valid test is an eth address can not have 0X as a prefix.

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -55,7 +56,7 @@ class SnapKeyringImpl implements SnapKeyringCallbacks {
     const addresses = await this.#messenger.call(
       'KeyringController:getAccounts',
     );
-    return addresses.includes(address.toLowerCase());
+    return addresses.includes(toFormattedAddress(address));
```
ccharly: Watchout, `:getAccounts` will return a list of lower-case addresses for EVM accounts. So comparing them with a checksum address will most likely fail.

We might want to `addresses.map` and format them too.

app/components/UI/Ramp/components/Account.tsx
```diff
@@ -58,12 +58,17 @@ const Account = ({
   const internalAccounts = useSelector(selectInternalAccounts);
 
   const selectedInternalAccount = address
-    ? internalAccounts.find((account) =>
-        toLowerCaseEquals(account.address, address),
+    ? internalAccounts.find(
+        (account) =>
+          toFormattedAddress(account.address) === toFormattedAddress(address),
```
wachunei: nit: I see a lot of `toFormattedAddress(a) === toFormattedAddress(b)`, did you consider creating `toFormattedAddressEquals(a,b)`? Seems like a more equivalent change in the code.

wachunei (dismissed):
LGTM for Ramps

app/util/activity/index.ts
```diff
@@ -13,9 +12,23 @@ export const isFromOrToSelectedAddress = (
...
+
+  if (from && toFormattedAddress(from) === formattedSelectedAddress) {
+    return true;
+  }
+
+  if (to && toFormattedAddress(to) === formattedSelectedAddress) {
+    return true;
+  }
+
+  return false;
```
ccharly: Very nit: but we could re-use the previous pattern here too IMO:

```suggestion
  return
    (from && toFormattedAddress(from) === formattedSelectedAddress) ||
    (to && toFormattedAddress(to) === formattedSelectedAddress);
```

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
ccharly: I guess that's because `data` could be `undefined` too.

So maybe this could work instead:
```ts
const isAccountEnabled = (address: string) =>
  data && toFormattedAddress(address) in data;
```

bfullam (dismissed):

app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
owencraston: Not vital to this change but we might as well include it for the sake of accuracy.

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
owencraston: 🤦 very true.

app/components/Views/Settings/NotificationsSettings/NotificationOptionToggle/index.tsx
owencraston: This is unnecessary since the formatAdress method is defined as follows...

```js
export const formatAddress = (rawAddress: string, type: FormatAddressType) => {
  let formattedAddress = rawAddress;

  if (type && type === 'short') {
    formattedAddress = renderShortAddress(rawAddress);
  } else if (type && type === 'mid') {
    formattedAddress = renderSlightlyLongAddress(rawAddress);
  } else {
    formattedAddress = renderFullAddress(rawAddress);
  }

  return formattedAddress;
};

...


/**
 * Returns short address format
 *
 * @param {String} address - String corresponding to an address
 * @param {Number} chars - Number of characters to show at the end and beginning.
 * Defaults to 5.
 * @returns {String} - String corresponding to short address format
 */
export function renderShortAddress(address: string, chars = 5) {
  if (!address) return address;
  const formattedAddress = toFormattedAddress(address);
  return `${formattedAddress.substr(0, chars + 2)}...${formattedAddress.substr(
    -chars,
  )}`;
}

export function renderSlightlyLongAddress(
  address: string,
  chars = 4,
  initialChars = 20,
) {
  const formattedAddress = toFormattedAddress(address);
  return `${formattedAddress.slice(
    0,
    chars + initialChars,
  )}...${formattedAddress.slice(-chars)}`;
}

/**
 * Returns short address format
 *
 * @param {String} address - String corresponding to an address
 * @param {Number} chars - Number of characters to show at the end and beginning.
 * Defaults to 5.
 * @returns {String} - String corresponding to short address format
 */
export function renderShortAddress(address: string, chars = 5) {
  if (!address) return address;
  const formattedAddress = toFormattedAddress(address);
  return `${formattedAddress.substr(0, chars + 2)}...${formattedAddress.substr(
    -chars,
  )}`;
}

export function renderSlightlyLongAddress(
  address: string,
  chars = 4,
  initialChars = 20,
) {
  const formattedAddress = toFormattedAddress(address);
  return `${formattedAddress.slice(
    0,
    chars + initialChars,
  )}...${formattedAddress.slice(-chars)}`;
}
```

As you can see we already call `toFormattedAddress` in this util.

app/components/Views/TransactionsView/index.js
owencraston: good point.

app/components/Views/Wallet/index.tsx
owencraston: good point.

app/core/BackgroundBridge/BackgroundBridge.js
owencraston: fair point, I have added a nullcheck.

app/core/SnapKeyring/SnapKeyring.ts
owencraston: great catch. I will put this into a .some()

app/util/activity/index.ts
owencraston: While I agree, @elribonazo preferred this simpler easier to read approach which is why I made the change.

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
owencraston: this is still failing the test. I do not want to dive too deep into fixing every unit test. If you are okay with the current solution then I would prefer to keep it.

app/components/UI/Ramp/components/Account.tsx
owencraston: Great suggestion @wachunei. I thought of this after I realized how many changes I was making. I think your point is valid but part of me prefers this explicit yet verbose approach since it communicates throughout the codebase that addresses cannot be lower cased and instead must be formatted according to their account type. I am also trying to make the changes in this PR as easy to read for reviewers as possible. If you feel strongly that I should add this function then I can do it but otherwise, for the sake of speed, I would prefer to move forward as it.

wachunei: I don't feel strongly as this is a cherry pick needed for the next release. It can be tackled later if needed 👍

app/components/UI/Bridge/Views/BridgeView/BridgeView.test.tsx
jiexi: it should be an array. `scopes: ['eip155:0']`

jiexi:
I'm wondering if we could approach this the other direction. Rather than ensuring consistency at the sinks, what if we instead did that at the sources? By doing so we pay the checksum price once, and our internal code becomes easier to read and maintain because it does not need to be covered in case conversion preventative code. I understand that the existing changes were already probably very tedious to make, and that to go the other direction would be difficult as well.

app/components/UI/Ramp/components/Account.tsx
owencraston: I may actually go forward with this since there are so many lowerCase assumptions in the code that its getting really hard to get all the test cases to pass. I will report back when I have a solution.

app/components/Views/LedgerSelectAccount/index.tsx
owencraston: Good catch.

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
owencraston: @elribonazo do you know what format the addresses are returned in as a response from the `useFetchAccountNotifications` hook? This hook derives its data from the `fetchAccountNotificationSettings` which is defined as follows...

```
/**
 * Account Notification Settings.
 * - Informs us which accounts have notifications enabled.
 * @param accounts - accounts to check
 * @returns Record of Address <> Boolean (for which accounts are enabled/disabled)
 */
export const fetchAccountNotificationSettings = async (accounts: string[]) => {
  assertIsFeatureEnabled();
  const accountsStatus =
    await Engine.context.NotificationServicesController.checkAccountsPresence(
      accounts,
    );

  return accountsStatus;
};
```

If the addresses are in the wrong format this will cause the lookup to fail.

owencraston: Perhaps for now, I will fallback to the lowercase lookup.

app/components/Views/Snaps/SnapSettings/SnapSettings.tsx
owencraston: great point.

app/components/Views/Wallet/index.tsx
```diff
@@ -674,13 +675,15 @@ const Wallet = ({
...
+        const formattedSelectedAddress = toFormattedAddress(selectedAddress);
+
         trackEvent(
           createEventBuilder(MetaMetricsEvents.WALLET_COLLECTIBLES).build(),
         );
         // Call detect nfts
         const { NftDetectionController, NftController } = Engine.context;
         const previousNfts = cloneDeep(
-          NftController.state.allNfts[selectedAddress.toLowerCase()],
+          NftController.state.allNfts[formattedSelectedAddress],
```
owencraston, line 686: Need to verify that the nft controller holds the account ids as checksummed addresses for evm accounts.

app/util/smart-transactions/index.ts
```diff
@@ -28,7 +28,7 @@ export const getTransactionType = (
     transactionMeta?.origin !== TransactionTypes.MMM &&
     transactionMeta?.origin !== process.env.MM_FOX_CODE;
 
-  const to = transactionMeta.txParams.to?.toLowerCase();
+  const to = transactionMeta.txParams.to?.toLowerCase(); // should this be lowercase?
```
owencraston: @infiniteflower can you help me clarify this?

infiniteflower: Looks like this is checked against the swaps contract addresses, which are all in lower case https://github.com/MetaMask/swaps-controller/blob/main/src/constants.ts#L50

infiniteflower: If you want to leave this one as lower case, maybe add a comment so we know for the future.

app/components/Views/LedgerSelectAccount/index.tsx
owencraston: made the correct change above.

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
```diff
@@ -19,10 +23,14 @@ export function useNotificationAccountListProps(addresses: string[]) {
...
   const isAccountLoading = (address: string) =>
-    accountsBeingUpdated.includes(address.toLowerCase());
+    accountsBeingUpdated.some(
+      (addr) => toFormattedAddress(addr) === toFormattedAddress(address),
+    );
 
   const isAccountEnabled = (address: string) =>
-    data?.[address.toLowerCase()] ?? false;
+    data?.[toFormattedAddress(address)] ??
+    data?.[address.toLowerCase()] ?? // fallback to lowercase address lookup
```
Prithpal-Sooriya, line 32: TY for checking both areas. Will need to double-check, but IIRC this notification data returns a `Record<lowerCaseAddress, boolean>`

owencraston:
This PR was accidentally closed. I reopened it [here](https://github.com/MetaMask/metamask-mobile/pull/16068).

PR #15754, opened 2025-05-27
fix: cp-7.47.0 persist completedOnboarding state value

### LINKED ISSUE #15631: [Bug]: Backup and Sync is not updating on lock/unlock
### Describe the bug

After updating to 7.47.0 (with the redesign), the accounts are not syncing(updating) after a lock/unlock when updated happen on a different client. 

### Expected behavior

When the User Storage has been updated by a different client. The current client should get the latest changes after a lock/unlock of the wallet. 

### Screenshots/Recordings

_No response_

### Steps to reproduce

        Given I have the same SRP set up on two MetaMask clients
        And "Backup and Sync" is enabled on both clients
        And the "Accounts" toggle is enabled on the "Manage what you sync" section
        And I have the following accounts on the first client:
            | Account Name | Type |
            | Account 1    | EVM  |
            | Account 2    | EVM  |
        When I add a new EVM account on the first client
        And I navigate to the second client
        And I lock and unlock the second client
        Then I should see the new account in the account list
        And the account names should match between clients

```shell

```

### Detection stage

In production (default)

### Version

7.47.0

### Build type

None

### Device

iPhone 14, Pixel 8 

### Operating system

Other (please elaborate in the "Additional Context" section)

### Additional context

Android and iOS

### Severity

_No response_

mathieuartu:
## **Description**

This PR un-blacklists the `onboarding` reducer from `persistConfig`.
Instead, it adds granular control over which `onboarding` keys we want to persist.
Since it was blacklisted before, and in order to keep the previous logic, `onboarding.events` is still not persisted.

This PR also deprecates migration 71, and moves the original migration logic to `onPersistComplete`, as it is needed to correctly populate `completedOnboarding` for upgrading users that don't have `state.onboarding` yet.

As a reference, but not introduced by this PR:
- For new users, `completedOnboarding` is set to `true` after a proper onboarding.
- For all users, `completedOnboarding` is set to `false` after deleting the wallet.

This needs to be cherry-picked in `release/7.47.0` as it is a **release blocker**.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/15631

## **Manual testing steps**

1. No manual testing steps


app/store/persistConfig.ts
```diff
@@ -122,12 +122,26 @@ const persistUserTransform = createTransform(
...
 );
 
+const persistOnboardingTransform = createTransform(
+  (inboundState: RootState['onboarding']) => {
+    const { events, ...state } = inboundState;
+    // Reconstruct data to persist
+    return state;
+  },
+  null,
+  { whitelist: ['onboarding'] },
```
Prithpal-Sooriya, line 132: Very nice transform! Yep this is extensible if we need more control over the onboarding slice.

Prithpal-Sooriya (dismissed):

app/store/persistConfig.ts
Prithpal-Sooriya: Is profile-sync a feature of 7.47.0? 

If no (it was released earlier), do we need to add back this onboarding state?

mathieuartu: Re-added migration 71 disguised as 81 in consequence @Prithpal-Sooriya 🫡 🖖

app/store/migrations/081.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  try {
+    const KeyringController = state.engine.backgroundState.KeyringController;
+
+    if (
+      KeyringController &&
+      hasProperty(KeyringController, 'vault') &&
+      hasProperty(state, 'onboarding') &&
+      isObject(state.onboarding)
```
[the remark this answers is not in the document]
mathieuartu: Oh you're actually right, might as well delete those two conditions. Will pick this up tomorrow morning!

Prithpal-Sooriya (approved):

app/store/index.ts
```diff
@@ -64,12 +65,22 @@ const createStoreAndPersistor = async () => {
...
     // This sets the basic functionality value from the persisted state when the app is restarted
     store.dispatch(
-      toggleBasicFunctionality(
-        store.getState().settings.basicFunctionalityEnabled,
-      ),
+      toggleBasicFunctionality(currentState.settings.basicFunctionalityEnabled),
     );
+
+    // This sets the completedOnboarding value based on the KeyringController state
+    // This cannot be done in a migration because `state.onboarding` was previously blacklisted in `persistConfig`
```
tommasini, line 77: Hmm, we will be running this unnecessarily multiple times if its not a migration, am I wrong?

 Also, it was blacklisted, but that shouldnt block us from creating it during the migration, right?

tommasini (approved):

PR #15878, opened 2025-05-30
feat: Hero NFT redesign confirmation

digiwand:
## **Description**

- Abstracted Hero presentational component
- Create HeroNFT component 
- (TokenHero to be renamed and relocated to a similar path: hero/token in a follow-up PR)

No QA Needed label added as this will be QA'd as a whole with the rest of the transfer redesign

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5020

## **Manual testing steps**

1. Turn on transaction redesign feature flags
2. Go to NFT panel
3. Send NFT
4. Observe top hero component

### **After**

CleanShot 2025-05-27 at 23 24 30
CleanShot 2025-05-27 at 19 23 41

https://github.com/user-attachments/assets/d200ee20-73c6-42b6-a224-18e778ff689b


app/components/Views/confirmations/components/hero/index.ts
```diff
@@ -0,0 +1,2 @@
+export { Hero } from './hero';
+// export { HeroNft } from './nft/nft';
```
OGPoyraz, line 2:
```suggestion
```

app/components/Views/confirmations/hooks/nft/useNft.ts
```diff
@@ -0,0 +1,49 @@
...
+import { toHex } from '@metamask/controller-utils';
+import { collectiblesSelector } from '../../../../../reducers/collectibles';
+import { selectAllNftContracts } from '../../../../../selectors/nftController';
+import { safeToChecksumAddress } from '../../../../../util/address';
+import { parseStandardTokenTransactionData } from '../../utils/transaction';
+import { useTransactionMetadataRequest } from '../transactions/useTransactionMetadataRequest';
+
+export interface UseNftResponse {
+  chainId: string;
+
```
OGPoyraz, line 12:
```suggestion
```

app/components/Views/confirmations/components/info/transfer/transfer.tsx
```diff
@@ -26,9 +28,15 @@ const Transfer = () => {
...
 
+  const { isNft } = useIsNft();
+
   return (
     <View>
-      <TokenHero />
+      {isNft ? (
+        <HeroNft />
+      ) : (
+        <TokenHero />
```
OGPoyraz, line 38: Minor, shall we do a minor rename here? Either `TokenHero` to `HeroToken` or `HeroNft` to `NftHero`?

app/components/Views/confirmations/hooks/nft/useNft.ts
```diff
@@ -0,0 +1,49 @@
...
+  const nft = tokenId ? nfts.find((c) => c.tokenId === tokenId.toString()) : undefined;
+
+  const nftContract = useNftContract(chainId, tokenAddress);
+
+  return {
+    chainId: toHex(chainId).toString(),
+    name: nftContract?.name,
+    nft,
+    tokenId,
+  };
```
OGPoyraz, line 48: This NFT hooks look brilliant 🤩

app/components/Views/confirmations/components/hero/nft/nft.tsx
OGPoyraz, line 1: As discussed before let's use a flat structure here.

app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
```diff
@@ -1,119 +1,75 @@
...
+        title={<AssetAmount amount={amount} styles={styles} />}
+        subtitle={showFiatOnTestnets ? fiat : undefined}
+        tooltipModalProps={{
+          content: amountPrecise,
+          isEnabled: isRoundedAmount,
+          hasTooltip: true,
+          testId: 'token-hero-amount',
+          title: strings('send.amount'),
+        }}
+      />
```
OGPoyraz, line 72: Can we not create another component here?

Then it would be much cleaner if we would just have two components as you put in the render method of `token-hero`, either `NFTHero` or `TokenHero` as you already done it in `transfer.tsx` and no more file hop in those components?

OGPoyraz (changes requested):
This PR looks awesome, nothing major on code changes! 

Just two small re-organisation request and we are good to go with it

app/components/Views/confirmations/components/info/transfer/transfer.tsx
```diff
@@ -26,9 +28,15 @@ const Transfer = () => {
   useMaxValueRefresher();
   useEffect(trackPageViewedEvent, [trackPageViewedEvent]);
 
+  const { isNft } = useIsNft();
+
   return (
     <View>
-      <TokenHero />
+      {isNft ? (
```
OGPoyraz, line 35: Ideally let's have a single `components/rows/transactions/hero-row` component that either renders `TokenHero` or `NftHero`, so we don't have to put render logic directly here. There some benefits to have row single row like this - we can easily check

app/components/Views/confirmations/components/info/transfer/transfer.tsx
digiwand: yep! the plan is to rename and relocate TokenHero to the new file path components/hero. I wanted to refrain from this update now to ensure the code changes are easier to review.

thanks for bringing this up here.

app/components/Views/confirmations/components/info/transfer/transfer.tsx
digiwand: I like the idea of a single component. We can do this. 

I'm confused about how we are defining "row" in the context of confirmations. Initially, we have had sections with rows in them. 

Sections have had rounded corners and the default background color (in light-mode this would be white). 

I'm thinking this component might make more sense in
`app/components/Views/confirmations/components/confirm/confirm-component.tsx`
where the <Title /> component lives, rather than living within `confirmations/components/info/`...

thoughts on this? 

In the meantime, I've created `components/hero/transfer` to hold this condition. I've also relocated the `components/hero` directory to `components/hero/base` since I think this may improve its interpretation.

digiwand: hm I just read this comment about the flat file structure https://github.com/MetaMask/metamask-mobile/pull/15878#discussion_r2117317374. flattening this last update

digiwand: @OGPoyraz 

to clarify, is this what you mean 
CleanShot 2025-06-02 at 23 54 33?

as opposed to 
CleanShot 2025-06-02 at 23 55 08

OGPoyraz: It's totally understandable that confusion on the sections vs rows. But in file structure we decided to use `row` instead of `section`.

You may have a question of "when we define the second level folder (when to use `components/FOLDER`)" . That's totally tied to domain folders. For example `gas` or `7702`. In that case we don't want to have `hero` folder containing multiple components.

To be honest - I fall into same pattern while doing gas components and decided to use real flattened components. 
So I mean more like separated `components/hero-base` (assuming this is for ERC20 and native) , `components/hero-nft` and `components/rows/hero-row`(as we use it as row)

OGPoyraz: Btw there maybe still examples against what I said above. We are still trying to find what fits best but I think we are close to have finalising all these

app/components/Views/confirmations/components/rows/transactions/token-hero/token-hero.tsx
digiwand: created components/hero/transfer

app/components/Views/confirmations/hooks/nft/useNft.ts
digiwand: ty! found this interesting

`collectiblesSelector` and `NftController.findNftByAddressAndTokenId` (not currently used in metamask-mobile), both returned an NFT object with metadata. `collectiblesSelector` reads from `nftControllerState.allNfts`. The name they provide appends its ID, e.g., "Example Nft #777".

we needed the name of the NFT without the appended number. This can be determined by retrieving the NFT contract details.

`useDisplayName` aggregates multiple assets to resolve the name. This returned the name we desired without the appended ID, "Example Nft". Since it is pulling from the contract, it does not return the image we need for ERC721. 

In our case, the hook `useDisplayName` derived the nft contract from `useWatchedNFTNames`. `useWatchedNFTNames` pulled from `selectAllNftContracts`, which reads from `nftControllerState.allNftContracts`.

I replaced the `useDisplayName` and `useWatchedNFTNames` logic with the creation of useNftContract to directly find the contract by address and chain id.

app/components/Views/confirmations/components/info/transfer/transfer.tsx
digiwand: @OGPoyraz

ty for the clarification! 

though I must admit I'm still working to understand the full vision.
> Btw there maybe still examples against what I said above.
++ I think this is adding to my confusion  

with the domain examples, gas and 7702, I only found these directly in the `hooks` directory, pattern `hooks/DOMAIN`. `components` seems to have `components/COMPONENT/DOMAIN` or `components/COMPONENT/COMPONENT/DOMAIN`.

in our case, since we want `components/COMPONENT/DOMAIN`, we can flatten the component to support this pattern as you mentioned.

re: row 
referencing rows https://www.notion.so/metamask-consensys/Confirmations-code-folder-structure-improvements-1c0f86d67d6880dfa74acc25f2f27832

> Strictly only for the info row components themselves

To reiterate, I'm not sure hero belongs in info. Would it make sense it belongs in confirm-component.tsx, similarly to where <Title /> is placed? 

Accounting for all of the above, I updated the code. I moved what was `components/hero/base` → `components/UI/hero`. 
Then this PR now has `components/hero-nft` and `components/hero/transfer`. Later, we will move `components/rows/transactions/token-hero` → `components/hero-token`. Hope this works for you too 🙏🏼  curious to hear more of your thoughts on this

OGPoyraz: That `components/UI/hero` was a solid idea @digiwand ! 
I should've clarify my thoughts clearly sorry about that, I quickly wrapped up a PR that re-org folders a bit please see the PR below - added screenshot. That PR is adjusting your PR - I will merge it here if you don't mind

https://github.com/MetaMask/metamask-mobile/pull/16057

app/components/Views/confirmations/hooks/nft/useNft.ts
OGPoyraz: That totally make sense

app/components/Views/confirmations/components/info/transfer/transfer.tsx
OGPoyraz: I've done this renaming here, fyi https://github.com/MetaMask/metamask-mobile/pull/16057

OGPoyraz (approved):

app/components/Views/confirmations/components/info/transfer/transfer.tsx
digiwand: sounds good to me! ty @OGPoyraz

app/components/Views/confirmations/components/UI/hero/hero.tsx
```diff
@@ -0,0 +1,108 @@
...
+}: HeroProps) => {
+  const [isModalVisible, setIsModalVisible] = useState(false);
+
+  const { styles } = useStyles(styleSheet, {
+    hasPaddingTop,
+  });
+
+  return (
+    <View style={styles.base}>
+      {componentAsset}
```
digiwand, line 84: we could consider changing this to {children}, or renaming componentAsset. It's so far used to display the asset, but we could be make this more generic to support other cases

OGPoyraz: Make sense @digiwand , let's do it on a follow up PR - as we need NFT's look broken on `main` now, merging.

pedronfigueiredo (approved):

PR #15901, opened 2025-05-30
feat: Adds missing metrics for `transfer` confirmation redesign

OGPoyraz:
## **Description**

This PR aims to add `transaction_transfer_usd_value` and `asset_type` into `transfer` tarnsaction confirmations.

Adding `No QA needed` as these transaction confirmations not live.
Adding `No Smoke e2e needed` as these transactions doesn't affect with no user flow or current e2e tests.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4969

## **Manual testing steps**

1. Trigger send flow with redesigned enabled to see event metrics.

### DIFF
--- a/app/components/Views/confirmations/hooks/useTokenAmount.ts
+++ b/app/components/Views/confirmations/hooks/useTokenAmount.ts
@@ -58,8 +58,10 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
   );
   const { networkNativeCurrency } = useNetworkInfo(chainId as Hex);
   const currencyRates = useSelector(selectCurrencyRates);
-  const usdConversionRate =
-    currencyRates?.[networkNativeCurrency as string]?.usdConversionRate ?? 0;
+  const usdConversionRateFromCurrencyRates =
+    currencyRates?.[networkNativeCurrency as string]?.usdConversionRate;
+  const usdConversionRate = usdConversionRateFromCurrencyRates ?? 0;
+
 
   const tokenAddress = safeToChecksumAddress(txParams?.to) || NATIVE_TOKEN_ADDRESS;
   const { value: decimals, pending } = useTokenDecimals(tokenAddress, networkClientId);
@@ -91,7 +93,7 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
       // Native
       fiat = amount.times(nativeConversionRate);
       const usdAmount = amount.times(usdConversionRate);
-      usdValue = usdAmount.isZero() ? null : usdAmount.toFixed(2);
+      usdValue = usdConversionRateFromCurrencyRates ? null : usdAmount.toFixed(2);
       break;
     }
     case TransactionType.contractInteraction:
@@ -101,7 +103,7 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
       fiat = amount.times(nativeConversionRate).times(contractExchangeRate);
 
       const usdAmount = amount.times(contractExchangeRate).times(usdConversionRate);
-      usdValue = usdAmount.isZero() ? null : usdAmount.toFixed(2);
+      usdValue = usdConversionRateFromCurrencyRates ? null : usdAmount.toFixed(2);
       break;
     }
     default: {
--- a/app/components/Views/confirmations/hooks/useTokenAmount.ts
+++ b/app/components/Views/confirmations/hooks/useTokenAmount.ts
@@ -31,7 +31,7 @@ interface TokenAmount {
   amountPrecise: string | undefined;
   amount: string | undefined;
   fiat: string | undefined;
-  usdValue: string | undefined;
+  usdValue: string | null;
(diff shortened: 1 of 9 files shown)


app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -82,13 +90,15 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
     case TransactionType.stakingUnstake: {
       // Native
       fiat = amount.times(nativeConversionRate);
+      usdValue = amount.times(usdConversionRate).toString();
       break;
     }
     case TransactionType.contractInteraction:
     case TransactionType.tokenMethodTransfer: {
       // ERC20
       const contractExchangeRate = contractExchangeRates?.[tokenAddress]?.price ?? 1;
```
pedronfigueiredo: `contractExchangeRate` and `usdConversionRate` should both default to 0 when they can't be determined.

app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -82,13 +90,15 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
...
       fiat = amount.times(nativeConversionRate).times(contractExchangeRate);
+      usdValue = amount.times(contractExchangeRate).times(usdConversionRate).toString();
```
pedronfigueiredo: `usdValue` should be null if `usdConversionRate` is 0. This way the PO can differentiate in the metrics whether a transaction had value 0 or the conversion rate couldn't be determined.

app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -82,13 +90,15 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
     case TransactionType.stakingUnstake: {
       // Native
       fiat = amount.times(nativeConversionRate);
+      usdValue = amount.times(usdConversionRate).toString();
```
pedronfigueiredo: On extension `sending_value` is determined from the value that is displayed to the user, which is rounded to 2 decimal places. Worth double checking with @bschorchit if we want the same here for consistency.

app/components/Views/confirmations/hooks/useTokenAmount.ts
OGPoyraz: Very good call - fixed this.

app/components/Views/confirmations/hooks/useTokenAmount.ts
OGPoyraz: Thanks for this catch - fixed.

app/components/Views/confirmations/hooks/useTokenAmount.ts
OGPoyraz: You're right, I think we want consistency here, fixed.

app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -82,13 +90,18 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
     case TransactionType.stakingUnstake: {
       // Native
       fiat = amount.times(nativeConversionRate);
+      const usdAmount = amount.times(usdConversionRate);
+      usdValue = usdAmount.isZero() ? null : usdAmount.toFixed(2);
```
pedronfigueiredo: This logic makes it so if the conversion rate is undefined on line 62, usdAmount is 0 here, which makes usd value null. If the amount is 0 and the conversion rate is something, usd value is exactly the same, null. That's not exactly what we want.

The case where no conversion rate was determined and where the transfer amount is zero should be different in terms of usdValue (null and 0 respectively).

OGPoyraz: 🤦‍♂️ you're right - fixing it now

OGPoyraz: Fixed - now exclusively checking `usdConversionRateFromCurrencyRates` and set `usdValue` into `null` only if it doesn't exists

app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -82,13 +92,18 @@ export const useTokenAmount = ({ amountWei }: TokenAmountProps = {}): TokenAmoun
     case TransactionType.stakingUnstake: {
       // Native
       fiat = amount.times(nativeConversionRate);
+      const usdAmount = amount.times(usdConversionRate);
+      usdValue = usdConversionRateFromCurrencyRates ? null : usdAmount.toFixed(2);
```
pedronfigueiredo: these ternaries are reversed right?

OGPoyraz: I guess I need to make myself another coffee right now

pedronfigueiredo (approved):

PR #15926, opened 2025-05-30
chore: prevent TokenList rerenders

Prithpal-Sooriya:
## **Description**

This migrates this external contribution internally so we can run the full CI test suite.
External PR: https://github.com/MetaMask/metamask-mobile/pull/15692

-----

When user switched to Popular Networks there were 2 extra rerenders because switching between networks will trigger refetch of TokenRatesController and TokenListController. These refetches caused some selectors to update, specifically ones used to generate sortedTokenKeys, but difference by update is usually so small, that order of sortedTokenKeys stays same anyway so it's unecessary to rerender to component.

We also agreed with @sahar-fehri that these updates are happening correctly (more info in Slack https://margelo.slack.com/archives/C08P7RJQU04/p1748003584532309?thread_ts=1747998202.840729&cid=C08P7RJQU04)

I moved generating sortedTokenKeys to separate selector which simplified component and thanks to deep equality it prevents rerenders unless order is really changed. I also added some displayNames and some missing useCallback.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MMASSETS-884

## **Manual testing steps**

Interact with the token list and ensure no bugs are due to these perf changes.

1. View popular networks, single network
2. Switch networks and accounts, view token list.
3. Interact with a token list item.

### DIFF
--- a/app/selectors/accountsController.ts
+++ b/app/selectors/accountsController.ts
@@ -25,7 +25,7 @@ import { isEqualCaseInsensitive } from '@metamask/controller-utils';
 
 export type InternalAccountWithCaipAccountId = InternalAccount & {
   caipAccountId: CaipAccountId;
-}
+};
 
 /**
  *
@@ -100,6 +100,14 @@ export const selectSelectedInternalAccount = createDeepEqualSelector(
 );
 
 /**
+ * A memoized selector that returns the selected internal account id
+ */
+export const selectSelectedInternalAccountId = createSelector(
+  selectSelectedInternalAccount,
+  (account): string | undefined => account?.id,
+);
+
+/**
  * A memoized selector that returns the internal accounts sorted by the last selected timestamp
  */
 export const selectOrderedInternalAccountsByLastSelected = createSelector(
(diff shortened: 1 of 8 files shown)


app/selectors/accountsController.ts
```diff
@@ -99,6 +99,14 @@ export const selectSelectedInternalAccount = createDeepEqualSelector(
   },
 );
 
+/**
+ * A memoized selector that returns the selected internal account id
+ */
+export const selectSelectedInternalAccountId = createSelector(
+  selectSelectedInternalAccount,
+  (account): string | undefined => account?.id,
```
salimtb, line 107: Why can this selector return `undefined` ? Could this potentially cause issues in other places where the selector is used ? should we fallback on a default value instead ?

Prithpal-Sooriya: Great point! The `undefined` part came from this input selector:
https://github.com/MetaMask/metamask-mobile/blob/b9c4155938a7e353571f27896f720efd89f0f6f0/app/selectors/accountsController.ts#L82-L100

I think we are okay since this is a new selector, and places that use this handle undefined 😄

sahar-fehri (approved):

PR #15992, opened 2025-06-02
feat: identity/profile multichain analytics

wenfix:
## **Description**

Adds pre-configured and newly added chain ids to profile analytics

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4988

## **Manual testing steps**

1. Start App
  1.1 Identity event should fire and include a chain_id_list property with all configured chain ids
2. Add a popular or custom new network
  2.1 Identity event should fire and include the previous list with the newly added chain id
3. Remove a network
  3.1 Identity event should fire and include the chain_id_list without the removed chain id

## **Screenshots/Recordings**

Segment events:

On app start:
identity_on_open

Added chain with ID 80094 (Berachain):
adding_berachain

Remove chain with ID 10 (Optimism):
removing_optimism


app/util/metrics/UserSettingsAnalyticsMetaData/UserProfileAnalyticsMetaData.types.ts
```diff
@@ -26,4 +27,5 @@ export interface UserProfileMetaData {
   [UserProfileProperty.CURRENT_CURRENCY]?: string;
   [UserProfileProperty.HAS_MARKETING_CONSENT]: string;
   [UserProfileProperty.NUMBER_OF_HD_ENTROPIES]: number;
+  [UserProfileProperty.CHAIN_IDS]: string[];
```
ffmcgee725: is it possible to use type `CaipChainId` instead of `string` here, or will it complicate things ?

app/util/metrics/MultichainAPI/getNetworksConfigured.ts
```diff
@@ -0,0 +1,23 @@
...
+  chainId: CaipChainId | string,
+): CaipChainId[] {
+  const reduxState = store.getState();
+
+  const _chainId = isCaipChainId(chainId)
+    ? chainId
+    : toCaipChainId('eip155', chainId);
+
+  const networks = selectNetworkConfigurations(reduxState);
+  const chainIds = Object.values(networks).map((n) => n.chainId);
```
jiexi: this will result in a mix of evm hex Chain IDs and CAIP Chain IDs

ffmcgee725: nice catch! I believe @jiexi is right on this one

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @returns The chain ID in CAIP format.
+ */
+function caipifyChainId(chainId: CaipChainId | string): CaipChainId {
+  const _chainId = isHexString(chainId)
+    ? convertHexToDecimal(chainId)
+    : chainId;
+
+  return isCaipChainId(_chainId)
+    ? _chainId
+    : toCaipChainId('eip155', _chainId.toString());
```
adonesky1: we should probably add a comment explaining why this would only every be an `eip155` chain

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+import { store } from '../../../store';
+import { UserProfileProperty } from '../UserSettingsAnalyticsMetaData/UserProfileAnalyticsMetaData.types';
+import { convertHexToDecimal } from '@metamask/controller-utils';
+
+/**
+ * Converts a chain ID to CAIP format.
+ * @param chainId - The chain ID to convert, either as a CAIP chain ID or string.
+ * @returns The chain ID in CAIP format.
+ */
+function caipifyChainId(chainId: CaipChainId | string): CaipChainId {
```
adonesky1, line 18: Can you explain why `toCaipChainId` imported from `@metamask/utils` doesn't work instead?

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+  return isCaipChainId(_chainId)
+    ? _chainId
+    : toCaipChainId('eip155', _chainId.toString());
+}
+
+/**
+ * Gets an array of all configured network chain IDs converted to CAIP format.
+ * @returns An array of CAIP chain IDs for all configured networks.
+ */
+export function getConfiguredChainIdsCaipified(): CaipChainId[] {
```
adonesky1: is this **all** networks or just EVM networks?
```suggestion
export function getConfiguredCaipChainIds(): CaipChainId[] {
```

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+export function getChainIdListProperty(chainId: CaipChainId | string): {
+  [UserProfileProperty.CHAIN_IDS]: CaipChainId[];
+} {
+  const caipChainId = caipifyChainId(chainId);
+  const chainIds = getConfiguredChainIdsCaipified();
+
+  return {
+    [UserProfileProperty.CHAIN_IDS]: [...chainIds, caipChainId],
+  };
+}
```
adonesky1: Why wouldn't we be able to just refetch the list of configured caip chainIds once this has been added? I'm worried that we're creating two sources of truth this way

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Gets an array of chain IDs excluding the provided chain ID from the configured networks.
+ * @param chainId - The chain ID to remove from the list of configured networks.
+ * @returns An array of CAIP chain IDs containing the configured networks without the provided chain ID.
+ */
+export function getChainIdListPropertyWithout(chainId: CaipChainId | string): {
```
adonesky1: Same question [as above](https://github.com/MetaMask/metamask-mobile/pull/15992/files#r2126957971)

app/util/metrics/MultichainAPI/networkMetricUtils.ts
wenfix: This function uses that under the hood but the issue was that we get the network id's from the configuration in either hex or CAIP version (solana/bitcoin). Using `toCaipChainId` directly would result in `eip:0x1` for example. It was also not possible to use toEvmCaipChainId given that there are already caip2 addresses in the config.

app/util/metrics/MultichainAPI/networkMetricUtils.ts
wenfix: All networks as we get solana and bitcoin caip chain ids

app/util/metrics/MultichainAPI/networkMetricUtils.ts
ffmcgee725: ➕

wenfix: done

app/util/metrics/MultichainAPI/getNetworksConfigured.ts
wenfix: thanks, updated!

app/util/metrics/MultichainAPI/networkMetricUtils.ts
adonesky1:
```suggestion
export function removeItemFromChainIdList(chainId: CaipChainId | string): {
```

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  return Object.values(networks).map((n) => caipifyChainId(n.chainId));
+}
+
+/**
+ * Gets an array of chain IDs including both the configured networks and the provided chain ID.
+ * @param chainId - The chain ID to append to the list of configured networks.
+ * @returns An array of CAIP chain IDs containing both the configured networks and the provided chain ID.
+ */
+export function getChainIdListProperty(chainId: CaipChainId | string): {
```
adonesky1:
```suggestion
export function addItemToChainIdList(chainId: CaipChainId | string): {
```

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @param chainId - The chain ID to convert, either as a CAIP chain ID or string.
+ * @returns The chain ID in CAIP format.
+ */
+function caipifyChainId(chainId: CaipChainId | string): CaipChainId {
+  const _chainId = isHexString(chainId)
+    ? convertHexToDecimal(chainId)
+    : chainId;
+
+  //For Non-EVM chains, the chainId should already be in CAIP format. For EVM chains, we convert to CAIP by adding the 'eip155' namespace
+  return isCaipChainId(_chainId)
```
adonesky1: could we check this and return early?

app/components/Views/Settings/NetworksSettings/NetworkSettings/index.js
```diff
@@ -1596,6 +1602,11 @@ export class NetworkSettings extends PureComponent {
     const [, networkConfiguration] = entry;
     const { NetworkController } = Engine.context;
     NetworkController.removeNetwork(networkConfiguration.chainId);
+
+    MetaMetrics.getInstance().addTraitsToUser(
+      getChainIdListPropertyWithout(networkConfiguration.chainId),
```
adonesky1: looks like this name needs to be the new one?
```suggestion
      removeItemFromChainIdList(networkConfiguration.chainId),
```

app/components/UI/NetworkModal/index.tsx
```diff
@@ -242,6 +244,8 @@ const NetworkModals = (props: NetworkProps) => {
         ],
       });
 
+      addTraitsToUser(addItemToChainIdList(chainId));
```
adonesky1, line 247: So... now that this call is coming after the `chainId` is added to the state I would expect the adding of the `chainId` with the `addItemToChainIdList` would be redundant?

have you been able to test to verify what the `selectNetworkConfigurations(reduxState)` returns right here? So whether the new chainId is in redux state yet?

wenfix: Yes, it isn't available yet. This is because the `NetworkController` emits a `NetworkController:stateChange` event to which the `Engine` subscribes (via `EngineService`) in order to update the store. While the promise resolves synchronously here, the event will be handled by in a separate microtask that (likely) won't be concluded before we call `addItemToChainIdList`.

app/components/Views/NetworkSelector/NetworkSelector.test.tsx
```diff
@@ -15,6 +15,37 @@ import { NetworkListModalSelectorsIDs } from '../../../../e2e/selectors/Network/
 import { isNetworkUiRedesignEnabled } from '../../../util/networks/isNetworkUiRedesignEnabled';
 import { mockNetworkState } from '../../../util/test/network';
 
+// Mock network metric utility functions
```
ffmcgee725: non blocking
I'm okay with removing these comments, since the line of code is pretty self descriptive itself

app/components/Views/Settings/NetworksSettings/NetworkSettings/index.test.tsx
```diff
@@ -16,6 +16,41 @@ import Engine from '../../../../../core/Engine';
...
+  useMetrics: () => ({
+    trackEvent: jest.fn(),
+    createEventBuilder: jest.fn(() => ({
+      addProperties: jest.fn(() => ({
+        build: jest.fn(),
+      })),
+    })),
+  }),
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  withMetricsAwareness: (Component: any) => Component,
```
ffmcgee725: perhaps typing as `unknown` instead of `any` removed the need for the eslint disable comment

app/util/metrics/MultichainAPI/networkMetricUtils.test.ts
```diff
@@ -0,0 +1,283 @@
...
+
+describe('networkMetricUtils', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('getConfiguredCaipChainIds', () => {
+    it('should return empty array when no networks are configured', () => {
+      // @ts-expect-error - mocking store for test
+      mockStore.getState.mockReturnValue({});
```
ffmcgee725: I see we use this multiple times, so perhaps we can

- declare a global variable in this module, and type cast it `{} as typeof mockStore.getState`
- use it inside the test blocks, removing the need for `@ts-expect-error`

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    return chainId;
+  }
+
+  //If its not a hex string, we assume its a decimal chainId
+  if (!isHexString(chainId)) {
+    return toCaipChainId('eip155', chainId);
+  }
+
+  // For EVM chains, we convert to CAIP by adding the 'eip155' namespace
+  return toEvmCaipChainId(chainId as Hex);
```
ffmcgee725, line 30:
```suggestion
  return toEvmCaipChainId(toHex(chainId));
```

you can import `toHex` from '@metamask/utils' from what I recall, removing the need for type casting

ffmcgee725 (commented):
Left some comments 👍🏾 Overall very good work!

app/util/metrics/MultichainAPI/networkMetricUtils.ts
wenfix: we're already certain it's a hex string because of line 26 above it. this way we're saving a runtime call by doing it at compile

ffmcgee725 (approved):

adonesky1 (approved):

NidhiKJha (approved):

app/util/metrics/MultichainAPI/networkMetricUtils.test.ts
```diff
@@ -0,0 +1,279 @@
...
+      // @ts-expect-error - mocking selector for test
+      mockSelectNetworkConfigurations.mockReturnValue(mockNetworks);
+
+      const result = getConfiguredCaipChainIds();
+
+      expect(result).toEqual(['eip155:1', 'eip155:137', 'eip155:43114']);
+    });
+
+    it('should handle decimal chain IDs', () => {
+      const mockNetworks = {
```
tommasini, line 75: Would be nice to get this typed, to make sure it respects the network configruation state property wdyt?

tommasini (approved):
LGTM only left a suggestion to try to type the mock data structures at the test

app/components/UI/NetworkModal/index.tsx
```diff
@@ -343,6 +347,8 @@ const NetworkModals = (props: NetworkProps) => {
         blockExplorerUrl,
       );
 
+      addTraitsToUser(addItemToChainIdList(chainId));
```
tommasini, line 350: Any logic behind this traits being added to metrics before `setActiveNetwork` being called? I'm assuming that we just want this on metrics if setActiveNetwork runs successfully

app/util/metrics/MultichainAPI/networkMetricUtils.ts
tommasini: I'm wondering if we should improve the toCaipChainId with some of this logic https://github.com/MetaMask/utils/blob/9dd5b12a7e1901b539782485afdb34c6425cc0cc/src/caip-types.ts#L379
And leave a comment here that this is temporary, would be great to have only one function to convert to caip format

app/components/UI/NetworkModal/index.tsx
adonesky1: It only matters if the network is added really, not if the network is switched to

app/util/metrics/MultichainAPI/networkMetricUtils.ts
tommasini: Hmm, I'm not sure that we are certain, we know that it's not a caipChainId and we know that it could be an Hex at this point, but we do not know for sure that it would be an hex, right? That's why the TS is screaming. If it's an hex the way to be sure it's by using the [toHex](https://github.com/MetaMask/core/blob/e4c1ffaebc1b2016534d879aea50eb5d4c70808d/packages/controller-utils/src/util.ts#L216) @ffmcgee725 mentioned. Not sure if we are saving that big overhead, because if it's 99% most likely a hex we are just asserting it.

app/util/metrics/MultichainAPI/networkMetricUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/**
+ * Gets an array of all configured network chain IDs converted to CAIP format.
+ * @returns An array of CAIP chain IDs for all configured networks.
+ */
+export function getConfiguredCaipChainIds(): CaipChainId[] {
```
tommasini, line 37: NIT: I guess this is a temporary function right, since in the future (hopefully near future), every chain id in the app would be a `CaipChainId`

PR #16068, opened 2025-06-04
fix: Do not lowercase account addresses cp-7.47.0

### LINKED ISSUE #15404: [Bug]: Do not `.toLowerCase` non-EVM addresses

owencraston:
## **Description**
Given the introduction of Solana accounts in 7.47.0, we must now be diligent about address comparison. Solana addresses are case sensitive so we must remove all cases of `.toLowerCase` AND `.toLowerCaseEquals`  when performing address comparison/lookup. Instead I created a new function  called `areAddressesEqual`  (code pasted below) which we can use to handle this case sensitive comparison use case. I maintained the lowercase comparison for eth addresses since there are many assumptions about lower case addresses already in the codebase. In places where comparison was not needed I opted for the `toFormattedAddress` which handles the account type case for us. Instead of calling `safeToChecksumAddress` we should call  `toFormattedAddress` which conditionally checksums the address if the account is EVM. Both of these function definitions are pasted below.

```js
export function areAddressesEqual(address1: string, address2: string) {
  if (!address1 || !address2) {
    return false;
  }

  const isAddress1Eth = isEthAddress(address1);
  const isAddress2Eth = isEthAddress(address2);

  // If one is an ETH address and the other is not, return false
  if (isAddress1Eth !== isAddress2Eth) {
    return false;
  }

  // If both are ETH addresses, do a lowercase comparison
  if (isAddress1Eth && isAddress2Eth) {
    return address1.toLowerCase() === address2.toLowerCase();
  }

  // If both are not ETH addresses, do a raw comparison.
  // This is important for non-EVM addresses since they are case sensitive.
  return address1 === address2;
}


export function toFormattedAddress(address: string) {
  return isEthAddress(address) ? toChecksumAddress(address) : address;
}
```

We want to avoid any potential issues that could occurs from comparing two unique addresses that are equal when lowercased. This could cause issues such as ...
- incorrect assets being displayed
- sending to the wrong address
- Incorrect transaction data

#### Note
- There were several changes to some of these files that were auto format changes from prettier. The only real changes in this PR are swapping out `.toLowerCase` for `toFormattedAddress(address)` and the equivalent unit test fixes. 

### For reviewers
- Ensure that your owned file changes make sense. If a component is EVM only then we may not need to perform this change.
- Look for other cases of `.toLowerCase` AND `.toLowerCaseEquals` and ensure that the ones I left in the code make sense. 
    - I left some that were related to ENS, transactions and asset data.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/15404
Resolves: https://consensyssoftware.atlassian.net/browse/MUL-239

## **Manual testing steps**

1. Create a wallet
2. create a sol account
3. import an new SRP
4. create a new Sol account
5. perform a send between your accounts
6. ensure that all the data is correct
7. ensure that all the assets being displayed are as expected
8. ensure that the transaction data being shown on the transaction details screen is correct for each account.

## **Screenshots/Recordings**

There should be no visible changes with this PR.

### **Before**

https://github.com/user-attachments/assets/0fa9230f-f9de-4524-9fce-f7fd11063043

### **After**

https://github.com/user-attachments/assets/02329abe-c945-4a35-97f2-09401a2a0bdb

#### Tested imported account with multiple tokens and NFTS and its working as expected

<image src="https://github.com/user-attachments/assets/6b590409-9917-4d24-8187-f945fbbeb39e" height="500" width="250" />

<image src="https://github.com/user-attachments/assets/e1f11be1-f2d7-4aac-b260-5c601e9b6290" height="500" width="250" />


app/components/Views/Wallet/index.tsx
```diff
@@ -674,13 +675,15 @@ const Wallet = ({
...
+        const formattedSelectedAddress = toFormattedAddress(selectedAddress);
+
         trackEvent(
           createEventBuilder(MetaMetricsEvents.WALLET_COLLECTIBLES).build(),
         );
         // Call detect nfts
         const { NftDetectionController, NftController } = Engine.context;
         const previousNfts = cloneDeep(
-          NftController.state.allNfts[selectedAddress.toLowerCase()],
+          NftController.state.allNfts[formattedSelectedAddress],
```
owencraston, line 686: Need to verify that the nft controller holds the account ids as checksummed addresses for evm accounts.

app/components/Views/Settings/NotificationsSettings/AccountsList.hooks.tsx
```diff
@@ -19,10 +23,14 @@ export function useNotificationAccountListProps(addresses: string[]) {
...
 
   const isAccountLoading = (address: string) =>
-    accountsBeingUpdated.includes(address.toLowerCase());
+    accountsBeingUpdated.some(
+      (addr) => toFormattedAddress(addr) === toFormattedAddress(address),
+    );
 
   const isAccountEnabled = (address: string) =>
-    data?.[address.toLowerCase()] ?? false;
+    data?.[toFormattedAddress(address)] ??
```
owencraston, line 31: From @Prithpal-Sooriya on the previous PR

"TY for checking both areas. Will need to double-check, but IIRC this notification data returns a `Record<lowerCaseAddress, boolean>`"

infiniteflower (dismissed):
Thank you for the useful util, looks great

wachunei (dismissed):
LGTM for ramp

bfullam (dismissed):

hmalik88 (dismissed):

OGPoyraz (dismissed):

Prithpal-Sooriya (dismissed):

ffmcgee725 (dismissed):

bfullam (approved):

wachunei (approved):

ffmcgee725 (approved):

elribonazo (approved):

mathieuartu (approved):
LGTM! 

As a side note, we investigated on Slack the originally failing identity E2E tests and found out we still need to lowercase addresses in identity E2E mocks. 
This is because we're tied with how `AccountsController` implements `InternalAccount`s.
As long as `AccountsController` automatically lowercases accounts' addresses, we'll have to keep doing it in identity mocks.

infiniteflower (approved):

Cal-L (approved):
LGTM

Prithpal-Sooriya (approved):

OGPoyraz (approved):

gantunesr (approved):

vinnyhoward (approved):

hmalik88 (approved):

PR #16140, opened 2025-06-05
fix: cp-7.47.0 accounts for scientific notation in amounts

### LINKED ISSUE #16062: [Bug]: App crash when swapping with certain tokens using many decimals - Solana
### Describe the bug

When swapping in the Solana network, there are certain tokens that allows until a certain number of decimals, for example USDC allows 6, PONK allows 5, and while that token is selected we can't input a higher number of decimals, but if we select another token like SOL, which allows more decimals, and input a higher number of decimals and then switch to a token like USDC or PONK, the app crashes.

### Expected behavior

We should get a warning about the decimals, but the app should not crash (like in mainnet).

### Screenshots/Recordings

USDC:
https://www.loom.com/share/8e077a3e2f9e4a278fdfc35dc59f0d74

PONK:
https://www.loom.com/share/803eee556e564e27a5574ee15bdacdd1

### Steps to reproduce

1. Select a Solana account
2. Go to swaps
3. While SOL token is selected as source input the amount "0.123123123"
4. Switch the source token to USDC or PONK (might be happening the same with more tokens, I found this 2)

```shell
View: Root
Error: fractional component exceeds decimals [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT ] (fault="underflow", operation="parseFixed", code=NUMERIC_FAULT, version=bignumber/5.8.0)
```

### Detection stage

During release testing

### Version

7.47.0

### Build type

None

### Device

iPhone 14

### Operating system

iOS

### Additional context

Was able to reproduce with Android.

One time while reproducing in Android the app got stuck and I had to re-install it, but I wasn't able to replicate this again. Usually closing and re-opening the app solves it.

### Severity

_No response_

bfullam:
## **Description**
Hook currently breaks if a token amount is so small that it uses scientific notation. This adds additional checks to normalize those values before passing into parseUnits.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/16062

## **Manual testing steps**

1. Go to Solana swap
2. Input source token amount of 0.00001 SOL
3. Select AI16Z as destination token
4. Wait for quote
5. Click "switch tokens" arrow button
6. See no crash

### DIFF
--- a/app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
+++ b/app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
@@ -10,6 +10,17 @@ interface UseIsInsufficientBalanceParams {
   token: BridgeToken | undefined;
 }
 
+const normalizeAmount = (value: string, decimals: number): string => {
+  // Check if the value is in scientific notation
+  if (value.includes('e') || value.includes('E')) {
+    // Convert to decimal notation using the token's decimal precision
+    const num = Number(value);
+    const result = num.toFixed(decimals);
+    return result;
+  }
+  return value;
+};
+
 const useIsInsufficientBalance = ({ amount, token }: UseIsInsufficientBalanceParams): boolean => {
   const { quoteRequest } = useSelector(selectBridgeControllerState);
   const latestBalance = useLatestBalance({
@@ -24,7 +35,9 @@ const useIsInsufficientBalance = ({ amount, token }: UseIsInsufficientBalancePar
 
   // Safety check for decimal places before parsing
   const hasValidDecimals = isValidAmount && (() => {
-    const decimalPlaces = amount.includes('.') ? amount.split('.')[1].length : 0;
+    // Convert scientific notation to decimal string if needed
+    const normalizedAmount = normalizeAmount(amount, token.decimals);
+    const decimalPlaces = normalizedAmount.includes('.') ? normalizedAmount.split('.')[1].length : 0;
     return decimalPlaces <= token.decimals;
   })();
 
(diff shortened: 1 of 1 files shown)


app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
```diff
@@ -10,6 +10,17 @@ interface UseIsInsufficientBalanceParams {
   token: BridgeToken | undefined;
 }
 
+const normalizeAmount = (value: string, decimals: number): string => {
+  // Check if the value is in scientific notation
+  if (value.includes('e') || value.includes('E')) {
+    // Convert to decimal notation using the token's decimal precision
+    const num = Number(value);
```
SteP-n-s, line 17: if `value` is not a valid numeric string, you might end up getting `NaN` which will throw an error when the `.toFixed` gets applied. I think the following code should prevent this from happening. 

```
const normalizeAmount = (value: string, decimals: number): string => {
  // Check if the value is in scientific notation
  if (value.toLowerCase().includes('e')) {
    // Convert to decimal notation using the token's decimal precision
    const num = Number(value);
    // Do we want to return '0' for invalid numbers?
    if (isNaN(num)) {
      return '0';
    }
    return num.toFixed(decimals);
  }
  return value;
};
```

Also, not sure if we want to remove any trailing zeros after decimal point in the `return num.toFixed(decimals)` line. If yes, we can apply `.replace(/\.?0+$/, '');` at the end.

SteP-n-s (changes requested):

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
bfullam: Agreed on all points. Done 👌

SteP-n-s (approved):

PR #16155, opened 2025-06-05
fix: cherry-pick 7.47.0 include gas and gas price in PPOM requests (#15836)

digiwand:
Cherry-picks https://github.com/MetaMask/metamask-mobile/pull/15836

Delta: 
- app/components/Views/confirmations/utils/transaction.ts does not have addMMOriginatedTransaction method yet. missing +1 -1 from this
- app/lib/ppom/ppom-util.test.ts 
  - add TransactionMeta import
  - multiple line diffs using former ppomUtil.validateRequest params
- app/lib/ppom/ppom-util.ts 1 line delta which was a line with 1 bracket
   ```
      }
   ```

----

> Similar to https://github.com/MetaMask/metamask-extension/pull/33180 with variations
> 
> - Extract gas properties from TransactionMeta.
> - Continue to pass all request properties to PPOM (Tests are expecting these, and I don't see how this would block the work in this PR)
> - Use constant names > magic names

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5005

### DIFF
--- a/app/core/RPCMethods/eth_sendTransaction.ts
+++ b/app/core/RPCMethods/eth_sendTransaction.ts
@@ -114,7 +114,9 @@ async function eth_sendTransaction({
     origin: hostname,
   });
 
-  ppomUtil.validateRequest(req as PPOMRequest, transactionMeta?.id);
+  ppomUtil.validateRequest(req as PPOMRequest, {
+    transactionMeta,
+  });
 
   res.result = await result;
 }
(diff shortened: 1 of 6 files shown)


app/core/RPCMethods/eth_sendTransaction.ts
```diff
@@ -114,7 +114,9 @@ async function eth_sendTransaction({
     origin: hostname,
   });
 
-  ppomUtil.validateRequest(req as PPOMRequest, transactionMeta?.id);
+  ppomUtil.validateRequest(req as PPOMRequest, {
+    transactionMeta,
+  });
```
digiwand, line 119: same code. +2 extra lines from original.

NicolasMassart (approved):

PR #16011, opened 2025-06-03
fix: cp-7.47.0 add loader if account balance is not defined 

salimtb:
## **Description**

This PR adds a loading indicator while the balance is being fetched, instead of showing `undefined`.

## **Related issues**

Fixes: #15396 

## **Manual testing steps**

1. Go Account list
2. check the balance

**in order to reproduce , you can test on android and create a slow device emulator, then add a new Solana account**

### **Before**

Screenshot 2025-06-03 at 10 52 41

### **After**

Screenshot 2025-06-03 at 10 52 34

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #16165, opened 2025-06-06
chore: cherry-pick fix: add loader if account balance is not defined (#16011)

salimtb:
cherry pick fix: cp-7.47.0 add loader if account balance is not defined #16011


NicolasMassart:
Wrong branch name

PR #16238, opened 2025-06-10
fix: cp-7.49.0 Make `confirmation_redesign` feature flags work as true kill-switch

OGPoyraz:
## **Description**

Refactored the confirmation redesign feature flag selector to use a simpler "kill switch" approach instead of validation-based fallbacks.

### **Simplified feature flag logic:**

- Replaced validation and default fallback pattern with direct `remoteValues?.property !== false` checks
- Remote flags now act purely as "kill switches" - they can disable features when explicitly set to `false`

The logic now defaults to `true` for features when remote flags are not explicitly set to `false`, rather than falling back to predefined default values. This maintains the same kill switch functionality while simplifying the implementation.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5110

## **Manual testing steps**

N/A

### DIFF
--- a/app/selectors/featureFlagController/confirmations/index.ts
+++ b/app/selectors/featureFlagController/confirmations/index.ts
@@ -19,7 +19,7 @@ export type ConfirmationRedesignRemoteFlags = {
  * is set, the remote flag value takes precedence. If a remote flag is explicitly
  * set to `false`, it can disable the feature remotely.
  *
- * ## Adding New Confirmation Types
+ * ## Adding New Confirmation Flag
  *
  * **During Development:**
  * Use a local environment variable with a default fallback:
--- a/app/selectors/featureFlagController/confirmations/index.ts
+++ b/app/selectors/featureFlagController/confirmations/index.ts
@@ -30,7 +30,7 @@ export type ConfirmationRedesignRemoteFlags = {
  * );
  * ```
  *
- * **After Development (Release):**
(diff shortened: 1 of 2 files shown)


app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -12,52 +11,63 @@ export type ConfirmationRedesignRemoteFlags = {
...
+ * ```
+ *
+ * **After Development (Release):**
+ * Replace the fallback with the remote kill switch:
+ *
+ * ```
+ * const isNewConfirmationTypeEnabled = getFeatureFlagValue(
+ *   process.env.FEATURE_FLAG_REDESIGNED_NEW_CONFIRMATION_TYPE,
+ *   remoteValues?.new_confirmation_type !== false,
+ * );
```
jpuri, line 39: Should we not remove env variable after development ?

OGPoyraz: I think it still adds value for development since you can still set it locally. 

However - if we validated that certain type is working properly for a while - meaning that we want to remove old code, I think only in that scenario we can remove that. For example `signatures` - we can remove that entirely both from flags and env variable.

Updated JSDoc.

jpuri (approved):

matthewwalsh0 (approved):

PR #16285, opened 2025-06-11
feat: update account list to support grouping

PatrykLucka:
## **Description**

This PR adds account grouping in accounts menu using the newly added AccountTreeController. All the changes are hidden by a feature flag
## **Related issues**

Jira ticket: https://consensyssoftware.atlassian.net/browse/MUL-199

## **Manual testing steps**

1. Make sure multichain accounts feature flag is enabled
2. Open accounts menu
3. Accounts should be grouped

### **Before**
Screenshot 2025-06-11 at 17 13 45

### **After**

Screenshot 2025-06-11 at 17 35 50


app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.styles.ts
```diff
@@ -9,14 +10,38 @@ import { StyleSheet } from 'react-native';
...
+    sectionHeader: {
+      paddingHorizontal: 16,
+      paddingVertical: 8,
+      flexDirection: 'row',
+      justifyContent: 'space-between',
+      alignItems: 'center',
+      backgroundColor: theme.colors.background.default,
+    },
+    sectionHeaderText: {
+      fontSize: 14,
```
gantunesr: I had the impression that font size was already part of the DS, since you are using the `Text` component you can use the prop `variant` with `TextVariant`. But let me know if this is not be enough for the case you have here

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -222,7 +260,74 @@ const EvmAccountSelectorList = ({
     [navigate],
   );
 
-  const renderAccountItem: ListRenderItem<Account> = useCallback(
+  const renderSectionHeader = useCallback(
+    ({ section: { title } }: { section: AccountSection }) => (
+      <View style={styles.sectionHeader}>
+        <Text style={styles.sectionHeaderText}>{title}</Text>
+        <Text style={styles.sectionDetailsLink}>
+          Details
```
gantunesr: This should be part of the `locales/languages` for translation

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -222,7 +260,74 @@ const EvmAccountSelectorList = ({
...
+
+    if (selectedAccount) {
+      // Find the section and item index for the selected account
+      let sectionIndex = -1;
+      let itemIndex = -1;
+
+      for (let i = 0; i < accountSections.length; i++) {
+        const section = accountSections[i];
+        const accountIndex = section.data.findIndex(
+          (acc) => acc.address.toLowerCase() === selectedAccount.address.toLowerCase()
```
gantunesr: Lets use `toFormattedAddress` instead of `toLowerCase` here

```suggestion
          (acc) => toFormattedAddress(acc.address) === toFormattedAddress(selectedAccount.address)
```

app/core/Authentication/Authentication.ts
```diff
@@ -441,12 +444,20 @@ class AuthenticationService {
...
-      
+
+      const isMultichainAccountsEnabled = isMultichainAccountsEnabledForState1(
+        RemoteFeatureFlagController.state.remoteFeatureFlags.enableMultichainAccounts as Json & MultichainAccountsFeatureFlag
+      );
+
+      if (isMultichainAccountsEnabled) {
+        AccountTreeController.init();
+        await AccountsController.updateAccounts();
+      }
```
gantunesr: I see this is repeated, should it be its own method?

app/multichain-accounts/selectors/accountTreeController.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!accountTreeState?.accountTree?.wallets) {
+      return [{
+        title: 'Default Group',
+        data: [],
+      }];
+    }
+
+    return Object.values(accountTreeState.accountTree.wallets).map((wallet) => ({
+      title: wallet.metadata.name,
+      data: Object.values(wallet.groups).map((group) => group.accounts).flat(),
```
gantunesr: Why not use a `flatMap` instead?
```suggestion
      data: Object.values(groups).flatMap((group) => group.accounts),
```

app/multichain-accounts/selectors/accountTreeController.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param state - Root redux state
+ * @returns AccountTreeController state
+ */
+const selectAccountTreeControllerState = (state: RootState) => state.engine.backgroundState.AccountTreeController;
+
+/**
+ * Get account sections from AccountTreeController
+ * For now, this returns a simple structure until the controller is fully integrated
+ */
+export const selectAccountSections = createSelector(
```
gantunesr: Lets add some unit tests to these selectors

infiniteflower (dismissed):

app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts
```diff
@@ -88,3 +89,14 @@ export const selectMultichainAccountsState1Enabled = createMultichainAccountsSta
...
+/**
+ * Checks if multichain accounts are enabled for state 1.
+ * @param remoteFeatureFlags - The remote feature flags object containing the multichain accounts feature flag.
+ * @returns True if the multichain accounts feature is enabled for state 1, false otherwise.
+ */
+export const isMultichainAccountsEnabledForState1 = (remoteFeatureFlags: Json & MultichainAccountsFeatureFlag) => (
+  [FEATURE_VERSION_1, FEATURE_VERSION_2].some((featureVersion) =>
+    isMultichainAccountsFeatureEnabled(remoteFeatureFlags, featureVersion)
+  )
+);
```
gantunesr: I noticed this new method Patryk, why `selectMultichainAccountsState1Enabled` is not enough?

PatrykLucka: `selectMultichainAccountsState1Enabled` is UI selector, `isMultichainAccountsEnabledForState1` is used by that selector, but also by `initializeAccountTree` in `AuthenticationService`

app/core/Authentication/Authentication.ts
```diff
@@ -225,6 +227,21 @@ class AuthenticationService {
     };
   };
 
+  /**
+   * Initializes the account tree and updates the accounts if multichain accounts is enabled
+   */
+  private initializeAccountTree = async (): Promise<void> => {
```
owencraston: why is this needed?

app/core/Authentication/Authentication.ts
```diff
@@ -225,6 +227,21 @@ class AuthenticationService {
...
+    const { AccountTreeController, AccountsController, RemoteFeatureFlagController } = Engine.context;
+    const isMultichainAccountsEnabled = isMultichainAccountsEnabledForState1(
+      RemoteFeatureFlagController.state.remoteFeatureFlags.enableMultichainAccounts as Json & MultichainAccountsFeatureFlag
+    );
+
+    if (isMultichainAccountsEnabled) {
+      AccountTreeController.init();
```
owencraston: I think this should be put in the Engine.ts

app/multichain-accounts/selectors/accountTreeController.ts
owencraston: createDeepEqualSelector

app/multichain-accounts/selectors/accountTreeController.ts
```diff
@@ -0,0 +1,30 @@
...
+  [selectAccountTreeControllerState],
+  (accountTreeState) => {
+    if (!accountTreeState?.accountTree?.wallets) {
+      return [{
+        title: 'Default Group',
```
owencraston: is this really what we want to return? Would this not result in an unusable wallet?

app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts
owencraston: would we be able to remove this if we moved it into the engine?

app/core/Authentication/Authentication.ts
PatrykLucka: I don't know how we can do that, we need to have access to accounts, so the vault needs to be created and all accounts discovered in order to make it work.

app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts
PatrykLucka: I don't know if it is even possible, see: https://github.com/MetaMask/metamask-mobile/pull/16285#discussion_r2143165425

app/core/Authentication/Authentication.ts
PatrykLucka: state for AccountTreeController is not persisted, so we need to recalculate tree on every login

app/multichain-accounts/selectors/accountTreeController.ts
PatrykLucka: good question, it should never really happen, I think I just added it for easier testing 🤔  but maybe we should just throw an error in that case? wdyt?

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -91,13 +94,48 @@ const EvmAccountSelectorList = ({
     shallowEqual,
   );
 
+  const multichainAccountsState1Enabled = useSelector(selectMultichainAccountsState1Enabled);
+  const accountTreeSections = useSelector(selectAccountSections);
```
owencraston, line 92: why do we need two selectors for this? Can we not get the entire response from one selector call?

app/core/Authentication/Authentication.ts
owencraston: Can't we re calculate it in the engine? This is how we init all controllers and putting it in here feels like an anti pattern.

app/core/Authentication/Authentication.ts
owencraston: I am not sure I follow, I was under the impression that the accounts in the engine get initialized after login already? If not, how could we init the AccountsController if the vault was not created?

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
PatrykLucka: True! We can return null in `selectAccountSections ` when the `multichainAccountsState1Enabled` is false, so we will be able to only use one selector

app/core/Authentication/Authentication.ts
```diff
@@ -225,6 +227,21 @@ class AuthenticationService {
     };
   };
 
+  /**
+   * Initializes the account tree and updates the accounts if multichain accounts is enabled
+   */
+  private initializeAccountTree = async (): Promise<void> => {
+    const { AccountTreeController, AccountsController, RemoteFeatureFlagController } = Engine.context;
+    const isMultichainAccountsEnabled = isMultichainAccountsEnabledForState1(
+      RemoteFeatureFlagController.state.remoteFeatureFlags.enableMultichainAccounts as Json & MultichainAccountsFeatureFlag
```
gantunesr: I think if you use `assertMultichainAccountsFeatureFlagType`, you ensure type safety and avoid the type cast

app/core/Authentication/Authentication.ts
owencraston: After some discussion on slack, I understand why this cannot occur in the engine file.

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
owencraston: Sorry for not being clear, I was actually referring to combining the accountTreeSections and the internalAccounts.

owencraston: I am not sure how I feel about hiding away the multichainAccountsState1Enabled from the the  view since it might confused other devs. I dont feel strongly. cc @gantunesr

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -91,13 +93,44 @@ const EvmAccountSelectorList = ({
...
+  const internalAccounts = useSelector(selectInternalAccounts);
+
+  const accountSections = useMemo((): AccountSection[] => {
+    if (accountTreeSections) {
+      const accountsById = new Map<string, Account>();
+      internalAccounts.forEach((account) => {
+        const formattedAddress =
+          getFormattedAddressFromInternalAccount(account);
+        const accountObj = accounts.find(
+          (a) => a.address === formattedAddress,
```
owencraston: nit: we can use areAddressesEqual here.

app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts
owencraston: Honestly I think its cleaner to just move this logic into the auth class since we aren't going to reuse this anywhere else and having the selector here might lead to people missusing it.

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -222,7 +255,74 @@ const EvmAccountSelectorList = ({
...
+      if (sectionIndex !== -1 && itemIndex !== -1) {
+        // Add a small delay to ensure the list is rendered
+        setTimeout(() => {
+          accountListRef?.current?.scrollToLocation({
+            sectionIndex,
+            itemIndex,
+            animated: true,
+            viewPosition: 0.5, // Center the item in the view
+          });
+        }, 100);
```
ccharly: Feels a bit hacky to me. `useEffect` with no deps should be triggered after the component got mounted and since we already are using a `useEffect` (below) when calling the function this should be enough, no?

app/core/Authentication/Authentication.ts
```diff
@@ -441,12 +458,13 @@ class AuthenticationService {
...
       this.dispatchPasswordSet();
-      
+
       // Try to complete any pending Solana account discovery
       ///: BEGIN:ONLY_INCLUDE_IF(solana)
       this.retrySolanaDiscoveryIfPending();
       ///: END:ONLY_INCLUDE_IF
-      
+
+      this.initializeAccountTree();
```
ccharly: Given the other discussions happening on this function, I'd add a `TODO` to make sure we get back to this and try to update the logic.

```suggestion
      // TODO: Move this logic to the Engine.
      this.initializeAccountTree();
```

app/core/Authentication/Authentication.ts
```diff
@@ -495,12 +513,13 @@ class AuthenticationService {
...
```
ccharly:
```suggestion
      // TODO: Move this logic to the Engine.
      this.initializeAccountTree();
```

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
ccharly: Yes please, that's a new pattern we're trying to introduce and "generalize" in the codebase.

owencraston (commented):
Looking good

Simulator Screenshot - iPhone 16 Plus - 2025-06-13 at 16 02 48

owencraston (dismissed):

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
owencraston: I discussed with Patryk on slack and we decided this is better since the feature flag will be removed eventually which reduces the conflicts in the view.

infiniteflower (dismissed):

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -47,7 +49,13 @@ import { ACCOUNT_SELECTOR_LIST_TESTID } from './EvmAccountSelectorList.constants
...
+import { selectInternalAccounts } from '../../../selectors/accountsController';
+import { getFormattedAddressFromInternalAccount } from '../../../core/Multichain/utils';
+
+interface AccountSection {
+  title: string;
+  data: Account[];
+}
 
 /**
  * @deprecated This component is deprecated in favor of the CaipAccountSelectorList component.
```
vinnyhoward, line 56: Do these changes also need to be incorporated in the [CaipAccountSelectorList](https://github.com/MetaMask/metamask-mobile/tree/main/app/components/UI/CaipAccountSelectorList)? This component is being deprecated in favor of the new component

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
vinnyhoward: If `CaipAccountSelectorList` ends up being updated as well it would make a stronger case to move this into one selector

vinnyhoward (dismissed):
Approved but left some comments. I also pulled it down and tested. Works great💪

Since this is behind a feature flag I don't think they are blockers. Unless @owencraston disagrees. Main things I see are:

- `CaipAccountSelectorList` needs an update



- We are missing various E2E tests to confirm the new behavior using the different import flows

But since these changes are behind a feature flag I think this is good to go. We can add these changes in a different PR. We should create issues to track these updates @PatrykLucka if they haven't been created already

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
PatrykLucka: Right now we were focusing just on the account menu, but that’s a good point — keeping it consistent makes sense. Let’s leave that for the next steps.

app/multichain-accounts/controllers/account-tree-controller/constants.ts
```diff
@@ -0,0 +1,2 @@
+// Default AccountTreeController state
+export const defaultAccountTreeControllerState = {};
```
Cal-L: Replace this with `getDefaultAccountTreeControllerState` from the account tree controller

app/multichain-accounts/messengers/account-tree-controller-messenger/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param baseControllerMessenger - The base controller messenger.
+ * @returns The AccountTreeControllerMessenger.
+ */
+export function getAccountTreeControllerMessenger(
+  baseControllerMessenger: BaseControllerMessenger,
+) {
+  return baseControllerMessenger.getRestricted({
+    name: 'AccountTreeController',
+    allowedEvents: [
+      'AccountsController:stateChange',
```
Cal-L: `AccountsController:stateChange` and `KeyringController:stateChange` doesn't seem to be allowed events on the AccountTreeController - https://github.com/MetaMask/core/blob/main/packages/account-tree-controller/src/AccountTreeController.ts

app/multichain-accounts/messengers/account-tree-controller-messenger/index.ts
```diff
@@ -0,0 +1,28 @@
...
+      'AccountsController:accountAdded',
+      'AccountsController:accountRemoved',
+      'KeyringController:stateChange',
+    ],
+    allowedActions: [
```
Cal-L, line 18: `AccountsController:listAccounts`, `KeyringController:getAccounts` do not seem to be allowed actions on AccountTreeController. Also wanted to verify - Is `SnapController:get` -> SnapControllerGetSnap action or is there a typo?

app/multichain-accounts/messengers/account-tree-controller-messenger/index.ts
```diff
@@ -0,0 +1,28 @@
+import { BaseControllerMessenger } from '../../../core/Engine/types';
+
+/**
+ * Get the AccountTreeControllerMessenger for the AccountTreeController.
+ *
+ * @param baseControllerMessenger - The base controller messenger.
+ * @returns The AccountTreeControllerMessenger.
+ */
+export function getAccountTreeControllerMessenger(
```
Cal-L, line 9: On another note - Did this give you TS errors when applying non existent events and/or actions?

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -47,7 +49,13 @@ import { ACCOUNT_SELECTOR_LIST_TESTID } from './EvmAccountSelectorList.constants
 import { toHex } from '@metamask/controller-utils';
 import AccountNetworkIndicator from '../AccountNetworkIndicator';
 import { Skeleton } from '../../../component-library/components/Skeleton';
-import { selectMultichainAccountsState1Enabled } from '../../../selectors/featureFlagController/multichainAccounts';
+import { selectInternalAccounts } from '../../../selectors/accountsController';
+import { getFormattedAddressFromInternalAccount } from '../../../core/Multichain/utils';
+
+interface AccountSection {
```
Cal-L: Extract into types file

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -91,13 +94,44 @@ const EvmAccountSelectorList = ({
     shallowEqual,
   );
 
+  const accountTreeSections = useSelector(selectAccountSections);
+  const internalAccounts = useSelector(selectInternalAccounts);
+
+  const accountSections = useMemo((): AccountSection[] => {
```
Cal-L, line 95: While this is memoized, this logic looks computationally heavy. A few things that we'd want to check here to prevent performance regression:
- Verify how often the `accountSections` value actually changes
- The frequency of re-renders on this component"
- Worth measuring how long the function takes to execute - https://www.notion.so/metamask-consensys/MetaMask-Mobile-Debugging-18cf86d67d68801688d8e7b77bd0bbb2?source=copy_link#1e0f86d67d688059b0c6e7a3c6452746

app/core/Authentication/Authentication.ts
```diff
@@ -225,6 +233,34 @@ class AuthenticationService {
     };
   };
 
+  /**
+   * Checks if multichain accounts are enabled for state 1.
+   * @param remoteFeatureFlags - The remote feature flags object containing the multichain accounts feature flag.
+   * @returns True if the multichain accounts feature is enabled for state 1, false otherwise.
+   */
+  private isMultichainAccountsEnabledForState1 = (remoteFeatureFlags: Json & MultichainAccountsFeatureFlag) => (
```
Cal-L: Both methods `isMultichainAccountsEnabledForState1` and `initializeAccountTree`  doesn't seem unique to authentication service and should probably be abstracted elsewhere.

app/core/Authentication/Authentication.ts
```diff
@@ -495,12 +533,14 @@ class AuthenticationService {
...
-      
+
       // Try to complete any pending Solana account discovery
       ///: BEGIN:ONLY_INCLUDE_IF(solana)
       this.retrySolanaDiscoveryIfPending();
       ///: END:ONLY_INCLUDE_IF
-      
+
+      //TODO: Move this logic to the Engine when the account tree state will be persisted
+      this.initializeAccountTree();
```
Cal-L: To keep the Authentication service clean to only manage authentication logic, we can move this to be called in sagas instead. For example, we can place it in `authStateMachine`

Cal-L (commented):
Left some comments

GustavoRSSilva (dismissed):

app/multichain-accounts/messengers/account-tree-controller-messenger/index.ts
PatrykLucka: yes it did, it's handled by `GlobalEvents` type

app/multichain-accounts/messengers/account-tree-controller-messenger/index.ts
PatrykLucka: good catch, thank you!

app/multichain-accounts/messengers/account-tree-controller-messenger/index.ts
PatrykLucka: thanks, yes `SnapControllerGetSnap ` is `SnapController:get`

app/core/Authentication/Authentication.ts
PatrykLucka: yes, we have a TODO comment to move it to engine, probably at state 2, where we will have some kind of persisted state of `AccountsTreeController`. It's here just for state 1 of this feature.

app/core/Authentication/Authentication.ts
PatrykLucka: Oh, that's cool, I was looking for such thing, let me try to move it there now then!

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
PatrykLucka: thanks for your vigilance, just checked that, it looks like it renders three times every time the account menu is opened and each render takes about 0-2.5ms (each of these 3 initial renders takes a bit more time) for 2 wallets with total of 13 accounts.

owencraston (approved):

vinnyhoward (approved):

Cal-L (approved):
Platform LGTM

infiniteflower (approved):

app/components/UI/EvmAccountSelectorList/EvmAccountSelectorList.tsx
```diff
@@ -352,42 +445,19 @@ const EvmAccountSelectorList = ({
...
-      onContentSizeChange={onContentSizeChanged}
-      data={accounts}
+      sections={accountSections}
       keyExtractor={getKeyExtractor}
+      // @ts-expect-error - This will work, for some reason typescript sees null as an option here, but it's not.
       renderItem={renderAccountItem}
+      renderSectionHeader={accountTreeSections ? renderSectionHeader : undefined}
+      renderSectionFooter={accountTreeSections ? renderSectionFooter : undefined}
       // Increasing number of items at initial render fixes scroll issue.
       initialNumToRender={999}
```
Prithpal-Sooriya, line 462: Best to change this from a SectionList/Flatlist to a FlashList..
This field specifically breaks virtualisation (it will only virtualise items after 999 items are rendered)

PR #16293, opened 2025-06-11
feat: deprecate snap confirmation page

bfullam:
## **Description**
Deprecates the solana snap confirmation page. Swap and bridge transactions will now be confirmed automatically after submission.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to Swap/bridge
2. Get quote
3. Submit tx
4. See snap confirmation screen does not appear

### **After**

https://github.com/user-attachments/assets/3422998e-9349-4b4f-b2dc-9f101ae2b895


GustavoRSSilva (approved):

tommasini:
Heyyy I've ran the e2e here, I think it would be great to have a last green flag!

app/core/Engine/Engine.ts
```diff
@@ -1191,6 +1191,7 @@ export class Engine {
           'AccountsController:getAccountByAddress',
           'SnapController:handleRequest',
           'TransactionController:getState',
+          'RemoteFeatureFlagController:getState',
```
tommasini, line 1194: We should bring bridge controller to the new init controller pattern

tommasini (approved):

app/core/Engine/Engine.ts
bfullam: Agreed, will do in another PR!

app/components/UI/Bridge/Views/BridgeView/index.tsx
```diff
@@ -277,9 +279,23 @@ const BridgeView = () => {
...
+            navigation.navigate(Routes.BRIDGE.MODALS.ROOT, {
+              screen: Routes.BRIDGE.MODALS.BLOCKAID_MODAL,
+              params: {
+                errorType: isValidationError ? 'validation' : 'simulation',
+                errorMessage: isValidationError ? validationResult.result.validation.reason : validationResult.error,
+              },
+            });
+            return;
+          }
+          // TEMPORARY: If tx originates from Solana, navigate to transactions view BEFORE submitting the tx
```
amitabh94, line 297: Do we still need to do this ?

bfullam: In theory, no. In the past the behavior was very flaky though, so chose to not scope this change into this PR. Definitely worth addressing and testing as a fast follow, though!

amitabh94: Can we mark this as a TODO here ? I am wondering if removing this improves our experience now since we dont have snap anymore. Were you able to test this after removing this ?

app/util/bridge/hooks/useValidateBridgeTx.ts
amitabh94, line 1: Should we add a test for this hook to make sure we capture the changes in request/response if any ?
Also, is there a reason we are calling the endpoint directly here and not port to the controller ?

app/components/UI/Bridge/Views/BridgeView/index.tsx
bfullam: I'd say we just create a ticket for it in Jira vs marking as TODO in the code, no?

Definitely would improve the experience, I did test it a few times after removal and it worked but when I originally built it it worked for my 20-30 test transactions then failed and caused issues for Heyse in his testing.

amitabh94: You were still seeing issues after removing the snap page and not navigating to activity page?

app/util/bridge/hooks/useValidateBridgeTx.ts
bfullam: We can, though I'd also postpone that from this PR for time reasons. Mobile team has been really slammed lately and it already took a full week to get this approved as-is. Trying to avoid any code changes since that would remove the existing approval and potentially delay the PR much further.

This was done in the client for time reasons as well, faster than moving to the controller and having to deal with version bumps, etc.

app/components/UI/Bridge/Views/BridgeView/index.tsx
bfullam: I didn't see any issues in my tests, no, but I also didn't see any issues in my testing when we still had the snap and didn't navigate to the activity page. Because of that flaky behavior and the time constraints, I chose not to scope that so it can be addressed when we have adequate time to test thoroughly and make sure everything is consistent.

amitabh94: I let you be the best judge regarding this since you have worked closely on this but if time permits we should tackle this before releasing if we can . I believe it will massively improve user experience .
Lets start by creating a ticket as a follow up on this .

app/util/bridge/hooks/useValidateBridgeTx.ts
amitabh94: Lets create a ticket for this as well and track it a part of this user story then .

amitabh94 (approved):

PR #16317, opened 2025-06-12
fix: add traces

salimtb:
## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/components/UI/CollectibleContracts/index.js
```diff
@@ -415,6 +421,7 @@ const CollectibleContracts = ({
       const previousNfts = cloneDeep(
         NftController.state.allNfts[selectedAddress.toLowerCase()],
       );
+      trace({ name: TraceName.DetectNfts });
```
Prithpal-Sooriya, line 434: I really like this trace, nice work!

app/components/UI/CollectibleContracts/index.js
```diff
@@ -529,6 +537,11 @@ const CollectibleContracts = ({
     ],
   );
 
+  // End trace when component has finished initial loading
+  useEffect(() => {
+    endTrace({ name: TraceName.CollectibleContractsComponent });
+  }, []);
```
Prithpal-Sooriya: Not 100% sure about this trace, mostly because the start trace is called on every render, but the end trace is only called once on mount.

Maybe we can add a ref so it is only on first render?
```ts
const Component = () => {
  const isFirstRender = useRef(true)
  if (isFirstRender.current) {
    startTrace(...)
  }
  

  // Effect is ran at end of render
  useEffect(() => {
    endTrace(...);
    isFirstRender.current = false;
  }, [])
  
  // return JSX stuff
  return ( ... )
}
```

Prithpal-Sooriya: Hmm maybe this can be a reusable hook? Makes it easier to inject into our components.
```ts

/*
  This tracing hook only runs on first render.
*/
const useFirstRenderTrace = (traceName: TraceName) => {
  const isFirstRender = useRef(true);
  if (isFirstRender.current) {
    startTrace({ name: traceName })
  }

  // Effects run in between renders
  // Render -> Effects/Hooks -> Render -> ...
  // So after the first render is complete, this effect will run before the second render.
  useEffect(() => {
    if (isFirstRender.current) {
      endTrace({ name: traceName })
      isFirstRender.current = false;
    }
  }, []);
};

/// USAGE
const Component = () => {
  useFirstRenderTrace(TraceName.Component)
  
  // ... rest of logic
}

```

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -364,6 +373,10 @@ const NetworkSelector = () => {
     parentSpan,
   });
 
+  useEffect(() => {
```
Prithpal-Sooriya, line 376: see prev comment on the component tracing (specifically on using refs so the startTrace is only done on first render.

Prithpal-Sooriya (approved):

darkwing (approved):

PR #16321, opened 2025-06-12
feat: Filter out duplicate bridging transactions on activity list

pedronfigueiredo:
## **Description**

Implements filtering system for outgoing transactions with the same hash as incoming transactions.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/16202

## **Manual testing steps**

1. Go to this page...
2.
3.

### **Before**

See bug report ticket

### **After**

Screenshot 2025-06-12 at 11 47 26


app/components/UI/Transactions/index.js
```diff
@@ -772,6 +772,50 @@ class Transactions extends PureComponent {
...
+    for (let i = 0; i < transactions.length - 1; i++) {
+      const currentTx = transactions[i];
+
+      if (currentTx.type === 'incoming' && currentTx.txParams?.value) {
+        const nextTx = transactions[i + 1];
+
+        if (
+          nextTx.type === 'bridge' &&
+          nextTx.txParams?.value &&
+          currentTx.txParams.value === nextTx.txParams.value
```
OGPoyraz: Shouldn't we have to compare `hash` instead of value? Theoretically values can be same.

pedronfigueiredo: I assume hashes will be different, because the transaction data is different. I believe comparing the values and checking for the type is what we want in this case.

OGPoyraz (dismissed):

app/components/UI/Transactions/index.js
OGPoyraz: I think hashes are same as they reported in the task. Btw does bridge is happening in the portfolio dapp or is there a native experience in the app? I couldn't replicate "Bridge" transaction in the transaction list

pedronfigueiredo: I am being redirected to portfolio as well now, but it was working natively a minute ago

app/components/UI/Transactions/index.js
```diff
@@ -772,6 +772,50 @@ class Transactions extends PureComponent {
...
+    const incomingTransactionsToFilter = new Set();
+
+    for (let i = 0; i < transactions.length - 1; i++) {
+      const currentTx = transactions[i];
+
+      if (currentTx.type === 'incoming' && currentTx.txParams?.value) {
+        const nextTx = transactions[i + 1];
+
+        if (
+          nextTx.type === 'bridge' &&
```
matthewwalsh0: Why is this specific to bridge?

I assumed this was a generic issue impacting any incoming transactions generated by an outgoing transaction from the wallet as we keep both in the state.

What happens if we do a transfer to the same account for example, do we see two transactions?

matthewwalsh0 (changes requested):

app/components/UI/Transactions/index.js
pedronfigueiredo: Interesting point. The bug report specified bridge in particular, so I assumed that was the context for the bug.

I just tried sending eth to myself and it only shows one transaction

Screenshot 2025-06-12 at 14 00 17

matthewwalsh0: Thanks for testing Pedro.

Even if we're handling that scenario, what this error suggests is that the accounts API is retrieving the transaction as expected and adding it to the state (since we currently store the incoming and outgoing).

But the current transaction list selector isn't filtering transactions with duplicate hashes hence we see both.

Meaning we shouldn't need anything bridge specific, but just to filter the rendered transactions to unique hashes. But the remaining mystery then is why this is only impacting bridge transactions and we aren't seeing duplicates of every outgoing transaction on supported accounts API chains.

I think as a separate piece of work in future, we should stop storing the incoming transaction also (to save space) if it matches a local hash, and just update the client transaction list selectors to also include transactions to the account that don't have the `incoming` type.

To complicate things further, I've recently changed the controller logic to add other transaction types (send, transfer etc) on transactions retrieved from the accounts API.

pedronfigueiredo: I've conducted a few more tests with swaps, and simple send transactions and didn't find any other occurrence of the duplication. However, I've changed the implementation from bridge specific to generic for any transaction type that is not incoming.

app/components/UI/Transactions/utils.js
```diff
@@ -0,0 +1,41 @@
...
+   * @param {string} transactions[].type - The type of transaction (e.g.,
+   * 'bridge', 'incoming')
+   * @param {string} transactions[].id - Unique identifier for the transaction
+   * @param {Object} transactions[].txParams - Transaction parameters
+   * @param {string} transactions[].txParams.value - The transaction value in
+   * wei
+   * @returns {Array<Object>} Filtered array of transactions with redundant
+   * incoming transactions removed
+   */
+export const filterRedundantBridgeTransactions = (transactions) => {
```
OGPoyraz:
```suggestion
export const filterRedundantIncomingTransactions = (transactions) => {
```
Minor one

OGPoyraz (dismissed):

OGPoyraz (dismissed):

app/components/UI/Transactions/utils.js
matthewwalsh0, line 1: Minor, if we're adding a new file, should it be Typescript?

app/components/UI/Transactions/utils.js
```diff
@@ -0,0 +1,41 @@
...
+  if (!transactions || transactions.length === 0) {
+    return transactions;
+  }
+
+  const incomingTransactionsToFilter = new Set();
+
+  for (let i = 0; i < transactions.length - 1; i++) {
+    const currentTx = transactions[i];
+
+    if (currentTx.type === 'incoming' && currentTx.hash) {
```
matthewwalsh0: Do we need to be specific to `incoming`?

Should we not filter out all transactions with the same hash?

Outgoing transactions from the accounts API will have alternate types with the latest `TransactionController` version, so may also need filtering.

app/components/UI/Transactions/utils.js
```diff
@@ -0,0 +1,41 @@
...
+      const nextTx = transactions[i + 1];
```
matthewwalsh0: Do we care about whether it's the next transaction or not?

Could we simplify and just skip if `findIndex` with the `hash` returns an index lower than the current index?

The `time` on the original transaction from the wallet should be earlier as it's generated locally rather than using submission time, meaning it should come first in the transaction state.

app/components/UI/Transactions/index.js
```diff
@@ -756,12 +757,14 @@ class Transactions extends PureComponent {
...
+    const filteredConfirmedTransactions = filterDuplicateOutgoingTransactions(confirmedTransactions);
+
     const transactions =
       submittedTransactions && submittedTransactions.length
         ? submittedTransactions
             .sort((a, b) => b.time - a.time)
-            .concat(confirmedTransactions)
-        : this.props.transactions;
+            .concat(filteredConfirmedTransactions)
+        : filterDuplicateOutgoingTransactions(this.props.transactions);
```
matthewwalsh0: Minor, could we instead filter `transactions` once at the end?

matthewwalsh0 (changes requested):

app/components/UI/Transactions/utils.js
vinistevam: Done.

app/components/UI/Transactions/utils.js
vinistevam: Removed `incoming` specific filter.

app/components/UI/Transactions/utils.js
vinistevam: It makes sense, pushed the suggestion.

app/components/UI/Transactions/index.js
vinistevam: Sure, filtered at the end.

vinistevam (dismissed):

app/components/UI/Transactions/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return transactions.filter((currentTx, currentIndex) => {
+    if (!currentTx.hash) {
+      return true; // Keep transactions without a hash
+    }
+
+    // Check if a transaction with the same hash exists later in the array
+    const laterIndex = transactions.findIndex(
+      (tx, index) =>
+        tx.hash === currentTx.hash &&
```
matthewwalsh0: Minor, do we need to bother normalising these to lower case?

matthewwalsh0 (dismissed):

vinistevam (dismissed):

jpuri (dismissed):

app/components/UI/Transactions/utils.ts
vinistevam: Done. https://github.com/MetaMask/metamask-mobile/pull/16321/commits/2bdba074565a7fc306f25e08d7fb746ef7cb02ef

vinistevam (dismissed):

vinistevam (approved):

OGPoyraz (approved):

PR #16344, opened 2025-06-12
feat: Base Wallet Details Screen

owencraston:
## **Description**
This PR implements the basic wallet details screen. This feature is hidden behind a remote feature flag. This screen shows all the accounts that are tied to a wallet along with some other information...
- wallet name
- Total wallet balance (aggregate balances of all accounts in the wallet). This is not the aggregate balance of all your accounts, just the accounts that belong to the wallet
- List of accounts that belong to that wallet with...
    - the account name
    - account icon (Blockies or Jazzicon)
    - individual account balance

Designs: 

Screenshot 2025-06-17 at 9 43 37 PM

#### Out of scope:
- Social recovery 
- SRP
- Adding an account
These changes will be made in subsequent PRs. 

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUL-211

## **Manual testing steps**

1. This feature is hidden behind a remote feature flag so to enable it you will either need to use a proxy (message me for details) or have [this](https://github.com/MetaMask/metamask-mobile/blob/b425a933ac6645f700525dfab89a0edabd4c21fb/app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts#L70) function return true as well as [this one](https://github.com/MetaMask/metamask-mobile/blob/b425a933ac6645f700525dfab89a0edabd4c21fb/app/multichain-accounts/AccountTreeInitService/index.ts#L26).
2. Create a wallet
3. Open the account list.
4. Since the multi-chain accounts state 1 feature flag is enabled, you should see your accounts grouped by wallet
5. click on the blue details icon beside the wallet name
6. it should take you to the wallet details screen
7. this screen should contain your wallet name, total wallet balance, and a list of your accounts in that wallet with their individual balances. 
8. Click on an account should take you to that accounts details screen
9. clicking back from that screen should take you back to wallet details
10. Clicking back on the wallet details screen should take you back to the accounts list.

### **Before**

N/A

### **After**

https://github.com/user-attachments/assets/b7e49e6f-587f-4470-8435-8b0f27ed2114

<image src="https://github.com/user-attachments/assets/8262ae1b-ba98-4f2a-ae80-be6ba038f23c" height="500" width="250" />

### DIFF
--- a/app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
+++ b/app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
@@ -5,12 +5,6 @@ import { UseWalletBalancesHook } from './useWalletBalances.types';
 import { formatWithThreshold } from '../../../../../util/assets';
 import I18n from '../../../../../../locales/i18n';
 
-const formatCurrency = (value: number, currency: string) =>
-  formatWithThreshold(value, 0, I18n.locale, {
-    style: 'currency',
-    currency: currency.toUpperCase(),
-  });
-
 export const useWalletBalances = (
   accounts: InternalAccount[],
 ): UseWalletBalancesHook => {
@@ -39,17 +33,18 @@ export const useWalletBalances = (
   }, [accounts, multichainBalancesForAllAccounts]);
 
   const formattedWalletTotalBalance = useMemo(() => {
-    if (isLoading) {
+    if (isLoading || accounts.length === 0) {
       return undefined;
     }
 
     // Get displayCurrency from the first account (all accounts should have the same currency)
-    const firstAccountId = accounts[0]?.id;
-    const displayCurrency = firstAccountId
-      ? multichainBalancesForAllAccounts[firstAccountId]?.displayCurrency
-      : 'USD';
+    const displayCurrency =
+      multichainBalancesForAllAccounts[accounts[0].id]?.displayCurrency;
 
-    return formatCurrency(walletTotalBalance, displayCurrency);
+    return formatWithThreshold(walletTotalBalance, 0, I18n.locale, {
+      style: 'currency',
+      currency: displayCurrency.toUpperCase(),
+    });
   }, [
     isLoading,
     walletTotalBalance,
--- a/app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
+++ b/app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
@@ -41,7 +41,7 @@ export const useWalletBalances = (
     const displayCurrency =
       multichainBalancesForAllAccounts[accounts[0].id]?.displayCurrency;
 
-    return formatWithThreshold(walletTotalBalance, 0, I18n.locale, {
+    return formatWithThreshold(walletTotalBalance, 0.01, I18n.locale, {
       style: 'currency',
       currency: displayCurrency.toUpperCase(),
     });
--- a/app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
+++ b/app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
@@ -1,23 +1,21 @@
 import { useMemo } from 'react';
-import { useSelector } from 'react-redux';
 import useMultichainBalancesForAllAccounts from '../../../../hooks/useMultichainBalances/useMultichainBalancesForAllAccounts';
-import { selectCurrentCurrency } from '../../../../../selectors/currencyRateController';
 import { InternalAccount } from '@metamask/keyring-internal-api';
 import { UseWalletBalancesHook } from './useWalletBalances.types';
+import { formatWithThreshold } from '../../../../../util/assets';
+import I18n from '../../../../../../locales/i18n';
 
 const formatCurrency = (value: number, currency: string) =>
-  new Intl.NumberFormat(undefined, {
+  formatWithThreshold(value, 0, I18n.locale, {
     style: 'currency',
-    currency,
-    currencyDisplay: 'symbol',
-  }).format(value);
+    currency: currency.toUpperCase(),
+  });
 
 export const useWalletBalances = (
   accounts: InternalAccount[],
 ): UseWalletBalancesHook => {
   const { multichainBalancesForAllAccounts } =
     useMultichainBalancesForAllAccounts();
-  const currentCurrency = useSelector(selectCurrentCurrency);
 
   const isLoading = useMemo(
     () =>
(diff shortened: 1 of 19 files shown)


app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable arrow-body-style */
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
gantunesr: Is this needed?

app/multichain-accounts/selectors/accountTreeController.test.ts
```diff
@@ -88,16 +88,51 @@ describe('AccountTreeController Selectors', () => {
...
+              name: 'Wallet 2',
+            },
+          },
           data: [],
         },
       ]);
     });
 
-    it('returns null when wallets object is empty', () => {
+    it('returns null when wallets  is empty', () => {
```
gantunesr:
```suggestion
    it('returns null when wallets is empty', () => {
```

app/multichain-accounts/selectors/accountTreeController.ts
```diff
@@ -16,13 +18,19 @@ const selectAccountTreeControllerState = (state: RootState) => state.engine.back
 export const selectAccountSections = createDeepEqualSelector(
   [selectAccountTreeControllerState, selectMultichainAccountsState1Enabled],
   (accountTreeState, multichainAccountsState1Enabled) => {
-    if (!multichainAccountsState1Enabled || !accountTreeState?.accountTree?.wallets) {
+    if (
+      !multichainAccountsState1Enabled ||
+      !accountTreeState?.accountTree?.wallets
+    ) {
       return null;
```
gantunesr, line 25: I wonder if it is better to return an empty array here instead of `null`

app/components/Views/MultichainAccounts/WalletDetails/utils/getInternalAccountsFromWallet.ts
```diff
@@ -0,0 +1,29 @@
...
+  Object.values(wallet.groups).forEach((group) => {
+    accountIds.push(...group.accounts);
+  });
+
+  // Fetch internal accounts for each account ID
+  const internalAccounts = accountIds
+    .map((accountId) => accounts[accountId])
+    .filter((account): account is InternalAccount => account !== undefined);
+
+  return internalAccounts;
```
gantunesr: I think you can slightly improve performance this way

```suggestion
return Object.values(wallet.groups)
    .flatMap((group) => group.accounts)
    .map((accountId) => accounts[accountId])
    .filter((account): account is InternalAccount => account !== undefined);
```

app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
```diff
@@ -0,0 +1,64 @@
...
+  const formattedWalletTotalBalance = useMemo(() => {
+    if (isLoading) {
+      return undefined;
+    }
+
+    // Get displayCurrency from the first account (all accounts should have the same currency)
+    const firstAccountId = accounts[0]?.id;
+    const displayCurrency = firstAccountId
+      ? multichainBalancesForAllAccounts[firstAccountId]?.displayCurrency
+      : 'USD';
```
gantunesr: I wonder if there are some edge cases we're not considering here like an user that has their default currency set to ETH, BTC, or BCH

gantunesr (commented):
code looks good in general! will follow up with some manual testing

app/components/Views/MultichainAccounts/WalletDetails/utils/getInternalAccountsFromWallet.ts
owencraston: Very true.

app/selectors/featureFlagController/multichainAccounts/enabledMultichainAccounts.ts
owencraston: No, I needed it to add this to mock the feature flags. Good catch.

app/multichain-accounts/selectors/accountTreeController.ts
owencraston: The null return was part of [Patryks pr](https://github.com/MetaMask/metamask-mobile/pull/16285). The null return is used to conditionally render the accounts group (based on the feature flags). For this use case, I think the null return makes a lot of sense.

app/components/Views/MultichainAccounts/WalletDetails/hooks/useWalletBalances.ts
owencraston: Good point. this brings up a larger point that I think the wallet view balance also breaks in this case.

owencraston: Confirmed this edge case also breaks the wallet view. (which makes sense since we are leveraging the same logic)

<image src="https://github.com/user-attachments/assets/53d441c1-8b23-4678-a6ed-63e0f8d87760" height="500" width="250" />

owencraston: I have created an [issue](https://github.com/MetaMask/metamask-mobile/issues/16479) for this. I was able to remove the default value of USD.

owencraston: @gantunesr in fact my component is working when the users primary currency is a native token. The issue comes with really small balances that are less then 0.001 of the currency. This should be fixed in the above issue I created. 

<image src="https://github.com/user-attachments/assets/2eeafcdb-b733-4fed-8b7c-0282b6bf3371" height="500" width="250" />

owencraston: I was able to fix the small balance issue for the aggregate wallet balance but in order to fix this issue for each account, it would involve making lots of changes to the account list. This should be handled in [this issue](https://github.com/MetaMask/metamask-mobile/issues/16479).

<image src="https://github.com/user-attachments/assets/a322232d-165f-4971-8fe9-16cfef3314aa" height="500" width="250" />

app/multichain-accounts/selectors/accountTreeController.ts
gantunesr: An empty array and null are falsy so both could do the job, in extension it is used as an empty array. But nevertheless this is ok, no need to change it

gantunesr (commented):
Code looks good, will do some manual testing

MarioAslau (approved):
Looks good!

gantunesr:
I was not able to fully test the wallet details feature because of a few bugs that are not related to this PR, the first related to when you activate the feature flag for multichain accounts. I see that the list is empty and because of this bug I can't verify that the balances are properly calculated,

Screenshot 2025-06-19 at 9 07 47 PM

https://github.com/user-attachments/assets/bfed8b9f-65c2-416f-b66f-00e569ad2e60

The second bug is some UI/UX issues when looking into the account details, here's a video showing some of these issues,

https://github.com/user-attachments/assets/d4ba3514-80ae-4618-96d8-2596eb5b6712

gantunesr (approved):
The changes on this PR looks good but we have to fix the bugs related to the other features

PR #16352, opened 2025-06-13
feat(deposit): adds initial deposit processor

wachunei:
## **Description**

This PR adds the initial Deposit order processor. It includes creating a Native Ramps SDK instance to call the getOrder method.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/components/UI/Ramp/Aggregator/components/OrderListItem/OrderListItem.tsx
```diff
@@ -76,8 +78,14 @@ function OrderListItem({ order }: Props) {
   const styles = createStyles();
   const amount = getOrderAmount(order);
   const isBuy = order.orderType === 'BUY';
+  const isDeposit = order.provider === 'DEPOSIT';
```
georgeweiler, line 81: why isn't this `order.orderType === 'DEPOSIT'`?

app/components/UI/Ramp/Deposit/orderProcessor/index.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      id: 'test-id',
+      provider: 'DEPOSIT',
+      createdAt: 1673886669608,
+      amount: 123,
+      fee: 9,
+      cryptoAmount: 0.012361263,
+      cryptoFee: 9,
+      currency: 'USD',
+      currencySymbol: '',
+      cryptocurrency: 'BTC',
```
georgeweiler, line 73: small nit because it's test data, but can we change to ETH? btc isn't a supported deposit currency and does not match the network below

georgeweiler: It seems like these FiatOrders are all the same - maybe we could reduce dupes by abstracting.

app/components/UI/Ramp/Deposit/sdk/index.tsx
```diff
@@ -52,6 +52,8 @@ if (isDevelopmentOrInternalBuild) {
   environment = TransakEnvironment.Staging;
 }
 
+export const DepositSDKOrders = new NativeRampsSdk({}, environment);
```
georgeweiler, line 55: Maybe `DepositSDKUnauthorized` or something similar would be better?

app/components/UI/Ramp/utils/networkChainIdEquals.ts
```diff
@@ -0,0 +1,15 @@
+import { isCaipChainId } from '@metamask/utils';
+
+function networkChainIdEquals(network: string, chainId: string): boolean {
```
georgeweiler, line 3: do we need edgecase validation for things like:

- undefined, null, empty string
- invalid CAIP-2 format like 'eip155:' without a chain ID or whitespace

maybe isCaipChainId is enough?

app/components/UI/Ramp/Deposit/orderProcessor/index.ts
```diff
@@ -1,10 +1,86 @@
...
+  excludeFromPurchases: false,
+  orderType: depositOrder.orderType,
+  errorCount: 0,
+  lastTimeFetched: Date.now(),
+  data: depositOrder,
+});
 
 export async function processDepositOrder(
   order: FiatOrder,
   _options?: ProcessorOptions,
```
georgeweiler, line 60: I don't think this is used.

app/components/UI/Ramp/Deposit/orderProcessor/index.ts
```diff
@@ -1,10 +1,86 @@
...
+};
+
+export const depositOrderToFiatOrder = (depositOrder: DepositOrder) => ({
+  id: depositOrder.id,
+  provider: FIAT_ORDER_PROVIDERS.DEPOSIT,
+  createdAt: depositOrder.createdAt,
+  amount: depositOrder.fiatAmount,
+  fee: depositOrder.totalFeesFiat,
+  cryptoAmount: depositOrder.cryptoAmount || 0,
+  cryptoFee: depositOrder.totalFeesFiat || 0,
```
georgeweiler, line 43: what does `cryptoFee` represent? `totalFeesFiat` doesn't feel right.

app/components/UI/Ramp/Deposit/orderProcessor/index.ts
```diff
@@ -1,10 +1,86 @@
...
+  currency: depositOrder.fiatCurrency,
+  currencySymbol: '',
```
georgeweiler, line 45: is it possible to fill this value from the deposit order?

app/components/UI/Ramp/Aggregator/components/OrderListItem/OrderListItem.tsx
wachunei: the `FiatOrder['provider']` property is a "higher level" property that determines what _system_ is providing this order, before the aggregator it was Wyre, Transak, then it was Aggregator and now we will have Deposit (as in the _system_) https://github.com/MetaMask/metamask-mobile/blob/7d831738b4281875e6efffdcc8b15c38e14df724/app/components/UI/Ramp/orderProcessor/index.ts#L18-L23 

This is not using `orderType` because I want to target orders coming from this system (which now all of them have `orderType` with value `DEPOSIT`, so it could be swapped technically, but don't mean the same). The same way if the aggregator ever has a `DEPOSIT` `orderType`, with this code we don't want to target those orders.

app/components/UI/Ramp/Deposit/orderProcessor/index.test.ts
wachunei: > btc isn't a supported deposit currency and does not match the network below
This is testing that we don't care about the symbol and we pass it back verbatim. Probably there is a BTC token deployed there actually.

> It seems like these FiatOrders are all the same - maybe we could reduce dupes by abstracting.
You mean the expected objects, the one passed to `toEqual` function?

app/components/UI/Ramp/Deposit/sdk/index.tsx
wachunei: I was reluctant to use `Unauthorized` or similar since the only method we allow to be used is the `getOrder` method. Such name would make it seem like all of them will be unauthorized, but calling other methods without auth will fail.

app/components/UI/Ramp/utils/networkChainIdEquals.ts
wachunei: Good suggestions, not sure if isCaipChainId will cover these, will check.

app/components/UI/Ramp/Deposit/orderProcessor/index.ts
wachunei: This is not used now, they are part of the `FiatOrder` interface  though, and can be used later if we add exponential backoff to this processor like we did on the aggregator. The data will be used in the order details view 🔜

app/components/UI/Ramp/Deposit/orderProcessor/index.ts
wachunei: We will get an ISO code so it is up to us to use the formatting tools to render the symbol now. If we wanted we could fill that in on the API, however that would be a cross reference between the deposit system and the aggregator system, since we will be using aggregator data to fill in a symbol, and we are avoiding this by flattening somes nested properties.

app/components/UI/Ramp/utils/networkChainIdEquals.ts
wachunei: >✓ returns false when network is eip155 and chainId is 1

so the invalid format is covered by `isCaipChainId`.

I will add the falsy values cases and code

app/components/UI/Ramp/Deposit/orderProcessor/index.ts
wachunei: This one is coming from the aggregator, those orders only have totalFeesFiat field for both fee and crypto fee, as it just depends on the type of order. In this case we will never have a crypto fee so we can ignore it or remove this assignment and use `fee` only, leaving this as 0. What do you think?

AxelGes (approved):

PR #16367, opened 2025-06-13
feat: Add support for gas fees component for batch transactions

pedronfigueiredo:
## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5165

## **Manual testing steps**

1. Go to this page...
2.
3.

### **After**

Screenshot 2025-06-13 at 13 44 01

### DIFF
--- a/app/components/Views/confirmations/components/rows/transactions/gas-fee-details-row/gas-fee-details-row.tsx
+++ b/app/components/Views/confirmations/components/rows/transactions/gas-fee-details-row/gas-fee-details-row.tsx
@@ -1,7 +1,5 @@
 import React, { useState } from 'react';
 import { TouchableOpacity, View } from 'react-native';
-import { TransactionMeta } from '@metamask/transaction-controller';
-
 import { ConfirmationRowComponentIDs } from '../../../../../../../../e2e/selectors/Confirmation/ConfirmationView.selectors';
 import Icon, {
   IconSize,
@@ -21,6 +19,8 @@ import AlertRow from '../../../UI/info-row/alert-row';
 import { RowAlertKey } from '../../../UI/info-row/alert-row/constants';
 import { GasSpeed } from '../../../gas/gas-speed';
 import styleSheet from './gas-fee-details-row.styles';
+import { useTransactionBatchesMetadata } from '../../../../hooks/transactions/useTransactionBatchesMetadata';
+import { useFeeCalculationsTransactionBatch } from '../../../../hooks/gas/useFeeCalculationsTransactionBatch';
 
 
 const EstimationInfo = ({
@@ -74,9 +74,15 @@ const ClickableEstimationInfo = ({
 const GasFeesDetailsRow = ({ disableUpdate = false }) => {
   const [gasModalVisible, setGasModalVisible] = useState(false);
   const { styles } = useStyles(styleSheet, {});
+
   const transactionMetadata = useTransactionMetadataRequest();
+  const transactionBatchesMetadata = useTransactionBatchesMetadata();
+
   const feeCalculations = useFeeCalculations(
-    transactionMetadata as TransactionMeta,
+    transactionMetadata,
+  );
+  const feeCalculationsTransactionBatch = useFeeCalculationsTransactionBatch(
+    transactionBatchesMetadata,
   );
   const hideFiatForTestnet = useHideFiatForTestnet(
     transactionMetadata?.chainId,
@@ -104,13 +110,13 @@ const GasFeesDetailsRow = ({ disableUpdate = false }) => {
             {disableUpdate ? (
               <EstimationInfo
                 hideFiatForTestnet={hideFiatForTestnet}
-                feeCalculations={feeCalculations}
+                feeCalculations={transactionBatchesMetadata ? feeCalculationsTransactionBatch : feeCalculations}
               />
             ) : (
               <ClickableEstimationInfo
                 onPress={() => setGasModalVisible(true)}
                 hideFiatForTestnet={hideFiatForTestnet}
-                feeCalculations={feeCalculations}
+                feeCalculations={transactionBatchesMetadata ? feeCalculationsTransactionBatch : feeCalculations}
               />
             )}
           </View>
--- a/app/components/Views/confirmations/components/rows/transactions/gas-fee-details-row/gas-fee-details-row.tsx
+++ b/app/components/Views/confirmations/components/rows/transactions/gas-fee-details-row/gas-fee-details-row.tsx
@@ -1,27 +1,27 @@
+import { TransactionBatchMeta, TransactionMeta } from '@metamask/transaction-controller';
 import React, { useState } from 'react';
 import { TouchableOpacity, View } from 'react-native';
 import { ConfirmationRowComponentIDs } from '../../../../../../../../e2e/selectors/Confirmation/ConfirmationView.selectors';
-import Icon, {
-  IconSize,
-  IconName,
-} from '../../../../../../../component-library/components/Icons/Icon';
 import { strings } from '../../../../../../../../locales/i18n';
-import { TOOLTIP_TYPES } from '../../../../../../../core/Analytics/events/confirmations';
-import { useStyles } from '../../../../../../../component-library/hooks';
+import Icon, {
+  IconName,
+  IconSize,
+} from '../../../../../../../component-library/components/Icons/Icon';
 import Text from '../../../../../../../component-library/components/Texts/Text/Text';
+import { useStyles } from '../../../../../../../component-library/hooks';
+import { TOOLTIP_TYPES } from '../../../../../../../core/Analytics/events/confirmations';
 import useHideFiatForTestnet from '../../../../../../hooks/useHideFiatForTestnet';
 import { useFeeCalculations } from '../../../../hooks/gas/useFeeCalculations';
-import { useTransactionMetadataRequest } from '../../../../hooks/transactions/useTransactionMetadataRequest';
+import { useFeeCalculationsTransactionBatch } from '../../../../hooks/gas/useFeeCalculationsTransactionBatch';
 import { useConfirmationMetricEvents } from '../../../../hooks/metrics/useConfirmationMetricEvents';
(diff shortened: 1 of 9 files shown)


app/components/Views/confirmations/components/rows/transactions/gas-fee-details-row/gas-fee-details-row.tsx
```diff
@@ -104,13 +110,13 @@ const GasFeesDetailsRow = ({ disableUpdate = false }) => {
...
                 hideFiatForTestnet={hideFiatForTestnet}
-                feeCalculations={feeCalculations}
+                feeCalculations={transactionBatchesMetadata ? feeCalculationsTransactionBatch : feeCalculations}
               />
             ) : (
               <ClickableEstimationInfo
                 onPress={() => setGasModalVisible(true)}
                 hideFiatForTestnet={hideFiatForTestnet}
-                feeCalculations={feeCalculations}
+                feeCalculations={transactionBatchesMetadata ? feeCalculationsTransactionBatch : feeCalculations}
```
OGPoyraz: If this is not clickable in case of `transactionBatchesMetadata`, do we need to change this line?

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
```diff
@@ -1,146 +1,74 @@
...
+  networkClientId: '',
+  txParams: {
+    from: '0x0000000000000000000000000000000000000000' as Hex,
+    gasPrice: HEX_ZERO,
+  },
+} as TransactionMeta;
+
+export const useFeeCalculations = (transactionMeta?: TransactionMeta) => {
+  // Use a default transaction meta when undefined to satisfy hook requirements
+  const safeTransactionMeta = transactionMeta || DEFAULT_TRANSACTION_META;
```
OGPoyraz: I am wondering if you encountered an issue or an edge case to go with this approach? Technically this hook shouldn't be rendered if there is no `transactionMetadata`, will never be executed.

app/components/Views/confirmations/hooks/gas/feeCalculationsShared.ts
OGPoyraz, line 1: Can we relocate this file to `app/components/Views/confirmations/utils/gas.ts` since these are purely utility functions?

app/components/Views/confirmations/hooks/gas/useFeeCalculationsTransactionBatch.ts
OGPoyraz, line 1: Do we want to add any unit test for this new hook?

app/components/Views/confirmations/hooks/transactions/useSupportsEIP1559.ts
```diff
@@ -1,14 +1,13 @@
...
   TransactionMeta,
 } from '@metamask/transaction-controller';
 
 import { checkNetworkAndAccountSupports1559 } from '../../../../../selectors/networkController';
 import { RootState } from '../../../../../reducers';
 
-export function useSupportsEIP1559(transactionMeta: TransactionMeta | TransactionBatchMeta) {
+export function useSupportsEIP1559(transactionMeta: TransactionMeta) {
   const { networkClientId } = transactionMeta;
   const isLegacyTxn = (transactionMeta as TransactionMeta)?.txParams?.type === TransactionEnvelopeType.legacy;
```
OGPoyraz:
```suggestion
  const isLegacyTxn = transactionMeta?.txParams?.type === TransactionEnvelopeType.legacy;
```

I guess this is not needed anymore

app/components/Views/confirmations/hooks/transactions/useTransactionBatchSupportsEIP1559.ts
```diff
@@ -0,0 +1,20 @@
...
+  TransactionBatchMeta
+} from '@metamask/transaction-controller';
+import { useSelector } from 'react-redux';
+import { RootState } from '../../../../../reducers';
+import { checkNetworkAndAccountSupports1559 } from '../../../../../selectors/networkController';
+
+export function useTransactionBatchSupportsEIP1559(transactionBatchMeta: TransactionBatchMeta) {
+  const { networkClientId } = transactionBatchMeta;
+  const isLegacyTxn = transactionBatchMeta.gasFeeEstimates?.type === GasFeeEstimateType.Legacy ||
+    transactionBatchMeta.gasFeeEstimates?.type === GasFeeEstimateType.GasPrice;
```
OGPoyraz, line 12: `GasEstimation` type doesn't suggest if this transaction is legacy. It's all depend on envelope type but I am also  unsure what props to check here since we don't have specific `txParams` for batch. Does it make sense to check at least one sub transaction params inside this batch maybe?

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
OGPoyraz, line 1: I think the changes in this file breaking the fee calculations. 

See that "advanced" is selected but USD and ETH amounts are not same.

Screenshot 2025-06-20 at 15 25 34

OGPoyraz (changes requested):

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
```diff
@@ -1,146 +1,74 @@
...
 
-export const useFeeCalculations = (transactionMeta: TransactionMeta) => {
+// Default transaction meta for when undefined is passed
+const DEFAULT_TRANSACTION_META: TransactionMeta = {
+  id: 'default',
+  status: 'unapproved' as const,
+  time: Date.now(),
+  chainId: '0x1' as Hex,
+  gasLimitNoBuffer: HEX_ZERO,
+  networkClientId: '',
```
vinistevam: Shouldn't we set the `networkClientId` here? Otherwise, I believe it might cause issues when the `NetworkController` tries to use it later

app/components/Views/confirmations/hooks/gas/useFeeCalculationsTransactionBatch.ts
```diff
@@ -0,0 +1,135 @@
...
+const HEX_ZERO = '0x0';
+
+// Default batch meta for when undefined is passed
+const DEFAULT_BATCH_META: TransactionBatchMeta = {
+  id: 'default-batch',
+  from: '0x0000000000000000000000000000000000000000' as Hex,
+  status: TransactionStatus.unapproved,
+  chainId: '0x1' as Hex,
+  gas: HEX_ZERO,
+  networkClientId: '',
```
vinistevam: Same here

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
pedronfigueiredo: In React, hooks cannot render conditionally. There were other options, such as rendering a separate component with the separate hook when transactionMetadata was undefined, but that required more code duplication, so I opted for this approach.

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
pedronfigueiredo: This default value will have no consequence apart from the hook it is used in, and the UI will only display the value that does not come from the default. I don't think this networkClientId will reach the NetworkController.

app/components/Views/confirmations/hooks/transactions/useTransactionBatchSupportsEIP1559.ts
pedronfigueiredo: See [these types](https://github.com/MetaMask/core/blob/main/packages/transaction-controller/src/types.ts#L1249-L1275).

To your suggestion, I am not sure if the type of the nested transactions all need to be the same and if the batch transaction has the same type as them.

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
OGPoyraz: I agree - we cannot render hooks conditionally, we are also on the same page rendering separate component is another approach. However, we put complexity and unexpected calls (like calling controllers with unfilled info) if we try to satisfy all hooks with defaulting values like this. 


We are guarding these hooks to be executed here - by rendering `null` in `info-root.ts` if there is no `Approval.type`
https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/confirmations/components/info-root/info-root.tsx#L74

This is why I was asking if there is an edge case that you encounter. It should've been enough to check `Approval.type` exists but if not, my suggestion is to include checks for missing `transactionMeta` or `transactionBatchMeta` in `info-root` meaning no info component will be rendered and these hooks won't be executed so we don't have this "default values" implementation in this hook.

It happens fast enough but low hanging fruit is to render skeleton confirmation in future instead of returning `null` so we don't see empty page between screens.

app/components/Views/confirmations/hooks/transactions/useTransactionBatchSupportsEIP1559.ts
OGPoyraz: Those entities (the types you've linked) are created from the results from `GasFeeController`.

It doesn't mean that transaction is a legacy if gas estimate type is legacy. For example this is a possible scenario that envelope type is `0x2` (feemarket / EIP-1559 / non-legacy) but gas estimate type is `gasPrice` or `legacy`.

That logic is implemented here: https://github.com/MetaMask/core/blob/main/packages/transaction-controller/src/helpers/GasFeePoller.ts#L427

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
OGPoyraz: https://github.com/user-attachments/assets/f60a0732-c41c-413c-9d5c-8b3c9df3beb4

It appears that this issue is currently on `main` - we will look at this separately

OGPoyraz (approved):

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
pedronfigueiredo: I ended up removing these defaults

vinistevam (approved):

PR #14602, opened 2025-04-11
fix: Add delay in checkPermissions for iOS to prevent UI crash

abretonc7s:
## **Description**

This PR addresses a UI crash on iOS in `AccountConnect.tsx` caused by a race condition in `permissionsController.getPermissions()`.

- Added `await wait(100)` in `checkPermissions` for iOS to resolve the race condition.

#### Debugging Notes
This issue was tricky to debug as it occurred more frequently in live DEV environments. Eventually, I tracked it down by launching the app via Xcode instead of the regular React Native CLI, which provided access to all native logs and revealed the race condition.

## **Related issues**

Fixes:

## **Manual testing steps**

- Verified on iOS: UI crash no longer occurs.
- Confirmed no impact on Android.

### **Before**
image

### **After**
- normal connection flow

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #16460, opened 2025-06-18
fix: cp-7.50.0 align sdk connection with permission system

### LINKED ISSUE #16831: [Bug]: WalletConnect Error updating session for new chain
### Describe the bug

When switching to a chain that hasn't been permitted yet, we're displaying a new connection modal(*) and even approving that and switching chains the connection ends up being revoked.

(*) The reason for this second connection screen to occur is that we're currently saving permitted networks per hostname instead of per origin. There's a separate PR that fixes this: https://github.com/MetaMask/metamask-mobile/pull/16460

### Expected behavior

_No response_

### Screenshots/Recordings

https://github.com/user-attachments/assets/1729da79-a8e4-4270-8ed9-506bb4c88f75

### Steps to reproduce

1. Go to https://appkit-lab.reown.com/library/wagmi/ on your desktop browser
2. Connect -> Approve Mainnet 
3. Switch chain to OP -> see issue

```shell

```

### Detection stage

During release testing

### Version

7.50.0

### Build type

None

### Device

iPhone 11

### Operating system

Android, iOS

### Additional context

_No response_

### Severity

_No response_

elribonazo:
## **Description**

This PR updates the SDK connection flow to consistently use the latest permissions system UI, removing legacy components and streamlining the connection experience.

- Removed the legacy `AccountConnectSingle` component in favor of using the newer `PermissionsSummary` component consistently
- Eliminated the conditional rendering between old and new UI based on `isSdkUrlUnknown`
- Simplified icon handling and URL state management
- Improved callback handling with proper memoization using `useCallback`
- Consolidated connection confirmation logic into a single `handleConfirm` function
- Use channelIdOrHostname to fetch and store permissions instead of just hostname
- Fixed lots of Wallet Connect issues as part of this PR those addresses in the videos mainly, apart from the minor UI change

## **Related issues**

- Depends on #14602 

## **Manual testing steps**

- SDK connection requests from dapps
- WalletConnect connection flows
- Permission granting for both known and unknown origins
- Network switching during connection
- Multiple account selection flows

## **Screenshots/Recordings**
We first want to alignt eh AccountConnect.tsx component to the latest permission UI.

| Before       | After      |  
|--------------|--------------|
|  image | image | 

## Additional recordings
Switching networks with pre-approved networks (addEthereumNetwork + SwitchEthereumNetwork) VS only SwitchEthereumNetwork
| Without pre-approved networks  | with pre-approved networks     |  
|--------------|--------------|
|  <video src="https://github.com/user-attachments/assets/0e0eb6ec-7a89-441c-83be-fc634bb0e89a" /> | <video src="https://github.com/user-attachments/assets/1f6943ee-13f5-4f4c-b902-0594c278c1de" /> | 

| WC Session restore on cold start |  
|--------------|
|  <video src="https://github.com/user-attachments/assets/8e8967cb-1241-4040-9c53-de259076ff0b" /> |

### DIFF
--- a/app/components/Views/AccountConnect/AccountConnect.tsx
+++ b/app/components/Views/AccountConnect/AccountConnect.tsx
@@ -314,7 +314,7 @@ const AccountConnect = (props: AccountConnectProps) => {
         channelIdOrHostname.split(AppConstants.MM_SDK.SDK_REMOTE_ORIGIN)[1],
       ).origin;
     } else if (isOriginWalletConnect) {
-      title = channelIdOrHostname;
+      title = wc2Metadata.lastVerifiedUrl ?? wc2Metadata.url;
       dappHostname = title;
     } else if (!isChannelId && (dappUrl || channelIdOrHostname)) {
       title = prefixUrlWithProtocol(dappUrl || channelIdOrHostname);
--- a/app/components/Views/AccountConnect/AccountConnect.tsx
+++ b/app/components/Views/AccountConnect/AccountConnect.tsx
@@ -24,14 +24,13 @@ import {
 } from '../../../component-library/components/Toast';
 import { ToastOptions } from '../../../component-library/components/Toast/Toast.types';
 import { USER_INTENT } from '../../../constants/permissions';
-import { MetaMetricsEvents } from '../../../core/Analytics';
+import { MetaMetricsEvents, store } from '../../../core/Analytics';
 import Engine from '../../../core/Engine';
 import { selectAccountsLength } from '../../../selectors/accountTrackerController';
 import {
   InternalAccountWithCaipAccountId,
   selectInternalAccountsWithCaipAccountId,
 } from '../../../selectors/accountsController';
-import { isDefaultAccountName } from '../../../util/ENSUtils';
 import Logger from '../../../util/Logger';
 import {
   getAddressAccountType,
@@ -42,7 +41,7 @@ import {
   getUrlObj,
   prefixUrlWithProtocol,
 } from '../../../util/browser';
-import { Account, useAccounts } from '../../hooks/useAccounts';
+import { useAccounts } from '../../hooks/useAccounts';
 
 // Internal dependencies.
 import { PermissionsRequest } from '@metamask/permission-controller';
@@ -67,7 +66,6 @@ import {
   AccountConnectScreens,
 } from './AccountConnect.types';
 import AccountConnectMultiSelector from './AccountConnectMultiSelector';
-import AccountConnectSingle from './AccountConnectSingle';
 import AccountConnectSingleSelector from './AccountConnectSingleSelector';
 import { PermissionsSummaryProps } from '../../../components/UI/PermissionsSummary/PermissionsSummary.types';
 import PermissionsSummary from '../../../components/UI/PermissionsSummary';
@@ -77,7 +75,7 @@ import {
   AvatarSize,
   AvatarVariant,
 } from '../../../component-library/components/Avatars/Avatar';
-import { selectNetworkConfigurationsByCaipChainId } from '../../../selectors/networkController';
+import { selectEvmNetworkConfigurationsByChainId, selectNetworkConfigurationsByCaipChainId } from '../../../selectors/networkController';
 import { isUUID } from '../../../core/SDKConnect/utils/isUUID';
 import useOriginSource from '../../hooks/useOriginSource';
 import {
@@ -112,6 +110,10 @@ import { WalletClientType } from '../../../core/SnapKeyring/MultichainWalletSnap
 import AddNewAccount from '../AddNewAccount';
 import { InternalAccount } from '@metamask/keyring-internal-api';
 import { getApiAnalyticsProperties } from '../../../util/metrics/MultichainAPI/getApiAnalyticsProperties';
+import { PermissionKeys } from '../../../core/Permissions/specifications';
+import { CaveatTypes } from '../../../core/Permissions/constants';
+import { useNetworkInfo } from '../../../selectors/selectedNetworkController';
+import { formatChainIdToCaip } from '@metamask/bridge-controller';
 
 const AccountConnect = (props: AccountConnectProps) => {
   const { colors } = useTheme();
@@ -668,6 +670,8 @@ const AccountConnect = (props: AccountConnectProps) => {
     ],
   );
 
+  const {chainId} = useNetworkInfo(hostname);
+
   const handleNetworksSelected = useCallback(
     (newSelectedChainIds: CaipChainId[]) => {
       setSelectedChainIds(newSelectedChainIds);
@@ -676,8 +680,63 @@ const AccountConnect = (props: AccountConnectProps) => {
     [setScreen, setSelectedChainIds],
   );
 
-  const hideSheet = (callback?: () => void) =>
-    sheetRef?.current?.onCloseBottomSheet?.(callback);
+  const hideSheet = useCallback(
+    (callback?: () => void) =>
+      sheetRef?.current?.onCloseBottomSheet?.(callback),
+    [sheetRef],
+  );
+
+  const handleUpdateNetworkPermissions = useCallback(async () => {
+    let hasPermittedChains = false;
+    let chainsToPermit = selectedChainIds.length > 0 ? selectedChainIds : [];
+    if (chainId && chainsToPermit.length === 0) {
+      //This chainId is in format 0x1 but now we have caip25 format, eip155:1
+      chainsToPermit = [
+        formatChainIdToCaip(chainId)
+      ];
+    }
+
+    try {
+      hasPermittedChains = Engine.context.PermissionController.hasCaveat(
+        channelIdOrHostname,
+        PermissionKeys.permittedChains,
+        CaveatTypes.restrictNetworkSwitching,
+      );
+    } catch {
+      // noop
+    }
+
+    if (hasPermittedChains) {
+      Engine.context.PermissionController.updateCaveat(
+        channelIdOrHostname,
+        PermissionKeys.permittedChains,
+        CaveatTypes.restrictNetworkSwitching,
+        chainsToPermit,
+      );
+    } else {
+      Engine.context.PermissionController.grantPermissionsIncremental({
+        subject: {
+          origin: channelIdOrHostname,
+        },
+        approvedPermissions: {
+          [PermissionKeys.permittedChains]: {
+            caveats: [
+              {
+                type: CaveatTypes.restrictNetworkSwitching,
+                value: chainsToPermit,
+              },
+            ],
+          },
+        },
+      });
+    }
+  }, [selectedChainIds, chainId, hostname]);
+
+  const handleConfirm = useCallback(async () => {
+    hideSheet();
+    await handleUpdateNetworkPermissions();
+    await handleConnect();
+  }, [handleUpdateNetworkPermissions, hideSheet, handleConnect]);
 
   /**
    * User intent is set on AccountConnectSingle,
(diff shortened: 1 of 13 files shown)


elribonazo:
Removing globalNS tag, we will tackle that in a separate PR

elribonazo:
Update: tests are passing but the parseDeepLink will still cause an exception so this is still blocked

android/app/build.gradle
```diff
@@ -188,7 +188,7 @@ android {
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionName "7.47.1"
-        versionCode 2004
+        versionCode 2049
```
tommasini: Can you revert thsi change?

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -1342,7 +1342,7 @@
 				CODE_SIGN_ENTITLEMENTS = MetaMask/MetaMask.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 2004;
+				CURRENT_PROJECT_VERSION = 2049;
```
tommasini: Can you revert this changes?

bitrise.yml
```diff
@@ -2503,13 +2503,13 @@ app:
       VERSION_NAME: 7.47.1
     - opts:
         is_expand: false
-      VERSION_NUMBER: 2004
+      VERSION_NUMBER: 2049
```
tommasini: Can you revert this changes?

ios/MetaMask.xcodeproj/project.pbxproj
elribonazo: done

bitrise.yml
elribonazo: done thanks!

android/app/build.gradle
elribonazo: done

app/util/url/index.ts
```diff
@@ -75,3 +76,33 @@ export const toPunycodeURL = (urlString: string) => {
     return urlString;
   }
 };
+
+
+export const getHostname = (uri: string): string => {
```
tommasini: Can we create unit tests for this function?

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -355,11 +355,8 @@ export class BackgroundBridge extends EventEmitter {
       DevLogger.log(
         `notifySelectedAddressChanged: ${selectedAddress} channelId=${this.channelId} wc=${this.isWalletConnect} url=${this.url}`,
       );
-      if (this.isWalletConnect) {
-        approvedAccounts = getPermittedAccounts(this.url);
-      } else {
-        approvedAccounts = getPermittedAccounts(this.channelId);
-      }
+      approvedAccounts = getPermittedAccounts(this.origin);
```
jiexi, line 358: are we sure this is right? Because afaik, the json-rpc pipeline itself is still operating off hostname, so this would never result in any permitted accounts

elribonazo: We are moving away from storing in hostname as much as we possibly can, most of our code for WC is writing into channelId, the reason is to allow instances of the same Dapp connected for example on different browsers and keep different states.

cc: @christopherferreira9

app/util/url/url.test.ts
```diff
@@ -123,4 +124,79 @@ describe('URL Check Functions', () => {
...
+
+  describe('getHostname', () => {
+    it('returns hostname for standard HTTPS URLs', () => {
+      // Arrange
+      const url = 'https://www.google.com';
+
+      // Act
+      const result = getHostname(url);
+
+      // Assert
```
jiexi: do we need these comments?

app/util/url/url.test.ts
```diff
@@ -123,4 +124,79 @@ describe('URL Check Functions', () => {
...
+
+    it('returns original string for URIs without protocol separator', () => {
+      const plainString = 'example.com';
+      expect(getHostname(plainString)).toBe('example.com');
+    });
+
+    it('handles malformed URLs gracefully', () => {
+      const malformedUrl = 'https://';
+      const result = getHostname(malformedUrl);
+      expect(result).toBe('https');
```
jiexi: sure it doesn't make more sense to throw or return nothing in the case of malformed?

app/core/WalletConnect/wc-utils.ts
```diff
@@ -271,69 +291,27 @@ export const checkWCPermissions = async ({
...
+  if (verifyContext) {
+    const {verified} = verifyContext;
+    if (verified) {
+      return verified.origin;
     }
   }
-
-  return true;
-};
+  return defaultOrigin;
```
jiexi: would something like `return request.verifyContext?.verified?.origin ?? defaultOrigin` work? or should defaultOrigin only be used if `!request.verifyContext?.verified`?

app/core/WalletConnect/wc-utils.ts
```diff
@@ -250,14 +219,65 @@ export const onRequestUserApproval = (origin: string) => async (args: any) => {
...
-  origin,
-  caip2ChainId,
-  allowSwitchingToNewChain = false,
-}: { origin: string; caip2ChainId: CaipChainId; allowSwitchingToNewChain?: boolean }) => {
+export const isSwitchingChainRequest = (request: WalletKitTypes.SessionRequest) => {
+  const {params: {request: {method}}} = request;
+  return method === 'wallet_switchEthereumChain';
+}
+
+export const getChainIdForCaipChainId = (caipChainId: CaipChainId) => {
```
jiexi, line 256: I know we're doing this because we can't assume the caipChainid is EVM and do a toHex(), but it still feels like if that's the case we should enforce this at a higher level.

app/components/Views/AccountConnect/AccountConnect.tsx
```diff
@@ -317,14 +314,13 @@ const AccountConnect = (props: AccountConnectProps) => {
         channelIdOrHostname.split(AppConstants.MM_SDK.SDK_REMOTE_ORIGIN)[1],
       ).origin;
     } else if (isOriginWalletConnect) {
-      title = channelIdOrHostname;
+      title = wc2Metadata?.lastVerifiedUrl ?? wc2Metadata?.url ?? channelIdOrHostname;
```
jiexi, line 312: do these url values need to be formatted at all (changed to just hostname?)

app/components/Views/AccountConnect/AccountConnect.tsx
```diff
@@ -668,6 +641,8 @@ const AccountConnect = (props: AccountConnectProps) => {
     ],
   );
 
+  const {chainId} = useNetworkInfo(hostname);
```
jiexi: not channelIdOrHostname?

app/components/Views/AccountConnect/AccountConnect.tsx
```diff
@@ -676,8 +651,61 @@ const AccountConnect = (props: AccountConnectProps) => {
...
+      sheetRef?.current?.onCloseBottomSheet?.(callback),
+    [sheetRef],
+  );
+
+  const hasPermittedChains = useMemo(() => {
+    try {
+      return Engine.context.PermissionController.hasCaveat(
+        channelIdOrHostname,
+        PermissionKeys.permittedChains,
+        CaveatTypes.restrictNetworkSwitching,
```
jiexi: this will never be true. the permitted chains permission doesn't exist anymore. You need to get the `endowment:caip25` permission, `authorizedScopes` caveat, and check the permitted scopes on it.

app/components/Views/AccountConnect/AccountConnect.tsx
```diff
@@ -676,8 +651,61 @@ const AccountConnect = (props: AccountConnectProps) => {
...
+      });
+    }
+  }, [channelIdOrHostname, selectedChainIds, chainId, hasPermittedChains]);
+
+  const handleConfirm = useCallback(async () => {
+    hideSheet();
+    await handleConnect();
+    await handleUpdateNetworkPermissions();
+
+  }, [handleUpdateNetworkPermissions, hideSheet, handleConnect]);
```
jiexi: I'm not fully understanding why `handleConnect` doesn't already handle this case. Is this the scenario where there are existing permissions that need to be updated? I.e. permitting a new chain to switch to? If so there is different UI/approval flow for that

app/util/url/url.test.ts
elribonazo: done

app/util/url/url.test.ts
elribonazo: getHostname is used for WC and wc uris have weird malformed URL as wc:12346576432 which would lead to an invalid url if that makes sense. 

as we have
```typescript
it('returns protocol for protocol-based URIs', () => {
      const wcUri = 'wc:some-session-data';
      expect(getHostname(wcUri)).toBe('wc');
});
```
We can probably remove this test, I'll remove the test and improve jsdocs to be more specific on how this utility fn should be used

Done

app/core/WalletConnect/wc-utils.ts
elribonazo: done

app/core/WalletConnect/wc-utils.ts
elribonazo: Any proposal on where we could find a space for this fn? also, I'd be more in favor of keeping this as part of technical debt

app/components/Views/AccountConnect/AccountConnect.tsx
elribonazo: Thats a hotfix, as u see in first screenshot if we show channelIdOrHostname, it shows as: 

Screenshot 2025-07-04 at 09 25 57

lastVerifiedUrl, will show the correct url instead as shown here

Screenshot 2025-07-04 at 09 32 09

To not delay this further, If i may, I would propose to threat this as technical debt and get it fixed soon in a hotfix for 7.50.X

app/components/Views/AccountConnect/AccountConnect.tsx
elribonazo: Ok, done

app/util/url/url.test.ts
elribonazo: In favor of keeping the scope of this PR smaller and easier to review I've rolled back the changes for getHostname. Initially we had test coverage and this function already defined for this, in wc-utils. 

We can close this item

app/util/url/url.test.ts
elribonazo: https://github.com/MetaMask/metamask-mobile/pull/16460#discussion_r2184694396

app/components/Views/AccountConnect/AccountConnect.tsx
elribonazo: Reworked the hasPermitted chains fn to fetch the caip25 caveat and then check if there's active accounts after pairing with @wenfix

abretonc7s (approved):

tommasini (approved):

wenfix (approved):

app/util/url/index.ts
elribonazo: No longer applicable

app/core/WalletConnect/wc-utils.ts
elribonazo: Solving, will threat this as part of our technical debt. Create issue next week and sync with you on what would your proposal be.

app/core/BackgroundBridge/BackgroundBridge.js
wenfix:
```js
get origin() {
    return this.isWalletConnect || this.isMMSDK
      ? this.channelId
      : this.hostname;
  }
```
`origin` still points to `hostname` in case it's not WC or MM-SDK. Assuming `channelId` works for both this cases, it should be ok for us.

PR #16495, opened 2025-06-19
fix: cp-7.50.0 Fix `ApprovalTagUrl` exception for legacy confirmations

OGPoyraz:
## **Description**

This PR aim to fix couple small bits.

- Remove `from` prop from `ApprovalTagUrl` component as it has no relation with that information. This was occasionally causing some exceptions, see Sentry logs in the task.
- Remove `origin` row from contract interaction if transaction origin is app. (This is fixing a case for Earn team)
- Removed some unnecessary conditions where we already default to `contract interaction` confirmation.

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5198

## **Manual testing steps**

Exception happens rarely - so there is no repro steps.

### DIFF
--- a/app/components/UI/ApprovalTagUrl/ApprovalTagUrl.tsx
+++ b/app/components/UI/ApprovalTagUrl/ApprovalTagUrl.tsx
@@ -1,13 +1,9 @@
-import { toChecksumAddress } from 'ethereumjs-util';
 import React, { useEffect, useMemo, useState } from 'react';
-import { useSelector } from 'react-redux';
 
 import { strings } from '../../../../locales/i18n';
 import TagUrl from '../../../component-library/components/Tags/TagUrl';
 import { useStyles } from '../../../component-library/hooks';
 import AppConstants from '../../../core/AppConstants';
-import { selectInternalAccounts } from '../../../selectors/accountsController';
-import { selectAccountsByChainId } from '../../../selectors/accountTrackerController';
 import { getHost, prefixUrlWithProtocol } from '../../../util/browser';
 import useFavicon from '../../hooks/useFavicon/useFavicon';
 import stylesheet from './ApprovalTagUrl.styles';
@@ -18,7 +14,6 @@ export const APPROVAL_TAG_URL_ORIGIN_PILL = 'APPROVAL_TAG_URL_ORIGIN_PILL';
 
 interface ApprovalTagUrlProps {
   currentEnsName?: string;
-  from: string;
   origin?: string;
   sdkDappMetadata?: {
     url: string;
@@ -28,7 +23,6 @@ interface ApprovalTagUrlProps {
 }
 
 const ApprovalTagUrl = ({
-  from,
   origin,
   url,
   sdkDappMetadata,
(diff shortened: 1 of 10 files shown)


app/components/UI/ApprovalTagUrl/ApprovalTagUrl.tsx
```diff
@@ -37,17 +31,12 @@ const ApprovalTagUrl = ({
   const { styles } = useStyles(stylesheet, {});
   const [isOriginDeepLink, setIsOriginDeepLink] = useState(false);
 
-  const accountsByChainId = useSelector(selectAccountsByChainId);
-
-  const internalAccounts = useSelector(selectInternalAccounts);
-  const activeAddress = toChecksumAddress(from);
-
```
OGPoyraz, line 44: This component have nothing to do with `from` property or accounts. Perhaps it's been added historically and forgotten to remove. Some edge cases `from` passed as `undefined` - then `toChecksumAddress` is throwing an exception.

Hence doing the cleanup here to prevent this to happen, please see task for similar Sentry errors.

Screenshot 2025-06-19 at 09 15 41

app/components/Views/confirmations/components/title/title.tsx
```diff
@@ -125,7 +114,14 @@ const getTitleAndSubTitle = (
...
       }
-      return {};
+
+      // Default to contract interaction
+      return {
+        title: strings('confirm.title.contract_interaction'),
+        subTitle: isBatched
+          ? ''
+          : strings('confirm.sub_title.contract_interaction'),
+      };
```
OGPoyraz, line 124: As we default to contract interaction confirmation now - title needs to be adjusted too.

app/components/Views/confirmations/components/info/contract-interaction/contract-interaction.tsx
```diff
@@ -15,12 +16,13 @@ import SwitchAccountTypeInfoRow from '../../rows/switch-account-type-info-row';
 
 const ContractInteraction = () => {
   const transactionMetadata = useTransactionMetadataRequest();
-
   const { trackPageViewedEvent } = useConfirmationMetricEvents();
   const { isBatchedUpgrade } = use7702TransactionType();
 
   useEffect(trackPageViewedEvent, [trackPageViewedEvent]);
 
+  const isDappContractInteraction = transactionMetadata?.origin !== MMM_ORIGIN;
```
matthewwalsh0: If we're changing this, should we also hide if missing origin or origin equals `ORIGIN_METAMASK`?

app/components/Views/confirmations/components/info/contract-interaction/contract-interaction.tsx
```diff
@@ -30,7 +32,7 @@ const ContractInteraction = () => {
         enableMetrics={false}
         isTransactionsRedesign
       />
-      <OriginRow isSignatureRequest={false} />
+      {isDappContractInteraction && <OriginRow isSignatureRequest={false} />}
```
matthewwalsh0: Is there any value in putting this inside the `OriginRow` so we can re-use this criteria in other confirmation types?

matthewwalsh0 (changes requested):

jpuri (dismissed):
Changes look good 👍

app/components/Views/confirmations/components/info/contract-interaction/contract-interaction.tsx
OGPoyraz: In fact `ORIGIN_METAMASK` is `MMM_ORIGIN` (`metamask`) but I get what you're asking - there are also some app transactions using `Metamask Mobile` as origin. Created a utility for this.

app/components/Views/confirmations/components/info/contract-interaction/contract-interaction.tsx
OGPoyraz: I am happy for both approach, we might change it if we have consensus in the team. However this is not the convention we have it today. There are multiple info components work with rows this way.

Also in my opinion it's readable and clear when devs look at the info components and tell if row component will be rendered or not.

matthewwalsh0 (approved):

PR #16505, opened 2025-06-19
fix: app crash in certain conditions when `KeyringControllerState.keyrings` is `undefined`

mathieuartu:
## **Description**

~~This fixes a `useAutoSignIn` issue that could happen when `keyringControllerState.keyrings` does not exist yet.~~

This PR defaults the value returned by `selectKeyrings` to an empty array if `keyringControllerState.keyrings` is nullish.

This investigation started with `useAutoSignIn` throwing an error about `keyrings` not being an array. This original issue can be found [here](https://github.com/MetaMask/mobile-planning/issues/2214).
When this issue was fixed, we realized a new error popped up, not related to Identity flows (see image below), but seemingly with the same root cause.

This signifies a more profound error around `KeyringControllerState.keyrings` being `undefined`.
We do not know where the original issue originates from, and we will communicate around it, but in the meantime this PR ensures that we prevent developers from having this issue during the current testing period.

share_924596069652336176

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/2214
Related to: https://consensyssoftware.atlassian.net/browse/IDENTITY-138

## **Manual testing steps**

1. Go to the change password page
2. Enter current password
3. Before confirming on the next screen, close app
4. reopen the app
5. Verify that no crashes happened

### DIFF
--- a/app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
+++ b/app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
@@ -44,11 +44,15 @@ export function useAutoSignIn(): {
   const previousKeyringsLength = useRef(keyrings.length);
 
   useEffect(() => {
+    if (!keyrings || !Array.isArray(keyrings)) {
+      return;
+    }
+
     if (keyrings.length !== previousKeyringsLength.current) {
       previousKeyringsLength.current = keyrings.length;
       setHasNewKeyrings(true);
     }
-  }, [keyrings.length]);
+  }, [keyrings]);
 
   const areBasePrerequisitesMet = useMemo(
     () =>
--- a/app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
+++ b/app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
@@ -41,17 +41,14 @@ export function useAutoSignIn(): {
   const isSignedIn = useSelector(selectIsSignedIn);
 
   const keyrings = useSelector(selectKeyrings);
-  const previousKeyringsLength = useRef(keyrings?.length || 0);
+  const previousKeyringsLength = useRef(keyrings.length);
 
   useEffect(() => {
-    if (
-      Array.isArray(keyrings) &&
-      keyrings.length !== previousKeyringsLength.current
-    ) {
+    if (keyrings.length !== previousKeyringsLength.current) {
       previousKeyringsLength.current = keyrings.length;
       setHasNewKeyrings(true);
     }
-  }, [keyrings]);
+  }, [keyrings.length]);
 
   const areBasePrerequisitesMet = useMemo(
     () =>
--- a/app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
+++ b/app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
@@ -41,14 +41,13 @@ export function useAutoSignIn(): {
   const isSignedIn = useSelector(selectIsSignedIn);
 
   const keyrings = useSelector(selectKeyrings);
-  const previousKeyringsLength = useRef(keyrings.length);
+  const previousKeyringsLength = useRef(keyrings?.length || 0);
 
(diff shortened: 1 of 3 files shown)


app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
```diff
@@ -44,11 +44,15 @@ export function useAutoSignIn(): {
   const previousKeyringsLength = useRef(keyrings.length);
 
   useEffect(() => {
+    if (!keyrings || !Array.isArray(keyrings)) {
```
Tyschenko: Can we cover it with tests?

mathieuartu: Since it does nothing per se apart from acting as some kind of a type guard, and considering the sheer complexity of simulating adding keyrings in this already bloated unit test file, I'd advocate not adding a new test case for this condition.

Other arguments:
- We'll strip down `useAutoSignIn` soon (2 days window) to sign every user in, the only remaining condition being basic functionality being enabled or not.
- We're very likely to remove this "keyring watching" logic and attach a simple `performSignIn()` when importing a new SRP.

Tyschenko: Maybe you can implement a simple E2E test which will login to the app using fingerprint? Without your changes it should crash the app

image

mathieuartu: Added a fix with [3a64d54](https://github.com/MetaMask/metamask-mobile/pull/16505/commits/3a64d542173264c125c4ad71d6c1e09cefa304a9) and UT coverage with [63c99f8](https://github.com/MetaMask/metamask-mobile/pull/16505/commits/63c99f861f558afdb68a318a1598db2ce43fd9b5).

This was manually verified both on iOS and Android locally to be working and fixing the issue.

As we don't have ways to do biometrics based login set up right now in our E2E helpers, and there's no clear relationship between doing a biometrics login and `KeyringControllerState.keyrings` being `undefined`, we're covering this scenario with a UT.

cmd-ob: I have tested this manually and the UT is also covering this case. We can add a follow up to add some E2E tests too. I can also add this scenario to the QA regression pack for our RC validation. 

If an e2e test is needed, I think we can follow up as this PR is actually fixing this problem

cmd-ob (dismissed):
LGTM. Thanks for fixing @mathieuartu 
Looks like we might still need to handle the biometrics flow. 
The issue is fixed for the steps reported in the ticket.

mathieuartu:
Doing another E2E test run, while having reverted my original changes and added a fallback value (`[]`) for `selectKeyrings` if it is `undefined`, as we feel this is not a @MetaMask/identity  issue (`keyring` is of type `KeyringObject[]` and should not be `undefined`).

cc @cmd-ob @Prithpal-Sooriya @0x2kNJ

mathieuartu:
Also tagging @mikesposito . If the above comments holds true when tested by @cmd-ob when the builds are ready, then I think this is something you'd like to be aware of

cmd-ob:
> Doing another E2E test run, while having reverted my original changes and added a fallback value (`[]`) for `selectKeyrings` if it is `undefined`, as we feel this is not a @MetaMask/identity issue (`keyring` is of type `KeyringObject[]` and should not be `undefined`).
> 
> cc @cmd-ob @Prithpal-Sooriya @0x2kNJ

@mathieuartu I have not been able to reproduce the issue on the newest build.

cmd-ob (approved):

app/util/identity/hooks/useAuthentication/useAutoSignIn.ts
mathieuartu: Resolving this conversation as it turns out it was never the root cause of the problem.
cc @Tyschenko

PR #16512, opened 2025-06-19
fix: adjust ui entry point logic to work for diff flag variations

nickewansmith:
## **Description**
Lending Misc UI/Logic fixes

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/STAKE-1048

## **Manual testing steps**

- Pooled staking and Lending flags off
  - Pooled staking flow should work as expected and only unstake should be available with earnings data
  - There should be no Earn CTA button on ETH
  - Pooled staking inEligible and disabled switched handled in ui and not the earn selectors

## **Pre-merge author checklist**

- [ ] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #16577, opened 2025-06-23
chore: cherry-pick fix: adjust ui entry point logic to work for diff flag variations 

tommasini:
cherry-pick fix: adjust ui entry point logic to work for diff flag variations (#16512)

## **Description**

This PR aims to fix earn UI


NicolasMassart (approved):
Looks good to me: this cherry pick matches the original PR.

PR #16578, opened 2025-06-23
chore: update solana welcome modal/view

zone-live:
## **Description**

We are updating our Solana modal for the new Solana GTM strategy, that means to use the full view in order to have the desired look and feel.

Screenshot 2025-07-03 at 13 41 33

## **Related issues**

Fixes: N/A

## **Manual testing steps**

1. Create a new fresh wallet
2. After the onboarding the new Solana modal will appear

### **Before**

N/A

### **After**

solana_fox


app/components/Views/Wallet/index.tsx
```diff
@@ -341,6 +338,13 @@ const Wallet = ({
     isParticipatingInMetaMetrics,
     navigate,
   ]);
+///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
+  useEffect(() => {
+    navigate(Routes.SOLANA_NEW_FEATURE_CONTENT, {
+      screen: Routes.SOLANA_NEW_FEATURE_CONTENT,
+    });
+  }, [navigate]);
```
aganglada: every time we land here we navigate?

zone-live: Yes to show the new solana view and in there we have the logic to decide if it's meant to show it or not, to have things centralized.

zone-live: Thinking again, this can be made more efficient. Left me do a small change, to check here if we need to navigate beforehand instead of navigating straight away. 👍🏼

zone-live: Thank you!

aganglada (dismissed):

app/components/UI/SolanaNewFeatureContent/SolanaNewFeatureContent.styles.ts
```diff
@@ -1,43 +1,68 @@
...
+      paddingVertical: 30,
+    },
+    image: {
+      alignSelf: 'center',
+      width: Device.isLargeDevice() ? 300 : 220,
+    },
+    largeFoxWrapper: {
+      alignItems: 'center',
+      paddingTop: Device.isLargeDevice() ? 30 : 20,
+      paddingBottom: Device.isLargeDevice() ? 50 : 30,
```
Daniel-Cross: This is probably fine, but there is a hook you can use that will give dynamic sizing based on the screen size, which will also adapt to users who change their screen size dynamically, such as flip phones etc. It might not be needed but you might want to check it out instead of hard coding sizes.

`const { width, height } = useWindowDimensions();`

zone-live: Ah, I'll check! Ty Sir! 💪🏼

zone-live: Yeah both work, but I've updated the code to use the new `useWindowDimensions`  hook. Long term it's best to go with the more modern and flexible approach. Ty! 👍🏼

tommasini (dismissed):

app/components/Nav/App/App.tsx
```diff
@@ -617,6 +618,19 @@ const MultichainAccountDetails = () => {
   );
 };
 
+const SolanaNewFeatureContentView = () => (
+  <Stack.Navigator
+    screenOptions={{
+      headerShown: false,
+    }}
+  >
+    <Stack.Screen
```
tommasini, line 627: Shouldn't the modal be at `RootModalFlow` stack?

app/components/Views/Wallet/index.tsx
```diff
@@ -337,6 +337,23 @@ const Wallet = ({
...
+        SOLANA_FEATURE_MODAL_SHOWN,
+      );
+
+      if (hasSeenModal !== 'true') {
+        navigate(Routes.SOLANA_NEW_FEATURE_CONTENT);
+      }
+    };
+
+    checkAndNavigateToSolanaFeature();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tommasini: We should stop disabling react rules, since we want to benefit from react compiler, this will not allow react compiler to take effect and will create a burden to enable it completely in the future, can we implement it in a way that we do not need to disable a react rule?

e2e/specs/multichain/solana/wallet-invokeMethod.spec.ts
```diff
@@ -37,8 +40,15 @@ describe(SmokeMultiChainAPI('Solana - wallet_invokeMethod'), () => {
...
-        await SolanaNewFeatureSheet.tapCreateAccountButton();
+        await SolanaNewFeatureSheet.tapNotNowButton();
+
+        await WalletView.tapIdenticon();
+        await Assertions.checkIfVisible(AccountListBottomSheet.accountList);
+        await AccountListBottomSheet.tapAddAccountButton();
+        await TestHelpers.delay(4000);
+        await AddAccountBottomSheet.tapAddSolanaAccount();
         await AddNewHdAccountComponent.tapConfirm();
+        await TestHelpers.delay(4000);
```
tommasini, line 51: 4 seconds delays on each, they show definitely something is not behaving as smooth as expected. Since there is a big effort on performance, do we know why this 8 seconds delay are needed?

e2e/specs/solana/create-solana-account.spec.js
```diff
@@ -56,22 +56,24 @@ describe(SmokeNetworkExpansion('Create Solana account'), () => {
...
-    );
-    await SolanaNewFeatureSheet.tapCreateAccountButton();
+  it('should create a Solana account from the bottom sheet', async () => {
+    await SolanaNewFeatureSheet.tapNotNowButton();
+  
+    await WalletView.tapIdenticon();
+    await Assertions.checkIfVisible(AccountListBottomSheet.accountList);
+  
+    await AccountListBottomSheet.tapAddAccountButton();
+    await TestHelpers.delay(4000);
```
tommasini, line 66: We are adding another 4-second delay here, which leads me to believe that we are having performance regressions on adding an account, do you think this is the scenario?
If there is a performance issue when adding an account, can we raise it please?

app/components/Views/Wallet/index.tsx
zone-live: TS would always complain with the lint error, which is silly, it shouldn't be so strict. But it's needed.

app/components/Nav/App/App.tsx
zone-live: Its not a modal anymore.

e2e/specs/multichain/solana/wallet-invokeMethod.spec.ts
zone-live: I'll check it with the team to figure out the need for them, and will update them in a separate PR 👍🏼

e2e/specs/solana/create-solana-account.spec.js
zone-live: I'll check it with the team to figure out the need for them, and will update them in a separate PR 👍🏼

tommasini (approved):

PR #16581, opened 2025-06-23
feat: calculate solana gas requirements for swaps

bfullam:
## **Description**
Accounts for minimum Solana gas requirements when checking for insufficient balance during swap/bridge flow.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to Swaps
2. Select SOL as source token
3. Input current total SOL balance
4. See insufficient funds button and red balance text

### **After**
Screenshot 2025-06-23 at 14 25 04

### DIFF
--- a/app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
+++ b/app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
@@ -1,5 +1,8 @@
 import { useSelector } from 'react-redux';
-import { selectBridgeControllerState, selectMinSolBalance } from '../../../../../selectors/bridgeController';
+import {
+  selectBridgeControllerState,
+  selectMinSolBalance,
+} from '../../../../../selectors/bridgeController';
 import { useLatestBalance } from '../useLatestBalance';
 import { parseUnits } from 'ethers/lib/utils';
 import { BridgeToken } from '../../types';
@@ -25,7 +28,10 @@ const normalizeAmount = (value: string, decimals: number): string => {
   return value;
 };
 
-const useIsInsufficientBalance = ({ amount, token }: UseIsInsufficientBalanceParams): boolean => {
+const useIsInsufficientBalance = ({
+  amount,
+  token,
+}: UseIsInsufficientBalanceParams): boolean => {
   const { quoteRequest } = useSelector(selectBridgeControllerState);
   const minSolBalance = useSelector(selectMinSolBalance);
   const latestBalance = useLatestBalance({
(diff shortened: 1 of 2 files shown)


app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
```diff
@@ -38,20 +45,49 @@ const useIsInsufficientBalance = ({ amount, token }: UseIsInsufficientBalancePar
...
+    isNativeAddress(token.address);
+
+  let isInsufficientBalance = quoteRequest?.insufficientBal || false;
+
+  if (isSOL) {
+    // For SOL: check if balance - inputAmount >= minSolBalance (rent exemption)
+    const minSolBalanceLamports = parseUnits(minSolBalance, token.decimals);
+    const remainingBalance = latestBalance.atomicBalance.sub(inputAmount);
+    isInsufficientBalance =
+      isInsufficientBalance || remainingBalance.lt(minSolBalanceLamports);
```
SteP-n-s, line 85: Can you elaborate on why we need to override the `isInsufficientBalance` and not just reassign a new value based on the balance condition? Can we have a case where `isInsufficientBalance` is `true` and the check for remaining balance is `false`?

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
```diff
@@ -38,20 +45,49 @@ const useIsInsufficientBalance = ({ amount, token }: UseIsInsufficientBalancePar
...
+  } else {
+    // For non-SOL: just check if inputAmount > balance
+    isInsufficientBalance =
+      isInsufficientBalance || inputAmount.gt(latestBalance.atomicBalance);
```
SteP-n-s, line 89: In this instance, does the `isInsufficientBalance ` take into account whether user can also pay the network fees?

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
```diff
@@ -38,20 +45,49 @@ const useIsInsufficientBalance = ({ amount, token }: UseIsInsufficientBalancePar
...
+    token.decimals,
+  );
+  const isSOL =
+    token.chainId &&
+    isSolanaChainId(token.chainId) &&
+    isNativeAddress(token.address);
+
+  let isInsufficientBalance = quoteRequest?.insufficientBal || false;
+
+  if (isSOL) {
```
micaelae, line 80: Does this only run when the srcToken is native? We don't need the min balance check otherwise

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
bfullam: No, this was de-scoped for v1 due to time pressure. Can (and should) be included as a fast follow, though!

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
bfullam: Yes, isSol includes a native token check 👌

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
bfullam: This was included to account for non-evm and non-sol networks (in case there are any in the future). EVM network fees are accounted for in the `quoteRequest?.insufficientBal` check, SOL network fees are accounted for in the `isSol` block.

app/components/UI/Bridge/hooks/useInsufficientBalance/index.ts
bfullam: Implemented this way to leave room for possible implementation of `quoteRequest?.insufficientBal` for SOL quotes on the API side, similar to how the EVM balance checks are handled. My understanding is that we're currently leaning on the snap for these checks, but ideally we'll transition away from the snap entirely and have this all handled directly with the quote.

micaelae (approved):

PR #16654, opened 2025-06-25
fix: cp-7.50.0 Fix redesigned transaction confirmations after rejecting legacy transaction confirmation

### LINKED ISSUE #16653: [Bug]: Rejecting a legacy transaction confirmation breaks subsequent redesigned transaction confirmations
### Describe the bug

When user rejects a legacy transaction confirmation, subsequent redesigned transaction confirmations are rendered as legacy - causing app to almost crash. See recording below

### Reason

The underlying reason behind this is the [onComplete callback](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Approvals/TransactionApproval/TransactionApproval.tsx#L33) in `TransactionApproval` component.

When user rejects a request - `onComplete` callback is executed - however it fails to update `modalVisible` state because at the same time component gets re-rendered because of the `approvalRequest` changes, basically it causes a race condition.

### Solution

We can get rid of `modalVisible` entirely because accepting or rejecting confirmations will affect `approvalRequest` returned by `useApprovalRequest` then it will either render redesigned or legacy confirmations as first early return suggests.

### Expected behavior

Previous legacy confirmation rejects or approvals should not affect subsequent redesigned transaction confirmations

### Screenshots/Recordings

https://github.com/user-attachments/assets/d0bb9c17-ab26-4061-ba63-cb2eb85d543d

### Steps to reproduce

1. Trigger any kind of token approval
2. Go to send flow, pick destination and type amount
3. Final confirmation screen must be redesigned but it shows legacy confirmation

```shell

```

### Detection stage

During release testing

### Version

7.50.0 and main

### Build type

None

### Device

Both iOS and Android

### Operating system

iOS, Other (please elaborate in the "Additional Context" section)

### Additional context

_No response_

### Severity

_No response_

OGPoyraz:
## **Description**

When user rejects or approves a legacy transaction confirmation, subsequent redesigned transaction confirmations are rendered as legacy - causing app to almost crash. 

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/16653

## **Manual testing steps**

1. Trigger any kind of token approval
2. Go to send flow, pick destination and type amount
3. Final confirmation screen must be redesigned but it shows legacy confirmation

### **Before**

https://github.com/user-attachments/assets/2d354bb8-44d2-4cb8-9e7f-ce6bb13e7be3

### **After**

https://github.com/user-attachments/assets/e3934368-7d83-4e37-8a78-04bc722eef4b

### DIFF
--- a/app/components/Approvals/TransactionApproval/TransactionApproval.tsx
+++ b/app/components/Approvals/TransactionApproval/TransactionApproval.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback } from 'react';
 import useApprovalRequest from '../../Views/confirmations/hooks/useApprovalRequest';
 import { ApprovalTypes } from '../../../core/RPCMethods/RPCMethodMiddleware';
 import Approval from '../../Views/confirmations/legacy/Approval';
@@ -26,26 +26,18 @@ export interface TransactionApprovalProps {
 const TransactionApprovalInternal = (props: TransactionApprovalProps) => {
   const { approvalRequest } = useApprovalRequest();
   const { isRedesignedEnabled } = useConfirmationRedesignEnabled();
-  const [modalVisible, setModalVisible] = useState(false);
   const { onComplete: propsOnComplete } = props;
   const isQRSigning = props.isSigningQRObject && !props.transactionType;
 
   const onComplete = useCallback(() => {
-    setModalVisible(false);
     propsOnComplete();
   }, [propsOnComplete]);
 
   if (
-    (approvalRequest?.type !== ApprovalTypes.TRANSACTION &&
-      !modalVisible &&
-      !isQRSigning) ||
+    (approvalRequest?.type !== ApprovalTypes.TRANSACTION && !isQRSigning) ||
     isRedesignedEnabled
   ) {
     return null;
-  }
-
-  if (!modalVisible) {
-    setModalVisible(true);
   }
 
   if (props.transactionType === TransactionModalType.Dapp) {


app/components/Approvals/TransactionApproval/TransactionApproval.tsx
```diff
@@ -26,28 +26,20 @@ export interface TransactionApprovalProps {
 const TransactionApprovalInternal = (props: TransactionApprovalProps) => {
   const { approvalRequest } = useApprovalRequest();
   const { isRedesignedEnabled } = useConfirmationRedesignEnabled();
-  const [modalVisible, setModalVisible] = useState(false);
   const { onComplete: propsOnComplete } = props;
   const isQRSigning = props.isSigningQRObject && !props.transactionType;
 
   const onComplete = useCallback(() => {
-    setModalVisible(false);
```
matthewwalsh0, line 34: This state was originally needed to ensure the legacy modal remains open until `onComplete` was called, rather than when the approval request had gone.

Won't this break the legacy flow?

Why would this cause the redesign not to show if `isRedesignedEnabled` still causes it to return `null`?

matthewwalsh0 (changes requested):

app/components/Approvals/TransactionApproval/TransactionApproval.tsx
OGPoyraz: No it won't break legacy flow because we are checking `isRedesignedEnabled` below. In case if it's not redesigned - legacy will work as is.

The problem is after getting first legacy confirmation - `onComplete` cannot finish updating `modalVisible` state to `false` and subsequent transaction confirmation is not early returning because `modalVisible` is still settled to `true`. After brief amount of time `isRedesignedEnabled` is settling to `true` but it's too late because legacy is already mounted. (This is clearly can be seen the "Error" video in the PR description)

Please see recording below (on `main`) - that basically logs when `modalVisible` value is changed. Reject/approve of confirmation does not update this value.


https://github.com/user-attachments/assets/baf2d95c-f412-40dc-bb6a-0ecabd2440ee

jpuri: I am wondering if this code could be reason for the issue
```
  if (!modalVisible) {
    setModalVisible(true);
  }
```

OGPoyraz: Exactly @jpuri - that settles it to `true` but never goes back to `false`.

But then if we remove that - it's unnecessary to have this state. Maybe there was a historical reason to have this in the `TransactionApproval`. But I don't see any meaningful reason to have it right now

matthewwalsh0 (approved):

PR #16698, opened 2025-06-26
chore: add performance tracing for earn experiences

ameliejyc:
## **Description**

This PR adds Sentry tracing for 5 high and medium priority Earn metrics following the guidelines for [Performance Tracing](https://github.com/MetaMask/contributor-docs/blob/main/docs/performance-tracing.md): 

1. Deposit Screen
2. Withdraw Screen
3. Earnings component (not currently in use)
4. FAQ
5. FAQ APYs

A full list of the Earn performance benchmarks can be found [here](https://docs.google.com/document/d/1-mq98T39GhHIjkrGteIb99P7YLU3yHXSK8kaRuJTTJU/edit?tab=t.0#heading=h.l1kgno957cpf).

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-574

## **Manual testing steps**

1. First get permissions to https://metamask.sentry.io/ and the metamask team within that
2. Ensure you have the `MM_SENTRY_DSN_DEV` key in your `.js.env`
3. Check out this branch and run the app on iPhone. Ensure metametrics are enabled.
6. Try various ways to Deposit, e.g. clicking the 'Earn' button from the token list, or the 'Earn' button within the token detail view for ETH and earn tokens. A full list of supported tokens is [here](https://consensyssoftware.atlassian.net/wiki/spaces/PUK/pages/399765831835/Tokens+networks+support)
7. See on the console that Sentry Logger is called for the Earn Deposit Screen transaction
8. See on the Sentry [dashboard](https://metamask.sentry.io/dashboard/137081/) for Earn Deposit Screen (iOS) that a new trace has appeared
9. Try to withdraw an aToken by clicking into the token detail view and pressing 'Withdraw'
10. See on the console that Sentry Logger is called for the Earn Withdraw Screen transaction
11. See on the Sentry [dashboard](https://metamask.sentry.io/dashboard/137081/) for Earn Withdraw Screen (iOS) that a new trace has appeared
12. ~Click into an aToken and wait for the Earnings component to be visible~
13. ~See on the console that Sentry Logger is called for the Earn Earnings transaction~
14. ~See on the Sentry [dashboard](https://metamask.sentry.io/dashboard/137081/) for Earn Earnings (iOS) that a new trace has appeared~
15. On the Deposit page for Lending press the FAQ button
16. See on the console that Sentry Logger is called for Earn FAQ and Earn FAQ APYs transactions
17. See on the Sentry [dashboard](https://metamask.sentry.io/dashboard/137081/) for Earn FAQ (iOS) and Earn FAQ APYs that new traces have appeared

### After

Expect to see these graphs in Sentry:
image


ameliejyc:
> ### Bug: Duplicate Trace IDs Cause Navigation Conflicts
> Multiple `trace` calls for `TraceName.EarnDepositScreen` and `TraceName.EarnWithdrawScreen` from various entry points (e.g., `EarnTokenList`) use the same default ID. This causes concurrent traces to overwrite each other in the internal trace storage Map when users rapidly navigate, leading to lost or hanging traces and inaccurate performance measurements.
> 
> `app/components/UI/Earn/components/EarnTokenList/index.tsx#L130-L141`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmUyYWUwODc1LTA0OTctNGQzZC1iMWE0LWQyYzJhZjNlMTAwMiIsImVuY3J5cHRpb25LZXkiOiJ0Z2gxeC1aZnpzRENlaGROWnFlek85SWhCNkRiZVYtTy13SUlyU2NNUkhjIiwiYnJhbmNoIjoiVEFULTU3NC1tbS1tb2JpbGUtYWRkLXBlcmZvcm1hbmNlLXRyYWNpbmctZm9yLWVhcm4tZXhwZXJpZW5jZXMifSwiaWF0IjoxNzUxNTM4NTI1LCJleHAiOjE3NTIxNDMzMjV9.l74vVRZfBKngOhihkBwdvHRrEDybp6gZJViG4MAw2ybLk1TnRP-284j2PZg8Vfd7k5-kbJSWyNaw7SJwMAjB9s0ksWJDDJxi8xjpC4br_wwCWZWGSElo71X-rhGWzEOmTLssgNKs28lLzycizCk3C7c3uI62bm5LMzjsvecL7ZN2bWTzYOiu5v4b8m1xqkfERZr4XE7HCPIN0gKWEidWRR0g-iNYdNDW5RbYc4cFNIwTvJE0g3QzmW9_H4uLo3BD_qEwKESPC33qD1ifKu97RdKcaxtAYMYrMol0xaoRzhO3ULsm6vKBHbO142ZngJ3CGn8WHHq1eQ0cdyJve0TE4Q)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

Given these two traces measure simple navigation page loads I'm not sure we'd encounter this concurrency issue where users are able to go forward and back between pages quick enough to overwrite existing traces. In the case that they do, some trace data might be dropped though I think we have enough users that this isn't an issue. In the case that the trace is left hanging, it times out by default at 5 minutes and I haven't seen this occur for these metrics on the dashboard (and this could happen for any trace regardless). It would not affect the metric we care about (p75) to have outlier hanging traces. 

If we did want to address this we could add an `id` to `trace` calls. Then we'd need to use the same `id` in the `endTrace` which would be difficult as they live in different components. So one solution would be to search for and end _all_ traces with that name.

app/components/UI/Earn/LendingLearnMoreModal/index.tsx
```diff
@@ -148,6 +149,11 @@ export const LendingLearnMoreModal = () => {
     sheetRef.current?.onCloseBottomSheet();
   };
 
+  useEffect(() => {
+    trace({ name: TraceName.EarnFaqApys });
```
Matt561: Thinking out loud: Some components (e.g. EarnInputView) support both pooled-staking and lending experiences while others (e.g. PoolStakingLearnMoreModal) are experience-specific. For experience-specific components, would it make more sense to use a different `TraceName`? Will it be more difficult to track down where performance issues are happening if we have the same `TraceName` in different components?

app/components/hooks/useEndTraceOnMount.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from 'react';
+import { endTrace, TraceName } from '../../util/trace';
+
+/**
+ * Custom hook that ends a performance trace when the component first mounts
+ *
+ * @param traceName - The name of the trace to end
+ */
+
+export const useEndTraceOnMount = (traceName: TraceName) => {
```
Matt561, line 10: Nice!

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -182,6 +186,7 @@ const EarnInputView = () => {
     }
 
     if (tokenExperience === EARN_EXPERIENCES.STABLECOIN_LENDING) {
+      trace({ name: TraceName.EarnFaq });
```
Matt561: Do we want to include `trace({ name: TraceName.EarnFaq })` for the pooled-staking experience above?

app/components/UI/Earn/LendingLearnMoreModal/index.tsx
Matt561: I don't have hands-on experience with tracing in MM clients and curious to know the following:
- What happens if a trace is opened but never closed? Does it timeout eventually?
- What happens if we call `endTrace` but a trace was never opened? Is it handled safely/ignored?.

I'm thinking about these cases because a trace often starts in one file and ends in another. Just want to make sure everything is okay If we accidentally forget to end a trace or end one that was never opened.

I can see it being easy to break tracing when resolving merge conflicts or implementing suggestions from an AI coding agent since tracing is 1 more thing to keep in mind when building. 😄

Matt561 (dismissed):
Approving now so you're not blocked. The comments are mostly questions and minor suggestions 👍

app/components/UI/Earn/LendingLearnMoreModal/index.tsx
ameliejyc: Yeah that's a great shout. We could add in a `data` field to the trace which gives more contextual info, or use a different trace name altogether. I'm more in favour of the former to match our segment analytics practice of utilising metadata properties rather than adding endless different metric names (assuming I can easily filter by the property in the Sentry dashboard). I'll look into it.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
ameliejyc: I must have missed that - yes!

app/components/UI/Earn/LendingLearnMoreModal/index.tsx
ameliejyc: Great questions

> What happens if a trace is opened but never closed? Does it timeout eventually?

There is a default timeout of 5 minutes. In our performance benchmarks we care about the p75 so hopefully any timeout outliers don't affect us. 

`export const TRACES_CLEANUP_INTERVAL = 5 * 60 * 1000; // 5 minutes`

> What happens if we call endTrace but a trace was never opened? Is it handled safely/ignored?.

Yes, in the trace code `endTrace` looks for the a trace based on the name provided and if it doesn't find one it simply returns:
```
export function endTrace(request: EndTraceRequest) {
  const { name, timestamp } = request;
  const id = getTraceId(request);
  const key = getTraceKey(request);
  const pendingTrace = tracesByKey.get(key);

  if (!pendingTrace) {
    log('No pending trace found', name, id);
    return;
  }

  pendingTrace.end(timestamp);

  clearTimeout(pendingTrace.timeoutId);
  tracesByKey.delete(key);

  const { request: pendingRequest, startTime } = pendingTrace;
  const endTime = timestamp ?? getPerformanceTimestamp();
  const duration = endTime - startTime;

  log('Finished trace', name, id, duration, { request: pendingRequest });
}
```

ameliejyc: Added a data field to the trace with `experience` metadata of type `EARN_EXPERIENCES` for all the FAQ traces and verified I can see it in Sentry.

Matt561: Thanks for the info! 😄

Matt561 (approved):
LGTM! 👍

PR #16712, opened 2025-06-26
feat: Process batched transaction with single nested transaction as regular transaction

jpuri:
## **Description**

Process batched transaction with single nested transaction as regular transaction

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5259

## **Manual testing steps**

1. Submit batched transaction with single nested transaction
2. Check that it is treated as regular confirmation

## **Screenshots/Recordings**
TODO


app/core/RPCMethods/eip5792.ts
```diff
@@ -298,6 +181,255 @@ export async function getCapabilities(address: Hex, chainIds?: Hex[]) {
...
+  const requiredCallCapabilities = calls.flatMap((call) =>
+    Object.keys(call.capabilities ?? {}).filter(
+      (name) => call.capabilities?.[name].optional !== true,
+    ),
+  );
+
+  const requiredCapabilities = [
+    ...requiredTopLevelCapabilities,
+    ...requiredCallCapabilities,
+  ];
```
jiexi, line 259: does it matter if this result has non unique values?

app/core/RPCMethods/eip5792.ts
```diff
@@ -298,6 +181,255 @@ export async function getCapabilities(address: Hex, chainIds?: Hex[]) {
...
+async function validateSendCalls(
+  sendCalls: SendCalls,
+  networkClientId: string,
+  keyringType?: KeyringTypes,
+) {
+  const { AccountsController, TransactionController } = Engine.context;
+  const { from: paramFrom, chainId } = sendCalls;
+
+  const from =
+    paramFrom ?? (AccountsController.getSelectedAccount()?.address as Hex);
```
jiexi, line 396: should this also consider if the account is permitted or not for the dapp? or is this validated somewhere upstream of this function?

jpuri: yep its validated at other place

app/core/RPCMethods/eip5792.ts
```diff
@@ -381,6 +263,255 @@ export async function getCapabilities(address: Hex, chainIds?: Hex[]) {
...
+      `EIP-7702 not supported on chain: ${chainId}`,
+    );
+  }
+}
+
+function validateCapabilities(sendCalls: SendCalls) {
+  const { calls, capabilities } = sendCalls;
+
+  const requiredTopLevelCapabilities = Object.keys(capabilities ?? {}).filter(
+    (name) => capabilities?.[name].optional !== true,
```
ffmcgee725: if `call.capabilities?.[name]` is `undefined`, we will get a runtime error at `call.capabilities?.[name].optional` when trying to access `optional` of `undefined`.

app/core/RPCMethods/eip5792.ts
```diff
@@ -381,6 +263,255 @@ export async function getCapabilities(address: Hex, chainIds?: Hex[]) {
...
+  );
+
+  const requiredCallCapabilities = calls.flatMap((call) =>
+    Object.keys(call.capabilities ?? {}).filter(
+      (name) => call.capabilities?.[name].optional !== true,
```
ffmcgee725: if `call.capabilities?.[name]` is `undefined`, we will get a runtime error at `call.capabilities?.[name].optional` when trying to access `optional` of `undefined`.

app/core/RPCMethods/eip5792.ts
ffmcgee725: we should consider wrapping `AccountsController.getSelectedAccount()?.address` in `toHex` instead of explicitly type casting it

app/core/RPCMethods/eip5792.ts
jpuri: `name` is obtained from `Object.keys(call.capabilities ?? {})`. It is not likely to be undefined as it is object keys.

app/core/RPCMethods/eip5792.ts
jpuri: `name` is obtained from `Object.keys(call.capabilities ?? {})`. It is not likely to be undefined as it is object keys.

app/core/RPCMethods/eip5792.ts
jpuri: It is actually Hex value only thus casting is not needed.

app/core/RPCMethods/eip5792.ts
ffmcgee725: "not likely to be undefined" suggests there is a possibility of it being `undefined` then ?
I'm making the case that, if we are going to optional chain `capabilities?.[name]`, we might as well optional chain `capabilities?.[name]?.optional` as well, to be on the safe side. I'm not 100% familiar with the context of work, just looking at the bare code and providing the feedback I can muster.

app/core/RPCMethods/eip5792.ts
ffmcgee725: Can we remove the cast then ?

app/core/RPCMethods/eip5792.ts
jpuri: I added optional chaining here

app/core/RPCMethods/eip5792.ts
jpuri: I added optional chaining here also

app/core/RPCMethods/eip5792.ts
jpuri: It gives TX issue if I remove cast. But all evm addresses as we know are Hex

ffmcgee725 (approved):

PR #16765, opened 2025-06-27
perf: add some formatted address memoization - cp-7.50.0

ccharly:
## **Description**

Some more memoization to avoid unnecessary computation.

## **Related issues**

N/A

## **Manual testing steps**

No user facing changes. So just test the app as usual.

## **Screenshots/Recordings**

Not applicable


ccharly:
Cursor is right about the `undefined` account/account ID, but I made the assumption that all our selectors are based on the current list of accounts, meaning we can safely assume they are always define.

However, if we want to be more doing "protective coding" to cover this, we could also fallback to the previous behavior of computing the formatted addresses dynamically if it cannot be found in the map? Something like:

```ts
const formattedAddress =
  formattedAddressByAccountId[internalAccount.id] ?? toFormattedAddress(internalAccount.address);
```

Any opinion on this?

- - -

Also, if we get that one in:
- https://github.com/MetaMask/metamask-mobile/pull/16829

The fallback will be really really fast too.

owencraston:
@ccharly I think having the fallback to compute (and then adding into the cache after its computed) is a totally good solution.

PR #16844, opened 2025-07-01
fix: cp-7.50.0 Earn lending blocking UI tweaks

nickewansmith:
## **Description**
- There are some UIX issues noticed during lending flow that should be addressed before the release

- Update styles to accommodate smaller phones
- Add polling to earn token list to pull in cross network tokens
- Ensure lending learn more modal always uses the underlying token context
- Ensure that the empty state CTA switches to the lending token network
- Improve lending confirmation view handling
- Remove position dependency to further increase response speed and reliability

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/16834
Fixes: https://consensyssoftware.atlassian.net/browse/TAT-137

## **Manual testing steps**

- General testing of the earn lending feature
- Deposit/Withdraw/Approve
- Popular Networks / Single Network actions
- On small phones the feature should not have hidden buttons or cut off views
- Learn more modal should list earn token symbol only for all popups (withdraw and receipt token asset details earn section)

### **After**

https://www.loom.com/share/343b169bc4e24d6997058c1f25c1410a?sid=c884a764-2b1c-4fdf-9180-322ec7a7d994


app/components/UI/Earn/components/EmptyStateCta/index.tsx
```diff
@@ -58,7 +59,21 @@ const EarnEmptyStateCta = ({ token }: EarnEmptyStateCta) => {
     earnToken?.experience?.estimatedAnnualRewardsFormatted ?? '0',
   ).toFixed(0);
   const apr = earnToken?.experience?.apr;
-  const navigateToLendInputScreen = () => {
+  const navigateToLendInputScreen = async () => {
```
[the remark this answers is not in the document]
nickewansmith: This is on line 93
```
navigate('StakeScreens', {
      screen: Routes.STAKING.STAKE,
      params: { token },
    });
    ```

File: app/components/UI/Earn/utils/token/index.ts [hunk 2]
```diff
@@ -76,6 +76,11 @@ export const sortByHighestApr = (tokensToSort: EarnTokenDetails[]) =>
     (a, b) => parseFloat(b.experience.apr) - parseFloat(a.experience.apr),
   );
 
+export const sortByHighestBalance = (tokensToSort: EarnTokenDetails[]) =>
+  [...tokensToSort].sort((a, b) =>
+    b.balanceFiatNumber > a.balanceFiatNumber ? 1 : -1,
```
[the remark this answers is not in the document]
nickewansmith: This is fine in this case

app/components/UI/Earn/LendingLearnMoreModal/index.tsx
```diff
@@ -154,6 +156,36 @@ export const LendingLearnMoreModal = () => {
...
+  ]);
+  useEffect(() => {
+    if (tokenSnapshot?.token?.symbol) {
+      setAssetSymbol(tokenSnapshot?.token?.symbol);
+    } else if (earnToken?.symbol || earnToken?.ticker) {
+      setAssetSymbol(earnToken?.symbol || earnToken?.ticker || null);
+    } else {
+      setAssetSymbol(null);
+    }
+  }, [tokenSnapshot?.token?.symbol, earnToken?.symbol, earnToken?.ticker]);
```
Matt561, line 187: Future improvement suggestion: Might be a good idea to create a HoC (e.g. `withEarnTokens`) or hook to avoid this snapshot duplication in different components. No need to do this now but it may be nice to have a single place for this logic.

May also be nice to have a shared earn token list context that the snapshots pipe to so that we don't need to rerun these blocks in all views or components that need them every time.

Let's create a tech debt ticket to come back to this post launch when/if we get cycles.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -774,40 +775,48 @@ const EarnInputView = () => {
...
-                text: 'Tooltip Opened',
-                location: EVENT_LOCATIONS.EARN_INPUT_VIEW,
-                tooltip_name: 'MetaMask Pool Estimated Rewards',
-              },
-            })}
-            isLoading={isLoadingEarnMetadata}
-          />
-        )}
-      </View>
+      <ScrollView
```
Matt561, line 778: Likely overthinking: Does having a scrollview on the input page feels odd in practice? I'm not able to test and curious since this view has the keypad and token selector. Just wouldn't want the view to scroll or shift when users press on the keypad buttons.

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/EarnLendingDepositConfirmationView.styles.ts
```diff
@@ -14,6 +14,9 @@ const styleSheet = (params: { theme: Theme }) => {
     contentContainer: {
       paddingHorizontal: 16,
     },
+    scrollView: {
+      height: '85%',
```
Matt561, line 18: Is explicitly setting the height of the scrollview necessary? Asking because using percentage values like this can be risky and not translate accurately to different screen sizes. 

I don't understand at a glance why we wouldn't want the scrollview to fill the parent completely if it contains the entire view.

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/components/DepositInfoSection/index.tsx
```diff
@@ -87,12 +87,16 @@ const DepositInfoSection = ({
...
                   {
                     estimatedAnnualRewardsForAmount?.estimatedAnnualRewardsFormatted
                   }
                 </Text>
-                <Text color={TextColor.Alternative}>
+                <Text
+                  color={TextColor.Alternative}
+                  numberOfLines={1}
+                  ellipsizeMode="head"
+                >
```
Matt561, line 99: What is the output when the text is truncated? I don't understand why we'd hide portions of the estimatedAnnualRewards

Could you provide an example?

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/index.tsx
```diff
@@ -379,7 +379,10 @@ const EarnLendingDepositConfirmationView = () => {
         'TransactionController:transactionSubmitted',
         () => {
           emitDepositTxMetaMetric(MetaMetricsEvents.EARN_TRANSACTION_SUBMITTED);
-          navigation.navigate(Routes.TRANSACTIONS_VIEW);
+          // There is variance in when navigation can be called across chains
+          setTimeout(() => {
+            navigation.navigate(Routes.TRANSACTIONS_VIEW);
+          }, 0);
```
Matt561, line 385: If you're using this to defer navigation until the js and native operations are settled it may be more robust to use `InteractionManager.runAfterInteractions()` instead. I believe this function was made for exactly this.

E.g.
```ts
  InteractionManager.runAfterInteractions(() => {
    navigation.navigate(Routes.TRANSACTIONS_VIEW);
  });
```

app/components/UI/Earn/Views/EarnLendingWithdrawalConfirmationView/EarnLendingWithdrawalConfirmationView.test.tsx
```diff
@@ -845,6 +845,10 @@ describe('EarnLendingWithdrawalConfirmationView', () => {
         }),
       );
 
+      await act(async () => {
```
Matt561, line 850: Let's add a comment here to describe the purpose in case it isn't intuitive for future developers. 

From what I understand you're waiting until the next event loop tick and for react to finish its operations (e.g. state updates, useEffect, etc)

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -151,6 +152,7 @@ const EarnWithdrawInputView = () => {
       !receiptToken?.chainId
     )
       return;
+    setMaxRiskAwareWithdrawalAmount(undefined);
```
Matt561, line 155: Does this fix the bug we were chatting about yesterday or is it for something else? 👀

app/components/UI/Earn/components/EarnTokenList/__snapshots__/EarnTokenList.test.tsx.snap
```diff
@@ -175,7 +175,7 @@ exports[`EarnTokenList render matches snapshot 1`] = `
                 }
               }
             >
-              Select a token
+              Select a token to deposit
```
Matt561, line 178: I see we added "to deposit" does this render "to withdraw" contextually as well? Just want to make sure it doesn't render "Select a token to deposit" when in the withdrawal context.

app/components/UI/Earn/hooks/useEarnNetworkPolling.ts
```diff
@@ -0,0 +1,123 @@
...
+            }),
+          );
+
+          const networkClientId =
+            Engine.context.NetworkController.findNetworkClientIdByChainId(
+              chainId,
+            );
+
+          if (networkClientId && tokensToImport.length > 0) {
+            await TokensController.addTokens(tokensToImport, networkClientId);
```
Matt561, line 127: Is this importing tokens even if the user has a zero balance? Does the `tokensState?.allDetectedTokens` implicitly only pick up tokens where a user has a balance?

Just want to make sure we're not adding zero balance tokens that a user may never need. We don't want to clutter a user's token list.

app/components/UI/Earn/hooks/useEarnNetworkPolling.ts
```diff
@@ -0,0 +1,123 @@
...
+      if (!selectedAccount?.address || !useTokenDetection) return;
+
+      const { TokensController } = Engine.context;
+      const allDetectedTokens = tokensState?.allDetectedTokens || {};
+
+      for (const chainId of LENDING_CHAIN_IDS) {
+        const chainDetectedTokens =
+          allDetectedTokens[chainId]?.[selectedAccount.address];
+        if (
+          chainDetectedTokens &&
```
Matt561, line 107: Should we skip if the token list already has the token imported?

app/components/UI/Earn/hooks/useEarnNetworkPolling.ts
```diff
@@ -0,0 +1,123 @@
...
+).map((chainId) => toHex(chainId as Hex));
+
+/**
+ * Hook that triggers multi-network polling for Earn/lending functionality.
+ * Only starts polling when the component using this hook mounts (lazy loading).
+ * Automatically stops when the component unmounts.
+ *
+ * This ensures that lending tokens from all supported networks are available
+ */
+export const useEarnNetworkPolling = () => {
```
Matt561, line 54: I don't understand the problem you're solving as much so this may be naive _but_ would this make more sense to call once on app boot instead of every time a component using this hook is rendered? Is there a reason to call this more than just on app boot? Maybe the `tokensDetected` is updated every time a new token with a non-zero balance is found?

Apologies if this is a dumb question

app/components/UI/Earn/utils/tempLending.ts
```diff
@@ -376,14 +376,15 @@ export const getAaveV3MaxSafeWithdrawal = async (
...
       // We want result in token decimals
-      const maxTokenAmount = totalCollateralBase
-        .mul(parseUnits('1', tokenDecimals)) // Scale up to token decimals
-        .div(assetPrice); // Divide by price (removes the 8 decimals from price)
+      // const maxTokenAmount = totalCollateralBase
+      //   .mul(parseUnits('1', tokenDecimals)) // Scale up to token decimals
+      //   .div(assetPrice); // Divide by price (removes the 8 decimals from price)
 
-      return maxTokenAmount.toString();
+      // return maxTokenAmount.toString();
```
Matt561: Let's delete this commented out code If we don't need to calculate the total collateral since zero debt always means the health factor won't be impacted afaik.

app/components/UI/Earn/utils/tempLending.ts
```diff
@@ -657,6 +663,8 @@ export const getAaveV3MaxRiskAwareWithdrawalAmount = async (
         getAaveV3MaxSafeWithdrawal(userData, receiptToken as EarnTokenDetails),
       ]);
 
+    if (!maxHealthFactorWithdrawalInTokens) return undefined;
```
Matt561: We likely don't want this since we still want to check poolLiquidity. If the `maxHealthFactorWithdrawalInTokens` is undefined. We can filter out undefined values before calling `BigNumber.min`

```ts
const values = [value1, value2, value3].filter(v => v !== undefined);
const min = BigNumber.min(...values);
```

app/components/UI/Earn/components/EarnTokenList/index.tsx
```diff
@@ -87,6 +88,10 @@ export interface EarnTokenListProps {
 }
 
 const EarnTokenList = () => {
+  // Start polling lending networks when this component mounts and stops when it unmounts
+  // This is currently the main component that needs data cross chain on boot
+  useEarnNetworkPolling();
```
Matt561, line 93: I think it would be valuable to explain why this hook is necessary in the hook's file for future us.

app/components/UI/Earn/components/EarnTokenSelector/EarnTokenSelector.styles.ts
```diff
@@ -16,11 +16,13 @@ export default (params: { theme: Theme }) => {
     startAccessoryContainer: {
       paddingRight: 16,
       flexDirection: 'row',
-      maxWidth: 210,
       alignItems: 'center',
+      maxWidth: '95%',
```
Matt561, line 20: These percentage values worry me a bit. I've had headaches in the past getting consistent UIs across screen sizes with this approach.

nickewansmith:
https://cursor.com/agents?selectedBcId=bc-1be99df1-f75e-4635-95fe-b54d72a51b52

app/components/UI/Earn/LendingLearnMoreModal/index.tsx
nickewansmith: Ideally we have no need for snapshots, if it was known that the tokens would not be available if the user had not imported them the design would have been different from the start.

We need tokens that are not in the wallet with full reactivity for rates and ideally we should be able to convert to EarnTokenDetails type. I'd have to think more about this and ask other teams for input on architecture there. 

I have some ideas, but we may need some polling mechanism that gets the rate info for all earn token pairs or something like that.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
nickewansmith: The scroll view is not on the keypad, only the top portion which has a lot of possible content. Instead of getting into device conditions, I just did this temporarily for the phones that are small so everything is still functional. On larger phones the view up top does not scroll.

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/EarnLendingDepositConfirmationView.styles.ts
nickewansmith: Iirc this was set to be compatible with a smaller phone while working with small medium and large phones on screen simultaneously. The scroll view section is small. I have checked a several phones and it has been ok.

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/components/DepositInfoSection/index.tsx
nickewansmith: On smaller phones and even on larger ones that row does not have that much space. If estimated annual rewards were $900.56 983.45677 USDC. We must do something to stay on the line. These would be rarer cases. I thought it safer to have these ellipses vs making the text split to two rows without design decision

nickewansmith: Example output $900.68 ...65 USDC

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/index.tsx
nickewansmith: I did try this, but it didn't work. This race condition seems to happen on faster networks and would require more investigation to get to the bottom of why that is happening.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
nickewansmith: Yes this fixes that bug by unsetting the max withdrawal amount in the parent caller when the user has not borrowed anything. Then if they have borrowed we do the other checks and return a value.

app/components/UI/Earn/components/EarnTokenList/__snapshots__/EarnTokenList.test.tsx.snap
nickewansmith: Yes, it renders the withdrawal text when on the withdrawal list

app/components/UI/Earn/hooks/useEarnNetworkPolling.ts
nickewansmith: We could call this on app boot, but this would add extra load to the app. When the app starts for a new user it starts on mainnet ETH. I don't want to introduce polling for other chains on app boot before they click on the popular networks filter. Instead we only use it when the earn token list is being viewed. The polling controller is handling de-duplication of polling input calls behind the scene.

> Maybe the tokensDetected is updated every time a new token with a non-zero balance is found?

This already happens due to global polling, but it won't happen for tokens on other networks when on the single network view as that is how the global pollers are currently programmed. This hook ensures that that is never a problem.

app/components/UI/Earn/hooks/useEarnNetworkPolling.ts
nickewansmith: These tokens can be detected at any point in time.

app/components/UI/Earn/hooks/useEarnNetworkPolling.ts
nickewansmith: No it won't import if they have a zero balance, it doesn't do any extra importing. The imports are the same as if one would have clicked in the popular networks view. Whatever would have imported will show in the earn token list.

app/components/UI/Earn/utils/tempLending.ts
nickewansmith: I did this for now to avoid showing the withdrawal amount text as MSO did not want this to show unless the user has borrowed positions. I believe he would rather these types of checks pushed until v2. If we really want to check this in the next version we can, but I am veering towards the safer side for now. More than likely these pools would not run out of liquidity.

I see what you mean though, and since I will have some feedback to address anyway I'll see what I can do.

app/components/UI/Earn/components/EarnTokenSelector/EarnTokenSelector.styles.ts
nickewansmith: Understood, they would worry me more as well if I wasn't testing with 3 distinct phone sizes. These percentages work better than what existed before. Anywhere I add these style changes are to accommodate android/ios small/medium/large. I'm sure it is not perfect, but the previous ones were not working well with longer text on either side. I am no css guru, but this does test well on the devices I used and is a minimal surface area.

app/components/UI/Earn/utils/tempLending.ts
nickewansmith: I have added back the pool liquidity check on withdraw

app/components/UI/Earn/components/EarnTokenSelector/EarnTokenSelector.styles.ts
Matt561: Sounds good!

Matt561 (approved):

nickewansmith:
This pr has passed tests previously and the current smoke failure is unrelated.

PR #16939, opened 2025-07-04
fix: wallet link in account details

montelaidev:
## **Description**

This PR removes the noop in the wallet link of the account details. 

## **Related issues**

Fixes:

## **Manual testing steps**

1. Enable multichain account design
2. Go to the account details
3. Click the wallet link and see that you are brought to the wallet details. 

### **After**

https://github.com/user-attachments/assets/0ba6d592-1e66-4c6e-af69-bad88eafd287


app/components/Views/MultichainAccounts/AccountDetails/AccountTypes/BaseAccountDetails/index.tsx
```diff
@@ -154,15 +162,15 @@ export const BaseAccountDetails = ({
           </Text>
           <TouchableOpacity
             testID={AccountDetailsIds.WALLET_NAME_LINK}
-            onPress={handleEditWalletName}
+            onPress={handleWalletClick}
```
ccharly: Given that we check for `!wallet` inside the callback, should we disable this part if `wallet` is not defined here too? 🤔 not sure if would really happen though...

app/multichain-accounts/selectors/accountTreeController.test.ts
```diff
@@ -486,4 +492,452 @@ describe('AccountTreeController Selectors', () => {
...
+          backgroundState: {
+            AccountTreeController: {
+              accountTree: {
+                wallets: {
+                  [WALLET_ID_1]: {
+                    id: WALLET_ID_1,
+                    metadata: { name: 'Wallet 1' },
+                    groups: {
+                      group1: {
+                        accounts: [ACCOUNT_ID_1],
```
ccharly: Nit: But normally group IDs are composed of their wallet ID + a unique identifier to make name unique. Also, groups have an `id` field.

I guess we don't see any type-error because of the `as unknown as RootState`.

Maybe we could have a `WALLET_1_GROUP_ID_1` similar to your `WALLET_ID_1`?

app/multichain-accounts/selectors/accountTreeController.ts
```diff
@@ -62,3 +63,40 @@ export const selectWalletById = createDeepEqualSelector(
...
+      }));
+
+      const accountWallet = accountWallets.find((wallet) =>
+        wallet.accounts.some((account) => account === accountId),
+      );
+
+      return accountWallet
+        ? accountTreeState.accountTree.wallets[accountWallet.walletId]
+        : null;
+    },
```
ccharly, line 103: I don't think we need to map anything here actually 🤔 given that we're just re-using data from the state directly, we can do it that way (haven't tested):

```suggestion
      const accountWallet = Object.values(
        accountTreeState.accountTree.wallets,
      ).find((wallet) => {
        return Object.values(wallet.groups).some((group) => {
          return group.accounts.some((account) => account.id === accountId);
        });
      };

      return accountWallet
        ? accountTreeState.accountTree.wallets[accountWallet.id]
        : null;
    },
```

app/multichain-accounts/selectors/accountTreeController.test.ts
```diff
@@ -486,4 +492,452 @@ describe('AccountTreeController Selectors', () => {
...
+            },
+            RemoteFeatureFlagController: {
+              remoteFeatureFlags: {
+                enableMultichainAccounts: {
+                  enabled: true,
+                  featureVersion: '1',
+                  minimumVersion: '1.0.0',
+                },
+              },
+            },
```
ccharly: To avoid repeating this several times, we could maybe have it in a `mockDefaultState` constant and expand it in every test cases that need it?

app/multichain-accounts/selectors/accountTreeController.test.ts
```diff
@@ -486,4 +492,452 @@ describe('AccountTreeController Selectors', () => {
...
+
+    it('returns correct wallet when account is in different groups', () => {
+      const mockWallet = {
+        id: WALLET_ID_1,
+        metadata: { name: 'Wallet 1' },
+        groups: {
+          'keyring:1:ethereum': {
+            accounts: [ACCOUNT_ID_1],
+          },
+          'snap:solana:mainnet': {
```
ccharly, line 459: Nit:
```suggestion
          `${WALLET_ID_1}:1`: {
            accounts: [ACCOUNT_ID_1],
          },
          `${WALLET_ID_1}:2: {
```

app/multichain-accounts/selectors/accountTreeController.test.ts
```diff
@@ -486,4 +492,452 @@ describe('AccountTreeController Selectors', () => {
...
+        id: WALLET_ID_WITH_GROUPS,
+        metadata: { name: 'Complex Wallet' },
+        groups: {
+          'keyring:1:ethereum': {
+            accounts: [ACCOUNT_ID_1, ACCOUNT_ID_2],
+          },
+          'snap:solana:mainnet': {
+            accounts: [ACCOUNT_ID_3],
+          },
+          'keyring:2:ethereum': {
```
ccharly, line 631: Nit:

```suggestion
          `${WALLET_ID_1}:1`: {
            accounts: [ACCOUNT_ID_1, ACCOUNT_ID_2],
          },
          `${WALLET_ID_1}:2`: {
            accounts: [ACCOUNT_ID_3],
          },
          `${WALLET_ID_1}:3`: {
```

app/multichain-accounts/selectors/accountTreeController.ts
```diff
@@ -62,3 +63,40 @@ export const selectWalletById = createDeepEqualSelector(
...
+ * Get a wallet by account ID from AccountTreeController
+ * Returns a selector function that can be called with an account ID to find the wallet containing that account
+ * @param state - Root redux state
+ * @param multichainAccountsState1Enabled - Whether multichain accounts feature is enabled
+ * @returns Selector function that takes an account ID and returns the containing wallet or null
+ **/
+export const selectWalletByAccount = createDeepEqualSelector(
+  [selectAccountTreeControllerState, selectMultichainAccountsState1Enabled],
+  (accountTreeState, multichainAccountsState1Enabled) =>
+    (accountId: AccountId): AccountWallet | null => {
```
ccharly, line 79: Question: Any advantage of using a function over a `Map` or any kind of mapping for account IDs <> wallets?

IMO, a function would require more processing than if we were doing a selector that returns a mapping (that would only react if one of its dependent selectors would change).

app/components/Views/MultichainAccounts/AccountDetails/AccountTypes/BaseAccountDetails/index.tsx
montelaidev: It would become a noop if its not defined, but it should never be undefined.

app/multichain-accounts/selectors/accountTreeController.ts
montelaidev: It was the pattern from the previous selector. I think this selector should be removed once the reverse mapping is available

app/multichain-accounts/selectors/accountTreeController.ts
montelaidev: It just flattens the accounts into an array. I think this selector will be removed once the reverse map is available.

app/multichain-accounts/selectors/accountTreeController.test.ts
```diff
@@ -12,33 +13,50 @@ const WALLET_ID_NONEXISTENT = 'keyring:nonexistent';
...
+const ACCOUNT_ID_1 = 'account1';
+const ACCOUNT_ID_2 = 'account2';
+const ACCOUNT_ID_3 = 'account3';
+const ACCOUNT_ID_NONEXISTENT = 'nonexistent-account';
+
+/**
+ * Helper function to create a base mock state with RemoteFeatureFlagController
+ */
+const createMockState = (
+  accountTreeController: Record<string, unknown> = {},
```
ccharly: Could have used `AccountTreeControlleState` (but that will probably require more change too, given that group IDs won't be matching in this tests). (Actually, `DeepPartial` might be better here 🤔 anyway, you get the idea).

```suggestion
  accountTreeController: Partial<AccountTreeControllerState> | undefined = {},
```

app/multichain-accounts/selectors/accountTreeController.ts
ccharly: Yes, but I know we sometimes forget about those things 🙈

Do you think we can add a `// TODO: Use reverse mapping once available, for fast indexing.`?

ccharly (approved):
Code LGTM, haven't tested though.

PR #16960, opened 2025-07-07
fix: fix request payments token list

### LINKED ISSUE #16958: [Bug]: Request Payments token list is sluggish
### Describe the bug

I’ve been seeing some lag when using the token list to request funds. When searching for a token, the autocomplete causes the app to freeze for a few seconds, and in some cases, it even crashes. It feels pretty unstable and gives off a bad UX. I can consistently reproduce it on both a physical device and a simulator. Keep in mind, token search on the Token import screen is 10x faster than request payment. 

### Expected behavior

Token autocomplete on the request flow should be as performant as token complete in the import token flow.  

### Steps to reproduce

Fresh install of the app (I used both main and the outgoing release v7.50.0 (2069) 
Create Wallet
Go to payment request flow
Try searching for "Li" 
Autocomplete search is a bit sluggish. 
Notice when you try tapping on the token as soon as it appears in the list, the app freezes a bit. 

```shell

```

### Detection stage

On main branch

### Version

main

### Build type

None

### Device

iphone 16 Pro

### Operating system

iOS

### Additional context

For what its worth, I asked AI to help me identify the root cause(s) and this is the output:

1. Debounced Search Delay: The search uses a 300ms debounce. 
2. FlashList Rendering Issues: FlashList requires proper configuration and can hang if not set up correctly.
3. Async State Updates: The search results are being updated asynchronously. 

### Severity

_No response_

cortisiko:
## **Description**

The request token flow test was getting stuck during the token search autocomplete. Looking at the logs, we could see tons of timer activity and the test would hang there. I was also able to reproduce this manually. 

### Why was it happening?
I compared the request payment search with the token import search (which works fine) and found the issue. The request payment was using a 300ms debounced search, while token import searches immediately. Plus, the request payment was using FlashList for rendering while the token import uses simple View components.

What I changed

Removed the debouncing delay
In PaymentRequest/index.js, I removed the 300ms debounce that was causing the test to wait unnecessarily. 

Switched from FlashList to simple rendering
In PaymentRequest/AssetList/index.tsx, I replaced the FlashList with the same simple map rendering that token import uses

### Why does this fix it?

The test was hanging because:
The 300ms debounce meant the test had to wait for each keystroke
FlashList was adding unnecessary complexity for a small list of results
The token import flow works fine with immediate search and simple rendering
Now the request payment search behaves exactly like the working token import search - immediate response, simple rendering, no delays.

app/components/UI/PaymentRequest/index.js - removed debouncing
app/components/UI/PaymentRequest/AssetList/index.tsx - replaced FlashList with simple rendering

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

### **Before**

https://github.com/user-attachments/assets/4ec42893-e20f-471f-be21-aa522c482603

### **After**

https://github.com/user-attachments/assets/c26a2175-d725-4f86-9880-19424ec8cf4a


app/components/UI/PaymentRequest/index.js
```diff
@@ -433,15 +433,14 @@ class PaymentRequest extends PureComponent {
...
-  debouncedHandleSearch = debounce(this.handleSearch, 300);
-
   handleSearchTokenList = (searchInputValue) => {
     if (typeof searchInputValue !== 'string') {
       searchInputValue = this.state.searchInputValue;
     }
     this.setState({ searchInputValue });
 
-    this.debouncedHandleSearch(searchInputValue);
+    // Remove debouncing for faster search like token import
```
darioAnongba: This comment should not be part of the codebase as it implies some knowledge of what was implemented before. Seems more like a commit message.

darioAnongba (dismissed):

app/components/UI/PaymentRequest/index.js
cortisiko: good call out, will update! thanks @darioAnongba

app/components/UI/PaymentRequest/AssetList/index.tsx
```diff
@@ -146,12 +123,26 @@ const AssetList = ({
...
       style={styles.listContainer}
     >
-      <FlashList
-        data={searchResults}
-        renderItem={renderItem}
-        estimatedItemSize={80}
-        keyExtractor={(_, index) => index.toString()}
-      />
+      {/* Use simple rendering like token import for better performance */}
+      {searchResults.slice(0, 6).map((item: { symbol?: string; name?: string; address?: string }, index: number) => {
```
Prithpal-Sooriya, line 127: Since we are limiting this to 6 items, it should be fine to not use a FlashList.
I don't think we need to resort to a FlatList either.

app/components/UI/PaymentRequest/index.js
```diff
@@ -433,15 +433,13 @@ class PaymentRequest extends PureComponent {
...
-  debouncedHandleSearch = debounce(this.handleSearch, 300);
-
   handleSearchTokenList = (searchInputValue) => {
     if (typeof searchInputValue !== 'string') {
       searchInputValue = this.state.searchInputValue;
     }
     this.setState({ searchInputValue });
 
-    this.debouncedHandleSearch(searchInputValue);
+    this.handleSearch(searchInputValue);
```
Prithpal-Sooriya, line 442: I do prefer the debounce (maybe with a shorter delay, or leading vs trailing).

However lets not over-optimise. See if there are any performance implications (I doubt there will be many now that we are only rendering 6 items)

Prithpal-Sooriya (approved):

PR #17015, opened 2025-07-09
feat: Use `EditSpendingCapModal` for batch approval editing

OGPoyraz:
## **Description**

This PR aims to re-use `EditSpendingCapButton` in order to prevent duplication and use more advanced modal with validations. It doesn't introduce any functional change - please see before/after recordings.

## **Changelog**

No changelog

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5297

## **Manual testing steps**

1. Have smart account 
2. Try swapping that requires both approval and swap transaction (for example do a swap from uniswap with a token that you didn't approve before)
3. You should be able to click whole value component as it's a default UX 

### **Before**

https://github.com/user-attachments/assets/1dfb5388-1377-42b9-8d1b-77e35df9cd8e

### **After**

https://github.com/user-attachments/assets/37b216e4-547b-4d77-b73f-308ea096b6d2


app/components/UI/SimulationDetails/BalanceChangeRow/BalanceChangeRow.tsx
```diff
@@ -7,36 +7,37 @@ import Text, {
...
-import EditRowValue from '../EditRowValue';
 import styleSheet from './BalanceChangeRow.styles';
 
 interface BalanceChangeRowProperties extends ViewProps {
+  approveMethod?: ApproveMethod;
   balanceChange: BalanceChange;
-  editTexts?: {
-    title: string;
-    description: string;
-  };
```
OGPoyraz, line 22: As both modals using same translations - removing them.

app/components/Views/confirmations/components/approve-static-simulations/approve-and-permit2/approve-and-permit2.tsx
```diff
@@ -86,6 +87,7 @@ export const ApproveAndPermit2 = () => {
                 decimals: decimals ?? 1,
                 spendingCap: rawAmount ?? '',
                 onSpendingCapUpdate,
+                tokenSymbol: tokenSymbol,
```
OGPoyraz: This was a forgotten implementation, so adding the ticker here and `app/components/Views/confirmations/components/approve-static-simulations/increase-decrease-allowance/increase-decrease-allowance.tsx`.

app/components/Views/confirmations/components/modals/edit-spending-cap-modal/edit-spending-cap-modal.test.tsx
OGPoyraz, line 1: Mostly correcting the translations here.

app/components/Views/confirmations/components/modals/edit-spending-cap-modal/edit-spending-cap-modal.tsx
```diff
@@ -83,7 +85,8 @@ export const EditSpendingCapModal = ({
           style={styles.balanceInfo}
           color={TextColor.Alternative}
         >
-          Account balance : {balance}
+          {strings('confirm.edit_spending_cap_modal.account_balance')} :{' '}
+          {balance} {tokenSymbol}
```
OGPoyraz: `tokenSymbol` and "Account balance" translations were forgotten but corrected now.

app/components/UI/SimulationDetails/BalanceChangeRow/BalanceChangeRow.tsx
```diff
@@ -50,21 +51,33 @@ const BalanceChangeRow: React.FC<BalanceChangeRowProperties> = ({
...
+              />
+            </EditSpendingCapButton>
+          ) : (
+            <AmountAndAddress
+              asset={asset}
+              amount={amount}
+              isAllApproval={isAllApproval}
+              isUnlimitedApproval={isUnlimitedApproval}
             />
           )}
```
OGPoyraz, line 80: Whole `value` should be clickable hence we are doing this change.

app/components/UI/SimulationDetails/BatchApprovalRow/BatchApprovalRow.tsx
```diff
@@ -38,12 +33,11 @@ const BatchApprovalRow = () => {
     <>
       {approveBalanceChanges.map((balanceChange, index) => (
         <BalanceChangeRow
+          approveMethod={balanceChange.approveMethod}
           key={`batch_balance_change-${index}-${balanceChange}`}
           balanceChange={balanceChange as BalanceChange}
-          enableEdit={balanceChange.asset.type === AssetType.ERC20}
```
OGPoyraz, line 43: Moving this into `BalanceChangeRow` as we are already passing `balanceChange`.

jpuri:
As discussed before I think it is not good approach to replace more general modal `EditRowValue` by `EditSpendingCapModal`.

OGPoyraz:
@jpuri please see the screenshot below. This isn’t a new approach - it's the exact same one we used in the extension.

Screenshot 2025-07-18 at 15 17 27

I understand and respect your intention behind introducing the `EditRowValue` component. However, in the context of confirmations, there’s currently no use case for single-value editing - even in the extension itself. Since the extension repository is already ahead in terms of features, we haven’t encountered a scenario where this approach was needed.

If a similar requirement arises in the future, we can certainly extend the implementation in a way that makes sense for the broader architecture. For now, though, I believe introducing this level of abstraction would be a case of premature optimization. It adds an extra layer of complexity without providing any tangible benefit at this stage. The current implementation is already clean, concise, and reuses the same component across both places without any duplication or divergence in logic. By avoiding unnecessary generalization, we keep the codebase more maintainable and focused on real, present use cases rather than hypothetical ones.

matthewwalsh0:
> As discussed before I think it is not good approach to replace more general modal `EditRowValue` by `EditSpendingCapModal`.

This is a great point worth clarifying, but I think in this instance the changes in this PR make sense.

Both components are great abstractions, but the current `EditRowValue` isn't really a generic row editor and in some ways is less generic as it requires a `BalanceChange` which isn't available in the single approval confirmation use case. It also still renders the token balance, plus uses approval specific strings such as `edit_value_balance_info`.

So really it feels like we just have two different edit approval balance modals that we need to combine, and we just happened to use `EditSpendingCapModal` as it has more usages already, and more primitive inputs so it can support all three scenarios.

I absolutely agree that if these inline editors become a common pattern for other data, we should create a generic component that this one could extend, but that would introduce unnecessary complexity currently as we only have a single use case, and the `EditSpendingCapModal` component is already small enough to not hinder readability as the core logic is still provided by a callback prop.

app/components/UI/SimulationDetails/BalanceChangeRow/BalanceChangeRow.tsx
jpuri: AmountPill also has a tooltip visible when we click it. Can the 2 click events create trouble ?

jpuri (approved):
Changes look good, this may need to be addressed:

https://github.com/MetaMask/metamask-mobile/pull/17015#discussion_r2238531276

PR #17025, opened 2025-07-09
chore: add performance traces for earn deposit and withdrawal flows

ameliejyc:
## **Description**

This PR adds more Sentry traces for Earn metrics following the guidelines for [Performance Tracing](https://github.com/MetaMask/contributor-docs/blob/main/docs/performance-tracing.md). 

- Earn Deposit Spending Cap Screen
- Earn Deposit Review Screen
- Earn Deposit Confirmation Screen (this is the bottom sheet on Lending, and the standalone confirmation on Staking)
- Earn Deposit Tx Confirmed (only for Lending currently)
- Earn Withdraw Review Screen
- Earn Withdraw Confirmation Screen (this is the bottom sheet on Lending, and the standalone confirmation on Staking)
- Earn Withdraw Tx Confirmed (only for Lending currently)
- Earn Token List
- Earn Claim Confirmation Screen

As there are differences between how confirmations are shown between Lending and Pooled Staking there are points to note here:

1. For PS there is no **Earn Deposit Review Screen** or **Earn Withdraw Review Screen** trace. This is because the PS flow goes straight from Input -> Confirmation screen. I haven't added traces on the old confirmation flow on PS as I believe these are to be removed (but please confirm if this is not the case!)
2. Currently there are no **Tx Confirmed** traces for PS (deposit, withdraw and claim). This is because the Earn team doesn't handle the transaction flow code from the point that the user sees the confirmation screen. I think this needs to be a point of discussion of how to implement traces moving forwards for transactions app-wide (if that is indeed something we want to trace) rather than putting in a custom Earn implementation. Let me know your thoughts on this as I'm still ramping up on mobile code.
3. I have however been able to add **Tx Confirmed** traces for Lending deposits and withdrawals in their respective Confirmation Screen components as currently the transaction listener and associated metrics code lives there. 

A full list of the Earn performance benchmarks can be found [here](https://docs.google.com/document/d/1-mq98T39GhHIjkrGteIb99P7YLU3yHXSK8kaRuJTTJU/edit?tab=t.0#heading=h.l1kgno957cpf).

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Part fixes: https://consensyssoftware.atlassian.net/browse/TAT-574

## **Manual testing steps**

- First get permissions to https://metamask.sentry.io/ and the metamask team within that
- Ensure you have the MM_SENTRY_DSN_DEV key in your .js.env
- Check out this branch and run the app on iPhone. Ensure metametrics are enabled.
- Deposit an earn token for lending, e.g. USDC on Linea. A full list of supported tokens is [here](https://consensyssoftware.atlassian.net/wiki/spaces/PUK/pages/399765831835/Tokens+networks+support)
- See on the console that Sentry Logger is called for the associated new Earn Deposit... traces
- See on the Sentry [dashboard](https://metamask.sentry.io/dashboard/137081) for Earn Deposit... boards that new traces have appeared
- Repeat for ETH staking, withdrawal and claim and Lending withdrawals

### **After**

Our dashboard is updated here: https://metamask.sentry.io/dashboard/137081/?statsPeriod=24h


Matt561:
> Currently there are no Tx Confirmed traces for PS (deposit, withdraw and claim). This is because the Earn team doesn't handle the transaction flow code from the point that the user sees the confirmation screen. I think this needs to be a point of discussion of how to implement traces moving forwards for transactions app-wide (if that is indeed something we want to trace) rather than putting in a custom Earn implementation. Let me know your thoughts on this as I'm still ramping up on mobile code.

This is a good callout and is worth discussing with the confirmations team. I'll add this to next Monday's Earn x Confirmations sync. You're welcome to join too if you don't already have an invitation. Your context here will be super useful.

The general impression I have right now, is that the Confirmations team creates generic and reusable components and helpers for experience-owning teams like us to create and own the actual confirmation screens.

I still think this is worth bringing up in our next sync with them. This call isn't until next Monday though and I recognize that you may want an answer sooner! 😄

app/components/Views/confirmations/external/staking/info/staking-withdrawal/staking-withdrawal.tsx
```diff
@@ -14,6 +14,8 @@ import StakingContractInteractionDetails from '../../components/staking-contract
 import { HeroRow } from '../../../../components/rows/transactions/hero-row';
```
Matt561, line 14: Not directly related to this PR, but would it make sense for the Earn team added as codeowners for confirmation screens we maintain?

Matt561 (dismissed):
LGTM! 👍

Matt561 (approved):

matthewwalsh0 (approved):

app/components/Views/confirmations/external/staking/info/staking-withdrawal/staking-withdrawal.tsx
ameliejyc: I've asked the team and will create a new PR for this if so 👍

PR #17106, opened 2025-07-11
feat: 2.12 backup keyring key instead of password cp-7.52.0

### LINKED ISSUE #16914: Update metamask/seedless-onboarding-controller to version 2.0.0
Please update metamask/seedless-onboarding-controller to version 2.0.0

### LINKED ISSUE #17317: [Bug]: User login with new password got the error because of mis-match version controller between extension and mobile
### Describe the bug

If user login to social account on browser and mobile
then change password on browser
mobile need to re-login again with new password
but got the error cannot login with new password

### Expected behavior

Should be able to login with new password

### Screenshots/Recordings

image

https://github.com/user-attachments/assets/4c001294-7ebb-4fb8-8bc2-b5c0704be537

### Steps to reproduce

1. Login with social account on browser and mobile
2. Change password on extension
3. Re-login again on mobile
4. Login with new password
5. See the error

```shell

```

### Detection stage

On main branch

### Version

7.51.0

### Build type

Beta

### Device

Honor Play, iPhone14Promax

### Operating system

iOS, Android

### Additional context

_No response_

### Severity

_No response_

ieow:
## **Description**
Previously, seedless onboarding was backing up the keyring password to allow for vault recovery after a password change. Now we backup the keyring encryption key.

See the https://github.com/MetaMask/decisions/pull/85 for more details.
This is part of the implementation of option 6.

This PR update with seedless controller v2 as well

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/components/Nav/Main/index.js
```diff
@@ -127,28 +127,37 @@ const Main = (props) => {
...
-          ),
-          type: 'error',
-          icon: IconName.Danger,
-          isInteractable: false,
-          onPrimaryButtonPress: async () => {
-            await Authentication.lockApp({ locked: true });
+    const checkIsSeedlessPasswordOutdated = async () => {
+      if (isSeedlessPasswordOutdated) {
+        const isOutdated =
+          await Authentication.checkIsSeedlessPasswordOutdated();
```
tuna1207: @ieow why do we need to check seedless password outdated 1 more time before showing modal when `isSeedlessPasswordOutdated` is already true ?

app/core/Authentication/Authentication.ts
```diff
@@ -468,7 +474,18 @@ class AuthenticationService {
         name: TraceName.VaultCreation,
         op: TraceOperation.VaultCreation,
       });
-      await this.loginVaultCreation(password);
+
+      if (authData.oauth2Login) {
+        // if seedless flow - rehydrate
+        await this.rehydrateSeedPhrase(password, authData);
+      } else if (await this.checkIsSeedlessPasswordOutdated()) {
```
tuna1207, line 488: @ieow we should check pw outdated skip cache in this case with
```typescript
checkIsSeedlessPasswordOutdated(true)
```
but i think we are updating the check pw outdated flow in this case to only check once at unlock page load instead of on every unlock call
cc @himanshuchawla009 @lwin-kyaw

app/components/Nav/Main/index.js
ieow: This is temporary workaround
the state isSeedlessPasswordOutdated is still true when navigate to main wallet screen after syncing the password.
I am still debugging the issue

app/core/Authentication/Authentication.ts
ieow: skipping cache will cause the login delay ( especially during network slow)
also if there is error when querying the network, it might cause the user stuck at unlock screen
also it increase the server's load

matthiasgeihs:
why does this PR have 260+ commits? can we rebase on top of `main` to make sure we only review the relevant changes? or is this too much work now?

app/core/Authentication/Authentication.ts
matthiasgeihs: at which points do we not skip cache currently?

matthiasgeihs, line 488: should we change the default of `checkIsSeedlessPasswordOutdated` to `skipCache = True`?
then by default we always get the correct answer from the backend and only intentionally skip the cache.

app/core/Authentication/Authentication.ts
```diff
@@ -663,51 +707,58 @@ class AuthenticationService {
...
-      // update seedlessOnboardingController to use latest global password
-      await SeedlessOnboardingController.syncLatestGlobalPassword({
-        oldPassword: currentDevicePassword,
-        globalPassword,
-      });
+      await KeyringController.submitPassword(globalPassword);
+      throw new SeedlessOnboardingControllerError(
+        'Password Recently Updated',
+        SeedlessOnboardingControllerErrorType.PasswordRecentlyUpdated,
+      );
```
matthiasgeihs, line 786: that's weird. You are throwing an error on the happy path? Can you explain what the code is supposed to do here?

app/core/Authentication/Authentication.ts
```diff
@@ -663,51 +707,58 @@ class AuthenticationService {
...
+        if (
+          error.code ===
+          SeedlessOnboardingControllerErrorType.PasswordRecentlyUpdated
+        ) {
+          throw error;
+        }
+      }
+    }
+
+    // const releaseLock = await this.syncSeedlessGlobalPasswordMutex.acquire();
```
matthiasgeihs: remove if not needed.
don't leave commented out lines in production code.

app/core/Authentication/Authentication.ts
```diff
@@ -663,51 +707,58 @@ class AuthenticationService {
...
-        ),
-        true,
-      );
-      await this.resetPassword();
-
-      // check password outdated again skip cache to reset the cache after successful syncing
-      await SeedlessOnboardingController.checkIsPasswordOutdated({
-        skipCache: true,
       });
+      await KeyringController.changePassword(globalPassword);
```
matthiasgeihs, line 813: shouldn't we do this only if password was changed?

app/core/Authentication/Authentication.ts
```diff
@@ -663,51 +707,58 @@ class AuthenticationService {
...
+      await this.syncKeyringEncryptionKey();
     } catch (err) {
       // lock app again on error after submitPassword succeeded
       await this.lockApp({ locked: true });
       throw err;
     }
+    this.resetPassword();
+    // releaseLock();
```
matthiasgeihs: remove if not needed. why is the lock no longer needed?

app/core/Authentication/Authentication.ts
```diff
@@ -663,51 +707,58 @@ class AuthenticationService {
...
-        skipCache: true,
       });
+      await KeyringController.changePassword(globalPassword);
+      await this.syncKeyringEncryptionKey();
     } catch (err) {
       // lock app again on error after submitPassword succeeded
       await this.lockApp({ locked: true });
       throw err;
     }
+    this.resetPassword();
```
matthiasgeihs: what does this do?

matthiasgeihs (dismissed):
i couldn't do a full review. i would need a PR intro and much more time as i'm not familiar with the mobile code base.

focused on usage of keyring controller.
please see my high level comments.
didn't review tests.

requesting changes in particular because there are two commented out lines relating to a lock, which should not go into production code. is this lock no longer needed? if so, remove the lines.

app/core/Authentication/Authentication.ts
ieow: User submitted old password when the global password has changed

app/core/Authentication/Authentication.ts
ieow: removed

app/core/Authentication/Authentication.ts
ieow: removed

app/components/Nav/Main/index.js
chaitanyapotti: Can u add a comment for the same @ieow

app/core/Authentication/Authentication.ts
ieow: reset biometric password in keychain/keystore

app/core/Authentication/Authentication.ts
ieow: we need to change the password when the password is outdated right?

app/core/Authentication/Authentication.ts
ieow: updated

himanshuchawla009: Hi @ieow ,  we recently decided to keep the cache time as 10 seconds for now and later optimise this value based on the load and also if network call fails don't block the user, just swallow the error.

himanshuchawla009: You can create a diff PR for this.

app/core/Authentication/Authentication.ts
himanshuchawla009: This flow is updated in extension, this is not the correct way to handle this. Refer to this PR: https://github.com/MetaMask/metamask-extension/pull/34441

himanshuchawla009: Can also create a separate PR for this, not a blocker for this PR.

app/core/Authentication/Authentication.ts
himanshuchawla009: yes so we should do on pwd outdated case but here we seems to be doing for all cases, plz refer to extension PR i attached above.

himanshuchawla009 (dismissed):
lgtm

chaitanyapotti (approved):
Found issues here which will be fixed in the next PR

himanshuchawla009 (approved):

app/core/Authentication/Authentication.ts
matthiasgeihs: my comment was more about the coding style. i am confused why this throwing an error on the happy path.

this is the full code block i'm referring to:
```ts
     try {
       await KeyringController.verifyPassword(globalPassword);
       throw new SeedlessOnboardingControllerError(
         'Password Recently Updated',
         SeedlessOnboardingControllerErrorType.PasswordRecentlyUpdated,
       );
     } catch (error) {
       if (error instanceof SeedlessOnboardingControllerError) {
         if (
           error.code ===
           SeedlessOnboardingControllerErrorType.PasswordRecentlyUpdated
         ) {
           throw error;
         }
       }
     }
```

what i see the code doing is the following.
attempt to verify keyring password.
if verification success, you throw a custom error.
if it fails, keyring controller throws an error.
so in any case this throws an error.
after that you check if the error is the custom seedless error and throw in that case.
in all other cases the error is dropped.

so as far as i can see this simply asserts that the keyring password is different than the global password.
couldn't you simply do the following instead?

suggestion:
```ts
let passwordValid = true;
try {
    await KeyringController.verifyPassword(globalPassword);
} catch (error) {
    passwordValid = false
}

if (passwordValid) {
    throw new Exception("Keyring password expected to be different from global password");
}
```

additionally, you could move the verification code into a helper function to make the main logic even more readable.
```ts
function verifyKeyringPassword(pw: string): boolean {
let passwordValid = true;
try {
    await KeyringController.verifyPassword(globalPassword);
} catch (error) {
    passwordValid = false
}
return passwordValid;
}
```

app/core/Authentication/Authentication.ts
matthiasgeihs: this relates to the comment i made above.
it wasn't obvious to me what the code does as it's written a bit convulted imo.
after going over it again, i realize that at this point we actually know that the keyring controller password is different from the global password.

app/core/Authentication/Authentication.ts
matthiasgeihs: i guess you mean store the new password to keychain? or what does "reset" mean?

app/core/Authentication/Authentication.ts
ieow: fixed in next pr https://github.com/MetaMask/metamask-mobile/pull/17469

PR #17157, opened 2025-07-14
refactor: use `@metamask/eth-qr-keyring`

mikesposito:
## **Description**

This PR uses the `@metamask/eth-qr-keyring` package to handle QR code keyring functionality, replacing the previous implementation from `@keystonehq/metamask-airgapped-keyring`. 

Along with this change, these controllers are being updated:
- `keyring-controller` from `^22.1.1` to `^23.1.0`
- `preferences-controller` from `^18.4.0` to `^19.0.0`
- `signature-controller` from `^32.0.0` to `^33.0.0`

And this patch is being deleted:
- `patches/@metamask+keyring-controller+22.1.1.patch` - no longer needed due to the update to `keyring-controller@^23.1.0`

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.


app/util/bridge/hooks/useSubmitBridgeTx.ts
```diff
@@ -19,6 +20,7 @@ export default function useSubmitBridgeTx() {
       },
       stxEnabled,
     );
+    Logger.log('Bridge transaction submitted', txResult);
```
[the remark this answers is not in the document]
mikesposito: Removed in 50503d072a5f

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -393,7 +390,21 @@ const RootRPCMethodsUI = (props) => {
...
+            );
+            if (currentApproval) {
+              autoSign(transactionMeta);
+            }
+          },
+          ({ pendingApprovals }) =>
+            Object.values(pendingApprovals).some(
+              ({ requestData }) => requestData.txId === transactionMeta.id,
+            ),
+        );
```
mikesposito: I'm not super fan of this approach, but it's the only solution I found to go around the issue that `TransactionController:unapprovedTransactionAdded` is fired consistently before the pending approval is added to `ApprovalsController`, which makes `autoSign` fail with `Error accepting pending approval [Error: Approval request with id '...' not found`.

patches/@metamask+bridge-status-controller+40.2.0.patch
```diff
@@ -0,0 +1,26 @@
...
++++ b/node_modules/@metamask/bridge-status-controller/dist/bridge-status-controller.mjs
+@@ -593,7 +593,7 @@ export class BridgeStatusController extends StaticIntervalPollingController() {
+                 // Submit EVM tx
+                 // For hardware wallets on Mobile, this is fixes an issue where the Ledger does not get prompted for the 2nd approval
+                 // Extension does not have this issue
+-                const requireApproval = __classPrivateFieldGet(this, _BridgeStatusController_clientId, "f") === BridgeClientId.MOBILE && isHardwareAccount;
++                const requireApproval = __classPrivateFieldGet(this, _BridgeStatusController_clientId, "f") === BridgeClientId.MOBILE && selectedAccount.metadata.keyring.type.includes('Ledger Hardware');
+                 // Handle smart transactions if enabled
+                 txMeta = await __classPrivateFieldGet(this, _BridgeStatusController_trace, "f").call(this, {
+                     name: isBridgeTx
```
mikesposito: Upstream PR: https://github.com/MetaMask/core/pull/6564

package.json
```diff
@@ -256,7 +255,7 @@
     "@metamask/json-rpc-middleware-stream": "^8.0.7",
     "@metamask/key-tree": "^10.1.1",
     "@metamask/keyring-api": "^21.0.0",
-    "@metamask/keyring-controller": "^22.1.1",
+    "@metamask/keyring-controller": "^23.1.0",
```
Gudahtt, line 261: This update will introduce a new peer dependency warning from `preferences-controller`, which is also one major version behind. Could you update that package here as well? That would resolve the warning, and there are no other breaking changes in that version so it should require no further changes

Gudahtt: Ah, we're in a similar situation with `signature-controller`. Perhaps we can bump that here as well

mikesposito: Done in 367ec3da9c14 and 9b597ba21cb8

vivek-consensys (dismissed):
LGTM! Tested all major e2e flows across iOS and Android. Adding the QA Passed label!

app/core/QrKeyring/QrKeyring.ts
```diff
@@ -0,0 +1,38 @@
...
+    keyring: QrKeyring;
+    metadata: KeyringMetadata;
+  }) => Promise<CallbackResult>,
+): Promise<CallbackResult> =>
+  Engine.context.KeyringController.withKeyring(
+    { type: ExtendedKeyringTypes.qr },
+    operation,
+    // TODO: Refactor this to stop creating the keyring on-demand
+    // Instead create it only in response to an explicit user action, and do
+    // not allow Ledger interactions until after that has been done.
```
danroc: Why Ledger interactions should not be allowed until the refactor is done?

app/core/redux/slices/qrKeyringScanner/index.ts
```diff
@@ -0,0 +1,33 @@
+import { QrScanRequest } from '@metamask/eth-qr-keyring';
+import { PayloadAction, createSlice } from '@reduxjs/toolkit';
+
+export interface QrKeyringScannerState {
+  pendingScanRequest?: QrScanRequest;
```
danroc, line 5: Does it mean we can have only one pending request?

app/core/QrKeyring/QrKeyring.ts
mikesposito: Oh, that's a mistake in the comment. This function was copied from the existing Ledger equivalent: https://github.com/MetaMask/metamask-mobile/blob/cc1ed751f91101be3d8dee5f28a9e7f9f4ef8ec0/app/core/Ledger/Ledger.ts#L33-L33 

Will change `Ledger` to `Qr`.

app/core/redux/slices/qrKeyringScanner/index.ts
mikesposito: Yeah, correct. I thought that having a queue system in place for QR scans wouldn't be needed at this point, but still viable in the future if we see the need for having multiple scan requests at the same time. Do you think we should implement a queue system now instead?

danroc: I checked with @matthewwalsh0 and it's not possible to enqueue requests on mobile, so we are fine with having only one pending request.

app/core/QrKeyring/QrKeyring.ts
mikesposito: Fixed in 8c969e34f120

app/components/Nav/Main/RootRPCMethodsUI.js
```diff
@@ -326,13 +326,16 @@ const RootRPCMethodsUI = (props) => {
...
-
         const isLedgerAccount = isHardwareAccount(
           transactionMeta.txParams.from,
           [ExtendedKeyringTypes.ledger],
         );
 
+        // As the `TransactionController:unapprovedTransactionAdded` event is emitted
+        // before the approval request is added to `ApprovalController`, we need to wait
+        // for the next tick to make sure the approval request is present when auto-approve it
+        await new Promise((resolve) => setTimeout(resolve, 0));
```
danroc, line 337: Is this problem new or was it already present in the existing code?

Gudahtt: Previously the `await KeyringController.resetQRKeyringState();` call was effectively a "workaround" for this issue. The issue surfaced after that step was removed.

I guess technically you might say it's not a "problem" per se, as there are definitely advantages to having the opportunity to run code after a transaction is added to state, but before it's queued up as an approval. But it's certainly awkward to workaround for triggering an operation on the approval.

danroc (dismissed):
LGTM!

app/components/Nav/Main/RootRPCMethodsUI.js
danroc: Indeed, "problem" wasn't the right term. Thanks for the explanation 👍

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -314,35 +195,39 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
-        const keyringAccounts = await keyring.getAccounts();
-        removeAccountsFromPermissions(keyringAccounts);
-      },
-    );
-    const { remainingAccounts } = await KeyringController.forgetQRDevice();
-    Engine.setSelectedAddress(remainingAccounts.at(-1));
+    const remainingAccounts = KeyringController.state.keyrings
+      .filter((keyring) => keyring.type !== ExtendedKeyringTypes.qr)
+      .flatMap((keyring) => keyring.accounts);
+    Engine.setSelectedAddress(remainingAccounts[remainingAccounts.length - 1]);
```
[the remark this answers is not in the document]
Gudahtt: Good point, but not applicable for us (it's not currently possible for all accounts to be QR accounts)

Gudahtt (dismissed):
LGTM!

danroc (dismissed):

tommasini (dismissed):

shane-t (dismissed):

Prithpal-Sooriya (dismissed):

danroc (dismissed):

danroc (approved):

Gudahtt (approved):

PatrykLucka (approved):

matthewwalsh0 (approved):

shane-t (approved):

ameliejyc (approved):

salimtb (approved):
assets changes lgtm

tommasini (approved):

PR #17333, opened 2025-07-17
fix: account discovery for non evm networks

Daniel-Cross:
## **Description**

Added account discovery for all non-EVM networks (Solana & Bitcoin currently).

## **Changelog**

CHANGELOG entry: Added account discovery for Non-EVM accounts

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/jira/software/c/projects/NWNT/boards/1561/backlog?selectedIssue=NWNT-431

## **Manual testing steps**

1. Do a fresh install of the app or import new wallet via SRP.
2. Import wallet that contains Solana and/or Bitcoin accounts
3. Open account selector to see imported accounts.

### **After**

Uploading Simulator Screen Recording - iPhone 16 Pro - 2025-07-17 at 18.29.27.mp4…


darioAnongba:
first thing I see from this PR is that we are "bundling" the notion of "non-EVM" into a single block of logic, but we should keep the logic separate for each network/Snap as we don't want to fail or retry if a network succeeds and an other one fails. So we need as many items as Snaps:

- `SOLANA_DISCOVERY_PENDING`
- `BITCOIN_DISCOVERY_PENDING`

same for error handling, we don't want an error that says "non-EVM discovery failed", but one for each Snap/network (like it was before).

Same logic applies for feature flags. Everything Bitcoin should be behind the `bitcoin` feature flag as that PR will start discovering Bitcoin accounts on prod I think

app/actions/multiSrp/index.ts
```diff
@@ -90,14 +92,38 @@ export async function importNewSecretRecoveryPhrase(mnemonic: string) {
...
+        );
+      }
+    });
 
-  discoveredAccountsCount = await multichainClient.addDiscoveredAccounts(
-    newKeyring.id,
-  );
+    const hasFailures = results.some((result) => result.status === 'rejected');
+    if (hasFailures) {
+      await StorageWrapper.setItem(NON_EVM_DISCOVERY_PENDING, 'true');
```
darioAnongba: we were not storing anything before, was that a bug on previous implementation?

darioAnongba (dismissed):

app/core/Authentication/Authentication.test.ts
```diff
@@ -71,6 +71,7 @@ jest.mock('../SnapKeyring/MultichainWalletSnapClient', () => ({
   },
   WalletClientType: {
     Solana: 'solana',
+    Bitcoin: 'bitcoin',
```
darioAnongba, line 68: Using the feature flag here will work I think since you iterate over the client types?

app/actions/multiSrp/index.ts
Daniel-Cross: I think so, it stores in local storage if it fails it will retry again on the next user unlock. This particular file handles discovery when importing a new SRP.

aganglada:
Closing as [this](https://github.com/MetaMask/metamask-mobile/pull/17566) solves the problem

PR #17360, opened 2025-07-18
fix: Fix gas inputs to handle comma instead of dot

OGPoyraz:
## **Description**

This PR aims to fix gas inputs to replace comma `,` input to dot `.`

## **Changelog**

CHANGELOG entry: Fixed a bug on comma inputs on gas text boxes.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/17195

## **Manual testing steps**

1. Trigger any transaction
2. Try changing gas values with decimal value - use a comma separator 
3. It should immediately replace with dot and shouldn't throw error.


app/components/Views/confirmations/components/gas/gas-price-input/gas-price-input.tsx
```diff
@@ -38,9 +38,10 @@ export const GasPriceInput = ({
 
   const handleChange = useCallback(
     (text: string) => {
-      validateGasPriceCallback(text);
-      setValue(text);
-      const updatedGasPrice = add0x(decGWEIToHexWEI(text) as Hex);
+      const normalizedText = text.replace(',', '.');
+      validateGasPriceCallback(normalizedText);
```
matthewwalsh0: Could we avoid the duplication if we just changed `validateValueIsNumber` to also accept the `,`?

As the platform will ensure the correct keyboard is shown?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/components/gas/gas-price-input/gas-price-input.tsx
OGPoyraz: This was also my initial thought however changing `validateValueIsNumber` validation won't simply be enough to fix this. Because `decGWEIToHexWEI` expects that value with a dot so this duplication will always be necessary. Please see that the utility cannot parse that value

Screenshot 2025-07-18 at 13 48 02

We can make user to input text it with comma instead of dot but then it may cause a confusion because suggested values are presented with dot as we show them with `parseFloat`

Screenshot 2025-07-18 at 13 51 12

I agree it duplicates a bit (necessary in every fix unless we want `decGWEIToHexWEI` to handle values with comma) but I believe this is the simplest fix that we can do.

matthewwalsh0: In that case, could we make a single `normalizeGasValue` util that includes the `replace` and maybe `add0x(decGWEIToHexWEI(` also?

OGPoyraz: Updated two things 
- Created `convertGasInputToHexWEI` and `normalizeGasInput`
- Used new utilities in the validations and gas inputs.

All these make it possible to use comma in the inputs. Here is the recording

https://github.com/user-attachments/assets/7a183c24-d709-4aa6-a689-64299be08462

matthewwalsh0 (approved):

PR #17373, opened 2025-07-18
perf: Enable react-freeze

andrepimenta:
## **Description**

This PR improves the performance of the app by:
- Suspending all screens that are not visible to the user (using [React Freeze](https://github.com/software-mansion/react-freeze))

We just need to enableFreeze because on React Navigation V5, the `detachInactiveScreens` defaults to true:

Boolean used to indicate whether inactive screens should be detached from the view hierarchy to save memory. 
Make sure to call enableScreens from [react-native-screens](https://github.com/software-mansion/react-native-screens) to make it work. **Defaults to true.**
Reference: https://reactnavigation.org/docs/5.x/bottom-tab-navigator/#detachinactivescreens

## **Recordings**

Before React Freeze you can see that the wallet view continues to re-render even if not active:

https://github.com/user-attachments/assets/5f8acca5-67b9-4530-bf8a-22fd895e921d

After React Freeze you can see that the wallet view doesn't re-render if not active:

https://github.com/user-attachments/assets/61b410d0-547d-4b61-b66d-d99ff4c4c9ed

## **Manual testing steps**

1. Go from screen A to screen B
2. Notice screen A doesn't render if not visible

## **What can go wrong?**

There might be some side effects from enabling react-freeze. To quote their docs:
Can freezing some of my app components break my app? And how?
There are few ways that we are aware of when <Freeze> can alter the app behavior:

> #### Can freezing some of my app components break my app? And how?
> 
> There are few ways that we are aware of when `<Freeze>` can alter the app behavior:
> 
> 1. When attempting to freeze parts of the app that is visible to the user at a given moment -- in this case the frozen part is going to be replaced by the placeholder (or just by nothing if no placeholder is provided). So unless you really want this behavior make sure to only set `freeze` to `true` when the given subtree should not be visible and you expect user to not interact with it. A good example are screens on the navigation stack that are down the stack hierarchy when you push more content.
> 2. When you rely on the frozen parts layout to properly position the unfrozen parts. Note that when component is in frozen state it gets replaced by a placeholder (or by nothing if you don't provide one). This may impact the layout of the rest of your application. This can be workaround by making placeholder take the same amount of space as the view it replaces, or by only freezing parts that are positioned absolutely (e.g. the component that takes up the whole screen).
> 3. When component render method has side-effects that relay on running for all prop/state updates. Typically, performing side-effects in render is undesirable when writing react code but can happen in your codebase nonetheless. Note that when subtree is frozen your component may not be rendered for all the state updates and render method won't execute for some of the changes that happen during that phase. However, when the component gets back from the frozen state it will render with the most up-to-date version of the state, and if that suffice for the side-effect logic to be correct you should be ok."

I'm more concerned with number 3, but E2E plus app regression testing should be enough to find potential issues. If those issues do exist, then they will show coding bad practices (side-effects in render) so we should address them.


Cal-L (commented):
Nice, code LGTM! can you share a video showing a before and after of some screens that shows the freeze in action?

Cal-L (approved):
Very nice. LGTM!

Cal-L:
E2E failures not caused by PR. We ran twice and the same ones that failed before passed

PR #17384, opened 2025-07-18
feat: Adding a very basic send page for new send implementation.

jpuri:
## **Description**

Adding a very basic send page hidden behind env variable.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5398

## **Manual testing steps**
NA

## **Screenshots/Recordings**
NA


app/components/Views/confirmations/components/send/index.ts
OGPoyraz, line 1: Essentially everything is component and this could go into `components` folder but what we think of putting it into `confirmations/pages` / `confirmations/screens` folder? Because unlike other components this is a route now and it would be much more clear to have it on separate folder.

app/components/Nav/Main/MainNavigator.js
```diff
@@ -833,9 +840,19 @@ const MainNavigator = () => {
...
         name="SendFlowView"
-        component={SendFlowView}
+        component={
+          process.env.MM_SEND_REDESIGNS_ENABLED === 'true'
+            ? SendComponent
+            : SendFlowView
+        }
         //Disabling swipe down on IOS
         options={{ gestureEnabled: false }}
       />
```
OGPoyraz, line 858: We have 3 `send` related route here now. As you already have a context, can you clarify which is what and why it's needed?

app/components/Views/confirmations/components/send/index.ts
jpuri: Let me address this in PR I am working on now, as restructuring this will result in conflicts in PR built on top of this.

app/components/Nav/Main/MainNavigator.js
jpuri: `SendComponent` is the the new one added. The multiple other legacy ones we will eventually get rid of.

app/components/Views/WalletActions/WalletActions.tsx
```diff
@@ -275,6 +275,12 @@ const WalletActions = () => {
         .build(),
     );
 
+    if (process.env.MM_SEND_REDESIGNS_ENABLED === 'true') {
```
matthewwalsh0, line 278: Minor, as we'll do this in a few places and development will take some time, is it worth a util?

app/components/Views/confirmations/components/send/index.ts
jpuri: We have not been using `confirmations/pages` in new designs. `confirm` is also inside `component` and I would prefer to keep this consistent.

app/components/Views/confirmations/components/send/send.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import Text from '../../../../../component-library/components/Texts/Text';
+
+const Send = () => (
+  <View>
+    <Text>SEND IMPLEMENTATION TO COME HERE</Text>
+  </View>
+);
+
+export default Send;
```
matthewwalsh0, line 12: Minor, do we prefer non-default exports to make auto-complete easier and support other exports like constants?

.js.env.example
```diff
@@ -153,8 +153,8 @@ export IOS_GOOGLE_REDIRECT_URI='com.googleusercontent.apps.615965109465-h6tp2h3c
 #export IOS_GOOGLE_CLIENT_ID="615965109465-89b2lmqgm5ka8j8t403qhooouv57id9b.apps.googleusercontent.com"
 #export IOS_GOOGLE_REDIRECT_URI="com.googleusercontent.apps.615965109465-89b2lmqgm5ka8j8t403qhooouv57id9b:/oauth2redirect/google"
 
-# Enable 7702 locally
-export MM_SMART_ACCOUNT_UI_ENABLED="true"
+# Enable send re-designs locally
+export MM_SEND_REDESIGNS_ENABLED="true"
```
matthewwalsh0, line 157: Minor, maybe just `MM_SEND_REDESIGN_ENABLED`?

app/components/Nav/Main/MainNavigator.js
OGPoyraz: If `SendComponent` is the new one, why do we use it twice?

If we are trying to redirect a new `SendComponent` when `MM_SEND_REDESIGNS_ENABLED` is settled why do we need the stack above?

Can't we let all other send navigation work as is and just keep 
```
component={
          process.env.MM_SEND_REDESIGNS_ENABLED === 'true'
            ? SendComponent
            : SendFlowView
        }
``` 
and remove the stack above?

app/components/Views/confirmations/components/send/index.ts
matthewwalsh0: While it is technically a screen or page, I think we benefit more by keeping all the send files in a single directory. In this case: `app/components/Views/confirmations/components/send/`

OGPoyraz: Yes essentially it's a component but this is a page or screen(whatever you prefer to call). `confirm` is a different context - it's a component that we use in our pages.

app/components/Nav/Main/MainNavigator.js
```diff
@@ -833,9 +840,19 @@ const MainNavigator = () => {
       <Stack.Screen name="Asset" component={AssetModalFlow} />
       <Stack.Screen name="Webview" component={Webview} />
       <Stack.Screen name="SendView" component={SendView} />
+      <Stack.Screen
```
matthewwalsh0, line 843: For simplicity, could we avoid adding a new path and just re-use the existing one with an ENV?

Outside the scope of this PR and predating us, but the number of send routes in here is wild 😄

matthewwalsh0 (approved):

app/components/Views/confirmations/components/send/index.ts
OGPoyraz: @matthewwalsh0 in fact - this may not be even a screen. Imagine this is a root `Stack.Navigator` have more screens under it if we are thinking to add similar pages like `SendAsset` `SendTo` `SendAmount`, in that case it's actually a stack. 

`app/components/Views/confirmations/routes/Send` => this should be a `Stack.Navigator` include following pages
`app/components/Views/confirmations/pages/SendAsset`
`app/components/Views/confirmations/pages/SendTo`
`app/components/Views/confirmations/pages/SendAmount`

If we dump all into `app/components/Views/confirmations/components/send/` folder this sounds like a components used in `send` context to me.

matthewwalsh0: We traditionally avoid alternate pages in favour of modals as we're ultimately building up a single transfer transaction, and could use modals to make specific selections. I think that's a simpler architecture than X different pages that have to share state globally (not ideal) and route between each other.

Regardless, does keeping all our send files / components in one directory not align with our new folder structure, the goals being to avoid deep nesting and lots of subdirectories, and favour simplicity? Is having all the send components in one flat directory not the easiest way to maintain them? 

Surely we want to reuse our successful confirmation architecture? Which was a single generic template component such as `Send`, that uses modals and components such as `Header`, `Footer`, `Amount`, `Token` etc?

OGPoyraz: > I think that's a simpler architecture than X different pages that have to share state globally (not ideal) and route between each other

So you're simply referring to a multi-step "wizard" like send screen where user progressively completing steps and see the `transfer` transaction as final. That's fine for me - in that case screens approach not really needed. 

Hopefully we don't have any difficulties in a case wants to start send flow from the middle. Because imagine user is navigating to send flow from a asset page (so they already picked the asset) and we need to skip first step etc. It's all technically possible to make it work but I thought navigating straight to step screens were much easier to maintain. As I said in the design meeting yesterday - there are lot of unknowns in the designs to decide architecture right now, we all have some ideas but my concern is design is still blurry.

> Surely we want to reuse our successful confirmation architecture? Which was a single generic template component such as Send, that uses modals and components such as `Header`, `Footer`, `Amount`, `Token` etc?

I don't think it's possible to use all components as `Send` flow will be outside of the context providers but surely can re-use UI components that are not tied to any context provider.

matthewwalsh0: I'd like to think having a central `Send` component makes that scenario easier, as we can just include some optional params to specify the token or other data, and skip modals or inputs based on that.

Like a confirmation, it becomes a single page with configurable elements, rather than the legacy pattern of a wizard / flow defined across many components, all sharing lots of global state and reducer logic.

> I don't think it's possible to use all components

I wasn't meaning the actual confirmation components, but that pattern of a single template component made up of smaller components that abstract the complexity for different types, scenarios, protocols etc.

app/components/Nav/Main/MainNavigator.js
jpuri: At some places I am passing asset in navigation as parameter and that is creating trouble with the old navigation as it does not expect a parameter. This asset passing will be visible in the following PR.

app/components/Views/confirmations/components/send/index.ts
jpuri: Things will be more clear in upcoming PR: https://github.com/MetaMask/metamask-mobile/pull/17453

Components are broken down and architecture can support both wizard like flow or single page as we have a context available across pages and a [nested navigator](https://github.com/MetaMask/metamask-mobile/pull/17453/files#diff-5fd4378476aeb826ac69758340c92bd9ce37292bea84336811f9ea32e82f602dR12).

app/components/Nav/Main/MainNavigator.js
jpuri: As I mentioned [here](https://github.com/MetaMask/metamask-mobile/pull/17384#discussion_r2225413971), at places I need to navigate to send path directly to be able to pass params to it.

OGPoyraz (approved):

app/components/Views/WalletActions/WalletActions.tsx
jpuri: Let me add that in following PR.

app/components/Views/confirmations/components/send/send.tsx
jpuri: 👍 let me address this in next PR

.js.env.example
jpuri: Good point, let me address in next PR.

jpuri:
@matthewwalsh0 : I will address the smaller feedbacks in the next PR

jpuri:
Feedbacks addressed here https://github.com/MetaMask/metamask-mobile/pull/17415

PR #17401, opened 2025-07-21
feat: perps order creation

abretonc7s:
## **Description**

  This PR implements three key user stories for Perps trading functionality:

  1. **TAT-1022**: Users can now open market orders using their perp account balance
  2. **TAT-1023**: Take Profit/Stop Loss (TP/SL) functionality has been implemented to
  help users manage investment risk
  3. **TAT-1031**: Users can now open limit orders with proper price setting

  ## **Changelog**

  CHANGELOG entry: Added Perps market and limit order creation with Take Profit/Stop
  Loss functionality

  ## **Related issues**

  Fixes: TAT-1022, TAT-1023, TAT-1031

  ## **Manual testing steps**

  ### Market Orders (TAT-1022):
  1. Navigate to Perps trading view
  2. Ensure "Market" order type is selected
  3. Enter order amount using slider or direct input
  4. Verify order uses perp account balance
  5. Place market order successfully

  ### Take Profit/Stop Loss (TAT-1023):
  1. In order creation view, tap on "Take Profit" or "Stop Loss"
  2. Verify new dual input design (USD and percentage fields)
  3. Test that changing USD updates percentage and vice versa
  4. Verify percentage quick buttons (+5%, +10%, +15%, +20%) work correctly
  5. Set TP/SL values and confirm they're applied to the order

  ### Limit Orders (TAT-1031):
  1. Select "Limit" order type
  2. Verify limit price shows "Set price" placeholder (not current price)
  3. Tap to set limit price
  4. Test -1% and -2% buttons adjust the entered limit price
  5. Try to place order without limit price - should show validation error
  6. Set limit price and successfully place order

  ## **Screenshots/Recordings**

  ### **Before**
  - No perps order creation functionality
  - No TP/SL risk management
  - No limit order support

  ### **After**
  - Full market order creation with perp account balance
  - Comprehensive TP/SL interface with dual inputs
  - Limit order creation with proper UX and validation

LONG BTC:
https://github.com/user-attachments/assets/7b35b2e9-a405-44ef-bd7d-383589cd03a7

SHORT ETH:
long_eth

  - [x] I've followed [MetaMask Contributor 
  Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding 
  Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CO
  DING_GUIDELINES.md).
  - [x] I've applied the right labels on the PR (see [labeling guidelines](https://githu
  b.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)).
  Not required for external contributors.

  - [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code
  being changed).
  - [ ] I confirm that this PR addresses all acceptance criteria described in the ticket
   it closes and includes the necessary testing evidence such as recordings and or
  screenshots.


app/components/UI/Perps/Views/PerpsDepositAmountView/PerpsDepositAmountView.tsx
```diff
@@ -243,7 +239,7 @@ const PerpsDepositAmountView: React.FC<PerpsDepositAmountViewProps> = () => {
       }
     }
 
-    return '-';
+    return '0x0';
```
gambinish: This looks like a magic string. Can we define this in a variable?

app/components/UI/Perps/Views/PerpsOrderSuccessView.tsx
```diff
@@ -0,0 +1,265 @@
...
+import { useTheme } from '../../../../util/theme';
+import type { Colors } from '../../../../util/theme/models';
+import { strings } from '../../../../../locales/i18n';
+import Routes from '../../../../constants/navigation/Routes';
+import { formatPrice } from '../utils/formatUtils';
+
+interface OrderSuccessParams {
+  asset: string;
+  direction: 'long' | 'short';
+  size: string; // Changed from amount to match what's passed
```
gambinish: Nit: Can remove comment

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+    ToastVariants: {
+      Account: 'account',
+      Network: 'network',
+      Transaction: 'transaction',
+      Simple: 'simple',
+      Success: 'success',
+      Error: 'error',
+    },
+  };
+});
```
gambinish: I don't think we should be mocking component library components. I was noticing that the agent likes to do this a lot.

Shouldn't be necessary though, can we remove?

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.test.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+jest.mock('../../utils/formatUtils', () => ({
+  formatPrice: jest.fn((price) => `$${price}`),
+}));
+
+// Mock network utils
+jest.mock('../../../../../util/networks', () => ({
+  getDefaultNetworkByChainId: jest.fn(() => ({ name: 'Arbitrum' })),
+  getNetworkImageSource: jest.fn(() => ({ uri: 'network-icon' })),
+}));
```
gambinish: Similarly, I think these mocks might be overly aggressive. I don't think we need to mock the components that this View consumes, because we want to be testing the components themselves.

I also don't think the utils need to be mocked to get these tests passing since these are local utilities. For  external libraries like react navigation, mocking makes a lot more sense IMO. We should be able to resolve this by mocking the application state instead.

I can help taking a look at this if you want.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+  const styles = createStyles(colors);
+  const navigation = useNavigation<NavigationProp<PerpsNavigationParamList>>();
+  const route = useRoute<RouteProp<{ params: OrderRouteParams }, 'params'>>();
+  const toastContext = useContext(ToastContext);
+
+  const toastRef = toastContext?.toastRef;
+  const { openTooltipModal } = useTooltipModal();
+
+  // Selectors
+  const tokenList = useSelector(selectTokenList);
```
gambinish, line 133: I'm leaving a note here to double check that this token list is the correct one to be using. Do we need to consider multichain tokens like SOL?

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+
+  // Get PerpsController methods and state
+  const { placeOrder, getMarkets } = usePerpsTrading();
+  const currentNetwork = usePerpsNetwork();
+  const cachedAccountState = usePerpsAccount();
+
+  // Get real HyperLiquid USDC balance
+  const availableBalance = parseFloat(
+    cachedAccountState?.availableBalance?.toString() || '0',
+  );
```
gambinish, line 152: We may want to consider not defaulting to 0 in some cases, because this could be misleading if cachedAccountState takes a while to resolve.

I think `null` make more sense, because we can rely on that to render a loading state util the actual balance resolves.

We're doing this in a few places I think. Might make sense to go over all of the Perps components at some point to polish this behavior.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+    amount: defaultAmountValue,
+    leverage: defaultLeverageValue,
+    balancePercent: Math.round(initialBalancePercent * 100) / 100,
+    takeProfitPrice: undefined,
+    stopLossPrice: undefined,
+    limitPrice: undefined,
+  });
+
+  const [isPlacingOrder, setIsPlacingOrder] = useState(false);
+  const [orderType, setOrderType] = useState<'market' | 'limit'>('market');
```
gambinish: Do we want an interface or enum for `MarketOrderType`?

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+  const enhancedUsdcToken = useMemo(() => {
+    if (!tokenList) return null;
+
+    return enhanceTokenWithIcon({
+      token: {
+        symbol: 'USDC',
+        address: '0xaf88d065e77c8cC2239327C5EDb3A432268e5831', // USDC on Arbitrum
+        decimals: 6,
+        chainId: '0xa4b1',
+        name: 'USD Coin',
```
gambinish: Hard coding this make me feel uncomfortable.

Is there a way that we can get this from state somehow? If that's not possible, at the very least we should use the same constants that are used throughout the rest of the application.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+      setMarketData(null);
+    }
+  }, [getMarkets, asset, toastRef, navigation]);
+
+  useEffect(() => {
+    setIsLoadingMarketData(true);
+    fetchMarketData().finally(() => {
+      setIsLoadingMarketData(false);
+    });
+  }, [fetchMarketData]);
```
gambinish: Is there a reason we aren't extracting this into it's own hook? I'm fine with it here, just curious

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+      selectedPaymentToken.chainId !== HYPERLIQUID_MAINNET_CHAIN_ID &&
+      selectedPaymentToken.chainId !== HYPERLIQUID_TESTNET_CHAIN_ID
+    ) {
+      errors.push(strings('perps.order.validation.only_hyperliquid_usdc'));
+    }
+
+    // Check if limit order has a limit price set
+    if (orderType === 'limit' && !orderForm.limitPrice) {
+      errors.push(strings('perps.order.validation.limit_price_required'));
+    }
```
gambinish: Thanks for adding this UI layer error handling. Are we also handing this error at the controller?

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+              isBold: true,
+            },
+          ],
+          iconName: IconName.CheckBold,
+          iconColor: IconColor.Success,
+          hasNoTimeout: false,
+        });
+
+        navigation.navigate(Routes.PERPS.ORDER_SUCCESS, {
+          orderId: result.orderId || 'unknown',
```
gambinish: Is there a case where an order doesn't have an orderId? Might make sense to handle the `orderId === 'unknown'` case separately if this is a valid state to enter into

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+          hasNoTimeout: true,
+          closeButtonOptions: {
+            label: strings('perps.order.error.dismiss'),
+            variant: ButtonVariants.Secondary,
+            onPress: () => toastRef?.current?.closeToast(),
+          },
+        });
+      }
+    } catch (error) {
+      toastRef?.current?.showToast({
```
gambinish, line 592: `toastRef?.current?.showToast` is repeated multiple times here. I wonder if we should create a hook or util that makes this more DRY and allows us to clean up this component a bit.

Not a blocker, can be cleaned up later

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+                  <ListItemColumn widthType={WidthType.Fill}>
+                    <Text variant={TextVariant.BodyLGMedium}>
+                      {strings('perps.order.limit_price')}
+                    </Text>
+                  </ListItemColumn>
+                  <ListItemColumn widthType={WidthType.Auto}>
+                    <Text variant={TextVariant.BodyLGMedium}>
+                      {orderForm.limitPrice
+                        ? formatPrice(orderForm.limitPrice)
+                        : 'Set price'}
```
gambinish, line 724: Needs to be internationalized

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+              <ListItem>
+                <ListItemColumn widthType={WidthType.Fill}>
+                  <View style={styles.detailLeft}>
+                    <Text variant={TextVariant.BodyLGMedium}>
+                      {strings('perps.order.leverage')}
+                    </Text>
+                    <TouchableOpacity
+                      onPress={() =>
+                        openTooltipModal(
+                          'What is Leverage?',
```
gambinish, line 687: Needs to be internationalized

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -0,0 +1,1037 @@
...
+        <View style={styles.infoSection}>
+          <View style={styles.infoRow}>
+            <View style={styles.detailLeft}>
+              <Text variant={TextVariant.BodyMD} color={TextColor.Alternative}>
+                {strings('perps.order.estimated_execution_time')}
+              </Text>
+              <TouchableOpacity
+                onPress={() =>
+                  openTooltipModal(
+                    'Execution Time',
```
gambinish, line 853: Needs to be internationalized

app/components/UI/Perps/Views/PerpsPositionDetailsView/PerpsPositionDetailsView.test.tsx
```diff
@@ -26,6 +26,28 @@ jest.mock('../../../../../core/SDKConnect/utils/DevLogger', () => ({
...
+    const { View, TouchableOpacity, Text } = jest.requireActual('react-native');
+    return (
+      <View testID="perps-tpsl-bottomsheet">
+        <TouchableOpacity onPress={onClose}>
+          <Text>Close</Text>
+        </TouchableOpacity>
+      </View>
+    );
+  },
+}));
```
gambinish, line 49: We probably shouldn't mock this component because it may result in false positive unit tests, since we aren't including the actual component that we are testing.

I guess the bottom sheet is an annoying thing to mock...I can timebox some time to look into this later today.

app/components/UI/Perps/Views/PerpsView.tsx
```diff
@@ -76,6 +139,9 @@ const PerpsView: React.FC<PerpsViewProps> = () => {
   const [isToggling, setIsToggling] = useState(false);
   const [result, setResult] = useState<string>('');
 
+  // Popular trading pairs
+  const POPULAR_ASSETS = ['BTC', 'ETH', 'SOL', 'ARB'];
```
gambinish, line 142: Can we put this in our HyperliquidConfig?

app/components/UI/Perps/Views/PerpsView.tsx
```diff
@@ -222,9 +291,17 @@ const PerpsView: React.FC<PerpsViewProps> = () => {
...
-              Cached Balance: ${cachedAccountState.totalBalance}
-            </Text>
+            <>
+              <Text variant={TextVariant.BodySM} color={TextColor.Muted}>
+                Balance: ${cachedAccountState.totalBalance}
+              </Text>
+              {parseFloat(cachedAccountState.totalBalance) === 0 && (
+                <Text variant={TextVariant.BodySM} color={TextColor.Warning}>
+                  No funds deposited. Use &apos;Deposit Funds&apos; to get
+                  started.
```
gambinish, line 304: Needs to be internationalized

app/components/UI/Perps/components/PerpsLeverageBottomSheet/PerpsLeverageBottomSheet.styles.ts
```diff
@@ -0,0 +1,148 @@
...
+    },
+    leverageTextLow: {
+      color: colors.text.default,
+    },
+    leverageTextMedium: {
+      color: colors.warning.default,
+    },
+    leverageTextHigh: {
+      color: colors.error.default,
+    },
```
gambinish, line 28: I think we need to be careful with styling text directly in stylesheets. My understanding is that we should rely on the `TextVariant` property of the `Text` component so that the design system team's updates will automatically propagate to whatever we make.

I think that the style sheet here should mostly be related to alignment/padding/margin

app/components/UI/Perps/components/PerpsLimitPriceBottomSheet/PerpsLimitPriceBottomSheet.styles.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    priceLabel: {
+      fontSize: 14,
+      color: colors.text.alternative,
+    },
+    priceValue: {
+      fontSize: 16,
+      fontWeight: '500',
+      color: colors.text.default,
+    },
```
gambinish, line 28: Similar note about text styling, I think there are other instances of this elsewhere in the PR that we may want to address as well before merging

app/components/UI/Perps/components/PerpsLimitPriceBottomSheet/PerpsLimitPriceBottomSheet.tsx
```diff
@@ -0,0 +1,286 @@
...
+import { useTheme } from '../../../../../util/theme';
+import Keypad from '../../../Ramp/Aggregator/components/Keypad';
+import { formatPrice } from '../../utils/formatUtils';
+import { createStyles } from './PerpsLimitPriceBottomSheet.styles';
+import { usePerpsPrices } from '../../hooks/usePerpsPrices';
+
+// Default bid/ask spread when real order book data is not available
+// This represents a 0.01% spread (1 basis point) which is typical for liquid markets
+const DEFAULT_BID_MULTIPLIER = 0.9999; // Bid price is 0.01% below current price
+const DEFAULT_ASK_MULTIPLIER = 1.0001; // Ask price is 0.01% above current price
```
gambinish: Can we put this in the HyperliquidConfig?

Also are these value correct? It seems to be slightly different values (precision) from what the comment suggests.

app/components/UI/Perps/components/PerpsLimitPriceBottomSheet/PerpsLimitPriceBottomSheet.tsx
```diff
@@ -0,0 +1,286 @@
...
+        : undefined;
+
+      setPriceSnapshot({
+        currentPrice: current,
+        markPrice: mark,
+        bestBid: bid,
+        bestAsk: ask,
+      });
+    }
+  }, [isVisible]); // eslint-disable-line react-hooks/exhaustive-deps
```
gambinish: Do we want to create a skeleton loading state? I need to test manually. From what I can tell, I'd imagine there is a state where the bottom sheet is open, but the price snapshot is still being loaded.

gambinish: Also, don't we also want to recompute the price snapshot if `currentPriceData` changes? We may want more dependencies in this `useEffect`

app/components/UI/Perps/components/PerpsOrderHeader/PerpsOrderHeader.tsx
```diff
@@ -0,0 +1,146 @@
...
+} from '../../../../../component-library/components/Icons/Icon';
+import Text, {
+  TextVariant,
+  TextColor,
+} from '../../../../../component-library/components/Texts/Text';
+import TokenIcon from '../../../Swaps/components/TokenIcon';
+// import Routes from '../../../../../constants/navigation/Routes'; // No longer needed
+import { formatPrice, formatPercentage } from '../../utils/formatUtils';
+import { HYPERLIQUID_ASSET_ICONS_BASE_URL } from '../../constants/hyperLiquidConfig';
+// import type { PerpsNavigationParamList } from '../../controllers/types'; // No longer needed
```
gambinish: Cleanup: Can remove these commented lines if they aren't being used.

app/components/UI/Perps/components/PerpsOrderHeader/PerpsOrderHeader.tsx
```diff
@@ -0,0 +1,146 @@
...
+      />
+      <View style={styles.headerCenter}>
+        <View style={styles.headerCenterRow}>
+          <TokenIcon
+            symbol={asset}
+            icon={assetIconUrl}
+            style={styles.tokenIcon}
+          />
+          <Text variant={TextVariant.HeadingMD} style={styles.headerTitle}>
+            {asset} {price > 0 ? formatPrice(price) : '$---'}
```
gambinish: This seems like it's a bit brittle. Can we put this in a function to catch some edge cases?

app/components/UI/Perps/components/PerpsOrderTypeBottomSheet/PerpsOrderTypeBottomSheet.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  useEffect(() => {
+    if (isVisible) {
+      bottomSheetRef.current?.onOpenBottomSheet();
+    }
+  }, [isVisible]);
+
+  const orderTypes = [
+    {
+      type: 'market' as const,
```
gambinish: If we create an `OrderType` enum or interface we should be able to avoid having to cast to const here

app/components/UI/Perps/components/PerpsSlider/PerpsSlider.tsx
```diff
@@ -0,0 +1,258 @@
...
+  );
+
+  // Update position when value changes
+  useEffect(() => {
+    if (widthRef.current > 0) {
+      const percentage = (value - minimumValue) / (maximumValue - minimumValue);
+      const newPosition = percentage * widthRef.current;
+      translateX.value = withSpring(newPosition, {
+        damping: 15,
+        stiffness: 400,
```
gambinish: Can probably put this into a config, or at least a variable definition

app/components/UI/Perps/components/PerpsSlider/PerpsSlider.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const { styles } = useStyles(styleSheet, {});
+
+  // Shared values for animations
+  const sliderWidth = useSharedValue(0);
+  const translateX = useSharedValue(0);
+
+  // Convert position to value
+  const positionToValue = useCallback(
+    (position: number, width: number) => {
+      'worklet';
```
gambinish, line 73: What is this string? Is this required?

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+      ? stopLossPrice.replace(/[$,]/g, '')
+      : undefined;
+
+    onConfirm(parseTakeProfitPrice, parseStopLossPrice);
+    onClose();
+  };
+
+  // Quick percentage buttons
+  const takeProfitPercentages = [5, 10, 15, 20];
+  const stopLossPercentages = [5, 10, 15, 20];
```
gambinish: Can probably put this into HyperliquidConfig

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+  const calculatePriceForPercentage = useCallback(
+    (percentage: number, isProfit: boolean) => {
+      if (!currentPrice) return '';
+
+      // For long positions: profit = price up, loss = price down
+      // For short positions: profit = price down, loss = price up
+      const isLong = direction === 'long';
+      const multiplier = isProfit
+        ? (isLong ? 1 + percentage / 100 : 1 - percentage / 100)
+        : (isLong ? 1 - percentage / 100 : 1 + percentage / 100);
```
gambinish: Can we put this into a utility function?

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+        // For profit: long needs higher price, short needs lower price
+        const isValidDirection = isLong ? priceDiff > 0 : priceDiff < 0;
+        return isValidDirection ? percentage.toFixed(2) : `-${percentage.toFixed(2)}`;
+      }
+      // For loss: long needs lower price, short needs higher price
+      const isValidDirection = isLong ? priceDiff < 0 : priceDiff > 0;
+      return isValidDirection ? percentage.toFixed(2) : `-${percentage.toFixed(2)}`;
+    },
+    [currentPrice, direction],
+  );
```
gambinish: Similar with this logic, it would be nice to have this in a utility to make testing easier

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+    if (stopLossPrice) {
+      const slPrice = parseFloat(stopLossPrice.replace(/[$,]/g, ''));
+      if (!isNaN(slPrice)) {
+        // Long: SL must be below current price, Short: SL must be above
+        isValid = isValid && (isLong ? slPrice < currentPrice : slPrice > currentPrice);
+      }
+    }
+
+    return isValid;
+  }, [currentPrice, direction, takeProfitPrice, stopLossPrice]);
```
gambinish: Same here re: utility

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+                    const percentage = calculatePercentageForPrice(
+                      sanitized,
+                      true,
+                    );
+                    setTakeProfitPercentage(percentage);
+                  } else {
+                    setTakeProfitPercentage('');
+                  }
+                  setSelectedTpPercentage(null);
+                }}
```
gambinish: I think that this should be put into a memoized callback to prevent unwanted re-renders

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+                      parseFloat(sanitized),
+                      true,
+                    );
+                    setTakeProfitPrice(price);
+                    setSelectedTpPercentage(parseFloat(sanitized));
+                  } else {
+                    setTakeProfitPrice('');
+                    setSelectedTpPercentage(null);
+                  }
+                }}
```
gambinish: It looks like much of this logic is repeated from above. Can we put this into the same memoized callback and handle the state updates separately?

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+                      parseFloat(sanitized),
+                      false,
+                    );
+                    setStopLossPrice(price);
+                    setSelectedSlPercentage(parseFloat(sanitized));
+                  } else {
+                    setStopLossPrice('');
+                    setSelectedSlPercentage(null);
+                  }
+                }}
```
gambinish: Same thing here about memoizing a callback to be reused

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+                  setSlPriceInputFocused(false);
+                  // Format on blur if there's a value
+                  if (stopLossPrice && !isNaN(parseFloat(stopLossPrice))) {
+                    setStopLossPrice(formatPrice(stopLossPrice));
+                  }
+                }}
+              />
+              <Text variant={TextVariant.BodyMD} color={TextColor.Alternative}>
+                USD
+              </Text>
```
gambinish: Note: It seems like we are hardcoding USD in several places in the Perps directory. We may want to consider having a separate cleanup PR that consolidates these all into a variable definition

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -0,0 +1,528 @@
...
+                placeholderTextColor={colors.text.muted}
+                keyboardType="numeric"
+                onFocus={() => setSlPriceInputFocused(true)}
+                onBlur={() => {
+                  setSlPriceInputFocused(false);
+                  // Format on blur if there's a value
+                  if (stopLossPrice && !isNaN(parseFloat(stopLossPrice))) {
+                    setStopLossPrice(formatPrice(stopLossPrice));
+                  }
+                }}
```
gambinish: Same thing here about the memoized callback

app/components/UI/Perps/hooks/usePerpsPaymentTokens.ts
```diff
@@ -0,0 +1,156 @@
...
+  );
+
+  // Get all chain IDs to search for tokens
+  const allChainIds = useMemo(() => {
+    if (!networkConfigurations) return [];
+    return Object.keys(networkConfigurations) as Hex[];
+  }, [networkConfigurations]);
+
+  // Get all tokens with balances across networks
+  const tokensWithBalance = useTokensWithBalance({ chainIds: allChainIds });
```
gambinish, line 51: We want to be careful that this doesn't brick performance of this hook. No action necessary here, but something to keep in mind.

app/components/UI/Perps/routes/index.tsx
```diff
@@ -2,20 +2,43 @@ import { createStackNavigator } from '@react-navigation/stack';
...
+    <ModalStack.Screen
+      name={Routes.PERPS.MODALS.SLIPPAGE_MODAL}
+      component={PerpsSlippageModal}
+    />
+    <ModalStack.Screen
+      name={Routes.PERPS.MODALS.QUOTE_EXPIRED_MODAL}
+      component={PerpsQuoteExpiredModal}
+    />
+  </ModalStack.Navigator>
+);
```
gambinish: I think that we should seriously consider not using modals, and instead using Screen navigation. The team at Margelo has pointed out that Modals like this a generally an antipattern that gets overused, and can hurt overall performance.

app/components/UI/Perps/services/PerpsConnectionManager.ts
```diff
@@ -0,0 +1,123 @@
...
+  getConnectionState() {
+    return {
+      isConnected: this.isConnected,
+      isConnecting: this.isConnecting,
+      isInitialized: this.isInitialized,
+    };
+  }
+}
+
+export const PerpsConnectionManager = PerpsConnectionManagerClass.getInstance();
```
gambinish, line 123: Can we add unit tests here?

app/components/UI/Perps/services/HyperLiquidSubscriptionService.ts
```diff
@@ -431,6 +459,85 @@ export class HyperLiquidSubscriptionService {
...
+    if (subscription) {
+      subscription.unsubscribe().catch(console.error);
+      this.globalL2BookSubscriptions.delete(symbol);
+      this.orderBookCache.delete(symbol);
+      DevLogger.log(
+        `HyperLiquid: Cleaned up L2 book subscription for ${symbol}`,
+      );
+    }
+  }
+
```
gambinish, line 540: Can we add unit test coverage to these new methods?

app/components/UI/Perps/Views/PerpsView.tsx
abretonc7s: These were just added for quick testing, will be removed probably from your PR

app/components/UI/Perps/hooks/usePerpsPaymentTokens.ts
abretonc7s: This will also be removed completely from yesterday's meeting with confirmation, it seems we will need different flow

app/components/UI/Perps/components/PerpsSlider/PerpsSlider.tsx
abretonc7s: this is reanimated specific and avoid blocking main js thread

app/components/UI/Perps/components/PerpsLimitPriceBottomSheet/PerpsLimitPriceBottomSheet.tsx
abretonc7s: We need to have a snapshot of the price at a given point, I initially added the live prices in the component but it was confusing as the prices were always changing. Eventually I kept it as is, we can discuss and adapt as we try

app/components/UI/Perps/Views/PerpsView.tsx
abretonc7s: specifically the PerpsView is meant to be completely different so Idont think we should care on i18n yet (until we actually implement it)

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
abretonc7s: En fait le paywith risque de sauter completement... pour v1

abretonc7s: Actually it seems for V1 the PayWith may be fully removed which would simplify..

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
abretonc7s: Not particularly it really depends how we want to breakdown the hooks, I initially have each one in separate hook in the POC. Then tried to group them by feature -> useTrading(). I don't think there is bad or good solution but yes we can add a useMarket() hook. But then we need to be consistent and completely break apart the useTrading and only have single minimum hooks.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
abretonc7s: I feel a utility for this would be excessive, this is single line access to the toast.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
abretonc7s: I keep going back and forth on internationalizing or waiting for the screens to be finalized and fix everything at once...

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
abretonc7s: I think we can have a separate skeleton, having 0 at least prevent parsing errors

app/components/UI/Perps/Views/PerpsView.tsx
gambinish: Yeah fair point about PerpsView being our scratchpad. I have a PR that refactors this to match the designs. Can maybe address it there

nickewansmith (changes requested):
This is looking good generally. I found a few things during e2e testing that we might want to look at before merging.

I am unable to edit TP/SL from the positions listing view, although I am able to set it in the actual order that is placed and see it in the success screen. It also does not show in the positions screen though TP/SL was set.

Screenshot 2025-07-22 at 1 46 46 PM

From this positions list view, if I click on a position and go to the position details screen, I am not able to edit TP/SL.

From the positions list, if I click the edit TP/SL button, the popup does show but upon setting values, the values do not persist. Also, the popup screen for the editing TP/SL shows underneath existing positions unless clicking on the edit TP/SL button for the last position listed

https://github.com/user-attachments/assets/921cfbff-5b21-4ec6-9372-0f286548e897

gambinish:
> This is looking good generally. I found a few things during e2e testing that we might want to look at before merging.
> 
> I am unable to edit TP/SL from the positions listing view, although I am able to set it in the actual order that is placed and see it in the success screen. It also does not show in the positions screen though TP/SL was set.
> 
> Screenshot 2025-07-22 at 1 46 46 PM
> From this positions list view, if I click on a position and go to the position details screen, I am not able to edit TP/SL.
> 
> From the positions list, if I click the edit TP/SL button, the popup does show but upon setting values, the values do not persist. Also, the popup screen for the editing TP/SL shows underneath existing positions unless clicking on the edit TP/SL button for the last position listed
> 
>  Simulator.Screen.Recording.-.iPhone.15.Pro.-.2025-07-22.at.13.52.24.mp4

I agree that we should persist the TP/SL on the PositionDetail view.

Additionally, I could set a long position fine, but was getting the "Insufficient Margin" on any short order. Is this expected? Just want to be sure we aren't introducing a short bug here.

abretonc7s:
> This is looking good generally. I found a few things during e2e testing that we might want to look at before merging.
> 
> I am unable to edit TP/SL from the positions listing view, although I am able to set it in the actual order that is placed and see it in the success screen. It also does not show in the positions screen though TP/SL was set.
> 
> Screenshot 2025-07-22 at 1 46 46 PM
> From this positions list view, if I click on a position and go to the position details screen, I am not able to edit TP/SL.
> 
> From the positions list, if I click the edit TP/SL button, the popup does show but upon setting values, the values do not persist. Also, the popup screen for the editing TP/SL shows underneath existing positions unless clicking on the edit TP/SL button for the last position listed
> 
>  Simulator.Screen.Recording.-.iPhone.15.Pro.-.2025-07-22.at.13.52.24.mp4

Yes good catch, I initially thought the edit tpsl was separate but after double checking it is also in this epic. I added it.

nickewansmith (dismissed):
Tested the new changes and It looks good to me! We may need a few more tests to pass coverage. 🤞🏽

nickewansmith (dismissed):

gambinish (approved):

nickewansmith (approved):

PR #17403, opened 2025-07-21
fix: cp-7.51.0 cp-7.52.0 Fix TokenHero component and put back e2e assertions

OGPoyraz:
## **Description**

This PR aims to fix broken `TokenHero` component for native assets. Please see recording that all type of assets are shown properly in `TokenHero` component now.

It also puts back e2e assertions in place. 

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5403

## **Manual testing steps**

1. Try sending ETH
2. See correct amount shown in the confirmation

### **After**

https://github.com/user-attachments/assets/e5c5aebc-8f6e-4963-9b97-452e3df7aa28


app/components/Views/confirmations/legacy/SendFlow/Amount/index.js
```diff
@@ -635,9 +635,15 @@ class Amount extends PureComponent {
...
       selectedAddress,
+      globalNetworkClientId,
     } = this.props;
     try {
-      return await NftController.isNftOwner(selectedAddress, address, tokenId);
+      return await NftController.isNftOwner(
+        selectedAddress,
+        address,
+        tokenId,
+        globalNetworkClientId,
```
OGPoyraz, line 645: This is a crucial fix to make this function work properly for ERC721 - probably broken for long.

e2e/specs/confirmations-redesigned/transactions/send-max-transfer.spec.ts
```diff
@@ -68,12 +66,7 @@ describe(SmokeConfirmationsRedesigned('Send Max Transfer'), () => {
         await AmountView.tapNextButton();
 
         // Check all expected elements are visible
-        await Assertions.checkIfVisible(
-          ConfirmationUITypes.FlatConfirmationContainer,
-        );
-        await Assertions.checkIfVisible(RowComponents.FromTo);
-        await Assertions.checkIfVisible(RowComponents.GasFeesDetails);
-        await Assertions.checkIfVisible(RowComponents.AdvancedDetails);
+        await Assertions.checkIfTextIsDisplayed('1,000 ETH');
```
OGPoyraz, line 69: Putting back amount assertion - others not really needed as we are already checking flow in another e2e tests.

app/components/Views/confirmations/components/rows/transactions/hero-row/hero-row.tsx
```diff
@@ -5,23 +5,26 @@ import { ConfirmationRowComponentIDs } from '../../../../../../../../e2e/selecto
...
   return (
     <View
       style={styles.wrapper}
       testID={ConfirmationRowComponentIDs.TOKEN_HERO}
     >
-      {!isPending && isNft === true && <HeroNft />}
-      {!isPending && isNft === false && <HeroToken amountWei={amountWei} />}
+      {isPending && <LoadingHeroRow />}
+      {!isPending &&
+        (isNft ? <HeroNft /> : <HeroToken amountWei={amountWei} />)}
```
OGPoyraz, line 27: Rather than showing nothing - it's better to have a simple circle with same height and weight for short amount of time.

app/components/Views/confirmations/hooks/nft/useIsNft.ts
```diff
@@ -17,6 +17,6 @@ export const useIsNft = (): { isNft?: boolean; isPending: boolean } => {
   const isNft =
     !isPending && details.standard !== undefined
       ? details.standard !== TokenStandard.ERC20
-      : undefined;
+      : false;
```
OGPoyraz: This line was causing the issue - it was never settled `true` or `false` for native assets but parent component was expecting boolean values.

e2e/specs/confirmations-redesigned/transactions/send-max-transfer.spec.ts
OGPoyraz: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/920ce15e-ec68-4747-bb00-6c16acd5de79

app/components/Views/confirmations/hooks/nft/useIsNft.ts
jpuri: while `isPending` is true returning true or false from here is not correct IMHO, parent component needs to handle loading state

this will return false even if it fails to get token details

OGPoyraz: I agree with you, @jpuri - but even when `isPending` is `false`, this can still end up as `undefined` during a native token transfer. In fact, it should be false, not `undefined`, which is why the conditions below were failing:

image

Returning `undefined` while `isPending` is `true` isn’t a big issue if the parent component is handling it and not rendering the children yet. So as you mentioned, there’s room for optimization here, but there’s no longer a tangible benefit.

jpuri: We need to check native in a separate condition, here defaulting to false can result in situations that token detection failed and we still return false.

If parent miss to check `isPending` which was happening before this resulted in `NFTHero` rendering `undefined` momentarily.

OGPoyraz: > If parent miss to check isPending which was happening before this resulted in NFTHero rendering undefined momentarily.

The problem was not checking `isPending`, it was expecting `isNft` is a `true` or `false` but for native it was returning `undefined`. Now it returns `false` plus it handles loading properly so no more empty `View` anymore.

vinistevam (dismissed):

vinistevam (approved):

jpuri (approved):

PR #17415, opened 2025-07-21
feat: Pass the asset to send page using routing

jpuri:
## **Description**

Passing the asset information to the send page if available.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5407

## **Manual testing steps**
NA

## **Screenshots/Recordings**
Screenshot 2025-07-21 at 5 43 27 PM


jpuri:
Skipping sonar could as test coverage in new components is >90%.

salimtb (approved):

Prithpal-Sooriya (approved):

app/components/Views/confirmations/components/send/send.tsx
```diff
@@ -1,12 +1,65 @@
...
 
-export default Send;
+export const Send = () => {
+  const navigation = useNavigation();
+  const from = useSelector(selectSelectedInternalAccount);
+  const route = useRoute<RouteProp<Record<string, { asset: Asset }>, string>>();
+  const { asset } = route?.params ?? {};
+  const { styles } = useStyles(styleSheet, {});
+
+  const submitSend = useCallback(() => {
```
matthewwalsh0, line 27: Minor, we try and name the handlers based on the events so `handleSubmitPress` or `handleCancelPress` etc.

app/components/Views/confirmations/utils/confirm.ts
```diff
@@ -18,3 +18,6 @@ export function isStakingConfirmation(requestType: string) {
     TransactionType.stakingClaim,
   ].includes(requestType as TransactionType);
 }
+
+export const isSendRedesignEnabled = () =>
```
matthewwalsh0, line 22: I think we should be very careful to avoid coupling any of this code to the confirmations code.

Should this be in a new `/utils/send.ts`?

app/components/Nav/Main/MainNavigator.js
```diff
@@ -622,7 +623,11 @@ const SendView = () => (
 
 const SendComponent = () => (
   <Stack.Navigator headerMode="screen">
-    <Stack.Screen name="Send" component={Send} />
+    <Stack.Screen name={Routes.SEND.ROOT} component={Send} />
+    <Stack.Screen
```
matthewwalsh0, line 627: Why is the redesigned confirmation in the send component?

app/components/UI/AssetOverview/AssetOverview.tsx
```diff
@@ -234,7 +235,16 @@ const AssetOverview: React.FC<AssetOverviewProps> = ({
     } else {
       dispatch(newAssetTransaction(asset));
     }
-    navigation.navigate('SendFlowView', {});
+    if (isSendRedesignEnabled()) {
```
matthewwalsh0, line 238: Rather than repeating this code and burden for every caller, could we make a simple `useSendFlow` hook that returns a `startSendFlow` function for example, that encapsulates the navigation, the routes, and the decision of whether we load the redesign or legacy?

matthewwalsh0 (changes requested):

app/components/Nav/Main/MainNavigator.js
jpuri: this is review page to which we navigate after creating transaction, adding it to same navigator helps to smooth page opening and also context keep state of confirm page when user comes back by clicking cancel.

next PR will add context on top of this

app/components/UI/AssetOverview/AssetOverview.tsx
jpuri: Not all components are functional react components, thus using hook is bit hard. Additionally each case is different in asset we pass to route.

matthewwalsh0: Four of the usages in this PR are in functional components, and naturally the asset would be an argument on the `startSendFlow` method.

It's much easier for us to maintain if we encapsulate as much knowledge as we can, and therefore minimise the coupling to other components.

This is a standard pattern as we already have `useSwapBridgeNavigation` for example.

matthewwalsh0: To be added in subsequent PR 👍

matthewwalsh0 (approved):

PR #17515, opened 2025-07-23
fix: eth_sendTransaction proper validation

ffmcgee725:
## **Description**

In the [following code](https://github.com/MetaMask/metamask-mobile/blob/60c955494345698d08d2d600e9a4b2bb4ae387e8/app/core/RPCMethods/eth_sendTransaction.ts#L106), `nChainId` will always resolve to '0x0' because no `chainId` property currently exists. So the suggested change retrieves the network client id from `req.networkClientId` and converts it into a valid `chainId`.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/4898

## **Manual testing steps**

1. Go to [Metamask Test Dapp](https://metamask.github.io/test-dapp/) via in app browser.
2. Connect to an EVM network.
3. Press send legacy transaction button (or any transaction button that calls `eth_sendTransaction`). Behaviour should stay the same functionally.

### DIFF
--- a/app/core/RPCMethods/eth_sendTransaction.ts
+++ b/app/core/RPCMethods/eth_sendTransaction.ts
@@ -11,6 +11,7 @@ import {
 } from '@metamask/transaction-controller';
 import { rpcErrors } from '@metamask/rpc-errors';
 import ppomUtil, { PPOMRequest } from '../../lib/ppom/ppom-util';
+import Engine from '../Engine';
 
 /**
  * A JavaScript object that is not `null`, a function, or an array.
@@ -99,13 +100,19 @@ async function eth_sendTransaction({
       message: `Invalid parameters: expected the first parameter to be an object`,
     });
   }
+
+  const networkConfig =
+    Engine.context.NetworkController.getNetworkConfigurationByNetworkClientId(
+      req.networkClientId,
+    );
+
   // TODO: Normalize chainId to Hex string
-  const nChainId = parseInt(req.networkClientId, 16);
+  const nChainId = parseInt(networkConfig?.chainId || '0x0', 16);
+
   await validateAccountAndChainId({
     from: req.params[0].from,
     chainId: nChainId,
   });
-
 
   const { result, transactionMeta } = await sendTransaction(req.params[0], {
     deviceConfirmedOn: WalletDevice.MM_MOBILE,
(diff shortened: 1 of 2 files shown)


adonesky1:
When you say no `chainId` property currently exists do you mean that we remove it upstream if the dapp sends it? There are some wallets that do accept `chainId` as part of an `eth_sendTransaction` request (looks [like wallet_connect does](https://github.com/ethereum/EIPs/issues/1637)). We don't but I guess I"m wondering have you verified that we remove extraneous properties upstream or theoretically could it have been sent and used here before?

ffmcgee725:
@adonesky1 

What I meant to convey is that I could only confirm the following
- calls made from our test dapps are either not passing in the `chainId` in params, OR it's what you said about us removing the field upstream (mere assumption, I have not manually confirmed this);
- the field is also marked as optional AND deprecated in our docs
```ts
/**
 * Standard data concerning a transaction to be processed by the blockchain.
 */
export type TransactionParams = {
    /**
     * Network ID as per EIP-155.
     *
     * @deprecated Ignored.
     * Use `networkClientId` when calling `addTransaction`.
     */
    chainId?: Hex;
```

I remember back when we created this issue you advised me that for this validation we should proceed as per the advise on the original issue ticket.

Since some wallets accept `chainId` as part of params, would you rather we do something like a conditional usage of the field in case it's defined ?

```ts
const networkConfig = 
  Engine.context.NetworkController.getNetworkConfigurationByNetworkClientId(
    req.networkClientId,
  );
const nChainId =
   req.params[0].chainId ?? parseInt(networkConfig?.chainId || '0x0', 16);
```

app/core/RPCMethods/eth_sendTransaction.ts
```diff
@@ -99,11 +100,15 @@ async function eth_sendTransaction({
       message: `Invalid parameters: expected the first parameter to be an object`,
     });
   }
+
+  const networkConfig =
+    Engine.context.NetworkController.getNetworkConfigurationByNetworkClientId(
+      req.networkClientId,
```
adonesky1, line 106: sorry last question here, you've validated that networkClientId is always defined here?

ffmcgee725: Seems to be the case

PR #17545, opened 2025-07-24
feat: Metamask mobile perps withdrawal

abretonc7s:
## **Description**

This PR implements the withdrawal functionality for the Perps trading feature, allowing users to withdraw USDC from their HyperLiquid perps account back to their Arbitrum wallet.

Key features:
- Direct withdrawal flow with amount input screen
- Real-time fee calculation showing $1 HyperLiquid fee 
- Toast notifications for withdrawal progress
-  Simplified and aligned deposit flow to match withdrawal

## **Changelog**

CHANGELOG entry: Added withdrawal functionality for Perps trading, allowing users to withdraw USDC from their perps account to Arbitrum

## **Related issues**

Fixes: TAT-1021

## **Manual testing steps**

1. Navigate to the Perps trading view and ensure you have a positive balance
2. Click the "Withdraw" button next to your perp balance
3. Enter an amount to withdraw (must be greater than $1.01 to cover fees)
4. Verify the fee breakdown shows:
   - $1 HyperLiquid withdrawal fee
5. Tap "Review" to see the withdrawal summary
6. Tap "Withdraw" to initiate the withdrawal
7. Observe the toast notification showing withdrawal progress
8. Wait ~5 minutes for the withdrawal to complete
9. Verify the USDC balance increases on Arbitrum

### **Before**

No withdrawal functionality was available.

### **After**

- Withdrawal screen with amount input and fee breakdown
- Toast notifications for withdrawal status

Simulator Screen Recording - iPhone 16 Pro - 2025-07-25 at 00 10 05


app/components/UI/Perps/Views/PerpsView.test.tsx
```diff
@@ -1,311 +1,44 @@
...
-      resetError: mockResetError,
-    });
-
-    render(<PerpsView />);
-
-    const retryButton = screen.getByText('Retry Connection');
-    fireEvent.press(retryButton);
-
-    expect(mockResetError).toHaveBeenCalled();
-    expect(mockConnect).toHaveBeenCalled();
```
gambinish, line 309: Just double checking that we want to remove these. I realize this is now a sandbox, but it might still be useful to have tests here for regressions?

I'm fine with removing it if you feel that it's redundant

app/components/UI/Perps/Views/PerpsView.tsx
```diff
@@ -337,6 +352,23 @@ const PerpsView: React.FC<PerpsViewProps> = () => {
             style={styles.button}
           />
 
+          {/* Show withdraw button only if there's available balance */}
+          {cachedAccountState?.availableBalance &&
+            parseFloat(
+              cachedAccountState.availableBalance
+                .replace('$', '')
+                .replace(',', ''),
```
gambinish: I'd prefer to handle this in a way without doing string replace. Can we stick this in a utility, and maybe leverage some of our other formatting utils instead?

app/components/UI/Perps/Views/PerpsView.tsx
```diff
@@ -337,6 +352,23 @@ const PerpsView: React.FC<PerpsViewProps> = () => {
...
+            ) > 0 && (
+              <Button
+                variant={ButtonVariants.Secondary}
+                size={ButtonSize.Lg}
+                width={ButtonWidthTypes.Full}
+                label={strings('perps.withdrawal.title')}
+                onPress={() => navigation.navigate(Routes.PERPS.WITHDRAW)}
```
gambinish: I think this might be better to put in a useCallback hook

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
```diff
@@ -0,0 +1,380 @@
...
+        'Your funds should arrive within 5 minutes',
+      'perps.withdrawal.error': 'Withdrawal failed',
+      'perps.withdrawal.error_generic': 'An error occurred during withdrawal',
+      'perps.withdrawal.invalid_amount': 'Please enter a valid amount',
+      'perps.deposit.max_button': 'Max',
+      'perps.deposit.done_button': 'Done',
+    };
+    return translations[key] || key;
+  }),
+}));
```
gambinish: We shouldn't need to mock these. I think if you remove it the test will still pass. This could lead to false positive tests if the strings need to change

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
```diff
@@ -0,0 +1,380 @@
...
+    }) => (
+      <View testID="perps-quote-details">
+        <Text testID="network-fee">{networkFee}</Text>
+        <Text testID="estimated-time">{estimatedTime}</Text>
+        <Text testID="rate">{rate}</Text>
+        <Text testID="metamask-fee">{metamaskFee}</Text>
+      </View>
+    ),
+  };
+});
```
gambinish: We should be able to remove this as well, and just test the depandent component itself

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
```diff
@@ -0,0 +1,380 @@
...
+  return {
+    __esModule: true,
+    TokenInputArea,
+    TokenInputAreaType: {
+      Source: 'source',
+      Destination: 'destination',
+    },
+    MAX_INPUT_LENGTH: 10,
+  };
+});
```
gambinish: if possible, I think that we should modify the test so that we test whatever UI gets rendered from the test runner rather than rebuilding the component in a mock here.

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,445 @@
...
+              tokenType={TokenInputAreaType.Destination}
+            />
+          </Box>
+
+          {shouldDisplayQuoteDetails && (
+            <Box style={styles.quoteContainer}>
+              <PerpsQuoteDetailsCard
+                networkFee={formattedQuoteData.networkFee}
+                estimatedTime={formattedQuoteData.estimatedTime}
+                rate={`1 ${USDC_SYMBOL} = 1 ${USDC_SYMBOL}`}
```
gambinish, line 368: This comparison seems odd, and doesn't tell me much as a user. Should these values come from state? Would we ever expect this to change?

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,445 @@
...
+        {isInputFocused && (
+          <View style={styles.floatingKeypadContainer}>
+            <Button
+              variant={ButtonVariants.Primary}
+              label={getButtonLabel()}
+              onPress={handleContinue}
+              style={styles.actionButton}
+              disabled={!hasValidInputs || isSubmittingTx}
+              loading={isSubmittingTx}
+              testID="continue-button"
```
gambinish: Nit: This testId should be sourced from the e2e/selectors file along with the other Perps testIds

app/components/UI/Perps/services/HyperLiquidWalletService.ts
```diff
@@ -44,6 +45,18 @@ export class HyperLiquidWalletService {
             }
             return [selectedAddress];
           }
+          case 'eth_chainId': {
+            // Return Arbitrum chain ID in hex format
+            // HyperLiquid operates on Arbitrum
+            const chainId = getChainId(this.isTestnet);
+            const hexChainId = `0x${parseInt(chainId, 10).toString(16)}`;
```
gambinish: I think we have an existing `toHexAddress` util we can use here, rather than doing a parseInt

gambinish (changes requested):
I've done an initial pass here. I'll do a deeper review tomorrow and test full withdraw functionality on my device.

app/components/UI/Perps/Views/PerpsView.tsx
abretonc7s: We should keep the callback for complex logic, when inlining I don't think it brings anything.

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
abretonc7s: This doesn't really do much currently, but can be useful when we withdraw + swap or bridge. I think we can leave it there for now and adjust based on feedback during internal testing.

app/components/UI/Perps/Views/PerpsView.test.tsx
abretonc7s: can remove

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
abretonc7s: I havent addressed all the comments on unit tests, I want to run a separate agents later to fix all of them consistently for all perps screen. for now priority is to get the coverage even if the tests arent perfect

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
abretonc7s: I havent addressed all the comments on unit tests, I want to run a separate agents later to fix all of them consistently for all perps screen. for now priority is to get the coverage even if the tests arent perfect

app/components/UI/Perps/hooks/useWithdrawValidation.ts
```diff
@@ -0,0 +1,96 @@
...
+  const cachedAccountState = usePerpsAccount();
+  const perpsNetwork = usePerpsNetwork();
+  const isTestnet = perpsNetwork === 'testnet';
+
+  // Available balance from perps account
+  const availableBalance = useMemo(() => {
+    const balance = cachedAccountState?.availableBalance || '0';
+    // Remove $ and parse
+    return balance.replace('$', '').replace(',', '');
+  }, [cachedAccountState]);
```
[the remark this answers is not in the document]
gambinish: I think that we should fix this to use `Intl.NumberFormat` rather than doing string replace

gambinish:
I've pushed a couple of fixes:

1. Hyperliquid USDC balance in TabView should show `accountState.availableBalance`, to align with balance shown on withdraw view
2. Moved `navigation.goback()` into catch block to prevent withdraw toast message from persisting infinitely (could not be dismissed)

gambinish (dismissed):
LGTM

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,449 @@
...
+              networkImageSource={destNetworkImage}
+              networkName="Arbitrum"
+              testID={PerpsWithdrawViewSelectorsIDs.DEST_TOKEN_AREA}
+              tokenType={TokenInputAreaType.Destination}
+            />
+          </Box>
+
+          {shouldDisplayQuoteDetails && (
+            <Box style={styles.quoteContainer}>
+              <PerpsQuoteDetailsCard
```
Matt561, line 365: The `'perps.deposit.metamask_fee_tooltip` local string used in `PerpsQuoteDetailsCard` currently has the word "deposit" hardcoded which doesn't make sense in the context of the Withdrawal view.

We'll likely want to update the `PerpsQuoteDetailsCard` to pass a param for the action (deposit/withdrawal) so that it makes sense contextually.

https://github.com/user-attachments/assets/6d7bcace-e44b-492d-9b26-0c68ff6a65c2

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,449 @@
...
+                  variant={ButtonVariants.Secondary}
+                  size={ButtonSize.Md}
+                  label="25%"
+                  onPress={() => handlePercentagePress(0.25)}
+                  style={styles.percentageButton}
+                />
+                <Button
+                  variant={ButtonVariants.Secondary}
+                  size={ButtonSize.Md}
+                  label="Max"
```
Matt561: Let's replace this hardcoded English string with a local variable.

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,449 @@
...
+                  onPress={handleMaxPress}
+                  style={styles.percentageButton}
+                />
+                <Button
+                  variant={ButtonVariants.Secondary}
+                  size={ButtonSize.Md}
+                  label="Done"
```
Matt561: Let's replace this hardcoded English string with a local variable.

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,449 @@
...
+                estimatedTime={formattedQuoteData.estimatedTime}
+                rate={`1 ${USDC_SYMBOL} = 1 ${USDC_SYMBOL}`}
+                metamaskFee={METAMASK_WITHDRAWAL_FEE_PLACEHOLDER}
+              />
+            </Box>
+          )}
+        </ScrollView>
+
+        {isInputFocused && (
+          <View style={styles.floatingKeypadContainer}>
```
Matt561, line 377: I don't think the continue button and percentage buttons are supposed to be displayed at the same time. My understanding of the designs is that the withdraw button replaces the percentage button when a valid amount is entered. This would also mean that the error message/minimum value message would need to go somewhere else.


image

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
```diff
@@ -0,0 +1,650 @@
...
+  onInputPress?: () => void;
+}
+
+jest.mock('../../../../UI/Bridge/components/TokenInputArea', () => {
+  const MockReact = jest.requireActual('react');
+  const View = jest.requireActual('react-native').View;
+  const TouchableOpacity = jest.requireActual('react-native').TouchableOpacity;
+  const Text = jest.requireActual('react-native').Text;
+
+  const TokenInputArea = MockReact.forwardRef(
```
Matt561, line 96: Are we mocking this component because it relies on external requests/calls?

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
```diff
@@ -0,0 +1,650 @@
...
+    TokenInputAreaType: {
+      Source: 'source',
+      Destination: 'destination',
+    },
+    MAX_INPUT_LENGTH: 10,
+  };
+});
+
+// Mock Keypad component
+jest.mock('../../../Ramp/Aggregator/components/Keypad', () => {
```
Matt561, line 144: How come we are mocking the components in this view? Wouldn't we want to use them as is to avoid needing to maintain these mocks and ensure breaking changes to underlying components are discovered immediately?

Unless there is an external request we're trying to avoid? If yes, wouldn't it make more sense to mock the external call?

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,449 @@
...
+          <Box style={styles.inputsContainer} gap={8}>
+            <TokenInputArea
+              ref={inputRef}
+              amount={withdrawAmount || undefined}
+              token={sourceToken}
+              tokenBalance={availableBalance}
+              networkImageSource={getNetworkImageSource({
+                chainId: sourceToken.chainId,
+              })}
+              networkName="Hyperliquid"
```
Matt561: Is this string currently or expected to be used a lot? Would it make sense to break it out into a constant?

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
```diff
@@ -0,0 +1,449 @@
...
+              <Box style={styles.arrowCircle}>
+                <Icon name={IconName.Arrow2Down} size={IconSize.Xl} />
+              </Box>
+            </Box>
+
+            <TokenInputArea
+              amount={destAmount}
+              token={destToken}
+              networkImageSource={destNetworkImage}
+              networkName="Arbitrum"
```
Matt561: Is this string currently or expected to be used a lot? Would it make sense to break it out into a constant?

app/components/UI/Perps/constants/hyperLiquidConfig.ts
```diff
@@ -101,6 +103,14 @@ export const FEE_RATES: FeeRatesConfig = {
 // MetaMask fee for deposits (temporary placeholder)
 export const METAMASK_DEPOSIT_FEE = '$0.00';
 
+// Withdrawal fees
+export const HYPERLIQUID_WITHDRAWAL_FEE = 1; // $1 USD fixed fee
+export const METAMASK_WITHDRAWAL_FEE = 0; // $0 - no MM fee for withdrawals
+export const METAMASK_WITHDRAWAL_FEE_PLACEHOLDER = '$0.00'; // Display format
+
+// Withdrawal timing
+export const WITHDRAWAL_ESTIMATED_TIME = '5 minutes';
```
Matt561, line 112: Let's use a local string variable instead of a hardcoded English string.

app/components/UI/Perps/constants/perpsConfig.ts
```diff
@@ -8,3 +8,13 @@ export const PERPS_CONSTANTS = {
...
 } as const;
+
+/**
+ * Withdrawal-specific constants
+ */
+export const WITHDRAWAL_CONSTANTS = {
+  DEFAULT_MIN_AMOUNT: '1.01', // Default minimum withdrawal amount in USDC
+  DEFAULT_FEE_AMOUNT: 1, // Default withdrawal fee in USDC
+  DEFAULT_FEE_TOKEN: 'USDC', // Default fee token
+  DEFAULT_ESTIMATED_TIME: '5 minutes', // Default estimated withdrawal time
```
Matt561: Let's use a local string variable instead of a hardcoded English string.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -641,11 +641,128 @@ export class PerpsController extends BaseController<
...
+          error,
+          reason: 'Missing assetId',
+          params,
+        });
+        this.update((state) => {
+          state.lastError = error;
+          state.lastUpdateTimestamp = Date.now();
+        });
+        return { success: false, error };
+      }
```
Matt561: Minor Nit: Creating a method to handle and log errors would make this function more concise.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -641,11 +641,128 @@ export class PerpsController extends BaseController<
...
+
+      if (!params.amount || parseFloat(params.amount) <= 0) {
```
Matt561: Minor nit: `params.amount` accepts any `string` but if `parseFloat` fails and returns `NaN` it'll be considered valid.

app/components/UI/Perps/hooks/usePerpsWithdrawQuote.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  const formattedQuoteData = useMemo<FormattedQuoteData>(() => {
+    // Parse amount
+    const amountNum = parseFloat(amount || '0');
+
+    // Get fees from route constraints or use defaults
+    const networkFee =
+      withdrawalRoute?.constraints?.fees?.fixed ??
+      WITHDRAWAL_CONSTANTS.DEFAULT_FEE_AMOUNT;
+    const metamaskFee = METAMASK_WITHDRAWAL_FEE; // $0 currently
```
Matt561, line 64: If `metamaskFee` expected to also be in USDC? Asking because comment mentions fiat.

app/components/UI/Perps/hooks/useWithdrawTokens.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  // Destination token (Arbitrum USDC)
+  const destToken = useMemo<PerpsToken>(() => {
+    const arbitrumChainId = toHex(
+      parseInt(ARBITRUM_MAINNET_CHAIN_ID, 10),
+    ) as Hex;
+
+    const baseToken: PerpsToken = {
+      symbol: USDC_SYMBOL,
+      address: ZERO_ADDRESS, // Will be actual USDC address on Arbitrum
```
Matt561: Should we replace the `ZERO_ADDRESS` in this PR?

Matt561 (commented):
@abretonc7s I've left some comments! Please ping me or the perps tech channel when you need another review.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -641,11 +641,128 @@ export class PerpsController extends BaseController<
...
+        params,
+        timestamp: new Date().toISOString(),
+        assetId: params.assetId,
+        amount: params.amount,
+        destination: params.destination,
+        activeProvider: this.state.activeProvider,
+        isTestnet: this.state.isTestnet,
+      });
+
+      // Validate required parameters
```
Matt561: I think we may be validating parameters in the underlying (hyperliquid) provider as well. Is it intended to validate in the PerpsController and underlying separately? Is this needed if params are the same?

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
abretonc7s: no but I want to do a full pass at all the unit tests at once during hardening and clean them up. Currently mostly generating them to get them to pass and get the PR to merge. Especially if we have considable changes coming after testing, test probably will have to be adjusted anyway

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
abretonc7s: I changed the prefix for i18n to perps.quote.*

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
abretonc7s: yes I left there intentionally, since the design will change to account for the error message

app/components/UI/Perps/controllers/PerpsController.ts
abretonc7s: yes good point, I have simplified. Keeping all the devlogger for now until we get to hardening the features.

app/components/UI/Perps/hooks/usePerpsWithdrawQuote.ts
abretonc7s: We won't have fee intiially, so I think better to handle this once we get a first version out for people to test

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
abretonc7s: Especially make sure we are consistent accross all ui components, not overmock etc. It is easier to do it once for all files then keep going back and forth on each PR.

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.tsx
Matt561: Sounds good!

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
Matt561: No worries, this makes sense.

app/components/UI/Perps/Views/PerpsWithdrawView/PerpsWithdrawView.test.tsx
Matt561: Resolving based on discussion in [this](https://github.com/MetaMask/metamask-mobile/pull/17545#discussion_r2238077315) thread.

Matt561 (approved):

PR #17551, opened 2025-07-24
fix: refactoring in send code

jpuri:
## **Description**

PR is addressing feedbacks on previous PR: https://github.com/MetaMask/metamask-mobile/pull/17415

## **Changelog**

CHANGELOG entry:

## **Related issues**
NA

## **Manual testing steps**
NA

## **Screenshots/Recordings**
NA


app/components/Views/confirmations/utils/send.ts
```diff
@@ -0,0 +1,23 @@
...
+    navigation.navigate(Routes.SEND.DEFAULT, {
+      screen: Routes.SEND.ROOT,
+      params: {
+        asset,
+      },
+    });
+  } else {
+    navigation.navigate('SendFlowView', {});
+  }
+};
```
jpuri, line 26: Hey @matthewwalsh0 : I added this utility function here. I am still a bit doubtful if this is good approach as it changes existing flow - which I prefer to avoid. This will also break some existing tests as at some places parameter to `SendFlowView` is `{}` and at others it is not passed.

Once we get rid of old send all of the pages calling this function will actually have refactor to cleanup adding transaction to state.

matthewwalsh0: Ideally this would be in a hook such as `useSendNavigation` or `useSendFlow` to avoid the caller having to provide the navigation object, and it could return a less explicit function such as `startSendFlow`.

We could still also export this function directly from the same hook though, for any non-functional components.

We have an opportunity to fix lots of things that are less than ideal with the legacy send flow, so unfortunately I think breaking existing tests is a worthwhile trade-off to allow us to do this unhindered.

The benefit here is simply decoupling and encapsulating knowledge. It lets us change the routes, the parameter format, add metrics; all without any changes to the many (4+) callers.

app/components/Views/confirmations/context/send-context/send-context.tsx
```diff
@@ -87,7 +87,7 @@ export const SendContextProvider: React.FC<{
     setTransactionParams({ ...transactionParams, ...params });
   };
 
-  const submitSend = useCallback(async () => {
+  const handleSubmitPress = useCallback(async () => {
```
matthewwalsh0, line 90: If these functions are now on a context, they shouldn't use the `handleX` syntax but be meaningful to the context and the outcome.

matthewwalsh0, line 90: This context should strictly provide a means to access and update the shared state, such as selected asset, destination, and amount.

Any logic such as this should be in a dedicated hook such as `useSubmitSend` so it can be abstracted and tested directly.

We also need to be careful to abstract the logic based on protocol, so will likely also need a `useSubmitSendEVM`.

app/components/Views/confirmations/utils/send.ts
```diff
@@ -0,0 +1,38 @@
+import { Nft } from '@metamask/assets-controllers';
+
+import Routes from '../../../../constants/navigation/Routes';
+import { AssetType } from '../types/token';
+
+interface NavigateType {
```
matthewwalsh0: Do we need to define this, there must be something internal such as `NavigationProp`?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/context/send-context/send-context.tsx
jpuri: The next PR moves these into their own hook [here](https://github.com/MetaMask/metamask-mobile/pull/17621/files#diff-41fe88aa94415a3020ffc482800dbd674506329d7c3b515c9ccd82a499d86edc).

jpuri: The next PR moves these into their own hook [here](https://github.com/MetaMask/metamask-mobile/pull/17621/files#diff-41fe88aa94415a3020ffc482800dbd674506329d7c3b515c9ccd82a499d86edc).

app/components/Views/confirmations/utils/send.ts
jpuri: I did checked but I could not find anything exported from navigation library. But I simplified this type declaration, and it looks better now.

app/components/Views/confirmations/context/send-context/send-context.tsx
jpuri: I am working on another PR to add infrastructure for non-evm networks.

jpuri:
I added [skip-sonar-cloud](https://github.com/MetaMask/metamask-mobile/issues?q=state%3Aopen%20label%3Askip-sonar-cloud) to the PR as there are small changes in existing code files that is causing sonar to complain.

matthewwalsh0 (approved):

OGPoyraz (approved):

salimtb (approved):

PR #17554, opened 2025-07-24
feat: Amount validation on send page

jpuri:
## **Description**

PR adds validations for amount field on send page.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5401

## **Manual testing steps**

1. Enable new send flow locally
2. Enter invalid amount value
3. Check error displayed

## **Screenshots/Recordings**
Screenshot 2025-07-24 at 6 22 13 PM
Screenshot 2025-07-24 at 6 22 05 PM


app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  asset,
+  contractBalances,
+  from,
+}: ValidateAmountArgs) => {
+  if (!asset) {
+    return;
+  }
+  if (amount === undefined || amount === null || amount === '') {
+    return;
+  }
```
OGPoyraz, line 68: Would it make sense to not hit validation if values are not defined? In fact this could be "amount" component responsibility

app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export interface ValidateAmountArgs {
+  accounts: Record<Hex, AccountInformation>;
+  amount?: string;
+  asset?: AssetType;
+  contractBalances: Record<Hex, Hex>;
+  from: Hex;
+}
+
+export const validateAmount = ({
```
OGPoyraz, line 56: Wouldn't it be easier to have `validateEVMAmount` `validateNonEVMAmount` etc? So that we can have clear separation and easier to maintainable validators per blockchain.

app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+}: ValidateAmountArgs) => {
+  if (!asset) {
+    return;
+  }
+  if (amount === undefined || amount === null || amount === '') {
+    return;
+  }
+  if (!isDecimal(amount) || Number(amount) < 0) {
+    return strings('transaction.invalid_amount');
+  }
```
OGPoyraz, line 71: Rather than having one big validation function, can we separate smaller re-usable validation function like we did it in [gas](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/confirmations/utils/validations/gas.ts#L4) or [approve](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/confirmations/utils/validations/approve.ts#L74) .

For example `validateEVMAmount` consist of 
- `validateEVMTokenAmount`
- `validateEVMNativeAmount`
- `validateSufficientAmount`
etc

That way we can easily extend chain agnostic validations much easier.

app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  let weiValue;
+  let weiBalance;
+  if (isNativeToken(asset)) {
```
OGPoyraz, line 74: Could we separate this if clause into `validateEVMNativeAmount`?

app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    ) as Hex;
+    const account = accountAddress ? accounts[accountAddress] : undefined;
+    // toWei can throw error if input is not a number: Error: while converting number to string, invalid number value
+    try {
+      weiValue = toWei(amount);
+    } catch (error) {
+      return strings('transaction.invalid_amount');
+    }
+    weiBalance = hexToBN(account?.balance ?? '0');
+  } else {
```
OGPoyraz, line 86: And `validateEVMTokenAmount` for this else block?

app/components/Views/confirmations/context/send-context/utils.ts
OGPoyraz, line 56: Also can we move this validation into `app/components/Views/confirmations/utils/validations/send` ?

app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  transactionParams: TransactionParams,
+) => {
+  const { from, to, value } = transactionParams;
+  const trxnParams: TransactionParams = { from };
+  if (isNativeToken(asset)) {
+    trxnParams.data = '0x';
+    trxnParams.to = to;
+    trxnParams.value = BNToHex(toWei(value ?? '0') as unknown as BN);
+  } else if (asset.tokenId) {
+    trxnParams.data = generateTransferData('transferFrom', {
```
OGPoyraz, line 29: Minor - shall we put a small comment this block is for NFTs and the one below is for ERC20?

app/components/Views/confirmations/context/send-context/send-context.tsx
```diff
@@ -95,7 +92,7 @@ export const SendContextProvider: React.FC<{
     const networkClientId = NetworkController.findNetworkClientIdByChainId(
       toHex(chainId),
     );
-    const trxnParams = prepareTransaction(asset, transactionParams);
+    const trxnParams = prepareEVMTransaction(asset, transactionParams);
```
OGPoyraz, line 95: This wasn't introduced in this PR, but would it make sense to have a `submitEVMTransaction` utility and move all EVM-related logic there? 
It would be beneficial to encapsulate all blockchain-related logic instead of keeping it within this component.

app/components/Views/confirmations/context/send-context/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+import {
+  hexToBN,
+  isDecimal,
+  toTokenMinimalUnit,
+  toWei,
+} from '../../../../../util/number';
+import { AssetType } from '../../types/token';
+import { isNativeToken } from '../../utils/generic';
+
+export const prepareEVMTransaction = (
```
OGPoyraz, line 18: Can we also relocate this into `app/components/Views/confirmations/utils/send` ?

OGPoyraz (changes requested):

app/components/Views/confirmations/context/send-context/utils.ts
jpuri: Imagine if I entered a wrong value and then cleared it, error should go away.

app/components/Views/confirmations/context/send-context/utils.ts
jpuri: This function is only about 30 lines, I would avoid breaking it down further.

app/components/Views/confirmations/context/send-context/utils.ts
jpuri: I am adding a hook for this function in upcoming PR. I would prefer to keep it in the hook.

app/components/Views/confirmations/context/send-context/utils.ts
jpuri: Le me do that as part of next PR.

app/components/Views/confirmations/context/send-context/send-context.tsx
jpuri: Upcoming PR is moving it to utils. I would prefer to delay renaming function to include `EVM` as it would be best to do when we add other networks, so there is more clarity of how code is structured.

app/components/Views/confirmations/context/send-context/utils.ts
jpuri: Yep next PR is already doing that

OGPoyraz (commented):
Lifting off my CR as I will be off for next two days, thanks for addressing suggestions @jpuri ❤️

app/components/Views/confirmations/context/send-context/utils.ts
jpuri: I am working on another PR to add infrastructure for non-evm send.

app/components/Views/confirmations/context/send-context/send-context.tsx
jpuri: That sounds good, I am working on another part to add infrastructure for non-evm send and will take it up as part of that.

app/components/Views/confirmations/context/send-context/utils.ts
OGPoyraz: In that case should we have a validation called `validateValueExists`? Anyways - it's not a blocker and minor we can fine tune later if needed.

OGPoyraz (dismissed):

OGPoyraz (approved):

PR #17566, opened 2025-07-24
feat: account discovery bitcoin

darioAnongba:
## **Description**

Add Bitcoin to account discovery and make the discovery chain agnostic by using config per Snap.

## **Changelog**

CHANGELOG entry: add Bitcoin to account discovery

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/NWNT-431

## **Manual testing steps**

1. Import wallet with Bitcoin account (ideally of multiple account types like segwit and taproot)
2. See accounts appear with right balance in account menu
3. Import SRP and verify similar process (SRP 2)


app/core/SnapKeyring/MultichainWalletSnapClient.ts
```diff
@@ -233,6 +252,7 @@ export abstract class MultichainWalletSnapClient {
               {
                 scope,
                 entropySource,
+                synchronize: false,
```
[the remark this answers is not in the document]
darioAnongba: resolved, synchronize is true by default.

app/core/Authentication/Authentication.ts
```diff
@@ -130,63 +127,63 @@ class AuthenticationService {
...
-      // Store flag to retry on next unlock
-      StorageWrapper.setItem(SOLANA_DISCOVERY_PENDING, TRUE);
-    });
-    ///: END:ONLY_INCLUDE_IF
+
+    for (const clientType of Object.values(WalletClientType)) {
+      const { discoveryStorageId } = WALLET_SNAP_MAP[clientType];
+
+      this.attemptAccountDiscovery(clientType).catch((error) => {
+        console.warn('Account discovery failed during wallet creation:', error);
```
orestiseth: nit: should we include the `clientType` here or is it going to be included on the `error`?

aganglada (dismissed):
Love the change ❤️

zone-live (dismissed):

app/core/Authentication/Authentication.ts
```diff
@@ -130,63 +127,63 @@ class AuthenticationService {
...
+    for (const clientType of Object.values(WalletClientType)) {
+      const { discoveryStorageId } = WALLET_SNAP_MAP[clientType];
+
+      try {
+        const isPending = await StorageWrapper.getItem(discoveryStorageId);
+        if (isPending === TRUE) {
+          await this.attemptAccountDiscovery(clientType);
+        }
+      } catch (error) {
+        console.warn('Failed to check/retry discovery:', error);
```
orestiseth: I think here we can gather all the pending items and then fire a promise.all to execute them in parallel. Obviously, it makes sense only if we can have a good number of pending tasks and/or `attemptAccountDiscovery` can be slow-ish.

app/core/Authentication/Authentication.ts
```diff
@@ -803,18 +797,18 @@ class AuthenticationService {
...
-      );
+    if (options.shouldImportAccounts) {
+      for (const clientType of Object.values(WalletClientType)) {
+        const { discoveryScope } = WALLET_SNAP_MAP[clientType];
+        const multichainClient =
+          MultichainWalletSnapFactory.createClient(clientType);
+
+        discoveredAccountsCount += await multichainClient.addDiscoveredAccounts(
+          id,
+          discoveryScope,
```
orestiseth: we can optimise this too.

darioAnongba: Great point! Applied to all.

app/core/Authentication/Authentication.ts
darioAnongba: Great point! Applied to all.

app/core/Authentication/Authentication.ts
```diff
@@ -796,20 +813,28 @@ class AuthenticationService {
...
+          nAccountsPerSnap.push(count);
+        }),
+      );
     }
 
+    const discoveredAccountsCount = nAccountsPerSnap.reduce(
+      (acc, count) => acc + count || 0,
+      0,
+    );
+
```
orestiseth: nit: We can avoid the intermediate array variable and collapse that to something like:
```typescript
 const discoveredAccountsCount = options.shouldImportAccounts
      ? (
          await Promise.all(
            Object.values(WalletClientType).map(async (clientType) => {
              const { discoveryScope } = WALLET_SNAP_MAP[clientType];
              const multichainClient =
                MultichainWalletSnapFactory.createClient(clientType);

              return await multichainClient.addDiscoveredAccounts(
                id,
                discoveryScope,
              );
            }),
          )
        ).reduce((acc, count) => acc + (count || 0), 0)
      : 0;
```

but up to you, it does not make a big difference.

orestiseth (dismissed):

app/core/Authentication/Authentication.ts
darioAnongba: better indeed, applied.

orestiseth (approved):

app/core/Authentication/Authentication.ts
```diff
@@ -802,19 +813,22 @@ class AuthenticationService {
...
+                MultichainWalletSnapFactory.createClient(clientType);
+
+              return await multichainClient.addDiscoveredAccounts(
+                id,
+                discoveryScope,
+              );
+            }),
+          )
+        ).reduce((acc, count) => acc + count || 0, 0)
+      : 0;
```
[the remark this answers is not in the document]
darioAnongba: good catch but this seems to be the expected behaviour from the current code where we do not try catch the failure of `addDiscoveredAccounts`. Meaning that for multiple networks, the same should be expected, unless that was a bug before.

montelaidev (approved):

PR #17604, opened 2025-07-25
feat: perps close position

abretonc7s:
## **Description**

This PR implements the ability for users to close perpetual positions (TAT-1826) and introduces protocol-agnostic fee computation for the perps trading system.

### Key Changes:

1. **Close Position Feature**
   - Added new `PerpsClosePositionBottomSheet` component with market/limit order tabs
   - Integrated position size slider for partial/full position closing
   - Displays position P&L, effective margin, and closing percentage
   - Automatic TP/SL closure when position is closed
   - Success animations and feedback upon trade execution

2. **Protocol-Agnostic Fee Computation**
   - Created unified `usePerpsOrderFees` hook for centralized fee calculation
   - Added async `calculateFees` method to IPerpsProvider interface
   - Supports future volume-based discounts and user tier calculations
   - Separated protocol fees from MetaMask fees for transparency
   - Fixed incorrect HyperLiquid fee rates (0.045% taker, 0.015% maker)

The protocol-agnostic fee system allows each provider to implement its own fee structure while maintaining a consistent interface. This sets the foundation for future enhancements like user tier discounts and staking benefits.

## **Changelog**

CHANGELOG entry: Added the ability to close perpetual trading positions with market or limit orders, including partial position closing and automatic TP/SL closure

## **Related issues**

Fixes: TAT-1826

## **Manual testing steps**

1. Navigate to the Perps trading screen and open a position
2. Tap the "Close" button on an open position
3. Test market order closing:
   - Select "Market" tab
   - Use slider to select partial amount (e.g., 50%)
   - Verify position size, effective margin, and P&L display correctly
   - Confirm and verify position is partially closed
4. Test limit order closing:
   - Select "Limit" tab  
   - Set a limit price
   - Use slider to select full position (100%)
   - Verify order is placed correctly
5. Verify fees are calculated correctly:
   - Check that market orders show 0.045% fee
   - Check that limit orders show 0.015% fee (when acting as maker)
6. Verify TP/SL orders are automatically cancelled when position is closed
7. Verify success animation and Fox points display after closing

### **Before**

Users could only open positions but had no way to close them through the UI.

### **After**

https://github.com/user-attachments/assets/a3196795-b60f-4362-ad43-7bab85ef4500

- Close position bottom sheet with market/limit tabs
- Position size slider for partial closing
- P&L and effective margin display
- Success animations after closing


nickewansmith (commented):
Some general thoughts/questions from functional testing

- In some cases the position may not close i.e. a partial close when the position is a $10 position. There have been a few other situations I have yet to pinpoint where the position does not close as expected. In this case, do we want to notify the user about why the position is not closed or partially closed? I'm not sure if we want to handle that in this ticket or not.

- As mentioned I noticed there are some cases where the position does not close
  - If the position is at some low amount and we attempt to close it partially
  - In the case of a limit close, after clicking close position, we stay on the same screen and nothing seems to happen unless it is closed. I'm not completely sure what is expected here or this particular PR since I don't see orders made and when they might be triggered. If I time it just right, with the price I see in the screen underneath, the position does close
  
- I noticed that after the position is closed sometimes we redirect to the positions page and sometimes to the create an order page. I'm not sure if this is intended or not.

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.styles.ts
```diff
@@ -0,0 +1,118 @@
...
+      backgroundColor: theme.colors.background.alternative,
+      padding: 16,
+      borderRadius: 8,
+    },
+    statisticsLabel: {
+      marginBottom: 4,
+    },
+    statisticsValue: {
+      fontSize: 16,
+      fontWeight: '600',
```
Matt561: ⚠️ UI Maintainability Flag: We'll want to do a pass before GA and replace any bespoke styles (e.g. typography, buttons, etc) with DS components. This way the Perps UI is updated automatically when DS changes occur.

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
```diff
@@ -0,0 +1,274 @@
+import {
```
Matt561: Was this file simply moved? I think this was added in your previous market statistics PR.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -277,16 +288,17 @@ const PerpsOrderView: React.FC = () => {
...
-            • Limit orders: 0.02% of position size
+            • Limit orders: Fee varies by order type and liquidity
           </Text>
           <Text variant={TextVariant.BodyMD} style={styles.tooltipItem}>
-            • Fees are deducted from your margin balance
+            • Base rates shown (volume discounts may apply)
           </Text>
           <Text variant={TextVariant.BodyMD}>
-            • No funding fees for the first 8 hours
+            • Fees deducted from margin balance
```
Matt561, line 301: We'll want user-facing text to always be replaced with locale variables. It may be worth adding to the `CLAUDE.md` file.

app/components/UI/Perps/hooks/usePerpsOrderFees.ts
```diff
@@ -0,0 +1,149 @@
...
+
+        // Get protocol-specific fees from the active provider
+        const result = await calculateFees({ orderType, isMaker, amount });
+        setProtocolFeeRate(result.feeRate);
+      } catch (err) {
+        setError(
+          err instanceof Error ? err.message : 'Failed to fetch protocol fees',
+        );
+        // Fallback to a reasonable default
+        setProtocolFeeRate(0.00045); // 0.045% taker fee as fallback
```
Matt561: Where are we pulling these maker and taker fees from? Is this specified in the HL docs? I'm seeing contradicting numbers on different sites. 

It's not hugely important for now since we only care about the logic but curious nonetheless.

Matt561 (commented):
I haven't tested locally but quickly glanced through the changes and left some comments 👍

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
abretonc7s: Yes, I hardcoded this until we have ifnal copy

app/components/UI/Perps/hooks/usePerpsOrderFees.ts
abretonc7s: this is official hyperliquid doc, but the fallback shouldnt be hardcoded config.

abretonc7s:
> Some general thoughts/questions from functional testing
> 
> * In some cases the position may not close i.e. a partial close when the position is a $10 position. There have been a few other situations I have yet to pinpoint where the position does not close as expected. In this case, do we want to notify the user about why the position is not closed or partially closed? I'm not sure if we want to handle that in this ticket or not.
> * As mentioned I noticed there are some cases where the position does not close
>   
>   * If the position is at some low amount and we attempt to close it partially
>   * In the case of a limit close, after clicking close position, we stay on the same screen and nothing seems to happen unless it is closed. I'm not completely sure what is expected here or this particular PR since I don't see orders made and when they might be triggered. If I time it just right, with the price I see in the screen underneath, the position does close
> * I noticed that after the position is closed sometimes we redirect to the positions page and sometimes to the create an order page. I'm not sure if this is intended or not.

It does a navigation.goBack() currently so it depends where you come from.
For the amount this is probably a design issue, there is a minimum order amount that disable the field but it is very visible in UI.
We may have some edge case to handle but I think we could handle separately, like if we do partial close at from 15$ and min order in 10$, we woudlnt be able to close the remaining .

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.styles.ts
abretonc7s: yes we are still missing the final ds styles on figma, will adjust all this in a single pass.

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
abretonc7s: I am not sure what this refers to?

Matt561 (approved):

PR #17629, opened 2025-07-25
fix: remove overreliance on tokenSnapshot in withdraw and deposit confirmation

### LINKED ISSUE #17628: [Bug]: Earn token withdraw/deposit components should use earnTokenPair for token pair info
### Describe the bug

In the Earn withdrawal / deposit confirmation forms, we are using the earnToken and outputToken vars from useEarnToken, but we should instead use the earnTokenPair as this is the source of truth for the pair. 

earnToken and outputToken are never both true, really a better name would be isEarnToken and isOutputToken, but this is beyond the scope of this bug.

The logic in the two confirmation components is reacting to one side of the pair always being false, and is calling the tokenSnapshot logic for backup. The result is that everything is still successful, but the tokenSnapshot is requested each time instead of what we expect, that it is only called if one side of the pair is missing which will be rare.

### Expected behavior

We should not call tokenSnaphot logic as a backup if we do not need to, it is meant to be used as a fallback and right now we are too reliant on it with this bug

### Screenshots/Recordings

n/a

### Steps to reproduce

1. Log output for tokenSnapshot var returned from useEarnToken in the two confirmation views for deposit and withdraw: `app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/index.tsx`, `app/components/UI/Earn/Views/EarnLendingWithdrawalConfirmationView/index.tsx`
2. Ensure that you have a token pair already. USDC/aUSDC for example. Then Go to the deposit flow and click through until deposit is confirmed
3. you will see tokenSnapshot data
4. Go to the withdraw flow and click through until withdraw is confirmed
5. you will see tokenSnapshot data

```shell

```

### Detection stage

During release testing

### Version

7.52.0

### Build type

Beta

### Device

Iphone 14 

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

nickewansmith:
## **Description**
- There is an over-reliance on tokenSnapshot logic due to a bug where we are not using the correct variable from useEarnTokens for earn token pairs. The flow will still work but we are reliant on tokenSnapshot logic for it to work. The tokenSnapshot logic is not meant to be used to this degree.
- Ensure that we are using the correct variable for earn token pairs from the useEarnToken hook, this eliminates the issue

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/17628

## **Manual testing steps**

Log output for tokenSnapshot var returned from useEarnToken in the two confirmation views for deposit and withdraw: app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/index.tsx, app/components/UI/Earn/Views/EarnLendingWithdrawalConfirmationView/index.tsx
Ensure that you have a token pair already. USDC/aUSDC for example. Then Go to the deposit flow and click through until deposit is confirmed
you will not see tokenSnapshot data
Go to the withdraw flow and click through until withdraw is confirmed
you will not see tokenSnapshot data


app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/index.tsx
```diff
@@ -99,8 +99,12 @@ const EarnLendingDepositConfirmationView = () => {
...
   );
 
-  const { earnToken, outputToken, getTokenSnapshot, tokenSnapshot } =
-    useEarnToken(token as TokenI);
+  const { earnTokenPair, getTokenSnapshot, tokenSnapshot } = useEarnToken(
+    token as TokenI,
+  );
+
+  const earnToken = earnTokenPair?.earnToken;
+  const outputToken = earnTokenPair?.outputToken;
```
Matt561, line 107: I think it's fine for now given the scope of this PR is purely to fix the bug but it may be confusing in the future to have `earnTokenPair`, `earnToken`, and `outputToken` returned from the `useEarnToken` hook (assuming this hasn't changed while I was away of course).

I'm not sure when we _wouldn't_ want to use `earnTokenPair` currently.

Matt561 (approved):

app/components/UI/Earn/Views/EarnLendingDepositConfirmationView/index.tsx
nickewansmith: There are a few cases where we want to know if a token is a lending or outputToken.

nickewansmith: https://consensyssoftware.atlassian.net/browse/TAT-1312

nickewansmith:
Smoke Tests have been hard to get through for a while now, I have checked and the current failures are unrelated.

PR #17642, opened 2025-07-28
fix: refactor recreatevault -> change password cp-7.53.0

### LINKED ISSUE #17618: [Bug]: Creating new Solana account when changing password, it will be disappeared when re-login
### Describe the bug

When user have 2 mobiles device, android and ios
on ios device, user see 3 account
on android device, user change the password
on ios device, lock the account, unlock, check and see 4 accounts
if reset the wallet on android and ios, the new solana wallet will be disappear

Note: if login same social account on extension, will not show any new solana account

### Expected behavior

Should not have this behavior

### Screenshots/Recordings

on ios device
Image

on extension

Image

### Steps to reproduce

1. When user have 2 mobiles device, android and ios
2. on ios device, user see 3 account
3. on android device, user change the password
4. on ios device, lock the account, unlock, check and see 4 accounts
5. reset the wallet on android and ios, the new solana wallet will be disappear

6. if login same social account on extension, will not show any new solana account

```shell

```

### Detection stage

On main branch

### Version

7.52.99

### Build type

Beta

### Device

Honor Play, iPhone14promax

### Operating system

iOS, Android

### Additional context

_No response_

### Severity

_No response_

ieow:
## **Description**
Refactor recreateNewVault with changePassword
changePassword is recently added in KeyringController, refactoring mobile's recreateNewVault to reduce the complexity of change password in mobile

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:
#17593
 #17618
## **Manual testing steps**

1. Login into Wallet
2. Go to Setting > Security Settings > Change Password
3. Submit old password 
4. Submit new password

### **After**

https://github.com/user-attachments/assets/724a7f84-f88c-4139-8d66-c43232ccfaff


app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
 
 import { selectSeedlessOnboardingLoginFlow } from '../selectors/seedlessOnboardingController';
 import { Authentication } from './Authentication/Authentication';
 import { endTrace, trace, TraceName, TraceOperation } from '../util/trace';
 
 /**
- * Restore the given serialized QR keyring.
- *
- * @param {unknown} serializedQrKeyring - A serialized QR keyring.
+ * Returns current vault seed phrase
```
himanshuchawla009:
```suggestion
 * Returns the seed phrase from vault for the provided `keyringId`.
```

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
+ * It does it using an empty password or a password set by the user
```
himanshuchawla009:
```suggestion
 * If the wallet is unlocked, the password can be an empty string. Otherwise, a password is required to retrieve the seed phrase from the vault.
```

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
-  } catch (e) {
     Logger.error(
-      e,
-      'error while trying to restore imported srp accounts on recreate vault',
+      error,
+      '[recreateVaultWithNewPassword] seedless onboarding pw change error',
     );
-  }
-};
+    // restore keyring with old password if seedless onboarding pw change fails
```
himanshuchawla009: this comment is not valid anymore.

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
-      );
-    }
 
-    return keyringId;
-  } catch (e) {
     Logger.error(
-      e,
-      'error while trying to restore imported srp accounts on recreate vault',
+      error,
+      '[recreateVaultWithNewPassword] seedless onboarding pw change error',
```
himanshuchawla009:
```suggestion
      '[seedlessChangePassword] seedless onboarding pw change error',
```

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
-  return await KeyringController.exportSeedPhrase(password, keyringId);
-};
-
 /**
  * Recreates a vault with the new password
  *
  * @param password - current password
  * @param newPassword - new password
- * @param selectedAddress
+ * @param selectedAddress - selected address
```
himanshuchawla009:
```suggestion
 * @param selectedAddress - current selected address in wallet.
```

app/core/Vault.js
```diff
@@ -24,326 +23,90 @@ import { Authentication } from './Authentication/Authentication';
...
-  )
-    ? await getSerializedKeyring(KeyringTypes.ledger)
-    : undefined;
-  const serializedQrKeyring = hasKeyringType(
-    KeyringController.state,
-    KeyringTypes.qr,
-  )
-    ? await getSerializedKeyring(KeyringTypes.qr)
-    : undefined;
+  await KeyringController.changePassword(newPassword);
```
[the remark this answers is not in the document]
himanshuchawla009: is this new `changePassword` function backward compatible for existing mobile users? @mikesposito

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
 /**
- * Restore the given serialized Ledger keyring.
+ * Changes the password of the seedless onboarding controller
  *
- * @param {unknown} serializedLedgerKeyring - A serialized Ledger keyring.
+ * @param {string} newPassword - new password
+ * @param {string} password - current password
  */
-export const restoreLedgerKeyring = async (serializedLedgerKeyring) => {
+export const seedlessChangePassword = async (newPassword, password) => {
```
himanshuchawla009:
```suggestion
export const recreateSeedlessVaultWithNewPassword = async (newPassword, password) => {
```

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
-      await KeyringController.createNewVaultAndRestore(
-        password,
-        primaryKeyringSeedPhrase,
-      );
-      seedlessChangePasswordError = new SeedlessOnboardingControllerError(
-        SeedlessOnboardingControllerErrorType.ChangePasswordError,
-        error || 'Password change failed',
-      );
-      await Authentication.syncKeyringEncryptionKey();
+      await KeyringController.changePassword(password);
```
mikesposito: If this step fails for some reason we'll be in a tough spot, where the `seedlessChangePassword` failed while the first `KeyringController.changePassword` was successful - or in other words, KeyringController will use the new password while SeedlessOnboardingController will use the old password.

At the same time, I can't think of a reason the first call would succeedv while the second fails.

app/core/Vault.js
```diff
@@ -1,349 +1,96 @@
...
+  if (isSeedlessFlow) {
     try {
-      trace({
-        name: TraceName.OnboardingResetPassword,
-        op: TraceOperation.OnboardingSecurityOp,
-      });
-      await SeedlessOnboardingController.changePassword(newPassword, password);
-      await Authentication.syncKeyringEncryptionKey();
-      specificTraceSucceeded = true;
+      await seedlessChangePassword(newPassword, password);
```
chaitanyapotti: we should keep the sentry traces still

app/core/Vault.js
mikesposito: nit: is this comment still relevant?

mikesposito (commented):
Thanks for taking on this work! Changes look overall good, though I left a comment regarding the only thing that concerns me with this refactor - even though, regardless, this new version seems infinitely safer than the original code we had.

app/core/Vault.js
himanshuchawla009: we also need to revert seedless controller encryption key as well are doing in extension.

```
    const revertedKeyringEncryptionKey = await exportEncryptionKey();
     await storeKeyringEncryptionKey(revertedKeyringEncryptionKey);
 
 ```

**Review by himanshuchawla009 — CHANGES_REQUESTED** [review]

File: app/core/Vault.js (hunk 8 above)
Comment (himanshuchawla009) [reply app/core/Vault.js]: in case of error , its reverting keyring controller back to old pwd but nevertheless there can be rare cases like user closes the window when seedless controller call in progress.

Comment (himanshuchawla009) [reply app/core/Vault.js]: Also we are already handling the case when keyring state is latest but seedless controller local state is outdated, in the change pwd flow let say seedless backend calls succeeds but local state is still old because user closes the app, then  when user will open the app again they will see the popup "Your password has been changed" and once they enter the new pwd keyring state will sync with seedless.

Comment (mikesposito) [reply app/core/Vault.js]: > we are already handling the case when keyring state is latest but seedless controller local state is outdated 

That's good to know. Given that, would it make sense to just throw an error in case `seedlessChangePassword` fails, instead of trying to revert the KeyringController password? This way, we would rely on the existing flow

File: app/core/Vault.js (hunk 6 above)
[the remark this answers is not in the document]
Comment (mikesposito) [reply app/core/Vault.js]: yes, the `changePassword` function works for both existing and new users.

**Comment by mikesposito** [comment]
> is this new changePassword function backward compatible for existing mobile users? @mikesposito

@himanshuchawla009 yes, the `changePassword` function works for both existing and new users

File: app/core/Vault.js (hunk 8 above)
Comment (ieow) [reply app/core/Vault.js]: [at the finally ](https://github.com/MetaMask/metamask-mobile/pull/17642/files#diff-9cf2bc6561abadc82def41b57791dc765fd05bf93563c4b198dd09d07f60ad1fR97)
we use the Authentication.syncKeyringEncryptionKey

File: app/core/Vault.ts [hunk 10]
```diff
@@ -0,0 +1,111 @@
...
+    // if it succed seedless change password but fail on the change password on local, we will prompt user password out of date
+    // and ask user to login with new password
+    if (isSeedlessFlow) {
+      await recreateSeedlessVaultWithNewPassword(newPassword, password);
+    }
+
+    await KeyringController.changePassword(newPassword);
+  } finally {
+    if (isSeedlessFlow) {
+      await Authentication.syncKeyringEncryptionKey();
```
himanshuchawla009: we shouldn't do this in case of error right? Finally block will be executed in all cases.

app/core/Vault.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+  Engine.setSelectedAddress(selectedAddress);
+};
+
+/**
+ * Recreates a vault with the same password for the purpose of using the newest encryption methods
+ *
+ * @param password - Password to recreate and set the vault with
+ */
+export const recreateVaultWithSamePassword = async (
```
himanshuchawla009: why diff function is required, is it being used anywhere?

app/core/Vault.ts
```diff
@@ -0,0 +1,111 @@
...
+ * @param {string} keyringId - Keyring ID to retrieve the seed phrase from the vault
+ * @returns {Promise<string>} - Seed phrase from the vault
+ */
+export const getSeedPhrase = async (password: string, keyringId: string) => {
+  const { KeyringController } = Engine.context;
+  return await KeyringController.exportSeedPhrase(password, keyringId);
+};
+
+/**
+ * Changes the password of the seedless onboarding controller
```
himanshuchawla009:
```suggestion
 * Changes the password of the seedless onboarding controller and re-creates seedless vault using new password.
```

app/core/Vault.ts
```diff
@@ -0,0 +1,111 @@
...
+  } finally {
+    endTrace({
+      name: TraceName.OnboardingResetPassword,
+      data: { success: specificTraceSucceeded },
+    });
+  }
+};
+
+/**
+ * Recreates a vault with the new password
```
himanshuchawla009:
```suggestion
 * Recreates keyring and seedless storage vault with the new password.
```

app/core/Vault.ts
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @param password - current password
+ * @param newPassword - new password
+ * @param selectedAddress - current selected address in wallet.
+ */
+export const recreateVaultWithNewPassword = async (
```
himanshuchawla009:
```suggestion
export const recreateVaultsWithNewPassword = async (
```

himanshuchawla009 (changes requested):

app/core/Vault.ts
ieow: removed

himanshuchawla009 (approved):
lgtm

PR #17643, opened 2025-07-28
test: Add preinstalled example Snap

FrederikBolding:
## **Description**

Add preinstalled example Snap for ease of testing with Flask.


GuillaumeRx:
We need to disable/hide it on production builds right ?

FrederikBolding:
> We need to disable it on production builds right ?

We don't currently publish Flask to production on mobile, so doesn't seem that important. I guess technically we are also including it in production on extension, just hidden in the UI 🤔

GuillaumeRx:
> > We need to disable it on production builds right ?
> 
> We don't currently publish Flask to production on mobile, so doesn't seem that important. I guess technically we are also including it in production on extension, just hidden in the UI 🤔

Yeah we just hide it, I guess if we're not publishing anything it's okay for now to let it as it is on Flask

GuillaumeRx (approved):

PR #17679, opened 2025-07-28
fix: cp-7.52.0 dapp permissions not properly updating after hardware wallet is forgotten 

### LINKED ISSUE #17611: [Bug] App crash and error when navigating into in-app browser
Build: 7.52 (2196)

Not sure if these steps are the cause of this error but have attached the state logs to investigate.

1. Add hardware wallet (QR device)
2. Import 2x QR accounts
3. Forget QR device
4. Click on in-app browser
5. Error displayed

Actual: 

State logs:
[state-logs-v7.52.0-(2197).json](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMmpZQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--8e2454e5378edb452e5812b29a165d590976ee8a/state-logs-v7.52.0-%282197%29.json)

Expected: User able to launch in-app browser

jiexi:
## **Description**

Mobile was failing to update the existing dapp permissions when a hardware wallet was forgotten. The updating of dapp permissions was occurring after the hardware wallet was forgotten. 

The helper used to update the dapp permissions relies on the account being targeted for removal to still exist in the AccountsController so that it can resolve a hex address to a CAIP Account Id. (This helper still operates on hex address because of prolific usage of hex address in the UI still). The solution here was to ensure the permissions are updated before the hardware wallet is forgotten.

Separately, the helper used to update the dapp permissions incorrectly logs an error when it encounters a subject that does not have a CAIP-25 endowment (likely a snap). This was fixed by making the CAIP-25 endowment not found a distinct error type. This is not crucial to resolving the bug.

## **Changelog**

CHANGELOG entry: Fixed a bug that was causing the in-app browser to crash when loading a dapp that was previously permitted with an account from a hardware wallet that was forgotten

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/17611

## **Manual testing steps**

1. Add a hardware wallet by opening the account selector on the main wallet page -> Add account or hardware wallet -> Hardware Wallet -> QR -> Scan QR Code (You can use Airgap Vault on iOS to make a QR wallet) -> Select some accounts to import
2. Go to the in-app browser
3. Connect a dapp with one of the newly added QR wallet accounts
4. Forget the hardware wallet by opening the account selector on the main wallet page -> Add account or hardware wallet -> Hardware Wallet -> QR -> Forget wallet
5. Go back to the in-app browser
6. The page should load fine, without an uncaught exception being presented to the user
7. Tap the top right permission, see that the previously permitted QR wallet accounts are no longer permitted

### **Before**

https://github.com/user-attachments/assets/32e22637-3b4e-4449-9c4d-a42daf4ee2e7

### **After**

https://github.com/user-attachments/assets/84d2ad5c-16ac-4fa5-933e-13d6446a7968


app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -323,14 +324,21 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
+    // This is because `removeAccountsFromPermissions` relies on the account
+    // existing in AccountsController in order to resolve a hex address
+    // back into CAIP Account Id. Hex addresses are used in
+    // `removeAccountsFromPermissions` because too many places in the UI still
+    // operate on hex addresses rather than CAIP Account Id.
+    await Engine.context.KeyringController.withKeyring(
+      { type: ExtendedKeyringTypes.qr },
+      async ({ keyring }) => {
+        const keyringAccounts = await keyring.getAccounts();
+        removeAccountsFromPermissions(keyringAccounts);
```
adonesky1, line 337: are these already checksummed. when returned directly from the keyring? Or is there a reason we don't need to checksum them anymore?

app/core/Ledger/Ledger.test.ts
```diff
@@ -179,6 +188,14 @@ describe('Ledger core', () => {
   });
 
   describe('forgetLedger', () => {
+    it('updates removes the accounts from existing permissions', async () => {
```
adonesky1:
```suggestion
    it('removes the accounts from existing permissions', async () => {
```

app/core/Ledger/Ledger.ts
```diff
@@ -94,6 +95,14 @@ export const closeRunningAppOnLedger = async (): Promise<void> => {
...
 export const forgetLedger = async (): Promise<void> => {
   await withLedgerKeyring(async ({ keyring }) => {
+    // Permissions need to be updated before the hardware wallet is forgotten.
+    // This is because `removeAccountsFromPermissions` relies on the account
+    // existing in AccountsController in order to resolve a hex address
+    // back into CAIP Account Id. Hex addresses are used in
+    // `removeAccountsFromPermissions` because too many places in the UI still
+    // operate on hex addresses rather than CAIP Account Id.
+    const accounts = await keyring.getAccounts();
+    removeAccountsFromPermissions(accounts);
```
adonesky1, line 105: This seems like a separate gap from what's mentioned in the PR description which seems to point more to the correction of the `onForget` handler? This is just another gap in our implementation where we made no previous attempt to update permissions at all?

app/core/Permissions/index.ts
```diff
@@ -465,12 +468,15 @@ export const removePermittedAccounts = (
...
 
 // The codebase needs to be refactored to use caipAccountIds so that this is easier to change
-export const removeAccountsFromPermissions = async (addresses: Hex[]) => {
+export const removeAccountsFromPermissions = (addresses: Hex[]) => {
   const { PermissionController } = Engine.context;
   for (const subject in PermissionController.state.subjects) {
     try {
       removePermittedAccounts(subject, addresses);
     } catch (e) {
+      if (e instanceof Caip25EndowmentMissingError) {
```
adonesky1, line 477: Nice catch here. Wondering, are there any types of errors **we would** actually want to fully stop this loop for?

jiexi: i don't think we'd ever want to just fully exit without trying to remove the accounts from all subjects

app/core/Ledger/Ledger.ts
jiexi: > This is just another gap in our implementation where we made no previous attempt to update permissions at all?

Sorry, I'm wrong about the previous implementation exiting early. It was just incorrectly logging an error when a caip-25 permission did not exist. The need to remove accounts before forgetting the keyring is still due to our implementation in a way. `removePermittedAccounts` will fail because it's unable to resolve the hex addresses via `AccountsController.getAccountByAddress` since those accounts no longer exist. Because of that, the ordering matters

app/components/Views/ConnectQRHardware/index.tsx
jiexi: doesn't matter if they are checksummed or not because we do a lowercase comparison later in the chain-agnostic-permission package

adonesky1: 🤔 should we doing a lowercase comparison since not all chains are address case-insensitive? I mean probably not really an issue in this particular context but

app/core/Permissions/index.ts
adonesky1: right so then should we catch any errors and continue on, not just this error type?

app/components/Views/ConnectQRHardware/index.tsx
```diff
@@ -323,14 +324,21 @@ const ConnectQRHardware = ({ navigation }: IConnectQRHardwareProps) => {
...
+      },
     );
-    removeAccountsFromPermissions(checksummedRemovedAccounts);
+    const { remainingAccounts } = await KeyringController.forgetQRDevice();
+    Engine.setSelectedAddress(remainingAccounts[remainingAccounts.length - 1]);
```
ffmcgee725: [nit]
```suggestion
Engine.setSelectedAddress(remainingAccounts.at(-1));
```
we can take advantage of `.at()` [javascript array method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) for simpler syntax here.

"The usual practice is to access [length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length) and calculate the index from that — for example, array[array.length - 1]. The at() method allows relative indexing, so this can be shortened to array.at(-1)." 

source : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

app/core/Permissions/index.ts
jiexi: right, this code does that already. The only difference is that `Failed to remove account from permissions after deleting account from wallet` will not get logged if the failure is due to the caip-25 permission not existing for the subject.

app/components/Views/ConnectQRHardware/index.tsx
jiexi: to clarify, we do a lowercase comparison for EVM. This eventually hits our account comparison helper that handles the EVM and non-EVM cases for address comparison

app/core/Permissions/index.ts
adonesky1: oh lol whoops I misread 👍

adonesky1 (dismissed):
LGTM

ffmcgee725 (dismissed):

adonesky1 (approved):
LGTM!

ffmcgee725 (approved):

Akaryatrh:
I can't reproduce anymore the issue described [here](https://github.com/MetaMask/metamask-mobile/issues/17611). Approving for QA team.

PR #17685, opened 2025-07-28
feat: 14414 Replace redux-persist with barebones JS persist system using FileSystemStorage

MarioAslau:
## **Description**

This PR aims at replacing our redux-persist with barebones JS Persistence using FileSystemStorage.

As a first step, all controller state is being removed from the regular redux-persist into FileSystemStorage, with each controller being written to a separate file vs storing the whole Engine object into one file.

**EngineService**:
* setupEnginePersistence:
  * Acts as the bridge between the Engine's controller system and the storage layer, setting up real-time persistence for individual controllers
  * Contains Redux synchronization by calling this.updateBatcher.add(controllerName) to efficiently batch Redux updates after successful persistence

**persistConfig** now has:
* ControllerStorage methods (getter and setter) for persisted controllers in FilesystemStorage
* Note: The existing persistUserTransform and persistOnboardingTransform remain for non-engine state; engine state is now blacklisted from redux-persist

**Migrations** now have:
* To keep developer experience the same when creating migrations, the migration system now includes:
  * deflateToControllersAndStrip → "repacking" the single object back into distributed files and safely preserving failed controllers to prevent data loss
  * inflateFromControllers → "unpacking" distributed files back into a single object for migration processing
* Migration 104: Checks if old persisted engine state is present, migrates each controller to individual storage files, then removes successfully migrated controllers from the old redux-persist state (preserving any that failed to migrate)

**Key Benefits:**
* ndividual controller failures don't affect other controllers
* Performance: Smaller, focused file operations instead of large monolithic writes
* Data Safety: Critical bug fix prevents silent data loss during migration failures

Note: if .env file value is not set right, this will trigger a Vault Recovery on dev environments after locking and reopening of the application. Due to this conditions we have for E2E: const state = isE2E ? reduxState?.engine?.backgroundState  : persistedState?.backgroundState ?? {};

## **Changelog**

CHANGELOG entry: 

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

---

> [!NOTE]
> Replaces engine redux-persist with per-controller filesystem storage, wires EngineService to persist and batch controller updates, and adds migration 104 with inflate/deflate helpers plus tests and docs.
> 
> - **State Persistence (core change)**
>   - Introduces `ControllerStorage` and `createPersistController` in `store/persistConfig/index.ts` for per-controller FilesystemStorage persistence (debounced writes).
>   - Blacklists `engine` from redux-persist; keeps `user` and `onboarding` transforms only.
>   - Adds `ControllerStorage.getAllPersistedState()` to load controller states at startup.
> - **EngineService**
>   - `start()` now prefers persisted controller state (non-E2E) and initializes controllers via `initializeControllers`.
>   - `setupEnginePersistence()` subscribes to `BACKGROUND_STATE_CHANGE_EVENT_NAMES`, batches Redux updates (`UPDATE_BG_STATE_KEY`), and persists filtered controller state.
>   - Renames `updateControllers` → `initializeControllers`; enhances batching logic.
> - **Migrations**
>   - Adds `104` to split legacy `engine.backgroundState` into per-controller files and preserve failed controllers.
>   - Updates `migrations/index.ts` to inflate controller files before 105+ and deflate back after last migration; strips `engine` only if all saves succeed.
> - **Tests & Tooling**
>   - Replaces old `persistConfig` tests; adds comprehensive tests for new persistence, migration 104, and EngineService (timers, batching, persistence, error handling).
>   - Updates test setup to mock `unstable_batchedUpdates` and related dependencies.
> - **Docs**
>   - Adds `STATE_FLOW_DOCUMENTATION.md` describing end-to-end state flow and new persistence architecture.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 60d270a32a6dee0ded1ac40069e2c0818eaa2540. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/store/migrations/091.ts
```diff
@@ -0,0 +1,129 @@
...
+            hasProperty(controllers, controllerName) &&
+            isObject(controllers[controllerName])
+          ) {
+            const controllerState = controllers[controllerName];
+
+            // Save raw controller data to individual file (filtering will be applied later)
+            const key = `persist:${controllerName}`;
+            const value = JSON.stringify(controllerState);
+
+            await FilesystemStorage.setItem(key, value, Device.isIos());
```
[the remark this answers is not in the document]
MarioAslau: This is wrong. The pre-existing data in the old storage is already filtered when it was stored there. Adding an additional filter inside the migration would be re-filtering filtered data. It is redundant.

app/store/migrations/index.ts
Cal-L, line 1: @MarioAslau mentioned that a new migration mechanism for the controller data will be introduced as a follow up. Let's keep an eye out for any teams introducing migrations for controller data after this PR goes in.

app/store/migrations/092.ts
```diff
@@ -0,0 +1,151 @@
...
+      return newState;
+    }
+
+    Logger.log(
+      'Migration 92: No existing engine data found, skipping migration',
+    );
+
+    return state;
+  } catch (error) {
+    Logger.error(
```
Cal-L: Logger.error already captures exceptions. Let's change all Logger.error to captureException so that errors are already reported

app/store/persistConfig.ts
```diff
@@ -13,6 +13,122 @@ import { getPersistentState } from './getPersistentState/getPersistentState';
 const TIMEOUT = 40000;
 const STORAGE_THROTTLE_DELAY = 200;
 
+const CONTROLLER_LIST = [
```
Cal-L: Export into constants file

app/store/persistConfig.ts
```diff
@@ -13,6 +13,122 @@ import { getPersistentState } from './getPersistentState/getPersistentState';
...
+                    `Invalid persisted data for ${controllerName}: not an object`,
+                  ),
+                );
+                return null;
+              }
+
+              const { _persist, ...controllerState } = parsedData;
+
+              // Only include controllers that have meaningful state (not empty objects)
+              if (Object.keys(controllerState).length > 0) {
```
Cal-L: Let's verify if it's possible to have a controller that exists in persistence without state

Cal-L: We can probably remove this condition to save on performance since this isn't likely to break things

app/store/persistConfig.ts
```diff
@@ -13,6 +13,122 @@ import { getPersistentState } from './getPersistentState/getPersistentState';
...
+            Logger.error(error as Error, {
+              message: `Failed to get controller state for ${controllerName}`,
+            });
+            return null; // Don't include controllers with errors
+          }
+        }),
+      );
+
+      // Combine all controller states into a single object, filtering out null values
+      const backgroundState = controllerStates
```
Cal-L: Instead of post processing controllerStates, you can just construct the object as it goes through the mapping. Then you can just return controllerStates here instead

app/store/persistConfig.ts
```diff
@@ -59,57 +175,56 @@ const MigratedStorage = {
...
-      Record<string, unknown>
-    > = {};
-    for (const [key, value] of Object.entries(controllers)) {
-      if (!value || typeof value !== 'object') continue;
-
-      const persistedState = getPersistentState(
-        value,
-        // @ts-expect-error - EngineContext have stateless controllers, so metadata is not available
-        Engine.context[key as keyof EngineContext]?.metadata,
+              ReduxService.store.dispatch({
```
Cal-L: Let's add some comments to indicate what this is doing

app/store/persistConfig.ts
```diff
@@ -59,57 +175,56 @@ const MigratedStorage = {
...
-    try {
-      // Check if Engine is initialized by trying to access context
-      if (Engine.context) {
-        // This is just to trigger the error if engine does not exist
-      }
-    } catch (error) {
-      // Engine not initialized, skipping transform
-      return inboundState;
-    }
+              await ControllerStorage.setItem(
```
Cal-L: Let's add some comments to indicate what this is doing

Cal-L (commented):
Left some comments

app/store/persistConfig.ts
```diff
@@ -13,6 +13,107 @@ import { getPersistentState } from './getPersistentState/getPersistentState';
...
+      );
+
+      return { backgroundState };
+    } catch (error) {
+      Logger.error(error as Error, {
+        message: 'Failed to gather controller states',
+      });
+      return { backgroundState: {} };
+    }
+  },
```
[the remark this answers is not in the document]
MarioAslau: This is very unlikley, the error case is low probability. We usually can have one controller missing, if more are missing, the app would break. Adding error propagation + handling adds overhead to every state persistence

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,245 @@
...
+                type: UPDATE_BG_STATE_KEY,
+                payload: { key: controllerName },
+              });
+            } catch (error) {
+              Logger.error(
+                error as Error,
+                `Failed to persist ${controllerName} state`,
+              );
+            }
+          }, 200),
```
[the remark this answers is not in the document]
MarioAslau: The race condition is less critical than underlined above:
- Individual controller scope: Each controller has its own debounced function
- Sequential execution within controller: Each execution waits for completion
- File overwrites are atomic: Filesystem writes are typically atomic operations
- Dispatch happens after persistence
- Only issue: rapid state changes might lose intermediate updates (but that's expected with debouncing)

The race condition exists but might be more of a theoretical issue than a practical problem in this specific case.

MarioAslau:
Bitrise tests before merging Main: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/7a46bb54-209b-4d15-9151-8983f2d66b07

MarioAslau:
This ticket was solved. We are waiting for the right time to merge it into Main

app/core/EngineService/EngineService.ts
```diff
@@ -72,7 +72,11 @@ export class EngineService {
       parentContext: getUIStartupSpan(),
       tags: getTraceTags(reduxState),
     });
-    const state = reduxState?.engine?.backgroundState ?? {};
+
+    const state = isE2E
```
MarioAslau: this isE2E can cause problems when developing locally if the env variable is not set/read well, which will result in vault recovery after re-opening the app.


Comment it out if causing issues and use  'const state = persistedState?.backgroundState ?? {};'

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,268 @@
...
+                // @ts-expect-error - EngineContext have stateless controllers, so metadata is not available
+                Engine.context[controllerName as keyof EngineContext]?.metadata,
+              );
+
+              ReduxService.store.dispatch({
+                type: UPDATE_BG_STATE_KEY,
+                payload: { key: controllerName },
+              });
+
+              await persistController(filteredState);
```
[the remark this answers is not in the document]
MarioAslau: This is false, redux is primary and should be executed immediately, persisting comes secondary and can be done after.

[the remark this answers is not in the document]
MarioAslau: Same as above, redux should happen first, persistence after

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,268 @@
...
+            } catch (error) {
+              Logger.error(
+                error as Error,
+                `Failed to process ${controllerName} state change`,
+              );
+            }
+          },
```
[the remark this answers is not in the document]
MarioAslau: Same reasoning as the others. Redux State is primary and the persist is secondary, coming after.

app/store/persistConfig/index.ts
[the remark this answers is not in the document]
MarioAslau: Redux should have priority, persitence coming secondary after

app/store/migrations/104.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+jest.mock('redux-persist-filesystem-storage');
+const mockFilesystemStorage = FilesystemStorage as jest.Mocked<
+  typeof FilesystemStorage
+>;
+
+jest.mock('../../util/device');
+const mockDevice = Device as jest.Mocked<typeof Device>;
+
+describe('Migration 103', () => {
```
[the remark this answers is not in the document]
MarioAslau: stale, file has been updated

app/store/persistConfig/index.ts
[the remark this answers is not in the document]
MarioAslau: Redux comes first, persistence second

app/store/migrations/index.ts
```diff
@@ -224,37 +226,124 @@ export const migrationList: MigrationsList = {
...
 
-        // If this is the last migration and we have a callback, run it
         if (
           onMigrationsComplete &&
           Number(migrationNumber) === Object.keys(inputMigrations).length - 1
         ) {
           onMigrationsComplete(migratedState);
         }
 
+        if (Number(migrationNumber) === lastVersion && lastVersion >= 102) {
```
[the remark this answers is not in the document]
MarioAslau: outdated

app/core/EngineService/EngineService.ts
```diff
@@ -73,7 +73,11 @@ export class EngineService {
...
+      ? reduxState?.engine?.backgroundState
+      : persistedState?.backgroundState ?? {};
+
```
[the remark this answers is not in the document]
MarioAslau: this isE2E can cause problems when developing locally if the env variable is not set/read well, which will result in vault recovery after re-opening the app.

Comment it out if causing issues and use 'const state = persistedState?.backgroundState ?? {};'

MarioAslau:
Bitrise PR E2E: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/a6377d95-9726-48d9-a648-b2cd1c312977

asalsys (dismissed):

app/core/EngineService/EngineService.ts
```diff
@@ -30,12 +35,6 @@ export class EngineService {
         if (key === INIT_BG_STATE_KEY) {
           // first-time init action
           ReduxService.store.dispatch({ type: INIT_BG_STATE_KEY });
-        } else {
```
Gudahtt, line 33: What was the reason for removing incremental controller state updates? I'd expect this to result in the Redux backgroundState slice getting stale after initialization 🤔

MarioAslau: @Gudahtt the have been moved to  app/store/persistConfig/index.ts in setupEnginePersistence()

Gudahtt: Ah I see, thanks. I knew there must be some equivalent somewhere, or all our tests would be failing.

MarioAslau: EngineService calls setupEnginePersistence() during startup, setupEnginePersistence() subscribes to all controller state changes. Every controller state change triggers UPDATE_BG_STATE_KEY dispatch.

Gudahtt: What was the reason for moving it there? It seems confusing to me that the UI background state updates would be coming from `persistConfig`, given that it has nothing to do with persistence. Especially given that the initial update is still here.

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,268 @@
...
+          } catch (error) {
+            Logger.error(error as Error, {
+              message: `Failed to persist ${controllerName} state`,
+            });
+          }
+        }, 200);
+
+        Engine.controllerMessenger.subscribe(
+          eventName,
+          // Debounce to prevent excessive filesystem writes during rapid state changes
```
Gudahtt: Nit: This comment seems to be on the wrong line. Shouldn't it be next to where debounce is used? Either where the debounced function is created, or where it's called.

Also the warning is a little unclear to me - what's the intended takeaway there? The manner we're using it in here has a degree of unpredictability, but it shouldn't cause any problematic race conditions.

It might exacerbate the "state loss due to app closing before state is persisted" issue, but that possibility will exist no matter what, and was a calculated risk with this decision I think.

app/store/migrations/104.ts
```diff
@@ -0,0 +1,101 @@
...
+          if (
+            hasProperty(controllers, controllerName) &&
+            isObject(controllers[controllerName])
+          ) {
+            const controllerState = controllers[controllerName];
+
+            const key = `persist:${controllerName}`;
+            const value = JSON.stringify(controllerState);
+
+            await FilesystemStorage.setItem(key, value, Device.isIos());
```
Cal-L, line 50: We recently encountered a race condition with redux persist where the same migrations can run again if the app is closed before re-persistence can happen. This makes it dangerous to update APIs outside of the redux persist context since those values could change the next time around. Is there a way for us to update file system storage in a different area?

Cal-L (commented):
Let's resolve the cursor comments and add JS docs in the changes for the migrations/index file

app/store/migrations/index.ts
```diff
@@ -224,37 +226,122 @@ export const migrationList: MigrationsList = {
...
       const asyncMigration = async (
         incomingState: Promise<unknown> | unknown,
       ) => {
-        const state = await incomingState;
+        let state = await incomingState;
+
+        if (!didInflate && Number(migrationNumber) > 104) {
+          state = await inflateFromControllers(state);
+          didInflate = true;
+        }
```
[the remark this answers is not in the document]
MarioAslau: We do not want the inflation/deflation to run on 104, just afterwards.

app/store/migrations/104.ts
MarioAslau: Cursor comments regarding the same issue keep popping up. I double checked with Andre, Redux is the primary one and should be executed immediately, while persistence is secondary and is done after.

Regarding the race condition. After reviewing the migration logic, I believe the current implementation is actually safe from data loss:

1. If Migration 104 completes successfully ->  engine.backgroundState gets cleared -> migration won't re-run
2. If Migration 104 doesn't complete (app crash) -> engine.backgroundState remains unchanged -> re-running copies the identical data

From what I understand even if the migration re-runs, it's copying the same data from the same source to the same destination. There's no window where newer data could be overwritten with older data.

The race condition causes redundant work (migration runs twice) but doesn't create data loss or corruption.

If we want to eliminate the redundant work, we could add a completion marker, something like checking if the file already exists.

Does this align with you, or am I missing a scenario where data loss could occur?

app/core/EngineService/EngineService.ts
```diff
@@ -73,7 +73,11 @@ export class EngineService {
       parentContext: getUIStartupSpan(),
       tags: getTraceTags(reduxState),
     });
-    const state = reduxState?.engine?.backgroundState ?? {};
+
+    const state = isE2E
```
MarioAslau: if env file value is not set right, this will trigger a Vault Recovery on dev environments after locking and reopening of the application.

app/store/persistConfig/index.ts
MarioAslau: I am addressing these in the PR fix that I will push shortly. I missed to delete the warning, that was form me during development time.

app/store/persistConfig/index.ts
sethkfman, line 1: There should be a ReadMe added to the persistConfig directory with a flow chart and written explanation on how the new persist mechanism works. This can be done in a fast follow PR.

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +141,68 @@ export class EngineService {
...
+            async (controllerState: StateConstraint) => {
+              try {
+                // Filter out non-persistent fields based on controller metadata
+                const filteredState = getPersistentState(
+                  controllerState,
+                  // @ts-expect-error - EngineContext has stateless controllers, so metadata is not available
+                  UntypedEngine.context[controllerName as keyof EngineContext]?.metadata,
+                );
+
+                ReduxService.store.dispatch({
```
Gudahtt: I see that we're no longer using the update batcher for updating the Redux engine slice. What was the reason for that? That batcher was a significant performance improvement.

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  async removeItem(key: string) {
+    try {
+      return await FilesystemStorage.removeItem(key);
+    } catch (error) {
+      Logger.error(error as Error, {
+        message: `Failed to remove item for ${key}`,
+      });
+    }
+  },
+  async getKey(): Promise<Record<string, unknown>> {
```
Gudahtt: What is the "key" referred to in the name `getKey`? I was expecting this meant "get property" or something, with the property being a parameter. But there is no parameter here. It looks like this gets all persisted state.

app/store/migrations/index.ts
```diff
@@ -224,29 +226,131 @@ export const migrationList: MigrationsList = {
...
 // Enable both synchronous and asynchronous migrations
 export const asyncifyMigrations = (
   inputMigrations: MigrationsList,
   onMigrationsComplete?: (state: unknown) => void,
-) =>
-  Object.entries(inputMigrations).reduce(
+) => {
+  const versions = Object.keys(inputMigrations)
+    .map((v) => Number(v))
+    .filter((v) => Number.isFinite(v))
```
Gudahtt: This is curious. Are there migrations with non-finite migration numbers?

The sort is also curious, since they should already be sorted

app/store/persistConfig/index.ts
MarioAslau: Will add it into this PR actually to make it less confusing for reviewers! I have been working on it, I have been switching back and forth with addressing PR comments.

app/store/migrations/index.ts
```diff
@@ -224,29 +226,131 @@ export const migrationList: MigrationsList = {
...
+        if (Number(migrationNumber) === lastVersion && lastVersion > 104) {
+          const s2 = migratedState as StateWithEngine;
+          const hasControllers = Boolean(
+            s2.engine?.backgroundState &&
+            Object.keys(s2.engine.backgroundState).length > 0,
+          );
+          if (hasControllers) {
+            return await deflateToControllersAndStrip(migratedState);
+          }
+        }
```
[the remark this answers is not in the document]
MarioAslau: This keeps repeating

app/core/EngineService/EngineService.ts
```diff
@@ -73,7 +77,11 @@ export class EngineService {
       parentContext: getUIStartupSpan(),
       tags: getTraceTags(reduxState),
     });
-    const state = reduxState?.engine?.backgroundState ?? {};
+
+    const state = isE2E
+      ? reduxState?.engine?.backgroundState
+      : persistedState?.backgroundState ?? {};
```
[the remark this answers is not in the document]
MarioAslau: Repeated

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,219 @@
...
+        JSON.stringify(filteredState),
+      );
+
+      Logger.log(`${controllerName} state persisted successfully`);
+    } catch (error) {
+      Logger.error(error as Error, {
+        message: `Failed to persist ${controllerName} state`,
+      });
+    }
+  }, debounceMs);
```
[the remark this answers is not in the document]
MarioAslau: Repetead

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +141,68 @@ export class EngineService {
...
+                  controllerState,
+                  // @ts-expect-error - EngineContext has stateless controllers, so metadata is not available
+                  UntypedEngine.context[controllerName as keyof EngineContext]?.metadata,
+                );
+
+                ReduxService.store.dispatch({
+                  type: UPDATE_BG_STATE_KEY,
+                  payload: { key: controllerName },
+                });
+
```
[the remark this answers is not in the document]
MarioAslau: Repetead

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,219 @@
...
+      );
+
+      return { backgroundState };
+    } catch (error) {
+      Logger.error(error as Error, {
+        message: 'Failed to gather controller states',
+      });
+      return { backgroundState: {} };
+    }
+  },
```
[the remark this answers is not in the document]
MarioAslau: repeated

app/core/EngineService/EngineService.ts
MarioAslau: Addressed in PR changes

app/store/migrations/104.ts
MarioAslau: This was resolved in the live review

joaoloureirop:
Multiple review requests were modified by Github Pull Requests vscode extension while I was adding myself to the list

I've manually added back the removed teams

Sorry for the confusion

app/core/EngineService/EngineService.ts
```diff
@@ -73,7 +83,11 @@ export class EngineService {
       parentContext: getUIStartupSpan(),
       tags: getTraceTags(reduxState),
     });
-    const state = reduxState?.engine?.backgroundState ?? {};
+
+    const state = isE2E
+      ? reduxState?.engine?.backgroundState
+      : persistedState?.backgroundState ?? {};
```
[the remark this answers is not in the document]
MarioAslau: Repeating Cursor Alert.

app/store/persistConfig/index.ts
MarioAslau: renamed

app/store/migrations/index.ts
MarioAslau: This was defensive logic when debugging things added a while back. Removing it. Thanks for the catch Mark!

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,203 @@
...
+
+const persistConfig = {
+  key: 'root',
+  version,
+  blacklist: [
+    'rpcEvents',
+    'accounts',
+    'confirmationMetrics',
+    'alert',
+    'engine',
```
joaoloureirop, line 189: missing `qrKeyringScanner` item from [main](https://github.com/MetaMask/metamask-mobile/blob/afd4024c732e2a29cc9e350199b6219f3ec99492/app/store/persistConfig.ts#L143)

app/store/persistConfig/persistConfig.test.ts
```diff
@@ -0,0 +1,529 @@
...
+    it('have correct basic configuration', () => {
+      expect(persistConfig.key).toBe('root');
+      expect(persistConfig.version).toBe(version);
+      expect(persistConfig.timeout).toBe(40000);
+      expect(persistConfig.blacklist).toEqual([
+        'rpcEvents',
+        'accounts',
+        'confirmationMetrics',
+        'alert',
+        'engine',
```
joaoloureirop, line 137: missing `qrKeyringScanner` item from [main](https://github.com/MetaMask/metamask-mobile/blob/afd4024c732e2a29cc9e350199b6219f3ec99492/app/store/persistConfig.test.ts#L120)

joaoloureirop (changes requested):

app/core/EngineService/EngineService.ts
```diff
@@ -73,7 +83,11 @@
...
```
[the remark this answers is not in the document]
MarioAslau: This keeps repeating over and over.

app/core/EngineService/EngineService.ts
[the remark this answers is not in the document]
MarioAslau: This is ok... it keeps popping up

app/store/persistConfig/index.ts
joaoloureirop: addressed in https://github.com/MetaMask/metamask-mobile/pull/17685/commits/a6a17589ef20ebe13d385a1bf0af7aeedea1792b

app/core/EngineService/EngineService.ts
```diff
@@ -72,7 +72,11 @@ export class EngineService {
...
+
```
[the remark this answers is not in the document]
joaoloureirop: outdated

app/store/migrations/104.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  it('should handle invalid state gracefully', async () => {
+    const invalidState = null;
+
+    const result = await migration102(invalidState);
+
+    expect(result).toBe(invalidState);
+    expect(mockFilesystemStorage.setItem).not.toHaveBeenCalled();
+  });
+});
```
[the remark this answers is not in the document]
joaoloureirop: unrelated to this PR

app/store/migrations/104.ts
```diff
@@ -0,0 +1,101 @@
...
+            ),
+          );
+
+          // Preserve failed controller state to prevent data loss
+          if (hasProperty(controllers, controllerName)) {
+            failedControllerStates[controllerName] =
+              controllers[controllerName];
+          }
+        }
+      }
```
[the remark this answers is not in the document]
joaoloureirop: outdated

joaoloureirop (dismissed):

joaoloureirop (approved):

app/store/persistConfig/STATE_FLOW_DOCUMENTATION.md
```diff
@@ -0,0 +1,288 @@
+# MetaMask Mobile State Management Flow
+
+This document describes the complete state management flow in MetaMask Mobile, from app initialization through migrations, Redux persist, Engine initialization, and the new persistence system.
+
+## 🔄 Complete Application State Flow
+
+```
```
Prithpal-Sooriya, line 7: Instead of ascii diagrams, can we use mermaid diagrams? Claude generated this - feel free to modify

<details><summary>Mermaid diagram</summary>


```mermaid
flowchart TD
    A[App Starts] --> B[Redux Store<br/>Created with<br/>persistReducer]
    B --> C[persistStore Called<br/>Triggers<br/>Rehydration]
    C --> D{Migration<br/>Needed?}
    D -->|No| H[Redux Store<br/>Rehydrated with<br/>Migrated State]
    D -->|Yes| E[Run Migrations<br/>000-104<br/>During<br/>Rehydration]
    E --> F{Migration<br/>= 104?}
    F -->|Yes| G[Migration 104:<br/>Split Legacy<br/>Engine State<br/>Into Individual<br/>Controller Files]
    F -->|No| I[Legacy<br/>Migration]
    G --> H
    I --> H
    H --> J[onPersistComplete<br/>Callback<br/>Executed]
    J --> K[EngineService<br/>start Called]
    K --> L[Get Redux State<br/>& Controller<br/>Storage Files]
    L --> M[Engine.init<br/>Called with<br/>Combined State]
    M --> N[Controllers<br/>Initialized<br/>in Engine<br/>Context]
    N --> O[initializeControllers<br/>Subscribe to<br/>State Changes]
    O --> P[setupEnginePersistence<br/>Setup New<br/>Persistence<br/>System]
    P --> Q[Subscribe to<br/>All Controller<br/>State Change<br/>Events]
    Q --> R[🔄 Runtime Loop<br/>Controller<br/>Updates to<br/>Redux + File<br/>Storage]
```

</details>

app/store/persistConfig/STATE_FLOW_DOCUMENTATION.md
Prithpal-Sooriya, line 1: I do like living documentation for complex architecture breakdown, as long as maintainers can use and maintain this.

Is this document at a high level (no low level code examples) that this can be maintained with minimal changes?

app/store/persistConfig/STATE_FLOW_DOCUMENTATION.md
MarioAslau: @Prithpal-Sooriya I will make an additional ticket to modify this once it's merged, great suggestion !

MarioAslau: https://github.com/MetaMask/mobile-planning/issues/2339

app/store/persistConfig/index.ts
```diff
@@ -0,0 +1,204 @@
...
+    } catch (error) {
+      Logger.error(error as Error, {
+        message: `Failed to remove item for ${key}`,
+      });
+    }
+  },
+});
+
+export const ControllerStorage = {
+  // Use the consolidated storage without AsyncStorage fallback
```
asalsys, line 73: shouldn't the naming convention be "StorageController"

asalsys (commented):
I would update the ControllerStorage naming to StorageController

asalsys (approved):

Prithpal-Sooriya (approved):

app/store/persistConfig/index.ts
MarioAslau: the initial storage was named MigratedStorage, thus ControllerStorage. It's not a controller, as the other convention is for controllers KeyringController, AssetsController etc.

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +147,65 @@ export class EngineService {
...
+  private setupEnginePersistence = () => {
+    try {
+      if (UntypedEngine.controllerMessenger) {
+        BACKGROUND_STATE_CHANGE_EVENT_NAMES.forEach((eventName) => {
+          const controllerName = eventName.split(':')[0];
+
+          // Skip CronjobController state change events
+          // as they are handled separately in the CronjobControllerStorageManager.
+          // This prevents duplicate updates to the Redux store.
+          if (eventName === 'CronjobController:stateChange') {
```
andrepimenta, line 170: We are skipping persisting CronjobController, I believe this was being persisted before

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +147,65 @@ export class EngineService {
...
+            async (controllerState: StateConstraint) => {
+              try {
+                // Filter out non-persistent fields based on controller metadata
+                const filteredState = getPersistentState(
+                  controllerState,
+                  // @ts-expect-error - Engine context has stateless controllers, so metadata may not be available
+                  UntypedEngine.context[controllerName]?.metadata,
+                );
+
+                this.updateBatcher.add(controllerName);
```
andrepimenta, line 187: Should we add this here, or use it where it was being used before? Or at least, we should do it right away, before the `filteredState` code

app/core/EngineService/EngineService.ts
Gudahtt: I believe it's still being persisted via redux-persist through the `cronjobController` slice. This slice is not included on the persist config blacklist, so it should still be persisted.

The controller itself shouldn't be persisting anything via controller state directly, as it has its `persist` metadata set to `false`: https://github.com/MetaMask/snaps/blob/9aa68cdff235c503149425c149c9f5f4f0b41964/packages/snaps-controllers/src/cronjob/CronjobController.ts#L183

app/core/EngineService/EngineService.ts
Gudahtt: Agreed, I've prototyped this change here: https://github.com/MetaMask/metamask-mobile/pull/21155

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +147,65 @@ export class EngineService {
...
+
+                await persistController(filteredState, controllerName);
```
Gudahtt, line 189: Note that `filteredState` can be empty; some controllers don't persist any state. It looks like that would result in us storing `{}` as a separate item in storage for little benefit (and overwriting it each time any of the UI state changes!)

It would be ideal to skip the setup of this persistence function for any controllers with zero persisted state.

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +147,65 @@ export class EngineService {
...
-      // as they are handled separately in the CronjobControllerStorageManager.
-      // This prevents duplicate updates to the Redux store.
-      if (eventName === 'CronjobController:stateChange') {
-        return;
+          UntypedEngine.controllerMessenger.subscribe(
+            eventName,
+            async (controllerState: StateConstraint) => {
+              try {
+                // Filter out non-persistent fields based on controller metadata
+                const filteredState = getPersistentState(
```
Gudahtt, line 181: We're filtering to just persisted state here, but changes to non-persisted state would still trigger an unnecessary write.

Rather than deriving the data we want here, we can derive it using a messenger subscription selector instead to ensure that the handler only gets triggered if a state change actually happens. e.g.:

```
controllerMessenger.subscribe(
  eventName,
  async (persistedControllerState: Json) => {
    ...
  },
  (state: StateConstraint) => getPersistedState(state, UntypedEngine.context[controllerName]?.metadata),
);
```

app/core/EngineService/EngineService.ts
Gudahtt: The CronjobController gets persisted via a separate slice rather than via the normal method because we really needed it to be independently persisted (for performance reasons) and we couldn't wait for state splitting.

Now that we have state splitting, we could consider handling it the normal way. But it would require controller changes, and extension doesn't have state splitting yet. Also we'd need to consider a migration. Maybe easiest to leave it as-is until we have state splitting on both clients, since it's a top priority for extension now as well.

app/core/EngineService/EngineService.ts
Gudahtt: This strategy breaks if we're using the same subscriber for updating the UI as well. But we can handle that separately, as mentioned here: https://github.com/MetaMask/metamask-mobile/pull/17685/files#r2430316734

app/core/EngineService/EngineService.ts
Gudahtt: If we do this, we don't need the `CronjobController` check anymore (since it has no persisted state).

app/core/EngineService/EngineService.ts
```diff
@@ -130,25 +147,65 @@ export class EngineService {
...
+          );
+        });
+        Logger.log(
+          'Individual controller persistence and Redux update subscriptions set up successfully',
+        );
       }
-      engine.controllerMessenger.subscribe(eventName, () =>
-        update_bg_state_cb(eventName.split(':')[0]),
+    } catch (error) {
+      Logger.error(
```
Gudahtt, line 204: I don't think it's safe to let wallet initialization continue in this case. Better to crash, so that we discover the problem sooner. The user will be in for a very unpleasant surprise if nothing is getting persisted.

Gudahtt: Logging the error and continuing makes a bit more sense if persistence fails for a specific state update, because it's at least possible that might be caused by a lower severity issue. But that could be a Sev1 too.

TBH I'd err on the side of crashing in that case too, but logging seems fine as long as we're confident we can detect and react quickly if that does happen.

app/store/migrations/index.ts
```diff
@@ -224,29 +227,169 @@ export const migrationList: MigrationsList = {
...
+            captureException(
+              new Error(
+                `deflateToControllersAndStrip: Failed to save ${controllerName} to individual storage: ${String(
+                  error,
+                )}`,
+              ),
+            );
+
+            // Preserve failed controller state to prevent data loss
+            failedControllerStates[controllerName] = controllerState;
```
Gudahtt, line 325: This is a pretty interesting strategy. I love the idea of saving state that we can't effectively migrate, but where does that leave the user _for the current session_ if this fails?

If we only succeed in migrating part of the state, then the failed controller state will remain persisted for us to fix with a later migration (in theory), but we're not crashing the app, so a user can still use the wallet. That will result in the failed controller state being replaced by runtime activity.

At best we'd have to merge two versions of the same state, which might not be possible to do safely. And the user would get no warning about the app potentially having data loss, or the dangers of proceeding despite that problem. At worst we could allow a lot more state corruption to happen (because of how controller state is inter-related), even if we do attempt to save a copy.

That is, I am concerned that this approach of stashing state that we weren't able to migrate leaves us in a much worse condition than crashing upon any migration failure.

MarioAslau: Concerns were addressed in: https://github.com/MetaMask/metamask-mobile/pull/21225

app/core/EngineService/EngineService.ts
MarioAslau: Concerns were addressed in: https://github.com/MetaMask/metamask-mobile/pull/21225

PR #17688, opened 2025-07-29
feat: perps transactions history

nickewansmith:
## **Description**
- We want to view the users transaction history for perps fills, orders and funding.
3. What is the improvement/solution?
- Adds the perps transaction view as a tab in the current transactions view

## **Changelog**
`CHANGELOG entry: Added a new transactions view for perps fills, orders and user funding fees`

CHANGELOG entry:

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1027

## **Manual testing steps**

1. Go to the transactions history view
2. click on the Perps tab
5. Click on the orders and funding tabs and go back to the fills tab
6. Each tab list item should have a detail screen according to the figma design in ticket - new designs are in the Design folder at the bottom

### **After**

https://github.com/user-attachments/assets/b918a87a-f090-4e53-ac65-506057865c40


app/components/UI/Perps/Views/PerpsView.tsx
```diff
@@ -272,6 +272,14 @@ const PerpsView: React.FC<PerpsViewProps> = () => {
...
     navigation.navigate(Routes.PERPS.POSITIONS);
   };
+  const handleTransactionsHistoryNavigation = async () => {
+    navigation.navigate(Routes.TRANSACTIONS_VIEW, {
+      screen: Routes.TRANSACTIONS_VIEW,
+      params: {
+        redirectToPerpsTransactions: true,
+      },
+    });
+  };
```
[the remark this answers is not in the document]
nickewansmith: If there are any suggestions on how to do this another way feel free to comment. I could not seem to find another way to get to the view with params.

app/components/UI/Perps/components/PerpsTransactionDetailAssetHero/PerpsTransactionDetailAssetHero.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const PerpsTransactionDetailAssetHero: React.FC<
+  PerpsTransactionDetailAssetHeroProps
+> = ({ transaction, styles }) => {
+  const { assetUrl } = usePerpsAssetMetadata(transaction.asset);
+
+  return (
+    <View
+      testID="perps-transaction-detail-asset-hero"
```
abretonc7s: maybe we should keep using the selector file consistently from `e2e/selectors/Perps/Perps.selectors.ts`

app/components/UI/Perps/components/PerpsTransactionItem/PerpsTransactionItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  renderRightContent,
+}) => {
+  const { assetUrl } = usePerpsAssetMetadata(item.asset);
+
+  return (
+    <TouchableOpacity
+      testID="transaction-item"
+      style={styles.transactionItem}
+      onPress={() => onPress(item)}
+      activeOpacity={0.7}
```
abretonc7s: We could leverage a common config for opacity around perps like           activeOpacity={PERPS_CHART_CONFIG.INTERVAL_SELECTOR_OPACITY}

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -173,7 +179,7 @@ export class HyperLiquidProvider implements IPerpsProvider {
   /**
    * Place an order using direct wallet signing (same as working debug test)
    */
-  async placeOrder(params: PerpsOrderParams): Promise<OrderResult> {
+  async placeOrder(params: OrderParams): Promise<OrderResult> {
```
abretonc7s, line 183: any reasons to rename here?

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -905,6 +911,190 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+            amountUsd: delta.usdc,
+            rate: delta.fundingRate,
+            timestamp: time,
+            transactionHash: hash,
+          };
+        });
+
+        return funding;
+      } catch (fundingError) {
+        DevLogger.log('userFunding failed', fundingError);
```
abretonc7s: I wonder if we should display an error instead of defaulting to empty array, might be confusing to users.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -905,6 +911,190 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+        return [];
+      }
+    } catch (error) {
+      DevLogger.log('Error getting user funding:', error);
+      return [];
```
abretonc7s, line 1090: same here

app/components/UI/Perps/controllers/types/index.ts
```diff
@@ -264,11 +264,15 @@ export interface PriceUpdate {
 export interface OrderFill {
   orderId: string; // Order ID that was filled
   symbol: string; // Asset symbol
-  side: string; // Order side (from SDK)
+  side: string; // Normalized order side ('buy' or 'sell')
   size: string; // Fill size
   price: string; // Fill price
+  pnl: string; // PNL
+  direction: string; // Direction of the fill
```
abretonc7s, line 273: is direction same as side?

app/components/UI/Perps/controllers/types/index.ts
```diff
@@ -264,11 +264,15 @@ export interface PriceUpdate {
...
   fee: string; // Fee paid
+  feeToken: string; // Fee token symbol
```
abretonc7s, line 275: is this always USDC in the case of hyperliquid?

app/components/UI/Perps/controllers/types/index.ts
```diff
@@ -368,6 +420,11 @@ export interface IPerpsProvider {
   // Note: deposit() is handled by PerpsController routing (blockchain operation)
   validateDeposit(params: DepositParams): { isValid: boolean; error?: string }; // Protocol-specific deposit validation
 
+  // Historical data operations
+  getOrderFills(params?: GetOrderFillsParams): Promise<OrderFill[]>; // Historical trade fills
+  getOrders(params?: GetOrdersParams): Promise<Order[]>; // Historical order lifecycle
+  getFunding(params?: GetFundingParams): Promise<Funding[]>; // Historical funding payments
+
```
abretonc7s: I think we should have some more comments on differentiating the different types of orders, maybe with some examples in the ts-doc. It would be easier when we get back to it as it is easy to forget the difference.

app/components/UI/Perps/routes/index.tsx
```diff
@@ -128,6 +128,20 @@ const PerpsScreenStack = () => (
       - PerpsOrderHistoryView (Order history)
       - PerpsOrderDetailsView (Order details)
     */}
+      {/*
```
abretonc7s: might not need to keep this comment

app/components/UI/Perps/utils/blockchainUtils.ts
abretonc7s, line 1: this should probably be something we put at the IperpsProvider level, like provider.getExplorerUrl() and it gives the correct one based on active network. Otherwise it won't work with other protocols.

app/components/UI/Perps/utils/transactionDetailStyles.ts
```diff
@@ -0,0 +1,98 @@
+import type { Theme } from '../../../../util/theme/models';
+
+/**
+ * Shared styles for Perps transaction detail views
+ * Extracts common styling patterns used across PerpsOrderTransactionView,
+ * PerpsPositionTransactionView, and PerpsFundingTransactionView
+ */
+export const createTransactionDetailStyles = (theme: Theme) => {
```
abretonc7s, line 9: we should return from StyleSheets.create to avoid having all the as const.

abretonc7s: Also we need to have another full pass on design to avoid manually setting the weight etc (leveraging design system tokens)

app/components/UI/Perps/utils/transactionTransforms.ts
abretonc7s, line 1: This seems like hyperliquid specific uitlities, maybe we can move them to a hyperliquid folder and prefix with hyperliquid?

app/components/UI/Perps/utils/transactionTransforms.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    // Create safe amount strings
+    const isPositive = BigNumber(amountUsd).isGreaterThan(0);
+    const amountUSDC = `${isPositive ? '+' : '-'}$${BigNumber(amountUsd)
+      .absoluteValue()
+      .toString()}`;
+
+    return {
+      id: `funding-${timestamp}-${symbol}`,
+      type: 'funding' as const,
```
abretonc7s: we could type the return object to avoid the const

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
```diff
@@ -0,0 +1,159 @@
...
+  // Set navigation title
+  navigation.setOptions(
+    getPerpsTransactionsDetailsNavbar(navigation, transaction.title),
+  );
+
+  const handleViewOnBlockExplorer = () => {
+    if (!selectedInternalAccount) {
+      return;
+    }
+    Linking.openURL(
```
abretonc7s: I wonder if we should open this from the in-app browser or the system

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
```diff
@@ -0,0 +1,159 @@
...
+              transaction.fundingAmount?.isPositive ? '+' : '-'
+            }${formatPerpsFiat(Math.abs(feeNumber))}`,
+    },
+    { label: 'Rate', value: transaction.fundingAmount?.rate },
+  ];
+
+  return (
+    <ScreenView>
+      <ScrollView
+        testID="perps-funding-transaction-view"
```
abretonc7s: perps selector for e2e ids

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
```diff
@@ -0,0 +1,159 @@
...
+                <Text variant={TextVariant.BodySM} color={TextColor.Default}>
+                  {detail.value}
+                </Text>
+              </View>
+            ))}
+          </View>
+
+          {/* Block explorer button */}
+          <Button
+            testID="block-explorer-button"
```
abretonc7s: same

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
```diff
@@ -0,0 +1,159 @@
...
+    Linking.openURL(
+      getHyperliquidExplorerUrl(perpsNetwork, selectedInternalAccount.address),
+    );
+  };
+
+  const feeNumber = transaction.fundingAmount?.feeNumber || 0;
+
+  // Funding detail rows based on design
+  const detailRows = [
+    { label: 'Date', value: formatTransactionDate(transaction.timestamp) },
```
abretonc7s: We may need to i18n those labels

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsOrderTransactionView.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const { totalFee, protocolFee, metamaskFee } = usePerpsOrderFees({
+    orderType: transaction?.order?.type ?? 'market',
+    amount: transaction?.order?.size ?? '0',
+  });
+
+  if (!transaction) {
+    return (
+      <ScreenView>
+        <View style={styles.content}>
+          <Text>Transaction not found</Text>
```
abretonc7s: i18n

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsOrderTransactionView.tsx
```diff
@@ -0,0 +1,219 @@
...
+      return;
+    }
+    Linking.openURL(
+      getHyperliquidExplorerUrl(perpsNetwork, selectedInternalAccount.address),
+    );
+  };
+
+  // Main detail rows based on design
+  const mainDetailRows = [
+    { label: 'Date', value: formatTransactionDate(transaction.timestamp) },
```
abretonc7s: i18n

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsOrderTransactionView.tsx
abretonc7s, line 1: we should make sure all strings are i18n'ed and using the e2e perps selector

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsPositionTransactionView.tsx
```diff
@@ -0,0 +1,229 @@
...
+        </View>
+      </ScreenView>
+    );
+  }
+
+  const handleViewOnBlockExplorer = () => {
+    if (!selectedInternalAccount) {
+      return;
+    }
+    Linking.openURL(
```
abretonc7s: this should be abstracted (not hyperliquid specific). Maybe a new hook usePerpsExplorerUrl()

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsPositionTransactionView.tsx
abretonc7s, line 1: same general comment on i18n and selector ids

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.styles.ts
abretonc7s, line 1: I see whitespace only changes, maybe we should run a yarn format for consistency

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
```diff
@@ -0,0 +1,587 @@
...
+// Union type for FlashList items (headers + transactions)
+type ListItem =
+  | { type: 'header'; title: string; id: string }
+  | { type: 'transaction'; transaction: PerpsTransaction; id: string };
+
+type FilterTab = 'Trades' | 'Orders' | 'Funding';
+
+interface PerpsTransactionsViewProps {}
+
+const styleSheet = (params: { theme: Theme }) => {
```
abretonc7s: should be in a separate file for the styles and consistency

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
```diff
@@ -0,0 +1,587 @@
...
+import { usePerpsConnection, usePerpsTrading } from '../../hooks';
+import {
+  transformFillsToTransactions,
+  transformFundingToTransactions,
+  transformOrdersToTransactions,
+} from '../../utils/transactionTransforms';
+import PerpsTransactionItem from '../../components/PerpsTransactionItem';
+
+// Perps transaction data structure matching the new design
+export interface PerpsTransaction {
```
abretonc7s: do we want to duplicate the interface or could we re-use the order types?

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
```diff
@@ -0,0 +1,587 @@
...
+          scrollEnabled={false}
+        >
+          {filterTabs.map(renderFilterTab)}
+        </ScrollView>
+      </View>
+
+      <FlashList
+        ref={flashListRef}
+        data={flatListData}
+        renderItem={renderListItem}
```
abretonc7s, line 358: we should probably use const for these hardcoded size elements

abretonc7s (dismissed):

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: No, will change back

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.test.tsx
```diff
@@ -0,0 +1,360 @@
...
+}));
+
+jest.mock('../../utils/blockchainUtils', () => ({
+  getHyperliquidExplorerUrl: () => mockGetHyperliquidExplorerUrl(),
+}));
+
+// jest.mock('../../utils/formatUtils', () => ({
+//   // formatPerpsFiat: (value: string | number) => mockFormatPerpsFiat(value),
+//   // formatTransactionDate: () => mockFormatTransactionDate(),
+// }));
```
gambinish: Can this be removed?

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
```diff
@@ -0,0 +1,159 @@
...
+import { useSelector } from 'react-redux';
+import { selectSelectedInternalAccount } from '../../../../../selectors/accountsController';
+import { usePerpsNetwork } from '../../hooks';
+import {
+  formatPerpsFiat,
+  formatTransactionDate,
+} from '../../utils/formatUtils';
+import { BigNumber } from 'bignumber.js';
+
+// Interface now imported from PerpsTransactionsView
```
gambinish: This looks like it can probably be removed

app/components/Views/ActivityView/index.js
```diff
@@ -224,6 +242,13 @@ const ActivityView = () => {
               tabLabel={strings('fiat_on_ramp_aggregator.orders')}
             />
           )}
+          {isPerpsEnabled && (
+            <PerpsConnectionProvider
+              tabLabel={strings('perps.transactions.title')}
+            >
+              <PerpsTransactionsView />
+            </PerpsConnectionProvider>
+          )}
```
gambinish, line 251: I looks like the title for this tab is `Perps` but since this is user facing, I think this should be `Perpetuals`

Screenshot 2025-08-05 at 11 49 40 AM

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -202,24 +202,24 @@
...
 		51AB7231D0E692F5EF71FACB /* Pods-MetaMask-QA.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-MetaMask-QA.debug.xcconfig"; path = "Target Support Files/Pods-MetaMask-QA/Pods-MetaMask-QA.debug.xcconfig"; sourceTree = "<group>"; };
 		57C103F40F394637B5A886FC /* FontAwesome5_Brands.ttf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = FontAwesome5_Brands.ttf; path = "../node_modules/react-native-vector-icons/Fonts/FontAwesome5_Brands.ttf"; sourceTree = "<group>"; };
-		5914F547268E4F7BB337A3DF /* Geist Bold Italic.otf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = undefined; includeInIndex = 0; lastKnownFileType = unknown; name = "Geist Bold Italic.otf"; path = "../app/fonts/Geist Bold Italic.otf"; sourceTree = "<group>"; };
+		5914F547268E4F7BB337A3DF /* Geist Bold Italic.otf */ = {isa = PBXFileReference; explicitFileType = undefined; fileEncoding = 9; includeInIndex = 0; lastKnownFileType = unknown; name = "Geist Bold Italic.otf"; path = "../app/fonts/Geist Bold Italic.otf"; sourceTree = "<group>"; };
```
gambinish: Can you help me understand why these encoding changes are necessary?

locales/languages/en.json
```diff
@@ -1139,6 +1139,9 @@
     "developer_options": {
       "hyperliquid_network_toggle": "Hyperliquid Network Toggle",
       "perps_sandbox_button": "Perps Sandbox"
+    },
+    "transactions": {
+      "title": "Perps"
```
gambinish:
```suggestion
      "title": "Perpetuals"
```

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
gambinish: ++

We should probably do a once over to make sure all hardcoded user facing strings are localized

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsOrderTransactionView.tsx
```diff
@@ -0,0 +1,219 @@
...
+import PerpsTransactionDetailAssetHero from '../../components/PerpsTransactionDetailAssetHero';
+import { usePerpsNetwork, usePerpsOrderFees } from '../../hooks';
+import { PerpsNavigationParamList } from '../../types/navigation';
+import { getHyperliquidExplorerUrl } from '../../utils/blockchainUtils';
+import {
+  formatPerpsFiat,
+  formatTransactionDate,
+} from '../../utils/formatUtils';
+
+// Interface now imported from PerpsTransactionsView
```
gambinish: This comment can probably be removed

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsPositionTransactionView.tsx
```diff
@@ -0,0 +1,229 @@
...
+      getHyperliquidExplorerUrl(perpsNetwork, selectedInternalAccount.address),
+    );
+  };
+
+  // Asset display component with metadata
```
gambinish: This comment still relevant? Seems like it's a floater

app/components/UI/Perps/Views/PerpsView.tsx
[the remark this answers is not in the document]
gambinish: I think this is fine to keep as is, especially since this is our sandbox env

app/components/UI/Perps/components/PerpsTransactionItem/PerpsTransactionItem.test.tsx
```diff
@@ -0,0 +1,319 @@
...
+        item={transactionWithoutSubtitle}
+        styles={mockStyles}
+        onPress={mockOnPress}
+        renderRightContent={mockRenderRightContent}
+      />,
+    );
+
+    // This would require checking the applied styles, which is complex in React Native Testing Library
+    // The test ensures the component renders without crashing when subtitle is empty
+    expect(true).toBeTruthy();
```
gambinish: This is an odd test. I don't think it's testing if the subtitle is centered, but just that it renders. This might be a use case for some kind of snapshot test?

gambinish: I generally don't think snapshot tests are super useful, we don't need to make this change if you don't think it's necessary. Also open to just removing the test as well

app/components/UI/Perps/utils/transactionDetailStyles.ts
gambinish: Yeah, generally we likely want to use the design tokens like `TextVariant` etc so that we stay up to date with the changes design team makes

app/components/UI/Perps/utils/transactionTransforms.ts
```diff
@@ -0,0 +1,215 @@
...
+      statusText = 'Canceled';
+    } else if (isRejected) {
+      orderStatusType = 'canceled'; // Map rejected to canceled
+      statusText = 'Rejected';
+    } else if (isTriggered) {
+      orderStatusType = 'filled'; // Map triggered to filled
+      statusText = 'Triggered';
+    } else {
+      orderStatusType = 'pending';
+      statusText = isOpened ? '' : 'Queued';
```
gambinish: Question: Do enums make sense here instead of the hard coded strings?

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
nickewansmith: I will try to slim this down a bit. I don't want to alter the design too much as there is a lot of code depending on it rn and that would require a good amount of refactoring. I'll see what I can do here.

app/components/Views/ActivityView/index.js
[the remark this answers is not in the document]
nickewansmith: This actually was placed on the `PerpsConnectionProvider` due to not displaying a level lower. It does work as expected currently.

nickewansmith: I agree with you, I'm fine with changing it, but got this from the newest figma design
Screenshot 2025-08-05 at 9 57 37 PM

ios/MetaMask.xcodeproj/project.pbxproj
nickewansmith: I will remove this, it keeps popping up locally and getting included.

locales/languages/en.json
nickewansmith: As mentioned, cautious about changing this as it was in latest designs but I do agree.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: Actually, I see that there is no name conflict, all other type names from the same module do not follow this pattern except for `PerpsMarketData` which maybe should be switched to `MarketData`.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: I have followed the existing pattern here for now. I agree that if there are errors from the client we should leave it up to the view as to whether or not to display those errors.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: I am following the existing pattern here. Maybe we can do a pass on the provider error handling as a whole in another PR?

app/components/UI/Perps/controllers/types/index.ts
nickewansmith: It is related but not the same. In the case of HL, `direction` is Closed Long, Closed Short, Opened Long, Opened Short, {Closed/Opened} Long > Short, {Closed/Opened}Short > Long. Because HL returns this flipped fill direction, I chose this as the abstract since other providers may not report flipped fills in this way.

app/components/UI/Perps/controllers/types/index.ts
nickewansmith: It could be HYPE, but I have only seen that with a spot trade which I am filtering out.

app/components/UI/Perps/utils/transactionTransforms.ts
nickewansmith: This is not HL specific, it is specific to the transactions list for any perps provider.

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
nickewansmith: I have tried both but they seem to look the same. Maybe we can leave as-is for now and circle back to this when we audit everything.

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
nickewansmith: I agree that this type could be cleaner, there is some repetition here. We have discussed it and agreed it will be ok. 

To further explain, this transformation of the perps types to a PerpsTransaction type is a transformation that would need to happen somewhere as this is basically what the view will display. The keys here are pretty similar but they are related moreso to the transaction list requirements.

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsFundingTransactionView.tsx
nickewansmith: It is now showing the SimpleWebView similar to typical transactions history links

app/components/UI/Perps/utils/transactionTransforms.ts
```diff
@@ -0,0 +1,221 @@
...
+      timestamp,
+      asset: symbol,
+      fill: {
+        shortTitle: `${action} ${
+          isFlipped
+            ? direction?.toLowerCase() || ''
+            : part2?.toLowerCase() || ''
+        }`,
+        amount: amountUSD,
+        amountNumber: parseFloat(parseFloat(amount).toFixed(2)),
```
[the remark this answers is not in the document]
nickewansmith: Ack and agreed, I'll take a look at this again now that things are more stable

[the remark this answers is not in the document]
nickewansmith: removed double parseFloat, and ensure default of zero

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsPositionTransactionView.tsx
```diff
@@ -0,0 +1,228 @@
...
+        ? `${
+            greaterThanCent
+              ? formatPnl(transaction.fill?.pnl)
+              : `+$${transaction.fill?.pnl}`
+          }`
+        : `${
+            lessThanNegCent
+              ? formatPnl(transaction.fill?.pnl)
+              : `-$${Math.abs(parseFloat(transaction.fill?.pnl))}`
+          }`,
```
gambinish, line 129: Non blocking: I wonder if we should eventually break this out into our utility file to make the component code a bit cleaner

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
```diff
@@ -0,0 +1,391 @@
...
+        })}
+      </Text>
+      <Text style={styles.emptyText}>
+        {strings('perps.transactions.empty_state.history_will_appear')}
+      </Text>
+    </View>
+  );
+
+  const filterTabs: FilterTab[] = useMemo(
+    () => ['Trades', 'Orders', 'Funding'],
```
gambinish: Do these need to be localized? Or are these not user facing?

ios/MetaMask.xcodeproj/project.pbxproj
gambinish: Reminder to remove this before merging

gambinish (dismissed):

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsTransactionsView.tsx
nickewansmith: I missed these, I'll add them

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsPositionTransactionView.tsx
nickewansmith: Agreed

app/components/UI/Perps/utils/transactionTransforms.ts
```diff
@@ -0,0 +1,224 @@
...
+    } else {
+      amountBN = BigNumber(fill.size).times(fill.price).plus(fill.fee);
+      amount = `${isPositive ? '+' : '-'}${amountBN.toString()} ${
+        fill.feeToken
+      }`;
+    }
+
+    const absAmount = Math.abs(parseFloat(amount)).toFixed(2);
+    const amountUSD = `${isPositive ? '+' : '-'}$${absAmount}`;
+
```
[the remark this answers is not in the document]
gambinish: Is this legit?

gambinish (dismissed):

app/components/UI/Perps/utils/transactionTransforms.ts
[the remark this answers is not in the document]
nickewansmith: I think it is actually, good catch. I stepped out for a sec but will fix when I get back.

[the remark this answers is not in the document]
nickewansmith: It should display the fee token not the symbol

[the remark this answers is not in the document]
nickewansmith: Looking even closer, this code was somewhat outdated with pieces left that were not needed. The tests passed bc it actually worked in the ui unintentionally due to stripping the string. I have simplified the logic and the test still passes now. Ui tests well.

gambinish (dismissed):

nickewansmith:
I have added a minimal change here that does not affect output. Bitrise tests were mostly passing before this across runs, and collectively across runs have passed all tests.

gambinish (approved):

app/components/UI/Perps/Views/PerpsTransactionsView/PerpsPositionTransactionView.tsx
```diff
@@ -0,0 +1,228 @@
...
+      textColor: BigNumber(transaction.fill?.pnl).isGreaterThanOrEqualTo(0)
+        ? TextColor.Success
+        : TextColor.Error,
+    });
```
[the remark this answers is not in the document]
nickewansmith: This is ok for now it should be greaterThanOrEqual to zero, we can address this when moving this logic into a util.

PR #17789, opened 2025-07-31
perf: Update FlashList to v2 cp-7.53.0

### LINKED ISSUE #17541: [Bug]: Performance degradation Android Accounts list
### Describe the bug

Performance degradation observed for Power user account (Using SRP with over 50 accounts):

- Accounts list slower to load (sometimes it does not load all accounts until clicking on it)
- Switching accounts slower to load
- 

Unable to reproduce on iOS, only Android

### Expected behavior

_No response_

### Screenshots/Recordings

https://github.com/user-attachments/assets/e66b72fc-cd17-45ac-98d7-e85b6c3a0e61

https://github.com/user-attachments/assets/f7e23517-1260-4cfc-9d8b-325ea43a0193

https://github.com/user-attachments/assets/7f90a8fb-b9d3-4787-a3ea-be9b05deb248

### Steps to reproduce

1. Import SRP with 50+ accounts
2. Click on Accounts list
3. Switch accounts

```shell

```

### Detection stage

During release testing

### Version

7.53.0

### Build type

None

### Device

Pixel 6 Pro

### Operating system

Android

### Additional context

_No response_

### Severity

_No response_

owencraston:
## **Description**
- Performance regressions was found during release testing for 7.53.0
- I believe the root cause of this issue is the following...
    - The initial state of the accounts controller is an empty array[]
    - FlashList uses the size of the list combined with the `estimatedItemSize` prop to draw the list
    - Since 7.51 and 7.53, we introduced multi chain accounts state 1 which changes how the accounts list is rendered. The size of elements is more dynamic (since now we render wallet name, account cell and footers). This combined with the fact that the initial list size starts at 0 was causing issues for the native android RecyclerListView. 
    - This warning was logged in the console while testing:
    - `WARN  FlashList's rendered size is not usable. Either the height or width is too small (<2px). Please make sure that the parent view of the list has a valid size. FlashList will match the size of the parent.1`
        - There are several issues online with [this warning](https://stackoverflow.com/questions/73516978/flashlists-rendered-size-is-not-usable).
- Why update to FlashList v2 instead of fixing the root cause of the issue?
    - Potential solution 1: Give the list a fixed height
        - This is not possible since the list lives in a bottom drawer popup that has a dynamic height
    - Potential Solution 2: Stop the useAccounts hook from returning an empty array as its initial state.
        - This is a much more difficult fix since that hook combines logic from over 7 selectors. 
        - hooks are expected to have state updates and thus handling the dynamic size of the data is a normal react construct. If anything it was a bug in FlashList.
- Update to FlashList v2 for the following reasons:
   - FlashList v2 handles the dynamic height and list size values without issues
   - Easier to use API (no more need for estimatedSize prop)
   - No more fixed height requirement 
   - Fixes the scroll to selected account issue that was present since we migrated to FlashList v1

### Resources
- [Migrating from FlashList v1 to v2](https://shopify.github.io/flash-list/docs/v2-migration)
- [FlashList v2: A Ground-Up Rewrite for React Native's New Architecture](https://shopify.engineering/flashlist-v2)
- [What's new in v2](https://shopify.github.io/flash-list/docs/v2-changes/)

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/17541

## **Manual testing steps**

1. Import an SRP with many accounts or create many accounts within your existing wallet
2. open the account list 
3. the list should open within a reasonable amount of time and not require a touch event to complete the rendering cycle
4. the list should scroll to the currently selected account

### **Before**

https://github.com/user-attachments/assets/ba7a15ff-5521-4239-a875-65c0ed88d491

### **After**

https://github.com/user-attachments/assets/01ab18ef-d0f4-4c94-8568-0374c936dbba


app/components/Views/AccountSelector/AccountSelector.tsx
```diff
@@ -52,9 +51,7 @@ import { ButtonProps } from '../../../component-library/components/Buttons/Butto
 import { useSyncSRPs } from '../../hooks/useSyncSRPs';
 
 const AccountSelector = ({ route }: AccountSelectorProps) => {
-  const { height: screenHeight } = useWindowDimensions();
```
owencraston, line 55: We no longer need to pass in a fixed height to the parent so we can remove this. The account list will now automatically take up the height it needs to render the accounts on the visible portion of the screen.

matthewwalsh0 (dismissed):

Matt561 (dismissed):
Approving `PerpsMarketListView` change on behalf of Perps 👍

package.json
```diff
@@ -312,7 +312,7 @@
     "@sentry/core": "~8.54.0",
     "@sentry/react": "~8.54.0",
     "@sentry/react-native": "~6.10.0",
-    "@shopify/flash-list": "^1.7.6",
+    "@shopify/flash-list": "2.0.1",
```
tommasini, line 315: What's the reasoning for not be "^2.0.1"?

tommasini (dismissed):
LGTM!

package.json
owencraston: I don't have a great answer here. Typically I like to pin our packages to specific versions to avoid anything unexpectedly breaking.

vinnyhoward (dismissed):

gantunesr (dismissed):

gambinish (dismissed):

gambinish (approved):

gantunesr (approved):

nickewansmith (approved):
LGTM

vinnyhoward (approved):

Cal-L (approved):
LGTM

sleepytanya (approved):

PR #17900, opened 2025-08-04
ci: Only do required setup for CI step

FrederikBolding:
## **Description**

Skip doing iOS / Android specific setup where possible, e.g. when running unit tests.

The PPOM build is always required and cannot be skipped anymore.


Prithpal-Sooriya (approved):
LFG

PR #17904, opened 2025-08-04
feat: add MultichainAccountSelectorList

PatrykLucka:
## **Description**
This PR introduces a new MultichainAccountSelectorList component to enable the multichain account management UI.

**In scope of this PR:**
- new MultichainAccountSelectorList component is used when the feature flag is enabled
- storybook for the new component
- "dumb" search input is added
- selecting accounts is possible

**Out of scope:**
- Account cell (it is already there)
- Create account
- Context menu
- Search
- Add wallet
- Balance display

## **Changelog**

CHANGELOG entry: Added new MultichainAccountSelectorList used under feature flag

## **Related issues**

Fixes:
Jira ticket: https://consensyssoftware.atlassian.net/browse/MUL-299

## **Manual testing steps**

1. Enable Multichain accounts feature flag
2. Go to account list 

## **Screenshots/Recordings**

https://github.com/user-attachments/assets/baf664d3-162b-435e-99e7-08f9f495475a


app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+  WalletSection,
+} from './MultichainAccountSelectorList.types';
+import createStyles from './MultichainAccountSelectorList.styles';
+import {
+  MULTICHAIN_ACCOUNT_SELECTOR_LIST_TESTID,
+  MULTICHAIN_ACCOUNT_SELECTOR_SEARCH_INPUT_TESTID,
+  MULTICHAIN_ACCOUNT_SELECTOR_EMPTY_STATE_TESTID,
+} from './MultichainAccountSelectorList.constants';
+
+const MultichainAccountSelectorList = ({
```
gantunesr, line 29: Why not call it `MultichainAccountList`?

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+    walletSections.forEach((section) => {
+      items.push({
+        type: 'header',
+        data: { title: section.title, walletName: section.walletName },
+      });
+
+      section.data.forEach((accountGroup) => {
+        items.push({
+          type: 'account',
```
gantunesr: I'd probably call this type `body` or `cell` to be more consistent with `header` and `footer`

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.types.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface MultichainAccountSelectorListProps
+  extends Partial<FlashListProps<FlattenedMultichainAccountListItem>> {
+  /**
+   * Callback when an account is selected
+   */
+  onSelectAccount?: (accountGroup: AccountGroupObject) => void;
+  /**
+   * Selected account group
+   */
+  selectedAccountGroup?: AccountGroupObject;
```
gantunesr: Unsure if this one should be optional, what was your reasoning for it?

app/components/Views/AccountSelector/AccountSelector.test.tsx
```diff
@@ -201,4 +234,23 @@ describe('AccountSelector', () => {
     );
     expect(addButton).toBeDefined();
   });
+
+  it('should render account selector with multichain support', () => {
```
gantunesr:
```suggestion
  it('renders account selector with multichain support', () => {
```

gantunesr (commented):
Looking good, left some minor comments

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
PatrykLucka: Wanted to keep it consistent. It replaces `EvmAccountSelectorList`, which should be replaced by `CaipAccountSelectorList`. Also it is used only in `AccountSelector` component, so I thought keeping "Selector" part makes sense

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
PatrykLucka: yea, `cell` makes sense, we will have multiple cells between header and footer, so I prefer that than having multiple "bodies" 😅

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.types.ts
PatrykLucka: `EvmAccountSelectorList` had selected list optional, but just checked that in `CaipAccountSelectorList` it's already required, so it probably makes sense to make it required 👍

app/components/Views/AccountSelector/AccountSelector.tsx
```diff
@@ -119,6 +127,27 @@ const AccountSelector = ({ route }: AccountSelectorProps) => {
...
 
+  // Handler for multichain account selection
+  const _onSelectMultichainAccount = useCallback(
+    (accountGroup: AccountGroupObject) => {
+      Engine.context.AccountTreeController.setSelectedAccountGroup(
+        accountGroup.id,
+      );
+      sheetRef.current?.onCloseBottomSheet();
+
+      // Track Event: "Switched Account" - may need to adjust this
```
david0xd: Can we resolve this concern on this PR, or put it under `TODO:`? This way it looks like a comment that no one would ever look into.

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -13,18 +18,30 @@ import { AccountWalletObject } from '@metamask/account-tree-controller';
...
 
+/**
+ * Get validated AccountTreeController state with feature flag and wallets checks
+ * @param state - Root redux state
+ * @returns Object with validation status and account tree state
+ */
+export const selectValidatedAccountTreeState = createDeepEqualSelector(
+  [selectAccountTreeControllerState, selectMultichainAccountsState1Enabled],
+  (accountTreeState, multichainAccountsState1Enabled) => ({
+    isValid:
```
david0xd: Why is this validation required? It looks confusing TBH. Wouldn't the account tree be always valid and available? 🤔

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -100,3 +110,65 @@ export const selectWalletByAccount = createDeepEqualSelector(
...
+        title: wallet.metadata.name,
+        wallet,
+        // Return account groups instead of flattened accounts
+        data: Object.values(wallet.groups) as AccountGroupObject[],
+      }),
+    );
+  },
+);
+
+export const selectSelectedAccountGroup = createDeepEqualSelector(
```
david0xd: Maybe this function could have docstring.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const handleSelectAccount = useCallback(
+    (accountGroup: AccountGroupObject) => {
+      // Prevent multiple rapid calls for the same account
+      if (selectedAccountGroup?.id === accountGroup.id) return;
+      onSelectAccount?.(accountGroup);
+    },
+    [onSelectAccount, selectedAccountGroup?.id],
+  );
+
+  const renderItem: ListRenderItem<FlattenedMultichainAccountListItem> =
```
david0xd, line 101: You might consider creating subcomponent that will render only the wallets with multichain accounts, keeping this one as a wrapper. Given that there is a lot of this logic going on for the element construction already, and that there will be more when search, pinned, hidden, create, new menu and stuff are implemented.

It doesn't have to be decoupled in this PR, as it is fine at the moment, but something we should consider when we start adding more.

app/components/Views/AccountSelector/AccountSelector.tsx
```diff
@@ -119,6 +127,27 @@ const AccountSelector = ({ route }: AccountSelectorProps) => {
     [accounts?.length, onSelectAccount, trackEvent, createEventBuilder],
   );
 
+  // Handler for multichain account selection
```
david0xd: Comment seems to be redundant as the function name is already self-explainable.

app/selectors/multichainAccounts/accountTreeController.ts
PatrykLucka: One reason is feature flag, as long as we are using it, there is an option that account tree won't be populated. Other checks are there just to make sure that everything is well defined, but maybe you're right we don't have to check that and if anything will go wrong something will just break and it might be easier to find the error.

david0xd: I somehow believe that account tree should be defined in state and populated with whatever data required before a user can even think about opening the account list page/menu and start selecting the accounts. It should be always present, IMO. And the user prevented from reaching to these parts of the app if it's not. Account tree and its state managed by the controller is the primary integrity of the system and should always be "stronger" than anything else. If it should fail for any reason, it should happen outside of the surface area which is the UI and basic selectors reading the state 🤔 

//cc @ccharly

app/components/Views/AccountSelector/AccountSelector.tsx
PatrykLucka: Confirmed with Thomas that the event should not change ✅

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+          autoFocus={false}
+          style={styles.searchTextField}
+        />
+      </View>
+      <View style={styles.listContainer} testID={testID}>
+        <FlashList
+          ref={listRef}
+          data={flattenedData}
+          renderItem={renderItem}
+          estimatedItemSize={70}
```
owencraston: where did you get this value?

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+        <View
+          style={styles.emptyState}
+          testID={MULTICHAIN_ACCOUNT_SELECTOR_EMPTY_STATE_TESTID}
+        >
+          <Text
+            variant={TextVariant.BodyMD}
+            color={TextColor.Muted}
+            style={styles.emptyStateText}
+          >
+            No accounts found
```
owencraston: This should be translated.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+          value={searchText}
+          onChangeText={setSearchText}
+          placeholder="Search accounts..."
+          testID={MULTICHAIN_ACCOUNT_SELECTOR_SEARCH_INPUT_TESTID}
+          autoFocus={false}
+          style={styles.searchTextField}
+        />
+      </View>
+      <View style={styles.listContainer} testID={testID}>
+        <FlashList
```
owencraston, line 179: we should pass a [getItemType](https://shopify.github.io/flash-list/docs/usage/#getitemtype) function to FlashList

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+    [onSelectAccount, selectedAccountGroup?.id],
+  );
+
+  const renderItem: ListRenderItem<FlattenedMultichainAccountListItem> =
+    useCallback(
+      ({ item }) => {
+        switch (item.type) {
+          case 'header': {
+            return (
+              <View style={styles.sectionHeader}>
```
owencraston: It would be nice if we extracted these into components such as AccountListHeader, AccountListCell, and AccountListFooter. This way the main component logic would be easier to maintain and we could more easily memoize each section.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+    );
+  }
+
+  return (
+    <>
+      <View style={styles.searchContainer}>
+        <TextFieldSearch
+          value={searchText}
+          onChangeText={setSearchText}
+          placeholder="Search accounts..."
```
owencraston: Needs to be transalted.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,196 @@
...
+        />
+      </View>
+      <View style={styles.listContainer} testID={testID}>
+        <FlashList
+          ref={listRef}
+          data={flattenedData}
+          renderItem={renderItem}
+          estimatedItemSize={70}
+          showsVerticalScrollIndicator={false}
+          keyExtractor={(item, index) => {
```
owencraston: extract this prop out into a function so that react can memoize it.

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -13,18 +18,27 @@ import { AccountWalletObject } from '@metamask/account-tree-controller';
 export const selectAccountTreeControllerState = (state: RootState) =>
   state.engine.backgroundState.AccountTreeController;
 
+/**
+ * Get validated AccountTreeController state with feature flag and wallets checks
+ * @param state - Root redux state
+ * @returns Object with validation status and account tree state
+ */
+export const selectValidatedAccountTreeState = createDeepEqualSelector(
```
owencraston: We should not be using createDeepEqualSelector here. Instead we should use plain old createSelector since it does a reference comparison which is much faster than deep equals and since this is a large object, the deep excessive deep equal calls will be a burden on performance. If the underlying object reference stays the same (which it should unless someone adds an account or changes the selected account), this will be more performant than deep equals.

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -13,18 +18,27 @@ import { AccountWalletObject } from '@metamask/account-tree-controller';
...
+    isValid: multichainAccountsState1Enabled,
+    accountTreeState,
+  }),
+);
+
 /**
  * Get account sections from AccountTreeController
  * For now, this returns a simple structure until the controller is fully integrated
  */
 export const selectAccountSections = createDeepEqualSelector(
```
owencraston: use createSelector instead of deep quals selector.

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -100,3 +107,71 @@ export const selectWalletByAccount = createDeepEqualSelector(
...
 );
+/**
+ * Get all account groups from all wallets in the AccountTreeController
+ * Returns a flat array of all account groups across all wallets
+ */
+export const selectAccountGroups = createDeepEqualSelector(
+  [selectValidatedAccountTreeState],
+  ({ isValid, accountTreeState }): AccountGroupObject[] => {
+    if (!isValid) {
+      return [];
```
owencraston: whenever we return an empty array or object we should opt for something like this...

```
const EMPTY_ARR: ReadonlyArray<never> = Object.freeze([]);
const EMPTY_OBJ: Readonly<Record<string, never>> = Object.freeze({});
...
if (!isValid) {
      return EMPTY_ARR;
}

```

This keeps the object reference stable (and reduces memory) so that it does not trigger a re render in the view.

owencraston: this pattern can be applied to all the selectors here.

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -100,3 +107,71 @@ export const selectWalletByAccount = createDeepEqualSelector(
...
+  (
+    { isValid, accountTreeState },
+    selectedAccountId,
+  ): AccountGroupObject | null => {
+    if (!isValid || !selectedAccountId) {
+      return null;
+    }
+
+    // Search through all wallets and groups to find the one containing the selected account
+    for (const wallet of Object.values(accountTreeState.accountTree.wallets)) {
```
owencraston: this selector as well as the `selectWalletByAccount` are iterating over the entire state. Instead we could create a reverse maps once per wallets identity:

	•	accountId → walletId
	•	accountId → group

This would look something like...

```ts
type AccountToWalletMap = Readonly<Record<AccountId, AccountWalletId>>;
type AccountToGroupMap = Readonly<Record<AccountId, AccountGroupObject>>;

export const selectAccountToWalletMap = createSelector(
  [selectWalletsMap],
  (wallets): AccountToWalletMap => {
    if (!wallets) return EMPTY_OBJ as AccountToWalletMap;
    const map: Record<string, string> = Object.create(null);
    for (const w of Object.values(wallets)) {
      for (const g of Object.values(w.groups)) {
        for (const id of g.accounts) map[id] = w.id;
      }
    }
    return map;
  },
);

export const selectAccountToGroupMap = createSelector(
  [selectWalletsMap],
  (wallets): AccountToGroupMap => {
    if (!wallets) return EMPTY_OBJ as AccountToGroupMap;
    const map: Record<string, AccountGroupObject> = Object.create(null);
    for (const w of Object.values(wallets)) {
      for (const g of Object.values(w.groups)) {
        for (const id of g.accounts) map[id] = g as AccountGroupObject;
      }
    }
    return map;
  },
);


export const selectWalletByAccount = createSelector(
  [selectWalletsMap, selectAccountToWalletMap, (_: RootState, accountId: AccountId) => accountId],
  (wallets, idx, accountId) => (wallets ? wallets[idx[accountId]] ?? null : null),
);

export const selectSelectedAccountGroup = createSelector(
  [selectIsMultichainEnabled, selectAccountToGroupMap, selectSelectedInternalAccountId],
  (enabled, idx, selectedId) => (!enabled || !selectedId ? null : idx[selectedId] ?? null),
);
```

app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -44,14 +58,11 @@ export const selectAccountSections = createDeepEqualSelector(
...
-  (accountTreeState, multichainAccountsState1Enabled) =>
+export const selectWalletById = createSelector(
+  [selectValidatedAccountTreeState],
+  ({ isValid, accountTreeState }) =>
     (walletId: AccountWalletId): AccountWalletObject | null => {
-      if (
-        !multichainAccountsState1Enabled ||
-        !accountTreeState?.accountTree?.wallets
-      ) {
+      if (!isValid) {
```
owencraston: what is this `isValid` I am seeing all over this file?

app/selectors/multichainAccounts/accountTreeController.ts
owencraston: I was confused about this entire isValid thing but I understand it now. I honestly think this is getting too complex to manage and we will have to remove all of this once we ship. I think it would be easier/more performant to ignore the feature flags in the selectors always return data) and instead conditionally call these selectors based on state 1/2 in the views. This will also give us a better sense of our performance since the selectors will remain the same after we launch. cc @gantunesr what do you think?

gantunesr: agreed, we shouldn't be using the feature flag at the selector level, just for the navigation and/or UI if necessary.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
PatrykLucka: It was just quick shot, as docs says "Doesn't have to be very accurate", but, just checked most elements should be 64px, so we can change it to this number.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -0,0 +1,194 @@
...
+import {
+  MULTICHAIN_ACCOUNT_SELECTOR_LIST_TESTID,
+  MULTICHAIN_ACCOUNT_SELECTOR_SEARCH_INPUT_TESTID,
+  MULTICHAIN_ACCOUNT_SELECTOR_EMPTY_STATE_TESTID,
+} from './MultichainAccountSelectorList.constants';
+import { strings } from '../../../../../locales/i18n';
+
+const MultichainAccountSelectorList = ({
+  onSelectAccount,
+  selectedAccountGroup,
```
montelaidev, line 31: Should we make this `selectedAccountGroups`? The `EvmAccountSelectorList` and `CaipAccountSelectorList` accepts a list of selected addresses

PatrykLucka: Good question, there are actually couple of other things that we could make consistent with the other components, but I decided to cut the scope of this already big PR. This can be added in next steps whenever (if ever) we will be adding support for the views that are using multi select

owencraston (dismissed):

owencraston (dismissed):

PatrykLucka:
Adding the `No E2E Smoke Needed`  since test https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/140b0190-2b22-42dd-bc76-3986f34839d4 is flaky and blocking CI, check the following bitrise run to verify that the rest of the E2E tests are passing

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
montelaidev: We will need this for the dapp connection, but we can leave it for a followup pr.

owencraston (approved):

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.stories.tsx
```diff
@@ -17,7 +17,7 @@ const SAMPLE_ACCOUNT_GROUP = {
 } as AccountGroupObject;
 
 const MultichainAccountRowMeta = {
-  title: 'Component Library / MultichainAccounts',
+  title: 'Component Library / MultichainAccounts/ AccountCell',
```
darkwing, line 20: nit: Space after `MultichainAccounts`

darkwing (approved):

owencraston:
Also resolves: https://consensyssoftware.atlassian.net/browse/MUL-465

PR #17943, opened 2025-08-05
chore(snaps): Bump Snaps packages

GuillaumeRx:
## **Description**

This PR bumps the snaps packages to the latest release of the snap monorepo (v123)

Notable changes:

- Adds the `onActive` and `onInactive` lifecycle hooks
- Fixes various issues with `CronjobController`
  - Fixes double scheduling of events
  - Use alternative storage
- Bumps SES version
- Fixes various issues with `SnapController`
  - Ignore terminate calls to not executing Snaps
  - Properly stop Snaps when clearing state
- Emit event when destroying execution environment

## **Changelog**

CHANGELOG entry:null

## **Related issues**

Closes: https://github.com/MetaMask/metamask-mobile/issues/16973

## **Manual testing steps**

```gherkin
Feature: my feature name
```


yarn.lock
```diff
@@ -6180,6 +6178,36 @@
     ses "^1.13.1"
     validate-npm-package-name "^5.0.0"
 
+"@metamask/snaps-utils@^11.2.0", "@metamask/snaps-utils@^11.3.0":
```
FrederikBolding: Dedupe?

app/core/Engine/Engine.ts
```diff
@@ -1751,6 +1764,22 @@ export class Engine {
...
+        if (state === 'active') {
+          this.controllerMessenger.call('SnapController:setClientActive', true);
+        }
+
+        if (state === 'background') {
+          this.controllerMessenger.call(
+            'SnapController:setClientActive',
+            false,
+          );
+        }
```
FrederikBolding: Could we simplify this to something like this?
```suggestion
        this.controllerMessenger.call('SnapController:setClientActive', state === 'active');
```

app/core/Engine/Engine.ts
```diff
@@ -786,6 +797,8 @@ export class Engine {
           ' and origin: ',
           origin,
         );
+
+        return null;
```
FrederikBolding, line 803: What's the reasoning for adding this?

GuillaumeRx: This function was returning `undefined` which isn't possible through json-rpc. That would crash any snap calling it. (We return the result of the hook in the rpc method implementation)

FrederikBolding: Fair enough 👍

app/core/Engine/Engine.ts
GuillaumeRx: There is a couple of other state which I'm not sure we want to react to 🤔 I'll try and see.

FrederikBolding: Have you tested that this actually triggers in time for the Snap to receive the message?

app/core/redux/slices/cronjobControllerStorage/index.ts
```diff
@@ -0,0 +1,36 @@
...
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { RootState } from '../../../../reducers';
+
+export interface CronjobControllerStorageState {
+  cronjobControllerStorage?: CronjobControllerState;
+}
+
+export const initialState: CronjobControllerStorageState = {
+  cronjobControllerStorage: undefined,
+};
```
FrederikBolding: Wondering if we should change the names a bit here so we don't have `cronjobControllerStorage.cronjobControllerStorage`. WDYT?

GuillaumeRx: Can do !

app/core/Engine/Engine.ts
GuillaumeRx: Yes 😄 

https://github.com/user-attachments/assets/05ff8098-17a8-4b25-b1d1-5bf45ad7eec4

GuillaumeRx: Tried with your suggested implementation but we need to specify the state to trigger `onInactive` otherwise we trigger it multiple time

Screenshot 2025-08-06 at 11 51 06

FrederikBolding: Can we just return early in case of other states than those two?

app/core/redux/slices/cronjobController/index.ts
FrederikBolding, line 1: Now that we have a separate slice for the cronjob controller, should/can we remove the controller from the regular Redux store? Otherwise, I would presume we get double the amount of state updates?

app/core/CronjobControllerStorageManager/CronjobControllerStorageManager.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    jest.resetAllMocks();
+    ReduxService.store = {
+      dispatch: jest.fn(),
+      getState: jest.fn(() => ({
+        // Mock the initial state for CronjobController
+        cronjobController: { storage: { events: {} } },
+      })),
+      subscribe: jest.fn(),
+      replaceReducer: jest.fn(),
+    } as unknown as ReduxStore;
```
FrederikBolding: Is there not a utility to create a proper store?

app/core/redux/slices/cronjobController/index.ts
```diff
@@ -0,0 +1,35 @@
...
+      state.storage = action.payload;
+    },
+  },
+});
+
+const { actions, reducer } = slice;
+
+export default reducer;
+
+export const selectCronjobControllerState = (state: RootState) => state[name];
```
FrederikBolding, line 34: Should this select the actual state (`storage`) instead to simplify the storage manager?

app/core/redux/slices/cronjobController/index.ts
GuillaumeRx: Mmmmmh It's not persisted on the "regular" store anymore and on the extension side we kept it because we need to recover the state if the new storage has never been used

GuillaumeRx: cc. @Mrtenz what do you think ?

app/core/CronjobControllerStorageManager/CronjobControllerStorageManager.test.ts
GuillaumeRx: I found multiple instances of this in tests yesterday, will see if I can find a utility

app/core/redux/slices/cronjobController/index.ts
GuillaumeRx: Also we still call `this.update()` in the controller so I think we still need some sort of state there ?

FrederikBolding: Maybe we can ignore calls to update its state from now on? It might get too hacky though.

app/core/Engine/Engine.ts
```diff
@@ -1751,6 +1765,20 @@ export class Engine {
...
+      'change',
+      (state: AppStateStatus) => {
+        if (state !== 'active' && state !== 'inactive') {
+          return;
+        }
+
+        return this.controllerMessenger.call(
+          'SnapController:setClientActive',
+          state === 'active',
+        );
```
FrederikBolding: `inactive` only exists on iOS, for some reason I thought we were checking for `background`. We may need to rethink this a bit then? It also warrants testing on Android.

https://reactnative.dev/docs/appstate#app-states

GuillaumeRx: Sorry, this is an oversight from me, I rewrote this multiple times and used `background` before. Will update

app/core/CronjobControllerStorageManager/CronjobControllerStorageManager.test.ts
```diff
@@ -0,0 +1,56 @@
...
+            },
+          },
+        },
+      };
+
+      manager.set(cronjobControllerState);
+
+      expect(dispatchSpy).toHaveBeenCalledWith(
+        setCronjobControllerState(cronjobControllerState),
+      );
```
FrederikBolding, line 79: Can we check that the getter now works?

app/core/Engine/Engine.ts
```diff
@@ -1751,6 +1765,20 @@ export class Engine {
...
     );
+
+    this.appStateListener = AppState.addEventListener(
+      'change',
+      (state: AppStateStatus) => {
+        if (state !== 'active' && state !== 'background') {
+          return;
+        }
+
+        return this.controllerMessenger.call(
```
FrederikBolding, line 1786:
```suggestion
        // Notifies Snaps that the app may be in the background. 
        // This is best effort as we cannot guarantee the messages are received in time.
        return this.controllerMessenger.call(
```

FrederikBolding (dismissed):

app/core/Engine/Engine.ts
```diff
@@ -2208,6 +2240,10 @@ export class Engine {
 
   removeAllListeners() {
     this.controllerMessenger.clearSubscriptions();
+
+    ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps)
+    this.appStateListener.remove();
+    ///: END:ONLY_INCLUDE_IF
```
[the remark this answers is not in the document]
FrederikBolding: Maybe just add an optional chain here in case it doesn't exist @GuillaumeRx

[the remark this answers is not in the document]
GuillaumeRx: Yeah it's planned

FrederikBolding (dismissed):

FrederikBolding (approved):

GuillaumeRx:
Skipping E2Es since there's a constant failure on one test but is unrelated to these changes. All the other E2Es including the Snaps one passes.

PR #18047, opened 2025-08-06
fix: copied to clipboard alert not dismissable

weitingsun:
## **Description**
**_This bug happens when the alert state is being persisted and saved to storage_**: When an address copy happens the showAlert action sets isVisible: true in the alert state. If user closes the app at this point before alert isVisible changes to false, the state isVisible: true is being persisted. When app restarts, the alert will still be visible and cannot be dismissed. The solution here is adding alert to blacklist in persistConfig.ts

Note: On Android, even though we can see the alert on app open, the alert is dismissible unlike iOS

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/17994 

## **Manual testing steps**

```gherkin
1. Click on copy address icon 
2. Close the app immediately while the "public address copied to clipboard" is still in view
3. Re-open the app 
4. Observe "public address copied to clipboard" alert should not be visible and app should function normally
```

### **Before**
iOS before

### **After**
Scenario 1: Users who don't have this problem, and they won't encounter it if they close the app when the alert is visible
iOS after

Scenario 2: Users who are stuck because of this alert, the alert should disappear after they install the fixed version
Clear alert

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -187,7 +187,7 @@ android {
         applicationId "io.metamask"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionName "7.54.0"
+        versionName "7.54.99"
         versionCode 2203
         testBuildType System.getProperty('testBuildType', 'debug')
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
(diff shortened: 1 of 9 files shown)


Cal-L (dismissed):
LGTM

Cal-L (approved):
Lgtm

PR #18113, opened 2025-08-08
feat: as a user i can view the perp tutorial and first time user state

nickewansmith:
## **Description**

  This PR implements a first-time user experience for the Perps trading feature. When users access the Perps tab for the first time, they will see a tutorial carousel that guides them through the key concepts of perpetual trading instead of the default trading interface.
  
 We determine if the user is a first time user by calling the userRoles info route to see if there is a role and if it is `"missing"` or `undefined`. This means the user has never touched the perps provider and needs to fund their account.

  What is the reason for the change?
  New users need onboarding guidance when first encountering Perps trading to understand the feature and how to get started.

  What is the improvement/solution?
  - Added first-time user detection logic in PerpsController and HyperLiquidProvider
  - Implemented conditional rendering in PerpsTabView to show tutorial content for new users vs. trading interface for returning users
  - Created PerpsTutorialCarousel component with educational content
  - Added comprehensive test coverage for all new functionality

## **Changelog**

CHANGELOG entry: Added first-time user tutorial carousel for Perps trading to help new users understand perpetual contracts

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1292
Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1247

## **Manual testing steps**

```gherkin
 Feature: Perps first-time user tutorial
    Scenario: user sees tutorial on first Perps tab visit
      Given user has never used Perps trading before - userRole returning will be null or 'missing'
      When user navigates to the Perps tab
      Then user should see the first time user state and a link to tutorial carousel instead of the trading balance
    Scenario: user sees normal interface on subsequent visits
      Given user has completed the Perps tutorial or used Perps before - userRoles will not be null
      When user navigates to the Perps tab
      Then user should see the normal trading interface with balance and positions
```

### **Before**
https://github.com/user-attachments/assets/6b986b5b-cf1d-4143-9565-f88028447996

### **After**
https://github.com/user-attachments/assets/71578f64-b306-41e5-8123-b77b3fd65ac3


app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +1937,42 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+      if (!userRole?.role || userRole.role === 'missing') {
+        return true;
+      }
+
+      return false;
+    } catch (error) {
+      // If there's an error checking user status, default to true (don't show tutorial)
+      DevLogger.log('Error checking if first time user', { error });
+      return true;
+    }
```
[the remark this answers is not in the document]
nickewansmith: Left over from testing 🙇🏽‍♂️

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +1937,42 @@ export class HyperLiquidProvider implements IPerpsProvider {
 
     return `${baseUrl}/explorer`;
   }
+
+  /**
```
abretonc7s: should this be at the PerpsController level or the provider level? What happens when we manage multiple providers?

app/components/UI/Perps/controllers/types/index.ts
```diff
@@ -407,6 +411,7 @@ export interface IPerpsProvider {
   // Unified asset and route information
   getDepositRoutes(params?: GetSupportedPathsParams): AssetRoute[]; // Assets and their deposit routes
   getWithdrawalRoutes(params?: GetSupportedPathsParams): AssetRoute[]; // Assets and their withdrawal routes
+  getIsFirstTimeUser(params?: GetIsFirstTimeUserParams): Promise<boolean>; // Check if this is the first time user is trading
```
abretonc7s: I think we should handle it in controller.

abretonc7s (dismissed):

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: I implement this aware that we do not know what this "is first time user" question/answer looks like with multiple providers. What I do know, is that each provider should be able to tell us if a user has ever used their platform in some way or another. 

I can imagine this being in the controller only and would prefer it, but in that case there would need to be some kind of state I believe, to know that the user has never used the perps experience.

Seeing that we only have 1 provider, I think this approach is ok as it can change quickly if we end up keeping some kind of state for if a user is a first time user across all providers, which will remove complexity.

With no state kept, the answer with multiple providers would be if any of them have been used, which each provider should be able to answer sufficiently. 

I don't think having it this way is harmful at the moment, although I do understand your concern @abretonc7s . 

Thoughts @Matt561 @gambinish?

app/components/UI/Perps/controllers/types/index.ts
nickewansmith: I agree we _should_ handle it in the controller outside of the providers to avoid complexity as it is as simple as saving a true or false flag to state that represents all perps providers, but for now without any state this is prob ok, as we still need a way to figure this out now. Having it as a provider level method keeps things cleanly separate for now with 1 provider. It would be an easy refactor to move to the controller handling this on its own once we have some state that we can persist.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
gambinish: I would agree that this is okay right now, and we shouldn't over-engineer for the time being. Good to keep in mind though to minimize migration to other providers.

We may want different onboarding flows, we may not

gambinish (dismissed):
This looks good to me. I assume we are planning on adding some animations in. Any idea when those will land?

One small nuance, and this might be a personal opinion:

As a first time user, if I got through onboarding, and decide to not fund my PerpsAccount yet, I don't think that I should have to go through onboarding again.

The current behavior: If I didn't fund my perps account during onboarding, I am forced to go through onboarding again to fund my account. This could create friction for new users getting familiar with Perps.

Given that we aren't persisting state yet, I'm not sure if this is possible in the transient setup we have now. Ideally, a user should only ever have to go through onboarding once? What do you think about considering this in the next iteration?

On the other hand, new users may actually feel that they want to go through onboarding again, or at least learn more about what they are doing before funding their account.

We can bring this up during a synchronous chat perhaps.

nickewansmith:
@abretonc7s to avoid delay on the Pr after approval we will move forward. But, if you still feel strongly let's discuss more Monday and go from there 🙏

gambinish (approved):

nickewansmith:
This PR has collectively passed the e2e smoke across runs.

Issue #17836, opened 2025-08-01
[Bug]: Send max ETH fails on EVM networks


Montoya:
### Describe the bug

Attempting to send max ETH fails. It does not appear to be deducting gas correctly. See video below. 

### Expected behavior

Send max should still deduct gas correctly and the transaction should go through. 

### Screenshots/Recordings

https://github.com/user-attachments/assets/f35f3367-5c9d-487a-8960-2e6f23fd7213

### Steps to reproduce

1. Go to any network, such as Arbitrum. 
2. Attempt to send max ETH from one address to another
3. Observe that the tx fails

### Error messages or log output

```shell

```

### Detection stage

During release testing

### Version

7.53.0

### Build type

Other (please specify exactly where you obtained this build in "Additional Context" section)

### Device

iPhone 16

### Operating system

iOS

### Additional context

Build 2215

bschorchit:
Closing as merged: https://github.com/MetaMask/metamask-mobile/pull/18258

PR #18128, opened 2025-08-08
fix: Tweak send max calculation to use max possible actual gas fees

### LINKED ISSUE #17865: [Bug]: Send max on Base and Sepolia leads to an infinite spinner
### Describe the bug

Send max on Sepolia leads to you to an infinite spinner after you approve the confirmation. Disabling smart transactions doesn't affect this issue. 

Update: the same is also happening on Base

### Expected behavior

_No response_

### Screenshots/Recordings

https://github.com/user-attachments/assets/39fa6023-e893-4d3a-9eff-6c631069cec3

### Steps to reproduce

1. Start a send max on Sepolia
2. Notice the infinite spinner

```shell

```

### Detection stage

During release testing

### Version

7.53.0

### Build type

None

### Device

iphone 11

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

pedronfigueiredo:
## **Description**

The way we calculated the value to send all native asset funds of the wallet, subtracted the estimated gas fees to the balance. However, this may not always be sufficient given the gas fee market fluctuations. The estimated fee is not guaranteed to be enough for the transaction to be confirmed.

Where the estimated fee is calculated from the minimumFeePerGas * the gas limit (and the minimumFeePerGas is the minimum value between the maxFeePerGas and the sum of the estimatedBaseFee and maxPriorityFeePerGas), the max fee is calculated from the maxFeePerGas * the gas limit directly. This means that the maxFee is greater or equal than the estimated fee.

This PR implements the calculation of the max fee so we can use it to calculate the total to transfer, lowering the likelihood of failing transactions.

Something must have changed on the gas API last week, because this is a recent regression only now reported, and the client code hasn't changed. In fact, I tested that this regression is now happening on old client code, at least as old as [#15074](https://github.com/MetaMask/metamask-mobile/pull/15074).

[#17865](https://github.com/MetaMask/metamask-mobile/issues/17865) should not happen anymore with this fix, but a try catch was wrapped around `Engine.acceptPendingApproval` in any case, to prevent instances of a transaction failure error bubbling up to the navigation code again.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: [#17865](https://github.com/MetaMask/metamask-mobile/issues/17865) and [#17836](https://github.com/MetaMask/metamask-mobile/issues/17836)

## **Manual testing steps**

```gherkin
Feature: my feature name
```

### DIFF
--- a/app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
+++ b/app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
@@ -100,12 +100,16 @@ export const useFeeCalculations = (transactionMeta: TransactionMeta) => {
   );
 
   // Max fee
-  const maxFee = useMemo(() => multiplyHexes(
-      supportsEIP1559
-        ? (decimalToHex(maxFeePerGas) as Hex)
-        : (txParamsGasPrice as Hex),
-      gasLimitNoBuffer as Hex,
-    ), [supportsEIP1559, maxFeePerGas, txParamsGasPrice, gasLimitNoBuffer]);
+  const maxFee = useMemo(
+    () =>
+      multiplyHexes(
+        supportsEIP1559
+          ? (decimalToHex(maxFeePerGas) as Hex)
+          : (txParamsGasPrice as Hex),
+        gasLimitNoBuffer as Hex,
+      ),
+    [supportsEIP1559, maxFeePerGas, txParamsGasPrice, gasLimitNoBuffer],
+  );
 
   const {
     currentCurrencyFee: maxFeeFiat,
--- a/app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
+++ b/app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
@@ -17,7 +17,35 @@ import { useGasFeeEstimates } from './useGasFeeEstimates';
 
 const HEX_ZERO = '0x0';
 
-export const useFeeCalculations = (transactionMeta: TransactionMeta) => {
+export const useFeeCalculations = (
+  transactionMeta: TransactionMeta,
+): {
+  estimatedFeeFiat: string | null;
+  estimatedFeeNative: string | null;
+  preciseNativeFeeInHex: string | null;
+  calculateGasEstimate: ({
+    feePerGas,
+    gasPrice,
+    gas,
+    shouldUseEIP1559FeeLogic,
+    priorityFeePerGas,
+  }: {
+    feePerGas: string;
+    gasPrice: string;
+    gas: string;
+    shouldUseEIP1559FeeLogic: boolean;
+    priorityFeePerGas: string;
+  }) => {
+    currentCurrencyFee: string | null;
+    nativeCurrencyFee: string | null;
+    preciseNativeCurrencyFee: string | null;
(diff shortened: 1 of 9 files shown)


app/components/Views/confirmations/hooks/useApprovalRequest.ts
```diff
@@ -29,12 +30,15 @@ const useApprovalRequest = () => {
...
-        opts,
-      );
+      try {
+        await Engine.acceptPendingApproval(
+          approvalRequest.id,
+          { ...approvalRequest.requestData, ...(value || {}) },
+          opts,
+        );
+      } catch (error: Error | unknown) {
+        captureException(error as Error);
```
matthewwalsh0: Should this instead be the responsibility of the caller, such as `Footer` as it then has the opportunity to present or render errors in a standard way?

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
```diff
@@ -98,10 +99,29 @@ export const useFeeCalculations = (transactionMeta: TransactionMeta) => {
     ],
   );
 
+  // Max fee
+  const maxFee = useMemo(() => multiplyHexes(
```
matthewwalsh0: Minor, this logic is great, but there is a lot in this single hook.

Not needed now but I added a `useTransactionMaxGasCost` and I wondered if we'd benefit from additional more granular hooks in future such as `useEstimatedGasCost`.

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
```diff
@@ -98,10 +99,33 @@ export const useFeeCalculations = (transactionMeta: TransactionMeta) => {
...
   );
 
+  // Max fee
+  const maxFee = useMemo(
+    () =>
+      multiplyHexes(
+        supportsEIP1559
+          ? (decimalToHex(maxFeePerGas) as Hex)
+          : (txParamsGasPrice as Hex),
+        gasLimitNoBuffer as Hex,
```
OGPoyraz: I think we should use `txParams.gas` instead of `gasLimitNoBuffer` here. 

The reason that we use `gasLimitNoBuffer` in the calculations is showing the minimal values in UI. But in reality `txParams.gas` value could be spend. This will cause more dust to be left over but it will be more reliable then.

app/components/Views/confirmations/hooks/gas/useFeeCalculations.ts
```diff
@@ -98,10 +127,38 @@ export const useFeeCalculations = (transactionMeta: TransactionMeta) => {
     ],
   );
 
+  // Max fee
+  const maxFee = useMemo(
+    () =>
+      multiplyHexes(
+        supportsEIP1559
+          ? (decimalToHex(maxFeePerGas) as Hex)
```
OGPoyraz, line 135: Minor - we can use directly `transactionMeta.txParams.maxFeePerGas` instead of converting decimal `maxFeePerGas`

OGPoyraz (dismissed):

matthewwalsh0 (dismissed):

OGPoyraz (approved):

PR #18164, opened 2025-08-10
chore: Fix selector & key prop warnings on home page

owencraston:
## **Description**

Remove the following warnings from the login/home page:

```
index.ts:63 The result function returned its own inputs without modification. e.g
`createSelector([state => state.todos], todos => todos)`
This could lead to inefficient memoization and unnecessary re-renders.
Ensure transformation logic is in the result function, and extraction logic is in the input selectors. 
```

```
Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, tabLabel: ..., navigation: ...};
  <Tokens {...props} />
```

```
index.tsx:274 Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, tabLabel: ..., navigation: ...};
  <Connect(CollectibleContracts) {...props} />
```

-  explicitly apply the key prop in the Tokens list and the CollectibleContracts
- Remove "wrapper" selectors that simply return the the result of their input

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **Before**

Screenshot 2025-08-10 at 4 25 36 PM

Screenshot 2025-08-10 at 4 27 36 PM

Screenshot 2025-08-10 at 4 28 04 PM

Screenshot 2025-08-10 at 4 28 27 PM

Screenshot 2025-08-10 at 4 29 04 PM

Screenshot 2025-08-10 at 4 30 14 PM

### **After**
(There are still warnings but they are not related to selectors or keys. the remaining warnings seem to be bugs in some of the packages that we are using)

https://github.com/user-attachments/assets/d8c64a6b-5ed3-422b-97f8-d462dbda82ee

### DIFF
--- a/app/components/Views/Wallet/index.tsx
+++ b/app/components/Views/Wallet/index.tsx
@@ -277,11 +277,14 @@ const WalletTokensTabView = React.memo(
           renderTabBar={renderTabBar}
           onChangeTab={onChangeTab}
         >
-          <Tokens {...tokensTabProps} />
+          <Tokens {...tokensTabProps} key={tokensTabProps.key} />
           {isPerpsEnabled && <PerpsTabView {...perpsTabProps} />}
           {defiEnabled && <DeFiPositionsList {...defiPositionsTabProps} />}
           {collectiblesEnabled && (
-            <CollectibleContracts {...collectibleContractsTabProps} />
+            <CollectibleContracts
+              {...collectibleContractsTabProps}
+              key={collectibleContractsTabProps.key}
+            />
           )}
         </ScrollableTabView>
       </View>
--- a/app/components/Views/Wallet/index.tsx
+++ b/app/components/Views/Wallet/index.tsx
@@ -278,8 +278,15 @@ const WalletTokensTabView = React.memo(
           onChangeTab={onChangeTab}
         >
           <Tokens {...tokensTabProps} key={tokensTabProps.key} />
-          {isPerpsEnabled && <PerpsTabView {...perpsTabProps} />}
-          {defiEnabled && <DeFiPositionsList {...defiPositionsTabProps} />}
+          {isPerpsEnabled && (
+            <PerpsTabView {...perpsTabProps} key={perpsTabProps.key} />
+          )}
+          {defiEnabled && (
+            <DeFiPositionsList
+              {...defiPositionsTabProps}
+              key={defiPositionsTabProps.key}
+            />
+          )}
           {collectiblesEnabled && (
             <CollectibleContracts
               {...collectibleContractsTabProps}
--- a/app/selectors/multichain/multichain.ts
+++ b/app/selectors/multichain/multichain.ts
@@ -185,12 +185,14 @@ export function selectMultichainAssetsMetadata(state: RootState) {
 }
 
 function selectMultichainAssetsRatesState(state: RootState) {
-  return state.engine.backgroundState.MultichainAssetsRatesController;
+  return state.engine.backgroundState.MultichainAssetsRatesController
+    .conversionRates;
 }
 
 export const selectMultichainAssetsRates = createDeepEqualSelector(
   selectMultichainAssetsRatesState,
-  (multichainAssetRates) => multichainAssetRates.conversionRates,
+  (conversionRates) => conversionRates,
+  { devModeChecks: { identityFunctionCheck: 'never' } },
 );
 
 export function selectMultichainHistoricalPrices(state: RootState) {
--- a/app/selectors/multichain/multichain.ts
+++ b/app/selectors/multichain/multichain.ts
@@ -184,14 +184,13 @@ export function selectMultichainAssetsMetadata(state: RootState) {
   return state.engine.backgroundState.MultichainAssetsController.assetsMetadata;
 }
 
-export function selectMultichainAssetsRatesState(state: RootState) {
-  return state.engine.backgroundState.MultichainAssetsRatesController
-    .conversionRates;
+function selectMultichainAssetsRatesState(state: RootState) {
+  return state.engine.backgroundState.MultichainAssetsRatesController;
 }
 
 export const selectMultichainAssetsRates = createDeepEqualSelector(
   selectMultichainAssetsRatesState,
-  (conversionRates) => conversionRates,
+  (multichainAssetRates) => multichainAssetRates.conversionRates,
 );
 
 export function selectMultichainHistoricalPrices(state: RootState) {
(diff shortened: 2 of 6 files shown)


gantunesr (dismissed):
Looks good, did not manually test it

app/selectors/transactionController.ts
```diff
@@ -6,23 +6,18 @@ import { selectPendingSmartTransactionsBySender } from './smartTransactionsContr
...
-
 const selectTransactionBatchesStrict = createSelector(
   selectTransactionControllerState,
   (transactionControllerState) => transactionControllerState.transactionBatches,
 );
 
 export const selectTransactions = createDeepEqualSelector(
-  selectTransactionsStrict,
-  (transactions) => transactions,
+  selectTransactionControllerState,
```
OGPoyraz: The reason of having `selectTransactionsStrict` selector here is to prevent shallow equality check on whole state in `createDeepEqualSelector` instead of just `transactions`. So this will be less performant if we do this change. 

And as `selectTransactions` used in the selectors below - it will also cause the performance degradation there too.

app/components/Views/Wallet/index.tsx
```diff
@@ -267,11 +267,14 @@ const WalletTokensTabView = React.memo(
           renderTabBar={renderTabBar}
           onChangeTab={onChangeTab}
         >
-          <Tokens {...tokensTabProps} />
+          <Tokens {...tokensTabProps} key={tokensTabProps.key} />
```
Prithpal-Sooriya, line 270: hmm, why is the key necessary here if we are already spreading all the `tokensTabProps` to this component?

app/components/Views/Wallet/index.tsx
```diff
@@ -267,11 +267,14 @@ const WalletTokensTabView = React.memo(
...
           {isPerpsEnabled && <PerpsTabView {...perpsTabProps} />}
           {defiEnabled && <DeFiPositionsList {...defiPositionsTabProps} />}
           {collectiblesEnabled && (
-            <CollectibleContracts {...collectibleContractsTabProps} />
+            <CollectibleContracts
+              {...collectibleContractsTabProps}
+              key={collectibleContractsTabProps.key}
```
Prithpal-Sooriya, line 276: likewise, is there a reason we need to explicitly pass in the key if it is spread inside this component through `{...collectibleContractsTabProps}`?

app/selectors/multichain/multichain.ts
```diff
@@ -183,14 +183,13 @@ export function selectMultichainAssetsMetadata(state: RootState) {
...
-    .conversionRates;
+function selectMultichainAssetsRatesState(state: RootState) {
+  return state.engine.backgroundState.MultichainAssetsRatesController;
 }
 
 export const selectMultichainAssetsRates = createDeepEqualSelector(
   selectMultichainAssetsRatesState,
-  (conversionRates) => conversionRates,
+  (multichainAssetRates) => multichainAssetRates.conversionRates,
 );
```
Prithpal-Sooriya: Deep eq is more expensive that the previous version.

- Before: the input selector result was previously only the conversion rate
    - Meaning the deep-equal input result validation was only on the returned conversion rates.
- After: now the input selector result takes the **entire** `MultichainAssetsRatesController` state.
    - The deep-equal input result validate now needs to check the entire `MultichainAssetsRatesController` state.

Prithpal-Sooriya: I get that we want to avoid the identity function output selector warning, but this new approach will incur more validation costs.

Prithpal-Sooriya: Does this really need to be a create-selector? I guess yes since the `deep-equal` check was done probably because this underlying state is volatile.

app/selectors/transactionController.ts
Prithpal-Sooriya: Agreed, similar to the above discussion.
https://github.com/MetaMask/metamask-mobile/pull/18164/files#r2266099607

app/components/Views/Wallet/index.tsx
Prithpal-Sooriya, line 280: Nice I saw the warning in the description:
```
Warning: A props object containing a "key" prop is being spread into JSX:
```

Can we do this for the other lists below? (perps, defi)

app/selectors/multichain/multichain.ts
Prithpal-Sooriya: I feel like for places that use the `deepEqualSelector`, there are cases where we intentionally want an identity output selector.

Thoughts on modifying the `createDeepEqualSelector` function to include better JSDocs for usage and suppress the identity warning?

```ts

/**
 * Creates a selector with deep equality checking for input comparisons.
 * 
 * Uses deep equality instead of reference equality to prevent unnecessary 
 * recalculations when input selectors return identical objects with different references.
 * 
 * @example
 * ```typescript
 * const selectUserPreferences = createDeepEqualSelector(
 *   selectUserPreferencesState,
 *   (preferences) => preferences
 * );
 * 
 * const selectFilteredItems = createDeepEqualSelector(
 *   [selectItems, selectComplexFilterConfig],
 *   (items, filterConfig) => items.filter(item => matchesFilter(item, filterConfig))
 * );
 * ```
 * 
 * **When to use:**
 * - Input selectors return complex objects that should be compared by value
 * - Result function returns identity/passthrough (acceptable here)
 * 
 * **Avoid when:**
 * - Working with primitives or when deep equality checks would be expensive
 */
export const createDeepEqualSelector = createSelectorCreator({
  memoize: lruMemoize,
  memoizeOptions: {
    equalityCheck: deepEqual,
  },
  devModeChecks: {
    identityFunctionCheck: 'never',
  },
});

```

app/components/Views/Wallet/index.tsx
owencraston: Done ✅

app/selectors/multichain/multichain.ts
owencraston: Great points, I worry that disabling this globally could lead to some misuse from people that do not understand the selectors as well as you guys. Instead I reverted my changes on your selectors and added the identityFunctionCheck: 'never' individually to each selector.

Prithpal-Sooriya: Ah good idea! If we find that we are doing this in multiple places then we can re-evaluate the global ver.

Prithpal-Sooriya (approved):

app/selectors/multichain/multichain.ts
owencraston: Agreed. I still think having the warning will make devs pause and think twice to see if the use case is valid.

app/selectors/transactionController.ts
owencraston: Good point, reverted my changes and disabled the warning on that specific selector.

gantunesr (approved):

PR #18165, opened 2025-08-10
feat: as a user i can subscribe to push notifications

nickewansmith:
## **Description**

Requirement for this to work without mocking is that the new centralized Notifications feature is enabled and we have perps notifications flowing through the notifications service controller. Here is a POC starter [PR](https://github.com/MetaMask/core/pull/6279), only adding the fields/methods to NotificationsServiceController that are missing that we use in this branch. I have made it so that this mobile branch is safe to merge so that we can get the code in even though the controller will need more work than just adding these fields/methods to allow the switch to go off/on (we could do that though for testing if @MetaMask/notifications team is in agreement).

We want to show a notifications tooltip to the user if they have completed their first order and have not enabled perps notifications. After their first order has been marked complete, this notification popup will no longer be triggered.
Adds a persistent state  for first order in PerpsController and uses this along with notifications feature enabled flag to possibly trigger the tooltip to ask the user if they want to turn on notifications for perps. After the first order success is marked, the user will no longer see this tooltip suggestion.

## **Changelog**

CHANGELOG entry: Added a new perps notifications tooltip for users who successfully complete an order for the first time

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1290
Relates to: https://github.com/MetaMask/core/pull/6279

## **Manual testing steps**

- you may need to set notifications feature flag to on manually in the code
- to test the toggle functionality you may need to switch the toggle key to something that actually exists as a toggle

```gherkin
Feature: perps notifications tooltip on first time order

  Scenario: user clicks on Perpetuals in the wallet menu
    Given user makes a trade order
    When user is making their first trade order
    Then a perps notification tooltip should pop up suggesting that they turn on notifications and then user has made first order variable should be set to true
    
  Scenario: user clicks on Perpetuals in the wallet menu
    Given user makes a trade order
    When user is making a trade order but it is not their first
    Then a perps notification tooltip should NOT pop up suggesting that they turn on notifications
    
  Scenario: user clicks on Perpetuals in the wallet menu
    Given user makes a trade order
    When user is making their first trade order but notifications feature is off
    Then a perps notification tooltip should NOT pop up suggesting that they turn on notifications and the user should be redirected to the appropriate page after order success which is existing functionality
```

### **After**

https://www.loom.com/share/34838c69f74a4befabf0cd0bceed6de6?sid=a1efcff5-6ad4-4ef9-8b96-5033122aaf0c


app/actions/notification/helpers/index.ts
```diff
@@ -62,6 +62,21 @@ export const toggleFeatureAnnouncements = async (
...
+/**
+ * Perps Notifications Switch
+ * - Enables/Disables Perps Notifications
+ * @param perpsNotificationsEnabled boolean to toggle on/off
+ */
+export const togglePerpsNotifications = async (
+  perpsNotificationsEnabled: boolean,
+) => {
+  assertIsFeatureEnabled();
+  // @ts-expect-error - setPerpsNotificationsEnabled not yet implemented
```
abretonc7s, line 74: I am not sure yet what is best for now here, if we want to keep this in

Prithpal-Sooriya: Open to either waiting until the CORE changes are merged, or we can keep this but add some guards so this doesn't break at runtime.

```ts
export const togglePerpsNotifications = async (perpsNotificationsEnabled: boolean) => {
  assertIsFeatureEnabled();
  
  // Temporary until perps notifications is added into controller
  if (
    'setPerpsNotificationsEnabled' in
    Engine.context.NotificationServicesController
  ) {
    // @ts-expect-error - setPerpsNotificationsEnabled not yet implemented
    await Engine.context.NotificationServicesController.setPerpsNotificationsEnabled(
      perpsNotificationsEnabled,
    );
  }
};
```

app/selectors/notifications/index.tsx
```diff
@@ -64,6 +64,12 @@ export const selectIsFeatureAnnouncementsEnabled = createSelector(
   (notificationServicesControllerState: NotificationServicesState) =>
     notificationServicesControllerState.isFeatureAnnouncementsEnabled,
 );
+export const selectIsPerpsNotificationsEnabled = createSelector(
+  selectNotificationServicesControllerState,
+  (notificationServicesControllerState: NotificationServicesState) =>
+    // @ts-expect-error - isPerpsNotificationsEnabled not yet implemented
+    notificationServicesControllerState.isPerpsNotificationsEnabled,
```
Prithpal-Sooriya, line 71: TY for this comment, I don't think we need any fallbacks for this, it would most likely return undefined as it is not currently in state.

app/components/UI/Perps/hooks/usePerpsSelector.ts
```diff
@@ -8,7 +8,7 @@ import type { PerpsControllerState } from '../controllers/PerpsController';
  * @returns Selected data from PerpsController state
  */
 export function usePerpsSelector<T>(
-  selector: (state: PerpsControllerState | undefined) => T,
+  selector: (state: PerpsControllerState) => T,
```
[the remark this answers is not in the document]
nickewansmith: I think this is ok.

app/util/notifications/hooks/useSwitchNotifications.ts
```diff
@@ -76,6 +78,31 @@ export function useFeatureAnnouncementToggle() {
   };
 }
 
+export function usePerpsNotificationToggle() {
```
Prithpal-Sooriya, line 81: Very clean hook to manage toggling for perps.
For this PR I don't think any extra changes are necessary. Once our backend APIs are setup I'll do a little cleanup on this hook.

1. Perps will only support push notifications, so we don't need to call `debounce(listNotifications)();` to refresh in-app notifs.

2. Perps toggle might be unique since we want to ensure that it is tied to push notifications. E.g. when toggling on, we also want to get users to toggle push notifications.
    - We can add this in a diff PR, either extending this hook or as a separate hook.

app/components/Views/Settings/NotificationsSettings/index.tsx
```diff
@@ -81,6 +82,7 @@ const NotificationsSettings = ({ navigation, route }: Props) => {
             styles={styles}
           />
           <FeatureAnnouncementToggle />
+          <PerpsNotificationToggle />
```
Prithpal-Sooriya: Can we add a condition/FF to ensure that this toggle is not visible if perp notifications are not yet supported?

nickewansmith: Added the perps flag and a perps notifications feature flag constant

Prithpal-Sooriya (approved):

app/components/Views/Settings/NotificationsSettings/index.tsx
Prithpal-Sooriya: TY

vinnyhoward (approved):
LGTM. Very clean.

app/components/UI/Perps/hooks/usePerpsSelector.ts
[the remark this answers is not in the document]
abretonc7s: any reasons for changing it?

app/components/Views/Settings/NotificationsSettings/PerpsNotificationToggle.tsx
```diff
@@ -0,0 +1,38 @@
...
+export function PerpsNotificationToggle() {
+  const { data: isEnabled, switchPerpsNotifications } =
+    usePerpsNotificationToggle();
+  const { trackEvent, createEventBuilder } = useMetrics();
+
+  const togglePerpsNotificationsEnabled = useCallback(async () => {
+    const newValue = !isEnabled;
+    await switchPerpsNotifications(newValue);
+
+    trackEvent(
```
abretonc7s, line 18: we should use the perps specific hooks to track 
```
// For user interactions
track(MetaMetricsEvents.PERPS_ORDER_SUBMITTED, properties);
```

abretonc7s (changes requested):

app/components/UI/Perps/hooks/usePerpsSelector.ts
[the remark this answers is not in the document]
nickewansmith: During merge conflict resolution, I had two of the same hook, I chose the one that more matches the rest of the code base in terms of the assumption that the state should not be undefined.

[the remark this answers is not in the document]
nickewansmith: During merge conflict resolution, I had two of the same hook, I chose the one that more matches the rest of the code base in terms of the assumption that the state should not be undefined.

app/actions/notification/helpers/index.ts
abretonc7s: @nickewansmith should we add @Prithpal-Sooriya comment?

app/components/Views/Settings/NotificationsSettings/PerpsNotificationToggle.tsx
nickewansmith: Maybe this small change can be added in a separate outstanding pr to avoid knocking of stamps here if that's ok.

abretonc7s (approved):

PR #18187, opened 2025-08-11
fix: Storage of File with Sensitive Data Under Web Root

kreeksec:
https://github.com/MetaMask/metamask-mobile/blob/09f00032972ec2b9539e87f939b82c4def52c63f/e2e/create-static-server.js#L9-L13
Libraries like express provide easy methods for serving entire directories of static files from a web server. However, using these can sometimes lead to accidental information exposure. If for example the node_modules folder is served, then an attacker can access the _where field from a package.json file, which gives access to the absolute path of the file.

fix the problem, we should avoid serving the entire `node_modules` directory. Instead, only expose specific subdirectories within `node_modules` that are required by the application (e.g., only the `dist` folders of certain packages like `jquery` or `bootstrap`). This can be achieved by checking the URL path and only allowing access to a whitelist of safe subfolders. In this file, update the logic in the `/node_modules/` handler to only serve files from explicitly allowed packages, and reject or return a 404 for any other requests.

Specifically, in `e2e/create-static-server.js`, replace the block that serves all of `node_modules` with logic that only serves specific subfolders (e.g., `jquery/dist`, `bootstrap/dist`). You may need to add a mapping of allowed packages to their public paths, and update the handler to check the requested path against this mapping.

### DIFF
--- a/e2e/create-static-server.js
+++ b/e2e/create-static-server.js
@@ -6,11 +6,29 @@ import serveHandler from 'serve-handler';
 const createStaticServer = function (rootDirectory) {
   return http.createServer((request, response) => {
     if (request.url.startsWith('/node_modules/')) {
-      request.url = request.url.substr(14);
-      return serveHandler(request, response, {
-        directoryListing: false,
-        public: path.resolve('./node_modules'),
-      });
+      // Only allow serving specific packages from node_modules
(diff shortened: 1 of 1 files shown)


kreeksec:

PR #18276, opened 2025-08-12
feat: perps builder and referral fee collection

nickewansmith:
### What is the reason for the change?

We would like to collect fees for perps trading activity on our platform

### What is the improvement/solution?

This commit implements a fee collection system for MetaMask's perpetuals trading feature, adding both builder fees and referral code functionality to generate revenue from trading activities.

Key Changes:
1. Fee Collection Infrastructure
- Builder Fee System: Added a 10 basis point (0.1%) builder fee that gets automatically collected on all trades
- Referral Code System: Implemented automatic referral code assignment to ensure MetaMask receives referral rewards
- Network-Specific Configuration: Separate configurations for testnet and mainnet environments
2. Core Implementation
- HyperLiquidProvider Enhancements:
- Added ensureBuilderFeeApproval() method to handle builder fee authorization
- Added ensureReferralSet() method to automatically set MetaMask referral codes
- Integrated fee collection into all order placement flows (market orders, limit orders, TP/SL orders)
- Added helper methods for fee calculations and approval checks
3. Configuration Updates
- BUILDER_FEE_CONFIG: Defines builder addresses and fee rates for both networks
- REFERRAL_CONFIG: Sets up referral codes ("METAMASK" for mainnet, "MSO" for testnet)
- Network-specific builder addresses: Separate wallets for testnet and production environments
4. User Experience Improvements
- Network-specific Tutorial Tracking: Modified isFirstTimeUser to track tutorial completion separately for testnet and mainnet
- Automatic Fee Handling: Users don't need to manually approve fees - the system handles this automatically
- Error Handling: Comprehensive error handling and logging for fee-related operations
5. Testing & Quality Assurance
- Comprehensive Test Coverage: Added extensive unit tests for the new fee collection functionality
- Integration Testing: Tests cover builder fee approval, referral code setting, and error scenarios
- Updated Test Snapshots: Updated existing test snapshots to reflect new state structure
6. Technical Details
- Fee Calculation: Builder fees are converted from basis points to tenths of basis points for HyperLiquid API compatibility
- Approval Flow: Automatic approval of builder fees before order placement
- Referral Validation: Checks if referral codes are ready and properly configured before use
- State Management: Updated Redux state to support network-specific tutorial tracking

## **Changelog**

CHANGELOG entry: Adds perps builder and referral fee collection for trading activity

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1253

## **Manual testing steps**

- It may be easier to test on mainnet, but also works on testnet
- Edit the config file depending on which netowrk you are choosing and which builder wallet, staking test wallet is already set up as a builder so that is a good one to use.

```
export const BUILDER_FEE_CONFIG = {
  // Test wallet address for builder fees, currently staking test wallet
  testnetBuilder: '0x316BDE155acd07609872a56Bc32CcfB0B13201fA' as `0x${string}`,
  // Production builder wallet
  mainnetBuilder: '0x316BDE155acd07609872a56Bc32CcfB0B13201fA' as `0x${string}`,
  // Fee in basis points (10 bp = 0.1% - this is the maximum allowed)
  feeBps: 10, // 10 basis points (0.1%)
};

// Referral code configuration
export const REFERRAL_CONFIG = {
  // Production referral code
  mainnetCode: 'MSO',
  // Development/testnet referral code
  testnetCode: 'MSO',
};
```

- now using a new account that has usdc on arbitrum and enough ArbETH to pay for gas, fund the perps trading account
- once you make a trade, you can check the dashboard to see the effect of the trade on the fees reported
- to check the fees in detail you can use this script to dl the json, replacing the user with the new user you are testing with

```
curl -s https://api.hyperliquid.xyz/info \                                                                         ✔  system   at 13:49:58  
  -H 'Content-Type: application/json' \
  -d '{"type":"userFillsByTime","user":"0x5A0e4c98Cb4086926edeE390d251ac60B5c27596","startTime":0,"endTime":'$end_ms'}' \
| tee hl_fills-0x5A0e4c98Cb4086926edeE390d251ac60B5c27596.json
```
- referrals are not counted for referrers that own the referrer code, so to test referrals you would need to be on a new or different account and not the staking test account

```gherkin
Feature: perps builder and referral fee collection

  Scenario: user is a new user to the perps feature
    Given the user has funded their account with usdc
    When user places an order
    Then the builder fee we charge will be approved on their behalf
    Then the referrer will be set to MM code if they do not yet have a referrer
    Then the order will execute and builder fees and referral fees will be accumulated
```

### **Before**

n/a

### **After**

https://www.loom.com/share/1508edb339b54aeba2eacc3fe8eb4cf8?sid=a44006ac-b80b-4e5b-83e7-683ecf1a5d89

[hl_fills-0x5A0e4c98Cb4086926edeE390d251ac60B5c27596.csv](https://github.com/user-attachments/files/21743035/hl_fills-0x5A0e4c98Cb4086926edeE390d251ac60B5c27596.csv)

Screenshot 2025-08-12 at 3 06 33 PM


app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith, line 1: The bulk of the changes in this PR are in this file. Basically, for builder fees, on every call to the exchange client that accepts a builder param we send that builder param and our fee with it. For Referral fees on every call to the exchange client that accepts a builder param, we are checking if the user has a referrer attached to their account, and if they don't we will set ourselves as their referrer. This is a one time thing, and from then on we are their referrer. This also means that if the user already has a referrer we cannot change it.

In terms of orders, when creating orders, updating orders and closing orders we attach a builder fee or may possibly attempt to set a referrer if not set. 

Note that updating orders does not necessarily incur a fee immediately, but instead updates the fee for when the order is filled whenever that is, be it immediately or otherwise.

Note that we approve a maximum fee of 10bps .1% for perps trading on behalf of the user with their private key, but we could send a lower fee amount per order/update/close if we wanted to under certain conditions i.e. MetaMask points give a discount. Currently for each order case, we send the maximum fee each time.

Ultimately the code ensures that we never miss a referral or builder fee as long as the builder account is funded with > 100USDC at all times and the builder account has done > 10k volume. We should ensure that these things are true before launching in production.

app/components/UI/Perps/constants/hyperLiquidConfig.ts
```diff
@@ -102,6 +102,24 @@ export const FEE_RATES: FeeRatesConfig = {
...
 };
 
+// Builder fee configuration
+export const BUILDER_FEE_CONFIG = {
+  // Test wallet address for builder fees, currently staking test wallet
+  testnetBuilder: '0x316BDE155acd07609872a56Bc32CcfB0B13201fA' as `0x${string}`,
+  // Production builder wallet
+  mainnetBuilder: '0xe95a5e31904e005066614247d309e00d8ad753aa' as `0x${string}`,
+  // Fee in basis points (10 bp = 0.1% - this is the maximum allowed)
+  feeBps: 10, // 10 basis points (0.1%)
```
Matt561: We have the following block inside of `perpsConfig.ts`

```ts
export const METAMASK_FEE_CONFIG = {
  // Trading fees (as decimal, e.g., 0.01 = 1%)
  TRADING_FEE_RATE: 0, // 0% currently, will be fetched from API later

  // Deposit/withdrawal fees
  DEPOSIT_FEE: 0, // $0 currently
  WITHDRAWAL_FEE: 0, // $0 currently

  // Future: These will be fetched from API based on:
  // - User tier/volume
  // - Promotional campaigns
  // - Protocol-specific agreements
  // - MetaMask points/rewards integration
} as const;
```

Which one should serve as the source of truth? This is the value we'll want to display in the fee tooltip when placing an order.

app/components/UI/Perps/constants/hyperLiquidConfig.ts
```diff
@@ -102,6 +102,24 @@ export const FEE_RATES: FeeRatesConfig = {
...
+};
+
+// Referral code configuration
+export const REFERRAL_CONFIG = {
+  // Production referral code METAMASK
+  mainnetCode: 'METAMASK',
```
Matt561: Apologies if I'm overthinking here but have we verified that these referral codes aren't in use yet? I'm not sure where the connection is between these codes and HL's backend to ensure we're the ones being paid.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
Matt561: Documentation: I almost feel like these things could be broken out into a Notion doc as a TL;DR knowledge base on Hyperliquid for future engineers or us if we lose context.

Points like:

>This is a one time thing, and from then on we are their referrer. This also means that if the user already has a referrer we cannot change it.

>Note that updating orders does not necessarily incur a fee immediately, but instead updates the fee for when the order is filled whenever that is, be it immediately or otherwise.

>we never miss a referral or builder fee as long as the builder account is funded with > 100USDC at all times and the builder account has done > 10k volume.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -177,6 +179,106 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+  }
+
+  /**
+   * Convert basis points to tenths of basis points (for HyperLiquid API)
+   * @param bps - Fee in basis points (1 bp = 0.01%)
+   * @returns Tenths of basis points (10 tenths = 1 bp)
+   */
+  private bpsToTenthsBps(bps: number): number {
+    return bps * 10;
+  }
```
Matt561: Could these be broken out into a general utils file since they're not Hyperliquid specific?

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -177,6 +179,106 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+
+  /**
+   * Check current builder fee approval for the user
+   * @param builder - Builder address to check approval for
+   * @returns Current max fee rate or null if not approved
+   */
+  private async checkBuilderFeeApproval(
+    builder: `0x${string}`,
```
Matt561: Would it be worth typing the builder arg more specifically so that only MM addresses are allowed in here or else we get TS errors? Possibly overkill but thought I'd mention.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -1489,17 +1495,28 @@ export class PerpsController extends BaseController<
...
+    const currentNetwork = this.state.isTestnet ? 'testnet' : 'mainnet';
+
     DevLogger.log('PerpsController: Marking tutorial as completed', {
       timestamp: new Date().toISOString(),
+      network: currentNetwork,
     });
 
     this.update((state) => {
-      state.isFirstTimeUser = false;
+      state.isFirstTimeUser[currentNetwork] = false;
```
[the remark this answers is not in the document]
nickewansmith: I'm not sure if this applies when in dev, but added one anyway.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+  private async ensureReferralSet(): Promise<void> {
+    const errorMessage = 'Error ensuring referral code is set';
+    try {
+      const isTestnet = this.clientService.isTestnetMode();
+      const network = isTestnet ? 'testnet' : 'mainnet';
+      const expectedReferralCode = this.getReferralCode(isTestnet);
+      const referrerAddress = this.getBuilderAddress(isTestnet);
+      const userAddress = await this.walletService.getUserAddressWithDefault();
+
+      if (userAddress.toLowerCase() === referrerAddress.toLowerCase()) {
```
Matt561, line 2067: I don't understand this check. Is this purely to avoid attaching the referral code if the MM wallet places an order? Is this a HL constraint where you can't refer yourself?

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+      if (!hasReferral) {
+        DevLogger.log('No referral set - setting MetaMask as referrer', {
+          network,
+          referralCode: expectedReferralCode,
+        });
+        await this.setReferralCode();
+      }
+    } catch (error) {
+      DevLogger.log(errorMessage, error);
+      console.error(errorMessage, error);
```
Matt561: Minor: Do we need both `DevLogger` and `console.error`?

app/store/migrations/094.test.ts
```diff
@@ -0,0 +1,320 @@
...
+  };
+}
+
+jest.mock('@sentry/react-native', () => ({
+  captureException: jest.fn(),
+}));
+
+jest.mock('./util', () => ({
+  ensureValidState: jest.fn(),
+}));
```
Matt561: Not-blocking in my opinion but with our whole team leaning on code generation we should make sure we don't over-mock either. If this is a simple util we probably don't need to mock it.

app/store/migrations/094.ts
```diff
@@ -0,0 +1,65 @@
...
+      if (typeof isFirstTimeUser === 'boolean') {
+        // Convert boolean to object format
+        // If user had completed tutorial (false), mark both networks as completed
+        // If user was first-time (true), mark both networks as first-time
+        perpsController.isFirstTimeUser = {
+          testnet: isFirstTimeUser,
+          mainnet: isFirstTimeUser,
+        };
+
+        // Migration completed successfully
```
Matt561: Minor: We can remove this

app/store/migrations/094.ts
```diff
@@ -0,0 +1,65 @@
...
+      isObject(state.engine.backgroundState) &&
+      hasProperty(state.engine.backgroundState, 'PerpsController') &&
+      isObject(state.engine.backgroundState.PerpsController) &&
+      hasProperty(
+        state.engine.backgroundState.PerpsController,
+        'isFirstTimeUser',
+      )
+    ) {
+      const perpsController = state.engine.backgroundState.PerpsController;
+      const isFirstTimeUser = perpsController.isFirstTimeUser;
```
Matt561: Is there a chance that we'd reach this line and `isFirstTimeUser` isn't set? Just wondering if we should default if it's undefined.

app/components/UI/Perps/constants/hyperLiquidConfig.ts
nickewansmith: The MetaMask fees are different from the builder and referral fees and are prob more general to perps, the builder/referral fees are specific to HL

app/components/UI/Perps/constants/hyperLiquidConfig.ts
nickewansmith: For the test staking wallet, the code is MSO. This code can only be set in the ui currently. In the PRD the code chosen for prod was METAMASK, if that changes then we can update it before launch. I think we will know for sure by eow per MSO. HL has an api for getting the info that is on the builder account. If we are not matching the order will go through but there will be an error in the logs about the mismatch.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: Yes you cannot refer yourself

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: DevLogger uses console.debug so if we want an actual sentry error possibly

app/store/migrations/094.ts
nickewansmith: It won't hurt to add this, will do

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: They aren't specific to HL but I left them here as prob nothing else would use them unless using this notation specifically, which is possible. I can break it out though.

app/components/UI/Perps/constants/hyperLiquidConfig.ts
abretonc7s: Normally the metamask fee would be 10bps but can be discounted depending on rewards / points integration

app/components/UI/Perps/constants/hyperLiquidConfig.ts
```diff
@@ -102,6 +102,24 @@ export const FEE_RATES: FeeRatesConfig = {
   maker: 0.00015, // 0.015% - Limit orders that add liquidity
 };
 
+// Builder fee configuration
+export const BUILDER_FEE_CONFIG = {
+  // Test wallet address for builder fees, currently staking test wallet
+  testnetBuilder: '0x316BDE155acd07609872a56Bc32CcfB0B13201fA' as `0x${string}`,
```
abretonc7s: Let's use Hex instead `0x${string}` for consistency

app/components/UI/Perps/constants/hyperLiquidConfig.ts
```diff
@@ -102,6 +102,24 @@ export const FEE_RATES: FeeRatesConfig = {
...
+  // Production builder wallet
+  mainnetBuilder: '0xe95a5e31904e005066614247d309e00d8ad753aa' as `0x${string}`,
```
abretonc7s: same here

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
abretonc7s: Why dont we directly set the value in the config in the expected format and avoid having to convert? This will simplify code

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
abretonc7s: lets use Hex and maybe we can maintain using object parameters for all new methods to maintain consistency. --> params: BuildFeeApprovalParam

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -192,6 +294,12 @@ export class HyperLiquidProvider implements IPerpsProvider {
 
       await this.ensureReady();
 
+      // Ensure builder fee approval and referral code are set before placing any order
+      await Promise.all([
+        this.ensureBuilderFeeApproval(),
+        this.ensureReferralSet(),
+      ]);
```
abretonc7s, line 281: I think this is fine for now, but may slow down a bit the order, we may want to cache that per account to avoid re-checking. Kepping this is mind for future improvement

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+        console.error('User is the referrer - skipping', {
+          network,
+          referrerAddress,
+          userAddress,
+        });
+        DevLogger.log('User is the referrer - skipping', {
+          network,
+          referrerAddress,
+          userAddress,
+        });
```
abretonc7s: we can remove devlogger since we have real error

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+
+      if (!hasReferral) {
+        DevLogger.log('No referral set - setting MetaMask as referrer', {
+          network,
+          referralCode: expectedReferralCode,
+        });
+        await this.setReferralCode();
+      }
+    } catch (error) {
+      DevLogger.log(errorMessage, error);
```
abretonc7s: no need to duplicate with console.error

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+      }
+      console.error('Referral code not ready', {
+        stage,
+        code,
+        referrerAddr,
+        referral,
+      });
+      return false;
+    } catch (error) {
+      DevLogger.log(errorMessage, error);
```
abretonc7s: remove devlogger

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+        code: referralCode,
+      });
+
+      DevLogger.log('Referral code set result:', result);
+
+      return result?.status === 'ok';
+    } catch (error) {
+      DevLogger.log(errorMessage, error);
+      console.error(errorMessage, error);
+      throw new Error(errorMessage);
```
abretonc7s: prevent duplicate log and probably on all errors we should leverage       return createErrorResult(error, { success: false });

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2057,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+      const referral = await infoClient.referral({ user: referrerAddr });
+
+      const stage = referral.referrerState?.stage;
+
+      if (stage === 'ready') {
+        const onFile = referral.referrerState?.data.code;
+        if (onFile.toUpperCase() !== code.toUpperCase()) {
+          throw new Error(
+            `Ready for referrals but there is a config code mismatch ${onFile} vs ${code}`,
+          );
```
abretonc7s, line 2126: return createErrorResult(error, { success: false });
maybe we can reuse this pattern

app/components/UI/Perps/constants/hyperLiquidConfig.ts
abretonc7s: Lets not use bps for unit, instead the correct decimal to avoid conversion

abretonc7s (changes requested):

abretonc7s (changes requested):
A few things for consistency and simplification

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: The methods in question are private and ultimately are used by the public interface methods which will return a proper createErrorResult if there is an error thrown. Errors thrown by these private methods display properly in the ui currently.

app/components/UI/Perps/constants/hyperLiquidConfig.ts
nickewansmith: switched to decimal and conversions in the config inline, also renamed to maxFee prefix

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: param removed completely

app/store/migrations/094.test.ts
nickewansmith: removed mock

app/store/migrations/094.ts
nickewansmith: I did not add bec there is a null check

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: The builder fee check will probably be required on each order as the user can revoke the approval at any time.

The referral check could be cached, that would be awesome. The controller state doesn't seem like the place for this though. I'm fine with revisiting as the call for referral is not that expensive and run in parallel with builder checks.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -1936,4 +2035,175 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+      const userAddress = await this.walletService.getUserAddressWithDefault();
+
+      if (userAddress.toLowerCase() === referrerAddress.toLowerCase()) {
+        // if the user is the builder, we don't need to set a referral code
+        console.error('User is the referrer - skipping', {
+          network,
+          referrerAddress,
+          userAddress,
+        });
+        DevLogger.log('User is the referrer - skipping', {
```
abretonc7s: still have double log

app/components/UI/Perps/constants/hyperLiquidConfig.ts
```diff
@@ -102,6 +102,30 @@ export const FEE_RATES: FeeRatesConfig = {
...
+    .toFixed(4)
+    .replace(/\.?0+$/, '')}%` as `${string}%`,
+};
+
+// Referral code configuration
+export const REFERRAL_CONFIG = {
+  // Production referral code METAMASK
+  mainnetCode: 'METAMASK',
+  // Development/testnet referral code
+  testnetCode: 'MSO',
```
abretonc7s, line 128: we should add a note to remove this

abretonc7s: FIXME: use official testnetCode as soon as available

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -718,6 +813,10 @@ export class HyperLiquidProvider implements IPerpsProvider {
       const result = await exchangeClient.order({
         orders,
         grouping: 'positionTpsl',
+        builder: {
+          b: this.getBuilderAddress(this.clientService.isTestnetMode()),
+          f: BUILDER_FEE_CONFIG.maxFeeTenthsBps,
```
abretonc7s, line 818: this value will need to be dynamic once we have reward (potential discount)

abretonc7s (dismissed):
just tiny fixes, and to track we need to change the testnetcode, or we make a jira task separately. Otherwise just to remove the log

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
nickewansmith: Yes, I am aware. For now though I believe we should be charging the max fee.

abretonc7s (approved):

abretonc7s (dismissed):
just need to fix the cursor bot bug

gambinish (dismissed):

gambinish (dismissed):

sethkfman (approved):
LGTM - reviewed migration and engine updates

gambinish (approved):

PR #18315, opened 2025-08-13
feat(assets): add account group balance selectors and components

amitabh94:
## **Description**

Implement aggregated balance and portfolio change on the Wallet home screen (feature-flagged) using core calculators to mirror the extension behavior.

- Aggregated balance selectors in `app/selectors/assets/balances.ts` using core calculators and narrowed controller slices:
  - `selectBalanceForAllWallets`
  - `selectSelectedGroupAggregatedBalance`
  - `selectBalanceByAccountGroup`
  - `selectBalanceByWallet`
  - `selectBalanceChangeForAllWallets`
  - `selectBalanceChangeByAccountGroup`
  - `selectBalancePercentChange`
  - `selectBalancePercentChangeByAccountGroup`

- Narrowed state extractors for accounts, tokens, currency rates, and account tree (only fields required by calculators)

- UI wiring in `app/components/Views/Wallet/index.tsx`:
  - When `isMultichainState2Enabled` is true, render new `AccountGroupBalance` component 
  - Keep legacy rendering when the feature flag is off`

- Tests:
  - `app/selectors/assets/balances.test.ts` covering aggregate-by-wallet and zero/fallback cases
  - `app/components/UI/Assets/BalanceChange/AccountGroupBalanceChange.test.tsx` covering positive/negative changes and privacy mode

> **P.S. The aggregated balances are missing the native balances and staked balances as of now which will automatically be added in the next assets controller release.**

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: N/A

## **Manual testing steps**

```gherkin
Feature: Aggregated balance and portfolio change (mobile)

  Scenario: Total balance and 1d change with feature flag enabled
    Given the feature flag "isMultichainState2Enabled" is enabled
    And the app is built with @metamask-previews/assets-controllers@73.1.0-preview-a89812f5
    And multiple accounts with tokens exist across chains
    When I open the Wallet home screen
    Then I see the total balance equal to the aggregated fiat value of all wallets
    And I see the 1d portfolio change row with amount and percent for the selected account group

  Scenario: Account group switch updates values
    Given the feature flag is enabled
    And I have at least two account groups with different holdings
    When I switch the selected account group
    Then the total balance and 1d change update to reflect the new group's aggregation only

  Scenario: Legacy behavior when feature flag disabled
    Given the feature flag "isMultichainState2Enabled" is disabled
    When I open the Wallet home screen
    Then the legacy balance and change components render instead of aggregated values

  Scenario: Privacy mode formatting parity
    Given privacy mode is enabled
    When the Wallet home screen renders
    Then the balance text uses the same obscured length and formatting as before

  Scenario: Empty state
    Given no token balances are available
    When the Wallet home screen renders
    Then the total balance displays 0 and portfolio change displays 0 amount and 0%
```

## **Screenshots/Recordings**

<div>
    <a href="https://www.loom.com/share/fcfc633b4b1b461fb48f92d9428c5476">
      <p>Aggregated Balances - BIP 44 - Watch Video</p>
    </a>
    <a href="https://www.loom.com/share/fcfc633b4b1b461fb48f92d9428c5476">
      
    </a>
  </div>


app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
```diff
@@ -30,6 +34,23 @@ const AccountCell = ({ accountGroup, isSelected }: AccountCellProps) => {
     });
   }, [navigate, accountGroup]);
 
+  const selectBalanceForGroup = useMemo(
+    () => selectBalanceByAccountGroup(accountGroup.id),
+    [accountGroup.id],
+  );
```
gantunesr, line 40: I'm unsure if we should be accessing redux selectors inside the DS components since we need them to be flexible in case we want to introduce a different use case.

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
amitabh94, line 1: This is the new component that will replace the PortfolioBalance component and remove all the aggregation logic from the components and hooks layer.

app/selectors/assets/balances.ts
amitabh94, line 1: This file contains all the new selectors for aggregated balances and balance changes by wallet, group and also for currently selected group.

app/components/Views/Wallet/index.tsx
```diff
@@ -980,7 +986,11 @@ const Wallet = ({
           </View>
         ) : null}
         <>
-          <PortfolioBalance />
+          {isMultichainAccountsState2Enabled ? (
```
amitabh94, line 1036: BIP44 feature flag that uses the new balance component when turned on.

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
```diff
@@ -0,0 +1,87 @@
...
+  );
+
+  const togglePrivacy = useCallback(
+    (value: boolean) => {
+      PreferencesController.setPrivacyMode(value);
+    },
+    [PreferencesController],
+  );
+
+  const displayBalance = (() => {
```
gantunesr: Should this be wrapped in a `useCallback` hook?

sethkfman (dismissed):
LGTM - Reviewed Wallet & CODEOWNER file

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
```diff
@@ -0,0 +1,87 @@
...
+                }
+                percentChange={balanceChange1d.percentChange}
+                userCurrency={balanceChange1d.userCurrency}
+              />
+            )}
+          </TouchableOpacity>
+        ) : (
+          <View style={styles.skeletonContainer}>
+            <Skeleton width={100} height={40} />
+            <Skeleton width={100} height={20} />
```
gantunesr, line 80: Should this be part of a styles file?

app/selectors/assets/balances.ts
```diff
@@ -0,0 +1,348 @@
...
+    accountsState: AccountsControllerState,
+    tokenBalancesState: TokenBalancesControllerState,
+    tokenRatesState: TokenRatesControllerState,
+    multichainRatesState: MultichainAssetsRatesControllerState,
+    multichainBalancesState: MultichainBalancesControllerState,
+    tokensState: TokensControllerState,
+    currencyRateState: CurrencyRateState,
+    enabledNetworkMap: Record<string, Record<string, boolean>> | undefined,
+  ) =>
+    calculateBalanceForAllWallets(
```
tommasini, line 140: I believe as a future performance improvement, this can be used in worklets 
https://github.com/margelo/react-native-worklets-core

I'm saying this because I believe that the time complexity and space complexity of the [calculateBalancesForWallets](https://github.com/MetaMask/core/blob/b7b8be4608b86a969996d4e3bcc21c4cc25c2de3/packages/assets-controllers/src/balances.ts#L301) function could have performance challenges (FPS drop), at Android lower end devices

We should double check the behavior there with a power user account

app/selectors/assets/balances.ts
```diff
@@ -0,0 +1,348 @@
...
+  calculateBalanceForAllWallets,
+  type TokenBalancesControllerState,
+  type TokenRatesControllerState,
+  type MultichainBalancesControllerState,
+  type MultichainAssetsRatesControllerState,
+  type TokensControllerState,
+  type CurrencyRateState,
+  type BalanceChangeResult,
+  calculateBalanceChangeForAllWallets,
+  calculateBalanceChangeForAccountGroup,
```
tommasini, line 12: Flagging here as well that this math calculations can be done under worklets, that will not block the main thread, and will give the app a more performative feel

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
owencraston: Pretty sure we should lift the state up to the parent. cc @brianacnguyen

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
```diff
@@ -30,6 +34,23 @@ const AccountCell = ({ accountGroup, isSelected }: AccountCellProps) => {
...
+  const groupBalance = useSelector(selectBalanceForGroup);
+  const displayBalance = useMemo(() => {
+    if (!groupBalance) {
+      return undefined;
+    }
+    const value = groupBalance.totalBalanceInUserCurrency;
```
owencraston: why not move the formatting logic inside the selector. this mimics a backend/front end model.

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
amitabh94: I believe I assumed that is the component specifically owned by accounts team as its usage was only in the account view . Looks like we need to enhance the component to pass down balance data in case this is a generic component. 

If this is a DS component , then I can move this portion of the account list out of this PR into a separate PR so that you guys can handle it properly in case I am missing context around how these views and components are structured for accounts team.

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
amitabh94: We actually can and return a formatted value . Will push a commit for this.

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
```diff
@@ -63,10 +84,7 @@ const AccountCell = ({ accountGroup, isSelected }: AccountCellProps) => {
           color={TextColor.Default}
           testID={AccountCellIds.BALANCE}
         >
-          {
```
amitabh94, line 66: @gantunesr @owencraston How are you guys thinking of replacing the balance here?

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
amitabh94: We mostly use the skeleton loader in other experiences .Any reason we would want to have this in the styles file?

app/selectors/assets/balances.ts
amitabh94: Noted. Will check this with help of performance team once we have a experimental build on a low end physical android device.

app/selectors/assets/balances.ts
amitabh94: Noted. Will check how worklets is currently  being used in the app and if not then how we can best use them going forward if this presents performance issues after testing for a power user.

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
amitabh94: Moved the formatting to within the selector itself

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
amitabh94: On second thought , I have decided to keep it out as formatting in selectors adds sync CPU on every recompute and creates new strings/objects, breaking memoization and causing extra re-renders. We’ll keep selectors numeric and format at the component edge (memoized) to keep lists and the JS thread responsive.

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
amitabh94: I have updated it to use useMemo() so as to avoid recompute on unrelated re renders.

app/component-library/components-temp/MultichainAccounts/AccountCell/AccountCell.tsx
gantunesr: Final decision is to refactor it in a follow up PR

app/components/UI/Assets/components/Balance/AccountGroupBalance.tsx
gantunesr: I thought it was the pattern we were following, if it is not the case then nevermind

gantunesr (approved):

owencraston (approved):

sethkfman (approved):

app/selectors/assets/balances.ts
Prithpal-Sooriya: @tommasini really nice callout, this is a great use-case! Only concern is:
1. Calling worklets inside selectors (selectors can fire frequently)
2. Worklets might make this data "async", and we would need to rethink how to calculate this data for use in components.

PR #18370, opened 2025-08-15
Rwds 175 add silent login support to the app

sophieqgu:
## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```


app/core/Engine/controllers/rewards-controller/hooks/useRewardsAuth.ts
```diff
@@ -1,39 +1,77 @@
...
 
-  const handleLogin = useCallback(async () => {
+  // Update subscription ID when address or auth state changes
+  useEffect(() => {
+    if (!address) {
+      setSubscriptionId(null);
+      return;
+    }
+    const id =
+      Engine.context.RewardsController.getSubscriptionIdForAccount(address);
```
VGR-GIT: Guidelines say that we have to use redux selectors as much as possible. If we did that, we wouldn't need to directly  use the rewards controller singleton. A bigger benefit would be that we can keep this reactive and use the useSelector hook.

app/core/Engine/controllers/rewards-controller/hooks/useRewardsAuth.ts
```diff
@@ -54,33 +92,52 @@ export const useRewardsAuth = ({
...
     Engine.context.RewardsController.setDevOnlyLoginAddress(null);
   }, [logout]);
 
-  const clearLoginError = useCallback(() => setLoginError(null), []);
+  const clearOptinError = useCallback(() => setOptinError(null), []);
+
+  // Check authentication state when address changes
+  useEffect(() => {
+    checkAuthState();
+  }, [checkAuthState, address]);
```
VGR-GIT, line 105: @sophieqgu Wondering why we need this, `checkAuthState` and the useEffect directly below if we're just reexposing the rewards controller state. Would it be enough to use a selector (define a selector in rewards controller + useSelector) and have our app react to changes?

app/core/Engine/controllers/rewards-controller/types.ts
```diff
@@ -171,11 +171,19 @@ export enum RewardClaimStatus {
 /**
  * State for the RewardsController
  */
+export interface SilentAuthState {
```
VGR-GIT: Do we need to make the `Silent` distinction? I'm wondering if it doesn't make more sense for both internal and external use to just talk about `AuthState`. The means to populate this is silently but that shouldn't dictate naming here imo.

app/core/Engine/messengers/rewards-controller-messenger/index.ts
```diff
@@ -15,11 +15,20 @@ export type RewardsControllerMessenger = RestrictedMessenger<
...
+    RewardsControllerMessengerActions,
+    RewardsControllerMessengerEvents
+  >,
 ): RewardsControllerMessenger {
   return messenger.getRestricted({
     name,
-    allowedActions: [],
-    allowedEvents: [],
+    allowedActions: [
+      'AccountsController:getSelectedAccount',
```
VGR-GIT: Wondering if we should use `getSelectedMultichainAccount` instead as `getSelectedAccount` seems to be scoped solely to EVM accounts? In contrast, `selectedAccountChange` is firing for both EVM and non-EVM accounts.

VGR-GIT: This would apply to everywhere we're using this fn

app/components/Views/RewardsView/RewardsHero.tsx
```diff
@@ -159,10 +159,10 @@ const RewardsHero: React.FC<RewardsHeroProps> = ({
 
         <Text style={styles.seasonEndText}>{formatSeasonEndDate(endDate)}</Text>
 
-        {loginError && (
+        {optinError && (
```
VGR-GIT, line 156: This feels like something we should be able to access through a selector on our controller state? Versus having to pass it through components as props.

app/core/Engine/controllers/rewards-controller/RewardsController.ts
```diff
@@ -50,6 +82,20 @@ export class RewardsController extends BaseController<
...
+      () => this.#handleAuthenticationTrigger('Account changed'),
+    );
+
+    // Subscribe to KeyringController unlock events to retry silent auth
+    this.messagingSystem.subscribe('KeyringController:unlock', () =>
+      this.#handleAuthenticationTrigger('KeyringController unlocked'),
+    );
+
+    // Initialize silent authentication on startup
+    this.#handleAuthenticationTrigger('Controller initialized');
```
VGR-GIT, line 118: @sophieqgu Haven't run this yet, but I was wondering if we're not having too much listeners here. Reading this I would assume we'd only need `AccountsController:selectedAccountChange`?

VGR-GIT (changes requested):

app/core/Engine/controllers/rewards-controller/RewardsController.ts
sophieqgu: This is to ensure that silent auth runs on app startup. There's also a state that the Controller has initiated but user has not unlocked the wallet yet (has to enter password). Then, the silent auth will return error " Failed to generate signature: [Error: KeyringController - The operation cannot be completed while the controller is locked.]" So I made it to also listen to the `KeyringController:unlock` event to ensure that silent auth runs again when the wallet is unlocked.

app/core/Engine/controllers/rewards-controller/hooks/useRewardsAuth.ts
sophieqgu: There is an async call inside `checkAuthState` that retrieves the token from vault, and it manages state updates `isAuthenticated` and `isAuthenticating` based on async results. I thought selectors can only deal with synchromous states.

app/component-library/components/Navigation/TabBar/TabBar.tsx
```diff
@@ -128,9 +128,6 @@ const TabBar = ({ state, descriptors, navigation }: TabBarProps) => {
     [
       descriptors,
       state.index,
-      colors.primary.default,
-      colors.primary.inverse,
-      colors.icon.muted,
```
brianacnguyen, line 133: why are these removed?
also can you post a screenshot of the after changes?

brianacnguyen (changes requested):
I'd recommend to pull main into `feat-rewards-page` to get the latest changes to the TabBar/Homepage. There's been quite a few changes that's not reflected in this branch

sophieqgu:
Will close in favor of: https://github.com/MetaMask/metamask-mobile/pull/18582

PR #18383, opened 2025-08-15
feat(perps): implement stream architecture for optimized rendering performance

abretonc7s:
## **Description**

  This PR implements a new stream architecture for Perps that optimizes rendering performance by allowing each component to subscribe to data updates with its own
  refresh interval. This prevents cascade re-rendering and gives components fine-grained control over their update frequency.

  The architecture introduces a centralized `PerpsStreamManager` that manages WebSocket subscriptions and distributes updates to components based on their
  individual debounce settings. This first PR focuses on optimizing `PerpsOrderView` with a 10-second refresh rate for stable price display during order entry.

  ## **Changelog**

  CHANGELOG entry: Improved Perps trading performance by optimizing component rendering and update frequencies

  ## **Related issues**

  Fixes: Performance issues with excessive re-renders in Perps trading views

  ## **Manual testing steps**
```gherkin
Feature: Perps Stream Architecture Performance Optimization

    Scenario: User places an order with optimized price updates
      Given the user is on the Perps Order View screen
      And the stream architecture is active

      When user enters an order amount
      Then prices update every 10 seconds (not on every WebSocket message)
      And the component does not re-render unnecessarily
      And WDYR logs show reduced re-renders

    Scenario: Multiple components with different refresh rates
      Given multiple Perps components are mounted
      And each has different refresh requirements

      When WebSocket receives price updates
      Then each component updates at its configured interval
      And components don't interfere with each other's update cycles
```

### **Before**

  Multiple re-renders on every WebSocket update, components interfering with each other's debounce timers, WDYR showing excessive re-renders for unchanged data

### **After**

  Price updates every 10 seconds in PerpsOrderView, no parent component re-renders, WDYR showing optimized render patterns, stable array references preventing
  unnecessary re-renders


app/components/UI/Perps/Views/__snapshots__/PerpsView.test.tsx.snap
```diff
@@ -752,12 +755,13 @@ exports[`PerpsView should render correctly when initializing 1`] = `
...
               $
-              2
+              ---
             </Text>
             <Text
               accessibilityRole="text"
             >
-              -3.00
+              +
+              0.00
```
nickewansmith, line 764: Are these snaps ok changing to zero values?

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -0,0 +1,338 @@
...
+    const controller = Engine.context.PerpsController;
+    const pollInterval = setInterval(async () => {
+      try {
+        const orders = await controller.getOrders();
+        this.cache.set('orders', orders);
+        this.notifySubscribers(orders);
+      } catch (error) {
+        // Handle error silently
+      }
+    }, 5000);
```
nickewansmith, line 255: Should we have some point where we error if connection fails?

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -238,9 +237,10 @@ const PerpsOrderViewContent: React.FC = React.memo(() => {
         });
       },
     });
-
-  const [selectedPaymentToken, setSelectedPaymentToken] =
-    useState<PerpsToken | null>(null);
+  // Update ref when orderType changes
+  useEffect(() => {
+    orderTypeRef.current = orderForm.type;
+  }, [orderForm.type]);
```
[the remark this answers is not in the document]
gambinish: This could cause a double re-render potentially?

app/components/UI/Perps/Views/__snapshots__/PerpsView.test.tsx.snap
```diff
@@ -163,13 +163,13 @@ exports[`PerpsView should render correctly in connected state 1`] = `
               accessibilityRole="text"
             >
               $
-              50,000
+              ---
```
gambinish, line 166: This diff seems kinda sus doesn't it? Did the price streaming changes cause this snapshot update?

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
```diff
@@ -104,7 +105,11 @@ const PerpsTPSLBottomSheet: React.FC<PerpsTPSLBottomSheetProps> = ({
   const { track } = usePerpsEventTracking();
 
   // Subscribe to real-time price only when visible and we have an asset
-  const priceData = usePerpsPrices(isVisible && asset ? [asset] : [], {});
+  // Use 1s debounce for TP/SL bottom sheet
+  const priceData = useLivePrices({
+    symbols: isVisible && asset ? [asset] : [],
+    debounceMs: 1000,
+  });
```
gambinish, line 112: why do we need to debounce this?

app/components/UI/Perps/hooks/stream/index.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Export types for components to use
+export type { PriceUpdate } from '../../controllers/types';
+
+// Future hooks to be added:
+// export function useLiveFunding(debounceMs: number = 5000): Funding[] { ... }
+// export function useLiveAccountState(debounceMs: number = 2000): AccountState { ... }
+// export function useLiveOrderBook(symbol: string, debounceMs: number = 100): OrderBook { ... }
+// export function useLiveTrades(symbol: string, debounceMs: number = 0): Trade[] { ... }
```
gambinish, line 177: In my charting integration, I'm subscribing to the current candle as it comes in, we may need a `useLiveCandle` hook as well to follow this pattern

app/components/UI/Perps/providers/PerpsStreamManager.tsx
abretonc7s: these will be removed with the upcoming PRs, currently just placeholder. It will be via websocket

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
abretonc7s: I dont think we need to have more than 1 price update per second, even that is probably too much and make ui confusing

app/components/UI/Perps/Views/__snapshots__/PerpsView.test.tsx.snap
abretonc7s: I would say so, probably wont change anything. Not sure actually where was the -3 from (maybe from mock)

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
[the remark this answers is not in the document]
abretonc7s: yes it got mixed up while rebasing

app/components/UI/Perps/components/PerpsTPSLBottomSheet/PerpsTPSLBottomSheet.tsx
abretonc7s: the idea is to control the rendering frequency, we can adjust the debounce value as we see fit

app/components/UI/Perps/Views/__snapshots__/PerpsView.test.tsx.snap
abretonc7s: Probably, I think it could be lower prioritiy to focus on the snapshot

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -0,0 +1,338 @@
...
+      if (this.subscribers.size === 0) {
+        DevLogger.log('StreamChannel: No subscribers left, disconnecting');
+        this.disconnect();
+      }
+    };
+  }
+
+  protected connect() {
+    // Override in subclasses
+  }
```
gambinish: Is this doing anything? Is it needed?

I suppose it is overriding, but it does seem weird to have an empty function here. If it's needed that's okay, but maybe we can update the comment to be a bit more descriptive?

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -0,0 +1,338 @@
...
+    if (this.wsSubscription) {
+      this.wsSubscription();
+      this.wsSubscription = null;
+    }
+  }
+
+  protected getCachedData(): T | null {
+    // Override in subclasses
+    return null;
+  }
```
gambinish: Likewise here

app/components/UI/Perps/providers/PerpsStreamManager.tsx
gambinish, line 255: We don't need to implement this here, but I wonder if we should have this kind of a polling fallback mechanism in place for all web socket connections in case of failure 🤔

abretonc7s: yes all this will be gone and replaced via websocket, I should have removed it to avoid confusion

app/components/UI/Perps/providers/PerpsStreamManager.tsx
abretonc7s: improved comment

app/components/UI/Perps/providers/PerpsStreamManager.tsx
abretonc7s: I think it is fine to keep it, otherwise it needs to change to an abstract class but I dont know if that helps and create more changes

gambinish (dismissed):

Matt561 (approved):

PR #18402, opened 2025-08-15
fix: enable network after swaps

### LINKED ISSUE #18395: [Bug]: [Remove GNS] After swapping with a network that isn't selected in the Network Manager, you don't see your latest transaction
### Describe the bug

When you perform a swap on a network that isn't enabled in the network manager and then navigate to your activity view, you won't see the latest transaction because the network manager hasn't enabled the network you just swapped on.

### Expected behavior

When a user performs a swap on a network that isn't enabled in the network manager, the system should automatically enable that network to ensure the transaction appears in the user's activity view. This prevents the scenario where users complete transactions on networks they haven't manually enabled, only to find those transactions missing from their transaction history.

### Specifically:
- When a swap transaction is initiated on a network that isn't currently enabled
- The network manager should automatically enable that network
- The transaction should then be visible in the user's activity/transaction history
- This ensures users don't lose track of their transactions due to network enablement state

### Screenshots/Recordings

_No response_

### Steps to reproduce

1. Goto uniswap swaps
2. Choose a asset with a network that isn't enabled in the network manager
3. Perform the swap
4. Navigate to the activity view
5. Observe that you cannot see your transaction because the network isn't enabled

```shell

```

### Detection stage

In production (default)

### Version

7.54

### Build type

None

### Device

All

### Operating system

Other (please elaborate in the "Additional Context" section)

### Additional context

_No response_

### Severity

_No response_

vinnyhoward:
## **Description**

This fix is for the features hidden under the `MM_REMOVE_GLOBAL_NETWORK_SELECTOR` feature flag

**Problem**: When you perform a swap on a network that isn't enabled in the network manager and then navigate to your activity view, you won't see the latest transaction because the network manager hasn't enabled the network you just swapped on.

**Solution**: Enable network after a swap

Why introduce `tryEnableEvmNetwork()` versus using other helper functions?
- Conditional logic: Unlike `enableNetwork()` which always enables a network or `toggle` which turns on/off a network, `tryEnableEvmNetwork()` includes checks and validation
- Safety: It prevents enabling already-enabled networks and validates input parameters
- Consistency: Uses the same CAIP-2 chain ID format as other network operations
- Reusability: Can be used across different transaction flows (swaps, confirmations) without duplicating logic

## **Changelog**

CHANGELOG entry: Fixed - Auto-enable networks in Network Manager after swaps and transaction confirmations 

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/18395

## **Manual testing steps**

Make sure to set these feature flags to test:
```bash
export MM_REMOVE_GLOBAL_NETWORK_SELECTOR="true"
export MM_UNIFIED_SWAPS_ENABLED="false"
export MM_BRIDGE_ENABLED="true"
```

```gherkin
Feature: Network Manager

  Scenario: User swaps on a network not enabled in Network Manager
    Given the user has enabled Binance chain in Network Manager and nothing else
    And they have assets in another chain like Ethereum
    When the user goes to a dapp like Uniswap and swaps on Ethereum
    And the user finishes the transaction
    Then the network should be automatically enabled in Network Manager
    And when they go to the activity view
    Then they should see their transaction because the network is now enabled
```

#### Dapp Swap
https://github.com/user-attachments/assets/e4c1318f-90c7-4fdc-8bf0-a5bcd0f46f56

#### Legacy Swap
https://github.com/user-attachments/assets/545ba624-41e8-4a14-ac7c-79450d3330ac

#### Dapp Swap
https://github.com/user-attachments/assets/07b1ad60-c5b3-4aff-b90c-2175ebd484d0

#### Legacy Swap
https://github.com/user-attachments/assets/28d72d4b-291a-41f1-8558-781bb8c3a282


app/components/UI/Swaps/QuotesView.js
```diff
@@ -1228,6 +1232,11 @@ function SwapsQuotesView({
       setIsHandlingSwap(false);
       navigation.dangerouslyGetParent()?.pop();
     }
+
+    // Enable the network if it's not enabled for the Network Manager
+    if (isRemoveGlobalNetworkSelectorEnabled()) {
+      tryEnableEvmNetwork(chainId);
+    }
```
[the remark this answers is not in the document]
vinnyhoward: Fixed

app/components/hooks/useNetworkEnablement/useNetworkEnablement.ts
```diff
@@ -94,6 +110,19 @@ export const useNetworkEnablement = () => {
...
+        const caipChainId = toCaipChainId(KnownCaipNamespace.Eip155, chainId);
+        const isEnabled = isNetworkEnabled(caipChainId);
+        if (!isEnabled) {
+          enableNetwork(caipChainId);
+        }
+      }
+    },
+    [isNetworkEnabled, enableNetwork],
+  );
+
```
micaelae, line 117: @vinnyhoward I think I'm missing some context here - what does it mean for a network to be enabled? How is it possible for the user to navigate to the Swap experience, get token lists, balances and submit a transaction without the network being enabled?

vinnyhoward: So we recently introduced the Network Manger/Filter into mobile behind a feature flag. It operates just like the extension version where users can sort their tokens, NFTs, and activity by networks. But sorting them does not affect critical flows such as send and swap. However when we complete a transaction, we want to add the network to the filter of the Network Manager if it wasn't added before. Because if the user swaps on a network that wasn't added to the filter, they wont see their transaction activity. I'd be happy to get on a call with you to walk through the flows

micaelae: Gotcha - that makes sense! Thanks for the explanation 🙏

micaelae (dismissed):

infiniteflower (dismissed):

app/components/Views/confirmations/hooks/useConfirmActions.ts
```diff
@@ -79,6 +82,11 @@ export const useConfirmActions = () => {
   );
 
   const onConfirm = useCallback(async () => {
+    // Enable the network if it's not enabled for the Network Manager
+    if (isRemoveGlobalNetworkSelectorEnabled()) {
+      tryEnableEvmNetwork(chainId);
+    }
```
OGPoyraz: Is there any reason that this action is happening before actual confirm action? Any downside if we call this after `onRequestConfirm`? 

Also this function will be called for signatures as well, which is not really necessary. In that case could it be pulled under `isTransactionReq` if block?

app/components/Views/confirmations/hooks/useConfirmActions.test.ts
```diff
@@ -302,4 +310,192 @@ describe('useConfirmAction', () => {
...
+        disableNetwork: jest.fn(),
+      } as unknown as ReturnType<
+        typeof NetworkEnablementHook.useNetworkEnablement
+      >['networkEnablementController'],
+      enableNetwork: jest.fn(),
+      disableNetwork: jest.fn(),
+      toggleNetwork: jest.fn(),
+      isNetworkEnabled: jest.fn(),
+      hasOneEnabledNetwork: false,
+    });
```
OGPoyraz: Can we reduce the duplication of this mock as it's duplicated 4 times?

app/components/Views/confirmations/hooks/useConfirmActions.test.ts
```diff
@@ -302,4 +310,192 @@ describe('useConfirmAction', () => {
...
+
+    const { result } = renderHookWithProvider(() => useConfirmActions(), {
+      state,
+    });
+
+    result?.current?.onConfirm();
+    await flushPromises();
+
+    expect(mockTryEnableEvmNetwork).toHaveBeenCalledWith('0x1');
+  });
```
OGPoyraz: Isn't enabling/disabling (depending of what chain is enabled) checked in the `tryEnableEvmNetwork` callback? Why we are unit testing that part here? Shouldn't we only check if `tryEnableEvmNetwork` executed or not?

app/components/Views/confirmations/hooks/useConfirmActions.test.ts
```diff
@@ -302,4 +310,192 @@ describe('useConfirmAction', () => {
...
+
+  it('enables network when ledger signing is in progress and feature flag is enabled', async () => {
```
OGPoyraz: I believe the switching part should have done after actual confirmation happens - if that's the case this ledger and QR signing tests are invalid.

app/components/Views/confirmations/hooks/useConfirmActions.ts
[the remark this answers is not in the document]
OGPoyraz: This is a valid comment.

OGPoyraz (changes requested):

app/components/Views/confirmations/hooks/useConfirmActions.ts
vinnyhoward: Good call, let me place it under the `isTransactionReq`

app/components/Views/confirmations/hooks/useConfirmActions.test.ts
vinnyhoward: This is actually needed. The `enableNetwork` and `disableNetwork` are separate methods. One belongs in the `networkEnablementController` and the other belongs in the `useNetworkEnablement` hook. You can check out the implementation details in [here](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/hooks/useNetworkEnablement/useNetworkEnablement.ts)

app/components/Views/confirmations/hooks/useConfirmActions.test.ts
vinnyhoward: You're right good catch. I will remove. The test coverage should only cover if it is a transaction request

app/components/Views/confirmations/hooks/useConfirmActions.ts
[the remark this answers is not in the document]
vinnyhoward: Will be moving to transaction req block like you mentioned in the other comment

OGPoyraz (dismissed):

micaelae (dismissed):

micaelae (dismissed):

OGPoyraz (dismissed):

pedronfigueiredo (approved):

micaelae (approved):

PR #18489, opened 2025-08-20
feat: Filter from account in the recipient page

OGPoyraz:
## **Description**

This PR aims to filter from account from the recipient list.
It also implements that selected internal account will be defined by the selected token as we are listing account groups asset page. 

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/mobile-planning/issues/2286

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### DIFF
--- a/app/components/Views/confirmations/components/send/amount/amount-keyboard/amount-keyboard.styles.ts
+++ b/app/components/Views/confirmations/components/send/amount/amount-keyboard/amount-keyboard.styles.ts
@@ -16,7 +16,7 @@ export const styleSheet = (params: {
       backgroundColor: continueDisabled
         ? theme.colors.error.default
         : theme.colors.primary.default,
-      marginBottom: 6,
+      marginBottom: 12,
     },
   });
 };
--- a/app/components/Views/confirmations/components/token-list/token-list.tsx
+++ b/app/components/Views/confirmations/components/token-list/token-list.tsx
@@ -7,9 +7,7 @@ import {
   TextVariant,
   TextColor,
 } from '@metamask/design-system-react-native';
-import { useSelector } from 'react-redux';
 
-import { selectInternalAccountsById } from '../../../../../selectors/accountsController';
 import Routes from '../../../../../constants/navigation/Routes';
 import { strings } from '../../../../../../locales/i18n';
 import { AssetType } from '../../types/token';
@@ -31,11 +29,10 @@ export function TokenList({
   onClearFilters,
 }: TokenListProps) {
   const { gotToSendScreen } = useSendScreenNavigation();
-  const { updateAsset, updateFromAccount } = useSendContext();
+  const { updateAsset } = useSendContext();
   const { captureAssetSelected } = useAssetSelectionMetrics();
   const [visibleTokenCount, setVisibleTokenCount] =
     useState(TOKEN_COUNT_PER_PAGE);
-  const accounts = useSelector(selectInternalAccountsById);
 
   const handleTokenPress = useCallback(
     (asset: AssetType) => {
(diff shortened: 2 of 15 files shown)


app/components/Views/confirmations/hooks/send/evm/useSelectedEVMAccountTokens.ts
OGPoyraz, line 1: Again I am not going to add any test for this file as it's a temporary hook.

app/components/Views/confirmations/context/send-context/send-context.tsx
```diff
@@ -53,10 +53,11 @@ export const SendContextProvider: React.FC<{
       value={{
         asset,
         chainId: chainId as string | undefined,
-        fromAccount: from as InternalAccount,
-        from: from?.address as string,
+        fromAccount,
+        from: fromAccount?.address as string,
```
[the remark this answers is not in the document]
OGPoyraz: Invalid as we always expect `fromAccount` to be filled in the flow

app/components/Views/confirmations/components/token-list/token-list.tsx
```diff
@@ -41,9 +44,17 @@ export function TokenList({
       );
       captureAssetSelected(asset, position.toString());
       updateAsset(asset);
+      updateFromAccount(accounts[asset.accountId as string]);
```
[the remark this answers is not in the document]
OGPoyraz: This will never be the case as we list tokens from those accounts already.

app/components/Views/confirmations/hooks/send/useSendActions.ts
```diff
@@ -28,7 +28,7 @@ export const useSendActions = () => {
       const toAddress = recipientAddress || to;
 
       if (isEvmSendType) {
-        await submitEvmTransaction({
+        submitEvmTransaction({
```
OGPoyraz, line 32: There is no downside doing this - as we are handling loading state instead of white screen in the confirmations and this feels way much snappier rather than waiting on the `recipient` screen.

OGPoyraz: https://github.com/user-attachments/assets/cb28ac1f-2f6e-4783-abb1-0c5432b35571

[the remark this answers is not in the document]
OGPoyraz: This is intentional

app/components/Views/confirmations/components/token-list/token-list.tsx
```diff
@@ -41,9 +44,17 @@ export function TokenList({
...
     },
-    [captureAssetSelected, gotToSendScreen, tokens, updateAsset],
+    [
+      captureAssetSelected,
+      gotToSendScreen,
+      tokens,
+      updateAsset,
+      updateFromAccount,
+      accounts,
+    ],
```
jpuri: We need to update from account even if user selects an asset and we directly take user to amount page.

Thus this may be a `useEffect` in the context to update from account, so even if asset change from anywhere it will update from account.

OGPoyraz: I am not sure if I follow - how could `asset - account` relationship can be changed if user is in the send flow?

OGPoyraz: Neither account group or the `accountId` in the `asset` cannot be changed while the user is in the send flow. 

I understand your concern but this can be a valid only if we are thinking a client like extension - it's not possible on mobile.

app/components/Views/confirmations/components/send/amount/amount-keyboard/amount-keyboard.styles.ts
```diff
@@ -16,7 +16,7 @@ export const styleSheet = (params: {
       backgroundColor: continueDisabled
         ? theme.colors.error.default
         : theme.colors.primary.default,
-      marginBottom: 6,
+      marginBottom: 12,
```
OGPoyraz, line 19: Minor update to keep "Continue" button to be same level with other percentage buttons

jpuri (dismissed):

jpuri (approved):

PR #18545, opened 2025-08-21
feat: add perps balance to total wallet balance, add pnl to perps tab

nickewansmith:
## **Description**

Integrates perps account balances into the wallet portfolio view, allowing users to see their total portfolio value including perps positions. 

**Reason for change:** Users couldn't see their complete financial position as perps balances were isolated from wallet totals.

**Improvement:** 
- Perps balances now included in total wallet balance
- WebSocket connections optimized - only active in perps environment
- Balances persist in Redux for offline viewing
- Real 24h historical data for accurate portfolio change calculations

## **Changelog**

CHANGELOG entry: Added perps account value to wallet total balance and unrealized P&L display

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1299
Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1044

## **Manual testing steps**

```gherkin
Feature: Perps Balance Integration

  Scenario: User views combined wallet and perps balance
    Given user has $1000 in wallet
    And user has $500 in perps account
    And perps feature is enabled

    When user opens wallet view
    Then total balance shows $1500
    And perps balance is included without WebSocket connection

  Scenario: User switches between wallet and perps tabs
    Given user is viewing wallet with perps balance included
    
    When user navigates to perps tab
    Then WebSocket connection establishes
    And real-time updates begin
    
    When user returns to wallet view
    Then WebSocket disconnects
    And perps balance persists from Redux

  Scenario: User restarts app with perps balance
    Given user has perps balance of $500
    And app is force quit
    
    When user reopens app
    Then perps balance of $500 displays immediately
    And no WebSocket connection is made until entering perps

  Scenario: User sees accurate 24h portfolio changes
    Given yesterday's perps balance was $400
    And today's perps balance is $500
    And wallet balance unchanged
    
    When user views portfolio percentage
    Then percentage reflects combined portfolio change
    And uses real historical data not estimates
```

### **Before**

- Perps balance not included in wallet total
- No historical data for accurate percentages

### **After**

- Unified portfolio balance (wallet + perps)
- Real 24h historical data for changes

Simulator Screenshot - iPhone 15 Pro - 2025-08-27 at 15 40 14


app/components/Views/Wallet/index.tsx
```diff
@@ -972,7 +974,9 @@ const Wallet = ({
           </View>
         ) : null}
         <>
-          <PortfolioBalance />
+          <PerpsStreamProvider>
```
abretonc7s: This should be part of the TabView, probably then the tab view can be broken into the separate components, otherwise we may have conneciton issue since the streamprovider also need to be within the perpsconnectionmanager

app/components/UI/Perps/hooks/usePerpsHistoricalPortfolio.ts
```diff
@@ -0,0 +1,34 @@
...
+  GetHistoricalPortfolioParams,
+  HistoricalPortfolioResult,
+} from '../controllers/types';
+import { usePerpsSelector } from './usePerpsSelector';
+
+/**
+ * Hook for accessing historical portfolio data with refresh capability
+ * Provides cached data from Redux state and method to refresh from API
+ */
+export function usePerpsHistoricalPortfolio() {
```
abretonc7s: should we have a loading state as well and potentially return error?

abretonc7s (changes requested):

app/components/Views/Wallet/index.tsx
nickewansmith: The wallet total is outside of the perps tab view. I left this connection provider out bc it worked and adding the connection provider was causing an infinite load. But now I see this has been fixed in a separate PR.

abretonc7s (dismissed):

gambinish (dismissed):
Mentioned this in Slack, but are we running the risk of having a aggregated balance discrepancy for the same account across mobile and extension? Since we aren't supporting Perps on extension, we would never include the balance there, and I wonder if the user would be confused about this.

I also think Assets team recently moved some balance calculation to core, so we might want to get their review here as well.

I'm also noticing that the info icon next to Unrealized P&L doesn't do anything. I think it's supposed to open a bottom sheet? Can be tacked on later.

https://github.com/user-attachments/assets/e847d1fe-8190-4c6a-8444-35a206ab14a3

amitabh94:
> Mentioned this in Slack, but are we running the risk of having a aggregated balance discrepancy for the same account across mobile and extension? Since we aren't supporting Perps on extension, we would never include the balance there, and I wonder if the user would be confused about this.
> 
> I also think Assets team recently moved some balance calculation to core, so we might want to get their review here as well.
> 
> I'm also noticing that the info icon next to Unrealized P&L doesn't do anything. I think it's supposed to open a bottom sheet? Can be tacked on later.
> 
>  Screen.Recording.2025-08-21.at.10.53.08.AM.mov

Thanks for raising this. Yes indeed the calculation for total balance has moved to core as part of the BIP 44 work. 
You can find the calculation [here](https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/balances.ts).

In order to add perps balances , you would need to make the changes in the assets controller which should be fairly straightforward if you have the perps balances and 1d change. Basically same thing you are doing in this PR but on the core side. Once you have the changes in the core, we can review that and then on the UI they should be available with very minor changes to selector inputs.  

The aggregation logic on the UI would not be used going forward and the PR for the same would be in main today or tomorrow depending on the CI gods so you will have more idea . 
Sharing the PR's for both mobile and extension to give you an idea of how things are changing. 

https://github.com/MetaMask/metamask-mobile/pull/18315
https://github.com/MetaMask/metamask-extension/pull/35012

TLDR: I wouldn't count too much on the portfolio balance component as it will soon be removed.

The question still remains that are we okay to display different balances on mobile and extension since perps balances will not be available on extension?

Thats said, we have precedent on this already where we add staking balances to total on mobile but not on extension .

gambinish:
>The question still remains that are we okay to display different balances on mobile and extension since perps balances will not be available on extension?

My understanding is that this is acceptable from a product perspective, and yes since we already have a similar discrepancy across clients related to some earn features like staking. We can follow up internally about this to double check

nickewansmith:
Oof ok, I'll take a look thanks @amitabh94 and @gambinish

amitabh94:
> > Mentioned this in Slack, but are we running the risk of having a aggregated balance discrepancy for the same account across mobile and extension? Since we aren't supporting Perps on extension, we would never include the balance there, and I wonder if the user would be confused about this.
> > I also think Assets team recently moved some balance calculation to core, so we might want to get their review here as well.
> > I'm also noticing that the info icon next to Unrealized P&L doesn't do anything. I think it's supposed to open a bottom sheet? Can be tacked on later.
> > Screen.Recording.2025-08-21.at.10.53.08.AM.mov
> 
> Thanks for raising this. Yes indeed the calculation for total balance has moved to core as part of the BIP 44 work. You can find the calculation [here](https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/balances.ts).
> 
> In order to add perps balances , you would need to make the changes in the assets controller which should be fairly straightforward if you have the perps balances and 1d change. Basically same thing you are doing in this PR but on the core side. Once you have the changes in the core, we can review that and then on the UI they should be available with very minor changes to selector inputs.
> 
> The aggregation logic on the UI would not be used going forward and the PR for the same would be in main today or tomorrow depending on the CI gods so you will have more idea . Sharing the PR's for both mobile and extension to give you an idea of how things are changing.
> 
> #18315 [MetaMask/metamask-extension#35012](https://github.com/MetaMask/metamask-extension/pull/35012)
> 
> TLDR: I wouldn't count too much on the portfolio balance component as it will soon be removed.
> 
> The question still remains that are we okay to display different balances on mobile and extension since perps balances will not be available on extension?
> 
> Thats said, we have precedent on this already where we add staking balances to total on mobile but not on extension .

Thinking more on this , I think it would be best to keep the logic for perps balances separate from the current aggregation logic in core. Maybe you can create a selector on the mobile codebase based on the changes you have already madeand then just add it to the aggregated balance selector and that way you can keep it under the feature flag as well and have more control whether to display in mobile vs extension . 

@nickewansmith This is something we might want to do for staked balances as well.

Think of it as like:

Mobile : `TotalBalanceSelector = (AggregatedBalanceSelector(assets Team) + StakedBalance(Earn Team) + Perps Balances(Perps Team)`

Extension : `TotalBalanceSelector = AggregatedBalanceSelector(assets Team)`

And then when we have these experiences on extension then you can just enhance the selectors on the UI.
This way , you don't have to deal with adding logic to asset controllers and can only focus on your balance and balance change sin your selectors.

nickewansmith:
The latest image shows the current state, we have removed the perps total from the wallet total for v1.

app/components/UI/Perps/components/PerpsCard/PerpsCard.tsx
```diff
@@ -54,14 +59,17 @@ const PerpsCard: React.FC<PerpsCardProps> = ({
...
     valueColor = pnlValue >= 0 ? TextColor.Success : TextColor.Error;
-    valueText = formatPnl(pnlValue);
-    const roeValue = parseFloat(position.returnOnEquity);
-    labelText = `${roeValue >= 0 ? '+' : ''}${roeValue.toFixed(2)}%`;
+    const notionalValue = BigNumber(position.size || '0')
+      .multipliedBy(position.entryPrice || '0')
+      .toString();
+    valueText = formatPrice(notionalValue, { maximumDecimals: 2 });
+    const roeValue = parseFloat(position.returnOnEquity) * 100;
+    labelText = `${formatPnl(pnlValue)} (${formatPercentage(roeValue, 1)})`;
```
gambinish, line 67: Nit: Non blocking, but it might be good to centralize these into a shared utility

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -1019,15 +1052,42 @@ export class PerpsController extends BaseController<
 
     try {
       const provider = this.getActiveProvider();
-      const accountState = await provider.getAccountState(params);
+
+      // Get both current account state and historical portfolio data
+      const [accountState, historicalPortfolio] = await Promise.all([
```
gambinish, line 1057: Would there be any benefit to using `Promise.allSettled` here?

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -1361,6 +1454,31 @@ export class PerpsController extends BaseController<
...
+        totalValue: accountState.totalValue || '0',
+        unrealizedPnl: accountState.unrealizedPnl || '0',
+        accountValue1dAgo:
+          accountValue1dAgo ||
+          existingBalance?.accountValue1dAgo ||
+          accountState.totalValue ||
+          '0',
+        lastUpdated: Date.now(),
+      };
+    });
```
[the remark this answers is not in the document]
gambinish: I think I may have run into this bug actually. When switching between testnet and mainnet I would often get an infinite loader, which could have been from corrupted/stale state.

I don't think you introduced this in your PR, so this doesn't need to block. But maybe this is something we should keep an eye on.

app/util/test/initial-background-state.json
```diff
@@ -391,7 +391,8 @@
     "hasPlacedFirstOrder": {
       "testnet": false,
       "mainnet": false
-    }
+    },
+    "perpsBalances": {}
```
gambinish, line 395: 🧠 ⛈️ : I wonder if we can use this and expand on it to support perps balances for each account?

The way I imagined it working, was to have some controller method do the following:

1. Keep track of which account groups have an EVM account that has gone through Perps onboarding
2. Keep track of each perp balance for each account group that has gone through onboarding

We might even need a separate controller for this, that's scoped only to persisting this data.

gambinish (approved):
✅ LGTM. I left a few comments, but mostly for discussion. We may want to start considering how the V2 balance updates will work across the app, and how they will integrate with bip44 workstream.

Also, we talked about this during standup, but the scrollable area on the Perps tab is pretty tiny, even on my Samsung Galaxy S23 Ultra:

https://github.com/user-attachments/assets/f8008f8b-e53e-491c-9ef4-6e68a397058a

I wonder if we can do some fancy UX here to make the pnl pill minimize while scrolling, or maybe have the pnl pill scroll with the positions

app/util/test/initial-background-state.json
nickewansmith: 1/2: Yes def something for discussion, but the onboarded yes/no flag seems like it could be app wide across all accounts. If a user onboards under any account in the app, they may not want/need to see that flow anymore. They could always view the tutorial again. But doing it how you mention would help us limit what accounts we are querying. I think any user could already have a balance though on any address, and then it's back to getting balances for all, which is prob not what we want to do.

app/components/UI/Perps/controllers/PerpsController.ts
nickewansmith: Yea I'm ok with either way, Promise.allSettled may end up being more verbose for not much gain

PR #18566, opened 2025-08-21
feat: use new bip44 token selectors

bergarces:
## **Description**

Use new BIP44 selectors to create an ordered list that contains all assets, as well as new list item component, based on the existing one, that uses the data from the new selector. 

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

  Scenario: user [verb for user action]
    Given [describe expected initial app state]

    When user [verb for user action]
    Then [describe expected outcome]
```

## **Pre-merge author checklist**

- [X] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [X] I've completed the PR template to the best of my ability
- [X] I’ve included tests if applicable
- [X] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [X] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #18733, opened 2025-08-25
feat: bip44 assets selector

bergarces:
## **Description**

Imports the selector from `@metamask/assets-controllers` so that it can be used by subsequent PRs by multiple teams. Including tokens list by assets and send flow by confirmations.

Preview WIP PR for tokens list using the same selector: https://github.com/MetaMask/metamask-mobile/pull/18566

I have verified that calling the selector returns the expected output.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```


app/selectors/tokenList.ts
```diff
@@ -72,3 +74,46 @@ export const selectSortedTokenKeys = createDeepEqualSelector(
...
+      ...(AccountTrackerController as {
+        accountsByChainId: Record<
+          Hex,
+          Record<
+            Hex,
+            {
+              balance: Hex | null;
+            }
+          >
+        >;
```
bergarces: The reason behind this assertion is that this field is needed in both clients, but `AccountTrackerController` has a different implementation in mobile and extension.

The type does not match because mobile version uses `string` instead of `Hex`, but the values are stored as `Hex`.

app/selectors/assets/assets-list.ts
```diff
@@ -0,0 +1,63 @@
...
+            Hex,
+            {
+              balance: Hex | null;
+            }
+          >
+        >;
+      }),
+    };
+  },
+  (filteredState) => _selectAssetsBySelectedAccountGroup(filteredState),
```
bergarces, line 62: There's already an acknowledgement that we need to measure performance for these new selectors. They will be audited soon.

https://github.com/MetaMask/metamask-mobile/pull/18315/files#r2294901635

app/selectors/assets/assets-list.ts
```diff
@@ -0,0 +1,63 @@
...
+    } = state.engine.backgroundState;
+
+    let multichainState = {
+      accountsAssets: {},
+      assetsMetadata: {},
+      balances: {},
+      conversionRates: {},
+    };
+
+    ///: BEGIN:ONLY_INCLUDE_IF(keyring-snaps)
```
owencraston, line 32: are the code fences necessary here?

owencraston (approved):

app/selectors/assets/assets-list.ts
bergarces: If those controllers are not initialised and the state is undefined, the app crashes, as the function is expecting that state to exist.

The initialisation of those controllers is behind the same code fence, so I didn't want to risk it and added a fallback with an empty state that prevents the app from crashing.

sethkfman (approved):
LGTM

PR #18763, opened 2025-08-26
feat(perps): update Perps withdrawal screen and remove Hyperliquid network

abretonc7s:
## **Description**

This PR implements a comprehensive redesign of the Perps withdrawal interface, addressing critical stability issues and enhancing user experience through modern UI patterns and improved network handling.

**1. What is the reason for the change?**
- Poor UX: Outdated interface lacking modern input methods (numeric keypad, percentage buttons) made amount entry cumbersome
- Incorrect messaging: Toast notifications displayed wrong currency (ETH instead of USDC), confusing users about their withdrawals
- Technical debt: Hyperliquid network references remained in the codebase despite migration to new infrastructure

**2. What is the improvement/solution?**
- Completely redesigned withdrawal UI with Tailwind CSS featuring large amount display, percentage buttons (10%, 25%, 50%, Max), and integrated numeric keypad
- Fixed toast notifications to correctly show USDC amounts and accurate timing ("Processing withdrawal" → "Withdrawal successful - X USDC will arrive within 5 minutes")
- Removed all Hyperliquid references, displaying "Perps Account" instead for cleaner branding

## **Changelog**

CHANGELOG entry: Updated Perps withdrawal screen with new design, fixed Arbitrum network validation crash, and improved withdrawal notifications to show correct USDC amounts

## **Related issues**

Fixes:
- [TAT-1379](https://consensyssoftware.atlassian.net/browse/TAT-1379): Remove Hyperliquid network from GNS
- [TAT-1396](https://consensyssoftware.atlassian.net/browse/TAT-1396): USDC logo doesn't load in withdrawal screen
- [TAT-1397](https://consensyssoftware.atlassian.net/browse/TAT-1397): Toast improvements for withdrawal notifications

## **Manual testing steps**

```gherkin
Feature: Perps Withdrawal Screen Update

  Scenario: User performs a withdrawal with new UI
    Given user is on the Perps main screen with a positive balance
    And Arbitrum network is either enabled or disabled initially
    
    When user taps on "Withdraw" button
    Then the new withdrawal screen appears with:
      - Large amount display at top
      - Available balance shown
      - Percentage buttons (10%, 25%, 50%, Max)
      - Numeric keypad for amount entry
      - USDC token display in receive section
      - Provider fee information
    
    When user enters an amount using the keypad or percentage buttons
    And taps "Withdraw"
    Then user is immediately navigated back to the previous screen
    And a toast appears saying "Processing withdrawal - This may take a few moments"
    And after withdrawal completes, a success toast appears saying "Withdrawal successful - X USDC will arrive within 5 minutes"

  Scenario: Arbitrum network auto-enablement
    Given user has Arbitrum network disabled
    
    When user enters the Perps environment
    Then Arbitrum network is automatically enabled
    And no crashes occur during deposit or withdrawal operations

  Scenario: Minimum amount validation
    Given user is on the withdrawal screen
    
    When user enters an amount below the minimum ($1.01)
    Then an error message appears: "Minimum withdrawal amount is $1.01"
    And the Continue button remains disabled
```

### **Before**
- Old withdrawal screen with Hyperliquid network display
- No percentage buttons or proper keypad
- Incorrect toast messages showing ETH instead of USDC
- App crashes when Arbitrum not enabled

### **After**
- New modern dark theme withdrawal UI matching design mockups
- Clean interface with amount display, percentage buttons, and integrated keypad
- Proper USDC withdrawal toasts
- Arbitrum auto-enablement prevents crashes
- "Perps Account" shown instead of Hyperliquid

https://github.com/user-attachments/assets/846aba53-646b-49eb-9b34-958d661c9e54

[TAT-1361]: https://consensyssoftware.atlassian.net/browse/TAT-1361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TAT-1379]: https://consensyssoftware.atlassian.net/browse/TAT-1379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -880,8 +936,20 @@ export class PerpsController extends BaseController<
...
+          error: errorMessage,
+        };
       });
 
+      // Clear withdrawal result after delay
+      setTimeout(() => {
+        this.update((state) => {
+          state.lastWithdrawResult = null;
+        });
+      }, 100);
```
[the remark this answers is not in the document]
abretonc7s: wanted, this is just long enough to update the state and show notification.

Matt561:
@abretonc7s Feedback after messing around with this branch on an Android simulator.

Functional Problems
1.  Withdraw button not disabled and no error message if entered amount greater than available balance
2. Receive tooltip missing `perps.tooltips.receive.title` translation
3. Opening Provider fee tooltip **crashes the app**. Provider fee tooltip requires context (temp solution to improve after beta)
4. Nit: For decimal number input, when clicking backspace keypad button (bottom-right) the decimal point is included in the removed characters but isn't visible which means the whole number is displayed but the user needs to press backspace again to clear the whole number.
	1. E.g. Enter 5.12 using the keypad
	2. Click the keypad's backspace twice. You will see `5` but you need to click backspace **again** to clear the remaining whole `5` (whole number)
5. Nit: If you enter a decimal number with trailing zeroes (e.g. 5.50005) the zeroes are automatically removed

Design Discrepancies
1. `Withdraw` button and amount input text should use theme to be either light (in dark mode) or dark (in light mode)
2. "Receive", "Provider fee", "You'll receive" should all be the same font-size, weight, and color (16px equivalent) and all left-aligned evenly
3. Icon in "Receive" section shouldn't have chip background
4. "Receive" USDC icon has `"?"` as the network icon (bottom-left)
5. Receive, Provider fee, You'll receive tooltips one size too small 
7. Withdraw button border radius should be lower to match designs
8. Withdraw text inside button should be larger to match designs
9. Perp account balance flashes a red/orange background when updating

Here's the Figma file I'm looking at - https://www.figma.com/design/Jp8owZibGXg26diIjDjVuq/Mobile-Perps?node-id=4604-9401&p=f&m=dev

Reviewing the code now!

app/components/Nav/App/App.tsx
```diff
@@ -1271,6 +1272,14 @@ const App: React.FC = () => {
...
               ],
             });
+
+            // Also enable the network in NetworkEnablementController
+            // This ensures Arbitrum (and others) are enabled by default
+            const caipChainId = `eip155:${parseInt(
+              network.chainId,
+              16,
+            )}` as `${string}:${string}`;
+            NetworkEnablementController.enableNetwork(caipChainId);
```
Matt561: There may be existing utils to handle this:

- `toEvmCaipChainId` from `@metamask/multichain-network-controller`
- `toCaipChainId` from `@metamask/utils`

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -851,7 +894,20 @@ export class PerpsController extends BaseController<
       this.update((state) => {
         state.lastError = result.error || PERPS_ERROR_CODES.WITHDRAW_FAILED;
         state.lastUpdateTimestamp = Date.now();
+        state.withdrawInProgress = false;
+        state.lastWithdrawResult = {
+          success: false,
+          error: result.error || PERPS_ERROR_CODES.WITHDRAW_FAILED,
```
Matt561, line 932: Will `result.error` be exposed to the end user? If yes, we should ensure this value is localized and added to the translation mapping you added previously.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -851,7 +894,20 @@ export class PerpsController extends BaseController<
...
+        };
       });
+
+      // Clear withdrawal result after delay
+      setTimeout(() => {
+        this.update((state) => {
+          state.lastWithdrawResult = null;
+        });
+      }, 100);
```
Matt561: If I'm understanding correctly, this clears automatically only when the withdrawal fails. Should we let the UI handle clearing? I don't have a strong opinion on this just thinking out loud. I'm wondering if we want an error toast to be displayed in the UI? If yes, could this introduce a race condition where the UI is slow to render the toast, the PerpsController clears, and the toast is rendered with no error message (empty)?

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -884,8 +940,20 @@ export class PerpsController extends BaseController<
       this.update((state) => {
         state.lastError = errorMessage;
         state.lastUpdateTimestamp = Date.now();
+        state.withdrawInProgress = false;
+        state.lastWithdrawResult = {
+          success: false,
+          error: errorMessage,
```
Matt561, line 971: Same thing here about whether we display this error the the end user ever? If yes, it'll need to be localized.

app/components/UI/Perps/hooks/usePerpsNetworkValidation.ts
```diff
@@ -0,0 +1,95 @@
...
+        // Determine which Arbitrum network to check based on environment
+        const chainId = isTestnet
+          ? ARBITRUM_TESTNET_CHAIN_ID
+          : ARBITRUM_MAINNET_CHAIN_ID;
+
+        // Convert to hex format for chain ID
+        const hexChainId = toHex(parseInt(chainId, 10));
+
+        // Create CAIP chain ID for network enablement check
+        const caipChainId = `eip155:${parseInt(chainId, 10)}` as CaipChainId;
```
Matt561: Might be able to reuse a util for this (mentioned in earlier message)

app/components/UI/Perps/hooks/usePerpsWithdrawStatus.ts
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      // Show error toast
+      toastRef?.current?.showToast({
+        variant: ToastVariants.Icon,
+        iconName: IconName.Danger,
+        hasNoTimeout: false,
+        labelOptions: [
+          {
+            label: `${strings('perps.withdrawal.error')}\n${
+              lastWithdrawResult.error ||
```
Matt561, line 98: Ah looks like we **do** display the error. We'll want to make sure these are translated correctly.

app/components/UI/Perps/controllers/PerpsController.ts
abretonc7s: currently not exposed but we can have on sentry

abretonc7s (commented):
reviewed

app/components/UI/Perps/hooks/usePerpsNetworkValidation.ts
```diff
@@ -0,0 +1,95 @@
...
+          {
+            error: error instanceof Error ? error.message : 'Unknown error',
+            isTestnet,
+          },
+        );
+      }
+    };
+
+    validateAndEnableArbitrum();
+  }, [enableNetwork, isNetworkEnabled, isTestnet, perpsNetwork]);
```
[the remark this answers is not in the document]
abretonc7s: not a bug

gambinish (dismissed):

app/components/Nav/App/App.tsx
```diff
@@ -1271,6 +1273,11 @@ const App: React.FC = () => {
                 },
               ],
             });
+
+            // Also enable the network in NetworkEnablementController
+            // This ensures Arbitrum (and others) are enabled by default
+            const caipChainId = toEvmCaipChainId(network.chainId);
+            NetworkEnablementController.enableNetwork(caipChainId);
```
gambinish: This seems very dangerous to me, and can lead to unintended behaviors. Are we essentially enforcing that all EVM networks get enabled by default?

I think that this has the potential to break other workstreams, especially with the `REMOVE_GNS` work that is slated for assets/ux. And potentially also with dApp connectivity.

Can you help me understand why this change is necessary? If it's absolutely necessary, we should make sure that we run e2e tests, and ensure that they pass since this impacts features outside of the Perps domain.

app/components/UI/Perps/hooks/usePerpsNetworkValidation.ts
```diff
@@ -0,0 +1,95 @@
...
+  ARBITRUM_TESTNET_CHAIN_ID,
+} from '../constants/hyperLiquidConfig';
+import { usePerpsNetwork } from './usePerpsNetwork';
+
+/**
+ * Hook to ensure Arbitrum network is enabled for perps operations
+ * This prevents crashes when trying to deposit/withdraw when Arbitrum is not enabled
+ *
+ * @returns Object with validation state
+ */
```
gambinish: I feel that a better UX is to have the user add Arbitrum as a network when going through Perps onboading.

The issue that we are trying to solve with this hook is enabling Arbitrum to allow for withdraw flows on that chain. This can be done directly as a call to `enableNetwork(caipChainId)`

However, if the user doesn't already have Arbitrum added, it cannot be enabled. And I do not think that enforcing all default networks be added and enabled is a solution we can go with for V1.

Happy to sync with you on this on a call, but we may need to rethink the approach here.

app/components/UI/Perps/integration/connectionLifecycle.test.tsx
```diff
@@ -190,6 +227,51 @@ describe('Connection Lifecycle Integration Tests', () => {
...
+                  BackgroundTimer.stop();
+                }
+              };
+
+              const subscription = AppState.addEventListener(
+                'change',
+                handleAppStateChange,
+              );
+              return () => subscription?.remove();
+            }, [onDisconnect]);
```
gambinish, line 269: I wonder if this can be broken out into a test util

gambinish (changes requested):
After reviewing this more carefully, I think that the changes introduced in this PR have the potential to break other core workstreams in the app. In particular, the remove GNS work that has been going on. If we are dealing with adding networks globally, we absolutely should be running smoke tests here.

I do not think that we can arbitrarily add/enable all default networks for all users, just so that we can support Perps withdrawals on Arbitrum.

Let's chat about the approach. Unless I've misunderstood, we may need to change the way we are handling this.

app/components/Nav/App/App.tsx
Matt561: cc: @abretonc7s

app/components/UI/Perps/hooks/usePerpsNetworkValidation.ts
Matt561: Please invite me to the sync if you do end up meeting. I want to stay informed 👍

abretonc7s:
> After reviewing this more carefully, I think that the changes introduced in this PR have the potential to break other core workstreams in the app. In particular, the remove GNS work that has been going on. If we are dealing with adding networks globally, we absolutely should be running smoke tests here.
> 
> I do not think that we can arbitrarily add/enable all default networks for all users, just so that we can support Perps withdrawals on Arbitrum.
> 
> Let's chat about the approach. Unless I've misunderstood, we may need to change the way we are handling this.

Good point this has been removed to separate PR

gambinish:
>For decimal number input, when clicking backspace keypad button (bottom-right) the decimal point is included in the removed characters but isn't visible which means the whole number is displayed but the user needs to press backspace again to clear the whole number.

I'm still noticing this bug, maybe this is a quick fix?
https://github.com/user-attachments/assets/c29c571c-71e1-48ab-b23d-814368ef526a

app/components/UI/Perps/integration/connectionLifecycle.test.tsx
abretonc7s: Potentially but I would prefer to do this separately

gambinish (dismissed):
Thanks for making those changes. This looks much better, and I'm happy we were able to catch the withdraw regression before this hit main.

I still think https://github.com/MetaMask/metamask-mobile/pull/18763#discussion_r2305201785 is worth discussing further, but we don't need to block the PR for this.

gambinish (approved):

PR #18824, opened 2025-08-27
chore: migrate swaps to BIP-44

bfullam:
## **Description**
Integrating the new BIP-44 account selectors into the Swaps/Bridge experience.

### Added

- **Source account selector**:  a new `selectSourceWalletAddress` selector has been added to the bridge store, which wraps `selectSelectedInternalAccountByScope`.
- **Destination account selector component**: filter by scope instead of explicit Solana checks.
- **Display the AccountGroup's name instead of the individual account's name**: check new selector `selectAccountGroups` for group name.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: BIP-44 selectors

  Scenario: user completes EVM <> NON-EVM swap
    Given user has an asset to swap

    When user inputs swap parameters
    Then proper destination account options are available and swap completes as expected
```


app/components/hooks/useAccounts/useAccounts.ts
```diff
@@ -166,10 +172,18 @@ const useAccounts = ({
           balanceError: undefined,
         };
 
+        let accountName = internalAccount.metadata.name;
+        if (isMultichainAccountsState2Enabled) {
+          accountName =
+            accountGroups.find(({ accounts: groupAccounts }) =>
```
owencraston: I really want to avoid modifying the useAccounts hook since its already so dam slow and we have had countless performance issues with it. This hook is used a lot too so the changes here have wide spread effects. Let me see if I can find a better way.

owencraston:
I've made some very crude modifications on this [draft PR](https://github.com/MetaMask/metamask-mobile/pull/18909) take a look and let me know what you think. My main goal was to remove the added logic from useAccounts.

app/components/UI/Bridge/components/DestinationAccountSelector.tsx/index.tsx
```diff
@@ -57,6 +61,21 @@ const DestinationAccountSelector = () => {
...
+    if (!destToken) return [];
+
+    const destCaipChainId = formatChainIdToCaip(destToken.chainId);
+    const isEvmChain = !isNonEvmChainId(destCaipChainId);
+
+    return accounts.filter(
+      (account) =>
+        account.scopes?.includes(destCaipChainId) ||
+        (isEvmChain && account.scopes?.includes(EthScope.Eoa)),
+    );
```
ccharly: We should use `isScopeEqual` and `isScopeEqualToAny` utils for those (if we decide to keep the filtering here of course):

Reference is here:
- https://github.com/MetaMask/accounts/blob/main/packages/keyring-utils/src/scopes.ts

app/components/hooks/useAccounts/useAccounts.ts
ccharly: Also, why would we need to rely on `useAccounts` anymore? 🤔

We're shifting the entire account-model from singular blockchain accounts to "account groups", so I'd say the components that are being migrated to "multichain accounts" have to use new selectors and not rely on the existing selectors/hooks for the old account model?

I do miss some context here maybe, but we should not "fix"/"alter" the old behaviors, we need to re-write/migrate the components that have to be compatible with "multichain accounts"/BIP-44.

app/selectors/bridge.ts
```diff
@@ -0,0 +1,51 @@
...
+  [(state: RootState) => state, selectDestToken],
+  (state, destToken): Set<AccountId> => {
+    if (!destToken) return new Set<AccountId>();
+
+    const destScope = formatChainIdToCaip(destToken.chainId);
+    const isEvm = destScope.startsWith('eip155:');
+
+    const byDestScope = selectInternalAccountsByScope(state, destScope);
+    const evmWildcard = isEvm
+      ? selectInternalAccountsByScope(state, 'eip155:0')
```
owencraston: nit: we should use EthScope.EOA here.

owencraston (commented):
With BIP44 all account should support solana so when I perform a swap while my current selected account is account three, I would expect the default swap/bridge destination account to be my current selected account. Here you can see in my screenshots, the default destination is not my current selected account. Another weird thing is that there are only 3 valid destination addresses. I think all account should support solana but I think these are missing due to my team not finishing the account reconciliation yet. 

Simulator Screenshot - iPhone 16 Plus - 2025-08-29 at 13 49 39

Simulator Screenshot - iPhone 16 Plus - 2025-08-29 at 13 49 32

Simulator Screenshot - iPhone 16 Plus - 2025-08-29 at 13 50 01

Simulator Screenshot - iPhone 16 Plus - 2025-08-29 at 13 50 14

app/selectors/bridge.ts
bfullam: Done

owencraston (approved):

app/selectors/bridge.ts
```diff
@@ -0,0 +1,52 @@
+import { createSelector } from 'reselect';
```
infiniteflower, line 1: I'm fine with keeping this, but we should eventually migrate the `app/core/redux/slices/bridge/index.ts` stuff into here or some other folder under the Bridge folder.

infiniteflower (approved):

PR #18910, opened 2025-08-29
chore: cp-7.54.1 add logging to get more info around blocking login biometrics 

tommasini:
## **Description**
* add logging to app trigger auth when biometrics can not get password from keychain
* check when migration happens if we have the vault in keyring controller when existing user is true 

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```


weitingsun (dismissed):

app/store/persistConfig.ts
```diff
@@ -18,6 +18,28 @@ const MigratedStorage = {
...
       if (res) {
+        try {
+          const state = JSON.parse(res);
+          const user =
+            typeof state?.user === 'string'
+              ? JSON.parse(state.user)
+              : state?.user;
+          const engine =
+            typeof state?.engine === 'string'
+              ? JSON.parse(state.engine)
```
joaoloureirop: This looks like an expensive operation to do. 
Do we need to hold getItem until the parsing finishes, or could we make the logging async (from `getIem` perspective)?

app/components/Nav/App/App.tsx
```diff
@@ -1116,6 +1116,13 @@ const App: React.FC = () => {
           errorMessage,
           `Unlock attempts: 1`,
         );
+        if (locked) {
+          Logger.error(
+            new Error(errorMessage),
```
joaoloureirop, line 1102: does `AUTHENTICATION_APP_TRIGGERED_AUTH_NO_CREDENTIALS` value serve to identify this specific error?

app/store/persistConfig.ts
tommasini: This happens only when we fresh open the app. I've measured it and it didn't seem a burden, but I had almost no data in the app.

This happens at the rehidration process inside of redux persist as far as I know it. 

The way I see it on perhaps checking this behavior is instead of here, at Engine initialization at engine service 🤔🤔

That actually seems a better idea!

app/components/Nav/App/App.tsx
tommasini: Yess! It's saying that was not able to get it from the keychain that it is what we are looking to see if it happens 🙏

zone-live (dismissed):

app/store/persistConfig.ts
tommasini: Done

app/store/migrations/util/index.test.ts
```diff
@@ -28,4 +43,278 @@ describe('ensureValidState', () => {
...
+          },
+        },
+      },
+      user: {
+        existingUser: true,
+      },
+      ...overrides,
+    });
+
+    it('should log vault existence when existingUser is true and vault exists', () => {
```
joaoloureirop: update tests description to follow contributor docs

> Do not use "should" at the beginning of the test name. 

[source](https://github.com/MetaMask/contributor-docs/blob/main/docs/testing/unit-testing.md#use-it-to-specify-the-desired-behavior-for-the-code-under-test)

app/store/migrations/util/index.tsx
```diff
@@ -40,6 +41,40 @@ export function ensureValidState<T>(
...
+        const vault = state.engine.backgroundState.KeyringController.vault;
+        hasVault = Boolean(vault);
+      }
+
+      Logger.log(
+        'Is vault defined at KeyringController at migration when existingUser',
+        hasVault,
+      );
+    } catch (error) {
+      captureException(
```
joaoloureirop: We should use `Logger.error` instead

joaoloureirop (changes requested):

app/store/migrations/util/index.test.ts
tommasini: Addressed [here](https://github.com/MetaMask/metamask-mobile/pull/18910/commits/1bb8873f878a75111e71887351a7f0709efe82c1)

app/store/migrations/util/index.tsx
tommasini: Addressed [here](https://github.com/MetaMask/metamask-mobile/pull/18910/commits/1bb8873f878a75111e71887351a7f0709efe82c1)

joaoloureirop (approved):

PR #18949, opened 2025-08-29
chore: improve Engine tests stability

Nodonisko:
## **Description**

Previously only some tests called `destroyEngine` method after each test and that caused that tests actually relied on running in order etc. Now each tests destroys engine instance so it's much more stable.

I still have one more failing test but maybe I should just update that fixture? I am not sure.

Screenshot 2025-08-29 at 15 15 42

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### DIFF
--- a/app/core/Engine/Engine.test.ts
+++ b/app/core/Engine/Engine.test.ts
@@ -190,6 +190,12 @@ describe('Engine', () => {
     // Create expected state by merging the static fixture with current AppMetadataController state
     const expectedState = {
       ...backgroundState,
+      AccountTrackerController: {
+        ...backgroundState.AccountTrackerController,
+        // This is just hotfix, because it should not be empty but it reflects current state of Engine code
+        // More info: https://github.com/MetaMask/metamask-mobile/pull/18949
+        accountsByChainId: {},
+      },
       AppMetadataController: {
         currentAppVersion,
         previousAppVersion: '', // This will be managed by the controller
--- a/app/core/Engine/Engine.test.ts
+++ b/app/core/Engine/Engine.test.ts
@@ -82,9 +82,10 @@ describe('Engine', () => {
     ReduxService.store = configureStore({});
   });
 
-  afterEach(() => {
+  afterEach(async () => {
     jest.restoreAllMocks();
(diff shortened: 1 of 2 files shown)


tommasini:
hey @Nodonisko I think you've just found a bug, that[ initializing the AccountTrackerController with an empty state](https://github.com/MetaMask/metamask-mobile/blob/a5f6e82a75cc853431acd2d64a278062d043ca90/app/core/Engine/Engine.ts#L898), it's not allowing it to initialize as it intends to do IMO. I will double check that, but doesn't make sense to me the default state be an empty object instead of getting the network from Network Controller

Nodonisko:
@tommasini Thanks for feedback, so correct fix would to put some default network there instead of empty object?

tommasini:
@Nodonisko The team responsible for addressing the issue was notified, feel free to address the test to the expected behaviour for now `accountsByChainId: {}`!
Thanks Nodonisko!

cc: @salimtb

Cal-L:
@Nodonisko Thanks for looking into this. Just bouncing thoughts on your question about the failing unit test for non matching AccountTrackerController data - I'm looking at the controller repo [code](https://github.com/MetaMask/core/blob/main/packages/assets-controllers/src/AccountTrackerController.ts#L267) and by default, the state should resemble the expected state that the test is showing. I would consider this a real test failure if that is being overridden to an empty object. @tommasini It looks like that data is just being wiped right?

tommasini:
Yeah, it seems that when I'm running the unit test on latest main right now, locally, I have the same error. I've talked with @salimtb  and @salimtb confirmed that the initialization of AccountsTrackerController is wrong, and will fix it. The issue is [here](https://github.com/MetaMask/metamask-mobile/blob/a5f6e82a75cc853431acd2d64a278062d043ca90/app/core/Engine/Engine.ts#L899)

I was suggesting to change the unit test since it's actually showing the actual result that is happening right now in Engine, because we are overriding if initial state is undefined

Nodonisko:
@tommasini I hotfixed that failing test

PR #18958, opened 2025-08-29
fix: prevent calling wallet_addEthereumChain only on persistent chains

baptiste-marchand:
## **Description**

`wallet_addEthereumChain` was throwing when trying to add any "default chains", in the sense of chains that are present by default when installing MetaMask Mobile. These default chains, like Base, can however be removed by users, so `wallet_addEthereumChain` shouldn't throw.

~This PR updates `wallet_addEthereumChain` to fail only for chains that can't be removed by users, like Ethereum Mainnet, Linea Mainnet and default testnets.~

This PR removes the out of date conditional where `wallet_addEthereumChain` can't be called with some chainIds. This API can be called with any `chainId` and will prompt the user to update the default `rpcUrl` for an enabled networks (including "default networks")

## **Changelog**

CHANGELOG entry: `wallet_addEthereumChain` no longer throws an error when called with "default network" chainIds. This API can be called with any `chainId` and will prompt the user to update the default `rpcUrl` for an enabled networks (including "default networks").

## **Related issues**

Fixes: [WAPI-408](https://consensyssoftware.atlassian.net/browse/WAPI-408)

## **Manual testing steps**

```gherkin
Feature: Add default chain

  Scenario: user tries to add a default chain previously removed
    Given a default chain (i.g. Base Mainnet) isn't part of the user's known networks

    When user visit an app that tries to switch to Base Mainnet
    Then it should correctly prompt to add Base Mainnet network
```

[WAPI-408]: https://consensyssoftware.atlassian.net/browse/WAPI-408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/core/RPCMethods/wallet_addEthereumChain.js
```diff
@@ -113,8 +113,9 @@ export const wallet_addEthereumChain = async ({
   };
 
   //TODO: Remove aurora from default chains in @metamask/controller-utils
-  const actualChains = { ...ChainId, aurora: undefined };
-  if (Object.values(actualChains).find((value) => value === chainId)) {
+  if (
+    Object.values(getPersistentNetworks()).find((value) => value === chainId)
+  ) {
```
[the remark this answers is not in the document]
wenfix: Think Cursor is correct in this comment, `chainId` will be a hex, so we never enter this condition.

Perhaps this check is no longer necessary? It seems better UX to allow calling `wallet_addEthereumChain` with any existing chain, which prompts the user to update its existing configuration.

Thoughts?

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -482,7 +483,12 @@ const NetworkSelector = () => {
           }
           buttonProps={{
             onButtonClick: () => {
-              openModal(chainId, false, LINEA_MAINNET, true);
+              openModal(
+                chainId,
+                !isPersistentNetwork(LINEA_MAINNET),
```
wenfix: Probably not related to this PR but the Edit button still shows up for both Linea and Mainnet regardless of how I flip this flag. Is this the intended behavior or am I missing something here?

wenfix:
Left a couple minor comments but overall looks good @baptiste-marchand

app/util/networks/index.js
```diff
@@ -208,6 +208,24 @@ export const getTestNetworks = () =>
 export const isMainnetNetwork = (networkType) =>
   getMainnetNetworks().includes(networkType);
 
+/**
+ * Returns a list of networks that can't be removed by users.
+ *
+ * @returns {string[]} - List of networks that can't be removed by users.
+ */
+export const getPersistentNetworks = () => [
+  ...getAllNetworks().slice(0, 2),
```
jiexi: does this cut Mainnet and Linea Mainnet off the list?

app/core/RPCMethods/wallet_addEthereumChain.js
adonesky1: I'm very unclear on why we do this. I think this may be very outdated logic since we should enable this call to go through and update the rpcUrl for "persistent"/"default" networks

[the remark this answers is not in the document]
adonesky1: Yes exactly what @wenfix said.

1) the types here are are in fact mismatched. I just ran into this while testing and adding logs. This condition will never hit because `getPersistentNetworks` returns a list of chainNames not chainIds
2) This condition should just be fully removed at this point which will have the same effect

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -415,7 +416,7 @@ const NetworkSelector = () => {
           buttonIcon={IconName.MoreVertical}
           buttonProps={{
             onButtonClick: () => {
-              openModal(chainId, false, MAINNET, true);
+              openModal(chainId, !isPersistentNetwork(MAINNET), MAINNET, true);
```
adonesky1: Can you explain what this change is about?

app/components/Views/NetworkSelector/NetworkSelector.tsx
```diff
@@ -425,7 +426,7 @@ const NetworkSelector = () => {
             })
           }
           onLongPress={() => {
-            openModal(chainId, false, MAINNET, true);
+            openModal(chainId, !isPersistentNetwork(MAINNET), MAINNET, true);
```
adonesky1: Can you explain what this change is about?

jiexi (dismissed):

jiexi (approved):

PR #19041, opened 2025-09-01
test: Add APIs used in Snaps tests to allowlist

FrederikBolding:
## **Description**

Fixes Snaps E2Es by adding missing mocks where applicable and adding some URLs to the allowlist that were not added when this was refactored. This makes all Snaps E2E tests pass again.

Also fixes a bug where Snap IDs would be converted to HTTPS URLs for phishing checks.

### DIFF
--- a/e2e/api-mocking/mock-e2e-allowlist.ts
+++ b/e2e/api-mocking/mock-e2e-allowlist.ts
@@ -50,5 +50,5 @@ export const ALLOWLISTED_URLS = [
   'https://client-config.api.cx.metamask.io/v1/flags?client=mobile&distribution=flask&environment=dev',
   'https://acl.execution.metamask.io/latest/registry.json',
   'https://acl.execution.metamask.io/latest/signature.json',
-  'https://signature-insights.api.cx.metamask.io/v1/signature?chainId=0x539',
+  'https://signature-insights.api.cx.metamask.io/v1/signature?chainId=0x1',
 ];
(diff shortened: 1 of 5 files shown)


Mrtenz (approved):

e2e/api-mocking/mock-e2e-allowlist.ts
```diff
@@ -47,4 +47,8 @@ export const ALLOWLISTED_URLS = [
   'https://nft.api.cx.metamask.io/collections?contract=0xb66a603f4cfe17e3d27b87a8bfcad319856518b8&chainId=1',
   'https://nft.api.cx.metamask.io/users/0x76cf1cdd1fcc252442b50d6e97207228aa4aefc3/tokens?chainIds=1&limit=50&includeTopBid=true&continuation=',
   'https://bridge.dev-api.cx.metamask.io/getTokens?chainId=1',
+  'https://client-config.api.cx.metamask.io/v1/flags?client=mobile&distribution=flask&environment=dev',
```
cmd-ob, line 50: Can we add a default mock for the flask remote feature flags please? We are trying to avoid adding new URLs to this as much as possible.

FrederikBolding: As far as I can tell we don't have one for main yet, once we do I think we can apply the same one to Flask.

cmd-ob (approved):

FrederikBolding:
Bitrise tests have already passed, skipping after rebase.

PR #19167, opened 2025-09-03
fix: false warnings for HyperEVM

Julink-eth:
## **Description**

Currently there are warnings when adding the HyperEVM network on the name/symbol because it uses the chain ID 999 which is the same as WAN CHAIN, this PR is intented to deal with those warnings as an exception specific to HyperEVM to remove those warnings.

## **Changelog**

CHANGELOG entry: Removed warnings when adding HyperEVM as a custom network

## **Related issues**

NA

## **Manual testing steps**

1. Go to https://chainlist.org/
2. Search for hyperevm
3. Click on connect wallet
4. It should ask you to approve adding the network with all the details about the network
5. No warnings should be shown 
6. Then go to the Networks menu
7. Click on the 3 dots menu next to HyperEvm and click on Edit
8. In the popup there should not be any warnings
9. Check the Tokens View with the network selected being HyperEVM
10. There should not be a warning next to the HYPE Token

## **Screenshots/Recordings**

NA

### **Before**

image

image

### **After**

image

image


app/util/networks/index.js
```diff
@@ -686,3 +696,20 @@ export const WHILELIST_NETWORK_NAME = {
...
+ * @param {string} symbol - The symbol to validate.
+ * @returns {boolean} - Whether the symbol is valid for the given chain ID.
+ */
+export const isValidSymbol = (chainId, symbol) => {
+  if (!chainId || !symbol) {
+    return false;
+  }
+
+  return WHITELIST_SYMBOL[chainId] === symbol;
+};
```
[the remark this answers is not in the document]
Julink-eth: this.state.chainId is always a hex string.

Battambang (dismissed):

vivek-consensys:
Tested with no warnings. 

iOS: 
image

image


Android:
image

image

jiexi:
In [app/core/RPCMethods/networkChecker.util.ts](https://github.com/MetaMask/metamask-mobile/blob/858c1e5af6cdaf8f10f89a08359112b13ed7ec0d/app/core/RPCMethods/networkChecker.util.ts#L35), `checkSafeNetwork` gets the safe chainlist from `https://chainid.network/chains.json`, which is technically a list of chains, meaning that it's already possible for multiple networks to exist on the same chainId, BUT we are only matching against the FIRST chain in the list via `.find()` a few lines below.

What are your thoughts on modifying `checkSafeNetwork` to append a constant of whitelisted networks to the `https://chainid.network/chains.json` result and then use `.some()` instead of `.find()` to ensure we check all possible matching chains?

There are other places that make a request to `https://chainid.network/chains.json`. I imagine it is important for Hyper EVM to also be included in those places too

Julink-eth:
> checkSafeNetwork

That's a good idea, that should work, I'll try this.

Julink-eth:
> > checkSafeNetwork
> 
> That's a good idea, that should work, I'll try this.

Actually I just tried but `some` returns a boolean when there is a match not the actual chain + for this to work I think we would have to do the check on more than just the chain ID (Or replace entirely the chain returned by the chainId website by the whitelisted one) because again, `999` is currently the chain ID of `wanchain` which is the chain that is being returned by https://chainid.network/chains.json

Let me know what you think

jiexi:
sorry to clarify, not only does `.find()` need to be replaced with `.some()`, but all the checks inside the `if (matchedChain) {` block also need to be included in body of the fn passed to `.some()`. This will mean it skip when it sees the `wanchain` because the name/rpc/token will not match, but will return true when it finds our manually inserted hyper evm entry. Does that make sense?

```
import axios from 'axios';
import { BannerAlertSeverity } from '../../component-library/components/Banners/Banner';
import { strings } from '../../../locales/i18n';
import { PopularList } from '../../util/networks/customNetworks';

import { toHex } from '@metamask/controller-utils';

export type Alert = {
  alertError: string;
  alertSeverity: BannerAlertSeverity;
  alertOrigin: string;
}

const findPopularNetwork = (rpcUrl: string, chainId: string) =>
  PopularList.some((network) => {
    const { origin } = new URL(network.rpcUrl);
    return origin === rpcUrl && network.chainId === chainId;
  });

const findPopularNetworkName = (name: string, chainId: string) =>
  PopularList.some(
    (network) =>
      network.nickname.toLowerCase() === name?.toLowerCase() &&
      network.chainId === chainId,
  );

const findPopularNetworkSymbol = (symbol: string, chainId: string) =>
  PopularList.some(
    (network) => network.ticker === symbol && network.chainId === chainId,
  );

const checkSafeNetwork = async (
  chainIdDecimal: string,
  rpcUrl: string,
  nickname: string,
  ticker: string,
): Promise<Alert[]> => {
  const EVM_NATIVE_TOKEN_DECIMALS = 18;

  const response = await axios.get('https://chainid.network/chains.json');
  const safeChainsList = response.data;

  const matchedChains = safeChainsList.filter(
    (chain: { chainId: number }) => chain.chainId.toString() === chainIdDecimal,
  );

  if (!matchedChains.length) {
    return [{
      alertError: strings('add_custom_network.unrecognized_chain_id'),
      alertSeverity: BannerAlertSeverity.Error,
      alertOrigin: 'unknown_chain',
    }]
  }

  const alertsForMatchedChains = matchedChains.map((matchedChain) => {
    const alerts = [];
    const { origin } = new URL(rpcUrl);
    if (
      !matchedChain.rpc
        ?.map((rpc: string) => new URL(rpc).origin)
        .includes(origin) &&
      !findPopularNetwork(origin, toHex(chainIdDecimal))
    ) {
      alerts.push({
        alertError: strings('add_custom_network.invalid_rpc_url'),
        alertSeverity: BannerAlertSeverity.Error,
        alertOrigin: 'rpc_url',
      });
    }
    if (matchedChain.nativeCurrency?.decimals !== EVM_NATIVE_TOKEN_DECIMALS) {
      alerts.push({
        alertError: strings('add_custom_network.invalid_chain_token_decimals'),
        alertSeverity: BannerAlertSeverity.Warning,
        alertOrigin: 'decimals',
      });
    }
    if (
      matchedChain.name?.toLowerCase() !== nickname?.toLowerCase() &&
      !findPopularNetworkName(nickname, toHex(chainIdDecimal))
    ) {
      alerts.push({
        alertError: strings('add_custom_network.unrecognized_chain_name'),
        alertSeverity: BannerAlertSeverity.Warning,
        alertOrigin: 'chain_name',
      });
    }
    if (
      matchedChain.nativeCurrency?.symbol !== ticker &&
      !findPopularNetworkSymbol(ticker, toHex(chainIdDecimal))
    ) {
      alerts.push({
        alertError: strings('add_custom_network.unrecognized_chain_ticker'),
        alertSeverity: BannerAlertSeverity.Warning,
        alertOrigin: 'chain_ticker',
      });
    }

    return alerts;
  })

  // less alerts means a better match
  const bestMatchedAlerts = alertsForMatchedChains.reduce((best, current) => {
    return current.length < best.length ? current : best;
  }, alertsForMatchedChains[0])

  return bestMatchedAlerts;
};

export default checkSafeNetwork;
```

something like this?

app/components/Views/Settings/NetworksSettings/NetworkSettings/index.test.tsx
```diff
@@ -1076,6 +1076,20 @@ describe('NetworkSettings', () => {
       expect(instance.state.warningSymbol).toBeUndefined(); // No warning for valid symbol
     });
 
+    it('should not validate the symbol when ticker is whitelisted for chainId', async () => {
```
jiexi: it does validate though, no? it just uses the new whitelist instead

Julink-eth: oh yeah the sentence is badly phrased. 
How about "should not show symbol warning for whitelisted ticker" ?

jiexi (approved):

vinnyhoward (approved):

adonesky1 (approved):
Ideally I'd love for us to find a more durable pattern than a whitelist but this looks fine for now. Do we need to do this on extension too? And if so should we have a consolidated list shared across both clients?

PR #19202, opened 2025-09-03
feat: add network connection banner with direct edit functionality

cryptodev-2s:
## **Description**

This PR introduces a new user experience for handling slow RPC connections by showing a non-blocking banner that allows users to directly edit the problematic RPC URL.

### 📊 Metrics & Analytics

The following metrics will be tracked to improve RPC reliability:
- **Banner Display Events**: When slow/unavailable RPC banner is shown.
- **User Interaction**: When users click "Edit RPC" button two different events one for slow and another for unavailable.

## **Changelog**

CHANGELOG entry:  **Slow RPC Connection Banner**:  Added Non-blocking banner appears after 5/30 seconds for slow/unavailable RPC connections, with direct "Edit RPC" functionality to quickly resolve connection issues.

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/17090

## **Manual testing steps**

```gherkin
Feature: my feature name

```

#### ***Scenario:*** RPC is slow after 5s, then becomes available after 30s, followed by clicking Update

https://github.com/user-attachments/assets/f776dff4-2daf-4809-9762-d6077ca227a6

#### ***Scenario:*** RPC is slow after 5s followed by clicking Update

https://github.com/user-attachments/assets/f09a7c3d-c9bd-444d-b831-130b7cdc7994


app/core/Engine/controllers/network-controller/monitor-network-initialization.ts
```diff
@@ -0,0 +1,125 @@
...
+        networkChainId as Hex,
+      );
+      if (networkClientId) {
+        lookupNetworkClientIds.push(networkClientId);
+        networkController.lookupNetwork(networkClientId);
+      }
+    }
+  }
+
+  // Show banner after 5 seconds for networks that aren't available
```
mcmire: Should we move this timer into the banner component? And should we create the Segment event from within the component instead of within the controller? That way all we are doing on the controller side is looking up networks. And we can dynamically start and stop the timer based on what `enabledNetworks` is and what `networksMetadata` is.

Then, the responsibility of this file changes, from monitoring the status of networks to just refreshing the metadata for the specified enabled networks (and the UI will do the monitoring).

What do you think about this approach?

app/core/Engine/controllers/network-controller/monitor-network-initialization.ts
```diff
@@ -0,0 +1,125 @@
...
+  networkController: NetworkController;
+  networkEnablementController: NetworkEnablementController;
+  trackEvent: (options: {
+    event: IMetaMetricsEvent | ITrackingEvent;
+    properties: JsonMap;
+  }) => void;
+}): void {
+  let hasShownBanner = false;
+
+  const networkEnablementControllerState = networkEnablementController.state;
```
mcmire: See here for how I am looking up enabled networks on the extension side: https://github.com/MetaMask/metamask-extension/pull/36045/files#diff-6fbff2cfe97ac01b77296ef2122c7e0a5b3ff6a84b584b4d1a87482f35eea3d6R1939

app/components/UI/SlowRpcConnectionBanner/SlowRpcConnectionBanner.tsx
mcmire, line 1: It looks like this is called "slow connection banner" but shouldn't this also apply to when the connection has failed as well? i.e. shouldn't we show a banner not only when the network is degraded but also unavailable?

app/components/UI/SlowRpcConnectionBanner/SlowRpcConnectionBanner.tsx
```diff
@@ -0,0 +1,112 @@
...
+ * Shows when any network takes more than 5 seconds to initialize.
+ */
+const SlowRpcConnectionBanner: React.FC = () => {
+  const navigation = useNavigation();
+  const dispatch = useDispatch();
+  const { trackEvent, createEventBuilder } = useMetrics();
+  const visible = useSelector(
+    (state: RootState) => state.modals.slowRpcConnectionBannerVisible,
+  );
+  const chainId = useSelector(
```
mcmire: Hmm, if we don't have a status in `networksMetadata` for when an endpoint becomes degraded I see why you'd need to persist whether the banner is open or closed in the app state. 

But I wonder if `state.modals` is the right place for this? Maybe we want to make a new bucket, such as `networkConnectionBanners`. So we could create `actions/networkConnectionBanners/index.js` and `reducers/networkConnectionBanners/index.js`.

cryptodev-2s: Thanks for catching this! It’s fixed now. I had originally placed the banner there, which was later removed, and forgot to move it.

app/core/Engine/controllers/network-controller/monitor-network-initialization.ts
cryptodev-2s: Yes, that’s better as you described. The change has been made.

app/core/Engine/Engine.ts
```diff
@@ -2251,6 +2251,30 @@ export class Engine {
...
+
+    await Promise.allSettled(
+      chainIds.map(async (networkChainId) => {
+        const networkClientId = NetworkController.findNetworkClientIdByChainId(
+          networkChainId as Hex,
+        );
+        return await NetworkController.lookupNetwork(networkClientId);
+      }),
+    );
+  }
```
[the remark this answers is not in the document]
mcmire: Technically true, but `findNetworkClientIdByChainId` should not return `undefined`. If it does, that's a bug. So I'm okay with this.

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        if (shouldShowBanner) {
+          dispatch(
+            showNetworkConnectionBanner({
+              chainId: firstUnavailableNetwork.chainId,
+              status: firstUnavailableNetwork.status,
+            }),
+          );
+
+          trackEvent(
```
mcmire: Since `visible`, `chainId`, and `status` come in through a selector, and they will get updated when we `dispatch`, what do you think about extracting the creation of the event to a separate `useEffect`? It would get called on the next re-render but I think that should be okay?

app/actions/networkConnectionBanners/index.ts
```diff
@@ -0,0 +1,63 @@
+import { Hex } from '@metamask/utils';
+import { Action } from 'redux';
+import { NetworkConnectionBannerStatus } from '../../components/UI/NetworkConnectionBanner/types';
+/**
+ * Different action types available for different RPC event flow
+ */
+export enum NetworkConnectionBannersActionType {
```
mcmire: Nit: Why make an enum if `ShowNetworkConnectionBannerAction` and `HideNetworkConnectionBannerAction` already have a `type`? It seems that we could get them from `ShowNetworkConnectionBannerAction['type']` and `HideNetworkConnectionBannerAction['type']`.

app/actions/networkConnectionBanners/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Action to hide the network connection banner
+ * No parameters needed - just hides the currently visible banner
+ */
+export interface HideNetworkConnectionBannerAction extends Action {
+  type: NetworkConnectionBannersActionType.HIDE_NETWORK_CONNECTION_BANNER;
+}
+
+export type NetworkConnectionBannersAction =
```
mcmire: Should this be:

``` suggestion
export type NetworkConnectionBannerAction =
```

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,198 @@
...
+              status: timeoutType,
+            };
+          }
+
+          hasUnavailableNetwork = true;
+          break; // Only show one banner at a time
+        }
+      }
+
+      if (hasUnavailableNetwork && firstUnavailableNetwork) {
```
mcmire: Do we need `hasUnavailableNetwork`? It seems like we can say:

``` suggestion
      if (firstUnavailableNetwork !== null) {

File: app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.test.tsx [hunk 9]
```diff
@@ -0,0 +1,585 @@
...
+  NetworkConfiguration,
+  RpcEndpointType,
+} from '@metamask/network-controller';
+
+import NetworkConnectionBanner from './NetworkConnectionBanner';
+import { useNetworkConnectionBanners } from '../../hooks/useNetworkConnectionBanners';
+import { strings } from '../../../../locales/i18n';
+
+jest.mock('../../hooks/useNetworkConnectionBanners');
+jest.mock('../../../../locales/i18n', () => ({
```
mcmire: Seems strange that we have to mock I18n this way. I feel like we ought to be able to set i18n to English in unit tests and that we way we don't have to do this :(

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,198 @@
...
+  const navigation = useNavigation();
+  const { enabledNetworksByNamespace } = useNetworkEnablement();
+  const { trackEvent, createEventBuilder } = useMetrics();
+  const { visible, chainId, status } = useSelector(
+    selectNetworkConnectionBannersState,
+  );
+  const visibleRef = useRef(visible);
+
+  useEffect(() => {
+    visibleRef.current = visible;
```
mcmire: Why do we have to set `visible` in a ref? It doesn't seem that we update `visibleRef.current` in this file, we just read it, so why can't we use `visible` directly?

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,198 @@
...
+  }, [visible]);
+  const networkConfigurationByChainId = useSelector(
+    selectEvmNetworkConfigurationsByChainId,
+  );
+  const evmEnabledNetworksChainIds = useMemo(
```
mcmire: Could this be a selector instead of having to memoize this here?

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,198 @@
...
+    () =>
+      Object.entries(enabledNetworksByNamespace[KnownCaipNamespace.Eip155])
+        .filter(([, isEnabled]) => isEnabled)
+        .map(([networkChainId]) => networkChainId as Hex),
+    [enabledNetworksByNamespace],
+  );
+  const currentNetwork = useMemo(
```
mcmire: Does this need to be memoized?

app/reducers/networkConnectionBanners/index.ts
```diff
@@ -0,0 +1,58 @@
...
+import { NetworkConnectionBannerStatus } from '../../components/UI/NetworkConnectionBanner/types';
+
+/**
+ * Interface for defining what properties will be defined in store
+ */
+export interface NetworkConnectionBannersState {
+  visible: boolean;
+  chainId?: Hex;
+  status?: NetworkConnectionBannerStatus;
+}
```
mcmire: What are your thoughts on having this type be:

``` suggestion
export type NetworkConnectionBannersState =
  | {
      visible: false;
    }
  | {
      visible: true;
      chainId: Hex;
      status: NetworkConnectionBannerStatus;
    };
```

Having both `chainId` and `status` be optional like this feels a bit inaccurate because it means that `status` could be set while `chainId` is not — but that's impossible.

This way if you don't need to check whether `chainId` or `status` is present — they will be, if `visible` is true.

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
```diff
@@ -0,0 +1,198 @@
...
+
+      let hasUnavailableNetwork = false;
+      let firstUnavailableNetwork: {
+        chainId: Hex;
+        status: NetworkConnectionBannerStatus;
+      } | null = null;
+
+      for (const evmEnabledNetworkChainId of evmEnabledNetworksChainIds) {
+        const networkClientId =
+          Engine.context.NetworkController.findNetworkClientIdByChainId(
```
mcmire: Hmm... since we are in UI-land, and especially since we are in a hook, should we be accessing Engine directly?

It feels to me that at the top of this hook, outside of any `useEffect`, we could use selectors to get the first unavailable network, and then have this `useEffect` be dependent on that value (so if `firstUnavailableNetwork` changes, we reset existing timers and then, `firstUnavailableNetwork !== null`, we start the timers over).

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,198 @@
...
+  showNetworkConnectionBanner,
+} from '../../../actions/networkConnectionBanners';
+import { selectEvmNetworkConfigurationsByChainId } from '../../../selectors/networkController';
+import { useNetworkEnablement } from '../useNetworkEnablement/useNetworkEnablement';
+import { NetworkConnectionBannerStatus } from '../../UI/NetworkConnectionBanner/types';
+
+const SLOW_BANNER_TIMEOUT = 5 * 1000; // 5 seconds
+const UNAVAILABLE_BANNER_TIMEOUT = 30 * 1000; // 30 seconds
+
+const useNetworkConnectionBanners = (): {
```
mcmire: It seems that the only place this hook is used is in the component itself. Is there a reason why you extracted it to a separate file? Was it to make testing easier or were you thinking we would use this in the future?

mcmire (commented):
This looks pretty good. I have yet to try this out, I will do that tomorrow, but I just looked through the code for now. I feel like there are some things we can do more simply, but I'm not sure. I tried to make some suggestions, but if this works and you're happy with it, no worries.

app/actions/networkConnectionBanners/index.ts
cryptodev-2s: Yes good catch

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
cryptodev-2s: Indeed useless flag I forget to remove after adding firstUnavailableNetwork

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
cryptodev-2s: Yes good catch

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
cryptodev-2s: Mainly due to v2, as we could also consider using it in the future to display a balance status when certain networks are slow or unavailable...

app/actions/networkConnectionBanners/index.ts
cryptodev-2s: Both ShowNetworkConnectionBannerAction and HideNetworkConnectionBannerAction use this enum for their type property, and it’s also referenced in the reducer’s switch case. Are you suggesting we could just stick with string literals instead of an enum? I followed the same pattern I saw used in other actions.

Also, the name NetworkConnectionBannersActionType might be misleading since it’s not only a type definition, it’s also used to reference the values:

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
cryptodev-2s: Not needed at all

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
```diff
@@ -0,0 +1,185 @@
...
+          );
+        }
+      } else if (networkConnectionBannersState.visible) {
+        // Hide banner if all networks are available
+        dispatch(hideNetworkConnectionBanner());
+      }
+    };
+
+    // Set up slow banner timeout (5 seconds)
+    const slowTimeout = setTimeout(() => {
```
mcmire, line 171: I just tested this PR while using multiple enabled networks, and it seems like the timer is getting cleared and reset every 5 seconds. So I'm not sure that this is working correctly :( I'm looking into it now.

app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+const NetworkConnectionBanner: React.FC = () => {
+  const { networkConnectionBannersState, currentNetwork, editRpc } =
+    useNetworkConnectionBanners();
+
+  if (!networkConnectionBannersState.visible || !currentNetwork) {
+    return null;
+  }
+
+  return (
+    <Box twClassName="bg-warning-muted border-t border-warning-muted px-4 py-3">
```
mcmire: Hmm I thought we had a Banner component. Maybe that's only the case on Extension 🤔 

Since we are creating one from scratch should we honor the design? In particular I think the most important thing is we need to wrap on multiple lines if the text is too long, otherwise the message isn't really readable:

Screenshot 2025-09-23 at 9 59 18 AM

app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+          </Text>
+        </Box>
+
+        <Button
+          variant={ButtonVariant.Tertiary}
+          onPress={editRpc}
+          disabled={false}
+          twClassName="shrink-0"
+        >
+          {strings('network_connection_banner.edit_rpc')}
```
mcmire: Should this say "Update RPC" rather than "Edit RPC"? It seems that this is the case in the design.

app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.tsx
cryptodev-2s: Yes indeed I am working on the wrapping behavior

app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.tsx
cryptodev-2s: Nice Catch, indeed not sure why I used edit instead of update

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
cryptodev-2s: Yes indeed it might be related to the recent changes I made and this bug raised by cursor https://github.com/MetaMask/metamask-mobile/pull/19202#discussion_r2372561161

Thats what I currently checking its because of the dependency of networkConnectionState in useEffect

app/actions/networkConnectionBanners/index.ts
mcmire: Oh I see I didn't realize this was used in multiple places, I was thinking it was just used in this file. No worries about this, then.

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
mcmire: I know that `useEffect`s can be tricky so if this is too hard to figure out, no problem. At this point I don't think I have enough time to exactly match in Extension what you are doing in Mobile so we are free to diverge implementations.

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
```diff
@@ -0,0 +1,185 @@
...
+    // Set up unavailable banner timeout (30 seconds)
+    const unavailableTimeout = setTimeout(() => {
+      checkNetworkStatus('unavailable');
+    }, UNAVAILABLE_BANNER_TIMEOUT);
+
+    return () => {
+      clearTimeout(slowTimeout);
+      clearTimeout(unavailableTimeout);
+    };
+  }, [evmEnabledNetworksChainIds, dispatch, networkConnectionBannersState]);
```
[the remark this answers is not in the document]
mcmire: Ha, yes, that's what I've been experiencing too :(

app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.tsx
cryptodev-2s: Fixed in https://github.com/MetaMask/metamask-mobile/pull/19202/commits/8171ddc1e9673daa90de45ce435ddf4fe0278f04

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
cryptodev-2s: Fixed in https://github.com/MetaMask/metamask-mobile/pull/19202/commits/8171ddc1e9673daa90de45ce435ddf4fe0278f04

app/components/hooks/useNetworkConnectionBanners/useNetworkConnectionBanners.ts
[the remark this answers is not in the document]
cryptodev-2s: Fixed in https://github.com/MetaMask/metamask-mobile/pull/19202/commits/8171ddc1e9673daa90de45ce435ddf4fe0278f04

app/components/UI/NetworkConnectionBanner/NetworkConnectionBanner.tsx
cryptodev-2s: Fixed here: https://github.com/MetaMask/metamask-mobile/pull/19202/commits/a6622302d8da5dedb37d08c2ca871f2371459132  
Turns out there’s already a banner component; my bad, I hadn’t used it directly.

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
```diff
@@ -0,0 +1,190 @@
...
+    // Set up unavailable banner timeout (30 seconds)
+    const unavailableTimeout = setTimeout(() => {
+      checkNetworkStatus('unavailable');
+    }, UNAVAILABLE_BANNER_TIMEOUT);
+
+    return () => {
+      clearTimeout(slowTimeout);
+      clearTimeout(unavailableTimeout);
+    };
+  }, [evmEnabledNetworksChainIds, dispatch]);
```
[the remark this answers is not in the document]
mcmire: I don't think this is true. `selectEVMEnabledNetworks` is memoized. So this should work now. I will give this a shot.

Cal-L (commented):
Curious about the underlying behavior prior to the banner appearing. Since the banner only appears a little later, does that mean that the RPC endpoint is connected before? If not, is there a reason that the banner wouldn't appear as long as the RPC endpoint isn't connected?

ajeetd:
> Curious about the underlying behavior prior to the banner appearing. Since the banner only appears a little later, does that mean that the RPC endpoint is connected before? If not, is there a reason that the banner wouldn't appear as long as the RPC endpoint isn't connected?

It doesn't necessarily mean the RPC endpoint was connected before, but it's also possible there is no issue so would not want it to display until that decision is made. It appears once it's established there seems to be a problem (slow to connect, or after a longer period of time signaling giving up attempting - failed to connect).

mcmire:
@cryptodev-2s A couple things I'm noticing:

- I made the banner appear, and when it did, I clicked "Update RPC", which closed the banner — but then I did not change any endpoints. Because the network was still unavailable, I expected the banner to reappear when navigating away and then returning to the home screen, but it did not. I am okay with closing the banner when clicking "Update RPC", but I do feel like it should reappear. I'm not quite sure why that's happening, I am testing that now, but can you confirm?
- I'm curious whether we should extend the timeout for the "slow" version of the banner from 5 seconds to 10 or 15 seconds? It feels a bit fast to me. And similar for "unavailable" — should it be 40 or 45 seconds?

mcmire:
@cryptodev-2s Mentioned this on Slack, but I think any issues we see with banners not reappearing or disappearing when they should may partially be caused by the fact that we don't run the block tracker on Mobile, so we don't always have an accurate picture of the current state of the network. There's nothing we can do about that now, though, so I think this behavior is fine for now.

However, one bug I think we should fix now is that if you've selected one network and the banner comes up because the network is unavailable, but then you go to the network filter and switch the network, the banner doesn't go away immediately, it will go away when the "slow" timer for the newly selected network runs.

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    // Tracking the event
+    trackEvent(
+      createEventBuilder(
+        status === 'slow'
+          ? MetaMetricsEvents.SLOW_RPC_UPDATE_RPC_CLICKED
+          : MetaMetricsEvents.UNAVAILABLE_RPC_UPDATE_RPC_CLICKED,
+      )
+        .addProperties({
+          chain_id_caip: `eip155:${hexToNumber(chainId)}`,
```
mcmire, line 75: I think we should add the RPC URL to the metrics event as well, but we should only keep the hostname for known URLs and group all custom URLs in the same bucket, like we are doing here: https://github.com/MetaMask/metamask-mobile/blob/4a8664a2ee70b93690d6af670073abb0df6a8c71/app/core/Engine/controllers/network-controller/messenger-action-handlers.ts#L153

app/core/Analytics/MetaMetrics.events.ts
```diff
@@ -501,6 +501,12 @@ enum EVENT_NAME {
   RPC_SERVICE_UNAVAILABLE = 'RPC Service Unavailable',
   RPC_SERVICE_DEGRADED = 'RPC Service Degraded',
 
+  // NETWORK CONNECTION BANNER
+  SLOW_RPC_BANNER_SHOWN = 'Slow RPC Banner Shown',
+  UNAVAILABLE_RPC_BANNER_SHOWN = 'Unavailable RPC Banner Shown',
+  SLOW_RPC_UPDATE_RPC_CLICKED = 'Slow RPC Update RPC Clicked',
+  UNAVAILABLE_RPC_UPDATE_RPC_CLICKED = 'Unavailable RPC Update RPC Clicked',
```
mcmire: Wondering if we should tweak these names a bit?

``` suggestion
  SLOW_RPC_BANNER_UPDATE_RPC_CLICKED = 'Slow RPC Banner: Update RPC Clicked',
  UNAVAILABLE_RPC_BANNER_UPDATE_RPC_CLICKED = 'Unavailable RPC Banner: Update RPC Clicked',

File: app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts [hunk 23]
```diff
@@ -0,0 +1,217 @@
...
+            const rpcUrl =
+              networkConfig.rpcEndpoints[
+                networkConfig.defaultRpcEndpointIndex || 0
+              ]?.url || networkConfig.rpcEndpoints[0]?.url;
+
+            firstUnavailableNetwork = {
+              chainId: evmEnabledNetworkChainId,
+              status: timeoutType,
+              networkName: networkConfig.name,
+              rpcUrl,
```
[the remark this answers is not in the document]
mcmire: We really ought to have an RPC URL by now, I would be surprised if we didn't. If we discover a bug later we can fix it.

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
```diff
@@ -0,0 +1,202 @@
...
+    // Set up unavailable banner timeout (30 seconds)
+    const unavailableTimeout = setTimeout(() => {
+      checkNetworkStatus('unavailable');
+    }, UNAVAILABLE_BANNER_TIMEOUT);
+
+    return () => {
+      clearTimeout(slowTimeout);
+      clearTimeout(unavailableTimeout);
+    };
+  }, [evmEnabledNetworksChainIds, dispatch]);
```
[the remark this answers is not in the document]
mcmire: We really ought to have an RPC URL, I would be surprised if we didn't. I think this is a non-issue, for now.

mcmire (dismissed):
This LGTM! (edit: oh, I see that e2e tests are failing, darn)

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
NicolasMassart: I agree that it should be stripped of infos that could identify the user as connecting to a specific node (addresses as parame or token or anything)

But also consider the domain or IP itself could be sensitive info, consider using the addSensistiveProperties() function of the builder if we absolutely want to track it.

Note that the example you provide, even if it's interesting on how it strips sensitive infos to only keep a "custom" string, is really not something we should apply as is here as it's not using the recommended way to create events using the event builder.

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
```diff
@@ -0,0 +1,217 @@
...
+            ? MetaMetricsEvents.SLOW_RPC_BANNER_SHOWN
+            : MetaMetricsEvents.UNAVAILABLE_RPC_BANNER_SHOWN,
+        )
+          .addProperties({
+            chain_id_caip: `eip155:${hexToNumber(
+              networkConnectionBannerState.chainId,
+            )}`,
+            rpc_endpoint_url: sanitizeRpcUrl(
+              networkConnectionBannerState.rpcUrl,
+            ),
```
NicolasMassart, line 204: C: same as the previous call, suggestion to only keep non critical part of the url and make it a sensitive property

app/components/hooks/useNetworkConnectionBanner/useNetworkConnectionBanner.ts
mcmire: Hi @NicolasMassart, great points here.

This `sanitizeRpcUrl` function really does two things as you allude to: 

1. Only keeps the hostname of the URL and discards the rest, which reduces the chance of including API keys or other sensitive information in the URL.
2. Filters out unknown URLs, which reduces the chance that private servers (IP addresses or domain names) will be tracked. A "known" URL is either a Infura URL with our API key at the end of it, or a hardcoded URL for a chain that we promote via the network list (e.g. Polygon, ZKSync, etc.) In the future we plan on extending this filter to include all RPC endpoints listed in the Chainlist directory.

Because we only restrict tracked URLs to official and popular networks, at the moment, I don't think it counts as PII. We don't track addresses in Segment (I think? I hope?), so all we'd know is that user A is using Mainnet and user B is using Polygon. 

When we add Chainlist as an allowlist, there could be a small risk that the URL is PII, say, if someone hacks the Chainlist directory to include a private server that they've obtained via a private Discord server or Telegram or something. So perhaps we could consider making a change then, would you be okay with that?

As for how `sensitiveProperties` works, I was reading the docs and it seems that if you use this property, we will create two events instead of just one. Won't this change how we need to treat metrics? i.e. if we are seeing 100k instances of one event and then add a property via `sensitiveProperties`, won't we now see 200k instances of the same event going forward?

NicolasMassart: Yes, using sanitize is ok.

About the sensitive properties use, we already have a lot of sensitive properties tracked. The 2 events are expected and when you analyse metrics you have to split in anonymous events and regular events. In the end, the numbers should be the same for each kind.

NicolasMassart (commented):
good for me on the tracking events side

NicolasMassart (approved):
Approving only the trackEvent parts, as discussed with @cryptodev-2s

salimtb (approved):
Assets changes lgtm

salimtb (approved):

PR #19236, opened 2025-09-04
feat: automatically increase metamask pay buffer

matthewwalsh0:
## **Description**

When retrieving bridge quotes for MetaMask pay, automatically increase the source token amount by a given amount until the token amount has reached the required minimum, or the max attempts are reached.

Will also stop if the token balance is reached, and will only apply to the first bridge if multiple bridges are being requested.

Uses new MetaMask pay feature flags in `confirmation_pay`, specifically:

- `attemptsMax` - The total number of quotes requests, including the first request.
- `bufferInitial` - The initial buffer percentage added to all quotes requests. For example, `0.025` is `2.5%`.
- `bufferStep` - The percentage to increase the source amount by on each request. For example, `0.025` is `2.5%`.
- `slippageInitial` - The slippage used for the first bridge.
- `slippageSubsequent` - The slippage used for additional bridges.

## **Changelog**

CHANGELOG entry: null

### DIFF
--- a/app/components/Views/confirmations/utils/bridge.ts
+++ b/app/components/Views/confirmations/utils/bridge.ts
@@ -111,8 +111,16 @@ async function getSufficientSingleBridgeQuote(
         index,
       );
 
+      const dust = new BigNumber(result.quote.minDestTokenAmount)
+        .minus(request.targetAmountMinimum)
+        .toString(10);
+
       log('Found valid quote', {
         attempt: i + 1,
+        target: targetTokenAddress,
+        targetAmount: result.quote.minDestTokenAmount,
+        goalAmount: request.targetAmountMinimum,
+        dust,
         quote: result,
       });
 
@@ -253,7 +261,7 @@ function getBestQuote(
   ).slice(0, 3);
 
   const quotesOverMinimumTarget = fastestQuotes.filter((quote) =>
-    new BigNumber(quote.minToTokenAmount.amount).isGreaterThanOrEqualTo(
+    new BigNumber(quote.quote.minDestTokenAmount).isGreaterThanOrEqualTo(
       request.targetAmountMinimum,
     ),
   );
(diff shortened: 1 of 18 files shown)


app/components/Views/confirmations/utils/bridge.ts
```diff
@@ -161,22 +253,56 @@ async function getGasFeeEstimates(chainId: Hex) {
...
     'asc',
   )[0];
 }
+
+function getFinalRequests(
+  requests: BridgeQuoteRequest[],
+): BridgeQuoteRequest[] {
+  return requests.map((request, index) => {
+    const isFirstRequest = index === 0;
+    const attemptsMax = isFirstRequest ? request.attemptsMax : 1;
```
pedronfigueiredo, line 285: Why do we only attempt once on requests after the first one?

app/components/Views/confirmations/utils/bridge.ts
```diff
@@ -56,12 +70,101 @@ export async function getBridgeQuotes(
   }
 }
 
+async function getSufficientSingleBridgeQuote(
```
pedronfigueiredo, line 73: Would it make sense to unit test some of these utils such as `getSufficientSingleBridgeQuote` and `getFinalRequests`?

pedronfigueiredo (approved):

app/components/Views/confirmations/utils/bridge.ts
matthewwalsh0: We get the quotes simultaneously to save time, meaning we can't know how much of the balance allocation that each bridge is using.

We could do attempts on both if we did them sequentially, but that would slow down the process for the user, plus the second request currently is just for the gas and usually 1 dollar.

app/components/Views/confirmations/utils/bridge.ts
matthewwalsh0: These aren't exported, and both should be covered extensively by all the new unit tests.

PR #19270, opened 2025-09-04
fix: use hash as txHistory key and txMetaId for Solana bridges

micaelae:
## **Description**

This PR bumps the bridge-status-controller version which includes these fixes:
- Use correct segment event names
- Use tx hash as txHistory key and txMetaId for Solana bridges

The PR also includes a migration script that updates existing Solana transactions with the correct txHistory key and txMetaId, which is used by the status controller to fetch the latest bridge tx status

## **Changelog**

CHANGELOG entry: `fix: use hash as txHistory key and txMetaId for Solana bridges`

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/SWAPS-2945

## **Manual testing steps**

```gherkin
Feature: Solana tx status polling

  Scenario: user submits a new Solana bridge transaction

    When user submits a new Solana bridge tx
    Then it should show up as a success, including the dest tx hash, within a few minutes


  Scenario: user updates the wallet
    Given they have existing Solana bridge transactions which have been pending for a while

    When user gets the latest wallet update
    Then status polling should restart for old solana bridge txs, and their dest tx hash should be visible in the activity log details

```


SteP-n-s (dismissed):

app/store/migrations/099.test.ts
```diff
@@ -0,0 +1,564 @@
...
+    expect(
+      newState.engine.backgroundState.BridgeStatusController,
+    ).toStrictEqual({
+      txHistory: {
+        '4ad572c0-5817-11f0-9a9f-65cbc1ba7703': evmTxHistoryItem,
+        '5Cs9btTX3G7a4H4mmbuYLPpEzyyj2t64aSnCCMyy1AKQZvBrLprbmcT7TF2mcTQNKGik9JTLRkqHTtkKjHUUi4af':
+          {
+            ...solanaTxHistoryItem,
+            txMetaId:
+              '5Cs9btTX3G7a4H4mmbuYLPpEzyyj2t64aSnCCMyy1AKQZvBrLprbmcT7TF2mcTQNKGik9JTLRkqHTtkKjHUUi4af',
```
infiniteflower, line 476: Isn't this the tx hash? I guess I'm a little confused why it's in the txMetaId. I know Solana txs are a bit different from EVM, but if there's this inconsistency we should probably leave a comment in the migration and possibly outside of the migration as well.

micaelae: it is. we only do this for Solana transactions since they don't have a txMetaId

When polling restarts, the controller expects the key and txMetaId to be equal: https://github.com/MetaMask/core/blob/main/packages/bridge-status-controller/src/bridge-status-controller.ts#L417

Prior to this migration solana keys were a random uuid, which broke txHistory lookup by hash

infiniteflower (dismissed):

vinistevam (dismissed):

package.json
```diff
@@ -223,8 +223,8 @@
     "@metamask/assets-controllers": "^74.3.2",
     "@metamask/base-controller": "^8.3.0",
     "@metamask/bitcoin-wallet-snap": "^1.0.0",
-    "@metamask/bridge-controller": "^41.4.0",
-    "@metamask/bridge-status-controller": "^40.2.0",
+    "@metamask/bridge-controller": "^42.0.0",
```
tommasini, line 226: @micaelae this controller have a peer dependency of Accounts controller for v33
Let's ask if accounts team agree that it's not a blocker pelase before moving further

app/store/migrations/099.ts
```diff
@@ -0,0 +1,63 @@
+import { captureException } from '@sentry/react-native';
+import { ensureValidState } from './util';
+import { isObject } from '@metamask/utils';
+import { BridgeStatusControllerState } from '@metamask/bridge-status-controller';
+import { isSolanaChainId } from '@metamask/bridge-controller';
+
+/**
+ * Migration: Update bridge txHistory for solana to use txHash as key and txMetaId
+ */
+const migration = async (state: unknown): Promise<unknown> => {
```
tommasini: Is there any special reason for this migration be async?

micaelae: it doesn't need to be - updating this

package.json
micaelae: @tommasini the accounts-controller is at v33 on mobile now so we're not blocked on this peer dependency. I also verified that there were no mismatched peer dependencies related to the bridge controllers during my latest local build

tommasini (approved):

vinistevam (approved):

SteP-n-s (approved):

PR #19417, opened 2025-09-08
feat: perp push notification deeplinks

Prithpal-Sooriya:
## **Description**

Adds perp push notification deeplink support.
1. Adds the "Push Notification Click" Event along with the notification data blob.
2. Adds the push link navigation logic to `DeeplinkManager`
3. Adds an interstitial whitelist so we can have specific whitelisted URLs that don't show the scary interstitial modal.

DevLog:
https://www.loom.com/share/f24919f900a24bf89284b80d0b108c86?sid=eb163e28-270a-48fb-a968-0eba00fac14e

## **Changelog**

CHANGELOG entry: add push notification deeplink support

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/ASSETS-1103

## **Manual testing steps**

This is a little tricky to test as you need to generate a perps notification (stop-loss, take profit, order executed, liquidated).
However once you have a push notification, you can then click on a push notification and it should navigate you once you unlock the wallet.

### **After**

https://www.loom.com/share/45a61639bd1747afb3506c1012eebf2f?sid=61960064-aeb1-41b5-81f0-3cc671364ca1

https://www.loom.com/share/c0c3cbe677ac4dd2beb3d71c0387cd2c?sid=4a3de0fd-0c1a-42e8-9732-0255c61b6fbd

https://www.loom.com/share/2441c2495a1e444e838d794a81c64b2d?sid=1705bac5-203a-4365-8597-bce367601f01

https://www.loom.com/share/9338fa9b305a4b2e8b13bb03c0a88f03?sid=44f8eacf-37bb-4140-ad26-c5e08ff7d49c


app/core/Analytics/MetaMetrics.events.ts
```diff
@@ -561,10 +565,6 @@ enum EVENT_NAME {
   PERPS_POSITION_CLOSE_PARTIALLY_FILLED = 'Perp Position Close Partially filled',
   PERPS_POSITION_CLOSE_FAILED = 'Perp Position Close Failed',
 
-  // Push Notifications Flow
-  PERPS_PUSH_NOTIFICATION_RECEIVED = 'Perp Push Notification Received',
-  PERPS_PUSH_NOTIFICATION_CLICKED = 'Perp Push Notification Clicked',
```
Prithpal-Sooriya, line 566: Push notification event is generic, we add additional information inside the event properties.

app/core/DeeplinkManager/DeeplinkManager.ts
```diff
@@ -171,6 +172,12 @@ class DeeplinkManager {
       }
     };
 
+    FCMService.onClickPushNotification().then((deeplink) => {
+      if (deeplink) {
+        handleDeeplink({ uri: deeplink });
+      }
+    });
```
Prithpal-Sooriya: On startup will capture the click push notification and navigate once unlocked.

app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts
```diff
@@ -36,6 +36,10 @@ enum SUPPORTED_ACTIONS {
   WC = ACTIONS.WC,
 }
 
+const interstitialWhitelist = [
+  `${PROTOCOLS.HTTPS}://${AppConstants.MM_IO_UNIVERSAL_LINK_HOST}/${SUPPORTED_ACTIONS.PERPS_ASSET}`,
+] as const;
```
Prithpal-Sooriya, line 41: Might be 🌶️ change.

Adds a whitelist for specific domains to not show the interstitial. Clicking a push notification that then opens the app and shows an interstitial is a little scary.

Currently the whitelist is on the specific perp universal link we support, but we can change this to be wider.

android/app/src/main/java/io/metamask/MainApplication.kt
```diff
@@ -44,6 +45,7 @@ class MainApplication : Application(), ShareApplication, ReactApplication {
                 packages.add(RCTMinimizerPackage())
                 packages.add(NativeSDKPackage())
                 packages.add(RNTarPackage())
+                packages.add(NotificationPackage())
```
Prithpal-Sooriya, line 48: Man, I tried everything to use `@react-native-firebase` to capture the Android app open intent for firebase messages, but it kept getting lost or removed 😭 

Opted to use a custom NativeModule that tracks the firebase messaging notification open intent and shapes it to be similar to `@react-native-firebase` expected object shape. This unblocks our JS layer to use the push notification when the app was killed.

Prithpal-Sooriya: Happy to revisit, but so much hours were sunken into debugging researching this. This fix should be satisfactory to enable push notification deeplinks on Android

android/app/src/main/java/io/metamask/nativeModules/NotificationModule.kt
```diff
@@ -0,0 +1,123 @@
...
+    )
+
+    fun saveNotificationIntent(intent: Intent?) {
+      intent?.extras?.let { extras ->
+        val hasFCMData = extras.keySet().any { key ->
+          FCM_KEYS.contains(key) ||
+            key.startsWith("google.") ||
+            key.startsWith("gcm.") ||
+            (key != "profile" && !key.contains("android"))
+        }
```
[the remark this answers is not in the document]
Prithpal-Sooriya: This is okay - I've vetted our push notification shapes and it seems valid.
We also have additional validation in the JS side to vet the push notification data deeplink.

android/app/src/main/java/io/metamask/nativeModules/NotificationModule.kt
```diff
@@ -0,0 +1,123 @@
...
+            data.getString(key)
+              ?: data.getLong(key, -1).takeIf { it != -1L }
+              ?: data.getInt(key, -1).takeIf { it != -1 }
+              ?: data.getBundle(key)
+              ?: data.getBoolean(key, false)
+          }
+
+          else -> null
+        } ?: return@forEach
+
```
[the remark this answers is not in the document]
Prithpal-Sooriya: This should actually be okay.
1. The notification push payload only allows string fields (realistically these are not needed)
2. The non-string fields are provided by FCM and are relatively safe.
3. Even if this fails and causes an exception in the Android Layer, it does not crash the JS application layer - will just throw an error, which we catch!

app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts
Montoya: I am not certain this is the right approach. It seems to me the ideal approach is to skip the interstitial when the referrer is an OS notification, while marketing should only be using signed deeplinks. Otherwise, with this approach, anyone can link to a perps page and there's no interstitial for 3rd party links.

bergarces (dismissed):

app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts
Prithpal-Sooriya: Good idea!

Push notifications do not have a referrer, but we can take this idea of drilling down some `isPushNotification` criteria to here to decide skip the interstitial.

I'll do some digging to see how feasible this is (quite a few places to jump through). Depending on the scope, I can have a fast follow PR on this.

android/app/src/main/java/io/metamask/MainApplication.kt
Prithpal-Sooriya: Fast follow PR to use whitelist push notifications by providing an accepted property.
https://github.com/MetaMask/metamask-mobile/pull/19451

app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts
Cal-L: I recall that one of the reasons for reusing deeplink logic is to leverage existing routing logic. I'd still recommend not using the entire deeplink flow though since the interstitial is a byproduct of what we're trying to achieve. Notifications are always trusted and can still be funneled through the routing logic - Is there a reason that we wouldn't just reuse the handlers directly instead?

Prithpal-Sooriya: Fast follow PR to use whitelist push notifications by providing an accepted property.
https://github.com/MetaMask/metamask-mobile/pull/19451

bergarces (approved):

salimtb (approved):

app/core/DeeplinkManager/DeeplinkManager.ts
```diff
@@ -171,6 +172,18 @@ class DeeplinkManager {
       }
     };
 
+    FCMService.onClickPushNotificationWhenAppClosed().then((deeplink) => {
```
Cal-L, line 175: Let's add a comment that this only triggers once when the app starts

app/util/notifications/services/FCMService.ts
```diff
@@ -177,5 +214,34 @@ class FCMService {
...
+    deeplinkCallback: (deeplink?: string) => void,
+  ) => {
+    try {
+      messaging().onNotificationOpenedApp((remoteMessage) => {
+        try {
+          analyticsTrackPushClickEvent(remoteMessage);
+          const deeplink = remoteMessage?.data?.deeplink?.toString();
+          deeplinkCallback(deeplink);
+        } catch {
+          // Do nothing
```
Cal-L, line 239: Do we need any error logging here?

android/app/src/main/java/io/metamask/MainApplication.kt
Prithpal-Sooriya: Areas to revisit:

- See what happens when we use the react-native-firebase package directly inside `MainActivity.kt`.
  - I don't see any direct methods inside firebase that we can call. It does [rely on the current activities intent](https://github.com/invertase/react-native-firebase/blob/v21.5.0/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java#L66-L69) when calling `getInitialNotification maybe this can be overwritten?

Cal-L (approved):
LGTM

app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts
Prithpal-Sooriya: Valid point @Cal-L!
Currently push notification deeplinks are only for perps, but this will expand to other links.

You are right, we don't really need to rely on full `handleUniversalLink` logic (which also handles the interstitial), realistically we only need this routing logic:
https://github.com/MetaMask/metamask-mobile/blob/33dd30701a82c42ac85a5c2d0e6f99f238063744/app/core/DeeplinkManager/ParseManager/handleUniversalLink.ts#L140-L193

I think reusing the routing logic would be a cleaner solution than trying to bypass the interstitial, I'll update this in my fast follow PR.

app/core/DeeplinkManager/DeeplinkManager.ts
Prithpal-Sooriya: Valid, will pencil this down and add in the fast follow PR

app/util/notifications/services/FCMService.ts
Prithpal-Sooriya: +1 - it would be a good idea to log these failures (can be very handy to see in sentry)

Will add in fast follow.

PR #19438, opened 2025-09-09
feat(perps): optimize Intl formatting performance to reduce tab switc…

abretonc7s:
## **Description**

This PR addresses critical performance issues in the Perps tab identified in the Margelo performance analysis. The main bottleneck was excessive creation of uncached `Intl.NumberFormat` and `Intl.DateTimeFormat` instances during market data transformation and component rendering.

**Root Cause:** Functions in `marketDataTransform.ts` and `formatUtils.ts` were creating new Intl instances on every call instead of using the existing cached formatters from `app/util/intl.ts`.

**Solution:** Replace all uncached Intl instantiations with cached formatters and optimize volume parsing logic with memoization.

## **Changelog**

CHANGELOG entry: Improved Perps tab performance by optimizing currency and number formatting

## **Related issues**

Fixes: 

## **Manual testing steps**

```gherkin
Feature: Perps tab performance optimization

  Scenario: user switches to Perps tab
    Given user is on any tab other than Perps
    And market data contains multiple assets with various volume formats
    
    When user taps on Perps tab
    Then tab should load faster than before
    And market list should render smoothly without lag
    And price formatting should display correctly
    And volume sorting should work as expected

  Scenario: user views individual asset details
    Given user is on Perps market list
    
    When user taps on any market row item  
    Then LivePriceHeader should render more quickly than before
    And price updates should be smooth and responsive
    And all formatting should display correctly
```

### **Before**
Performance issues identified in Margelo analysis:
- `transformMarketData` function takes ~2s causing slow Perps tab loading
- `LivePriceHeader` takes ~30ms to render due to uncached Intl usage
- `PerpsMarketRowItem` takes ~60ms per render with multiple Intl instantiations
- `usePerpsMarkets` parseVolume function takes 100-300ms

### **After**
- Replaced all uncached Intl instances with cached formatters
- Added volume parsing cache to avoid repeated string processing
- All existing functionality preserved with optimized performance

- [x] I've included tests if applicable (all existing tests pass - 134+ tests)

## **Technical Details**

**Files Modified:**
- `app/components/UI/Perps/utils/marketDataTransform.ts`: Replaced 6 Intl instances with cached formatters
- `app/components/UI/Perps/utils/formatUtils.ts`: Replaced 4 Intl instances with cached formatters  
- `app/components/UI/Perps/hooks/usePerpsMarkets.ts`: Added volume parsing cache and optimized regex
- `app/components/UI/Perps/components/TradingViewChart/TradingViewChartTemplate.tsx`: Added explanatory comment


app/components/UI/Perps/hooks/usePerpsMarkets.ts
```diff
@@ -65,36 +65,55 @@ export const usePerpsMarkets = (
   const [isRefreshing, setIsRefreshing] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
+  // Cache for parsed volume values to avoid repeated parsing
+  const volumeCache = useRef(new Map<string, number>()).current;
```
Matt561: Given our limited time, I'm approving this but think there's probably a simpler and more effective way to do this. I'd also prefer if we set an upper bound on the cache size. 

This implementation doesn't cache across components/views since the cache lives in a hook. This seems like it'll optimize re-rendering the same screen but as soon as the user navigates away the cache is lost.

If I'm on the market list, click on a market, and go back to the list the cache is lost.

Matt561 (dismissed):

app/components/UI/Perps/hooks/usePerpsMarkets.ts
Matt561: cc: @abretonc7s @dylanbutler1 - tagging so my comment isn't lost.

Matt561 (approved):

PR #19505, opened 2025-09-10
feat(SDKConnectV2): Connection Resumption & App State Handling

chakra-guy:
## **Description**

This pull request enhances the **Mobile Wallet Protocol (MWP)** V2 integration by introducing connection lifecycle management. It builds upon the existing connection handling by adding session resumption on app startup and proactive reconnection when the app returns from the background.

**Project Roadmap:**
- [x] Scaffolding
- [x] Connection Handling
- [x] UI Integration
- [x] **Connection Resumption & App State Handling (This PR)**
- [ ] RPC Message Handling (Bridge Integration)

**This PR:**
*   **Session Resumption on Startup:** Implements an `initialize` method in `ConnectionRegistry` to load and resume persisted V2 sessions from storage when the app starts, ensuring session continuity.
*   **Proactive Reconnection:** Adds an `AppState` listener to `ConnectionRegistry` that triggers a `reconnectAll` method when the app becomes active. This prevents stale connections after the app has been backgrounded.
*   **Connection Service Updates:** Introduces a `resume` method to the `Connection` class. The `create` factory method is also updated to support rehydrating connections from persisted data.
*   **Dependency Upgrade:** Upgrades `@metamask/mobile-wallet-protocol-core` and `@metamask/mobile-wallet-protocol-wallet-client` to `^0.1.1` to leverage new session management capabilities like `resume()` and `reconnect()`.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

No manual testing steps are needed for this PR. The V2 connection code path is not fully integrated end-to-end, making these new lifecycle methods currently unreachable in a standard build. Comprehensive manual testing will be performed in subsequent PRs once the full connection flow is established.

### **Before**

N/A (Logic change only)

### **After**

N/A (Logic change only)


app/core/SDKConnectV2/services/connection.ts
```diff
@@ -30,19 +31,21 @@ export class Connection {
...
    */
   public static async create(
-    connreq: ConnectionRequest,
+    data: ConnectionRequest | PersistedConnection,
     keymanager: IKeyManager,
     relayURL: string,
   ): Promise<Connection> {
-    const id = connreq.sessionRequest.id;
-    const metadata = connreq.metadata;
+    const id = 'sessionRequest' in data ? data.sessionRequest.id : data.id;
```
jiexi, line 46: thoughts on splitting the `data` param into `id` and `metadata`?

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -29,6 +31,36 @@ export class ConnectionRegistry {
...
+        const conn = await Connection.create(
+          c,
+          this.keymanager,
+          this.RELAY_URL,
+        );
+        await conn.resume();
+        this.connections.set(conn.id, conn);
+      } catch (error) {
+        console.error(
+          `[SDKConnectV2] Failed to resume session ${c.id}, removing.`,
```
jiexi: is an explicit removal supposed to happen in this catch block?

app/components/Views/SDK/SDKDisconnectModal/SDKDisconnectModal.tsx
```diff
@@ -105,6 +111,35 @@ const SDKDisconnectModal = ({ route }: SDKDisconnectModalProps) => {
...
   };
 
+  const onConfirmV2 = async () => {
+    try {
+      // Case: Disconnect a single account from a V2 session.
+      // This is a pure permission management action.
+      if (account && channelId) {
+        removePermittedAccounts(channelId, [toHex(account)]);
+        // If it's the last account, escalate to a full session termination.
+        if (accountsLength && accountsLength <= 1) {
```
jiexi: if `account` is provided, can it be assumed that `accountsLength` is as well? There's potential for getting into the state where the permission is removed but the session is not disconnected otherwise

app/core/SDKConnectV2/adapters/host-application-adapter.ts
```diff
@@ -38,9 +35,33 @@ export class HostApplicationAdapter implements IHostApplicationAdapter {
...
+    }, {} as SDKSessions);
+
+    store.dispatch(setSdkV2Connections(v2Sessions));
+  }
+
+  revokePermissions(connectionId: string): void {
+    const allAccountsForOrigin = getPermittedAccounts(connectionId);
+    if (allAccountsForOrigin.length > 0) {
+      removePermittedAccounts(connectionId, allAccountsForOrigin);
+    }
```
jiexi: if we're ensuring that a connection only remains active if there are permitted accounts, is it safe to just assume the permission does exist and call `engine.PermissionController.revokePermission()` instead?

app/components/Views/SDK/SDKDisconnectModal/SDKDisconnectModal.tsx
chakra-guy: This is not part of the PR, however valid comments. Created a ticket for this.

app/core/SDKConnectV2/adapters/host-application-adapter.ts
chakra-guy: This is not part of the PR, however valid comments. Created a ticket for this.

app/core/SDKConnectV2/services/connection-registry.ts
chakra-guy: Nope. I fixed the console error (it shouldn't say "removing"). 
Error handling is done in the following way: If a connection fails, do not remove or delete, it might be just a network issue that would resolve. And if it's a permanent error, we also do not have to explicitly handle it, as it will be cleaned up by the storage once it expires.

app/core/SDKConnectV2/services/connection.ts
chakra-guy: Yeah, I thought about that approach too, but in the end I went with this for the following reason: 
It better represents what are the scenarios/inputs it handles from a business logic perspective, making the code (in my opinion) better describe itself. "ID" and "metadata" doesn't really say anything, while "connection request" or "persisted connection" does.

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -107,17 +135,63 @@ export class ConnectionRegistry {
...
+          isColdStart = false;
+          return;
+        }
+
+        // For all subsequent 'active' events, we reconnect, but only after
+        // the initial setup is guaranteed to be complete to avoid race conditions.
+        this.ready.then(() => this.reconnectAll());
+      },
+    );
+  }
```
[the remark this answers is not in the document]
chakra-guy: As this class is a singleton, it's a none issue

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -107,17 +135,63 @@ export class ConnectionRegistry {
       throw new Error('[SDKConnectV2] Payload too large (max 1MB).');
     }
 
-    let connreq: unknown;
-    try {
-      connreq = JSON.parse(payload);
-    } catch (error) {
-      throw new Error('[SDKConnectV2] Invalid JSON in payload.');
-    }
+    const connreq: unknown = JSON.parse(payload);
```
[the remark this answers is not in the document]
chakra-guy: This was removed, as sonarcloud was complaining about this.

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -29,6 +31,38 @@ export class ConnectionRegistry {
...
+  private async initialize(): Promise<void> {
+    console.warn(
+      '[SDKConnectV2] Initializing and resuming persisted connections...',
+    );
+
+    const persisted = await this.store.list();
+
+    const promises = persisted.map(async (c) => {
+      try {
+        const conn = await Connection.create(
```
Cal-L, line 50: Optimization nit - Not sure how likely it is for a large amount of connects to be happening at the same time. Do you think it's possible to lazy load the connections once a persisted connection ID is recognized or does the connection have to ready at the time the request is received? If we can lazy load them, we should be able to save resources on unused connections.

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -107,17 +135,63 @@ export class ConnectionRegistry {
...
+  /**
+   * Proactively refreshes all active connections. This is the primary mechanism
+   * for preventing stale/zombie connections after the app was put in the background.
+   */
+  private async reconnectAll(): Promise<void> {
+    console.warn('[SDKConnectV2] Proactively refreshing all connections...');
+
+    const connections = Array.from(this.connections.values());
+
+    const promises = connections.map((conn) =>
```
Cal-L, line 184: Same question here regarding potentially lazy loading connections if they're still interacted with

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -107,17 +135,63 @@ export class ConnectionRegistry {
...
     return connreq;
   }
+
+  /**
+   * Sets up the listener for app state lifecycle events to handle reconnection.
+   */
+  private setupAppStateListener(): void {
+    let isColdStart = true;
+
+    AppState.addEventListener(
```
Cal-L, line 153: Is there scenario where the SDKConnect service needs to be destroyed? If so, we'll need to unsubscribe from this.

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -29,6 +31,38 @@ export class ConnectionRegistry {
     this.keymanager = keymanager;
     this.hostapp = hostapp;
     this.store = store;
+    this.ready = this.initialize();
```
Cal-L, line 34: The naming for this can be misleading since it resembles a boolean. Should we call this something like `this.resumedConnections` instead?

Cal-L (commented):
left some comments

app/core/SDKConnectV2/services/connection-registry.ts
chakra-guy: Thanks for raising the point on lazy loading. It would be nice, but unfortunately, it's not technically feasible. 
Unlike a typical request-response model, WebSockets are persistent connections where the Relay server can push a message to the wallet at any time. For our wallet to receive a request, the WebSocket connection must already be established and actively listening. If we wait to load the connection, there's no channel to receive the incoming request in the first place. Therefore, the current approach of initializing connections on app start/resume is a technical requirement.

app/core/SDKConnectV2/services/connection-registry.ts
chakra-guy: The short answer is no. The `ConnectionRegistry` is used as a singleton service within the app. It's instantiated once when the app launches and persists for the entire lifecycle of the app, so there's no scenario where it would be destroyed. Because of this, a teardown or unsubscribe is not needed.

app/core/SDKConnectV2/services/connection-registry.ts
chakra-guy: That's a fair point on the naming. My initial inspiration for the this.ready property came from other libraries, for example: the Service Worker API, which uses a very similar pattern (navigator.serviceWorker.ready) to provide a promise that resolves when the service is active and ready to handle events (which is very similar to our usecase).

That said, I completely understand the concern that it could be misleading. If you still feel it's not readable enough, I'm happy to change it. Some alternatives could be: this.initialization, this.readyPromise, this.initDone

app/core/SDKConnectV2/services/connection-registry.ts
chakra-guy: Update: I have started working on the next PR, which is the RPC Message handling part and there, we should be able to lazy init the background bridge client!

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -29,6 +31,38 @@ export class ConnectionRegistry {
...
+    this.setupAppStateListener();
+  }
+
+  /**
+   * One-time initialization to resume all persisted connections on app cold start.
+   */
+  private async initialize(): Promise<void> {
```
jiexi, line 41: i think this should return `true` or reject with a friendlier error rather than an array of settled promises. thoughts?

app/core/SDKConnectV2/services/connection-registry.ts
jiexi: Thoughts on some of these alternatives?
1. wrap the instantiation of the ConnectionRegistry into a async helper that ensures the registry is fully initialized before returning it to the caller
2. In each of the public methods, use `this.ready` as a guard by awaiting at the top of each public function

app/core/SDKConnectV2/services/connection-registry.ts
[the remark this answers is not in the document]
jiexi: it seems like these try catch blocks may be useful to keep still?

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -107,17 +135,63 @@ export class ConnectionRegistry {
...
+    AppState.addEventListener(
+      'change',
+      (nextAppState: AppStateStatus): void => {
+        console.warn('[SDKConnectV2] App state changed to:', nextAppState);
+
+        if (nextAppState !== 'active') {
+          return;
+        }
+
+        // First 'active' event on a cold start is ignored
```
jiexi, line 162: why is that?

chakra-guy: This is to avoid reconnecting all connections on a cold start

app/core/SDKConnectV2/services/connection-registry.ts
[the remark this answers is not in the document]
chakra-guy: Cursor's analysis is wrong here. Even though it throws native TypeError or SyntaxError, the caller does catch it, handle it and log [SDKConnectV2] error. So it should be fine after all. Also not sure why sonarcloud had an issue with the in the first place, seems silly...

app/core/SDKConnectV2/services/connection-registry.ts
chakra-guy: Updated this, so that it won't throw any error. It will be silently ignored, but the I think the promise needs to be there to avoid race conditions

Cal-L (dismissed):
Lgtm

app/core/SDKConnectV2/services/connection.ts
jiexi: I guess it just feels too hacky to not assert the type of `data` entirely before using it, but I won't block on this

jiexi (approved):

Cal-L (approved):
LGTM

PR #19547, opened 2025-09-10
chore: less rerenders for Perps deposit

Nodonisko:
## **Description**

Added proper memoization to some of the components to slightly reduce rendering time and also reduce number of rerenders.

There is still quite big room for more improvements but that would require much more deeper dive into so many nested hooks that are used there.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### DIFF
--- a/app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
+++ b/app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
@@ -1,4 +1,10 @@
-import React, { createRef, useCallback, useEffect, useState } from 'react';
+import React, {
+  createRef,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { TextInput, View } from 'react-native';
 import { useTokenAmount } from '../../hooks/useTokenAmount';
 import { useStyles } from '../../../../../component-library/hooks';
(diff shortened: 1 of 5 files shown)


app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
```diff
@@ -26,148 +32,156 @@ export interface EditAmountProps {
   onKeyboardDone?: () => void;
 }
 
-export function EditAmount({
```
Nodonisko, line 40: Not sure why GH is showing so big diff, I just added few memos and it changed indentation only

app/components/Views/confirmations/external/perps-temp/hooks/usePerpsDepositView.ts
```diff
@@ -25,7 +26,10 @@ export function usePerpsDepositView({
   const { payToken } = useTransactionPayToken();
   const { amounts: sourceAmounts } = useTransactionPayTokenAmounts();
 
-  const amountValue = new BigNumber(amountUnformatted ?? '0');
+  const amountValue = useMemo(
+    () => new BigNumber(amountUnformatted ?? '0'),
```
Nodonisko, line 30: I am not sure but I think there could be a bug, because `amountUnformatted` is all the time `0`. Is it possible that `useTokenAmount();` few lines above is missing param?

app/components/Views/confirmations/components/rows/pay-with-row/pay-with-row.tsx
```diff
@@ -43,6 +43,14 @@ export function PayWithRow() {
     });
   }, [navigation, totalFiat]);
 
+  const tokenBalance = useMemo(
+    () =>
+      payToken?.balance
+        ? formatAmount(I18n.locale, new BigNumber(payToken.balance))
+        : '0',
+    [payToken?.balance],
+  );
```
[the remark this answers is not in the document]
Nodonisko: Fixed

app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
```diff
@@ -37,189 +43,193 @@ export interface EditAmountProps {
...
-  return (
-    <View style={styles.container} testID="edit-amount-skeleton">
-      <View style={styles.primaryContainer}>
-        <View style={styles.inputContainer}>
-          <Skeleton height={70} width={80} />
+    );
+  },
+);
+
+export const EditAmountSkeleton = React.memo(
```
sethkfman, line 216: @Nodonisko Can you move this above the `EditAmount` function to address the lint error?

abretonc7s:
tagging @matthewwalsh0 as it falls under confirmations ownership

app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
```diff
@@ -37,189 +43,193 @@ export interface EditAmountProps {
...
+
+    const { styles } = useStyles(styleSheet, {
+      amountLength,
+      hasAlert,
+    });
+
+    const { tokenFiatAmount } = payToken ?? {};
+    const hasAmount = amountFiat !== '0';
+
+    const amountHuman = useMemo(
```
matthewwalsh0, line 95: Will this have any performance impact given it's a single division and returning a primitive?

app/components/Views/confirmations/components/rows/pay-with-row/pay-with-row.tsx
```diff
@@ -50,15 +50,15 @@ export function PayWithRow() {
     });
   }, [canEdit, navigation, totalFiat]);
 
+  const tokenBalance = useMemo(
```
matthewwalsh0, line 53: Same query here, this is a primitive, I usually try to avoid putting a `useMemo` around every piece of logic in a render?

app/components/Views/confirmations/external/perps-temp/hooks/usePerpsDepositView.ts
```diff
@@ -25,7 +26,10 @@ export function usePerpsDepositView({
   const { payToken } = useTransactionPayToken();
   const { amounts: sourceAmounts } = useTransactionPayTokenAmounts();
 
-  const amountValue = new BigNumber(amountUnformatted ?? '0');
+  const amountValue = useMemo(
```
matthewwalsh0, line 29: This is no longer in this hook anymore, I think could be a bad merge adding it back?

app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
```diff
@@ -37,189 +43,193 @@ export interface EditAmountProps {
...
-
-    onKeyboardShow?.();
-  }, [amountFiat, inputRef, onKeyboardShow, setIsFooterVisible]);
-
-  useEffect(() => {
-    if (autoKeyboard && !inputChanged) {
-      handleInputPress();
-    }
-  }, [autoKeyboard, inputChanged, handleInputPress]);
+export const EditAmount = React.memo(
```
matthewwalsh0, line 46: I can't recall if we're using React compiler in the mobile stack, does that not automatically memoise the components?

matthewwalsh0 (changes requested):
Thanks for these changes @Nodonisko! 

I'm afraid I'm in the middle of a large refactor of these components to make them more generic for re-use with Predict meaning their structure will be considerably different.

As for the number of React hooks in use, that was a side effect of the timelines and I'm not a fan either so I've also been migrating the business logic and calculation to a dedicated backend controller in the `core` repository.

app/components/Views/confirmations/external/perps-temp/hooks/usePerpsDepositView.ts
matthewwalsh0: Oh I see this PR is not targeting `main`, should it be?

app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
Nodonisko: JS implementation of BigNumber is incredibly slow compared to normal numbers. Initialization of new instance and that operations could easily take anything from 0.5ms to 5ms, compared to simple memo check that's like 0.0001 ms. This could easily stack up when component rerenders often.

It’s better to memoize more rather than less (that’s actually what the React compiler does, memoizing everything). `useMemo` is very cheap, and I’ve never seen performance issues caused by too much `useMemo`, but I’ve often seen the opposite, nearly in every project I’ve audited so far.

app/components/Views/confirmations/components/edit-amount/edit-amount.tsx
Nodonisko: It should be enabled but it's hard to tell which component is optimized and which one is skipped by compiler. Eslint plugin helps a bit but even when it doesn't complain it could skip component...

app/components/Views/confirmations/external/perps-temp/hooks/usePerpsDepositView.ts
abretonc7s: @sethkfman mentionned it is supposed to target a special branch before to merge to main since it comes from external contributor.

matthewwalsh0 (approved):

abretonc7s (changes requested):
It seems the PR needs to be updated, has linting issue now

Nodonisko:
There is too much changes and conflicts in most of files, I will need to profile this again. Closing this for now.

PR #19590, opened 2025-09-11
feat: Unified activity list for BIP-44

pedronfigueiredo:
## **Description**

Adding a unified activity list for both EVM and nonEvm transactions.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/5747

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### DIFF
--- a/app/components/UI/TransactionElement/index.js
+++ b/app/components/UI/TransactionElement/index.js
@@ -66,6 +66,12 @@ const createStyles = (colors, typography) =>
       backgroundColor: colors.background.default,
       flex: 1,
     },
+    rowWithBorder: {
+      backgroundColor: colors.background.default,
+      flex: 1,
+      borderBottomWidth: 1,
+      borderBottomColor: colors.border.muted,
+    },
     actionContainerStyle: {
       height: 25,
       padding: 0,
@@ -678,15 +684,9 @@ class TransactionElement extends PureComponent {
     return (
       <>
         <TouchableHighlight
-          style={[
-            styles.row,
-            // eslint-disable-next-line react-native/no-inline-styles
-            this.props.showBottomBorder && {
-              // eslint-disable-next-line react-native/no-inline-styles
-              borderBottomWidth: 1,
-              borderBottomColor: colors.border.muted,
-            },
-          ]}
+          style={
+            this.props.showBottomBorder ? styles.rowWithBorder : styles.row
+          }
           onPress={this.onPressItem}
           underlayColor={colors.background.alternative}
           activeOpacity={1}
--- a/app/components/UI/TransactionElement/index.js
+++ b/app/components/UI/TransactionElement/index.js
@@ -32,6 +32,7 @@ import {
 import { ThemeContext, mockTheme } from '../../../util/theme';
 import { selectTickerByChainId } from '../../../selectors/networkController';
 import { selectSelectedInternalAccount } from '../../../selectors/accountsController';
+import { selectSelectedAccountGroupInternalAccounts } from '../../../selectors/multichainAccounts/accountTreeController';
 import { selectPrimaryCurrency } from '../../../selectors/settings';
 import { selectSwapsTransactions } from '../../../selectors/transactionController';
 import { swapsControllerTokens } from '../../../reducers/swaps';
@@ -57,6 +58,7 @@ import { selectConversionRateByChainId } from '../../../selectors/currencyRateCo
 import { selectContractExchangeRatesByChainId } from '../../../selectors/tokenRatesController';
 import { selectTokensByChainIdAndAddress } from '../../../selectors/tokensController';
 import Routes from '../../../constants/navigation/Routes';
+import { selectMultichainAccountsState2Enabled } from '../../../selectors/featureFlagController/multichainAccounts';
 
 const createStyles = (colors, typography) =>
   StyleSheet.create({
@@ -159,6 +161,10 @@ class TransactionElement extends PureComponent {
     */
     selectedInternalAccount: PropTypes.object,
     /**
+     * Internal accounts for the selected account group
+     */
+    selectSelectedAccountGroupInternalAccounts: PropTypes.array,
+    /**
      * Current element of the list index
      */
     i: PropTypes.number,
@@ -196,6 +202,14 @@ class TransactionElement extends PureComponent {
     navigation: PropTypes.shape({
       navigate: PropTypes.func.isRequired,
     }).isRequired,
+    /**
+     * Whether multichain accounts state 2 is enabled
+     */
+    isMultichainAccountsState2Enabled: PropTypes.bool,
+    /**
+     * Whether to render a bottom border for row separation (used in unified list)
+     */
+    showBottomBorder: PropTypes.bool,
   };
 
   state = {
@@ -244,7 +258,7 @@ class TransactionElement extends PureComponent {
 
   onPressItem = () => {
     const { tx, i, onPressItem } = this.props;
-    onPressItem(tx.id, i);
+    onPressItem && onPressItem(tx.id, i);
 
     const isUnifiedSwap =
       tx.type === TransactionType.swap &&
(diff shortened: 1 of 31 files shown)


app/selectors/multichainAccounts/accountTreeController.ts
```diff
@@ -332,3 +332,30 @@ export const selectAccountGroupWithInternalAccounts = createSelector(
...
+/**
+ * Selector to get internal accounts associated with the currently selected account group.
+ *
+ * This composes `selectAccountGroupWithInternalAccounts` and `selectSelectedAccountGroup`
+ * to return the list of internal accounts belonging to the selected group.
+ *
+ * @param state - The Redux root state
+ * @returns A readonly array of internal accounts for the selected account group
+ */
+export const selectSelectedAccountGroupInternalAccounts = createSelector(
```
gantunesr, line 376: Do you think you can add the unit test to this selector?

app/selectors/multichain/multichain.ts
```diff
@@ -420,6 +421,64 @@ export const selectNonEvmTransactions = createDeepEqualSelector(
   },
 );
 
+/**
+ * Returns non-EVM transactions for all internal accounts in the selected account group.
+ * Flattens transactions across all supported non-EVM chains within the group.
+ */
+export const selectNonEvmTransactionsForSelectedAccountGroup =
```
gantunesr, line 454: Similarly here, we should ensure 100% coverage on our selectors. If it's too much screep cope then we can link a ticket here to add unit test

app/selectors/multichain/multichain.ts
```diff
@@ -420,6 +421,64 @@ export const selectNonEvmTransactions = createDeepEqualSelector(
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const entries = Array.isArray((accountTx as any)?.transactions)
+          ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            [accountTx as any]
+          : // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            Object.values(accountTx as Record<string, any>);
+
+        for (const entry of entries) {
+          const txs = entry?.transactions ?? [];
+          aggregated.transactions = aggregated.transactions.concat(txs);
```
gantunesr: You can use `push` here, I think this is more efficient since `concat` creates a new array on every call. Quick article about it: https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki

```suggestion
          aggregated.transactions = aggregated.transactions.push(...txs);
```

owencraston:
> Does it still make sense to have the network filter in this flow? The current network filter seems out of place for a multichain account.
> 
> Screenshot 2025-09-11 at 5 06 12 PM

I realized that I needed to add the flag `export MM_REMOVE_GLOBAL_NETWORK_SELECTOR="true"` to my js.env.

owencraston (commented):
There is some weird behaviour going on with this screen and network selector.

When the user has the 'Popular networks" or "all networks" or "any evm" network selected, the drop down is clickable. When the user has solana selected, the drop down it no longer clickable. 

then I noticed that when you multi select some networks (example OP and Solana) it shoes as "all networks" in the drop down menu. I am not sure if this is related to your change but its defiantly buggy. The transaction list itself seems to be working.

https://github.com/user-attachments/assets/0471b03d-bf3e-46a7-bc5d-1e843cf6b394

app/selectors/multichainAccounts/accountTreeController.ts
pedronfigueiredo: Added unit tests now.

app/selectors/multichain/multichain.ts
pedronfigueiredo: Added unit tests now.

app/selectors/multichain/multichain.ts
pedronfigueiredo: great shout, thanks for pointing that out!

pedronfigueiredo:
@owencraston thanks for testing the change! There is indeed some instability in the network picker at the moment, I believe this is something that the assets team will be looking into fixing soon. I'd prefer to keep that separate from this already big PR.

app/components/UI/TransactionElement/index.js
```diff
@@ -641,7 +678,15 @@ class TransactionElement extends PureComponent {
     return (
       <>
         <TouchableHighlight
-          style={styles.row}
+          style={[
+            styles.row,
+            // eslint-disable-next-line react-native/no-inline-styles
```
tommasini: Instead of disabling can't we create the styled separately?

app/components/Views/UnifiedTransactionsView/useUnifiedTxActions.ts
```diff
@@ -0,0 +1,221 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
tommasini: Should this component be inside of confirmations folder?

I believe we shouldn't disable the explicit any from here

app/components/Views/UnifiedTransactionsView/useUnifiedTxActions.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  const cancelUnsignedQRTransaction = async (tx: any) => {
+    await (Engine.context as any).ApprovalController.reject(
+      tx.id,
+      providerErrors.userRejectedRequest(),
+    );
+  };
+
+  return {
+    // state
```
tommasini: Forgot comments

app/selectors/multichain/multichain.test.ts
```diff
@@ -784,6 +784,171 @@ describe('MultichainNonEvm Selectors', () => {
...
+            },
+            // Noise: account not in the selected group
+            [MOCK_ACCOUNT_BIP122_P2WPKH_TESTNET.id]: {
+              [BtcScope.Testnet]: {
+                transactions: [{ id: 'tbtc-1', timestamp: 999 }],
+                next: null,
+                lastUpdated: 1200,
+              },
+            },
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tommasini: We shouldn't disable any

app/components/Views/UnifiedTransactionsView/useUnifiedTxActions.ts
pedronfigueiredo: We will want to refactor this component soon, this will be addressed then.

app/selectors/multichain/multichain.test.ts
pedronfigueiredo: We will want to refactor this component soon, this will be addressed then.

Matt561 (dismissed):
Approving change to single Earn-owned file 👍

jpuri (dismissed):

matthewwalsh0 (dismissed):

owencraston (dismissed):

app/components/Views/UnifiedTransactionsView/useUnifiedTxActions.ts
```diff
@@ -0,0 +1,356 @@
...
+    setRetryIsOpen((prev) => !prev);
+    setRetryErrorMsg(msg);
+  };
+
+  const getGasPriceEstimate = () => {
+    if (!gasFeeEstimates) {
+      return addHexPrefix(String(decGWEIToHexWEI('0')));
+    }
+
+    if ('type' in (gasFeeEstimates as object)) {
```
tommasini, line 90: The reason why we are type casting the gasFeeEstimates seems to show that the selector accurately typed, but we are just fetching the gasFeeEstimates from the controller. I couldn't find the reason why we loose the type on the selector 
image


But we could type it only on the selector return for now, and avoid all the casting down the root. Ideally we fix the why the selector is not returning.

tommasini (approved):

app/components/Views/UnifiedTransactionsView/useUnifiedTxActions.ts
pedronfigueiredo: Saving this for the future refactor as it would involve making changes to a selector we don't need to otherwise. We will refactor de activity list on both clients in Q4!

owencraston (approved):

Matt561 (approved):

jpuri (approved):

PR #19705, opened 2025-09-12
fix: add timeout to `net_version` request when getting the provider state

jiexi:
## **Description**

During investigation of a suspected issue with `chainChanged` and `accountsChanged` events, we found that a hanging `net_version` call can block the `getProviderState` call from responding. This has the effect of making the `window.ethereum` provide not properly propagating subsequent chainChanged events. This PR introduces a timeout to the underlying `net_version` request that could previous cause `getProviderState` to hang.

## **Changelog**

CHANGELOG entry: Fixed a bug where sometimes an unresponsive RPC endpoint could cause the UI to be stuck on a loading spinner and/or the `window.ethereum` provider to be stuck in a bad state. 

## **Related issues**

See: https://consensyssoftware.atlassian.net/browse/WAPI-421
See: https://github.com/MetaMask/metamask-extension/pull/35516

## **Manual testing steps**

Testing this can be tricky as it's not possible to add an RPC endpoint when it's already non-responsive. Using ganache/hardhat and then turning off the server is one way to get around this.

```gherkin
Feature: Add timeout to networkVersion requests

   Scenario: user opens the in app browser selected network unresponsive 
    Given the RPC is unresponsive (to net_version, or all requests in general)

    When user opens the in app browser and visits a dapp
    Then `window.ethereum._state.initialized` should return true in the console
```

### DIFF
--- a/app/core/BackgroundBridge/BackgroundBridge.js
+++ b/app/core/BackgroundBridge/BackgroundBridge.js
@@ -89,6 +89,9 @@ import PPOMUtil from '../../lib/ppom/ppom-util';
 import { isRelaySupported } from '../RPCMethods/transaction-relay';
 import { selectSmartTransactionsEnabled } from '../../selectors/smartTransactionsController';
 
+// Timeout for fetching deprecated network version (5 seconds)
+const GET_DEPRECATED_NETWORK_VERSION_TIMEOUT = 5000;
+
 const legacyNetworkId = () => {
   const { networksMetadata, selectedNetworkClientId } =
     store.getState().engine.backgroundState.NetworkController;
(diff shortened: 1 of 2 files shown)


app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -89,6 +89,9 @@ import PPOMUtil from '../../lib/ppom/ppom-util';
 import { isRelaySupported } from '../RPCMethods/transaction-relay';
 import { selectSmartTransactionsEnabled } from '../../selectors/smartTransactionsController';
 
+// Timeout for fetching deprecated network version (5 seconds)
```
ffmcgee725, line 92: [nit] okay with not so much explicitness in this comment ? Name is self descriptive enough I believe, at best I would keep something about the unit the integer is in (milliseconds, which represent the 5 seconds)

app/core/BackgroundBridge/BackgroundBridge.js
```diff
@@ -313,28 +316,53 @@ export class BackgroundBridge extends EventEmitter {
...
 
     return {
       chainId,
       networkVersion: networkVersion ?? 'loading',
     };
   }
 
+  async getDeprecatedNetworkVersion({ networkClient, networkClientId }) {
+    const networkVersion = this.deprecatedNetworkVersions[networkClientId];
+    if (networkVersion !== undefined) {
```
ffmcgee725, line 332:
```suggestion
    if (networkVersion)
```

would this be okay or could it possibly result in underlying issues ?

ffmcgee725: nevermind, read the rest of the code below, `null` is a valid value for this and we want it to go through.

app/core/BackgroundBridge/BackgroundBridge.test.js
```diff
@@ -219,7 +222,7 @@ describe('BackgroundBridge', () => {
 
   describe('notifySolanaAccountChangedForCurrentAccount', () => {
     it('emits nothing if there is no CAIP-25 permission', () => {
-      const url = 'https:www.mock.io';
+      const url = 'https://www.mock.io';
```
ffmcgee725, line 225: good catch... 👀

ffmcgee725 (approved):

app/core/BackgroundBridge/BackgroundBridge.js
jiexi: what about this instead?
`const GET_DEPRECATED_NETWORK_VERSION_TIMEOUT = 5000; // ms`

ffmcgee725: Sounds good yeah 👍🏾

PR #19717, opened 2025-09-14
fix: tp sl solution in perps proposal B

nickewansmith:
## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

  Scenario: user [verb for user action]
    Given [describe expected initial app state]

    When user [verb for user action]
    Then [describe expected outcome]
```

## **Pre-merge author checklist**

- [ ] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [ ] I've completed the PR template to the best of my ability
- [ ] I’ve included tests if applicable
- [ ] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #19726, opened 2025-09-15
fix: tp sl solution in perps proposal B

nickewansmith:
## **Description**
- https://consensys.slack.com/archives/C08U6DYNJ1G/p1757700508107169
- In MetaMask mobile perps, we associate TP/SL orders with a position, which isn’t the case on Hyperliquid order book.
  - This creates complexity when users execute trades with an open position or an open order.

For example, let’s consider this scenario:
First, user opens a Long position of size Z, with a TP at $X and SL at $Y
Then user opens a Long position of size Z’, with a TP at $X’ and SL at $Y’

In Hyperliquid, user will have 1 position of size Z + Z’ and 4 open orders (see [recording](https://www.loom.com/share/8a570ab01322419c967ad0275008b025)):
TP of size Z at price $X
TP of size Z’ at price $X’
SL of size Z at price $Y
SL of size Z’ at price $Y’

From there, if the user creates a limit order with a TP/SL, it will create 3 new orders (cf. [recording](https://www.loom.com/share/5330714a85984a839073dbda215876e5))

The current MetaMask mobile perps UI doesn’t support this: we associate one position with one TP/SL, and this TP/SL is expected to apply to the entire position. Given multiple orders are created in Hyperliquid order book, MetaMask UI doesn’t know which TP/SL to associate with this position, and returns TP/SL as “Not set”

https://consensyssoftware.atlassian.net/browse/TAT-1691 Explains Solution A

Solution B is similar to A except it does not block the user from adding a tpsl on the position page when there are multiple orders and with B we use position bound tpsl which resizes with the position. When a user places a tpsl order and then resizes their position, they do not need to worry about adjusting tpsls that no longer match the order size.
  - If tpsl order size is lower then the current position size after buying more of a token pair, then in A the user is left to handle that on their own. 
  - Because of this possible mismatch, in A we block the user from making position tpsl if there is not a match for the position size or higher. We tell them to close all of their tpsl orders instead
 

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1691

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **After**
https://github.com/user-attachments/assets/9eec2173-19cf-45a0-9540-55cf6f8daacf

### DIFF
--- a/app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
+++ b/app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
@@ -120,19 +120,87 @@ const PerpsMarketDetailsView: React.FC<PerpsMarketDetailsViewProps> = () => {
 
   const [refreshing, setRefreshing] = useState(false);
 
+  // TP/SL order selection state - track TP and SL separately
+  const [activeTPOrderId, setActiveTPOrderId] = useState<string | null>(null);
+  const [activeSLOrderId, setActiveSLOrderId] = useState<string | null>(null);
+
   const account = usePerpsAccount();
 
   usePerpsConnection();
   const { depositWithConfirmation } = usePerpsTrading();
   const { ensureArbitrumNetworkExists } = usePerpsNetworkManagement();
   // Get real-time open orders via WebSocket
-  const ordersData = usePerpsLiveOrders({ hideTpSl: true }); // Instant updates with TP/SL filtered
-
+  const ordersData = usePerpsLiveOrders({});
   // Filter orders for the current market
   const openOrders = useMemo(() => {
     if (!ordersData?.length || !market?.symbol) return [];
     return ordersData.filter((order) => order.symbol === market.symbol);
   }, [ordersData, market?.symbol]);
+
+  // Filter orders that have TP/SL data for chart integration
+  const ordersWithTPSL = useMemo(
+    () =>
+      openOrders.filter((order) => {
+        // Check if order has TP/SL prices directly
+        if (order.takeProfitPrice || order.stopLossPrice) return true;
+
+        // Check if it's a trigger order (TP/SL orders are stored as trigger orders)
+        if (order.isTrigger && order.detailedOrderType) {
+          const orderType = order.detailedOrderType.toLowerCase();
+          return (
+            orderType.includes('take profit') || orderType.includes('stop')
+          );
+        }
+
+        return false;
+      }),
+    [openOrders],
+  );
+
+  // Determine which TP/SL lines to show on the chart
+  const selectedOrderTPSL = useMemo(() => {
+    // Find the active TP order
+    let activeTPOrder = ordersWithTPSL.find(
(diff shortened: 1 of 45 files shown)


nickewansmith:
https://github.com/MetaMask/metamask-mobile/pull/19717

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
```diff
@@ -126,8 +126,9 @@ const PerpsMarketDetailsView: React.FC<PerpsMarketDetailsViewProps> = () => {
   const { depositWithConfirmation } = usePerpsTrading();
   const { ensureArbitrumNetworkExists } = usePerpsNetworkManagement();
   // Get real-time open orders via WebSocket
-  const ordersData = usePerpsLiveOrders({ hideTpSl: true }); // Instant updates with TP/SL filtered
-
+  // const ordersData = usePerpsLiveOrders({ hideTpSl: true }); // Instant updates with TP/SL filtered
```
gambinish: Can remove this if it's not needed

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
```diff
@@ -126,8 +126,9 @@ const PerpsMarketDetailsView: React.FC<PerpsMarketDetailsViewProps> = () => {
...
+  const ordersData = usePerpsLiveOrders({});
+  // console.log('ordersData', JSON.stringify(ordersData, null, 2));
```
gambinish: Remove log

app/components/UI/Perps/components/PerpsMarketTabs/PerpsMarketTabs.tsx
```diff
@@ -365,6 +435,7 @@ const PerpsMarketTabs: React.FC<PerpsMarketTabsProps> = ({
               expanded
               showIcon
               onTooltipPress={handleTooltipPress}
+              onTpslCountPress={() => handleTabChange('orders')}
```
gambinish: This will create a new function reference on every render. Might make sense to have a dedicated useCallback handler to memoize it

app/components/UI/Perps/components/PerpsPositionCard/PerpsPositionCard.tsx
```diff
@@ -163,6 +198,99 @@ const PerpsPositionCard: React.FC<PerpsPositionCardProps> = ({
...
+          formatPerpsFiat(position.stopLossPrice, {
+            ranges: PRICE_RANGES_POSITION_VIEW,
+          })
+        ) : (
+          strings('perps.position.card.not_set')
+        )}
+      </Text>
+    );
+  };
+
```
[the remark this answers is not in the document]
nickewansmith: Will clean up a bit, not set is working as inteded but the ternary is confusing

gambinish (approved):

PR #19730, opened 2025-09-15
fix: account switching in account permission list

montelaidev:
## **Description**

This PR fixes the issue where the use is not able to switch their selected account from the account permission screen. This pr also fixes the issue where after confirming a change in permissions, the client will navigate to the home screen. 

## **Changelog**

CHANGELOG entry: Fix active account selection from the bip44 account permission screen. 

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUL-844

## **Manual testing steps**

```gherkin
Feature: dapp connection
 Scenario: user manages existing dapp permissions
    Given the user has previously connected accounts to a dapp
    And the user is on the dapp page
    And at least one account group is connected

    When user taps the account button in the browser
    And user sees the permissions summary screen

    Then user should see the list of connected account groups
    And user taps an account group
    And user taps confirm

    Then user should see that the tapped account group becomes the selected account group. 
```

### **Before**

https://github.com/user-attachments/assets/8af8d1d3-a056-4e43-8103-5f381841f848

### **After**

https://github.com/user-attachments/assets/ddc0dbee-1633-4886-9813-be5f3d344ba2


app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -163,33 +162,9 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
-  );
-
   const requestedCaipChainIdsWithDefaultSelectedChainIds = useMemo(
-    () =>
-      Array.from(
-        new Set([...requestedCaipChainIds, ...defaultSelectedChainIds]),
-      ),
-    [requestedCaipChainIds, defaultSelectedChainIds],
+    () => Array.from(new Set([...requestedCaipChainIds, ...allNetworksList])),
+    [requestedCaipChainIds, allNetworksList],
```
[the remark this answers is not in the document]
adonesky1: This comment looks valid to me. This appears to fully remove some important defaulting logic?

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
```diff
@@ -158,6 +166,13 @@ export const useAccountGroupsForPermissions = (
       }
     });
 
+    const updatedCaipAccountIdsToConnect = newChainsRequested
+      ? getCaip25AccountFromAccountGroupAndScope(
+          [...priorityConnectedGroups, ...connectedAccountGroups],
+          requestedCaipChainIds,
+        )
+      : connectedAccountIds;
+
```
jiexi, line 201: i don't think this takes into account `requestedCaipAccountIds`

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -227,7 +229,9 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
     if (connectedAccountGroups.length > 0) {
       return {
         suggestedAccountGroups: connectedAccountGroups,
-        suggestedCaipAccountIds: existingConnectedCaipAccountIds,
+        suggestedCaipAccountIds: hasNewChainsRequested
+          ? updatedCaipAccountIdsToConnect
+          : existingConnectedCaipAccountIds,
```
jiexi: is this ternary required since `updatedCaipAccountIdsToConnect` is already branching off if new chains were requested or not?

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
```diff
@@ -184,5 +201,9 @@ export const useAccountGroupsForPermissions = (
     supportedAccountGroups: result.supportedAccountGroups,
     /** CAIP account IDs that are already connected via existing permissions */
     existingConnectedCaipAccountIds: result.connectedCaipAccountIds,
+    /** CAIP account IDs that should be connected */
+    updatedCaipAccountIdsToConnect: result.updatedCaipAccountIdsToConnect,
+    /** Whether new chains are requested for existing account groups */
+    hasNewChainsRequested: newChainsRequested,
```
jiexi: feels like this doesn't belong in this hook/helper

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
```diff
@@ -158,6 +163,17 @@ export const useAccountGroupsForPermissions = (
...
+    // This are caip account ids of connected account groups with the requested chains/namespaces
+    // It would also include newly requested chain requests.
+    const updatedCaipAccountIdsToConnect = Array.from(
+      new Set([
+        ...getCaip25AccountFromAccountGroupAndScope(
+          [...priorityConnectedGroups, ...connectedAccountGroups],
+          requestedCaipChainIds,
+        ),
+      ]),
+    );
```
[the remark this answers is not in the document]
montelaidev: The requestedCaipAccountIds are processed when generating the connectedAccountGroups. If there is an account group and that matches the caip account id, it would be added into the connectedAccountGroups.

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
montelaidev: Yea, this can be removed

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
montelaidev: Just updated to fix this.

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -227,7 +227,7 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
     if (connectedAccountGroups.length > 0) {
       return {
         suggestedAccountGroups: connectedAccountGroups,
-        suggestedCaipAccountIds: existingConnectedCaipAccountIds,
+        suggestedCaipAccountIds: updatedCaipAccountIdsToConnect,
```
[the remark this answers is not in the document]
montelaidev: when calculating supported account groups, it would check if the `requestedCaipAccountIds`  are supported and then generate the caipAccountIds.

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -213,21 +233,49 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
+      ); // If the selected account group does not include any requested accounts, we need to show more account groups
+
     // if there are no connected account groups, show the first supported account group
-    const [firstSupportedAccountGroup] = supportedAccountGroups;
+    const selectedAndRequestedAccountGroups = supportedAccountGroups.slice(
+      0,
+      selectedAccountGroupExcludesRequestedAccounts
+        ? requestedCaipAccountIds.length + 1
+        : requestedCaipAccountIds.length,
+    );
```
jiexi: I'm not understanding this section entirely. At this point `supportedAccountGroups` should include any requested caip account ids. It should not be possible for caipAccountIds that don't exist in the wallet to make it to this point in the component/flow, so an account group would have been matched. Are you seeing otherwise?

I also don't quite understand why we are slicing the supportAccountGroups to match the length of requestedCaipAccountIds in the first place

montelaidev: Yes you're the `supportedAccountGroups` would include the requested caip account ids. We are slicing here because we don't want to show all the supported account groups during initial connection, only the requested and the currently selected account groups.

jiexi: should this be a value that is formed and returned by `useAccountGroupsForPermissions` instead? This feels like complexity that this component should not have be aware of

jiexi: >  only the requested and the currently selected account groups

the currently selected account group, or the first supported account group? `useAccountGroupsForPermissions` puts the currently selected account group at the front of the list, but that may not be a supported account group (for example, it's already possible to import and PK that will create an account group without Solana in it)

jiexi: how about something like this? https://github.com/MetaMask/metamask-mobile/pull/20194

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
```diff
@@ -169,24 +176,52 @@ export const useAccountGroupsForPermissions = (
...
+
+    // If there are no supported account groups, then return an empty array because the selected account group is not supported
+    const selectedOrFirstSupportedAccountGroup =
+      supportedAccountGroups.length > 0
+        ? [selectedAccountGroup || supportedAccountGroups[0]]
+        : [];
```
jiexi, line 206: just want to make sure this is right. The wording in a previous comment implied to me that you wanted the selectedAccountGroup here, but following the code revealed that it was possible that it was actually the first supported account group which may not actually be the selected account group. I'm okay if this is intentional, but just making sure the `supportedAccountGroups.length > 0` guard is correct

jiexi (approved):

app/components/hooks/useAccountGroupsForPermissions/useAccountGroupsForPermissions.ts
montelaidev: Yes this guard is correct. The selectedAccountGroup may not be a supported account because it may just be a private keyring or hardware wallet account. These account groups do not have solana accounts, so they wouldn't show up in the `supportedAccountGroup` array when a solana chain id is requested.

owencraston (approved):

PR #19742, opened 2025-09-15
chore: upgrade assets controllers to v75.2.0

salimtb:
## **Description**

The goal of this PR is to:
	•	Bump the assets-controllers package.

## **Changelog**

CHANGELOG entry: bump assets controllers

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```


Prithpal-Sooriya (dismissed):

app/components/Views/OnboardingSuccess/index.test.tsx
```diff
@@ -222,24 +222,6 @@ describe('OnboardingSuccess', () => {
       // wait for the useEffect side-effect to call addNetwork
       await waitFor(() => {
         expect(Engine.context.NetworkController.addNetwork).toHaveBeenCalled();
-        expect(
```
owencraston: why are these removed?

patches/@metamask+assets-controllers+74.3.2.patch
```diff
@@ -1,163 +0,0 @@
...
-     // if we've reached this point, we have successfully fetched some data for nftInformation
-@@ -766,9 +769,10 @@ async function _NftController_getNftInformationFromTokenURI(contractAddress, tok
-             standard: standard || null,
-             favorite: false,
-             tokenURI: tokenURI ?? null,
-+            error: 'URI import error',
-         };
-     }
--    const isDisplayNFTMediaToggleEnabled = __classPrivateFieldGet(this, _NftController_openSeaEnabled, "f");
-+    const isDisplayNFTMediaToggleEnabled = __classPrivateFieldGet(this, _NftController_displayNftMedia, "f");
```
owencraston: can you explain the need for this patch?

salimtb: good catch , this should not be removed , it should be replaced with patch on v75.1.0 , will update now

app/components/Views/OnboardingSuccess/index.test.tsx
salimtb: hhmm i think this need to stay , i just improved the way balances has been fetched ( read the index.tsx file ) , update pushed

app/components/Views/OnboardingSuccess/index.tsx
```diff
@@ -372,18 +372,9 @@ export const OnboardingSuccess = () => {
...
-                Logger.error(
-                  error as Error,
-                  `Failed to update balances for ${chainId}`,
-                ),
-              ),
-            ),
-          );
+          await TokenBalancesController.updateBalances({
+            chainIds: addedChainIds,
+          });
```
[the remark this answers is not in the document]
salimtb: this case is handled on the controller level , if we have any balance failure we catch this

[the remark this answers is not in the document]
salimtb: this is a perf improvement , updateBalances can take multiples chains as params , no need to use Promise.all

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -189,6 +192,19 @@ const MultichainAccountSelectorList = ({
     }
   }, [lastCreatedAccountId, flattenedData, listRefToUse]);
 
+  useEffect(() => {
+    const updateBalances = async () => {
```
Cal-L: Would recommend abstracting this into a hook instead, which will allow us to easily test the business logic

app/core/Engine/Engine.ts
```diff
@@ -906,6 +907,10 @@ export class Engine {
           assetsContractController,
         ),
       includeStakedAssets: true,
+      accountsApiChainIds: selectAssetsAccountApiBalancesEnabled(
+        store.getState(),
```
Cal-L: We should be accessing state that's passed in from the constructor instead

app/core/Engine/Engine.ts
```diff
@@ -1459,10 +1464,12 @@ export class Engine {
         // TODO: This is long, can we decrease it?
         interval: 180000,
         state: initialState.TokenBalancesController,
-        useAccountsAPI: false,
         allowExternalServices: () => isBasicFunctionalityToggleEnabled(),
         queryMultipleAccounts:
           preferencesController.state.isMultiAccountBalancesEnabled,
+        accountsApiChainIds: selectAssetsAccountApiBalancesEnabled(
+          store.getState(),
```
Cal-L: Use store from constructor instead

Cal-L (commented):
Left some comments

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
salimtb: moved to a separated PR https://github.com/MetaMask/metamask-mobile/pull/19922

owencraston (dismissed):

patches/@metamask+assets-controllers+75.2.0.patch
```diff
@@ -161,3 +161,28 @@ index 4a81bbc..6cd5476 100644
...
+ exports.calculateBalanceChangeForAccountGroup = exports.calculateBalanceChangeForAllWallets = exports.calculateBalanceForAllWallets = void 0;
+-const account_api_1 = require("@metamask/account-api");
+ const keyring_api_1 = require("@metamask/keyring-api");
+ const utils_1 = require("@metamask/utils");
+ const AssetsContractController_1 = require("./AssetsContractController.cjs");
+@@ -26,8 +25,11 @@ const isChainEnabledByMap = (map, id) => {
+     const { namespace } = (0, utils_1.parseCaipChainId)(id);
+     return Boolean(map[namespace]?.[id]);
+ };
++const getWalletIdFromGroupId = (groupId) => {
```
Gudahtt: What is the purpose of this addition to the patch?

app/core/Engine/Engine.ts
```diff
@@ -1457,12 +1462,14 @@ export class Engine {
           ],
         }),
         // TODO: This is long, can we decrease it?
-        interval: 180000,
+        interval: 10000,
```
Gudahtt: This will making polling 18x more frequent, significantly increasing the number of network calls. We have a lot of problems and regressions in this area right now (there are **way** more network calls made than are reasonable across the entire client), so I am not sure changing this here in the same PR is a good idea.

Maybe we can consider this change separately so that it's easier to test in isolation and revert if need be.

Gudahtt: Also if we decrease this, we should remove the TODO comment about decreasing it 😅

salimtb: ooh we wanted to refresh balances more frequently , but i'm fine to put it back to 3min , will do it now

Gudahtt: Yeah, refreshing balances more frequently than 3 minutes is probably a good idea! Just trying to be cautious about how we do it.

package.json
```diff
@@ -225,7 +225,7 @@
     "@metamask/address-book-controller": "^6.1.0",
     "@metamask/app-metadata-controller": "^1.0.0",
     "@metamask/approval-controller": "^7.1.3",
-    "@metamask/assets-controllers": "^74.3.2",
+    "@metamask/assets-controllers": "^75.2.0",
```
Gudahtt, line 228: Could you update `bridge-controller` and `bridge-status-controller` to the latest major version here as well? They have a peer dependency on `assets-controllers`, so if they are not updated here they will emit a new peer dependency warning on install

patches/@metamask+assets-controllers+75.2.0.patch
salimtb: i added this because i wanted to avoid the bump of the accout-api , fixed now

package.json
salimtb: for bridge controller , there's already a PR for here https://github.com/MetaMask/metamask-mobile/pull/19913 , will do it for the bridge status controller

salimtb: done [8c8cd98](https://github.com/MetaMask/metamask-mobile/pull/19742/commits/8c8cd982e8972a8c62ddfb9dde4a749fac334624)

Gudahtt: That PR does update the `bridge-controller`, but we can't update these independently without introducing a peer dependency warning (which to clarify are meant to guard against introducing incompatible versions, and we intended on making these errors in the near future)

When packages have peer dependencies on each other like this, they should be updated in the same PR. In this case, the only breaking changes to the bridge packages is the `assets-controllers` bump. Therefore there should be no functional changes needed, just the update.

salimtb:
e2e tests are also failng on main:
https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/9dd96f6b-d31f-49c8-82f9-8e7008e9f948?tests_filter_status=failed

Cal-L (dismissed):
LGTM

bfullam (approved):

Prithpal-Sooriya (approved):

app/core/Engine/Engine.ts
```diff
@@ -1459,10 +1464,12 @@ export class Engine {
         // TODO: This is long, can we decrease it?
         interval: 180000,
         state: initialState.TokenBalancesController,
-        useAccountsAPI: false,
         allowExternalServices: () => isBasicFunctionalityToggleEnabled(),
         queryMultipleAccounts:
           preferencesController.state.isMultiAccountBalancesEnabled,
+        accountsApiChainIds: selectAssetsAccountApiBalancesEnabled({
+          engine: { backgroundState: initialState },
+        }) as `0x${string}`[],
```
tommasini, line 1472: What's the reason for the selector do not return the expected type?

salimtb: because the selector can’t guarantee the type because the value comes from a feature flag that may hold anything (string, number, invalid). That’s why TS infers a broad type.

tommasini (approved):

jvbriones (approved):

pnarayanaswamy (approved):

PR #19880, opened 2025-09-16
fix: Initial scroll to selected account for MultichainAccountSelectorList

owencraston:
## **Description**

- fixes a bug that causes the list to constantly scroll back to the selected account, even if the user is already scrolling on the list
- instead of a useEffect to trigger the scroll in favour of the [initialScrollIndex](https://shopify.github.io/flash-list/docs/usage/#initialscrollindex) prop from FlashList. This allows FlashList to handle the state of the selected account which means if does not get triggered on dependancy updates. This also offers slightly better UX as the list immediately renders with the selected account visible instead of scrolling to the selected index. 

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: N/A

## **Manual testing steps**

```gherkin
Feature: my feature name

  Scenario: user User scrolling on Multichain account selector list
    Given I am a user with a wallet that has many accounts (over 10)
    And I am testing on a lower end device

    When user open the account list
    Then the list opens immediately with the selected account in view
    And when the user scrolls immediately after the list renders, there is not scroll event triggered

```

### **Before**

https://github.com/user-attachments/assets/2280f121-af31-4fcc-84e8-cd7ecd12edc9

### **After**

https://github.com/user-attachments/assets/e8669781-b24f-44dd-ba47-3bae27e93cf0

https://github.com/user-attachments/assets/0cba0625-2b30-4c41-acd8-bb370d6da7d0

### DIFF
--- a/app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.test.tsx
+++ b/app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.test.tsx
@@ -1,13 +1,12 @@
 import React from 'react';
-import { fireEvent, waitFor } from '@testing-library/react-native';
+import { fireEvent, waitFor, within } from '@testing-library/react-native';
+import '@shopify/flash-list/jestSetup';
 import {
   AccountGroupObject,
   AccountWalletObject,
 } from '@metamask/account-tree-controller';
 import { InternalAccount } from '@metamask/keyring-internal-api';
 import MultichainAccountSelectorList from './MultichainAccountSelectorList';
-import { FlashListRef } from '@shopify/flash-list';
-import { FlattenedMultichainAccountListItem } from './MultichainAccountSelectorList.types';
 import renderWithProvider from '../../../../util/test/renderWithProvider';
 import {
   MULTICHAIN_ACCOUNT_SELECTOR_SEARCH_INPUT_TESTID,
@@ -20,6 +19,7 @@ import {
   createMockInternalAccountsFromGroups,
   createMockInternalAccountsWithAddresses,
 } from '../test-utils';
+import { ReactTestInstance } from 'react-test-renderer';
 
 jest.mock('../../../../core/Engine', () => ({
   context: {
@@ -117,7 +117,7 @@ describe('MultichainAccountSelectorList', () => {
     const { getByText } = renderComponentWithMockState(
       [wallet1, wallet2],
       internalAccounts,
-      [account1],
+      [],
     );
 
     expect(getByText('Wallet 1')).toBeTruthy();
@@ -145,7 +145,7 @@ describe('MultichainAccountSelectorList', () => {
     const { getByText } = renderComponentWithMockState(
       [srpWallet, snapWallet],
       internalAccounts,
-      [srpAccount],
+      [],
     );
 
     expect(getByText('Wallet 1')).toBeTruthy();
@@ -173,7 +173,7 @@ describe('MultichainAccountSelectorList', () => {
     const { getByText } = renderComponentWithMockState(
       [srpWallet, ledgerWallet],
       internalAccounts,
-      [srpAccount],
+      [],
     );
 
     expect(getByText('Wallet 1')).toBeTruthy();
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -339,7 +339,7 @@
     "@sentry/core": "~8.54.0",
     "@sentry/react": "~8.54.0",
     "@sentry/react-native": "~6.10.0",
-    "@shopify/flash-list": "2.0.1",
+    "@shopify/flash-list": "2.0.3",
```
owencraston, line 342: [Release info](https://github.com/Shopify/flash-list/releases/tag/v2.0.3). Needed to fix the unit tests since the mock that was exported from flash-list was returning undefined.

PatrykLucka (approved):
That's a great update! LGTM ✅

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.test.tsx
```diff
@@ -173,7 +173,7 @@ describe('MultichainAccountSelectorList', () => {
     const { getByText } = renderComponentWithMockState(
       [srpWallet, ledgerWallet],
       internalAccounts,
-      [srpAccount],
+      [],
```
owencraston, line 176: I had to remove this since having the selected account here meant that the other list items were not in the initial view and therefore not searchable. Nothing functionally changed though.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.test.tsx
```diff
@@ -717,7 +721,7 @@ describe('MultichainAccountSelectorList', () => {
       expect(getByText('Create account')).toBeTruthy();
     });
 
-    it('scrolls to the first selected account', async () => {
+    it('positions the list so the first selected account is initially visible', () => {
```
owencraston, line 724: I verified that this test fails if we remove the auto scroll logic

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -54,7 +54,8 @@ const MultichainAccountSelectorList = ({
   const [lastCreatedAccountId, setLastCreatedAccountId] = useState<
     string | null
   >(null);
-  const internalListRef = useRef(null);
+  const internalListRef =
+    useRef<FlashListRef<FlattenedMultichainAccountListItem>>(null);
```
owencraston, line 58: just better typing on this file. Not strictly needed.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -155,6 +156,16 @@ const MultichainAccountSelectorList = ({
     return items;
   }, [filteredWalletSections]);
 
+  // Compute first selected account index for initial positioning only
+  const initialSelectedIndex = useMemo(() => {
+    const targetId = selectedAccountGroups?.[0]?.id;
+    if (!targetId) return undefined;
+    const idx = flattenedData.findIndex(
```
owencraston, line 163: returns the first available index that matches the search query so this will not break for multi select.

app/component-library/components-temp/MultichainAccounts/MultichainAccountSelectorList/MultichainAccountSelectorList.tsx
```diff
@@ -155,6 +156,16 @@ const MultichainAccountSelectorList = ({
...
+      (item) => item.type === 'cell' && item.data.id === targetId,
+    );
+    return idx > 0 ? idx : undefined;
```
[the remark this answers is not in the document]
owencraston: I thought this too but after some thought this makes more sense. The default behaviour of the FlashList is to render the list at the top (0 index). Therefore by returning undefined when the selected account is the 0 index, we are simply saying, "resort to your default behaviour". I did not verify this but my assumption was that there are some small performance benefits to not providing that prop in this default case. This is why I opted to return undefined here.

PR #19954, opened 2025-09-18
feat: claim reward flow

sophieqgu:
## **Description**
- Refactor reward items and tiers to navigate to bottom sheet show
   -- shortDescription and longDescription for locked rewards
   -- shortUnlockedDescription and longUnlockedDesciption for unlocked rewards
   -- CTA to claim for unlocked and claimable rewards 
   -- input field and claimUrl depending on reward types

- Fix onboarding navigation
- Add cache invalidation logic to RewardsController

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **After**

https://github.com/user-attachments/assets/29770f24-9a7c-4992-ba0d-43e8e636a2fd

https://github.com/user-attachments/assets/df9059e6-a2b0-4c73-99cb-5c43bbffcabe

Simulator Screenshot - E2E Test  - 2025-09-17 at 13 18 55
Simulator Screenshot - E2E Test  - 2025-09-17 at 14 05 04
Simulator Screenshot - E2E Test  - 2025-09-18 at 14 14 47
Simulator Screenshot - E2E Test  - 2025-09-18 at 18 33 58


app/components/UI/Rewards/components/Onboarding/OnboardingStep4.tsx
```diff
@@ -56,9 +56,8 @@ const OnboardingStep4: React.FC = () => {
   }, [optin, referralCode]);
 
   const handlePrevious = useCallback(() => {
-    // Reset onboarding state so we can jump from step 1 to 5 next time
-    dispatch(setOnboardingActiveStep(OnboardingStep.INTRO));
-    navigation.navigate(Routes.WALLET_VIEW);
+    dispatch(setOnboardingActiveStep(OnboardingStep.STEP_3));
+    navigation.navigate(Routes.REWARDS_ONBOARDING_3);
```
VGR-GIT, line 60: Was this done based on a comment from our own tests? I had a comment from Tomas Santos last week where he thought it was very strange to go back in the onboarding flow, instead of going back to the wallet view.

app/components/UI/Rewards/hooks/useClaimReward.ts
```diff
@@ -0,0 +1,80 @@
...
+      if (!subscriptionId) {
+        setClaimRewardError('No subscription found. Please try again.');
+        return;
+      }
+
+      try {
+        setIsClaimingReward(true);
+        setClaimRewardError(undefined);
+
+        // Call the RewardsController method directly since the action handler is not registered yet
```
VGR-GIT: redundant comment

app/components/UI/Rewards/components/Tabs/LevelsTab/RewardsClaimBottomSheetModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }, [
+    handleClaimReward,
+    handleModalClose,
+    isClaimingReward,
+    isLocked,
+    rewardType,
+  ]);
+
+  return (
+    <RewardsBottomSheetModal
```
VGR-GIT: I would have this be its own thing and keep the implementation of RewardsBottomSheetModal as-is. We're trying to make it too generic and it's already become somewhat of a frankenstein component.

We also need to a way to show, on the confirm button, that we're loading/claim is in progress. + Do we also need to show toasters for these actions? I would have all these things live in this component.

app/constants/navigation/Routes.ts
```diff
@@ -95,6 +95,7 @@ const Routes = {
     MULTICHAIN_ACCOUNTS_INTRO: 'MultichainAccountsIntroModal',
     MULTICHAIN_ACCOUNTS_LEARN_MORE: 'MultichainAccountsLearnMoreBottomSheet',
     REWARDS_BOTTOM_SHEET_MODAL: 'RewardsBottomSheetModal',
+    REWARDS_CLAIM_BOTTOM_SHEET_MODAL: 'RewardsClaimBottomSheetModal',
```
VGR-GIT, line 98: One thing that seems weird to me from a UX POV is that these bottom sheet modals cause us to actually navigate back to a view when closed/confirmed. This can lead to weird situations where you were on a specific part of a view (say you scrolled down to a specific reward) and then when you close the claim modal, the view reloads and you're back at top part of that view.

I'm not sure if Andrew knows it works like that. If this isn't a problem then OK, but otherwise I think we need to surface this issue (also applies to other usages of `REWARDS_BOTTOM_SHEET_MODAL`) and maybe look at something similar but without the navigation

app/components/UI/Rewards/components/Tabs/LevelsTab/UpcomingRewards.tsx
```diff
@@ -31,58 +41,187 @@ interface TierAccordionProps {
...
+        return '';
+    }
+  }, [seasonReward.rewardType]);
+
+  const rewardClaimUrl = useMemo(() => {
+    if (isLocked || !currentAccountAddress || !seasonReward.claimUrl)
+      return undefined;
+    return seasonReward.claimUrl.replace(
+      '{address}',
+      currentAccountAddress || '',
```
VGR-GIT: I couldn't find such an endpoint in the backend? Is this something that is yet to be implemented or what agreed to be built soon? Or am I missing something?

Also, we should already base this around the bip-44 structure. That means:

- Use an account tree controller selector to get the current account group 
- Figure out what the network is for this reward (if any?) to figure out the correct account in the group to use when claiming.

app/selectors/rewards/index.ts
```diff
@@ -41,5 +41,15 @@ export const selectRewardsActiveAccountHasOptedIn = createSelector(
     rewardsControllerState.activeAccount?.hasOptedIn ?? null,
 );
 
+export const selectRewardsActiveAccountAddress = createSelector(
```
VGR-GIT, line 44: we don't need if we use the approach that I proposed here https://github.com/MetaMask/metamask-mobile/pull/19954/files#r2357979685

app/components/UI/Rewards/hooks/useSeasonStatus.ts
```diff
@@ -46,4 +46,34 @@ export const useSeasonStatus = (): void => {
   useEffect(() => {
     fetchSeasonStatus();
   }, [fetchSeasonStatus]);
+
+  // Listen for account linked events to trigger refetch
+  useEffect(() => {
+    Engine.controllerMessenger.subscribe(
```
VGR-GIT: I'm not sure how these kind of usages in multiple hooks will scale if we link a lot of accounts in 1 CTA press. We might need a different event for that, so that's one way to mitigate. We can leave it as-is for now.

VGR-GIT: One thing we could improve here is the reduction of duplicate code across all these hooks. Ideally there'd be some kind of generic rewards hook (i.e. useInvalidateByRewardEvents) that:

- takes an array of reward controller specific events (can likely be type checked)
- takes a function to be executed when any of these trigger.

Internally it would do pretty much what you do here, but for all the events that were specified. (both subscribing and unsubscribing):

```
 useEffect(() => {
    Engine.controllerMessenger.subscribe(
      'RewardsController:rewardClaimed',
      fetchSeasonStatus,
    );

    return () => {
      Engine.controllerMessenger.unsubscribe(
        'RewardsController:rewardClaimed',
        fetchSeasonStatus,
      );
    };
  }, [fetchSeasonStatus]);
```

VGR-GIT (changes requested):
See comments

app/components/UI/Rewards/components/Onboarding/OnboardingStep4.tsx
sophieqgu: Oh! I didn't know there was a comment about this. I thought it was a mistake during a merge conflict or something..

sophieqgu: And I realized the designs have changed.. It now contains a 'Skip button' that I assume goes directly to STEP4 and no back button

VGR-GIT: Weird, I assume users should always have the ability to go back to main wallet view since the onboarding flow is full screen?

sophieqgu: Looks like the changes are big enough to do in a later PR. https://consensys.slack.com/archives/C09372GAN59/p1758204242369369

app/components/UI/Rewards/components/Tabs/LevelsTab/UpcomingRewards.tsx
sophieqgu: This is very specific to the reward 'Priority support' which includes a claimUrl in the format 'https://support.metamask.io?address={address}'

app/constants/navigation/Routes.ts
sophieqgu: Yes this is very noticeable and all the components are re-rendered after navigating back, which is annoying. It looks like other modals such as clicking on the 'Trade' button in bottom nav doesn't do that. It simply overlays on the top and the background components are not re-rendered. I'll look into those.

app/selectors/rewards/index.ts
sophieqgu: I think we need an active address that is associated with a subscription, not just any account that is currently selected?

app/components/UI/Rewards/components/Tabs/LevelsTab/RewardsClaimBottomSheetModal.tsx
sophieqgu: Reverted changes to RewardsBottomSheetModal and kept RewardsClaimBottomSheetModal separate component

app/selectors/rewards/index.ts
sophieqgu: Waiting on feedback from Christian: https://consensys.slack.com/archives/C09372GAN59/p1758213707460859?thread_ts=1758211204.660839&cid=C09372GAN59

app/components/UI/Rewards/hooks/useSeasonStatus.ts
sophieqgu: Extracted to a generic useInvalidateByRewardEvents hook

VGR-GIT (approved):

PR #19970, opened 2025-09-18
feat: enhance wallet alignment process in MultichainAccountsIntroModal

PatrykLucka:
## **Description**
This PR adds `alignWallets()` call to "View accounts" button handler of Multichain Accounts educational modal.

## **Changelog**

CHANGELOG entry: Added loading state to View accounts call from Multichain Accounts educational modal

## **Related issues**

Fixes: 
Jira tickets:
https://consensyssoftware.atlassian.net/browse/MUL-897
https://consensyssoftware.atlassian.net/browse/MUL-894

## **Manual testing steps**

```gherkin
Feature: Show Multichain Accounts educational modal

  Scenario: user unlocks the app
    Given app has just been updated and user hasn't seen the modal

    When user unlocks the app
    Then new educational modal is shown with the loading state after pressing "View accounts"
```

## **Screenshots/Recordings**

https://github.com/user-attachments/assets/5a268085-ba9c-4b8c-89fa-9bd56980cfdd


app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
```diff
@@ -21,22 +21,67 @@ import { useStyles } from '../../../../component-library/hooks';
...
-    navigation.navigate(...createAccountSelectorNavDetails({}));
-  }, [navigation, dispatch]);
+  const handleViewAccounts = useCallback(async () => {
+    if (isAligning) return;
+
+    setIsAligning(true);
+
+    try {
+      const timeoutPromise = new Promise<void>((resolve) => {
+        setTimeout(() => resolve(), 2000);
```
mathieuartu: Can we maybe store this in a constant with a comprehensive name?

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
```diff
@@ -21,22 +21,67 @@ import { useStyles } from '../../../../component-library/hooks';
...
+      });
+
+      const alignWalletsPromise =
+        Engine.context.MultichainAccountService.alignWallets();
+
+      // Use Promise.all to wait for both the alignment and the minimum timeout
+      await Promise.all([alignWalletsPromise, timeoutPromise]);
```
mathieuartu: Hmmm... What about if `alignWalletsPromise` resolves earlier than 2 secs? Should we use `Promise.race`?

mathieuartu: @PatrykLucka what's the purpose of this? Do we want users to wait for at least 2 seconds before navigating to the next screen?

gantunesr: This is just a lower threshold and it was an arbitrary decision on extension that is being also used un mobile, I believe the purpose of it is merely for UX. We could change the solution to just use `Promise.race`

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
```diff
@@ -21,22 +21,67 @@ import { useStyles } from '../../../../component-library/hooks';
...
+
+    setIsAligning(true);
+
+    try {
+      const timeoutPromise = new Promise<void>((resolve) => {
+        setTimeout(() => resolve(), 2000);
+      });
+
+      const alignWalletsPromise =
+        Engine.context.MultichainAccountService.alignWallets();
```
gantunesr, line 68: We should trigger the alignment when the modal is displayed, the process should be running while the user is still reading. If they try to "View accounts" then we can show the loading state

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
fabiobozzo: Something like I did on extension? 
```
const MINIMUM_LOADING_DURATION = 2000; // ms
const alignWalletsPromise = Engine.context.MultichainAccountService.alignWallets();
const timeoutPromise = new Promise<void>((resolve) => {
  setTimeout(() => resolve(), MINIMUM_LOADING_DURATION);
});
await Promise.race([alignWalletsPromise, timeoutPromise]);
```

fabiobozzo (dismissed):
LGTM. 

I wonder whether we should start alignment when the modal opens, and show loading then only if still in progress. 🤔 (also for 'my' extension dialog PR)

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
fabiobozzo: Ah, I wrote a comment about that too, but didn't notice Guto wrote the same already 😓

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
PatrykLucka: exactly - wait at least 2 seconds, even if alignement finishes early

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
```diff
@@ -21,22 +21,87 @@ import { useStyles } from '../../../../component-library/hooks';
...
+    try {
+      // Wait for both the alignment promise and minimum timeout
+      await Promise.all([
+        alignWalletsPromise,
+        new Promise<void>((resolve) => {
+          setTimeout(() => resolve(), WALLET_ALIGNMENT_MINIMUM_TIMEOUT_MS);
+        }),
+      ]);
+    } catch (error) {
+      Logger.error(error as Error, 'Error in multichain accounts intro modal');
```
ccharly: I wonder if we want to capture this error in Sentry 🤔 The alignment should never really fail, but if it does, then the wallet might be a in weird state.

So having some error reporting on such critical operation could be valuable (especially the first one for the migration)

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.test.tsx
```diff
@@ -170,4 +189,102 @@ describe('MultichainAccountsIntroModal', () => {
...
+        fireEvent.press(viewAccountsButton);
+      });
+
+      // Wait for completion - this should take at least 2 seconds due to minimum timeout
+      await waitFor(
+        () => {
+          expect(mockGoBack).toHaveBeenCalledTimes(1);
+        },
+        { timeout: 5000 },
+      );
```
ccharly: I wonder if we should have track of the timing here and use `expect` to check that the whole rendering + action (or maybe just the click?) took "at least 2s" (I would use the constant `WALLET_ALIGNMENT_MINIMUM_TIMEOUT_MS` in here too, so it's pretty explicit)

ccharly (dismissed):
Code LGTM! I'll try to make a quick test also, but pre-approving in case we run out of time 😄

app/components/Views/MultichainAccounts/IntroModal/MultichainAccountsIntroModal.tsx
```diff
@@ -21,22 +21,89 @@ import { useStyles } from '../../../../component-library/hooks';
...
+            {strings('multichain_accounts.intro.setting_up_accounts')}
+          </Text>
+        </View>
+      );
+    }
+    return strings('multichain_accounts.intro.view_accounts_button');
+  };
 
   const handleClose = useCallback(() => {
     dispatch(setMultichainAccountsIntroModalSeen(true));
```
ccharly, line 84: Actually, just realized that now, but closing should still be waiting for alignment to proceed? Or is that ok to "force-close" this? (the alignment has started already anyway)

ccharly (approved):

PR #19985, opened 2025-09-18
fix: non blocking Account sync & discovery operations when importing a new SRP

mathieuartu:
## **Description**

This PR moves the account syncing & discovery operations to an IIFE and executes it in a non-blocking way.
This lets the user imports a new SRP without having to wait for backup & sync and discovery to complete.
Analytics are sent through a callback when all the operations are finished.

## **Changelog**

CHANGELOG entry: improve import srp performance

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUL-902

## **Manual testing steps**

```gherkin
Feature: Import new SRP

  Scenario: user adds a new SRP
    Given any wallet state

    When user taps "Continue" after importing
    Then it goes directly to the home page without a long wait
```


app/actions/multiSrp/index.ts
```diff
@@ -133,20 +139,25 @@ export async function importNewSecretRecoveryPhrase(
...
+        // Then we discover accounts
+        discoveredAccountsCount = await discoverAccounts(newKeyring.id);
+      } catch (error) {
+        captureException(
+          new Error(`Unable to sync, discover and create accounts: ${error}`),
+        );
+        discoveredAccountsCount = 0;
+      } finally {
+        // We trigger the callback with the results, even in case of error (0 discovered accounts)
+        callback?.({ address: newAccountAddress, discoveredAccountsCount });
```
ccharly: Nit: I wonder if we should also capture `error` and then it to the `callback` 🤔 Now that we are not awaiting this promise anymore, we don't have any way to know if the import has been complete and really successful.

WDYT?

app/actions/multiSrp/index.ts
```diff
@@ -133,20 +139,25 @@ export async function importNewSecretRecoveryPhrase(
     }
   }
 
-  let discoveredAccountsCount: number;
+  let discoveredAccountsCount: number = 0;
```
ccharly, line 147:
```suggestion
  // If state 2 is enabled, this function will return 0 discovered account
  // immediately, so we have to use the `callback` instead to get this
  // information. 
  let discoveredAccountsCount: number = 0;
```

app/actions/multiSrp/index.ts
```diff
@@ -26,16 +26,22 @@ import { discoverAccounts } from '../../multichain-accounts/discovery';
...
+  discoveredAccountsCount: number;
+}
+
 export async function importNewSecretRecoveryPhrase(
   mnemonic: string,
   options: ImportNewSecretRecoveryPhraseOptions = {
     shouldSelectAccount: true,
   },
-) {
+  callback?: (options: ImportNewSecretRecoveryPhraseReturnType) => void,
```
ccharly: I think this function has to be `async`, so the caller can use `await` if needed. WDYT?

app/components/Views/ImportNewSecretRecoveryPhrase/index.test.tsx
```diff
@@ -259,6 +303,35 @@ describe('ImportNewSecretRecoveryPhrase', () => {
     );
   });
 
+  it('tracks IMPORT_SECRET_RECOVERY_PHRASE_COMPLETED event on successful import (state 2)', async () => {
```
ccharly: Nit: We have used this in other "state 2 only" tests:

```suggestion
  it('(state 2) - tracks IMPORT_SECRET_RECOVERY_PHRASE_COMPLETED event on successful import', async () => {
```

ccharly (approved):
LGTM + Tested 💪

PR #19990, opened 2025-09-18
chore: enhanced tab to lazy load and scrollable cp-7.56.0

### LINKED ISSUE #20368: [Bug]: Wallet line animation under tabs starts working incorrectly after using the app for a while
### Describe the bug

Just like it happens for this bug https://github.com/MetaMask/metamask-mobile/issues/20363 after after minimizing the app for a while and doing something else, when you reopen the app the line animation below the tabs starts working incorrectly. See video

### Expected behavior

_No response_

### Screenshots/Recordings

https://github.com/user-attachments/assets/f4b54de6-a1c7-4a30-bf37-a638b743ec0f

### Steps to reproduce

1. Use the app for a while 
2. Notice that the line animation under the tabs stop working correctly (see video)

```shell

```

### Detection stage

In production (default)

### Version

7.56.0

### Build type

None

### Device

iPhone Pro

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

brianacnguyen:
## **Description**
This PR enhances the new tab to lazy load inactive contents and enable the scrollable swipe between tabs

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/19869, https://github.com/MetaMask/metamask-mobile/issues/20368

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **After**
https://github.com/user-attachments/assets/6f52482a-3ba0-41ae-bf11-0a670a1305c1


app/component-library/components-temp/Tabs/TabsList/TabsList.tsx
```diff
@@ -92,150 +189,247 @@ const TabsList = forwardRef<TabsListRef, TabsListProps>(
         const firstEnabledIndex = tabs.findIndex((tab) => !tab.isDisabled);
         setActiveIndex(firstEnabledIndex >= 0 ? firstEnabledIndex : -1);
       }
-      // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [initialActiveIndex, tabs]);
+    }, [initialActiveIndex, tabs, activeIndex]);
```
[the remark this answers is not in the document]
brianacnguyen: This is needed

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
```diff
@@ -137,6 +137,185 @@ describe('Tab', () => {
...
+      expect(toJSON()).toMatchSnapshot();
+    });
+
+    it('shows alternative text color when inactive and not disabled', () => {
+      const { toJSON } = render(
+        <Tab {...defaultProps} isActive={false} isDisabled={false} />,
+      );
+      expect(toJSON()).toMatchSnapshot();
+    });
+  });
```
georgewrmarshall: non-blocking: These tests aren't really asserting the specific text colors they're just checking that the snapshot doesn't change. I think they can be removed.

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
```diff
@@ -137,6 +137,185 @@ describe('Tab', () => {
...
+          accessibilityLabel="Custom accessibility label"
+          accessibilityHint="Custom accessibility hint"
+          testID="accessible-tab"
+        />,
+      );
+
+      const tab = getByTestId('accessible-tab');
+      expect(tab).toBeOnTheScreen();
+    });
+  });
```
georgewrmarshall, line 248: non-blocking: this test isn't asserting for the accessibility props at all. Either update to assert for the accessibilityLabel and accessibilityHint it or remove it.

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
```diff
@@ -137,6 +137,185 @@ describe('Tab', () => {
...
+
+  describe('Edge Cases', () => {
+    it('handles empty label gracefully', () => {
+      const { toJSON } = render(<Tab {...defaultProps} label="" />);
+      expect(toJSON()).toMatchSnapshot();
+    });
```
georgewrmarshall, line 257: non-blocking: this test could be improved by using a test id and asserting that the tab is in the document or defined instead of a snapshot.

app/component-library/components-temp/Tabs/Tab/Tab.tsx
```diff
@@ -19,52 +19,77 @@ const Tab: React.FC<TabProps> = ({
   isDisabled = false,
   onPress,
   testID,
+  onLayout,
```
georgewrmarshall, line 22: Are we missing `onLayout` from `TabProps` in Tab.types.ts?

app/component-library/components-temp/Tabs/Tab/Tab.tsx
```diff
@@ -19,52 +19,77 @@ const Tab: React.FC<TabProps> = ({
...
-        )
+  // Use callback ref and memoized layout handler to minimize unnecessary re-renders and improve tab performance.
+
+  // Simple callback ref to store the view reference
+  const handleViewRef = useCallback((view: View | null) => {
+    // Store ref for other uses
+    if (viewRef.current !== view) {
+      (viewRef as React.MutableRefObject<View | null>).current = view;
+    }
+  }, []);
```
georgewrmarshall: This callback ref pattern seems unnecessarily complex why aren't we just passing the ref directly to the `View`?

```
<View ref={viewRef}
```

app/component-library/components-temp/Tabs/Tab/Tab.tsx
```diff
@@ -19,52 +19,77 @@ const Tab: React.FC<TabProps> = ({
...
-      </Text>
+        {/* Hidden bold text that determines layout size */}
+        <Text
+          variant={TextVariant.BodyMd}
+          fontWeight={FontWeight.Bold}
+          numberOfLines={1}
+          style={tw.style('opacity-0')}
+        >
+          {label}
+        </Text>
```
georgewrmarshall, line 61: non-blocking: Is this really the only way we can ensure that there isn't any layout jumping? Seems like a lot to render an invisible element just for layout purposes. We can revisit this approach in a separate PR

app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx
```diff
@@ -139,6 +214,14 @@ const TabsBar: React.FC<TabsBarProps> = ({
     animateToTab(activeIndex);
   }, [activeIndex, animateToTab]);
 
+  // Trigger animation when all layouts become available
+  useEffect(() => {
+    if (hasAllTabLayouts && activeIndex >= 0) {
+      // All layouts available, trigger animation
+      animateToTab(activeIndex);
+    }
+  }, [hasAllTabLayouts, activeIndex, animateToTab]);
```
georgewrmarshall: non-blocking: I noticed this file has a lot of useEffects that call `animatedToTab`. I wasn't sure if this would trigger a lot of re-renders or possibly an infinite loops so added some logs. It seems to be ok but we might be able to get further performance gains in a subsequent PR.

https://github.com/user-attachments/assets/7a49c69c-a27c-4743-a0dd-c281513e508a

georgewrmarshall (commented):
Looking good! A couple small things and I can't seem to test on the app because of this issue in mobile https://consensys.slack.com/archives/C02U025CVU4/p1758747378027869. Is it possible to get the e2e tests to pass?

app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx
brianacnguyen: streamlined

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
```diff
@@ -137,6 +137,162 @@ describe('Tab', () => {
...
+      const { getAllByText } = render(
+        <Tab {...defaultProps} onPress={mockOnPress} />,
+      );
+
+      const tab = getAllByText('Test Tab')[0];
+      expect(() => {
+        fireEvent.press(tab);
+      }).not.toThrow();
+      expect(mockOnPress).toHaveBeenCalledTimes(1);
+    });
```
georgewrmarshall, line 285: non-blocking: I don't understand what this test is doing. `handles undefined onPress gracefully` but onPress is being passed a mock function so it's not `undefined`?

georgewrmarshall: non-blocking: I don't understand what this test is doing. `handles undefined onPress gracefully` but onPress is being passed a mock function so it's not `undefined`?

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
```diff
@@ -137,6 +137,162 @@ describe('Tab', () => {
...
+
+      const tab = getByTestId('ref-tab');
+      expect(tab).toBeOnTheScreen();
+
+      // Re-render to trigger ref callback
+      rerender(<Tab {...defaultProps} label="Updated Tab" testID="ref-tab" />);
+
+      const updatedTab = getByTestId('ref-tab');
+      expect(updatedTab).toBeOnTheScreen();
+    });
```
georgewrmarshall: non-blocking: I think this test can be removed. The ref callback has been removed and this test is still passing. It's not asserting for the ref or the ref callback

georgewrmarshall (dismissed):
Looking a lot better nice work! Couple small things but happy to approve this

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
brianacnguyen: Good catch!

app/component-library/components-temp/Tabs/Tab/Tab.test.tsx
brianacnguyen: Good catch!

georgewrmarshall (dismissed):

juanmigdr (commented):
Hey! Nice work, performance is much better than before! I have been testing it and noticed a few things:

**ISSUE 1:**
- There is no way to refresh tabs by swiping up (expected) and that is why on-tab-change we are supossed to refresh the list of tokens. When switching between Tokens and Perps I can see the Tokens view is still present and not re-rendered (expected) but I do not see any kind of loading spinner so I am just wondering if a data refresh is even being triggered at all

**ISSUE 2**
- As you ill see in the video, the perps tab is rendered once and when coming back to the tab it is re-rendered again (this is not expected)
-  BUT what is very weird is that if I navigate to it for a third time, it is not being re-rendered (which is the correct behavior) so it seems flaky

**NOTE**
As you can see if we switch between the Nfts and Perps tabs, the NFTs UI is not re-rendered and only data is being fetched and that is why you only see the loading spinner being re-rendered. This is the expected behavior in Tokens and what I would expect to see, a spinner at the bottom of the list showing data is being re-fetched


https://github.com/user-attachments/assets/507dee08-f941-4fe9-a50e-edaccda049dd

juanmigdr:
**ISSUE 3.1**

When using an account with quite a few NFTs, the whole tab navigation becomes very laggy even if that tab is not opened. It seems like the tab navigation is rendering all tabs as soon as the user opens the app and even though they are on the Tokens screen, the whole navigation becomes laggy due to rendering such a big amount of NFTs.

In the following video I click a tab and move my mouse away and you will see the lag when navigating between screens:

https://github.com/user-attachments/assets/db311119-d115-4b6e-875a-3f1ee87ce724

**ISSUE 3.2**
I can see that whenever the home screen is loaded, all tabs are being rendered. I would have expected that they are lazy loaded when each tab is opened for the first time but as seen below as soon as the app is opened the nfts tab is loaded:

https://github.com/user-attachments/assets/e837746d-8b33-4f0c-b5f8-5f5fb52ea939

Even if this is fixed, I believe once the NFTs view is opened for the first time, the same lagging is going to appear when switching in-between tabs. But I still think we should not render nfts if the tab is not opened. WDYT?

juanmigdr:
**ISSUE 4**
I have just identified that we are executing NFT detection whenever we switch between the `Tokens` and `Perps` tabs (Look at `console.warn` on line 1171) this should not be the case since we are triggering quite a few api calls for no reason:

https://github.com/user-attachments/assets/ee6e3faf-6b7b-47a4-afb9-3f1f130ebcea

juanmigdr:
Retested:
- ISSUE 1 - I do not think this is resolved, but maybe I am missing something
- ISSUE 2 - Still can be reproduced
- ISSUE 3.1 - Now the lag is even more noticeable because there is an animated tab moving and if you have a few NFTs it can be noticed easily
- ISSUE 3.2 - ✅ Fixed!
- ISSUE 4 - Can still be reproduced

juanmigdr:
I tested it and it LGTM, great work!! 🚀 

PS: As we mentioned in the call, there are 2 improvements that should be made after merging this PR:
- Rn, there is no way to refresh tokens rather tan opening and closing the app, something like a refresh control or tab switch refresh should be implemented
- NFT detection degrades performance of the tabs due to the amount of API requests being made, this is an existing issue

juanmigdr:
Hey, update on this point:

> 2/ We noticed that, if the NFT list is large, it becomes laggy. When we comment out the NFT detection, the laggy behavior goes away. This further points to the issue that the NFT detection is currently expensive/operationally heavy and is affecting other UIs

After further testing its true that commenting out detectNFTs improves performance but the experience is still laggy when having a large list of nfts. Look at the following video where I commented out nft detection

https://github.com/user-attachments/assets/85cba3dd-3af1-4f59-af71-14ab208a45e0

app/components/Views/Wallet/index.tsx
```diff
@@ -1194,6 +1194,15 @@ const Wallet = ({
...
+      } else if (tabLabel === strings('wallet.perps')) {
+        // Track Perps tab selection - no NFT detection needed
+        // Note: Add Perps analytics tracking here if needed
+      } else if (tabLabel === strings('wallet.predict')) {
+        // Track Predictions tab selection - no NFT detection needed
+        // Note: Add Predictions analytics tracking here if needed
+      } else {
+        // Handle any other tabs that might be added in the future
+        // Note: Only add NFT detection if specifically needed for that tab
       }
```
georgewrmarshall, line 1197: non-blocking: I think we could remove or comment out these empty `else if` statements. Do we know if there are plans to implement tracking?

app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx
```diff
@@ -314,31 +397,32 @@ const TabsBar: React.FC<TabsBarProps> = ({
...
+              shouldShow && (
+                <Animated.View
+                  style={tw.style('absolute bottom-0 h-0.5 bg-icon-default', {
+                    width: underlineWidthAnimated,
+                    transform: [{ translateX: underlineAnimated }],
+                  })}
+                />
+              )
+            );
+          })()}
```
georgewrmarshall, line 425: non-blocking: The animation for the tab underline seems to be a little laggy on iOS simulator and real android device on the first load of each tab. After the tab is loaded it seems to be a lot snappier

https://github.com/user-attachments/assets/157cb176-f69a-499f-ad5b-17e8120838bb

https://github.com/user-attachments/assets/e914fbf8-6de1-4d8e-b3e6-c7389d7b7519

georgewrmarshall (commented):
Great work on this! Approving to unblock but I am experiencing some issues:

1. Underline seems somewhat laggy on first tab load (low)
2. The tab layout scrolling is the same size regardless of the content (medium)

https://github.com/user-attachments/assets/e1cd9555-18fa-4443-a4b6-5e6e120f0f89

georgewrmarshall (dismissed):

app/components/Views/Wallet/index.tsx
brianacnguyen: Removed

georgewrmarshall (approved):

app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx
brianacnguyen: This will need to be dealt with with the addition of the skeleton loader, which can be added later

cortisiko (approved):

amitabh94 (approved):

PR #20211, opened 2025-09-23
fix: cp-7.56.0 speed up source token picker network change, skip token fetching

infiniteflower:
## **Description**

This PR does the following:
1. Fixes an issue where Swaps source token picker would make an API request to fetch tokens after selecting a token when it was not needed
2. Use `TokenListController`'s token cache as a first hit before reaching out to Bridge API
4. Memoizes all `fetchBridgeTokens` calls
5. Uses `FlatList` for both iOS and Android now. `FlashList` seemed to have some initial rendering issues when memoizing `fetchBridgeTokens`.
6. Fix an issue where if you were partially scrolled down on the destination token picker and then changed dest networks, the scroll position would remain, which caused rendering issues.

## **Changelog**

CHANGELOG entry: Fixed some performance issues with Swaps asset picker

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/20177
https://consensyssoftware.atlassian.net/browse/SWAPS-3067

## **Manual testing steps**

```gherkin
Feature: my feature name

  Scenario: user [verb for user action]
    Given [describe expected initial app state]

    When user [verb for user action]
    Then [describe expected outcome]
```

## **Screenshots/Recordings**

Before

https://github.com/user-attachments/assets/e572ed95-3f10-4dcf-8b25-27864c178299

After

https://github.com/user-attachments/assets/c672a932-0bca-413b-80b7-d4a5c8451eba

No /getTokens call

https://github.com/user-attachments/assets/d454b6a3-dbbd-4f90-8097-fd192200c0b5

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
[...]

PR #20241, opened 2025-09-23
feat: cache destination assets

bfullam:
## **Description**
The BridgeDestTokenSelector component experienced significant performance issues when loading token lists, particularly for networks with large token sets like Ethereum (11,000+ tokens). Users faced 2-3 second loading times every time they opened the token selector, regardless of whether they had previously loaded tokens for that network. This poor performance was caused by:

1. **Synchronous API dependency**: Every component mount triggered a fresh API call to the Bridge service to fetch available tokens
2. **No persistence**: Token data was discarded when the component unmounted, requiring complete re-fetching on subsequent opens
3. **No offline resilience**: Component failed completely when network requests failed

## **Solution Summary**

A comprehensive token caching system using MMKV persistent storage to dramatically improve loading performance while maintaining data freshness. The solution consists of:

**Core Components:**

- **BridgeTokensCacheManager**: Persistent cache layer with 30-minute TTL and chain-specific validation
- **Enhanced useTopTokens hook**: Progressive loading strategy that shows cached data immediately while refreshing in the background
- **Immediate state clearing**: Prevents displaying wrong network tokens during chain switches

**Key Features:**

- **Progressive Loading**: Cache-first strategy provides instant token display (~50ms) for subsequent loads
- **Background Refresh**: Fresh data fetched asynchronously to keep cache current
- **Chain Validation**: Multi-layer validation ensures cached tokens belong to the correct network
- **Graceful Degradation**: Automatic fallback to fresh API calls when cache is invalid or missing
- **Offline Resilience**: Users can access previously loaded tokens even without network connectivity

Further documentation [here](https://www.notion.so/metamask-consensys/Bridge-Destination-Asset-Selector-Cache-277f86d67d6880509ce6e83f43199825?source=copy_link)

## **Changelog**

CHANGELOG entry: Added caching system for swap destination assets

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/20177

https://consensyssoftware.atlassian.net/browse/SWAPS-3067

## **Manual testing steps**

```gherkin
Feature: my feature name

```


app/components/UI/Bridge/utils/BridgeTokensCacheManager.ts
```diff
@@ -0,0 +1,172 @@
...
+  ): Promise<void> {
+    try {
+      const cacheKey = this.getCacheKey(chainId);
+      const cacheData: CachedBridgeTokens = {
+        tokens,
+        timestamp: Date.now(),
+        chainId,
+      };
+
+      await StorageWrapper.setItem(cacheKey, JSON.stringify(cacheData));
```
Nodonisko, line 113: I would be very careful with JSON, if data are huge, JSON stringify/parse operations could be very slow. I am not sure if it's worth using persistent storage if you are caching for just 30 minutes. Consider in-memory cache only without JSON step.

bfullam: Is there another solution here you would suggest as a better alternative?

Cache length can be arbitrarily extended, since it refreshes in the background anyway.

infiniteflower (commented):
Would something simpler like `lodash.memoize` work? It means less code for us to maintain. 

I was using that in this PR https://github.com/MetaMask/metamask-mobile/pull/20211

app/components/UI/Bridge/utils/BridgeTokensCacheManager.ts
Nodonisko: I would just use Map/WeakMap or something like that to store it in memory without need for JSON serialization.

PR #20243, opened 2025-09-23
chore: rwds-468 fix stale balance

VGR-GIT:
## **Description**

Balance via season status can get out of sync with point events seen in activity tab. So we have a mechanism now in the rewards controller that invalidates the necessary concepts (like season status) by emitting a balanceUpdated event. It does this by checking if the earliest points event it finds is after either the `season status cache.lastFetched `or `season status cache.balance.updatedAt `+ a small threshold (as point event & balance update happen almost at the same time but balance is often a bit sooner)

## **Changelog**

CHANGELOG entry: null

### DIFF
--- a/app/core/Engine/controllers/rewards-controller/RewardsController.ts
+++ b/app/core/Engine/controllers/rewards-controller/RewardsController.ts
@@ -824,6 +824,80 @@ export class RewardsController extends BaseController<
   }
 
   /**
+   * Triggers a balance update if newer points events are detected
+   * @param pointsEvents - The points events response from the API
+   * @param params - The original request parameters containing seasonId and subscriptionId
+   */
+  private triggerBalanceUpdateIfNeeded(
+    pointsEvents: PaginatedPointsEventsDto,
+    params: GetPointsEventsDto,
+  ): void {
+    // Only proceed if there are results to process
+    if (pointsEvents.results.length === 0) {
+      return;
+    }
+
+    // Find the latest updatedAt from the points events response
+    const latestEvent = maxBy(pointsEvents.results, (event) =>
+      (event.updatedAt instanceof Date
+        ? event.updatedAt
+        : new Date(event.updatedAt)
+      ).getTime(),
+    );
+
+    if (!latestEvent) {
+      return;
+    }
+
+    const latestUpdatedAt =
+      latestEvent.updatedAt instanceof Date
+        ? latestEvent.updatedAt.getTime()
+        : new Date(latestEvent.updatedAt).getTime();
+
+    // Check if we have an active season status cache for the requested seasonId and subscription id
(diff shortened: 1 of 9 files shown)


app/core/Engine/controllers/rewards-controller/RewardsController.ts
```diff
@@ -840,6 +848,80 @@ export class RewardsController extends BaseController<
...
+          );
+
+          if (cachedSeasonStatus) {
+            // Compare cache timestamps with latest updatedAt
+            // Add 300ms delay to the balance updated timestamp as it's always going to be earlier than the event it was updated for.
+            const cacheBalanceUpdatedAt =
+              (cachedSeasonStatus.balance.updatedAt || 0) + 300;
+
+            const cacheSeasonStatusTimestamp =
+              cachedSeasonStatus.lastFetched || 0;
```
jonson: Since `cachedSeasonStatus.lastFetched` is using the device clock as the the timestamp source (via `Date.now()`, this logic is fragile.   We should avoid comparing timestamps from the device clock with timestamps generated by the backend, as we cannot assume the device's clock is perfectly in-sync with the backend.  If it's off by even 1s, it would invalidate the check we're doing with this value below.

app/core/Engine/controllers/rewards-controller/RewardsController.ts
```diff
@@ -840,6 +848,80 @@ export class RewardsController extends BaseController<
         'RewardsDataService:getPointsEvents',
         params,
       );
+
+      // Check if we should emit balance updated event
+      if (pointsEvents.results.length > 0) {
```
jonson: The `getPointsEvents` function is now 100 lines long and it's hard to follow what all this new logic is doing -- the majority of the logic in the function is now around invalidating the season status.  Can we extract it into a separate function (or even better, multiple smaller, well named functions) that better describe what it's doing?   `triggerBalanceUpdateIfNeeded` or `invalidateSeasonStatusIfNeeded` or something more obvious?

jonson (dismissed):

app/core/Engine/controllers/rewards-controller/RewardsController.ts
VGR-GIT: Yeah that makes sense. I think we can just keep this scoped to the balance update date. There's still the issue of the short gap between this date and its associated points event, but we can add some MS padding to the points event date to mitigate for that.

VGR-GIT: I'm going to go ahead with this approach (without checking vs device time). If we need to do any changes we can do so in a follow-up approach. I think the best approach would still be something like returning the last points event id as part of the season status, and checking that vs the earliest points event in the list

jbblanc (approved):

PR #20317, opened 2025-09-24
fix: cp-7.56.0 allow tpsl form to enter profit in sl and loss in tp for existing position

nickewansmith:
## **Description**
- When a user has an open position that has already generated a gain, and set a SL that is between current price and entry price, the “% loss” component display “0”
- When a user has an open position that has already generated a loss, and set a TP that is between current price and entry price, the “% loss” component display “0”

- % gain and % loss display a value whether it’s >0 or <0
- % profit/loss is calculated based on entry price

- If the TP generates a gain
  - Display a “+” in the % profit box, along with the %

- If the TP generates a loss (only feasible when modifying the TP/SL on an open position)
  - Display a “-” in the % profit box, along with the %

- If the SL generates a loss
  - Display a “-” in the % profit box, along with the %

- If the SL generates a gain (only feasible when modifying the TP/SL on an open position)
  - Display a “+” in the % profit box, along with the %

- When user inputs the %profit/loss in the component, via keyboard
  - By default, we use + for a TP and - for a SL

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1517

## **Manual testing steps**

```gherkin
Feature: fix tp loss and sl profit

  Scenario: user has an existing order
    Given user edits tpsl via position card

    When user enters a price that results in a positive stop loss, or a negative tp
    Then it is allowed and the user can save the value if valid
    Then there should be appropriate signs according to figma design and ticket

    Scenario: user has no existing order or is creating a new market or limit order
    Given user edits tpsl via order view

    When user enters a price that results in a positive stop loss or a negative tp
    Then functionality remains as before, the percentage will be zeroed out.
    Then there should be appropriate signs according to figma design and ticket
```

### **After**

https://github.com/user-attachments/assets/f1445fbd-6ce9-4075-b7f4-1d4a20a3dad6

Most recent video and most updated:

https://github.com/user-attachments/assets/4e04bfb0-566b-4b0a-a044-242958cd9674


app/components/UI/Perps/hooks/usePerpsTPSLForm.ts
```diff
@@ -438,41 +481,71 @@ export function usePerpsTPSLForm(
...
+          // Mixed signs - keep only the first sign
+          const firstSignMatch = finalValue.match(/[+-]/);
+          if (firstSignMatch) {
+            const sign = firstSignMatch[0];
+            const restOfString = finalValue.substring(1);
+            const numberPart = restOfString.replace(/[+-]/g, '');
+            finalValue = numberPart ? sign + numberPart : sign;
+          }
+        }
+      }
```
gambinish: It looks like this is some duplicated code from `handleTakeProfitPercentageChange` can this be consolidated?

gambinish (approved):

PR #20386, opened 2025-09-25
improve snap name resolution by using a callback and cleanup

GuillaumeRx:
This refactors the snap name lookup hook and does some cleanup around the logic that uses it.

### DIFF
--- a/app/components/Snaps/hooks/useSnapNameResolution.ts
+++ b/app/components/Snaps/hooks/useSnapNameResolution.ts
@@ -1,10 +1,10 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useCallback } from 'react';
 import { useSelector } from 'react-redux';
 import {
   getChainIdsCaveat,
   getLookupMatchersCaveat,
 } from '@metamask/snaps-rpc-methods';
-import { AddressResolution, DomainLookupResult } from '@metamask/snaps-sdk';
+import { DomainLookupResult } from '@metamask/snaps-sdk';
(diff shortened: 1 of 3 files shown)


app/components/Views/confirmations/hooks/send/useToAddressValidation.ts
```diff
@@ -72,7 +73,7 @@ export const useToAddressValidation = () => {
     warning?: string;
     loading?: boolean;
     resolvedAddress?: string;
-  }>(async () => validateToAddress(), [validateToAddress]);
+  }>(async () => validateToAddress(), [validateToAddress, chainId, to]);
```
GuillaumeRx, line 76: This is vital to ensure that any request made prior to a typing change gets canceled.

app/components/Snaps/hooks/useSnapNameResolution.ts
GuillaumeRx, line 1: Refactored to return a function that can be called when needed. This reduces the calls to the snap by only calling them if we are sure the value is a domain name.

app/components/Views/confirmations/hooks/send/useToAddressValidation.ts
jpuri: `chainId, to` are already dependencies for `validateToAddress` callback.

GuillaumeRx: But there is no cancel logic in the `validateToAddress` callback, without this I could see previous requests overwriting the current request if the snap was slow.

PR #20456, opened 2025-09-26
fix(SnapKeyring): do not use approval flow if confirmation is skipped

PatrykLucka:
## **Description**
This PR fixes a bug causing infinite loader when there was more than 2 solana accounts to be created (during alignement process).

Specifically, it prevents from using approval flow when there is `skipConfirmation` passed as an option to the `addAccount` function in `SnapKeyring`. Thanks to the approval flow won't be used, so the `FlowLoaderModal` won't be unnecessarily shown.

## **Changelog**

CHANGELOG entry: Fixed a bug that was causing infinite loader when updating version of the app

## **Related issues**

Fixes:
Jira ticket: https://consensyssoftware.atlassian.net/browse/MUL-1038

## **Manual testing steps**

```gherkin
Feature: Update to latest version

  Scenario: user updates application
    Given current prod: 7.54.2(2390) version is installed

    When user imports wallet with a lot of accounts
    Then kills the app
    Then updates to latest version 7.57.0 (2517)
    Then launches the app
    Then everything should work without blocking the app
```

### DIFF
--- a/app/core/SnapKeyring/SnapKeyring.ts
+++ b/app/core/SnapKeyring/SnapKeyring.ts
@@ -118,15 +118,15 @@ class SnapKeyringImpl implements SnapKeyringCallbacks {
     handleUserInput,
     accountNameSuggestion,
     skipAccountNameSuggestionDialog,
-    skipConfirmation,
+    skipApprovalFlow,
   }: {
     snapId: SnapId;
     accountNameSuggestion: string;
     handleUserInput: (accepted: boolean) => Promise<void>;
     skipAccountNameSuggestionDialog: boolean;
-    skipConfirmation: boolean;
+    skipApprovalFlow: boolean;
   }): Promise<{ accountName?: string }> {
-    if (skipConfirmation) {
+    if (skipApprovalFlow) {
       const { accountName } = await this.getAccountNameFromSuggestion(
         accountNameSuggestion,
       );
@@ -153,14 +153,14 @@ class SnapKeyringImpl implements SnapKeyringCallbacks {
     address: _address,
     snapId,
     skipSetSelectedAccountStep,
-    skipConfirmation,
+    skipApprovalFlow,
     onceSaved,
     accountName,
   }: {
     address: string;
     snapId: SnapId;
     skipSetSelectedAccountStep: boolean;
-    skipConfirmation: boolean;
+    skipApprovalFlow: boolean;
     onceSaved: Promise<string>;
     accountName?: string;
   }) {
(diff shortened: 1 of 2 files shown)


app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -222,6 +238,7 @@ class SnapKeyringImpl implements SnapKeyringCallbacks {
     // Only pre-installed Snaps can skip the account name suggestion dialog.
     const skipAccountNameSuggestionDialog =
       isSnapPreinstalled(snapId) && !displayAccountNameSuggestion;
+    const skipConfirmation = isSnapPreinstalled(snapId) && !displayConfirmation;
```
ccharly: I wonder if we should also use `&& skipAccountNameSuggestionDialog` here too.

Technically speaking, we cannot "skip the confirmations" if we need the get the name from the dialog

Though... we literally don't use those dialogs anymore, so maybe that's fine?

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -118,12 +118,21 @@ class SnapKeyringImpl implements SnapKeyringCallbacks {
...
     skipAccountNameSuggestionDialog: boolean;
+    skipApprovalFlow: boolean;
   }): Promise<{ accountName?: string }> {
+    if (skipApprovalFlow) {
+      const { accountName } = await this.getAccountNameFromSuggestion(
+        accountNameSuggestion,
+      );
+      await handleUserInput(true);
+      return { accountName };
+    }
```
fabiobozzo, line 135:
```suggestion
if (skipApprovalFlow) {
  try {
    const { accountName } = await this.getAccountNameFromSuggestion(
      accountNameSuggestion,
    );
    await handleUserInput(true);
    return { accountName };
  } catch (error) {
    await handleUserInput(false);
    throw new Error('Account creation failed');
  }
}
```

no error handling if getAccountNameFromSuggestion fails?

app/core/SnapKeyring/SnapKeyring.ts
```diff
@@ -203,7 +214,11 @@ class SnapKeyringImpl implements SnapKeyringCallbacks {
         // This part of the flow is not awaited, so we just log the error for now:
         Logger.error(error, 'Error occurred while creating snap account');
       }
-    });
+    };
+    if (skipApprovalFlow) {
+      return finalizeFn();
```
fabiobozzo, line 219:
```suggestion
      return await finalizeFn();
```
am I wrong?

fabiobozzo (approved):
@montelaidev I left a couple of comments just for my understanding, feel free to ignore them if they don't make sense 😊

PR #20909, opened 2025-10-07
feat: speed up swap quote availability with SSE

micaelae:
## **Description**

This PR integrates the bridge-api's getQuoteStream endpoint for fetching quotes, which is used instead of getQuotes when the LD `sse` flag is enabled for the build's version

To test this locally before the flags are enabled, set `clientVersion='8.0.0'` in the bridge controller init

This speeds up the time it takes to see the first quote. Some examples of slow quotes are available [here for testing purposes](https://mixpanel.com/project/2212883/view/2760759/app/insights#XfmYSQuCcoUb)

## **Changelog**

CHANGELOG entry: feat: use SSE to stream swap quotes

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/SWAPS-3243

## **Manual testing steps**

```gherkin
Feature: swap e2e flow should not have any user-facing changes

```

### **Before**

N/A

### **After**

N/A

---

> [!NOTE]
> Adds a bridge fetch wrapper to use expo/fetch for SSE quote streams and updates BridgeView to hide quotes while keypad is shown, adjust loading/continue states, and expose loading status from the quote hook.
> 
> - **Engine/Bridge Controller**:
>   - Add `handleBridgeFetch` to route `*Stream` endpoints via `expo/fetch`, others via `handleFetch`.
>   - Wire `BridgeController` `fetchFn` to `handleBridgeFetch`; add unit tests.
> - **Bridge UI**:
>   - Update keypad/quote visibility: show keypad when input is focused/inputs invalid/no active quote; hide quote details when keypad is visible.
>   - Blur input once quotes load or when `activeQuote.quote.requestId` exists.
>   - Bottom content: show "Fetching quote" only when `isLoading && !activeQuote`; hide Continue button when keypad is shown.
>   - `TokenInputArea` destination loading only when no `destTokenAmount` and `isLoading`.
> - **Quote Data Hook (`useBridgeQuoteData`)**:
>   - Return `quotesLoadingStatus`; define `isLoading` from it.
> - **Tests**:
>   - Adjust BridgeView tests for new loading/activeQuote behavior and keypad/continue visibility; add case for quote available while others load; minor fixes (clear mocks, timestamps).
>   - Add tests for `handleBridgeFetch` SSE routing.
> - **Docs/Setup**:
>   - Mock `expo/fetch` in test setup; minor README formatting tweaks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ce09bb4a1743a13ca59842707ded01e5aeb09306. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


yarn.lock
```diff
@@ -3649,6 +3649,23 @@ __metadata:
...
+  version: 1.0.1
+  resolution: "@expo/browser-polyfill@npm:1.0.1"
+  dependencies:
+    expo-2d-context: ^0.0.3
+    fbemitter: ^2.1.1
+    text-encoding: ^0.7.0
+    uuid: ^8.3.2
+    xmldom-qsa: ^1.0.3
+  peerDependencies:
+    expo-file-system: ^13.2.0
```
Gudahtt: This package requires older versions of these three packages, which we don't currently use. This may not function correctly with the newer versions.

Additionally, it's bringing in quite an old copy of `fbemitter`, which is indirectly pulling in older versions of a bunch of utility libraries that we already have copies of in the dependency tree.

Is there an equivalent alternative we can use instead? This one is clearly not well maintained.

micaelae: I added `@expo/browser-polyfill` to polyfill the `document` and `ReadableStream` objects specifically. I tried a few other libs before but haven't found one that works with `fetch-event-source`. Will try out some newer polyfills and get back to you

micaelae: I tried out some other polyfills for ReadableStream (`web-streams-polyfill` seems to be the top recommendation) and other ways to define `document`. However none of them successfully fixed the missing references

I came up with a hacky workaround though - I patched the SSE fetching dependency such that the usages of `document` are removed. This is ok since those lines are for stopping requests when the document is hidden (and restarting them when visible), which is not a usecase on mobile. And I confirmed the expected fetching behavior is preserved

^ This brought the bundle size back down to <45MB because it enabled me to remove the polyfill

I'm circling back with the swaps team to see what they think of this solution though. Also discussing potentially replacing the lib with our own utility so we don't need to patch. Moving this PR back to Draft state in the meantime

app/core/Engine/controllers/bridge-controller/bridge-controller-init.ts
```diff
@@ -20,6 +21,17 @@ import packageJSON from '../../../../../package.json';
 
 const { version: clientVersion } = packageJSON;
 
+export const handleBridgeFetch = (
+  url: RequestInfo | URL,
+  options: RequestInit = {},
+) => {
+  if (url.toString().includes('Stream')) {
+    // @ts-expect-error types are different
```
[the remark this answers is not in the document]
micaelae: it's always going to be a string as defined in the controller

SteP-n-s (approved):
Can we also reduce the debounce threshold to match the one used by the extension (300ms)? Right now, it’s set to 700ms, which is too high (see DEBOUNCE_WAIT in useBridgeQuoteRequest).

PR #21225, opened 2025-10-15
feat: New Persistence System Improvements

MarioAslau:
## **Description**

This PR addresses comments post-merge from https://github.com/MetaMask/metamask-mobile/pull/17685

Bitrise Build: https://app.bitrise.io/app/be69d4368ee7e86d/pipelines/fb54115a-4f82-4b48-9758-f561f98983ec

Fixes included in this PR:
- I've made it to skip the setup of the persist function for ones with zero persisted state and also removed the CronjobController check
- Now crashing app if initialisation fails
- Now crashing migration instead of saving state
- Using a messenger subscription selector to ensure unnecessary writes, only if a state change actually happens

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

I've tested the upgrade path for Persistence Improvements from:
* 7.54.0 (2344) → 7.58.9(2726)
* 7.51.0 (2157) → 7.58.9(2726)

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> Selective controller persistence with Redux update subscriptions and crash-on-failure migration flow, plus comprehensive tests.
> 
> - **EngineService**:
>   - Set up immediate Redux updates for controller state changes via `controllerMessenger.subscribe`, excluding `CronjobController`.
>   - Add persistence gating using `hasPersistedState` to skip controllers without persistent fields; log skips.
>   - Persistence handler filters state with `getPersistentState` and logs (without crashing) on per-update persistence errors.
>   - Strengthen setup error handling: throw critical error if persistence setup fails.
>   - Additional logs for missing `KeyringController.metadata.vault` during INIT/UPDATE.
> - **Migrations**:
>   - `inflateFromControllers`: on load failure, capture and throw critical error (crash) to prevent running with incomplete data.
>   - `deflateToControllersAndStrip`: crash immediately on any controller save failure; strip `engine` only after all saves succeed; improved error reporting.
>   - Minor fix to `hasControllers` check formatting.
> - **Utils**:
>   - New `utils/persistence-utils.ts` with `hasPersistedState` helper and tests.
> - **Tests**:
>   - Expanded `EngineService` tests for batching, subscriptions, vault checks, persistence gating, error logging, and Cronjob exclusion.
>   - New/expanded migration tests covering critical error paths and mixed-version flows.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d4a4f1355b56fbe795d63cc3ee60cc1a00bcb254. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/core/EngineService/EngineService.ts
```diff
@@ -49,6 +49,78 @@ export class EngineService {
     }),
   );
 
+  /**
+   * Checks if a controller has any persistent state based on its metadata.
+   * 
+   * @param metadata - The controller's state metadata
+   * @returns true if the controller has properties marked for persistence, false otherwise
+   */
+  private hasPersistedState = (metadata: any): boolean => {
```
Gudahtt: Why `any` here? This is against our TypeScript guidelines, and I thought we had a lint rule to disallow it as well

Gudahtt: Likewise for the `any` on line 63

Pum482930 (approved):

app/core/EngineService/EngineService.ts
```diff
@@ -122,89 +195,70 @@ export class EngineService {
...
                   `Failed to process ${controllerName} state change`,
                 );
+                // Individual persistence failure could be Sev1, user data at risk
+                // Better to crash and get immediate attention than silently fail
+                throw new Error(
+                  `Critical: Failed to persist ${controllerName} state. User data at risk. ${(error as Error).message}`,
+                );
               }
             },
+            // Selector: only trigger handler when persistent state changes, preventin unnecessary writes
```
Gudahtt: Sorry this was a bad suggestion, this doesn't work properly because `getPersistentState` returns a new object each time (so it always triggers a write, even if nothing really changed).

We had some ideas on how to improve this selector feature to handle cases like this (e.g. to compare top-level properties of the object rather than the root object itself), but that's not something supported today. Unfortunately we don't currently have a way to skip calling the handler when only non-persisted properties change.

MarioAslau: Changes reverted

app/store/migrations/index.ts
```diff
@@ -406,4 +414,4 @@ export const migrations = asyncifyMigrations(migrationList, (state) => {
 }) as unknown as MigrationManifest;
 
 // The latest (i.e. highest) version number.
-export const version = Object.keys(migrations).length - 1;
+export const version = Object.keys(migrations).length - 1;
```
Gudahtt: Nit: Trailing newline was removed here for some reason

app/store/migrations/index.test.ts
```diff
@@ -196,3 +212,238 @@ describe('migrations', () => {
     expect((migratedState as Record<string, unknown>).test).toEqual('sync');
   });
 });
+
+describe('Critical Error Handling', () => {
+  const mockedControllerStorage = ControllerStorage as jest.Mocked<typeof ControllerStorage>;
+  const mockedCaptureException = captureException as jest.MockedFunction<typeof captureException>;
```
Gudahtt: Jest has a utility function for this, we don't need to cast here:
```suggestion
  const mockedControllerStorage = jest.mocked(ControllerStorage);
  const mockedCaptureException = jest.mocked(captureException);
```

app/core/EngineService/EngineService.test.ts
```diff
@@ -360,6 +385,291 @@ describe('EngineService', () => {
     });
   });
 
+  describe('hasPersistedState', () => {
+    // Type for accessing private methods
+    interface EngineServiceWithPrivateMethods {
+      hasPersistedState: (metadata: Record<string, unknown> | undefined) => boolean;
```
Gudahtt: This practice of testing with private methods is strongly discouraged by our TypeScript and testing guidelines. I don't think we should bother change it here for now, as I see it's already a well established pattern in this file. But I just want to flag this as something we should avoid in the future, and move away from here eventually, ideally.

Gudahtt: Usually if we're testing a private function, it's a sign that we're testing the implementation rather than the API. Our test will then be fragile and likely to break in the future, adding friction rather than preventing bugs. It also might not model how the internal function is called correctly, leading to bugs/missed test cases. That's what our guidelines mainly focus on for this guideline.

In this case though, IMO testing `hasPersistedState` directly makes sense, it's a pure function, a utility function. In this case it would be better moved to a separate module I think. Our test guidelines assume each module under test is "one thing", you know. Something with a single responsibility that is expressed (and hopefully well documented!) via the public API. If you get the urge to test internals as though it's an independent component, it should probably be an independent component.

Alternatively (or additionally? I wouldn't want to delete these tests, these are good tests) we could test `hasPersistedState` indirectly by seeing if controllers without persisted state trigger writes.

MarioAslau: I fixed it! Thanks for the explanation @Gudahtt !

app/core/EngineService/EngineService.ts
```diff
@@ -49,6 +51,57 @@ export class EngineService {
...
+    );
+
+    // Set up immediate Redux updates for all controller state changes
+    // This ensures Redux is updated right away when controllers change
+    const update_bg_state_cb = (controllerName: string) => {
+      if (!engine.context.KeyringController.metadata?.vault) {
+        Logger.log('keyringController vault missing for UPDATE_BG_STATE_KEY');
+      }
+      this.updateBatcher.add(controllerName);
+    };
```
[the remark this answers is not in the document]
Gudahtt: It's true that this condition isn't really in the ideal place, but we should address this separately. It's pre-existing and not related to the changes in this PR.

app/core/EngineService/EngineService.ts
```diff
@@ -49,6 +50,57 @@ export class EngineService {
...
+      () => !this.engineInitialized,
+    );
+
+    // Set up immediate Redux updates for all controller state changes
+    // This ensures Redux is updated right away when controllers change
+    const update_bg_state_cb = (controllerName: string) => {
+      if (!engine.context.KeyringController.metadata?.vault) {
+        Logger.log('keyringController vault missing for UPDATE_BG_STATE_KEY');
+      }
+      this.updateBatcher.add(controllerName);
```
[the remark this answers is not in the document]
Gudahtt: Valid but pre-existing bug

app/core/EngineService/EngineService.ts
```diff
@@ -122,52 +175,26 @@ export class EngineService {
...
 
-          // Skip CronjobController state change events
-          // as they are handled separately in the CronjobControllerStorageManager.
-          // This prevents duplicate updates to the Redux store.
-          if (eventName === 'CronjobController:stateChange') {
+          // Check if controller has any persistent state before setting up persistence
+          // @ts-expect-error Engine context has stateless controllers, so metadata may not be available
+          const controllerMetadata = UntypedEngine.context[controllerName]?.metadata;
+          if (!hasPersistedState(controllerMetadata)) {
+            Logger.log(`Skipping persistence setup for ${controllerName}, no persistent state`);
```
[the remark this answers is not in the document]
Gudahtt: Luckily it does not

app/core/EngineService/EngineService.ts
```diff
@@ -184,27 +210,35 @@ export class EngineService {
...
-                this.updateBatcher.add(controllerName);
-
                 await persistController(filteredState, controllerName);
               } catch (error) {
                 Logger.error(
                   error as Error,
                   `Failed to process ${controllerName} state change`,
                 );
+                // Individual persistence failure could be Sev1, user data at risk
+                // Better to crash and get immediate attention than silently fail
```
Gudahtt: This could happen as a result of any state change, right? And we reach this line if the write operation fails?

This would be expected to happen in low disk space scenarios. I don't think crashing is the right thing to do here, because this error is expected and not something we can prevent. i.e. it doesn't necessarily indicate a bug, getting our attention won't help.

app/core/EngineService/EngineService.ts
```diff
@@ -260,4 +294,4 @@ export class EngineService {
 /**
  * EngineService class used for initializing and subscribing to the engine controllers
  */
-export default new EngineService();
+export default new EngineService();
```
Gudahtt: Nit: no trailing newline (I'm surprised the linter doesn't complain about this 🤔 )

app/core/EngineService/EngineService.test.ts
```diff
@@ -360,6 +385,228 @@ describe('EngineService', () => {
     });
   });
 
+
+  describe('initializeControllers edge cases', () => {
+    // Type for accessing private methods
+    interface EngineServiceWithInitializeControllers {
+      initializeControllers: (engine: { context: null; controllerMessenger: { subscribeOnceIf: jest.MockedFunction<(...args: unknown[]) => void> } }) => void;
```
Gudahtt: Looks like Prettier hasn't been run on this file

app/core/EngineService/utils/persistence-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @example
+ * ```typescript
+ * const metadata = {
+ *   field1: { persist: true, anonymous: false },
+ *   field2: { persist: false, anonymous: true },
+ * };
+ * hasPersistedState(metadata); // returns true
+ * ```
+ */
+export const hasPersistedState = (metadata: StateMetadata<StateConstraint> | undefined): boolean => {
```
tommasini, line 24: Interesting this was used for efficiency, nice! Should we propose migrate this to base controller instead of mobile codebase?

tommasini (dismissed):

tommasini (dismissed):

joaoloureirop (dismissed):
There are comments that we could remove as they do not add value.
But that alone is not worth of another CI run. 

Let's address them in the next revision.

joaoloureirop (approved):

PR #21347, opened 2025-10-17
fix: cp-7.58: add subscriptionid user metrics trait

### LINKED ISSUE #21346: [Bug]: Track subscription ID as an identity
View issue in GitHub: https://github.com/MetaMask/metamask-mobile/issues/21346

**Describe the bug**

Without putting subscription ID in the metametrics identity, we have no way to correlate data from metametrics with data in our Rewards backend. We need this field in identity and for the RewardsController to populate this after a user opts in. We should also backfill this information since we are making this change after v1 launch. A user could opt out of rewards and then opt in later, so we should upsert on opt-in too.

**Steps to reproduce**

1. Opt into rewards program
2. We don't associate the subscription_id with the user via traits.

**Error messages or log output**

```shell

```

**Detection stage**

In production (default)

**Version**

7.58

**Build type**

None

**Device**

Moto Edge 60 Pro

**Operating system**

Android, iOS

VGR-GIT:
## **Description**

Whout putting subscription ID in the metametrics identity, we have no way to correlate data from metametrics with data in our Rewards backend. We need this field in identity and for the RewardsController to populate this after a user opts in. We should also backfill this information since we are making this change after v1 launch. A user could opt out of rewards and then opt in later, so we should upsert on opt-in too. 

This change adds the `rewards_subscription_id` user trait:

- when opting in
- when visiting the rewards dashboard
- when opting out (clearing it)

## **Changelog**

CHANGELOG entry: track rewards_subscription_id as user metric

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/21346

---

> [!NOTE]
> Adds/updates metrics to set rewards_subscription_id on opt-in and dashboard view, and clear it on opt-out, with supporting types and tests.
> 
> - **Rewards (metrics/analytics)**:
>   - **Dashboard (`app/components/UI/Rewards/Views/RewardsDashboard.tsx`)**:
>     - Fetch `RewardsController:getFirstSubscriptionId` on mount and add user trait via `addTraitsToUser`.
>     - Keep existing view/tab tracking; update hook deps.
>   - **Opt-in (`useOptIn.ts`)**:
>     - After successful opt-in, add user traits including `has_rewards_opted_in`, `rewards_subscription_id`, and referral traits when present.
>   - **Opt-out (`useOptout.ts`)**:
>     - On successful opt-out, clear `rewards_subscription_id` trait and set `has_rewards_opted_in` to `OFF`.
>   - **Types (`UserProfileAnalyticsMetaData.types.ts`)**:
>     - Add `UserProfileProperty.REWARDS_SUBSCRIPTION_ID`.
>   - **Tests**:
>     - Update/add tests for dashboard analytics, opt-in/opt-out trait behavior; mock `Engine.controllerMessenger.call` and `useMetrics`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a23788bc279dca6141eb658a62ddeb4abbb38444. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Brunonascdev (dismissed):

jbblanc (dismissed):

Vahid-flipside:
@Brunonascdev @Montoya  I am not sure what the requirement is after opt-out when you say "clearing it".
I believe we will just stop firing data for that mmid which opts out but we don't clear anything unless the user decides to exercise its right to be forgotten [which afaik, its the dark "Delete Metametrics Data" in settings], afterwhich we'd need to delete their data in Bigquery and Mixpanel.
Am I getting this right @weslleymoura ?

Montoya:
@Vahid-flipside Rwds has its own "right to be forgotten" feature which is opt out. We remove as much information as we possibly can for the user when they opt out, including the subscription. So we would like to remove that subscription ID from the identity as well, with the intention that the user's "right to be forgotten" includes removing the mapping between metrics and Rwds backend. 

Let us know if we should make the string empty: "" or we can make it `undefined`

Vahid-flipside:
> know if we should make the string empty: "" or we can make it `undefined`

Thanks. undefined would be good but since the property is already defined as "string" in segment schema, I am not sure if that's possible. maybe empty string is easier.

Brunonascdev (approved):

PR #21408, opened 2025-10-20
fix: Prevent relative paths in RNTar native modules

### LINKED ISSUE #8096: some suggestions for the RNTar module.
### What is this about?

There is a potential ZIP slip vulnerability in the code segment in RNTar.java.   This vulnerability arises because the code does not properly filter the entry name of the compressed package. Although this module is not invoked from the UI and runs locally on the user's machine,an attacker can deceive a user into importing a malicious project file.   Once this file is imported, it could lead to security risks.   Therefore, it is recommended to validate the entry name of the compressed package as shown below.

### Scenario

_No response_

### Design

_No response_

### Technical Details

In the extractTgzFile method of the RNTar.java file located at android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java, you can add a validation for entry.getName() before the code File outputFile = new File(outputPath, entry.getName())
like this:
if (n.contains("/..") || n.contains("../")) {
          throw new RuntimeException("Entry with an illegal name");
        }

### Threat Modeling Framework

_No response_

### Acceptance Criteria

_No response_

- [ ] Engineering (needed in most cases)
- [ ] Design
- [ ] Product
- [X] Security
- [ ] Legal
- [ ] Marketing
- [ ] Other (please specify)

### References

_No response_

### LINKED ISSUE #13765: [Bug]: 🐛 Arbitrary file access during archive extraction ("Zip Slip")
### Describe the bug

https://github.com/MetaMask/metamask-mobile/blob/aa3124907bff139f1ce3355712c5f29dff40be30/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java#L89-L89

Extracting files from a malicious zip file, or similar type of archive, is at risk of directory traversal attacks if filenames from the archive are not properly validated.

Zip archives contain archive entries representing each file in the archive. These entries include a file path for the entry, but these file paths are not restricted and may contain unexpected special elements such as the directory traversal element (..). If these file paths are used to create a filesystem path, then a file operation may happen in an unexpected location. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

if a zip file contains a file entry `..\sneaky-metamask`, and the zip file is extracted to the directory `c:\output`, then naively combining the paths would result in an output file path of `c:\output\..\metamask-file`, which would cause the file to be written to `c:\metamask-file`.

### Expected behavior

_No response_

### Screenshots/Recordings

_No response_

## POC
In this vulnerable, a file path taken from a zip archive item entry is combined with a destination directory. The result is used as the destination file path without verifying that the result is within the destination directory. If provided with a zip file containing an archive path like `..\metamask-file`, then this file would be written outside the destination directory.
```java
void writeZipEntry(ZipEntry entry, File destinationDir) {
    File file = new File(destinationDir, entry.getName());
    FileOutputStream fos = new FileOutputStream(file); // BAD
    // ... write entry to fos ...
}
```
To fix this vulnerability, we need to verify that the normalized `file` still has `destinationDir` as its prefix, and throw an exception if this is not the case.
```java
void writeZipEntry(ZipEntry entry, File destinationDir) {
    File file = new File(destinationDir, entry.getName());
    if (!file.toPath().normalize().startsWith(destinationDir.toPath()))
        throw new Exception("Bad zip entry");
    FileOutputStream fos = new FileOutputStream(file); // OK
    // ... write entry to fos ...
}
```
## References
- [Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal).
- [CWE-22](https://cwe.mitre.org/data/definitions/22.html).

```shell

```

### Detection stage

In production (default)

### Version

v7.40.0

### Build type

None

### Device

Android

### Operating system

Android

### Additional context

_No response_

### Severity

_No response_

FrederikBolding:
## **Description**

Prevent unsafe unpacking of tarballs that use relative paths in both the iOS and Android native modules used for `RNTar.unTar`.

This was not abusable in production since the code path is not currently reachable, but since the code will be used in the near future this should be fixed.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Closes: https://github.com/MetaMask/metamask-mobile/issues/8096

Closes: https://github.com/MetaMask/metamask-mobile/issues/13765

---

> [!NOTE]
> Add path validation during tar extraction on Android and iOS to block entries escaping the output directory.
> 
> - **Security: Tar extraction hardening**
>   - **Android (`android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java`)**:
>     - Validate each entry with canonical paths; reject if `outputFile.getCanonicalPath()` does not start with the canonical `outputPath`.
>   - **iOS (`ios/Light-Swift-Untar-V2/Light-Swift-Untar.swift`)**:
>     - Resolve symlinks and validate extracted `filePath`/`directoryPath` stay under resolved root; throw `UntarError.invalidPaths` on violation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5f08804a4af704909ec1d701ba2f4e00fbf40850. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/needs-e2e-build.yml
+++ b/.github/workflows/needs-e2e-build.yml
@@ -109,6 +109,7 @@ jobs:
               - 'sonar-project.properties'         # SonarQube config
               - 'scripts/*'                        # Build/utility scripts
               - 'scripts/*/**'                     # Subdirectories in scripts/   
+              - '!scripts/build.sh'                # But NOT build.sh
               - 'patches/**'                       # Dependency patches
               - 'ppom/**'                          # PPOM sub-project
 
--- a/.github/workflows/needs-e2e-build.yml
+++ b/.github/workflows/needs-e2e-build.yml
@@ -109,7 +109,6 @@ jobs:
               - 'sonar-project.properties'         # SonarQube config
               - 'scripts/*'                        # Build/utility scripts
               - 'scripts/*/**'                     # Subdirectories in scripts/   
-              - '!scripts/build.sh'                # But NOT build.sh
               - 'patches/**'                       # Dependency patches
               - 'ppom/**'                          # PPOM sub-project
 
--- a/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
+++ b/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
@@ -78,7 +78,7 @@ public class RNTar extends ReactContextBaseJavaModule {
       }
 
       File outPathFile = new File(outputPath);
-      String outPathCanonical = outPathFile.getCanonicalPath() + File.seperator;
+      String outPathCanonical = outPathFile.getCanonicalPath() + File.separator;
 
       // Set up the input streams for reading the .tgz file
       try (FileInputStream fileInputStream = new FileInputStream(tgzPath);
@@ -91,7 +91,7 @@ public class RNTar extends ReactContextBaseJavaModule {
         while ((entry = (TarArchiveEntry) tarInputStream.getNextEntry()) != null) {
           File outputFile = new File(outputPath, entry.getName());
 
-          if (!outPathFile.getCanonicalPath().startsWith(outPathCanonical)) {
+          if (!outputFile.getCanonicalPath().startsWith(outPathCanonical)) {
             throw new IOException("Tarball failed to extract due to invalid paths.");
           }
 
--- a/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
+++ b/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
@@ -77,7 +77,7 @@ public class RNTar extends ReactContextBaseJavaModule {
         throw new IOException("The output directory is not readable and/or writable.");
       }
 
-      File outPathFile = new File(outPathFile);
+      File outPathFile = new File(outputPath);
       String outPathCanonical = outPathFile.getCanonicalPath() + File.seperator;
 
       // Set up the input streams for reading the .tgz file
--- a/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
+++ b/android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
@@ -77,6 +77,9 @@ public class RNTar extends ReactContextBaseJavaModule {
         throw new IOException("The output directory is not readable and/or writable.");
       }
 
+      File outPathFile = new File(outPathFile);
+      String outPathCanonical = outPathFile.getCanonicalPath() + File.seperator;
+
       // Set up the input streams for reading the .tgz file
       try (FileInputStream fileInputStream = new FileInputStream(tgzPath);
            GZIPInputStream gzipInputStream = new GZIPInputStream(fileInputStream);
@@ -87,6 +90,10 @@ public class RNTar extends ReactContextBaseJavaModule {
         // Loop through the entries in the .tgz file
         while ((entry = (TarArchiveEntry) tarInputStream.getNextEntry()) != null) {
           File outputFile = new File(outputPath, entry.getName());
+
+          if (!outPathFile.getCanonicalPath().startsWith(outPathCanonical)) {
+            throw new IOException("Tarball failed to extract due to invalid paths.");
+          }
 
           // If it is a directory, create the output directory
           if (entry.isDirectory()) {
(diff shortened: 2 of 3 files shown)


.github/workflows/needs-e2e-build.yml
```diff
@@ -109,7 +109,6 @@ jobs:
               - 'sonar-project.properties'         # SonarQube config
               - 'scripts/*'                        # Build/utility scripts
               - 'scripts/*/**'                     # Subdirectories in scripts/   
-              - '!scripts/build.sh'                # But NOT build.sh
```
FrederikBolding, line 112: This is a temporary change to get CI to run on this branch

android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
```diff
@@ -77,6 +77,9 @@ private String extractTgzFile(String tgzPath, String outputPath) throws IOExcept
         throw new IOException("The output directory is not readable and/or writable.");
       }
 
+      File outPathFile = new File(outPathFile);
```
NEllusion: Should this be 
```suggestion
      File outPathFile = new File(outputPath);
```

NEllusion (commented):
Thanks for putting up a PR with this fix!

I'm not too familiar with how this is used, but I was curious to know the reason we chose to resolve symlinks when determining the root path rather than using something like `.standardizedFileURL` in swift or `.normalize()` in java which dont? 

It seems to be fine, but it adds some complexity that can introduce race conditions (e.g what happens if at the time we check the path, a symlink resolves it to our correct folder, but later something changes it to point to somewhere else when we go to write)

android/app/src/main/java/io/metamask/nativeModules/RNTar/RNTar.java
FrederikBolding: It should, I'm surprised it works like this

FrederikBolding:
> Thanks for putting up a PR with this fix!
> 
> I'm not too familiar with how this is used, but I was curious to know the reason we chose to resolve symlinks when determining the root path rather than using something like `.standardizedFileURL` in swift or `.normalize()` in java which dont?
> 
> It seems to be fine, but it adds some complexity that can introduce race conditions (e.g what happens if at the time we check the path, a symlink resolves it to our correct folder, but later something changes it to point to somewhere else when we go to write)

A tar file can contain symlinks as well, so seems worth it to resolve them, since read or write likely will later in the process. I don't think an NPM package ever needs a symlink though, so it is strictly to guard against reading/writing in an unexpected place outside of the archive we are extracting.

.github/workflows/needs-e2e-build.yml
FrederikBolding: Will revert this after getting a clean E2E run

joaoloureirop (approved):

NEllusion (commented):
Only risk to note is that there is a small delay between when we resolve symlinks for a path, and when we actually perform the write. If something that previously was under our `rootPath` changes (due to a symlink being created after this resolution), it would allow escaping the intended target directory.

I can't imagine this feasibly happening though as we don't process symlinks from tar files today.
 
Can we add tests to ensure our checks do catch any errors that we may not expect? 

There was an issue I had noted down from taking a look recently, but it looks like you caught it  yesterday (e.g [this](https://github.com/MetaMask/metamask-mobile/pull/21408/commits/4374058251b73ca81103a565fa6f108da01b537c)) 🎉 

However, it goes to show how subtle some of the changes here can be. Having the regression tests will ensure the next time someone makes changes here, we can flag if it introduces a gap in our security.

FrederikBolding:
> Only risk to note is that there is a small delay between when we resolve symlinks for a path, and when we actually perform the write. If something that previously was under our `rootPath` changes (due to a symlink being created after this resolution), it would allow escaping the intended target directory.
> 
> I can't imagine this feasibly happening though as we don't process symlinks from tar files today.
> 
> Can we add tests to ensure our checks do catch any errors that we may not expect?
> 
> There was an issue I had noted down from taking a look recently, but it looks like you caught it yesterday (e.g [this](https://github.com/MetaMask/metamask-mobile/pull/21408/commits/4374058251b73ca81103a565fa6f108da01b537c)) 🎉
> 
> However, it goes to show how subtle some of the changes here can be. Having the regression tests will ensure the next time someone makes changes here, we can flag if it introduces a gap in our security.

FWIW we test the NPM unpacking flow in E2E, but an issue with the CI caused it to use an older build which means my typos above were uncaught. It broke instantly when running locally 😄 

We don't seem to have unit tests of any of the native code setup, I think it is out of scope for this PR to set that up.

PR #21412, opened 2025-10-20
refactor: Simplify NPM fetching logic

### LINKED ISSUE #3617: Review NPM installation implementation on mobile
I think it assumes one installation at a time right now, will need to be fixed if we ship OTA.

FrederikBolding:
## **Description**

Simplify the NPM fetching implementation to make it more readable, while also fixing a couple of issues:
- Use the cache directory for the app instead of a shared directory
- Re-use the cache key for the unpacked directory to allow for multiple installs at a time
- Properly clean up files on the filesystem after install
- Match error messages and states from the extension implementation

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Closes: https://github.com/MetaMask/snaps/issues/3617 

---

> [!NOTE]
> Refactors NPM tarball fetching with direct blob fetch, status/size validation, improved error handling, and robust cleanup; updates tests to match.
> 
> - **Core (NPM fetch)**:
>   - Rewrite `NpmLocation.fetchNpmTarball` to use `ReactNativeBlobUtil.config({ appendExt: 'tgz' }).fetch` and `RNTar.unTar`.
>   - Add response validation: assert 404 not-found, require 200 OK, enforce `content-length` against `TARBALL_SIZE_SAFETY_LIMIT`.
>   - Improve errors using `getErrorMessage`; accept `DetectSnapLocationOptions` via constructor.
>   - Ensure cleanup in `finally`: `response.flush()` and `fs.unlink` of unpacked path.
>   - Read files as UTF-8 strings in `readAndParseAt`; remove `stringToBytes`.
>   - Remove helpers: `decompressFile`, `fetchAndStoreNPMPackage`, `cleanupFileSystem`; drop `SNAPS_NPM_LOG_TAG`.
> - **Tests**:
>   - Update mocks to include `respInfo` with `status` and `content-length`; adjust FS mocks; align with new fetch/untar flow.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 46c1840b24153959f60b5d262f4b20edc434c982. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/core/Snaps/location/npm.ts
+++ b/app/core/Snaps/location/npm.ts
@@ -1,7 +1,7 @@
 /* eslint-disable import/prefer-default-export */
 ///: BEGIN:ONLY_INCLUDE_IF(preinstalled-snaps,external-snaps)
 import { VirtualFile } from '@metamask/snaps-utils';
-import { assert, getErrorMessage, stringToBytes } from '@metamask/utils';
+import { assert, getErrorMessage } from '@metamask/utils';
 import { NativeModules } from 'react-native';
 import ReactNativeBlobUtil, { FetchBlobResponse } from 'react-native-blob-util';
 import {
@@ -25,9 +25,7 @@ const findAllPaths = async (path: string): Promise<string[]> => {
 };
 
 const readAndParseAt = async (path: string) => {
-  const contents = stringToBytes(
-    await ReactNativeBlobUtil.fs.readFile(path, 'utf8'),
-  );
+  const contents = await ReactNativeBlobUtil.fs.readFile(path, 'utf8');
   return { path, contents };
 };
 
--- a/app/core/Snaps/location/npm.ts
+++ b/app/core/Snaps/location/npm.ts
@@ -6,6 +6,7 @@ import { NativeModules } from 'react-native';
 import ReactNativeBlobUtil, { FetchBlobResponse } from 'react-native-blob-util';
 import {
   BaseNpmLocation,
+  DetectSnapLocationOptions,
   getNpmCanonicalBasePath,
   TARBALL_SIZE_SAFETY_LIMIT,
 } from '@metamask/snaps-controllers';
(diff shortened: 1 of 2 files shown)


app/core/Snaps/location/npm.ts
```diff
@@ -41,87 +25,91 @@ const findAllPaths = async (path: string): Promise<string[]> => {
...
-};
-
-const cleanupFileSystem = async (path: string) => {
-  ReactNativeBlobUtil.fs.unlink(path).catch((error) => {
-    throw new Error(
-      `${SNAPS_NPM_LOG_TAG} cleanupFileSystem failed to clean files at path with error: ${error}`,
-    );
-  });
+  const contents = stringToBytes(
+    await ReactNativeBlobUtil.fs.readFile(path, 'utf8'),
```
Mrtenz: Why read as UTF-8 just to turn it into bytes again? Does it not support reading as buffer directly?

app/core/Snaps/location/npm.ts
```diff
@@ -41,87 +25,91 @@ const findAllPaths = async (path: string): Promise<string[]> => {
...
-          path: normalizedPath,
-          data: { canonicalPath: new URL(path, canonicalBase).toString() },
-        }),
+    let response: FetchBlobResponse | null = null;
+    let untarPath: string | null = null;
+    try {
+      response = await ReactNativeBlobUtil.config({
+        fileCache: true,
+        appendExt: 'tgz',
+      }).fetch('GET', tarballUrl.toString());
```
Mrtenz: Is it necessary to call `config` every time?

Mrtenz (changes requested):

app/core/Snaps/location/npm.ts
FrederikBolding: Probably not, will check

app/core/Snaps/location/npm.ts
FrederikBolding: Seemingly only supports utf8, ascii and base64.

FrederikBolding: But I think we handle both so we may be able to just use string, I'll check

Mrtenz (dismissed):

Mrtenz (dismissed):

Mrtenz (approved):

PR #21869, opened 2025-10-29
chore: Consolidate link types and implement normalizer (Phase 1 - PR 1)

smilingkylan:
# Deep Link Consolidation - Phase 1: Core Types & Normalizer

Fixes:
https://github.com/MetaMask/mobile-planning/issues/2343
https://consensyssoftware.atlassian.net/jira/polaris/projects/MWMR/ideas/view/8845676?selectedIssue=MWMR-10

## Overview
This PR introduces the foundational infrastructure for consolidating MetaMask Mobile's fragmented deep link handling system. It creates a unified representation and normalization layer for all deep links, establishing a consistent pattern for handling `metamask://` and `https://` protocols.

## Problem
The current deep link system has routing logic spread across multiple files using nested switch-case statements, making it difficult to:
- Understand the complete set of supported deep links
- Maintain consistency across different protocols
- Add new deep link actions
- Test deep link handling comprehensively

## Solution
This PR implements:
1. **Unified Type System** - Single source of truth for deep link structure
2. **CoreLinkNormalizer** - Converts all deep link formats to a standardized representation
3. **Protocol Conversion** - Bidirectional transformation between `metamask://` and `https://` formats
4. **Comprehensive Testing** - 30 test cases covering all functionality

### Run Tests
```bash
yarn jest app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```

**Expected Output:**
```
PASS app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
  ✓ All 30 tests passing
```

### Run Linting
```bash
yarn eslint 'app/core/DeeplinkManager' --fix
```

### Run TypeScript Check
```bash
yarn lint:tsc
```

All commands should pass without errors.

### Normalizing a Deep Link
```typescript
import { CoreLinkNormalizer } from './core/DeeplinkManager/CoreLinkNormalizer';

// Normalize a metamask:// link
const link = CoreLinkNormalizer.normalize(
  'metamask://swap?from=ETH&to=DAI&amount=100',
  'qr-code'
);

console.log(link);
// {
//   protocol: 'metamask',
//   action: 'swap',
//   params: {
//     from: 'ETH',
//     to: 'DAI',
//     amount: '100',
//     swapPath: 'swap?from=ETH&to=DAI&amount=100'
//   },
//   source: 'qr-code',
//   timestamp: 1698765432000,
//   originalUrl: 'metamask://swap?from=ETH&to=DAI&amount=100',
//   normalizedUrl: 'https://link.metamask.io/swap?from=ETH&to=DAI&amount=100',
//   isValid: true,
//   isSupportedAction: true,
//   isPrivateLink: false,
//   requiresAuth: false
// }
```

### Converting Protocols
```typescript
// Convert to metamask:// protocol
const metamaskUrl = CoreLinkNormalizer.toMetaMaskProtocol(link);
// 'metamask://swap?from=ETH&to=DAI&amount=100'
```

### Building Deep Links
```typescript
// Build a new deep link
const sendLink = CoreLinkNormalizer.buildDeeplink('metamask', 'send', {
  to: '0x1234567890abcdef',
  value: '1000000000000000000',
  hr: true
});
// 'metamask://send?to=0x1234567890abcdef&value=1000000000000000000&hr=1'
```

### Validating Deep Links
```typescript
// Check if a URL is a supported deep link
const isValid = CoreLinkNormalizer.isSupportedDeeplink('metamask://swap');
// true

const isInvalid = CoreLinkNormalizer.isSupportedDeeplink('metamask://unknown-action');
// false
```

## Backward Compatibility
This PR introduces new infrastructure but **does not modify existing deep link handling**. The new system runs in parallel with the existing system and has no impact on current functionality.

The next PR (Legacy Adapter) will provide the bridge between this new system and the existing deep link handlers.

### Follows Guidelines
- ✅ AAA test pattern (Arrange, Act, Assert)
- ✅ No "should" in test descriptions
- ✅ One behavior per test
- ✅ Comprehensive edge case coverage
- ✅ Proper TypeScript types (no `any`)
- ✅ All dependencies mocked
- ✅ JSDoc comments on public methods
- ✅ MetaMask coding standards

## Next Steps
**Phase 1, PR 2**: Legacy Adapter
- Create `LegacyLinkAdapter` to bridge new and old systems
- Implement conversion between CoreUniversalLink and legacy formats
- Enable gradual migration without breaking changes

**Phase 2**: Universal Router
- Create `UniversalRouter` singleton
- Implement handler registry
- Define handler interfaces

**Phase 3**: Integration
- Integrate router into DeeplinkManager
- Add feature flag for gradual rollout
- Full migration from legacy system

## Risk Assessment
**Risk Level**: Low

**Rationale:**
- No modifications to existing code
- Isolated to new files
- Comprehensive test coverage
- No runtime impact until integration

## Related Documentation
- Project Prompt: `/DEEPLINK_CONSOLIDATION_PROMPT.md`
- Implementation Summary: `/.kylan-docs/deep-link-consolidation-pr1-summary.md`

---

## Reviewer Checklist
- [ ] All tests pass (`yarn jest app/core/DeeplinkManager/CoreLinkNormalizer.test.ts`)
- [ ] No linting errors (`yarn eslint 'app/core/DeeplinkManager' --fix`)
- [ ] TypeScript compiles without errors (`yarn lint:tsc`)
- [ ] Code follows MetaMask coding guidelines
- [ ] Test coverage is comprehensive (30 test cases)
- [ ] No breaking changes to existing functionality
- [ ] JSDoc comments are clear and accurate
- [ ] Type definitions are appropriate and safe

---

> [!NOTE]
> Introduces a unified deep link normalizer and core link types, adds tests, updates hr handling, and registers the ramp action.
> 
> - **Core deep link infrastructure**:
>   - **CoreLinkNormalizer** (`app/core/DeeplinkManager/CoreLinkNormalizer.ts`): normalizes deeplinks, converts protocols, validates support, and builds links.
>   - **Types** (`app/core/DeeplinkManager/types/CoreUniversalLink.ts`): defines `CoreUniversalLink`/`CoreLinkParams` and action/protocol groupings (auth-required, SDK, ramp, perps, supported protocols, default action).
>   - **Tests** (`app/core/DeeplinkManager/CoreLinkNormalizer.test.ts`): covers normalization, protocol conversion, validation, and param handling.
> - **Deeplink handling tweaks**:
>   - Ensures `hr` is treated boolean (`!!params.hr`) in `handleMetaMaskDeeplink` and parsed from `'1'` in `extractURLParams`.
> - **Constants**:
>   - Adds `ACTIONS.RAMP` to `app/constants/deeplinks.ts`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eb455f3dca17d7ad3da6cb321063bd825319e45b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/core/DeeplinkManager/types/CoreUniversalLink.ts
```diff
@@ -0,0 +1,162 @@
...
+  (SDK_ACTIONS as readonly string[]).includes(action);
+
+export const isWhitelistedAction = (action: string): boolean =>
+  (WHITELISTED_ACTIONS as readonly string[]).includes(action);
+
+export const isRampAction = (action: string): boolean =>
+  (RAMP_ACTIONS as readonly string[]).includes(action);
+
+export const isPerpsAction = (action: string): boolean =>
+  (PERPS_ACTIONS as readonly string[]).includes(action);
```
weitingsun: why not pass in ACTIONS as second argument so we can reduce these 5 functions to 1? I can see arguments on both side so this is not a blocker ofc

weitingsun (dismissed):
LGTM

app/core/DeeplinkManager/types/CoreUniversalLink.ts
smilingkylan: I actually think it's nice syntactic sugar since these functions tell you what they do (instead of something vague like `const isTypedAction = (action: string, actionList: string[] => ...`

smilingkylan: And at that point why even have the functions at all?

weitingsun (dismissed):

app/core/DeeplinkManager/types/CoreUniversalLink.ts
NicolasMassart: This is what Fowler calls [Code as Doc](https://martinfowler.com/bliki/CodeAsDocumentation.html), I like it, and it makes AI understand code better. Naming things is sometimes the most important part.

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+    it('converts hr parameter from string to boolean', () => {
+      const url = 'metamask://home?hr=1';
+      const source = 'test';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.params.hr).toBe(true);
+    });
+
+    it('handles hr parameter with value 0', () => {
```
NicolasMassart: S: avoid use of generic meaningless words like handle, manage, ...
```suggestion
    it('uses 0 value hr as a boolean', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = 'metamask://home?hr=0';
+      const source = 'test';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.params.hr).toBe(false);
+    });
+
+    it('handles SDK parameters', () => {
```
NicolasMassart:
```suggestion
    it('extracts SDK parameters', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+
+      expect(result.action).toBe('connect');
+      expect(result.params.channelId).toBe('123');
+      expect(result.params.comm).toBe('socket');
+      expect(result.params.pubkey).toBe('abc');
+      expect(result.params.v).toBe('2');
+      expect(result.requiresAuth).toBe(false);
+    });
+
+    it('handles attribution parameters', () => {
```
NicolasMassart:
```suggestion
    it('extracts attribution parameters', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = 'metamask://send?to=0x123';
+      const source = 'test';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.action).toBe('send');
+      expect(result.requiresAuth).toBe(true);
+    });
+
+    it('handles ramp actions with paths', () => {
```
NicolasMassart:
```suggestion
    it('extracts ramp actions with paths', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.action).toBe('buy-crypto');
+      expect(result.params.rampPath).toBe('buy-crypto?amount=100&currency=USD');
+      expect(result.params.amount).toBe('100');
+      expect(result.params.currency).toBe('USD');
+    });
+
+    it('handles perps actions', () => {
```
NicolasMassart:
```suggestion
    it('extracts perps actions params', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = `https://${AppConstants.MM_IO_UNIVERSAL_LINK_HOST}/perps-asset/ETH-USD`;
+      const source = 'perps';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.action).toBe('perps-asset');
+      expect(result.params.perpsPath).toBe('perps-asset/ETH-USD');
+    });
+
+    it('defaults to home action when no action specified', () => {
```
NicolasMassart: Defaulting and when no action specified are redundant to me
```suggestion
    it('defaults to home action', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = `https://${AppConstants.MM_IO_UNIVERSAL_LINK_HOST}/`;
+      const source = 'test';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.action).toBe('home');
+    });
+
+    it('handles invalid URLs gracefully', () => {
```
NicolasMassart:
```suggestion
    it('identifies invalid URLs as invalid and unsupported', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = 'not-a-valid-url';
+      const source = 'test';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.isValid).toBe(false);
+      expect(result.isSupportedAction).toBe(false);
+      expect(result.originalUrl).toBe(url);
```
NicolasMassart: S: Returning original url is not specific to invalid urls, both invalid and valid return have it

```suggestion
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+    });
+
+    it('handles message parameter with spaces', () => {
```
NicolasMassart: C: replacing spaces with `+` is called form-style URL encoding
```suggestion
    it('encodes spaces as form style', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+        // Already a boolean, keep as is
+      } else {
+        params.hr = false;
+      }
+    }
+
+    // Clean up message parameter
+    if (params.message) {
+      params.message = params.message.replace(/ /g, '+');
+    }
```
NicolasMassart, line 210: S: I would advise not to manually sanitize url by replacing spaces with + but use encodeURIComponent() or build a new URLSearchParams

Also, not manually dealing with encoding and using qs or URL/URLSearchParams everywhere would makes encoding, decoding, and rebuilding symmetric and spec-compliant and this code useless.

But if you want to preserve it, here is a suggestion:

```suggestion
    if (params.message) {
      params.message = encodeURIComponent(params.message);
    }
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.params.from).toBe('ETH');
+      expect(result.params.to).toBeUndefined();
+      expect(result.params.amount).toBeUndefined();
+      expect(result.params.empty).toBeUndefined();
+    });
+
+    it('handles account parameter format', () => {
```
NicolasMassart:
```suggestion
    it('extracts account parameter', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = 'metamask://send?account=0x123@1';
+      const source = 'test';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.params.account).toBe('0x123@1');
+    });
+
+    it('handles wc action', () => {
```
NicolasMassart:
```suggestion
    it('extracts wc action', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = 'metamask://wc?uri=wc:123';
+      const source = 'wallet-connect';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.action).toBe('wc');
+      expect(result.params.uri).toBe('wc:123');
+    });
+
+    it('handles onboarding action', () => {
```
NicolasMassart:
```suggestion
    it('extracts onboarding action', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+      const url = `https://${AppConstants.MM_IO_UNIVERSAL_LINK_HOST}/onboarding/step1`;
+      const source = 'onboarding';
+
+      const result = CoreLinkNormalizer.normalize(url, source);
+
+      expect(result.action).toBe('onboarding');
+      expect(result.params.onboardingPath).toBe('onboarding/step1');
+    });
+
+    it('handles create-account action', () => {
```
NicolasMassart:
```suggestion
    it('extracts create-account action', () => {
```

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -0,0 +1,367 @@
...
+        isValid: true,
+        isSupportedAction: true,
+        isPrivateLink: false,
+        requiresAuth: false,
+      };
+
+      const result = CoreLinkNormalizer.toMetaMaskProtocol(link);
+
+      expect(result).toBe('metamask://home?hr=1');
+    });
```
NicolasMassart: S: group with "handles hr parameter with value 0'" test

NicolasMassart (dismissed):
First pass review, haven't had the time to go through all the code.
The code is over complicated imo. This can be simplified.
One low hanging improvement would be to rely on standards for encoding instead of trying to deal with cases individually and manually. Converting spaces with replace should not even be an issue. See my comments on encoding in the file.

app/core/DeeplinkManager/CoreLinkNormalizer.test.ts
```diff
@@ -75,16 +75,36 @@ describe('CoreLinkNormalizer', () => {
...
+        isValid: true,
+        isSupportedAction: true,
+        isPrivateLink: false,
+        requiresAuth: false,
+      };
+
+      const generatedMetamaskLink =
+        CoreLinkNormalizer.toMetaMaskProtocol(generatedLink);
+
+      expect(generatedMetamaskLink).toBe('metamask://home?hr=1');
```
NicolasMassart: I meant group under the same `describe`, but not int he same `it`.
Also try to avoid breaking the AAA pattern.

bfullam (dismissed):

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+    // Clean up message parameter
+    if (params.message) {
+      params.message = encodeURIComponent(params.message);
+    }
+
+    return params;
+  }
+
+  private static parseQueryString(search: string): Partial<CoreLinkParams> {
+    if (!search || search === '?') {
```
NicolasMassart: S: you can use searchParams to know if there's params.

```ts
    if ([...url.searchParams].length > 0) {...}
```
This would force to use a URL type param for parseQueryString instead of an untyped string, more typed is better

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+      return {};
+    }
+
+    try {
+      const parsed = qs.parse(search.substring(1), { arrayLimit: 99 });
```
NicolasMassart: Q: Why 99? Seems like a magic number without real reason to be. The risk is that it could just be too big and then defeats the purpose of preventing abuse or being too low and break when we have more values. A limit like this has to be defined base on actual need.

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+      params.rewardsPath = actionPath;
+    } else if (action === 'home') {
+      params.homePath = actionPath;
+    } else if (action === 'onboarding') {
+      params.onboardingPath = actionPath;
+    } else if (action === 'create-account') {
+      params.createAccountPath = actionPath;
+    } else if (action === 'deposit') {
+      params.depositCashPath = actionPath;
+    }
```
NicolasMassart: S: for this kind of test, I would favour a switch/case. Why not?

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+      'swapPath',
+      'dappPath',
+      'sendPath',
+      'perpsPath',
+      'rewardsPath',
+      'homePath',
+      'onboardingPath',
+      'createAccountPath',
+      'depositCashPath',
+    ];
```
NicolasMassart: Does this list have to be maintained? If so, how? What's the TS mechanism to enforce update need?

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+        value !== undefined &&
+        value !== ''
+      ) {
+        if (key === 'hr' && typeof value === 'boolean') {
+          filteredParams[key] = value ? '1' : '0';
+        } else {
+          filteredParams[key] = String(value);
+        }
+      }
+    });
```
NicolasMassart, line 321: C: this strangely looks like code on line 214-227. Can you see if there's a way to extract a helper function? (and test it)

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,329 @@
...
+
+    return Object.keys(filteredParams).length > 0
+      ? qs.stringify(filteredParams)
+      : '';
+  }
+
+  private static isSupportedAction(action: string): boolean {
+    // Check against all action constants
+    // Import is done dynamically to avoid circular dependencies
```
NicolasMassart: Q: Can you explain this? If we have circular dep, is it an architecture design issue that should be fixed instead of working around it?

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,337 @@
...
+    [ACTIONS.CREATE_ACCOUNT]: 'createAccountPath',
+    [ACTIONS.DEPOSIT]: 'depositCashPath',
+    [ACTIONS.PERPS]: 'perpsPath',
+    [ACTIONS.PERPS_MARKETS]: 'perpsMarketsPath',
+    [ACTIONS.PERPS_ASSET]: 'perpsAssetPath',
+    [ACTIONS.BUY]: 'buyPath',
+    [ACTIONS.SELL]: 'sellPath',
+    [ACTIONS.BUY_CRYPTO]: 'buyCryptoPath',
+    [ACTIONS.SELL_CRYPTO]: 'sellCryptoPath',
+  };
```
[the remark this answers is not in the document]
smilingkylan: For the record Ramps and Perps are special cases because they have multiple possible actions underneath them

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,334 @@
...
```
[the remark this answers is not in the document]
smilingkylan: Ramps and perps are special cases because they have multiple possible actions

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -256,26 +285,17 @@ export class CoreLinkNormalizer {
...
-    } else if (action === 'home') {
-      params.homePath = actionPath;
-    } else if (action === 'onboarding') {
-      params.onboardingPath = actionPath;
-    } else if (action === 'create-account') {
-      params.createAccountPath = actionPath;
-    } else if (action === 'deposit') {
-      params.depositCashPath = actionPath;
+    } else {
+      const pathKey = this.ACTION_PATH_MAP[action];
```
NicolasMassart, line 288: Q: why ACTION_PATH_MAP not usable for rampPath and perpsPath ?

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -0,0 +1,334 @@
+/**
+ * CoreLinkNormalizer - Unified deep link normalization
+ *
+ * Converts various deep link formats into a standardized CoreUniversalLink format
+ */
+import qs from 'qs';
+import { PROTOCOLS, ACTIONS } from '../../constants/deeplinks';
```
NicolasMassart, line 6: Q: So the circular dep disappeared? What happened?

smilingkylan: I'm not sure why that part was added in earlier. The deeplinks.ts file doesn't import anything so no risk of circular dependency

app/core/DeeplinkManager/CoreLinkNormalizer.ts
smilingkylan: Because there are multiple actions that fall under the ramp and perps path

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -85,8 +85,9 @@ export class CoreLinkNormalizer {
         isPrivateLink: false, // Will be determined by signature verification
         requiresAuth: isAuthRequiredAction(action),
       };
-    } catch (error) {
-      // Return invalid link representation
+    } catch (_error) {
+      // for some reason SonarQube doesn't think this
+      // counts as "handling" the error
```
NicolasMassart: Because it's actually what happens. You do things but the error variable is not used. So for Sonar, you are doing something but ignoring the error. Which is true, you kinda ignore it because you don't even look at waht error it is and just fallback. So you can mark this sonar issue as accepted.

app/core/DeeplinkManager/CoreLinkNormalizer.ts
```diff
@@ -274,9 +275,10 @@ export class CoreLinkNormalizer {
     const path = pathSegments.join('/');
     const query = urlObj.search;
 
-    return path
+    const output = path
       ? `${action}${path ? '/' + path : ''}${query}`
       : `${action}${query}`;
+    return output;
```
NicolasMassart: I don't think removing the direct return will fix this sonar issue.
The real problem is the 2 ternaries inside each other.

ternary 1: ``path ? `${action}${[ternary 2]}${query}` : `${action}${query}`;``
ternary 2: ``path ? '/' + path : ''``

Anyway if you think this "matrioshka" ternary is fine, just mark the Sonar issue as accepted

app/core/DeeplinkManager/CoreLinkNormalizer.ts
smilingkylan: I marked it as accepted. It's just annoying because we also can't leave the `(error)` part blank or we'll get other errors

app/core/DeeplinkManager/CoreLinkNormalizer.ts
smilingkylan: Ah ok I misread the issue

app/core/DeeplinkManager/CoreLinkNormalizer.ts
NicolasMassart: When you say "I'm not sure why that part was added in earlier." what do you mean? It's been added automatically? By IDE?

app/core/DeeplinkManager/CoreLinkNormalizer.ts
smilingkylan: @NicolasMassart When I tried using `encodeURIComponent` the AI stuff didn't like like it:

The `message` parameter contains **base64-encoded data**, not plain text that needs URL encoding.

When base64 strings are transmitted through URLs, the `+` character (which is part of the base64 alphabet) gets converted to a space by URL parsers. This corrupts the base64 string.

Using `encodeURIComponent()` would percent-encode the base64 string (e.g., `SGVsbG8+V29ybGQ=` → `SGVsbG8%2BV29ybGQ%3D`), which would break the downstream decoder that expects a valid base64 string:

```javascript
// From DeeplinkProtocolService.ts:669
const parsedMessage = Buffer.from(params.message, 'base64').toString('utf-8');
```

The current `replace(/ /g, '+')` approach simply reverses the URL parsing damage by restoring spaces back to `+`, preserving the valid base64 format needed for successful decoding. This is a common pattern when handling base64 data in URLs.

NicolasMassart: You're right that `+` is valid in base64 and gets turned into spaces by URL parsers. But instead of fixing it with a regex, the proper approaches are:

1. **Best:** switch to **base64url** encoding — it’s URL-safe (no `+`, `/`, or `=`).
2. **If keeping base64:** use `encodeURIComponent` when building the URL and `decodeURIComponent` before `Buffer.from(..., 'base64')`. That’s the standard, safe way.
3. **If legacy format must stay:** configure `qs` to preserve `+` instead of post-processing `message`.

Regex replacement is only a quick patch — better to solve it at the encoding or parsing level.

weitingsun:
LGTM from my review, defer to Nico's comments to be resolved

app/core/DeeplinkManager/CoreLinkNormalizer.ts
smilingkylan: @NicolasMassart The regex follows the pattern that has been in our code base for two years.

This PR is part of a much larger refactor and I'd prefer to not change too much of the inner-workings until after the refactor is complete. We have to be careful about changing encoding / decoding to make sure we are backwards compatible but we could definitely make a new issue for investigating ways to clean up that code

https://github.com/MetaMask/metamask-mobile/blob/3b9734560e0feab65d7518b3fcf2ade99fc107c3/app/core/DeeplinkManager/ParseManager/extractURLParams.ts#L65

app/core/DeeplinkManager/ParseManager/extractURLParams.ts
```diff
@@ -59,7 +59,11 @@ function extractURLParams(url: string) {
       const parsedParams = qs.parse(urlObj.query.substring(1), {
         arrayLimit: 99,
       });
-      params = { ...params, ...parsedParams, hr: parsedParams.hr === '1' };
+      params = {
+        ...params,
+        ...parsedParams,
+        hr: ['1', 'true'].includes(parsedParams.hr as string),
```
smilingkylan: From @EdouardBougon "this value defaults to false if not provided (0=false, 1=true, undefined=false). So I am going to enforce this rule going forward

weitingsun (approved):

PR #22023, opened 2025-10-31
feat: mcwp176 add auto tracking for related feature flags and override

asalsys:
## **Description**

This PR implements automatic tracking of related feature flags in analytics events. When feature flags are accessed via `getFeatureFlag()`, their states are automatically captured as snapshots. These snapshots are then included in all analytics events tracked through the `useMetrics` hook, providing better context for understanding user behavior in relation to feature flag configurations.

**Key changes:**
1. **Feature Flag Snapshot Tracking**: Added snapshot functionality to `FeatureFlagOverrideContext` that captures feature flag states (including key, value, type, description, override status) with timestamps when flags are accessed
2. **Automatic Event Enhancement**: Modified `useMetrics` hook to automatically include related feature flags in all analytics events via `generateGenericProperties()`

**Benefits:**
- Enables correlation between user actions and feature flag states
- Helps identify which feature flags influence user behavior
- Provides automatic context without requiring manual property addition in each tracking call
- Supports data-driven decision making for feature flag rollouts

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: 
[MCWP-176 Jira Ticket](https://consensyssoftware.atlassian.net/browse/MCWP-176)
[github Issue](https://github.com/MetaMask/mobile-planning/issues/2360)
## **Manual testing steps**

```gherkin
Feature: Related feature flags in analytics events

  Scenario: Feature flags are automatically included in analytics events
    Given the app is running with feature flag override functionality enabled
    When a feature flag is accessed via getFeatureFlag()
    Then a snapshot of that flag (key, value, type, description, override status, timestamp) is captured

  Scenario: Analytics events include related flags
    Given the app has accessed one or more feature flags
    When an analytics event is tracked using useMetrics().trackEvent()
    Then the event properties automatically include a relatedFlags array containing snapshots of accessed flags
    And each snapshot includes the flag key, value, originalValue, type, description, isOverridden status, and timestamp

  Scenario: Multiple events capture the same flags
    Given a feature flag has been accessed
    When multiple analytics events are tracked
    Then each event includes the relatedFlags property with the same flag snapshots
```

---

> [!NOTE]
> Automatically captures boolean feature flag states via `useMetrics.addTraitsToUser` when `getFeatureFlag` is called, including evaluated values for `boolean with minimumVersion`, with comprehensive tests added.
> 
> - **Feature flags/analytics integration**:
>   - In `app/contexts/FeatureFlagOverrideContext.tsx`, integrate `useMetrics` and call `addTraitsToUser` when `getFeatureFlag` is invoked:
>     - Snapshots for `boolean` flags using current value.
>     - Snapshots for `boolean with minimumVersion` using the validated result from `validateMinimumVersion`.
>   - Update `getFeatureFlag` dependencies to include `addTraitsToUser`.
> - **Tests**:
>   - In `app/contexts/FeatureFlagOverrideContext.test.tsx`:
>     - Mock `useMetrics` and verify snapshots are sent for boolean flags (including min-version cases) and not for non-boolean flags.
>     - Add `waitFor` usage for async assertions.
>     - Switch environment checks from `process.env.NODE_ENV` to `process.env.METAMASK_ENVIRONMENT` in version-related tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f8412fc9aae5ee23dc0f90b5536fa3260e42b193. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


NicolasMassart (changes requested):
I think inserting properties at the hook level is not a good idea.
- it only applies to the hook but metrics are used also directly from the instance
- it's out of the scope of the hook to deal with properties: we already have properties generated and they are passed when needed to the event builder, but passing them at the hook level is a risk of confusion and override when other props are passed later.

If data has to be added so we know what flags user has, it has to be on profile to me using `addUSerTraits`.

See how the profile is enriched with data using app/util/metrics/UserSettingsAnalyticsMetaData or app/util/metrics/DeviceAnalyticsMetaData

Can you provide more detail on how necessary it is to update any of the metametrics code for this feature and why it should be updated on every event sent with the hook?

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -51,13 +55,25 @@ interface FeatureFlagOverrideProviderProps {
...
   // Local state for overrides
   const [overrides, setOverrides] = useState<FeatureFlagOverrides>({});
+  const [featureFlagSnapshots, setFeatureFlagSnapshots] = useState<
+    FeatureFlagInfoSnapshot[]
+  >([]);
+
+  useEffect(() => {
+    if (featureFlagSnapshots.length > 0) {
+      addTraitsToUser({
+        relatedFlags: JSON.stringify(featureFlagSnapshots),
```
NicolasMassart: Q: Stringify this means that it will very hardly be searchable and filterable in analytics like Mixpanel. Is that what we want instead of flattening the values as profile props directly?

S: If so, comment the code to explain the reason for future devs to know why it's done this way.

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -66,6 +82,16 @@ export const FeatureFlagOverrideProvider: React.FC<
...
 
+  const takeSnapshot = useCallback((featureFlagInfo: FeatureFlagInfo) => {
+    setFeatureFlagSnapshots((prev) => [
+      ...prev,
+      {
+        timestamp: Date.now(),
+        ...featureFlagInfo,
+      },
+    ]);
+  }, []);
```
NicolasMassart: Q: I'm not sure, why do you accumulate featureFlagInfo?

- it will increate the size of the feature flag info json in identify prop.
- this creates a risk of max event size to be reached ([Segment has 32Kb limit per message](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/faq/#is-there-a-size-limit-on-requests)) 
- the memory leak noticed by bugbot is real

The only reason I see to do that is to keep history of changes of the values. Do we want to keep history? Why?
If so, can't we find a better way than feeding it into Segment as a cumulated json? If we sent a snapshot of the flags to segment user profile as traits, can we not handle the history server side by feeding a table with new values without erasing old ones?

NicolasMassart (changes requested):
This code change makes me uncomfortable as the initial issue is not asking for any history recording of the flags, but that's what the code does.
I think the implementation is not what we need and has too many side effect (memory leak, makes it hard to use data as it's json stringified, ...)
We have to also take the Segment limits into account as this implementation will lead to issues about message size limit at some point.
I would suggest a live pair review of this and discuss alternative option.
I will DM you in Slack.

app/contexts/FeatureFlagOverrideContext.tsx
asalsys: Your right I will just keep a timestamp of the latest request.

app/contexts/FeatureFlagOverrideContext.tsx
asalsys: removing

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -19,6 +21,8 @@ import {
   ToastVariants,
 } from '../component-library/components/Toast';
 import { MinimumVersionFlagValue } from '../components/Views/FeatureFlagOverride/FeatureFlagOverride';
+import useMetrics from '../components/hooks/useMetrics/useMetrics';
+import { JsonValue } from '@segment/analytics-react-native';
```
NicolasMassart: C: You should not have this low level dependency here. The props for addUserTraits is UserTraits which includes JsonMap which includes JsonValue.


```suggestion
import { UserTraits } from '@segment/analytics-react-native';
```

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -48,16 +52,31 @@ interface FeatureFlagOverrideProviderProps {
...
 
   // Local state for overrides
   const [overrides, setOverrides] = useState<FeatureFlagOverrides>({});
+  const [featureFlagSnapshots, setFeatureFlagSnapshots] =
+    useState<RelatedFlagsSnapshotType>({});
+
+  useEffect(() => {
+    if (Object.keys(featureFlagSnapshots).length > 0) {
+      addTraitsToUser({
+        relatedFlags: featureFlagSnapshots as unknown as JsonValue,
```
NicolasMassart:
```suggestion
        relatedFlags: featureFlagSnapshots as unknown as UserTraits,
```

NicolasMassart: Q: This will push an object value into relatedFlags in profile. It's definitely better than a JSON string, but not the best either as this is discouraged in the [data guidelines](https://www.notion.so/consensys/Event-Naming-Conventions-32b151ea6dac4495bef0928deae1d077?source=copy_link#4d96767df83f464f84851f3a7acc9c6d) if this can't be done with snake case props, please check with @Consensys/data-council team.

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -66,6 +85,16 @@ export const FeatureFlagOverrideProvider: React.FC<
     }));
   }, []);
 
+  const takeSnapshot = useCallback((featureFlagInfo: FeatureFlagInfo) => {
```
joaoloureirop: The use of `useCallback` here does more harm than good. Let's remove it.

joaoloureirop (changes requested):

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -66,6 +80,22 @@ export const FeatureFlagOverrideProvider: React.FC<
...
+      }));
+    } else if (featureFlagInfo.type === 'boolean with minimumVersion') {
+      setFeatureFlagSnapshots((prev) => ({
+        ...prev,
+        [featureFlagInfo.key]: (
+          featureFlagInfo.value as MinimumVersionFlagValue
+        ).enabled,
+      }));
+    }
+  }, []);
```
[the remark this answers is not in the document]
asalsys: we will only support tracking the related flags for  'boolean' and 'boolean with minimumVersion'. remote config values should be recorded

app/contexts/FeatureFlagOverrideContext.tsx
asalsys: 83:9  error  The 'takeSnapshot' function makes the dependencies of useCallback Hook (at line 219) change on every render. Move it inside the useCallback callback. Alternatively, wrap the definition of 'takeSnapshot' in its own useCallback() Hook  react-hooks/exhaustive-deps

when i remove it I get the following error

joaoloureirop: I see it was [added later on](https://github.com/MetaMask/metamask-mobile/pull/22023/commits/6d59f95ad9babc1a2e2c09fcd2691c4f9892c18c) to an useEffect dependency array

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -51,13 +53,25 @@ interface FeatureFlagOverrideProviderProps {
...
+    [key: string]: boolean;
+  }>({});
+
+  useEffect(() => {
+    if (Object.keys(featureFlagSnapshots).length > 0) {
+      addTraitsToUser({
+        ...featureFlagSnapshots,
+      });
+    }
+  }, [featureFlagSnapshots, addTraitsToUser]);
```
NicolasMassart: Q: why using a state and a useEffect instead of just calling addTraitsToUser in getFeatureFlag()? You would call it just twice, once for each flag type, but this would remove the need for state and effect.
Given the user profile traits are permanent in Segment, if you set them at the same time or incrementally, it's the same, eventually it's part of the profile.

app/contexts/FeatureFlagOverrideContext.test.tsx
```diff
@@ -55,6 +70,7 @@ describe('FeatureFlagOverrideContext', () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
+    mockAddTraitsToUser.mockClear();
```
NicolasMassart, line 73: C: you already have a `clearAllMocks()` call, no need for individual clearing IMO

NicolasMassart (approved):
Check the clear mock comment please, but otherwise LGTM.
Nice works, thanks for your care on reworking on this PR many times!

joaoloureirop (approved):

PR #22026, opened 2025-10-31
fix: Investigate perf regression in Perps 7.59

gambinish:
## **Description**

We noticed a performance regression on low end Android devices. It's unclear if this is a regression in 7.59, or if it was introduced earlier, and made more obvious in 7.59 with the latest controller changes and hip-3 support.

This PR proposes two performance improvements to the `PerpsStreamManager` and `HyperLiquidSubscriptionService` to address a potential memory leak caused by uncleared timers, and further optimizing our websocket subscription message processing.

1. `PerpsStreamManager`: The first change is to clear timers before we ever disconnect a stream connection. Without this we run the risk of exponentially maintaining orphaned timers that aren't closed before socket disconnection. This could be the reason for the memory creep that we were seeing, where performance on low end devices with limited memory would degrade over time.

2. `HyperLiquidSubscriptionService`: Based on what I could tell, we are currently subscribed to ALL market symbols (200-300+) on every websocket message, even when only 5-10 symbols actually need to be displayed or subscribed to. This should reduce the amount of data getting processed per websocket message by only subscribing to symbols that we care about.

## **Changelog**

CHANGELOG entry: Performance optimizations to `PerpsStreamManager` and `HyperLiquidSubscriptionService`

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### DIFF
--- a/app/components/UI/Perps/providers/PerpsStreamManager.tsx
+++ b/app/components/UI/Perps/providers/PerpsStreamManager.tsx
@@ -70,7 +70,6 @@ abstract class StreamChannel<T> {
       // Store pending update
       subscriber.pendingUpdate = updates;
 
-      // CRITICAL FIX: Clear existing timer before creating new one
       // This prevents timer accumulation and memory leaks, especially on Android devices
       if (subscriber.timer) {
         clearTimeout(subscriber.timer);
@@ -133,7 +132,6 @@ abstract class StreamChannel<T> {
   }
 
   public disconnect() {
-    // CRITICAL FIX: Clear all pending timers before disconnecting
     // This prevents orphaned timers from continuing to run after disconnect
     this.subscribers.forEach((subscriber) => {
       if (subscriber.timer) {
@@ -174,7 +172,6 @@ abstract class StreamChannel<T> {
   }
 
   public clearCache(): void {
-    // CRITICAL FIX: Clear all timers BEFORE disconnecting
     // This ensures no timers are orphaned during the disconnect/reconnect cycle
     this.subscribers.forEach((subscriber) => {
       // Clear any pending updates and timers
--- a/app/components/UI/Perps/providers/PerpsStreamManager.tsx
+++ b/app/components/UI/Perps/providers/PerpsStreamManager.tsx
@@ -70,16 +70,21 @@ abstract class StreamChannel<T> {
       // Store pending update
       subscriber.pendingUpdate = updates;
 
-      // Only set timer if one isn't already running
-      if (!subscriber.timer) {
-        subscriber.timer = setTimeout(() => {
-          if (subscriber.pendingUpdate) {
-            subscriber.callback(subscriber.pendingUpdate);
-            subscriber.pendingUpdate = undefined;
-            subscriber.timer = undefined;
-          }
-        }, subscriber.throttleMs);
+      // CRITICAL FIX: Clear existing timer before creating new one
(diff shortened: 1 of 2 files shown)


app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -70,16 +70,20 @@ abstract class StreamChannel<T> {
...
-        subscriber.timer = setTimeout(() => {
-          if (subscriber.pendingUpdate) {
-            subscriber.callback(subscriber.pendingUpdate);
-            subscriber.pendingUpdate = undefined;
-            subscriber.timer = undefined;
-          }
-        }, subscriber.throttleMs);
+      // This prevents timer accumulation and memory leaks, especially on Android devices
+      if (subscriber.timer) {
+        clearTimeout(subscriber.timer);
```
gambinish, line 75: Check if we have an existing timer, if we do, clear it before setting another timeout

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -126,6 +132,15 @@ abstract class StreamChannel<T> {
...
   public disconnect() {
+    // This prevents orphaned timers from continuing to run after disconnect
+    this.subscribers.forEach((subscriber) => {
+      if (subscriber.timer) {
+        clearTimeout(subscriber.timer);
+        subscriber.timer = undefined;
+      }
+      subscriber.pendingUpdate = undefined;
+    });
+
```
gambinish, line 143: This cleans up any orphaned timers after we disconnect

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -157,6 +172,16 @@ abstract class StreamChannel<T> {
...
+    // This ensures no timers are orphaned during the disconnect/reconnect cycle
+    this.subscribers.forEach((subscriber) => {
+      // Clear any pending updates and timers
+      if (subscriber.timer) {
+        clearTimeout(subscriber.timer);
+        subscriber.timer = undefined;
+      }
+      subscriber.pendingUpdate = undefined;
+    });
+
```
gambinish, line 184: Similarly here. When we clear the stream cache we likely also want to make sure we cleanup any orphaned timers as well

app/components/UI/Perps/services/HyperLiquidSubscriptionService.ts
```diff
@@ -1227,17 +1227,47 @@ export class HyperLiquidSubscriptionService {
         wsMetrics.messagesReceived++;
         wsMetrics.lastMessageTime = Date.now();
 
-        // Update cache for ALL available symbols
-        Object.entries(data.mids).forEach(([symbol, price]) => {
-          this.cachedPriceData ??= new Map<string, PriceUpdate>();
+        // Initialize cache if needed
+        this.cachedPriceData ??= new Map<string, PriceUpdate>();
```
gambinish, line 1231: It's unclear to me why we need to update the cache with all available symbols? Isn't that a bit excessive?

If you can help me understand why this is necessary, perhaps we don't need this change.

gambinish:
Closing this PR since it's been incorporated into a more comprehensive perf improvement here: https://github.com/MetaMask/metamask-mobile/pull/22038/files#r2483808961

PR #22100, opened 2025-11-04
feat(perps): use price updates to compute live positions

michalconsensys:
## **Description**

This PR implements real-time PnL (Profit and Loss) calculation for perpetual futures positions by enriching position data with live price updates.

**What is the reason for the change?**
Previously, `unrealizedPnl` and `returnOnEquity` values displayed in the UI were only updated when position data arrived via WebSocket (every few seconds). However, price updates arrive much more frequently, causing the displayed PnL to appear stale and not reflect real-time market movements.

**What is the improvement/solution?**
The `usePerpsLivePositions` hook now subscribes to both position updates and price updates. When prices change, it recalculates `unrealizedPnl` and `returnOnEquity` using the current mark price, providing users with live PnL updates that accurately reflect market movements.

Key technical improvements:
- Dual subscription system: positions stream (structural changes) + prices stream (live PnL)
- Performance optimization using `useRef` for intermediate state to prevent excessive re-renders
- Uses mark price (oracle-based) with fallback to mid price for accurate PnL calculation
- Exported `enrichPositionsWithLivePnL` utility function for potential reuse

## **Changelog**

CHANGELOG entry: Improved perps positions to display live profit/loss based on real-time price updates

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1946

## **Manual testing steps**

```gherkin
Feature: Live PnL updates for perps positions

  Scenario: user views position with live price updates
    Given user has an open perpetual futures position
    And the position tab is visible

    When market prices change rapidly
    Then the unrealized PnL updates in real-time without waiting for position stream refresh
    And the Return on Equity percentage reflects the live PnL calculation
```

### **Before**

https://github.com/user-attachments/assets/286925ce-95ef-47c5-a3fc-6cfc6e596e5e

### **After**

https://github.com/user-attachments/assets/89c48348-61f8-405a-a7cc-245fbc1f8645

---

> [!NOTE]
> Adds optional price-stream subscription to `usePerpsLivePositions` to compute live `unrealizedPnl` and `returnOnEquity`, exports utility, enables in Positions view, and updates tests/mocks.
> 
> - **Hooks**:
>   - Enhance `usePerpsLivePositions` to optionally subscribe to `stream.prices` (`useLivePnl` flag) and recompute `unrealizedPnl`/`returnOnEquity` on price updates.
>   - Add `enrichPositionsWithLivePnL(positions, priceData)` and export it from `hooks/stream/index`.
> - **UI**:
>   - `PerpsPositionsView`: use `usePerpsLivePositions` with `{ throttleMs: 1000, useLivePnl: true }` for live PnL display.
> - **Tests**:
>   - Add comprehensive tests for subscription lifecycle, throttling, and live PnL math (long/short, mark vs mid, fallbacks) in `useLivePositions.test.ts`.
>   - Update view tests/mocks to include `prices.subscribe` where needed (`PerpsMarketDetailsView.test.tsx`, `PerpsOrderView.test.tsx`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2678a2918c1542e7f5a2a55c202fe6ef871e782e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
```diff
@@ -12,21 +12,72 @@ export interface UsePerpsLivePositionsOptions {
...
+    ) {
+      return position;
+    }
+
+    // Calculate unrealized PnL: (markPrice - entryPrice) * size
+    const calculatedUnrealizedPnl = (markPrice - entryPrice) * size;
+
+    // Calculate ROE: (unrealizedPnl / marginUsed) as decimal (not percentage)
+    const calculatedRoe =
+      marginUsed > 0 ? calculatedUnrealizedPnl / marginUsed : 0;
```
michalconsensys, line 67: This needs to be doublechecked if the approach is correct. Otherwise I'm not sure how else we could reduce the latency as we're getting the pnl from a WS

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
```diff
@@ -12,21 +12,72 @@ export interface UsePerpsLivePositionsOptions {
...
+  return positions.map((position) => {
+    const priceUpdate = priceData[position.coin];
+    if (!priceUpdate) {
+      return position;
+    }
+
+    // Use mark price if available, fallback to mid price
+    const markPrice = priceUpdate.markPrice
+      ? Number.parseFloat(priceUpdate.markPrice)
+      : Number.parseFloat(priceUpdate.price);
```
Matt561, line 44: @abretonc7s - Are we comfortable falling back to mid price if mark price isn't available? My understanding is that under normal circumstances they're nearly equal but there are scenarios in which they can drift.

Imo, accuracy is more important that displaying the value. If we cannot get the price value we want we shouldn't display a potentially inaccurate one.

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
```diff
@@ -12,21 +12,72 @@ export interface UsePerpsLivePositionsOptions {
...
   positions: Position[];
   /** Whether we're waiting for the first real WebSocket data (not cached) */
   isInitialLoading: boolean;
 }
 
+/**
+ * Enrich positions with recalculated unrealizedPnl and returnOnEquity
+ * based on current mark price from live price feed
+ */
+export function enrichPositionsWithLivePnL(
```
Matt561, line 27: We may want to consider adding a `"skipLivePnl"` or `"useLivePnl"` option to the `UsePerpsLivePositionsOptions`.

The following don't seem to use the live pnl value and rerender on every price change now:

// Only check existence/leverage - PERFORMANCE ISSUE
1. usePerpsOrderForm (throttleMs: 0)
   - Line 64: Only needs existingPositionLeverage
   - Line 74-77: Just checks if position exists for asset
   - Re-renders on EVERY price update unnecessarily

2. useHasExistingPosition (no throttle, defaults to 0)
   - Line 38-43: Only needs to check if position exists
   - Re-renders on EVERY price update unnecessarily

3. usePerpsDataMonitor (no throttle, defaults to 0)  
   - Line 91-92: Only monitors structural changes (size, entryPrice)
   - Line 119-147: Checks if position size/entryPrice changed
   - Re-renders on EVERY price update unnecessarily

I suspect performance will degrade in one or more of these hooks.

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
Matt561: Do we include funding payments in the `calculatedRoe`? Just want to make sure if our previous `unrealizedPnl` did include them we should too since those can change the pnl over time.

Either way, I think it would be helpful to add a comment here to explain if the funding payments and/or any other fees are included in this unrealizedPnl. It's possible that when we received `unrealizedPnl` from Hyperliquid that it included these fees.

cc: @abretonc7s

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
```diff
@@ -57,27 +123,31 @@ export function usePerpsLivePositions(
...
+      unsubscribe();
+    };
+  }, [stream, throttleMs, updateEnrichedPositions]);
+
+  // Subscribe to price updates for real-time PnL recalculation
+  useEffect(() => {
+    const unsubscribe = stream.prices.subscribe({
+      callback: (newPriceData) => {
+        priceDataRef.current = newPriceData;
+        updateEnrichedPositions();
```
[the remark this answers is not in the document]
Matt561: @michalconsensys Is this legit?

Matt561 (dismissed):
I'm seeing the live updates when testing 👍 

I left a few comments/suggestions

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
michalconsensys: This is a good question, no fees are includes in my calculations, however the backend service we use may include these fees or use a different formula. Who is best to check these?

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
michalconsensys: Should we in this case only display markPrice?

abretonc7s: @Matt561 is right we should just keep the markPrice and skip the value otherwise (keeping undefined)

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
abretonc7s: Good call for the option, performance should remain stable as it re-use existing cache instead of creating new websocket subscriptions (all the data is already there we just do the mapping per position

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
```diff
@@ -79,6 +137,24 @@ export function usePerpsLivePositions(
...
+  useEffect(() => {
+    if (!useLivePnl) {
+      return undefined;
+    }
+
+    const unsubscribe = stream.prices.subscribe({
+      callback: (newPriceData) => {
+        setPriceData(newPriceData);
+      },
+      throttleMs,
```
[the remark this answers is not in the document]
abretonc7s: This is fine

app/components/UI/Perps/hooks/stream/usePerpsLivePositions.ts
abretonc7s: the roe shouldnt include all fees and come from hyperliquid directly we dont need to add them manually

abretonc7s: The best way to QA is to compare the value of your position vs what is displayed on hyperliquid

gambinish (approved):

PR #22201, opened 2025-11-05
fix: Deduplicate y-axis values in TradingView chart

gambinish:
## **Description**

We noticed that during certain candle intervals (eg SOL-USD) that there wasn't enough distinction between the y-axis intervals to register as a significant step increment in the chart.

For assets like SOL-USD trading around $160, the chart would display multiple candles with slightly different prices within a small range (e.g., 160.12, 160.34, 160.56, 160.78). When these prices were formatted with only 1 decimal place, they would all round to the same values

The bug was in the `priceFormatter` function, where we weren't being smart enough about calculating the visible range.

Instead of using a fixed number of decimals based solely on the price magnitude, this fix implements dynamic precision that adapts to the visible price range:

- Min price in visible range
- Max price in visible range
- Span (max - min)

Then, based on this span, we can dynamically set the target decimals. It's important to handle this in the webview itself, since the values are dynamic as the user navigates the chart, rather than relying on post messages to pass props from React into the WebView.

## **Changelog**

CHANGELOG entry: Dynamically calculate y-axis precision in Perps chart

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-1919

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **Before**

Screenshot 2025-11-05 at 8 32 58 AM

### **After**

https://github.com/user-attachments/assets/fa0fcea1-0c66-4982-87e5-28bb7462e8b4

---

> [!NOTE]
> Adjusts Y-axis price label precision dynamically from the visible price range, updating on zoom and initial data load to prevent duplicate labels.
> 
> - **Perps TradingView chart**:
>   - **Dynamic Y-axis precision**: Replaces magnitude-based `priceFormatter` with range-aware formatting using `window.visiblePriceRange`; adds rounding to avoid FP artifacts and 4-sig-fig handling for values < 1.
>   - **State & utilities**: Introduces `window.visiblePriceRange` and new `window.updateVisiblePriceRange()` to compute min/max/span from visible candles.
>   - **Lifecycle hooks**: Calls `updateVisiblePriceRange()` after initial data load and on `applyZoom()` to keep labels consistent during navigation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fede60cb3dab7d9c1e0cc76e28e759102628583f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/components/UI/Perps/components/TradingViewChart/TradingViewChart.html
+++ b/app/components/UI/Perps/components/TradingViewChart/TradingViewChart.html
@@ -107,6 +107,7 @@ PERFORMANCE OPTIMIZATIONS:
         window.lastDataKey = null; // Track the last dataset to avoid unnecessary autoscaling
         window.visibleCandleCount = 45; // Default visible candle count
         window.allCandleData = []; // Store all loaded data for zoom functionality
+        window.visiblePriceRange = null; // Track visible price range for dynamic decimal precision
         
         // Zoom limits - consistent with chart configuration
         window.ZOOM_LIMITS = {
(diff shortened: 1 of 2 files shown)


michalconsensys (approved):
I've tested it on a couple of assets and it seems to work well

PR #22208, opened 2025-11-05
fix: Updated tabsbar to update when font size pref changes

brianacnguyen:
## **Description**

Users were reporting that the TabsBar component in the Wallet view was not horizontally scrollable when tabs overflowed, particularly when device font size settings were increased.

**Root causes identified:**

1. **Overflow detection didn't account for container padding** - The component has `px-4` (32px total) padding, but the overflow calculation compared total content width against the full container width, not the available space minus padding.

2. **Underline positioning race condition** - With font scaling, layout measurements would race with font rendering, causing the underline to have inconsistent width/position on initial load.

3. **No dynamic recalculation** - When font size changed after initial load, the overflow detection would not recalculate, leaving tabs non-scrollable even when they overflowed.

**Solution:**

- Fixed overflow detection to account for container's 32px horizontal padding in both calculation points
- Added layout change detection to re-animate underline when tabs resize due to font scaling
- Implemented dynamic recalculation of scroll state when any tab's layout changes significantly (>1px)
- Used refs to avoid circular dependencies and performance issues

## **Changelog**

CHANGELOG entry: Fixed TabsBar not being horizontally scrollable when tabs overflow, especially with increased font size settings

## **Related issues**

Fixes: 

## **Manual testing steps**

```gherkin
Feature: TabsBar horizontal scrolling with increased font size

  Scenario: user increases device font size with many tabs
    Given user has multiple tabs visible in the Wallet view (Tokens, Perps, Predictions, NFTs)
    And tabs fit within the screen width at default font size

    When user increases device font size in Settings > Display & Brightness > Text Size
    Then tabs should become horizontally scrollable
    And the underline indicator should correctly match the width of active tab
    And user should be able to swipe horizontally to see all tabs

  Scenario: user views tabs with default font size
    Given user has multiple tabs in the Wallet view
    And device font size is at default setting
    And tabs do not overflow the screen width

    When user views the Wallet screen
    Then tabs should not be scrollable
    And all tabs should be visible without scrolling
    And the underline should correctly align with active tab

  Scenario: user switches between tabs after font size increase
    Given user has increased device font size
    And tabs are horizontally scrollable due to overflow

    When user taps on a different tab or swipes to navigate
    Then the underline should animate smoothly to the new tab
    And the ScrollView should automatically scroll to show the active tab
    And the underline width should match the tab width accurately
```

### **Before**
https://github.com/user-attachments/assets/3359232b-c54a-4d6b-9124-c6e194de7fc1

### **After**
https://github.com/user-attachments/assets/b83a5cc9-9232-4fb1-bbde-3e73d6c75332

---

> [!NOTE]
> Improves TabsBar by accounting for container padding in overflow detection and re-animating the underline when tab layouts change (e.g., font scaling).
> 
> - **TabsBar (`app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx`)**
>   - **Overflow detection**: Accounts for container padding (`-32px`) when computing `shouldScroll`.
>   - **Dynamic layout handling**:
>     - Detects significant tab layout changes (>1px) and recalculates scroll state.
>     - Re-animates active tab underline via `requestAnimationFrame`; added `activeIndexRef` to sync with `activeIndex`.
>   - **Stability/cleanup**: Cancels in-flight animations and pending RAF callbacks on updates/unmount.
>   - **Deps**: Updates effect/callback deps to include `animateToTab`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f60d9618cd7c29b3dbe6514067702df98aa72576. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx
```diff
@@ -176,7 +183,8 @@ const TabsBar: React.FC<TabsBarProps> = ({
         const gapsWidth = (tabs.length - 1) * 24; // Account for gaps between tabs
         const calculatedContentWidth = totalTabsWidth + gapsWidth;
 
-        const shouldScroll = calculatedContentWidth > containerWidth;
+        // Account for container's px-4 padding (16px * 2 = 32px)
+        const shouldScroll = calculatedContentWidth > containerWidth - 32;
```
georgewrmarshall, line 187: non-blocking: There’s some maintenance overhead here. If the container’s padding changes, we’ll need to update this magic number. Could we instead move the horizontal padding to the `ScrollView`, and to the wrapping `Box` when `shouldScroll` is false?

georgewrmarshall (approved):
Nice work on the fix here! Tested on iOS and it works as expected. Approving to unblock ✅

I do have a question regarding the architecture of this component though! I’m curious as to why are we conditionally rendering the `ScrollView` instead of always rendering it with `scrollEnabled` and using the `onContentSizeChange` handler? Is there a specific limitation that prevents that approach? Just wondering, as it seems like it could help reduce some of the code duplication.

PR #22522, opened 2025-11-12
chore: broadcast staking TRX transaction cp-7.61.3

zone-live:
## **Description**

Staking and unstaking methods connected to the tron-snap:

https://github.com/user-attachments/assets/a7f66644-1a44-4035-9e06-64c2ff71322f

Screenshot 2025-11-19 at 17 54 46
Example: https://tronscan.org/address/TGXFnQBLAdbdkupHUGSpeBfbxB72hkMsh2#/transaction/85bfd85b620e6a8fc5966458cbfcf3a9757454917f78c36b675e223b68564499

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> Integrates TRON staking and unstaking via Snap, updating Earn flows, lists, and selectors to support TRX with new validation, previews, and navigation.
> 
> - **TRON staking/unstaking integration**:
>   - Add `useTronStake` and `useTronUnstake` hooks for validation/confirmation via `tron-staking-snap`.
>   - New utils: `tron-staking-snap` (Snap RPCs), `tron` (navigation/result handling, token builder, staked total).
>   - Add `TronStakePreview` UI for fee preview; reuse `ResourceToggle`.
> - **Earn Input/Withdraw flows** (`EarnInputView.tsx`, `EarnWithdrawInputView.tsx`):
>   - Wire TRX validation on keypad input; confirm flows call Snap and navigate to success/error sheets.
>   - Show Tron-specific UI (resource toggle, fee preview) and simplified button labels; adjust disabled/loading states.
>   - Limit gas-cost warning to `isETH` only; debounce validation handler.
> - **Token list and selection**:
>   - `EarnTokenList`: include TRX native in deposit list (even with zero balance when enabled); navigate without EVM network switch; sorting updated.
>   - `EarnTokenSelector`: render output token for withdraw; earn token for stake.
> - **Selectors and data plumbing**:
>   - Support non-EVM (TRX) balances/tokens in earn selectors; default TRX pooled-staking APR to `0`.
>   - Add unified multichain token selector including non-EVM; filter Tron resource/testnet assets; use `getDecimalChainId`.
> - **Utilities/constants/i18n**:
>   - Add `normalizeToDotDecimal`; `TronResourceType` constant; new TRON strings (fee, success/error copy).
> - **Tests**: Extensive new/updated tests for views, hooks, lists, buttons, selectors, and utils; snapshot updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0989d7f70663d0b76a07eedff09830cae2c0f9d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/selectors/multichain/multichain.ts
```diff
@@ -336,6 +342,82 @@ export const selectMultichainTokenListForAccountsAnyChain =
       return tokens;
     },
   );
+
+/**
+ * Unified selector: EVM tokens (native + ERC20) for the selected EVM address
+ * plus non-EVM tokens (e.g., TRX) across all accounts in the selected account group.
+ * Returns a map keyed by chainId (hex for EVM, CAIP-2 for non-EVM) to TokenI[].
+ */
+export const selectAccountTokensAcrossChainsUnified = createDeepEqualSelector(
```
Matt561, line 351: Let's ask the assets team to give this a look. 

cc: @Prithpal-Sooriya

app/components/UI/Earn/components/EarnTokenSelector/index.tsx
```diff
@@ -46,10 +46,13 @@ const EarnTokenSelector = ({
   const earnToken = getEarnToken(someEarnToken);
   const outputToken = getOutputToken(someEarnToken);
 
-  const tokenToRender = (earnToken || outputToken || someEarnToken) as
-    | EarnTokenDetails
-    | TokenI
-    | undefined;
+  // For withdrawal flows we want to display the asset the user is
+  // actually interacting with (e.g. sTRX), not a paired underlying.
```
Matt561: Are TRX and sTRX not a pair in the context of Tron staking?

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -128,9 +142,17 @@ const EarnInputView = () => {
   const { getEarnToken } = useEarnTokens();
 
   ///: BEGIN:ONLY_INCLUDE_IF(tron)
-  const [resourceType, setResourceType] = useState<ResourceType>('energy');
-  const isTronNative =
-    token.ticker === 'TRX' && String(token.chainId).startsWith('tron:');
+  const [resourceType, setResourceType] = useState<ResourceType>(
+    TRON_RESOURCE.ENERGY,
+  );
+  const isTronNative = token.isNative && isTronChainId(String(token.chainId));
```
Matt561: Can this state and logic live within `useTronStake` or another hook? Ideally we want to encapsulate as much of the Tron logic as possible.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -139,18 +161,13 @@ const EarnInputView = () => {
...
-        ...token,
-        isETH: false,
-        balanceMinimalUnit: '0',
-        balanceFormatted: token.balance ?? '0',
-        balanceFiat: token.balanceFiat ?? '0',
-        tokenUsdExchangeRate: 0,
-        experiences,
-        experience: experiences[0],
-      } as EarnTokenDetails;
+    const tronEarnToken = buildTronEarnTokenIfEligible(token, {
```
Matt561: Can this logic live within the existing `useEarnTokens` if necessary? I see the `selectEarnTokens` has been updated to include TRX. 

The high level Earn architecture is:
EarnController (part of core) -> selectors (e.g. `selectEarnTokens`) -> reusable hooks (e.g. `useEarnTokens`).


cc: @nickewansmith What are your thoughts?

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -598,6 +616,18 @@ const EarnInputView = () => {
...
+    ///: BEGIN:ONLY_INCLUDE_IF(tron)
+    if (isTrxStakingEnabled && isTronNative) {
+      const result = await tronConfirm?.(
+        amountToken,
+        resourceType as TronResourceType,
+        String(token.chainId),
+      );
+      handleTronStakingNavigationResult(navigation, result, 'stake');
+      return;
+    }
```
Matt561: In the `selectEarnTokensSelector` Tron is assigned the `POOLED_STAKING` experience but this conflicts with handling in this function.

I think in the future it will make more sense to create bespoke handlers (e.g. `handleEthPooledStaking`, `handleTronStaking`) and have `handleLendingFlow` and `handleStakingFlow` to serve as routers to these lower level functions.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -716,7 +753,10 @@ const EarnInputView = () => {
   };
 
   const balanceText = strings('stake.balance');
-  const buttonLabel = getButtonLabel();
+  const buttonLabel =
+    isTrxStakingEnabled && isTronNative
+      ? strings('stake.stake')
+      : getButtonLabel();
```
Matt561: Would it be better to have this logic encapsulated in `getButtonLabel`?

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -922,7 +973,14 @@ const EarnInputView = () => {
...
         // Debounce used to avoid error message flicker from recalculating gas fee estimate
-        onChange={debounce(handleKeypadChange, 1)}
+        onChange={debounce((data) => {
+          handleKeypadChange(data);
+          ///: BEGIN:ONLY_INCLUDE_IF(tron)
+          if (isTrxStakingEnabled && isTronNative && !isFiat) {
+            tronValidate?.(String(data.value), String(token.chainId));
+          }
+          ///: END:ONLY_INCLUDE_IF
+        }, 1)}
```
Matt561: Can we break this out into a dedicated event handler?

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -81,39 +96,55 @@ const EarnWithdrawInputView = () => {
...
+  const tronResources = useSelector(selectTronResourcesBySelectedAccountGroup);
+  const stakedTrxTotal = useMemo(
+    () =>
+      isTrxStakingEnabled && isTronAsset
+        ? getStakedTrxTotalFromResources(tronResources)
+        : 0,
+    [isTrxStakingEnabled, isTronAsset, tronResources],
+  );
+  const { validate: tronValidateUnstake, confirm: tronConfirmUnstake } =
+    useTronUnstake();
```
Matt561: Can any of this state and/or business logic be moved into `useTronUnstake` for better encapsulation?

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -81,39 +96,55 @@ const EarnWithdrawInputView = () => {
...
-        ...token,
-        isETH: false,
-        balanceMinimalUnit: '0',
-        balanceFormatted: token.balance ?? '0',
-        balanceFiat: token.balanceFiat ?? '0',
-        tokenUsdExchangeRate: 0,
-        experiences,
-        experience: experiences[0],
-      } as EarnTokenDetails;
+    const tronEarnToken = buildTronEarnTokenIfEligible(token, {
```
Matt561: Can this be migrated to `useEarnTokens` or `useEarnTokenMetadata` if this is enriching the Tron token? I'm mostly wanting keep token-specific logic out of this component if feasible.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -81,39 +96,55 @@ const EarnWithdrawInputView = () => {
...
+    stakedTrxTotal,
     token,
     ///: END:ONLY_INCLUDE_IF
   ]);
 
-  const receiptTokenToUse: EarnTokenDetails | undefined =
-    receiptToken as EarnTokenDetails;
+  const receiptTokenToUse: EarnTokenDetails | undefined = receiptToken
+    ? (receiptToken as EarnTokenDetails)
+    : (earnToken as EarnTokenDetails);
```
Matt561: The `earnToken` should never be considered a receipt token. The Earn token pairs are made of the `earnToken` and `receiptToken` (or `outputToken`).

Example:
- ETH: staked ETH (pooled-staking)
- USDC: aUSDC (lending)

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -544,6 +575,17 @@ const EarnWithdrawInputView = () => {
...
+    ///: BEGIN:ONLY_INCLUDE_IF(tron)
+    if (isTrxStakingEnabled && isTronAsset) {
+      const result = await tronConfirmUnstake?.(
+        amountToken,
+        resourceType as TronResourceType,
+        String(token.chainId),
+      );
+      handleTronStakingNavigationResult(navigation, result, 'unstake');
+      return;
+    }
```
Matt561: It reads strangely that TRX is assigned the `POOLED_STAKING` experience _but_ follows its own dedicated logic path. Ideally we'd handle this within the pooled-staking code.

This may require general restructuring of the earn code and may not be feasible right now given time constraints.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -613,8 +664,13 @@ const EarnWithdrawInputView = () => {
       return strings('stake.enter_amount');
     }
     if (isOverMaximum.isOverMaximumToken) {
+      const tickerForError =
+        isTrxStakingEnabled && isTronAsset
+          ? (token.ticker ?? token.symbol ?? '')
+          : (withdrawalToken?.ticker ?? withdrawalToken?.symbol ?? '');
```
Matt561: Is it feasible to use the existing code instead of forcing a separate path specifically for Tron? If not it's okay and we will refactor later.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -625,14 +681,20 @@ const EarnWithdrawInputView = () => {
       return strings('earn.amount_exceeds_safe_withdrawal_limit');
     }
 
-    return strings('stake.review');
+    return isTrxStakingEnabled && isTronAsset
+      ? strings('stake.unstake')
+      : strings('stake.review');
```
Matt561: It seems likely that we'd want "review" for all experiences. Is this being used as a placeholder until Tron has a dedicated confirmation screen?

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -774,7 +843,7 @@ const EarnWithdrawInputView = () => {
           <View style={styles.earnTokenSelectorContainer}>
             <View style={styles.spacer} />
             <EarnTokenSelector
-              token={withdrawalToken as TokenI}
+              token={token as TokenI}
```
Matt561, line 826: I'm curious to see if this causes bugs with pooled-staking and/or lending.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -786,7 +855,18 @@ const EarnWithdrawInputView = () => {
...
+          ///: BEGIN:ONLY_INCLUDE_IF(tron)
+          if (isTrxStakingEnabled && isTronAsset && !isFiat) {
+            tronValidateUnstake?.(
+              String((data as { value: string }).value),
+              resourceType as TronResourceType,
+              String(token.chainId),
+            );
+          }
+          ///: END:ONLY_INCLUDE_IF
+        }}
```
Matt561: Let's break this into a dedicated event handler.

app/components/UI/Earn/components/EarnTokenList/index.tsx
```diff
@@ -223,14 +247,20 @@ const EarnTokenList = () => {
...
       const hasTokenBalance = new BN4(token.balanceMinimalUnit).gt(new BN4(0));
-      // show at least ETH if no other tokens have balance
-      if (hasTokenBalance || token.isETH) {
+      // show ETH always (existing behavior). Also show TRX native when TRX staking is enabled.
+      const isTrxNative =
+        Boolean(token.isNative) && isTronChainId(String(token.chainId));
+      if (
+        hasTokenBalance ||
+        token.isETH ||
+        (isTrxNative && isTrxStakingEnabled)
```
Matt561, line 256: Do we want to display Tron staking in the token list if the user has a zero balance?

app/components/UI/Earn/components/Tron/StakePreview/TronStakePreview.tsx
```diff
@@ -0,0 +1,55 @@
...
+    justifyContent={BoxJustifyContent.Between}
+    twClassName="mt-1"
+  >
+    <Text variant={TextVariant.BodyMd}>{label}</Text>
+    <Box>{typeof value === 'string' ? <Text>{value}</Text> : value}</Box>
+  </Box>
+);
+
+const TronStakePreview = ({
+  // resourceType = TRON_RESOURCE.ENERGY,
```
Matt561: Let's remove this if it's not being used.

app/components/UI/Earn/components/Tron/StakePreview/TronStakePreview.tsx
```diff
@@ -0,0 +1,55 @@
...
+    ? fee[0]
+    : fee;
+
+  return (
+    <Box
+      twClassName="w-full bg-default pl-6 pr-6 rounded-lg"
+      style={tw.style()}
+    >
+      <Row
+        label={strings('stake.tron.fee')}
```
Matt561: Not needed for this PR but it would be nice to keep the i18n strings under the earn namespace.

app/components/UI/Earn/hooks/useTronStake.ts
```diff
@@ -0,0 +1,118 @@
...
+      const { errors: validationErrors, ...rest } = validation ?? {};
+      let nextPreview: Record<string, unknown> | undefined =
+        rest && Object.keys(rest).length > 0 ? rest : undefined;
+
+      try {
+        // const fee = await computeTronFee(selectedTronAccount, {
+        //   transaction: 'stake',
+        //   accountId: selectedTronAccount.id,
+        //   scope: chainId,
+        // });
```
Matt561: Let's remove if this isn't needed.

app/components/UI/Earn/hooks/useTronStake.ts
```diff
@@ -0,0 +1,118 @@
...
+          asset: {
+            unit: 'TRX',
+            type: 'TRX',
+            amount: '0.01',
+            fungible: true,
+          },
+        };
+        nextPreview = { ...(nextPreview ?? {}), fee };
+      } catch {
+        // ignore for now
```
Matt561, line 109: Do we want this to fail silently? Will this be handled later?

app/components/UI/Earn/utils/tron-staking.ts
```diff
@@ -0,0 +1,143 @@
...
+import { TronResourceType } from '../../../../core/Multichain/constants';
+
+const controllerMessenger = Engine.controllerMessenger;
+
+export const TRON_WALLET_SNAP_ID: SnapId =
+  'npm:@metamask/tron-wallet-snap' as SnapId;
+
+export const TRON_WALLET_NAME: string = 'Tron';
+
+export interface TronStakeValidateParams {
```
Matt561: Might make more sense for the types to be in a separate types file and for this to remain a utils file.

app/components/UI/Earn/utils/tron-staking-snap.ts
```diff
@@ -0,0 +1,143 @@
+import { HandlerType } from '@metamask/snaps-utils';
```
Matt561, line 1: Would it make more sense to rename this file to represent that it's interfacing with the snap?

app/components/UI/Stake/components/StakeButton/index.tsx
```diff
@@ -205,6 +205,12 @@ const StakeButtonContent = ({ asset }: StakeButtonProps) => {
     if (primaryExperienceType === EARN_EXPERIENCES.STABLECOIN_LENDING) {
       return handleLendingRedirect();
     }
+    ///: BEGIN:ONLY_INCLUDE_IF(tron)
+    // Fallback for TRX (stake flag on) when no earnToken metadata is present yet
+    if (isTronNative && isTrxStakingEnabled) {
+      return handleStakeRedirect();
+    }
```
Matt561: TRX is assigned the `POOLED_STAKING` experience. Would it make more sense to handle this code path inside of `handleStakeRedirect`?

app/selectors/earnController/earn/index.ts
```diff
@@ -214,15 +213,25 @@ const selectEarnTokens = createDeepEqualSelector(
...
-        : hexToBN(
-            tokenBalances?.[selectedAddress as Hex]?.[token.chainId as Hex]?.[
-              token.address as Hex
-            ],
-          );
+      const tokenBalanceMinimalUnit = (() => {
+        if (token.isETH) {
+          return token.isStaked ? stakedBalanceWei : balanceWei;
+        }
+        if (isTronNative) {
```
Matt561: Minor: Do we have a way to detect non-EVM vs EVM tokens easily? Might be better for the check here to better support other non-EVM tokens.

app/selectors/earnController/earn/index.ts
```diff
@@ -13,14 +13,19 @@ import { TokenI } from '../../../components/UI/Tokens/types';
 import { deriveBalanceFromAssetMarketDetails } from '../../../components/UI/Tokens/util/deriveBalanceFromAssetMarketDetails';
```
Matt561, line 13: @nickewansmith Curious to get your thoughts on the changes in this selector.

app/selectors/earnController/earn/index.ts
```diff
@@ -246,17 +255,28 @@ const selectEarnTokens = createDeepEqualSelector(
...
-          token,
-          tokenExchangeRates,
-          tokenBalances?.[selectedAddress as Hex]?.[token.chainId as Hex] || {},
-          ethToUserSelectedFiatConversionRate,
-          currentCurrency || '',
-        );
-
-      let assetBalanceFiatNumber = balanceFiatNumber;
-      if (!token.isETH) {
+      const tronOrDerived = isTronNative
```
Matt561: Can we update this to avoid the ternary expression and use regular if statements? I think this will quickly get crazy once more staking tokens are added.

app/selectors/earnController/earn/index.ts
```diff
@@ -271,11 +291,15 @@ const selectEarnTokens = createDeepEqualSelector(
       // TODO: we could add direct validator staking as an additional earn experience
 
       if (isStakingToken || isStakingOutputToken) {
+        const aprForExperience = token.isETH
+          ? pooledStakingVaultAprForChain
+          : // TODO: Comeback after we have a TRX APR value
+            '0';
```
Matt561: A helper here to get the appropriate APR based on the token would be useful to avoid this type of condition where if it _isn't_ ETH it's assumed to be Tron. I can see these being difficult to spot when we need to debug or add new tokens.

app/components/UI/Earn/components/EarnTokenList/index.tsx
```diff
@@ -159,6 +162,27 @@ const EarnTokenList = () => {
   const handleRedirectToInputScreen = async (token: TokenI) => {
     const { NetworkController } = Engine.context;
 
+    // For non-EVM do not try to switch EVM network and just navigate directly
```
nickewansmith: Given that Tron Staking is an Earn product we may want to preserve the events and traces in the flow if possible to give us visibility around Tron Staking usage.

Separately if switching the network is the only part of the code that is to be skipped, maybe instead can have a util method earnNetworkSwitch or similar that will switch networks if the token is not a Tron token. In this way we can keep the events and traces and also have less code and less code duplication.

app/components/UI/Earn/components/EarnTokenSelector/index.tsx
nickewansmith: We should try to make it such that TRX is seen as the earnToken and STRX is the outputToken. Then there is no need to change this logic with Tron being the core reason.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
nickewansmith: Agreed, if there is Tron logic already in the selectEarnTokens selector, then that is great as this can be handled there such that Tron is already an earn token.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
```diff
@@ -889,31 +929,42 @@ const EarnInputView = () => {
           currencyToggleValue={currencyToggleValue}
         />
         <View style={styles.rewardsRateContainer}>
-          {isStablecoinLendingEnabled && !isTrxStakingEnabled ? (
-            <>
-              <View style={styles.spacer} />
-              <EarnTokenSelector
-                token={token}
-                action={EARN_INPUT_VIEW_ACTIONS.DEPOSIT}
+          {!isTrxStakingEnabled &&
```
nickewansmith: If Tron is enabled, do we want to show the EstimatedAnnualRewardsCard? I don't think this would work for Earn tokens if Earn lending was enabled.

app/selectors/earnController/earn/index.ts
```diff
@@ -180,21 +186,14 @@ const selectEarnTokens = createDeepEqualSelector(
...
-        token.isNative &&
-        token.chainId?.startsWith('tron:') &&
-        token.ticker === 'TRX';
+        token.isNative && isTronChainId(token.chainId as Hex);
 
       const isStakingToken =
         (token.isETH && !token.isStaked && isStakingSupportedChain) ||
-        (isTrxStakingEnabled && isTronNative && !token.isStaked);
+        (isTronNative && isTrxStakingEnabled && !token.isStaked);
       const isStakingOutputToken =
```
nickewansmith, line 224: Curious why we wouldn't add sTRX as the staking output token here? Can we assume that isTronNative and isStaked means that the token is sTRX?

app/components/UI/Earn/components/EarnTokenSelector/index.tsx
zone-live: No it isn't, TRX and sTRX are not a pair in the traditional sense, sTRX is a derivative token that just represents staked TRX. I'll check if it possible to do this association (TRX to be seen as the earnToken and STRX as the outputToken) 👍🏼

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
zone-live: Will do 👍🏼

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
zone-live: Makes sense, you're 100% right 👍🏼

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
zone-live: Makes sense, it would be more explicit, I don't mind doing that separation as a followup task as well.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
zone-live: True, it would be cleaner, will do.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
zone-live: Yap!

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: Will make it happen! 🫡

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: Since `useTronUnstake` already has `stakedTrxTotal`, I'll extend it and do it there. 👍🏼

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: True, I'll change that. Will make it clearer by using something like a `tronWithdrawalToken` so that it represents beter the staked position and acts as the receipt token.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: Yap, lets address it with a separate ticket like the one in the EarnInputView. 👍🏼

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: I think it can be simplified... The only tron specific thing is the final button label.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: It's just not something to worry for this task, since we have an other specific to handle the dedicated review/confirmation afterwards.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: I think this is fine because the whole withdraw flow is accessed from the staked position.

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: Will do.

app/components/UI/Earn/components/EarnTokenList/index.tsx
zone-live: We want it to do the same as ETH right now. And since for ETH it will show, for TRX it will as well. We can also revisit this block if later we decide we shouldn't show any token there with a zero balance.

app/components/UI/Earn/hooks/useTronStake.ts
zone-live: This will be added in a followup task, because right now we don't have the correct snap method to get the fees for the staking flow. But yes, I'll remove it since its not needed.

app/components/UI/Earn/hooks/useTronStake.ts
zone-live: Not relevant for now, we'll come back to this fee method later in a next task,

app/components/UI/Earn/utils/tron-staking.ts
zone-live: Will do.

app/components/UI/Earn/utils/tron-staking-snap.ts
zone-live: Will do!

app/components/UI/Stake/components/StakeButton/index.tsx
zone-live: I think it can be removed since the `handleStakeRedirect` already handles TRX case.

app/selectors/earnController/earn/index.ts
zone-live: True, we can use the `isNonEvmChainId` from multichain utils. It will check the chainid agains [this](https://github.com/MetaMask/core/blob/main/packages/multichain-network-controller/src/constants.ts#L32).

app/selectors/earnController/earn/index.ts
zone-live: I've updated logic here and this is not the case anymore. 👍🏼

app/selectors/earnController/earn/index.ts
zone-live: Will add 💪🏼

app/components/UI/Earn/components/EarnTokenList/index.tsx
zone-live: Fair point, and good idea. I'll create that method 💪🏼

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
zone-live: Not sure honestly, need to confirm, but I'll revisit this in a next ticket 👍🏼

app/selectors/earnController/earn/index.ts
zone-live: Yap, I've added it to the isStakingOutputToken. 👍🏼  And its based on:

`const isTronStakedToken =
        token.isStaked &&
        isTronChainId(token.chainId as Hex) && isTrxStakingEnabled;`

app/components/UI/Earn/components/EarnTokenSelector/index.tsx
zone-live: TRX is already in `earnTokensByChainIdAndAddress`, so I added sTRX  into earnOutputTokensByChainIdAndAddress. This way one is returned by `getEarnToken(TRX)` and the other `getOutputToken(sTRX)`.

app/selectors/multichain/multichain.ts
zone-live: Will ask for his input.

app/components/UI/Earn/Views/EarnInputView/EarnInputView.tsx
nickewansmith: I have covered this in another PR don't worry about it

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
```diff
@@ -81,40 +80,41 @@ const EarnWithdrawInputView = () => {
...
+    validating: isTronUnstakeValidating,
+    validate: tronValidateUnstake,
+    confirmUnstake: tronConfirmUnstake,
+  } = useTronUnstake({ token });
   ///: END:ONLY_INCLUDE_IF
 
-  const earnToken = React.useMemo(() => {
-    if (earnTokenFromMap) return earnTokenFromMap;
+  // Flag to conditionally show Tron-specific UI
+  let showTronUnstakingUI = false;
```
Matt561, line 95: How come this separate "showTronUnstakingUI" for the stake and unstake views is needed? Are we not able to rely directly on the `isTronEnabled` variable? 

Apologies if I'm missing something!

app/components/UI/Earn/components/EarnTokenList/index.tsx
```diff
@@ -156,27 +162,45 @@ const EarnTokenList = () => {
     });
   };
 
-  const handleRedirectToInputScreen = async (token: TokenI) => {
-    const { NetworkController } = Engine.context;
+  /**
+   * Prepares the network for the token if needed.
+   * EVM tokens require switching to the correct network.
+   * Non-EVM tokens (TRX, SOL, BTC) don't require network switching.
```
Matt561, line 168: Good to know! 👍. I wonder if we have a generic util for this functionality already? All good if we don't.

app/components/UI/Earn/hooks/useTronStake.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  const [validating, setValidating] = useState(false);
+  const [errors, setErrors] = useState<string[] | undefined>(undefined);
+  const [preview, setPreview] = useState<Record<string, unknown> | undefined>(
+    undefined,
+  );
+
+  const chainId = String(token.chainId);
+
+  const validate = useCallback<UseTronStakeReturn['validate']>(
```
Matt561, line 78: Nit: validate is ambiguous. Maybe something like `validateStakeInput`, `validateStakeAmount`, etc. would be more intuitive.

Non-blocking though.

app/components/UI/Earn/types/tron-staking.types.ts
```diff
@@ -0,0 +1,62 @@
...
+  assetId: CaipAssetType;
+  options: { purpose: TronResourceType.ENERGY | TronResourceType.BANDWIDTH };
+}
+
+export interface TronUnstakeConfirmParams {
+  value: string;
+  accountId: string;
+  assetId: CaipAssetType;
+  options: { purpose: TronResourceType.ENERGY | TronResourceType.BANDWIDTH };
+}
```
Matt561, line 39: These types appear identical. Do we need 2 separate types for this unstaking context?

I also get avoiding coupling if they're used for totally different things!

Non-blocking.

Matt561 (approved):
Leaving some minor comments but they're non-blocking. 👍

nickewansmith (approved):

app/components/UI/Earn/Views/EarnWithdrawInputView/EarnWithdrawInputView.tsx
zone-live: Its because of the code fences, we can't use `isTronEnabled` directly because it doesn't exist as a variable in non-tron builds (outside the fence). The `showTronUnstakingUI` variable ensures the code compiles and runs correctly in both cases, with and without the tron fence. So `showTronUnstakingUI` always exists, just assumes different values depending if the tron fence exists or not. I'll add a more explicit comment there in my next PR 💪🏼

app/components/UI/Earn/components/EarnTokenList/index.tsx
zone-live: I'll check and if we do, I'll add it in the next PR as well.

app/components/UI/Earn/hooks/useTronStake.ts
zone-live: Noted, I'll add it in next.

app/components/UI/Earn/types/tron-staking.types.ts
zone-live: Yeah, It was just for semantic "clarity" 😄 I can merge them with a type alias, no worries.

PR #23487, opened 2025-12-01
feat: import new remote feature flag controller with override functionality

asalsys:
## **Description**

This PR adds comprehensive override functionality to the `RemoteFeatureFlagController` that was added iby bumping remote-feature-flag-controller to version 3.1.0, allowing developers and users to locally override remote feature flags for testing and debugging purposes. The implementation also includes enhanced A/B test visibility features that provide access to available test groups.

**What is the reason for the change?**
The need to override remote feature flags locally for development, testing, and debugging scenarios without modifying the remote configuration.

**What is the improvement/solution?**
- Added local override system that takes precedence over remote flags
- Enhanced state management to track both remote flags and local overrides separately
- Added A/B test raw flag storage to provide visibility into available test groups
- Implemented comprehensive API methods for managing overrides and accessing flag information

## **Changelog**

CHANGELOG entry: Added override functionality to remote feature flags with methods to set and clear overrides and access A/B test groups

## **Related issues**

Fixes: 

## **Manual testing steps**

Feature: Remote Feature Flag Override Functionality

  Scenario: user sets a local override for a feature flag
    Given the RemoteFeatureFlagController is initialized with remote flags
    When user calls setFlagOverride('testFlag', true)
    Then testFlag selector should return the override value true
    And getFlagOverride('testFlag') should return true

  Scenario: user clears a specific flag override
    Given a feature flag has a local override set
    When user calls clearFlagOverride('testFlag')
    Then testFlag selector should return the original remote value
    And getFlagOverride('testFlag') should return undefined

  Scenario: user accesses A/B test groups
    Given remote flags contain A/B test configurations
    When user calls rawProcessedRemoteFeatureFlags selector
    Then it should return an array of available test groups with names and values

### **After**

https://github.com/user-attachments/assets/225f41fc-7073-4e5a-9c07-35e22460f12e

https://github.com/user-attachments/assets/9933bc3c-9dcb-4e60-9267-57e6d85dbd01

---

> [!NOTE]
> Adds local override support and A/B test handling for feature flags across selectors, context, UI, and stats; updates controller to v3.1.0.
> 
> - **State/Selectors**:
>   - Merge local overrides into `selectRemoteFeatureFlags`; add `selectLocalOverrides`, `selectRawFeatureFlags`, and `selectRawRemoteFeatureFlags`.
>   - Seed initial/background state and logs with `localOverrides` and `rawRemoteFeatureFlags`.
> - **Context** (`FeatureFlagOverrideContext`):
>   - Refactor to read flags via selectors and call Engine methods (`setFlagOverride`, `removeFlagOverride`, `clearAllFlagOverrides`).
>   - Expose `featureFlags`, `featureFlagsList`, `overrides`, and `getOverrideCount`; remove description/getter utilities.
>   - Bulk-send raw flags to metrics on change.
> - **Utils** (`util/feature-flags`):
>   - Introduce `FeatureFlagType` enum; detect new types: `boolean nested`, `boolean with minimumVersion`, `abTest`.
>   - Keep version support check; remove description helper.
> - **UI** (`FeatureFlagOverride`):
>   - Support new types: boolean, boolean nested, minimumVersion, arrays, objects, and A/B tests (select options from raw flags).
>   - Improve input editing/reset syncing; show `Clear All Overrides (count)` and filtered counts; boolean filter updated.
> - **Stats Hook**:
>   - Count booleans across types; add `abTest` to stats.
> - **Tests**:
>   - Comprehensive updates for context, selectors, UI behaviors, version gating, A/B tests, and stats.
> - **Dependencies**:
>   - Bump `@metamask/remote-feature-flag-controller` to `^3.1.0`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 04dea273451523bb59a0db772063a4fdd68087f2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Cal-L (commented):
The engine ready checks might not be necessary assuming this context is used in the right place. We already have a `ControllersGate` that blocks the majority of components from rendering before the engine is initialized. Can we rely on that instead?

asalsys:
> The engine ready checks might not be necessary assuming this context is used in the right place. We already have a `ControllersGate` that blocks the majority of components from rendering before the engine is initialized. Can we rely on that instead?

Yes, correct,  I don't think it is necessary at all. I had that logic because I was having trouble getting it to start I thought the issue was that the controller wasn't loading but it was because the controller was broken and wasn't building

app/components/Views/FeatureFlagOverride/FeatureFlagOverride.tsx
```diff
@@ -114,6 +129,41 @@ const FeatureFlagRow: React.FC<FeatureFlagRowProps> = ({ flag, onToggle }) => {
             ios_backgroundColor={theme.colors.border.muted}
           />
         );
+      case 'abTest': {
+        interface AbTestType {
```
Cal-L: Move type outside of component

app/components/Views/FeatureFlagOverride/FeatureFlagOverride.tsx
```diff
@@ -114,6 +129,41 @@ const FeatureFlagRow: React.FC<FeatureFlagRowProps> = ({ flag, onToggle }) => {
             ios_backgroundColor={theme.colors.border.muted}
           />
         );
+      case 'abTest': {
```
Cal-L: While extracting types, also cleaner to extract these into enums

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -54,70 +92,71 @@ export const FeatureFlagOverrideProvider: React.FC<
...
+
+  // Get overrides from controller state via Redux
+  const overrides = useSelector(selectLocalOverrides);
   const toastContext = useContext(ToastContext);
   const toastRef = toastContext?.toastRef;
 
-  // Local state for overrides
-  const [overrides, setOverrides] = useState<FeatureFlagOverrides>({});
+  // Subscribe to controller state changes to ensure we stay in sync
+  useEffect(() => {
```
Cal-L, line 56: This doesn't seem like it should do anything since it's triggering an empty function from inside of the useEffect. What's the intended purpose here - is it to get the latest state every time the feature flag changes?

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -20,11 +24,48 @@ import {
...
+export interface ExtendedRemoteFeatureFlagController
+  extends RemoteFeatureFlagController {
+  setFlagOverride: (key: string, value: Json) => void;
+  clearFlagOverride: (key: string) => void;
+  getAllFlags: () => FeatureFlagOverrides;
+  clearAllOverrides: () => void;
+}
+
+// Helper to safely access the RemoteFeatureFlagController with proper typing
+const getRemoteFeatureFlagController = ():
```
Cal-L: This does not need to be a function since it's just returning a reference

app/contexts/FeatureFlagOverrideContext.tsx
```diff
@@ -20,11 +24,48 @@ import {
...
+  | ExtendedRemoteFeatureFlagController
+  | undefined => Engine.context?.RemoteFeatureFlagController as
+    | ExtendedRemoteFeatureFlagController
+    | undefined;
+
+// Helper to safely execute controller methods with error handling
+const withRemoteFeatureFlagController = (
```
Cal-L: the `with` naming convention is usually used for higher order wrappers, which typically used to wrap classes or functional components. Following the logic, it looks like this is intended to handle catching error on method calls for the remote feature flag controller. A few suggestions to make this cleaner:
- Create a remote feature flag hook instead, `useRemoteFeatureFlagController`
- In the hook, return remote feature flag functions that gracefully handle method call failures

app/selectors/featureFlagController/index.ts
```diff
@@ -1,17 +1,57 @@
...
+  selectRemoteFeatureFlagControllerState,
+  (remoteFeatureFlagControllerState: unknown) =>
+    (remoteFeatureFlagControllerState as RemoteFeatureFlagControllerState)
+      ?.remoteFeatureFlags ?? {},
+);
+
 export const selectRemoteFeatureFlags = createSelector(
   selectRemoteFeatureFlagControllerState,
-  (remoteFeatureFlagControllerState) => {
+  (remoteFeatureFlagControllerState: unknown) => {
```
Cal-L: We should use the state type from `RemoteFeatureFlagController`

app/util/feature-flags/index.ts
```diff
@@ -12,6 +12,7 @@ export interface FeatureFlagInfo {
     | 'array'
     | 'boolean with minimumVersion'
     | 'boolean nested'
+    | 'abTest'
```
Cal-L: Cleaner to abstract these into an enum

Cal-L (commented):
Left some comments

app/selectors/featureFlagController/index.ts
```diff
@@ -1,17 +1,57 @@
...
     if (isRemoteFeatureFlagOverrideActivated) {
```
mcmire: Should `isRemoteFeatureFlagOverrideActivated` be extracted to a selector itself? Usually it's a no-no to reference things inside of a selector function that are set elsewhere. Selector functions are changed so if external variables are changed then the result of the selector function won't be recalculated.

For instance if we had

``` typescript
export const selectRemoteFeatureFlags = createSelector(
  selectRemoteFeatureFlagControllerState,
  selectIsRemoteFeatureFlagOverrideActivated,
  (remoteFeatureFlagControllerState, isRemoteFeatureFlagOverrideActivated) => {
    // ...
  }
)
```

then this selector will be better behaved.

package.json
```diff
@@ -266,7 +267,7 @@
     "@metamask/react-native-payments": "^2.0.0",
     "@metamask/react-native-search-api": "1.0.1",
     "@metamask/react-native-webview": "patch:@metamask/react-native-webview@npm%3A14.5.0#~/.yarn/patches/@metamask-react-native-webview-npm-14.5.0-b34fed6d50.patch",
-    "@metamask/remote-feature-flag-controller": "^3.0.0",
```
mcmire, line 269: I see that you removed that caret (`^`), but that is necessary in order for the resolution to override this line. As it is the resolution isn't doing anything so you are not really using the preview build. Can you add this back?

app/contexts/FeatureFlagOverrideContext.tsx
asalsys: I will remove the entire function

weitingsun (dismissed):

app/components/Views/FeatureFlagOverride/FeatureFlagOverride.tsx
```diff
@@ -114,8 +135,36 @@ const FeatureFlagRow: React.FC<FeatureFlagRowProps> = ({ flag, onToggle }) => {
...
+                label: option.name,
+                value: option.name,
+              }))}
+              label={flag.key}
+              defaultValue={(localValue as AbTestType).name}
+              onValueChange={handleSelectOption}
+              selectedValue={(localValue as AbTestType).name}
+            />
+          </Box>
+        );
```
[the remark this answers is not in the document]
asalsys: Return default display if A/B test options are unavailable on line 149

app/components/Views/FeatureFlagOverride/FeatureFlagOverride.tsx
```diff
@@ -114,8 +135,36 @@ const FeatureFlagRow: React.FC<FeatureFlagRowProps> = ({ flag, onToggle }) => {
...
+            (option: { name: string }) => option.name === name,
+          );
+          setLocalValue(selectedOption);
+          onToggle(flag.key, selectedOption);
+        };
+
+        return (
+          <Box twClassName="flex-1 ml-2 justify-center min-w-[160px]">
+            <SelectOptionSheet
+              options={abTestOptions.map((option: AbTestType) => ({
```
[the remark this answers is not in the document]
asalsys: this is resolved

MarioAslau (approved):
Looks good to me!

weitingsun (approved):

Cal-L (approved):
Lgtm

PR #23586, opened 2025-12-03
fix: hide trending section if empty

sahar-fehri:
## **Description**

This PR hides trending tokens section if empty.

The UI needs to show again the tokens section if it gets data on refresh

To test this; i updated the useTrendingRequest hook:

```
const isFirstCall = currentRequestId === 1;
const resultsToStore = isFirstCall
  ? []
  : await getTrendingTokens({
      chainIds: stableChainIds,
      sortBy,
      minLiquidity,
      minVolume24hUsd,
      maxVolume24hUsd,
      minMarketCap,
      maxMarketCap,
    });
```

Result:

https://github.com/user-attachments/assets/9247e4ba-fc3e-458d-a5a7-206bdd427de4

## **Changelog**

CHANGELOG entry: Hides trending section if empty

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **After**

https://github.com/user-attachments/assets/8010c17a-4079-4de3-a4d2-14ca887176bc

---

> [!NOTE]
> Hides empty Explore sections and shows an empty error state for Trending Tokens, while defaulting trending fetch to loading with 24h trending sort.
> 
> - **Trending View (Explore)**:
>   - Hide sections with empty data via `emptySections` state; keep components mounted to detect new data.
>   - Pass `toggleSectionEmptyState` to each section (`SectionCard`, `SectionCarrousel`) and wire through `sections.config`.
>   - Update `QuickActions` to only show visible sections.
> - **Trending Tokens Full View**:
>   - Show skeletons only while `isLoading`; when no results, render `EmptyErrorTrendingState` with retry.
>   - Disable Price Change control when no results.
> - **Hook**:
>   - `useTrendingRequest`: default `sortBy` to `h24_trending`; initial `isLoading` set to `true`.
> - **New Component**:
>   - `EmptyErrorTrendingState` with tests and localized strings.
> - **Tests & i18n**:
>   - Update tests to use `useTrendingSearch` and validate new empty/error state and controls.
>   - Add `trending.empty_error_trending_state.*` strings.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e96b2f05c1f6f4b047a4e5c703b08f1168c0ca97. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/components/UI/Trending/hooks/useTrendingRequest/useTrendingRequest.ts
+++ b/app/components/UI/Trending/hooks/useTrendingRequest/useTrendingRequest.ts
@@ -48,7 +48,7 @@ export const useTrendingRequest = (options: {
     Awaited<ReturnType<typeof getTrendingTokens>>
   >([]);
 
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
 
   const [error, setError] = useState<Error | null>(null);
 
--- a/app/components/UI/Trending/hooks/useTrendingRequest/useTrendingRequest.ts
+++ b/app/components/UI/Trending/hooks/useTrendingRequest/useTrendingRequest.ts
@@ -22,7 +22,7 @@ export const useTrendingRequest = (options: {
 }) => {
   const {
     chainIds: providedChainIds = [],
-    sortBy,
+    sortBy = 'h24_trending',
     minLiquidity = 0,
     minVolume24hUsd = 0,
     maxVolume24hUsd,
--- a/app/components/Views/TrendingTokens/TrendingTokensFullView/TrendingTokensFullView.tsx
+++ b/app/components/Views/TrendingTokens/TrendingTokensFullView/TrendingTokensFullView.tsx
@@ -41,6 +41,7 @@ import {
 } from '../../../UI/Trending/components/TrendingTokensBottomSheet';
 import { sortTrendingTokens } from '../../../UI/Trending/utils/sortTrendingTokens';
 import { useTrendingSearch } from '../../../UI/Trending/hooks/useTrendingSearch/useTrendingSearch';
+import EmptyErrorTrendingState from '../../TrendingView/components/EmptyErrorState/EmptyErrorTrendingState';
 
 interface TrendingTokensNavigationParamList {
   [key: string]: undefined | object;
@@ -112,6 +113,9 @@ const createStyles = (theme: Theme) =>
       fontWeight: '600',
       lineHeight: 19.6, // 140% of 14px
       fontStyle: 'normal',
+    },
+    controlButtonDisabled: {
+      opacity: 0.5,
     },
   });
 
@@ -312,8 +316,12 @@ const TrendingTokensFullView = () => {
             <TouchableOpacity
               testID="price-change-button"
               onPress={handlePriceChangePress}
-              style={styles.controlButton}
+              style={[
+                styles.controlButton,
+                searchResults.length === 0 && styles.controlButtonDisabled,
+              ]}
               activeOpacity={0.2}
+              disabled={searchResults.length === 0}
             >
               <View style={styles.controlButtonContent}>
                 <Text style={styles.controlButtonText}>
@@ -366,12 +374,14 @@ const TrendingTokensFullView = () => {
         </View>
       ) : null}
 
-      {isLoading || (searchResults as TrendingAsset[]).length === 0 ? (
+      {isLoading ? (
         <View style={styles.listContainer}>
           {Array.from({ length: 12 }).map((_, index) => (
             <TrendingTokensSkeleton key={index} />
           ))}
         </View>
+      ) : (searchResults as TrendingAsset[]).length === 0 ? (
+        <EmptyErrorTrendingState onRetry={handleRefresh} />
       ) : (
         <View style={styles.listContainer}>
           <TrendingTokensList
--- a/app/components/Views/TrendingView/TrendingView.tsx
+++ b/app/components/Views/TrendingView/TrendingView.tsx
@@ -79,18 +79,15 @@ const TrendingFeed: React.FC = () => {
   );
 
   // Memoize callbacks for each section to prevent infinite loops
-  const sectionEmptyCallbacks = useMemo(
-    () =>
-      HOME_SECTIONS_ARRAY.reduce(
-        (acc, section) => {
-          acc[section.id] = (isEmpty: boolean) =>
-            handleSectionEmptyChange(section.id, isEmpty);
-          return acc;
-        },
-        {} as Record<SectionId, (isEmpty: boolean) => void>,
-      ),
-    [handleSectionEmptyChange],
-  );
+  const sectionCallbacks = useMemo(() => {
+    const callbacks: Record<SectionId, (isEmpty: boolean) => void> =
+      {} as Record<SectionId, (isEmpty: boolean) => void>;
+    HOME_SECTIONS_ARRAY.forEach((section) => {
+      callbacks[section.id] = (isEmpty: boolean) =>
+        handleSectionEmptyChange(section.id, isEmpty);
+    });
+    return callbacks;
+  }, [handleSectionEmptyChange]);
 
   const handleBrowserPress = useCallback(() => {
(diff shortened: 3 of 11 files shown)


app/components/UI/Trending/hooks/useTrendingRequest/useTrendingRequest.ts
```diff
@@ -48,7 +48,7 @@ export const useTrendingRequest = (options: {
     Awaited<ReturnType<typeof getTrendingTokens>>
   >([]);
 
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
```
sahar-fehri, line 51: I had to set this initially to true because there was a race condition happening with the new logic
In SectionCard.tsx, there's this effect:
```
useEffect(() => {
  if (!isLoading && onEmptyChange) {
    onEmptyChange(data.length === 0);
  }
}, [data.length, isLoading, onEmptyChange]);
```

Component mounts → isLoading = false, data = []
Effect runs immediately → onEmptyChange(true) → section hidden! 
API call starts...
API returns data → data = [...], isLoading = false
Effect runs → onEmptyChange(false) → but section is already hidden
The section was being marked as empty before the API even had a chance to fetch data.

TL;DR: We wait for the actual API response before deciding if the section is empty, rather than assuming it's empty while loading.

app/components/Views/TrendingView/TrendingView.tsx
```diff
@@ -58,6 +61,37 @@ const TrendingFeed: React.FC = () => {
...
+          next.delete(sectionId);
+        }
+        return next;
+      });
+    },
+    [],
+  );
+
+  // Memoize callbacks for each section to prevent infinite loops
+  const sectionEmptyCallbacks = useMemo(
```
sahar-fehri: Removing this would cause an infinite loop in re-render

Prob is that every render creates a new arrow function → child's useEffect re-runs (setting toggleSectionEmptyState) → calls callback → updates parent state → re-render → new function → infinite loop

app/components/Views/TrendingTokens/TrendingTokensFullView/TrendingTokensFullView.tsx
```diff
@@ -366,12 +374,14 @@ const TrendingTokensFullView = () => {
...
       ) : null}
 
-      {isLoading || (searchResults as TrendingAsset[]).length === 0 ? (
+      {isLoading ? (
         <View style={styles.listContainer}>
           {Array.from({ length: 12 }).map((_, index) => (
             <TrendingTokensSkeleton key={index} />
           ))}
         </View>
+      ) : (searchResults as TrendingAsset[]).length === 0 ? (
```
juanmigdr, line 383: Love that you put it on a separate component!

app/components/Views/TrendingView/components/QuickActions/QuickActions.tsx
```diff
@@ -10,22 +10,31 @@ import {
   TextVariant,
 } from '@metamask/design-system-react-native';
 import { useTailwind } from '@metamask/design-system-twrnc-preset';
-import { SECTIONS_ARRAY } from '../../config/sections.config';
+import { SECTIONS_ARRAY, SectionId } from '../../config/sections.config';
+
+interface QuickActionsProps {
+  /** Set of section IDs that have empty data and should be hidden */
+  emptySections?: Set<SectionId>;
```
juanmigdr: Thinking if we should make this non-nullable

app/components/Views/TrendingView/components/QuickActions/QuickActions.tsx
```diff
@@ -10,22 +10,31 @@ import {
...
  * centralized sections configuration. When a new section is added to SECTIONS_CONFIG,
  * a corresponding button will automatically appear here.
  */
-const QuickActions: React.FC = () => {
+const QuickActions: React.FC<QuickActionsProps> = ({ emptySections }) => {
   const navigation = useNavigation();
   const tw = useTailwind();
 
+  const visibleSections = emptySections
+    ? SECTIONS_ARRAY.filter((s) => !emptySections.has(s.id))
```
juanmigdr: I think sections is always provided so we should always filter, you dont need a null check

app/components/Views/TrendingView/TrendingView.tsx
```diff
@@ -138,14 +172,25 @@ const TrendingFeed: React.FC = () => {
...
-
-          {HOME_SECTIONS_ARRAY.map((section) => (
-            <React.Fragment key={section.id}>
-              <SectionHeader sectionId={section.id} />
-              <section.Section refreshTrigger={refreshTrigger} />
-            </React.Fragment>
-          ))}
+          <QuickActions emptySections={emptySections} />
+
+          {HOME_SECTIONS_ARRAY.map((section) => {
```
juanmigdr:
```suggestion
          {HOME_SECTIONS_ARRAY.filter((section) => emptySections.has(section.id)).map((section) => {
            <React.Fragment key={section.id}>
              <SectionHeader sectionId={section.id} />
              <section.Section refreshTrigger={refreshTrigger} />
            </React.Fragment>
          })}
```

app/components/Views/TrendingView/TrendingView.tsx
juanmigdr: Can we make it more readable?
```typescript
  const sectionCallbacks = useMemo(() => {
    const callbacks: Record<SectionId, (isEmpty: boolean) => void> =
      {} as Record<SectionId, (isEmpty: boolean) => void>;
    HOME_SECTIONS_ARRAY.forEach((section) => {
      callbacks[section.id] = (isEmpty: boolean) =>
        handleSectionEmptyChange(section.id, isEmpty);
    });
    return callbacks;
  }, [handleSectionEmptyChange]);
```

juanmigdr: Actually we can remove both: sectionEmptyCallbacks and handleSectionEmptyChange

And replace it with:
```typescript
  const sectionCallbacks = useMemo(() => {
    const callbacks = {} as Record<SectionId, (isEmpty: boolean) => void>;
    HOME_SECTIONS_ARRAY.forEach((section) => {
      callbacks[section.id] = (isEmpty: boolean) => {
        setEmptySections((prev) => {
          const next = new Set(prev);
          if (isEmpty) {
            next.add(section.id);
          } else {
            next.delete(section.id);
          }
          return next;
        });
      };
    });
    return callbacks;
  }, []);
```

Then we just use it like `toggleSectionEmptyState={sectionCallbacks[section.id]}`

app/components/Views/TrendingView/components/QuickActions/QuickActions.tsx
sahar-fehri: actually makes sense since now all sections have this

app/components/Views/TrendingView/TrendingView.tsx
sahar-fehri: i think you meant this
```
          {HOME_SECTIONS_ARRAY.filter(
            (section) => !emptySections.has(section.id),
          ).map((section) => (
            <Box key={section.id}>
              <SectionHeader sectionId={section.id} />
              <section.Section
                refreshTrigger={refreshTrigger}
                toggleSectionEmptyState={sectionCallbacks[section.id]}
              />
            </Box>
          ))}
```

sahar-fehri: Actually, this was causing a bug where after refresh, even if data present it does not show the section again;
The filter does not work in this case; 
Doing that hidden: const isHidden = emptySections.has(section.id); is what did the trick
with that the component stays mounted in a way that when refresh triggers and child reports back to parent it re-adds the section back;
with the filter; the component is no longer mounted, hence even if we get data the section stays hidden

juanmigdr (approved):

app/components/Views/TrendingView/TrendingView.tsx
juanmigdr: Oh nice! Good one! 🚀

PR #24040, opened 2025-12-15
fix: make sure pending connection approvals can't get stuck in queue

adonesky1:
## **Description**

These changes are aiming to fix a case of stuck pending approvals we've been running into in our (@MetaMask/wallet-integrations) work on the MMConnect (new SDK). In some edge cases you can get approvals stuck in state where the `isProcessing` flag on the `PermissionApproval` component is never reset resulting in approvals queued behind it never showing and any subsequent request getting permanently stuck in this queue where nothing is shown and the user has no opportunity to approve or reject anything in the queue. As you can see in the "AFTER" video below, the external browser connection flows we support with the SDK create the possibility of multiple connection requests getting queued (in app browser avoids this since the confirmation blocks further action) since the SDK does not support per origin request rate limiting. 

**Root Cause:**
The `PermissionApproval` component used a `useRef`-based `isProcessing` flag to prevent duplicate processing of approval requests. However, in certain edge cases (especially with external browser paths where multiple requests can queue), this flag would never reset to `false`, causing:
1. The current approval to be blocked from processing
2. All queued approvals behind it to never display
3. Subsequent requests to get permanently stuck in the queue with no UI shown to the user

**Solution:**
- Removed the `isProcessing` ref-based guard mechanism that was causing the stuck state
- Removed the early return check for `ApprovalTypes.REQUEST_PERMISSIONS` and `eventSource` that was resetting the flag inappropriately
- Added `pendingApprovalList` to the `useEffect` dependency array so the effect properly re-runs when the approval queue changes for further assurances that the queue does not get permanently stuck.

This change allows the component to reactively process approval requests based on the actual state of the approval queue, rather than relying on a ref-based flag that can get out of sync.

## **Changelog**

CHANGELOG entry: Fixed an issue where pending approval requests could become permanently stuck in the queue, preventing users from approving or rejecting subsequent connection requests.

## **Related issues**

Fixes: n/a

## **Manual testing steps**

```gherkin
Feature: Permission approval queue handling

  Scenario: Multiple connection requests from external browser
    Given the user has MetaMask Mobile open
    And an external browser (not in-app browser) is used to initiate connection requests
    
    When multiple connection requests are queued in quick succession
    Then all queued approval requests should be displayed sequentially
    And users should be able to approve or reject each request
    And no requests should become permanently stuck in the queue

  Scenario: Approval queue processing after navigation
    Given a connection request is in the approval queue
    And the user navigates away from the approval screen
    
    When the user returns or a new approval request is added
    Then the approval queue should continue processing correctly
    And all pending approvals should be accessible to the user
```

### **Before**

https://github.com/user-attachments/assets/b4df84ee-2742-464f-8af3-fe0a99837a63

### **After**

https://github.com/user-attachments/assets/e74dda41-d57f-4a6a-90fe-045e3f0bfb1c

---

> [!NOTE]
> Remove ref-based processing guard and re-run effect when pending approvals change to avoid stuck permission approvals; add targeted tests and selector-driven mocks.
> 
> - **Approvals**
>   - `PermissionApproval.tsx`:
>     - Remove `isProcessing` ref guard and related reset logic.
>     - Use `useSelector(selectPendingApprovals, isEqual)` and add `pendingApprovals` to `useEffect` deps to re-run on queue changes.
>     - Null-safe check for `isEip1193Request`.
>     - Keep navigation to `AccountConnect` and analytics dispatch intact.
>   - `PermissionApproval.test.tsx`:
>     - Refactor to mock `selectPendingApprovals` and `selectAccountsLength` instead of full background state.
>     - Mock `getApiAnalyticsProperties` and `getAllScopesFromPermission`.
>     - Add tests verifying effect re-runs when `pendingApprovals` changes and after queue is cleared, ensuring navigation occurs; retain analytics and guard-case tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a11326d4ee5aea9c4ad32f767fd25c8981cb111e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/Approvals/PermissionApproval/PermissionApproval.tsx
```diff
@@ -87,6 +72,7 @@ const PermissionApproval = (props: PermissionApprovalProps) => {
     trackEvent,
     createEventBuilder,
     eventSource,
+    pendingApprovalList,
```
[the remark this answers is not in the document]
jiexi: this is a fair point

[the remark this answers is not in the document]
jiexi: nit. add a comment explaining why this is needed for future devs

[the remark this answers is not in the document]
adonesky1: addressed here: https://github.com/MetaMask/metamask-mobile/pull/24040/commits/5080ba378287032e6243d658d1b67fc60f3de585

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
```diff
@@ -87,6 +72,11 @@ const PermissionApproval = (props: PermissionApprovalProps) => {
     trackEvent,
     createEventBuilder,
     eventSource,
+    // Include pendingApprovals to ensure the effect re-runs when the approval queue changes.
+    // This prevents the queue from getting permanently stuck and handles cases where new approvals
+    // are added to the list, ensuring previous approvals that weren't rendered for other reasons
+    // can be processed. Ideally we can identify all cases where approval fail to render and then remove this dependency.
+    pendingApprovals,
   ]);
```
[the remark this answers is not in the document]
adonesky1: Perhaps we should guard against the duplicate `trackEvents` but I believe the navigation is not an issue since the `approvalRequest` returned by `useApprovalRequest` will always be the first in the queue meaning subsequent queued requests while the first one is "processing" will not cause a navigation away from the currently processing approval.

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
```diff
@@ -21,24 +20,14 @@ export interface PermissionApprovalProps {
...
-      approvalRequest?.type !== ApprovalTypes.REQUEST_PERMISSIONS ||
-      !eventSource
-    ) {
-      isProcessing.current = false;
-      return;
-    }
-
     const requestData = approvalRequest?.requestData;
 
     if (!requestData?.permissions?.[Caip25EndowmentPermissionName]) return;
```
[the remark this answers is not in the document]
adonesky1: restored this check/gate here: https://github.com/MetaMask/metamask-mobile/pull/24040/commits/3d0f90edc000905b80d1ae52c9ff63630bbc7b0d

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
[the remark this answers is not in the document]
jiexi: looking at the dependencies, this shouldn't rerun unless the approvals have changed. Likely not an issue

ffmcgee725 (dismissed):

app/components/Views/confirmations/hooks/useApprovalRequest.ts
```diff
@@ -12,17 +12,15 @@ export type ApprovalRequestType = ApprovalRequest<any>;
 
 const useApprovalRequest = () => {
   const pendingApprovals = useSelector(selectPendingApprovals, isEqual);
-  const pendingApprovalList = Object.values(pendingApprovals ?? {});
+  const pendingApprovalList = useMemo(
```
matthewwalsh0: Is it not slower to clone the whole list if we only need the first element, and we'd now be cloning whenever anything changes, even irrelevant approvals?

app/components/Views/confirmations/hooks/useApprovalRequest.ts
```diff
@@ -63,6 +61,7 @@ const useApprovalRequest = () => {
     pageMeta,
     onConfirm,
     onReject,
+    pendingApprovals,
```
matthewwalsh0: Does this not break encapsulation since the intent of this hook is to locate the "current" approval request?

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
```diff
@@ -87,6 +80,11 @@ const PermissionApproval = (props: PermissionApprovalProps) => {
     trackEvent,
     createEventBuilder,
     eventSource,
+    // Include pendingApprovals to ensure the effect re-runs when the approval queue changes.
+    // This prevents the queue from getting permanently stuck and handles cases where new approvals
+    // are added to the list, ensuring previous approvals that weren't rendered for other reasons
+    // can be processed. Ideally we can identify all cases where approval fail to render and then remove this dependency.
+    pendingApprovals,
```
matthewwalsh0, line 90: If the `approvalRequest` is already memoised from the full list, then I'm not sure I understand why contriving a dependence on the entire list would make any functional difference?

Since the content of the other approvals is irrelevant, and when one approval request is removed, the `useEffect` should already re-fire since `approvalRequest` is now a new reference?

matthewwalsh0 (dismissed):

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
matthewwalsh0: I think I follow the intent. We have a different bug causing these not to render, so we're just using future approvals as a trigger to re-render and workaround that.

Should we use `selectPendingApprovals` directly to avoid changing the return of `useApprovalRequest`?

jiexi: 🤔

app/components/Views/confirmations/hooks/useApprovalRequest.ts
jiexi: i figured i'd copy the original deepClone on the single pendingApproval that was returned. This is no longer relevant of course

app/components/Views/confirmations/hooks/useApprovalRequest.ts
jiexi: yeah, true. this is overloaded. no longer relevant now

app/components/Approvals/PermissionApproval/PermissionApproval.tsx
jiexi: addressed in the commits below

app/components/Approvals/PermissionApproval/PermissionApproval.test.tsx
```diff
@@ -43,13 +58,20 @@ const NAV_DETAILS_MOCK = [
...
     useApprovalRequest as jest.MockedFn<typeof useApprovalRequest>
   ).mockReturnValue({
     approvalRequest,
+    pendingApprovals: pendingApprovals || {},
     onConfirm: jest.fn(),
+    onReject: jest.fn(),
+    pageMeta: {},
     // TODO: Replace "any" with type
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   } as any);
```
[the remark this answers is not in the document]
adonesky1: addressed here: https://github.com/MetaMask/metamask-mobile/pull/24040/commits/1ff85c391f1caaa967a5adeafd1b93b518ddfbbc

jiexi (approved):

PR #24388, opened 2026-01-09
chore: remove environment specific naming in action

sethkfman:
## **Description**

This PR updates the push-eas-update workflow:
- added new env mapping to accept all env values needed to build RC and Production applications
- removes yarn cache restore
- addresses PR hash pinning and validation 

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> Introduces commit-hash–based publishing with validation and tightens OTA workflow gating.
> 
> - Adds `./.github/scripts/validate-pr-commit.sh` to verify `COMMIT_HASH` is the PR head and base branch matches
> - Updates `push-eas-update.yml` to accept `commit_hash`, check out the exact commit for fingerprinting and publishing, and record fingerprints for that commit vs `BASE_BRANCH`
> - New `validate-pr` job exposes validated `pr_number` for the approval step; `approval` and `push-update` now depend on both fingerprint match and successful validation
> - Streamlines steps (removes yarn cache hints and extraneous output), expands env/secret mappings, and adjusts summary outputs
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e7e0b876d2dfdd0198f5bdeea10b18711244a6e2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


.github/workflows/push-eas-update.yml
```diff
@@ -50,46 +62,66 @@ jobs:
...
+          echo "🔎 Base branch HEAD: $(git rev-parse --abbrev-ref HEAD || echo 'detached HEAD')"
+          BASE_SHA=$(git rev-parse HEAD)
+          echo "🔎 Base branch commit: $BASE_SHA"
+          echo "sha=$BASE_SHA" >> "$GITHUB_OUTPUT"
+
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
           node-version: '20'
           cache: 'yarn'
```
NEllusion: Instead of using actions/setup node with `cache`, can we use https://github.com/MetaMask/action-checkout-and-setup with `is-high-risk-environment: true` ? If not we can just remove caching all together. See https://github.com/MetaMask/MetaMask-planning/issues/3925 for risk of using GitHub's actions/cache

.github/workflows/push-eas-update.yml
```diff
@@ -36,8 +36,20 @@ env:
   TARGET_CHANNEL: ${{ inputs.channel }}
 
 jobs:
+  setup-dependencies:
+    name: Setup Dependencies
+    uses: ./.github/workflows/setup-node-modules.yml
+    with:
+      ref: refs/pull/${{ inputs.pr_number }}/head
```
NEllusion: Instead of the ref being a PR number (where the latest commit on that PR can change at any time), can we extract and use the commit SHA of the last commit on the branch instead?

If I were an attacker with write access, I would trigger an OTA update to this repository by setting `pr_number` to my PR or branch name with malicious content on it. The workflow would then generate an artifact based off my branch pulling in malicious dependencies..

I would then update my branch to show only safe, legitimate looking code that doesn't pose a security risk. Then a few hours later, the approval step might happen where a member of the release team reviews the PR I am planning to use, and approves it as it looks secure.  The result is that my malicious code from earlier gets used in the OTA release.

Using the commit SHA everywhere ensures that between the time that this `setup-dependencies` occurs, and when the actual build happens, we are using the same code.

.github/workflows/push-eas-update.yml
```diff
@@ -194,15 +243,20 @@ jobs:
           node-version: '20'
           cache: 'yarn'
```
NEllusion: Similar to the comment above (https://github.com/MetaMask/metamask-mobile/pull/24388/changes#r2683850875) we should avoid using caching in release workflows

NEllusion: As node 20 is almost at EOL I would recommend using a later LTS version of node as well: https://endoflife.date/nodejs

.github/workflows/setup-node-modules.yml
```diff
@@ -0,0 +1,127 @@
...
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+          fetch-depth: ${{ inputs.fetch-depth }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'yarn'
```
NEllusion: See https://github.com/MetaMask/metamask-mobile/pull/24388/changes#r2683850875. We should remove caching from this file

.github/workflows/push-eas-update.yml
```diff
@@ -36,8 +36,20 @@ env:
...
 jobs:
+  setup-dependencies:
+    name: Setup Dependencies
+    uses: ./.github/workflows/setup-node-modules.yml
+    with:
+      ref: refs/pull/${{ inputs.pr_number }}/head
+      fetch-depth: 0
+      upload-artifact: true
+      artifact-name: node-modules-eas-update
+      artifact-retention-days: 1
```
NEllusion: Here we generate the artifact, then wait for approval later on in the process. Just flagging that if that approval takes more than 24 hours, the artifact will expire. It may make sense to begin this artifact generation after the PR has been approved.

.github/workflows/setup-node-modules.yml
```diff
@@ -0,0 +1,127 @@
...
+          echo "Cache key: $CACHE_KEY"
+
+      - name: Restore node_modules cache
+        id: cache-node-modules
+        uses: actions/cache@v4
+        with:
+          path: |
+            node_modules
+            .yarn/cache
+            .yarn/install-state.gz
```
NEllusion: We currently gitignore these files in the mobile repo. Are they needed here?

.github/workflows/push-eas-update.yml
NEllusion, line 13: Are there specific branches we want to release from? Enabling this workflow by default to be capable of operating on any PR + relying on humans to approve / deny this work flow will be prone to social engineering, or human error. 

Could you clarify each spot along the release workflow where an approval is required?

NEllusion: This is going to be executing code from an un-merged pull request in an environment that has access to sensitive secrets. Are we able to ensure that we are only building releases from branches that have branch protection rules + controls on them?

NEllusion, line 314: Would we be missing out on any security protections during the execution of the scripts below?

NEllusion, line 17: A malicious internal attacker could set both the `pr_number` and the `base_branch` so in theory, they could set something that allows for the expo fingerprinting to pass. Would the worst case scenario here be  just that the app is broken since native changes were not included?

NEllusion: Similar to above, we should use a SHA so that we are operating on the same set of code that existed at the time of setup, and time of approval.

.github/workflows/setup-node-modules.yml
```diff
@@ -0,0 +1,127 @@
...
+    name: Setup Node Modules
+    runs-on: ubuntu-latest
+    outputs:
+      cache-hit: ${{ steps.cache-node-modules.outputs.cache-hit }}
+      cache-key: ${{ steps.cache-key.outputs.key }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref || github.ref }}
```
NEllusion: What is the value of github.ref in this instance?

.github/workflows/setup-node-modules.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Upload node_modules artifact
+        if: inputs.upload-artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: |
+            node_modules
+            .yarn/cache
+            .yarn/install-state.gz
```
NEllusion: Similar to above, we should look at if we need to include these in our artifact (as they are gitignored in the repo by default)

tommasini: Yeahhh, I guess if we are not installing again, we might not need this, this is cache to speed up the dependencies install step.  will test it!

.github/workflows/setup-node-modules.yml
tommasini: inputs.ref
* Optional input passed when calling this reusable workflow
* If provided, uses that specific ref (commit hash, branch, PR ref, etc.)
* Example: ref: ${{ inputs.commit_hash }} → checks out that commit

github.ref
* Default ref of the calling workflow
* The branch/ref where the calling workflow is running
* Example: if called from main, github.ref = refs/heads/main


So I don't think we want the input.ref to be the source of truth right? It should be the branch used to trigger the workflow

tommasini: It really doesn't matter the fingerprint should match!

.github/workflows/setup-node-modules.yml
```diff
@@ -0,0 +1,115 @@
...
+        with:
+          ref: ${{ inputs.ref }}
+          fetch-depth: ${{ inputs.fetch-depth }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Enable Corepack
```
tommasini: I don't think we need enable corepack at the end of day

.github/workflows/push-eas-update.yml
```diff
@@ -201,24 +277,37 @@ jobs:
...
         uses: actions/setup-node@v4
         with:
           node-version: '20'
-          cache: 'yarn'
 
-      - name: Install dependencies
+      - name: Enable Corepack
+        run: corepack enable && corepack prepare yarn@4.10.3 --activate
+
+      - name: Clear yarn state before artifact download
```
tommasini: I'm mostly sure that we don't need to clear the state either before downloading node modules since the reason why it was failing it was the state.tz file

.github/workflows/push-eas-update.yml
```diff
@@ -201,24 +277,37 @@ jobs:
...
           fetch-depth: 0
 
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
           node-version: '20'
-          cache: 'yarn'
 
-      - name: Install dependencies
+      - name: Enable Corepack
```
tommasini: Mostly sure that we can remove the corepack

sethkfman (commented):
LGTM after @weitingsun updates

weitingsun (approved):

Cal-L (approved):
LGTM

PR #24554, opened 2026-01-15
feat: cp-7.62.0 Use `ConnectivityController` to suppress network banners and events when device is offline

cryptodev-2s:
## Description
Adds device connectivity detection to distinguish between "device offline" and "network/RPC degraded" states.

## Problem
When a device goes offline, the mobile app incorrectly:

- Shows "Network Degraded/Unavailable" banners (misleading - it's the device, not the network)
- Tracks RPC endpoint failure metrics (false positives polluting analytics)

## Solution
Introduces `ConnectivityController` that tracks device online/offline status using `@react-native-community/netinfo`.

### Implementation Details

The mobile implementation uses `NetInfoConnectivityAdapter` which checks both connectivity indicators:

- **`isInternetReachable`** (preferred): Detects actual internet connectivity - ensures the device can reach the internet, not just connected to a network
- **`isConnected`** (fallback): Used when `isInternetReachable` is null (still determining connectivity) - checks if device has any network connection (WiFi, cellular, etc.)

This dual-check approach is critical because:
- A device can be `isConnected: true` (connected to WiFi/cellular) but `isInternetReachable: false` (no actual internet access)
- We need to detect true offline state, not just network interface availability
- The existing `OfflineMode` view already handles full-screen offline scenarios, and `ConnectivityController` ensures network banners don't show when internet is not reachable

### Key benefits:

- **Suppressed false-positive metrics**: RPC endpoint unavailable/degraded events are no longer tracked when the device is offline
- **Hidden misleading banners**: Network connection banners are hidden when internet is not reachable (even if device shows network connection)
- **Accurate offline detection**: Checks both `isInternetReachable` and `isConnected` to properly distinguish between "connected to network" vs "has internet access"

### Integration with Existing Offline UI

Mobile already has an `OfflineMode` view ([app/components/Views/OfflineMode/index.js](https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/OfflineMode/index.js#L61)) that handles device offline scenarios with a full-screen view. This view:
- Uses `NetInfo.useNetInfo()` to detect when the device has no connectivity
- Displays a dedicated offline screen with an astronaut image
- Provides a "Try Again" button that checks connectivity before allowing navigation
- Handles both regular offline scenarios and Infura blocked cases

The `ConnectivityController` integration complements this existing UI by:
- Hiding network connection banners when internet is not reachable (even if `isConnected` is true)
- Ensuring analytics accurately distinguish between device offline vs. RPC endpoint issues
- Providing a consistent connectivity state across the app that can be used by other components
- Using the same dual-check logic (`isInternetReachable` + `isConnected`) to ensure accurate offline detection

This ensures users see the appropriate `OfflineMode` screen for device connectivity issues, while network/RPC-specific banners only appear when the device has internet connectivity but is experiencing RPC endpoint problems.

## **Changelog**

CHANGELOG entry: Fixed misleading "Network Degraded" banners and false-positive RPC endpoint metrics when the device has no internet connection

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/WPC-210

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> Introduces device connectivity awareness and gates network degradation UX/metrics on actual internet reachability.
> 
> - Adds `ConnectivityController` with `NetInfoConnectivityAdapter`; wires into Engine context, messengers, types, constants, initial background state
> - Exposes selectors `selectConnectivityStatus`/`selectIsDeviceOffline`; `useNetworkConnectionBanner` now hides/suppresses degraded/unavailable banners and timers when offline
> - Updates NetworkController init to delegate `ConnectivityController:getState` and pass `isOffline` to RPC service options, preventing false RPC failure analytics
> - Bumps deps (`@metamask/network-controller` -> ^29, adds `@metamask/connectivity-controller`) and adds comprehensive unit tests for adapter, controller init, Engine, and banner hook
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 117302063e88f91814406cea018856fa5e45fa65. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/core/Engine/controllers/connectivity/netinfo-connectivity-adapter.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export class NetInfoConnectivityAdapter implements ConnectivityAdapter {
+  #unsubscribe: NetInfoSubscription | null = null;
+  #onConnectivityChangeCallbacks: ((status: ConnectivityStatus) => void)[] = [];
+  #currentState: NetInfoState | null = null;
+
+  /**
+   * Returns the current connectivity status.
+   *
+   * For initial state, we default to online since NetInfo requires async
```
mcmire, line 45: ~This works for now, but we should consider refactoring this after the release. I feel like we can fix this by making `getStatus` async.~ See below.

app/core/Engine/controllers/connectivity/netinfo-connectivity-adapter.ts
```diff
@@ -0,0 +1,108 @@
...
+        cb(
+          isOnline
+            ? CONNECTIVITY_STATUSES.Online
+            : CONNECTIVITY_STATUSES.Offline,
+        ),
+      );
+    });
+
+    // Subscribe to NetInfo state changes
+    this.#unsubscribe = netInfoAddEventListener((state: NetInfoState) => {
```
mcmire: Along with the proposed refactor above, perhaps we should add an event listener in the constructor of the service? Then `onConnectivityChange` will be very simple.

app/selectors/connectivityController.ts
```diff
@@ -0,0 +1,29 @@
...
+const selectConnectivityControllerState = (state: RootState) =>
+  state?.engine?.backgroundState?.ConnectivityController;
+
+/**
+ * Selector to get the connectivity status from ConnectivityController.
+ *
+ * @param state - The root state.
+ * @returns The connectivity status ('online' | 'offline').
+ */
+export const selectConnectivityStatus = createSelector(
```
mcmire, line 14: Ideally, selectors (aside from `selectConnectivityControllerState`) should be located in controller packages so that all clients can use them. We should consider moving these to the `connectivity-controller` package.

mcmire (commented):
I see that there are some concessions we had to make, but putting those aside, this looks good.

Nit: The PR title says that adding the controller is the main change here but that is more of a side effect I think? It seems that the main change is really about hiding the degraded/unavailable banners when the user is offline.

cryptodev-2s:
@SocketSecurity ignore npm/@metamask/eth-json-rpc-middleware@23.0.0

cryptodev-2s:
@SocketSecurity ignore npm/@metamask/eth-block-tracker@15.0.1

app/core/Engine/controllers/connectivity/netinfo-connectivity-adapter.ts
mcmire: Actually, maybe we should do this now? The Cursor bot comment below is correct; it _seems_ that we support `onConnectivityChange` being called more than once, but in fact we do not, because 1) `#unsubscribe` is overwritten each time and 2) another event listener is added via `netinfo`, and inside of that listener we iterate through all callbacks. So it just seems strange.

I think if would be better if we set up this event listener in the constructor:

``` typescript
constructor() {
  this.#removeNetInfoEventListener = netInfoAddEventListener((state: NetInfoState) => {
    this.#update(state);
  });
}
```

Then `onConnectivityChange` can look like:

``` typescript
onConnectivityChange(callback: (status: ConnectivityStatus) => void): void {
  this.#onConnectivityChangeCallbacks.push(callback);

  // Fetch initial state first
  netInfoFetch().then((state: NetInfoState) => {
    this.#update(state);
  });
}
```

We can also add an `#update` method:

``` typescript
#update(state: NetInfoState) {
  this.#currentState = state;
  const status = this.getStatus();

  this.#onConnectivityChangeCallbacks.forEach((callback) => callback(status));
}

Comment (cryptodev-2s) [reply app/core/Engine/controllers/connectivity/netinfo-connectivity-adapter.ts]: fixed here https://github.com/MetaMask/metamask-mobile/pull/24554/changes/a78217b4aa7f056f7973d027647169b9c51a18fe

File: app/selectors/connectivityController.ts [hunk 4]
```diff
@@ -0,0 +1,29 @@
+import { createSelector } from 'reselect';
+import { CONNECTIVITY_STATUSES } from '@metamask/connectivity-controller';
+import { RootState } from '../reducers';
+
+const selectConnectivityControllerState = (state: RootState) =>
+  state?.engine?.backgroundState?.ConnectivityController;
```
mcmire, line 6: Nit: Isn't the state guaranteed to be there?

``` suggestion
  state.engine.backgroundState.ConnectivityController;
```

app/selectors/connectivityController.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Selector to get the connectivity status from ConnectivityController.
+ *
+ * @param state - The root state.
+ * @returns The connectivity status ('online' | 'offline').
+ */
+export const selectConnectivityStatus = createSelector(
+  [selectConnectivityControllerState],
+  (connectivityState) =>
+    connectivityState?.connectivityStatus ?? CONNECTIVITY_STATUSES.Online,
```
mcmire, line 17: Nit: We should trust the state is there, no?

``` suggestion
  (connectivityState) => connectivityState.connectivityStatus;
```

app/selectors/connectivityController.ts
```diff
@@ -0,0 +1,29 @@
...
+);
+
+/**
+ * Selector to check if the device is offline.
+ *
+ * @param state - The root state.
+ * @returns True if the device is offline, false otherwise.
+ */
+export const selectIsDeviceOffline = createSelector(
```
mcmire, line 26: Nit: Do we need this if we also have `selectConnectivityStatus`?

mcmire (approved):
A few more comments but they are non-blocking.

LGTM.

app/selectors/connectivityController.ts
cryptodev-2s: I don't know why this pattern is being followed in other selectors

app/selectors/connectivityController.ts
cryptodev-2s: I will cleanup these selector in a follow PR, while moving to selector to core (ConnectivityController)

app/selectors/connectivityController.ts
cryptodev-2s: I will cleanup these selector in a follow PR, while moving to selector to core (ConnectivityController)

app/core/Engine/controllers/connectivity/connectivity-controller-init.ts
```diff
@@ -0,0 +1,37 @@
...
+  const connectivityAdapter = new NetInfoConnectivityAdapter();
+
+  const controller = new ConnectivityController({
+    messenger: controllerMessenger,
+    connectivityAdapter,
+  });
+
+  return {
+    controller,
+  };
```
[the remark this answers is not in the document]
mcmire: That's okay, it is not the responsibility of the controller to clean up the adapter.

app/selectors/connectivityController.ts
mcmire: Okay. Seems a bit strange that we need to be so defensive at runtime, when we are using TypeScript to guarantee that data is there.

But sure, we can address this later, anyway.

tommasini (approved):
LGTM

sethkfman (approved):
LGTM

PR #24885, opened 2026-01-19
feat: Add Multichain API support to Snaps

FrederikBolding:
## **Description**

Adds support for the multichain API to Snaps. Snaps can now use the multichain API as long as they specify the `endowment:multichain-provider` in their permissions. To support this, the SnapBridge was expanded to setup a CAIP provider stream. Additionally an E2E test was added for all of this functionality.

This PR also bumps the Snaps packages and brings in a couple bug fixes.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

https://consensyssoftware.atlassian.net/browse/WPC-70
https://consensyssoftware.atlassian.net/browse/WPC-345

## **Manual testing steps**

1. Go to https://metamask.github.io/snaps/test-snaps/latest/
2. Install the multichain provider Snap
3. Check that each action works

---

> [!NOTE]
> Adds Multichain API support for Snaps by extending the Snap bridge and tests.
> 
> - SnapBridge: introduces a `metamask-multichain-provider` CAIP stream with a dedicated JSON-RPC engine gated by `endowment:multichain-provider`; validates multichain requests, wires `wallet_createSession/getSession/invokeMethod/revokeSession`, routes non‑EVM via `MultichainRouter`, blocks unsupported methods, logs requests, and forwards to the correct network client by `networkClientId`.
> - Execution service: initializes bridge with `isWalletConnect: false`.
> - MultichainAccountConnect: excludes Snap IDs from phishing scanning to avoid false positives.
> - Tests: unit test covers multichain RPC handling; adds E2E spec for the Multichain provider (Ethereum, Sepolia, Solana) including chain ID/genesis/accounts and signing flows; updates selectors, Solana confirmation control, mocks genesis hashes, and bumps test-snaps to 3.4.0.
> - Dependencies: bumps Snaps packages (`snaps-sdk`, `snaps-controllers`, `snaps-execution-environments`, `snaps-rpc-methods`, `snaps-utils`) and updates lockfile/allowlist.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d80cb4ad1fdff0d520ecdd7bb8cbca7d8c5b3c02. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -513,7 +514,7 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
     let url = dappUrl || channelIdOrHostname || '';
 
     const checkOrigin = async () => {
-      if (isProductSafetyDappScanningEnabled()) {
+      if (isProductSafetyDappScanningEnabled() && !isSnapId(url)) {
```
FrederikBolding, line 526: Mirror implementation from https://github.com/MetaMask/metamask-mobile/blob/e3951e9751800c1a666fa770c64854d31996bc63/app/components/Views/AccountConnect/AccountConnect.tsx#L333 to prevent erroneous requests

app/core/Engine/controllers/snaps/execution-service-init.ts
```diff
@@ -51,7 +51,7 @@ export const executionServiceInit: ControllerInitFunction<
           fromHomepage: { current: false },
           toggleUrlModal: () => null,
           tabId: false,
-          isWalletConnect: true,
+          isWalletConnect: false,
```
GuillaumeRx, line 54: Why is this set to `false` now ?

app/core/Snaps/SnapBridge.ts
```diff
@@ -166,6 +190,143 @@ export default class SnapBridge {
     return engine;
   }
 
+  #setupProviderEngineCaip() {
```
GuillaumeRx, line 197: missing JSDoc

app/core/Snaps/SnapBridge.ts
```diff
@@ -166,6 +190,143 @@ export default class SnapBridge {
...
+        JsonRpcParams,
+        Json
+      >,
+    );
+
+    engine.push((req, _res, next, end) => {
+      const hasPermission = PermissionController.hasPermission(
+        this.#snapId,
+        SnapEndowments.MultichainProvider,
+      );
```
GuillaumeRx, line 224: Why do we check if the snap has the permission to use the Multichain API here ? Isn't that handled by the snaps platform ? I couldn't find a similar logic for the eip1193 provider 🤔

app/core/Snaps/SnapBridge.ts
```diff
@@ -166,6 +190,143 @@ export default class SnapBridge {
...
+      walletRevokeSession,
+      // @ts-expect-error These types are currently incompatible, but work in practice.
+      walletGetSession,
+      // @ts-expect-error These types are currently incompatible, but work in practice.
+      walletInvokeMethod,
+      // @ts-expect-error These types are currently incompatible, but work in practice.
+      walletCreateSession,
+    ]);
+
+    engine.push(
```
GuillaumeRx, line 252: Crazy that there's no abstraction for that already ? 🤔

app/core/Engine/controllers/snaps/execution-service-init.ts
FrederikBolding: It seems like a mistake that this was ever set to true, I just noticed it during this PR 😓

app/core/Snaps/SnapBridge.ts
FrederikBolding: The multichain API is available via `snap.request` which is given by default, so we need to enforce the permission at the JSON-RPC layer instead. With 1193, the permission grants the `ethereum` endowment, so we don't need to validate at the JSON-RPC layer

app/core/Snaps/SnapBridge.ts
FrederikBolding: What do you mean?

GuillaumeRx: This logic is a dupe of https://github.com/MetaMask/metamask-mobile/blob/6027478eaf4f633193f1f364fba7124f953f4689/app/core/BackgroundBridge/BackgroundBridge.js#L734. It could be easy to extract it in a single middleware that we can import to avoid the duplication ?

app/core/Snaps/SnapBridge.ts
```diff
@@ -166,6 +190,147 @@ export default class SnapBridge {
...
+        ) =>
+          Engine.controllerMessenger.call('MultichainRouter:handleRequest', {
+            ...params,
+            origin: this.#snapId,
+          }),
+        getNonEvmAccountAddresses: Engine.controllerMessenger.call.bind(
+          Engine.controllerMessenger,
+          'MultichainRouter:getSupportedAccounts',
+        ),
+        trackSessionCreatedEvent: undefined,
```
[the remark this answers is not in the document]
FrederikBolding: The hook is used with optional chaining so it won't throw

app/core/Snaps/SnapBridge.ts
FrederikBolding: https://github.com/MetaMask/metamask-mobile/pull/24885/commits/988168044ffeceaffea7027b8f9ae24b4f4f3b40

GuillaumeRx (dismissed):

GuillaumeRx (approved):

ccharly (approved):
LGTM for accounts

e2e/pages/Browser/TestSnaps.ts
```diff
@@ -24,7 +24,7 @@ import { RetryOptions } from '../../../tests/framework';
 import { Json } from '@metamask/utils';
 
 export const TEST_SNAPS_URL =
-  'https://metamask.github.io/snaps/test-snaps/3.1.0/';
+  'https://metamask.github.io/snaps/test-snaps/3.4.0/';
```
seaona, line 27: I'm wondering how is it possible to update the snaps URL and it doesn't need to be added in the allowlist (or it's included in the allowlisted hosts) 🤔 

https://github.com/MetaMask/metamask-mobile/blob/main/tests/api-mocking/mock-e2e-allowlist.ts

(not introduced by this PR,  but this does seem like a bug in the testing setup right? or maybe I'm missing something, as I'm not that familiar with Mobile codebase)

FrederikBolding: Yeah, that does seem like a bug. We don't have a local mock setup for test-snaps yet, so it should be tagged as a live request.

seaona: indeed! thank you, I'll open a ticket for investigating this

seaona (approved):

PR #24935, opened 2026-01-20
feat: integrate Merkl Distributor contract for claimed rewards retrieval

PatrykLucka:
## Description

Fixes an issue where the claimable Merkl rewards amount doesn't update immediately after claiming. The Merkl API has delayed updates, so we now read the claimed amount directly from the blockchain.

### Changes

**Core Fix: On-Chain Reading**
- Added `getClaimedAmountFromContract` to read claimed amount directly from the Merkl Distributor contract instead of relying on the delayed API
- Fixed ABI to properly decode the struct return type `(uint208 amount, uint48 timestamp, bytes32 merkleRoot)`

**UX Improvements**
- **Optimistic updates**: Rewards section hides immediately on claim, then verifies in background
- **Auto-refresh token balance**: After claim confirmation, token balance updates automatically (retries until changed)
- **Loading state**: Shows processing indicator while claim is being verified
- **Transaction handling**: Listens for confirmed/failed/dropped events via TransactionController

## Result

UI provides immediate feedback through optimistic updates, then verifies the claim in the background. Both rewards section and token balance update automatically without manual refresh.

## **Changelog**

CHANGELOG entry: Fixed Merkl rewards claimable amount not updating immediately after claiming by reading from blockchain and implementing optimistic UI updates

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUSD-224

## **Manual testing steps**
```gherkin
Feature: Merkl rewards claimable amount updates immediately after claiming

Scenario: User claims Merkl rewards
  Given user has claimable Merkl rewards displayed
  When user taps claim and approves the transaction
  Then rewards section immediately hides
  And token balance updates automatically after confirmation
  And no manual refresh is required
```

### **After**

https://github.com/user-attachments/assets/77c11bee-a570-41c8-8e20-780b7acb2767

---

> [!NOTE]
> Improves Merkl rewards claiming accuracy and UX by reading claimed amounts directly from chain and adding optimistic updates.
> 
> - Adds `getClaimedAmountFromContract` (Merkl Distributor `claimed(...)`) and correct ABIs; `useMerklRewards` now derives claimable from on-chain claimed, exposes `clearReward`, `refetch`, and `refetchWithRetry`, and tracks `isProcessingClaim`
> - Updates `MerklRewards` to optimistically hide rewards on claim, retry-verify in background, and auto-refresh token balance via RPC with retries; shows processing state in `PendingMerklRewards`
> - Enhances `ClaimMerklRewards` to emit analytics (`MUSD_CLAIM_BONUS_BUTTON_CLICKED`), wire `onClaimSuccess`, and disable/loading button; `useMerklClaim` now subscribes to tx confirmed/failed/dropped events, keeps loading until terminal state, and returns tx hash
> - Extends eligible tokens/addresses, adds feature flag `MM_EARN_MERKL_CAMPAIGN_CLAIMING`, i18n copy for processing, and comprehensive tests across hooks/components
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eebad9338b450e8f41e8d3dbe7872a5c4ed27811. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Earn/components/MerklRewards/ClaimMerklRewards.test.tsx
```diff
@@ -136,7 +200,12 @@ describe('ClaimMerklRewards', () => {
       error: null,
     });
 
-    const { UNSAFE_root } = render(<ClaimMerklRewards asset={mockAsset} />);
+    const { UNSAFE_root } = render(
```
Matt561: Nit: We should avoid using `UNSAFE_` methods in our tests. I don't want to block this PR though if it's not straightforward to fix.

Is it possible to assign the button a testId and use `getByTestId` to target it?

app/components/UI/Earn/components/MerklRewards/ClaimMerklRewards.tsx
```diff
@@ -1,38 +1,70 @@
...
     error: claimError,
   } = useMerklClaim({
     asset,
+    // This callback is triggered when the transaction is confirmed on-chain
+    onTransactionConfirmed: onClaimSuccess,
   });
 
   const handleClaim = async () => {
+    trackEvent(
+      createEventBuilder(MetaMetricsEvents.EARN_LENDING_WITHDRAW_BUTTON_CLICKED)
```
Matt561: `EARN_LENDING_WITHDRAW_BUTTON_CLICKED` for a yield claim could be confusing/unintuitive. If we don't see an existing `MetaMetricsEvents` event that makes sense we could create our own (e.g. `CLAIM_BUTTON_CLICKED`). 

Just a reminder though that if you _do_ end up creating a new event that you'll need to add a new `.yaml` schema to our segment-schema [directory](https://github.com/Consensys/segment-schema/tree/main/libraries/events/metamask-musd). 

Events that aren't in the segement-schema repo and mobile tracking plan risk being filtered out.

app/components/UI/Earn/components/MerklRewards/ClaimMerklRewards.tsx
```diff
@@ -1,38 +1,70 @@
...
+        .addProperties({
+          action_type: 'claim_rewards',
+          token: asset.symbol,
+          chain_id: asset.chainId,
+          network: network?.name,
+          location: 'asset_details',
+          experience: EARN_EXPERIENCES.STABLECOIN_LENDING,
```
Matt561: This is a bit misleading since the mUSD yield claiming isn't part of the lending experience.
```suggestion

```

app/components/UI/Earn/components/MerklRewards/ClaimMerklRewards.tsx
```diff
@@ -1,38 +1,70 @@
...
 
   const handleClaim = async () => {
+    trackEvent(
+      createEventBuilder(MetaMetricsEvents.EARN_LENDING_WITHDRAW_BUTTON_CLICKED)
+        .addProperties({
+          action_type: 'claim_rewards',
+          token: asset.symbol,
+          chain_id: asset.chainId,
+          network: network?.name,
+          location: 'asset_details',
```
Matt561: We can use an existing [constant](https://github.com/MetaMask/metamask-mobile/blob/ee02bf70dc013c8f3749d8d12c6b056e7dd6fe73/app/components/UI/Earn/constants/events/musdEvents.ts#L8) for this value
```suggestion
          location: EVENT_LOCATIONS.ASSET_OVERVIEW,
```

app/components/UI/Earn/components/MerklRewards/MerklRewards.tsx
```diff
@@ -7,38 +9,117 @@ import {
...
   // Fetch claimable rewards data
-  const { claimableReward } = useMerklRewards({
-    asset,
-  });
+  const { claimableReward, isProcessingClaim, clearReward, refetchWithRetry } =
+    useMerklRewards({
+      asset,
+    });
+
+  // Helper to get current balance from Engine state (avoids closure issues)
```
Matt561: Could we use an existing selector or helper for this balance?

I may be misunderstanding the reasoning for this.

app/components/UI/Earn/components/MerklRewards/MerklRewards.tsx
```diff
@@ -7,38 +9,117 @@ import {
...
+
+    // Token address might also need lowercase lookup
+    const tokenAddressLower = asset.address.toLowerCase() as Hex;
+    return (
+      chainBalances[asset.address as Hex] ?? chainBalances[tokenAddressLower]
+    );
+  }, [selectedAddress, asset.address, asset.chainId]);
+
+  // Update navigation params with new balance, retrying until it changes
+  const updateAssetBalanceWithRetry = useCallback(async () => {
```
Matt561, line 43: Instead of short-polling, we could potentially subscribe to asset balance changes.

I see a couple of ways to do this:
1. `useTokenWithBalance` - re-render when the asset balance changes. This seems like it could work but we'd want to make sure that we're not hiding the claim button on any balance change in case the user receives mUSD.
2. `useLatestBalance` - Bypasses potentially stale cached controller state and fetches the latest balance for a given token using RPC `getBalance` call.

What are we trying to solve in `updateAssetBalanceWithRetry`?
- Is the `TokenBalancesController` _not_ updating when we expect it to and we want to force it to update balances?
- Are the balances already automatically updating _but_ we want to detect a balance change event?

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
```diff
@@ -83,20 +105,68 @@ export const useMerklClaim = ({ asset }: UseMerklClaimOptions) => {
...
         origin: 'merkl-claim',
+        type: TransactionType.stakingClaim,
       });
 
-      return result;
+      const { id: transactionId } = transactionMeta;
+
+      // Set up listeners BEFORE awaiting result to avoid race condition
+      // where transaction confirms before listeners are set up
+      Engine.controllerMessenger.subscribeOnceIf(
```
Matt561: Do these subscriptions get cleaned up if the user navigates away before the tx has confirmed/failed/dropped? I think `subscribeOnceIf` returns a cleanup function. It may be worth calling this cleanup on unmount in a useEffect.

app/components/UI/Earn/components/MerklRewards/hooks/useMerklRewards.ts
```diff
@@ -127,29 +168,137 @@ export const useMerklRewards = ({
...
+      if (retryInProgressRef.current) {
+        return;
+      }
+
+      retryInProgressRef.current = true;
+      const maxRetries = options?.maxRetries ?? 5;
+      const delayMs = options?.delayMs ?? 3000;
+
+      try {
+        for (let attempt = 1; attempt <= maxRetries; attempt++) {
```
Matt561, line 266: Instead of short-polling, we could potentially subscribe to asset balance changes.

I see a couple of ways to do this:

useTokenWithBalance - re-render when the asset balance changes. This seems like it could work but we'd want to make sure that we're not hiding the claim button on any balance change in case the user receives mUSD.
useLatestBalance - Bypasses potentially stale cached controller state and fetches the latest balance for a given token using RPC getBalance call.
What are we trying to solve in updateAssetBalanceWithRetry?

Is the TokenBalancesController not updating when we expect it to and we want to force it to update balances?
Are the balances already automatically updating but we want to detect a balance change event?

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
```diff
@@ -83,20 +105,68 @@ export const useMerklClaim = ({ asset }: UseMerklClaimOptions) => {
...
 
-      // Submit transaction
-      const result = await addTransaction(txParams, {
+      // Submit transaction - resolves after user approves in the wallet UI
+      // Use stakingClaim type to get proper toast notifications
+      const { result, transactionMeta } = await addTransaction(txParams, {
         deviceConfirmedOn: WalletDevice.MM_MOBILE,
         networkClientId,
         origin: 'merkl-claim',
+        type: TransactionType.stakingClaim,
```
Matt561: This isn't a `stakingClaim` and will be misleading in both new claim event tracking and existing staking events. If an existing `TransactionType` doesn't meet our use case we can create a new one.

```suggestion
        type: TransactionType.stakingClaim,
```

Matt561 (changes requested):
Left some comments! 👍

app/components/UI/Earn/components/MerklRewards/MerklRewards.tsx
PatrykLucka: Well this approach allows us to use it in the retry loop without worrying about selector updates. I haven't found existing helper function for this. But I think the most interesting thing why we need that retry loop in the first place and I'll answer in the next question below : )

app/components/UI/Earn/components/MerklRewards/MerklRewards.tsx
PatrykLucka: So I actually tried both approaches and was not able to make them work. As you guessed, the TokenBalancesController was not updating when we expected it. Current approach with retries is a bit hacky, but that's the only thing that was working + when we can get the balance with the function defined above, it is quite simple to understand what it does and hot it gets the value and if we decide we want to use the retries, we can't really use hooks to get latest values.

However there is a chance that something that was not working with my first approach using selectors was actually fixed AFTER I implemented the retry function (like using navigation.setParams to update the balance in the UI). We could get back to it, just wanted to release working version quicker.

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
PatrykLucka: Good catch! Thank you!

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
PatrykLucka: yea, good point, let's keep it as TransactionType.contractInteraction for now, there is a separate ticket to add a new transaction type

app/components/UI/Earn/components/MerklRewards/MerklRewards.tsx
PatrykLucka: As per your suggestion I found `fetchEvmAtomicBalance` in the `useLatestBalance` file 🙌 . It simplified the component, thanks! But the retry loop is still there, do you think it's fine to keep it like this?

app/core/Analytics/MetaMetrics.events.ts
```diff
@@ -360,6 +360,7 @@ enum EVENT_NAME {
   EARN_LENDING_DEPOSIT_MORE_BUTTON_CLICKED = 'Earn Lending Deposit More Button Clicked',
   EARN_LENDING_WITHDRAW_BUTTON_CLICKED = 'Earn Lending Withdraw Button Clicked',
   EARN_LENDING_WITHDRAW_CONFIRMATION_BACK_CLICKED = 'Earn Lending Withdraw Confirmation Back Clicked',
+  MUSD_CLAIM_BONUS_BUTTON_CLICKED = 'mUSD Claim Bonus Button Clicked',
```
Matt561, line 363: Nice 🔥 - https://github.com/Consensys/segment-schema/commit/927bc5e866dec3bdc1e46726f8b7c8a0237d530d

Matt561 (approved):

PR #24958, opened 2026-01-20
chore: Add a script for using preview builds more easily

mcmire:
## **Description**

Some MetaMask repos (e.g. `core`) allow upcoming package changes to be tested in Extension and/or Mobile prior to release by publishing "preview builds".

However, using a preview build can be complicated, and specific steps must be followed depending on whether the upcoming changes are breaking or non-breaking.

To make following this workflow easier, this commit adds a package script which can be run to use a preview build. The general usage for this script is:

    yarn use-preview-build <package-name> <preview-version> --type <breaking | non-breaking>

This will automatically update the `resolutions` sections in `package.json` with the appropriate entries given what already exists for the package in `dependencies` and your choice of `type`.

Examples:

- You want to use a preview build for the package `@metamask/permission-controller`; the preview build version is `13.0.2-preview-940c934`; and you anticipate making non-breaking changes. You would say:

      yarn use-preview-build @metamask/permission-controller 13.0.2-preview-940c934 --type breaking

  This would add resolutions which repoint `@metamask/permission-controller` 13.x across the dependency tree to `@metamask-previews/permission-controller@13.0.2-preview-a4747b2b`.

- You want to use a preview build for the package `@metamask/network-controller`; the preview build version is `29.0.0-preview-a4747b2b`; and you anticipate making breaking changes. You would say:

      yarn use-preview-build @metamask/network-controller 29.0.0-preview-a4747b2b --type non-breaking

  This would add a resolution which repoints `@metamask/network-controller` to `@metamask-previews/network-controller@29.0.0-preview-a4747b2b` for whichever version is being used at the root level (it would not apply this resolution across the dependency tree).

## **Changelog**

CHANGELOG entry: null

## **Related issues**

https://consensyssoftware.atlassian.net/browse/WPC-199

## **Manual testing steps**

- Run `yarn use-preview-build @metamask/permission-controller 12.2.0-preview-a4747b2b --type non-breaking`. The script should succeed, and should add 4 resolutions.
- Run `yarn use-preview-build @metamask/permission-controller 12.2.0-preview-a4747b2b --type breaking`. The script should succeed, removing the previously added resolutions, and adding only 1 new resolution.

## **Screenshots/Recordings**

(N/A)

---

> [!NOTE]
> Introduces a helper to apply preview NPM builds via Yarn resolutions.
> 
> - Adds `scripts/use-preview-build.mjs` to update `package.json` `resolutions` for a given MetaMask package and preview version, handling both breaking and non-breaking cases; installs deps after changes
> - Wires script as `yarn use-preview-build` in `package.json`; includes script in `tsconfig.json`
> - Adds required deps (`@yarnpkg/core`, `yargs`, `@types/yargs`) and updates `yarn.lock`; ignores `@types/yargs` in `.depcheckrc.yml`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d527a8dcccbf4edc98e6995945af5f911ee52d20. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/scripts/use-preview-build.mjs
+++ b/scripts/use-preview-build.mjs
@@ -1,0 +1,615 @@
+#!/usr/bin/env node
+
+// This script is copied between Extension and Mobile,
+// and it needs to stay relatively the same between the two.
+/* eslint-disable import/no-nodejs-modules,import/no-namespace,no-console */
+
+import { structUtils } from '@yarnpkg/core';
+import { ESLint } from 'eslint';
+import { execa } from 'execa';
+import * as fs from 'fs';
+import * as path from 'path';
+import { fileURLToPath } from 'url';
+import * as semver from 'semver';
+import yargs from 'yargs';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+/**
+ * The scope that all MetaMask packages are published under.
+ */
+const METAMASK_NPM_SCOPE = 'metamask';
+
+/**
+ * Valid values for the `--type` option.
+ * @typedef {'breaking' | 'non-breaking'} ChangeType
+ */
+
+/**
+ * Captures the arguments passed to this script.
+ * @typedef {Object} CommandLineArguments
+ * @property {string} packageName
+ * @property {string} previewNpmScope
+ * @property {string} previewVersion
+ * @property {ChangeType} type
+ */
+
+/**
+ * The contents of a package's `package.json`.
+ * @typedef {Object} Manifest
+ * @property {string} name
+ * @property {Record<string, string>} [dependencies]
+ * @property {Record<string, string>} [devDependencies]
+ * @property {Record<string, string>} [resolutions]
+ */
+
+/**
+ * The shape of each line returned by `yarn why --json`.
+ * @typedef {Object} YarnWhyEntry
+ * @property {string} value
+ * @property {Record<string, { locator: string; descriptor: string }>} children
+ */
+
+/**
+ * Different interpretations of a SemVer-compatible version range,
+ * e.g. `^1.2.3`.
+ * @typedef {Object} ParsedVersionRange
+ * @property {semver.Range} semverRange
+ * @property {semver.SemVer} primarySemver
+ * @property {number} majorVersion
+ * @property {string} string
+ */
+
+/**
+ * Parses command line arguments using yargs.
+ *
+ * @returns {Promise<CommandLineArguments>} The parsed command line arguments.
+ */
+async function parseArgs() {
+  const argv = await yargs(process.argv.slice(2))
+    .usage(
+      'Usage: $0 <package-name> <preview-version> --type <breaking|non-breaking> [options]',
+    )
+    .positional('package-name', {
+      describe: 'The name of the package',
+      type: 'string',
+    })
+    .option('preview-npm-scope', {
+      describe: 'NPM organization for preview builds',
+      type: 'string',
+      default: 'metamask-previews',
+    })
+    .positional('preview-version', {
+      describe: 'The preview version string (e.g., "1.1.4-preview-e2df9b4")',
+      type: 'string',
+    })
+    .option('type', {
+      alias: 't',
+      describe: 'Type of change',
+      choices: ['breaking', 'non-breaking'],
+      demandOption: true,
+    })
+    .example(
+      '$0 @metamask/controller-utils 1.1.4-preview-e2df9b4 --type non-breaking',
+      'Add resolutions for non-breaking changes',
+    )
+    .example(
+      '$0 @metamask/network-controller 12.4.9-preview-e2df9b4 --type breaking',
+      'Add resolution for breaking changes',
+    )
+    .check((args) => {
+      const positionals = args._;
+      if (positionals.length < 2) {
+        throw new Error(
+          'Both <package-name> and <preview-version> are required',
+        );
+      }
+      return true;
+    })
+    .string('_')
+    .help().argv;
+
+  const [uncorrectedPackageName, previewVersion] = argv._;
+
+  const packageName = uncorrectedPackageName.startsWith(
+    `@${METAMASK_NPM_SCOPE}/`,
+  )
+    ? uncorrectedPackageName
+    : `@${METAMASK_NPM_SCOPE}/${uncorrectedPackageName}`;
+  const previewNpmScope = argv.previewNpmScope.startsWith('@')
+    ? argv.previewNpmScope.slice(1)
+    : argv.previewNpmScope;
+
+  return {
+    packageName,
+    previewNpmScope,
+    previewVersion,
+    type: argv.type,
+  };
+}
+
+/**
+ * Reads and parses a `package.json` file.
+ *
+ * @param {string} manifestPath - Path to the `package.json` file.
+ * @returns {Promise<Manifest>} The parsed `package.json` contents.
+ */
+async function readManifest(manifestPath) {
+  const content = await fs.promises.readFile(manifestPath, 'utf-8');
+  return JSON.parse(content);
+}
+
+/**
+ * Writes a `package.json` file.
+ *
+ * @param {string} manifestPath - Path to the `package.json` file.
+ * @param {Manifest} contents - The package.json contents to write.
+ * @returns {Promise<void>}
+ */
+async function writeManifest(manifestPath, contents) {
+  await fs.promises.writeFile(
+    manifestPath,
+    `${JSON.stringify(contents, null, 2)}\n`,
+    'utf-8',
+  );
+}
+
+/**
+ * Formats a `package.json` file.
+ *
+ * @param {string} manifestPath - Path to the `package.json` file.
+ * @returns {Promise<void>}
+ */
+async function formatManifest(manifestPath) {
+  const eslint = new ESLint({
+    // This is necessary in order to run this script outside of this project,
+    // e.g., for testing purposes.
+    cwd: path.resolve(__dirname, '..'),
+    ignore: false,
+    fix: true,
+  });
+  console.log('Formatting package.json...');
+  const results = await eslint.lintFiles(manifestPath);
+  await ESLint.outputFixes(results);
+}
+
+/**
+ * Saves a formatted version of the given `package.json` with the given
+ * resolutions updated.
+ *
+ * @param {string} manifestPath - The path to `package.json`.
+ * @param {Record<string, string>} resolutions - The new resolutions to load into `package.json`.
+ * @returns {Promise<void>}
+ */
+async function updateResolutionsInManifest(manifestPath, resolutions) {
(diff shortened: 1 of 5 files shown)


scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+  }
+
+  await writeManifest(manifestPath, {
+    ...manifest,
+    resolutions: {
+      ...manifest.resolutions,
+      ...resolutions,
+    },
+  });
+  // await formatManifest(manifestPath);
```
Mrtenz, line 199: Should this be uncommented?


```suggestion
  await formatManifest(manifestPath);
```

scripts/use-preview-build.mjs
Mrtenz, line 1: Wondering if this should be a Yarn plugin, since we get easier access to the Yarn API and dependencies then, rather than needing `execa`.

mcmire: @Gudahtt suggested that at first. And, maybe this is better in the long run. But I looked into this briefly and wasn't sure how exactly to do this. Yarn does provide a [`reduceDependency`](https://github.com/yarnpkg/berry/blob/60b83ace11bd36270c1ed117319139eaa6971db3/packages/yarnpkg-core/sources/Plugin.ts#L116C3-L116C19) hook, which is how the `resolutions` field works, but it seems to be useful only if you want to replace one dependency with another. In the case of the non-breaking workflow, I want to replace one dependency with multiple (e.g. if `package.json` contains `@metamask/transaction-controller@^10.0.0` then it would be as if it contained `"@metamask/transaction-controller@^10.0.0`": "@metamask-previews/transaction-controller@10.0.0-preview-abcd123"`, `"@metamask/transaction-controller@^10.1.0`": "@metamask-previews/transaction-controller@10.0.0-preview-abcd123"`, `"@metamask/transaction-controller@^10.2.0`": "@metamask-previews/transaction-controller@10.0.0-preview-abcd123"`). I'd have to do more research on this, and didn't know if I wanted to spend that time right now.

Mrtenz: Fair enough. I didn't dive into it much more to see if it was possible either.

Gudahtt: Yarn plugins can be independent commands as well, which still grants access to internal APIs. And that would make it easier to share the same script between clients.

The tutorial explains how to do this pretty well: https://yarnpkg.com/advanced/plugin-tutorial#adding-commands

Yarn uses clipanion rather than yargs, but otherwise it might be relatively easy to port this.

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+    !(
+      semverRange.set.length === 1 &&
+      semverRange.set[0] !== undefined &&
+      ((semverRange.set[0].length === 2 &&
+        semverRange.set[0][0] !== undefined &&
+        semverRange.set[0][1] !== undefined &&
+        semverRange.set[0][0].operator === '>=' &&
+        semverRange.set[0][1].operator === '<' &&
+        semverRange.set[0][1].semver.major ===
+          semverRange.set[0][0].semver.major + 1) ||
```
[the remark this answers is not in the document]
mcmire: Hmm, this is true, I will fix this.

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+ *
+ * @param {string} manifestPath - The path to `package.json`.
+ * @param {Record<string, string>} resolutions - The new resolutions to load into `package.json`.
+ * @returns {Promise<void>}
+ */
+async function updateResolutionsInManifest(manifestPath, resolutions) {
+  console.log('Adding resolutions:');
+  const manifest = await readManifest(manifestPath);
+  for (const [key, value] of Object.entries(resolutions)) {
+    console.log(`  "${key}": "${value}"`);
```
Gudahtt, line 189: Was this log left in intentionally?

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+  );
+}
+
+/**
+ * Formats a `package.json` file.
+ *
+ * @param {string} manifestPath - Path to the `package.json` file.
+ * @returns {Promise<void>}
+ */
+async function formatManifest(manifestPath) {
```
Gudahtt, line 164: Interesting idea, but for something as simple as adding resolutions, is it likely we'd run into any formatting errors?

I would assume we're preserving key order when we read/write the manifest, which is most commonly how `package.json` changes conflict with lint rules. I'm not sure what other lint violations are possible here.

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+ * Verifies that the the given version range is SemVer-compatible and that it is
+ * a version range that we can handle, and then parses the major version from
+ * the version range.
+ *
+ * Currently, this is the list of supported version ranges:
+ *
+ * - "18.0.0"
+ * - "=18.0.0"
+ * - ">=18.0.0,<19.0.0"
+ * - "^18.0.0"
```
Gudahtt, line 213: It says here that caret ranges are supported, but it's not obvious to me that it's true from reading the massive conditional below.

Maybe we can break that conditional up for readability? And ensure that caret ranges work.

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+ * Currently does not support patched dependencies.
+ *
+ * @param {string} rawDescriptor - The string to parse.
+ * @param {string} dependencyName - The name of the dependency that `rawDescriptor`
+ * represents.
+ * @returns {ParsedVersionRange} A parsed representation of the version range within `rawDescriptor`.
+ * @throws {Error} If `rawDescriptor` represents a patched dependency.
+ */
+function getParsedVersionRangeForDescriptor(rawDescriptor, dependencyName) {
+  validateNonPatchDescriptor(rawDescriptor, dependencyName);
```
Gudahtt, line 315: Is this the only reason `dependencyName` is needed here?

Isn't this validation redundant, given that we ensure the package isn't patched already in `main`?

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+
+  console.log('Removing resolutions for existing preview builds:');
+  for (const [key, value] of Object.entries(resolutionsToRemove)) {
+    console.log(`  "${key}": "${value}"`);
+  }
+
+  const newResolutions = Object.keys(resolutionsToRemove).reduce(
+    (workingResolutions, resolutionName) => {
+      if (resolutionName in workingResolutions) {
+        const clone = { ...workingResolutions };
```
Gudahtt, line 415: Nit: It would be more efficient to clone once (when setting the initial accumulator, on line 421 at the moment) rather than every loop

scripts/use-preview-build.mjs
mcmire: No, this one is intentional. This prints the resolutions that were added (they could be different depending on the `--type` provided).

scripts/use-preview-build.mjs
Gudahtt: Though getting rid of `execa` might be challenging. It looks like we're currently writing the manifest and re-installing dependencies as a way of updating the lockfile before we plan the new preview resolutions.

We'd need to learn a bit more about internal Yarn APIs to do that with fewer disk writes and no CLI calls.

Maybe something to consider later.

mcmire: Okay, I'll look into this further.

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+      return workingResolutions;
+    },
+    resolutions,
+  );
+  await writeManifest(manifestPath, {
+    ...manifest,
+    resolutions: newResolutions,
+  });
+  // await formatManifest(manifestPath);
+  await installDependencies();
```
Gudahtt, line 428: Nit: Probably should mention in the function description that it also installs dependencies (presumably to update the lockfile?)

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+  manifestPath,
+}) {
+  const manifest = await readManifest(manifestPath);
+  const rootPackageName = manifest.name;
+  const parsedVersionRange = getVersionRangeForDependency(
+    manifest,
+    packageName,
+  );
+
+  const resolutionKey = `${rootPackageName}@workspace:./${packageName}@${parsedVersionRange.string}`;
```
Gudahtt, line 508: This looks monorepo-specific. Does this work in mobile/extension?

scripts/use-preview-build.mjs
```diff
@@ -0,0 +1,615 @@
...
+  });
+  // await formatManifest(manifestPath);
+  await installDependencies();
+
+  console.log('');
+}
+
+/**
+ * Uses `yarn why` to get all version ranges of the given package that are being
+ * used across the dependency tree and which are compatible with the given major
```
Gudahtt, line 435: This assumes that we're using `^` ranges.

Probably an OK assumption to make? Though we could be a bit more precise by checking the range.

If we're going to assume this, we should at least state the assumption here in this comment. Maybe also in the usage docs or something.

scripts/use-preview-build.mjs
Gudahtt: Oh I see, sorry I missed the "Adding resolutions:" log above it somehow

scripts/use-preview-build.mjs
mcmire: I've made a new PR that creates a Yarn plugin instead: https://github.com/MetaMask/metamask-mobile/pull/25086

I will close this PR if that one seems like a better idea.

scripts/use-preview-build.mjs
mcmire: The new PR I made to add a Yarn plugin rather than a script supports more version ranges: https://github.com/MetaMask/metamask-mobile/pull/25086

scripts/use-preview-build.mjs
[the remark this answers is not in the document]
mcmire: The new PR I made to add a Yarn plugin handles this: https://github.com/MetaMask/metamask-mobile/pull/25086

scripts/use-preview-build.mjs
mcmire: The new PR I made to add a Yarn plugin rather than a script supports more version ranges: https://github.com/MetaMask/metamask-mobile/pull/25086

scripts/use-preview-build.mjs
mcmire: It does! Even if you don't have multiple workspaces, Yarn still uses the concept of workspaces internally, and you can target a dependency declared in `package.json` by prepending the package name + version range with `<root-package>@workspace:./`.

mcmire:
Oh right, this was superseded by #25086.

PR #24964, opened 2026-01-21
feat(perps): allow users to trade with any token PoC

michalconsensys:
> [!NOTE]
> This PR is a PoC of the trade with any token feature in Perps. The intention in this PR is to allow the product team to test the feature while it's still hidden for users. Further improvements to this feature will be made according to the technical specification here https://consensyssoftware.atlassian.net/browse/TAT-1030. This includes alerts, UI changes, tooltips, etc.

> [!NOTE]
>  Known issues:
> - When pay with token is used, and the user leaves the screen, the toast notification may stay on the screen, the trade may not be executed, the designs are not matching figma, and it does not offer to cancel the trade if the deposit takes too long https://consensyssoftware.atlassian.net/browse/TAT-1030
> - When pay with token is used, and the user does not have enough founds to cover the swap/trade, there is no warning shown or when the quotas are being fetched, we are not disabling the button to execute the trade https://consensyssoftware.atlassian.net/browse/TAT-2429
> - When user presses down on the asset menu to choose an asset to pay with, then the user can not revert back to perps balance https://consensyssoftware.atlassian.net/browse/TAT-2429
> - The designs do not match the specification https://consensyssoftware.atlassian.net/browse/TAT-1030
> - The feature is missing analytics https://consensyssoftware.atlassian.net/browse/TAT-2442
> - Increasing/removing margin is not supported https://consensyssoftware.atlassian.net/browse/TAT-2263 
> - Improved error handling, multiple button clicks, and race conditions https://consensyssoftware.atlassian.net/browse/TAT-1030

## **Description**

This PR implements a one-click trading feature for Perps that allows users to deposit funds and execute trades seamlessly within the order view. Previously, users had to navigate away from the order view to deposit funds, then return to place their order. 

## **Changelog**

CHANGELOG entry: Added one-click trading for Perps, allowing users to deposit funds and execute trades seamlessly within the order view

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-2180

## **Manual testing steps**

```gherkin
Feature: Perps One-Click Trading

  Scenario: User deposits funds and executes trade in one flow
    Given user is on the Perps Order View with a configured order
    And the "Trade With Any Token" feature flag is enabled
    And user has insufficient USDC balance on Hyperliquid

```

### **After**

https://github.com/user-attachments/assets/7ebeddc9-27b2-41cc-9d86-d3fb21eeeb36

---

> [!NOTE]
> **Medium Risk**
> Touches Perps order placement, transaction creation/confirmation routing, and toast/status tracking across multiple hooks and UI layers; regressions could affect deposit confirmation, navigation, or duplicate/missing toasts.
> 
> **Overview**
> **Adds a new Perps one-click “deposit + place order” flow (PoC).** Introduces `TransactionType.perpsDepositAndOrder` via a patched `@metamask/transaction-controller`, wires it through confirmations/metrics/activity, and updates Perps deposit progress/toast listeners to treat it like a deposit.
> 
> **Perps order UI can now drive the deposit confirmation and then auto-place the trade.** `PerpsOrderView` integrates confirmation/pay hooks, shows a new `PerpsPayRow` + `PerpsDepositFees` when the feature flag is enabled and a custom token is selected, prefills deposit amount from margin, and uses a new `usePerpsOrderDepositTracking` + `PerpsStreamManager` “active deposit handler” guard to avoid duplicate toasts.
> 
> **Controller/navigation changes to support the combined flow.** `PerpsController.depositWithConfirmation` gains a `depositAndPlaceOrder` path that uses `addTransaction` (no immediate navigation) and a convenience `depositWithOrder`; `usePerpsNavigation` now starts `depositWithOrder()` before navigating to redesigned confirmations.
> 
> **Related robustness fixes.** `usePerpsPaymentTokens` now returns all tokens (no min-balance filter) and excludes Hyperliquid chain IDs to avoid icon errors; Bridge token icon resolution is wrapped in try/catch for unsupported chains; confirmations add `disableSafeArea`/styling support and hide the network filter in Pay With modal for `perpsDepositAndOrder`; e2e selector updated to use `confirm-button`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 54257e55d59f1d7fb0569f808f7496231bef08af. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,339 @@
...
+  if (hasTransactionType(transaction, [TransactionType.predictWithdraw])) {
+    return strings('confirm.deposit_edit_amount_predict_withdraw');
+  }
+
+  if (hasTransactionType(transaction, [TransactionType.musdConversion])) {
+    return strings('earn.musd_conversion.convert');
+  }
+
+  return strings('confirm.deposit_edit_amount_done');
+}
```
aganglada: These can be separate from the view, so we follow the single responsibility file

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,339 @@
+import { getNativeTokenAddress } from '@metamask/assets-controllers';
```
aganglada: this file in general is a bit congested hand hard to follow I would separate

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
michalconsensys: This is now refactored 👍

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
michalconsensys: That's now refactored

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,254 @@
...
+    return () => {
+      Engine.controllerMessenger.unsubscribe(
+        'TransactionController:transactionStatusUpdated',
+        handleTransactionStatusUpdated,
+      );
+      Engine.controllerMessenger.unsubscribe(
+        'TransactionController:transactionFailed',
+        handleTransactionFailed,
+      );
+    };
```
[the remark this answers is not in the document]
michalconsensys: Fixed in https://github.com/MetaMask/metamask-mobile/pull/24964/commits/f11c6602167eca313c951738c32319b2aeaec011

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,295 @@
...
+        primaryRequiredToken?.address.toLowerCase() &&
+      token.chainId === primaryRequiredToken?.chainId,
+  );
+
+  const assetId = toCaipAssetType(
+    'eip155',
+    Number(primaryRequiredToken?.chainId ?? '0x0').toString(),
+    'erc20',
+    asset?.assetId ?? '0x0',
+  );
```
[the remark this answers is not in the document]
michalconsensys: This is the same as in custom-amount-info.tsx

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -1243,6 +1307,21 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
             />
           </View>
 
+          {isTradeWithAnyTokenEnabled &&
+            depositAmount &&
+            depositAmount.trim() !== '' &&
+            activeTransactionMeta && (
```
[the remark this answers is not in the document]
michalconsensys: We create the TX in the parent component, so this would be redundant

infiniteflower (dismissed):

app/components/UI/Bridge/hooks/useTokensWithBalance/index.ts
```diff
@@ -219,17 +219,27 @@ export const useTokensWithBalance: ({
         const evmTokenFiatAmount = evmBalances?.[i]?.tokenFiatAmount;
         const nonEvmTokenFiatAmount = Number(token.balanceFiat);
 
+        // Safely get token icon URL - formatAddressToAssetId may throw for unsupported chains
+        let tokenImage = token.image;
+        try {
+          const assetId = formatAddressToAssetId(token.address, chainId);
+          if (assetId) {
+            tokenImage =
+              getTokenIconUrl(assetId, isNonEvmChainId(chainId)) || token.image;
```
gambinish, line 255: I think one limitation here will be that the tokenImage that we host from the tokens API may be different than the icon for the hyperliquid asset pair.

This may be unavoidable, and might actually be preferable in certain cases.

app/components/UI/Perps/hooks/usePerpsDepositStatus.ts
```diff
@@ -16,6 +16,24 @@ import { usePerpsLiveAccount } from './stream/usePerpsLiveAccount';
...
+  skipDefaultToastTransactionIds.add(transactionId);
+};
+
+/**
+ * Unmark a transaction ID (cleanup)
+ */
+export const unmarkTransactionSkipDefaultToast = (transactionId: string) => {
+  skipDefaultToastTransactionIds.delete(transactionId);
+};
+
```
gambinish: Can you help me understand why these are needed?

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,263 @@
...
+ * 3. Showing success toast when funds arrive
+ * 4. Handling transaction failures
+ * 5. Executing the order after funds arrive
+ *
+ * This ensures the order is placed automatically after the deposit completes.
+ */
+export const usePerpsOrderDepositTracking = ({
+  onDepositComplete,
+  activeTransactionMeta,
+}: UsePerpsOrderDepositTrackingParams) => {
```
gambinish: I noticed this hook shares some patterns with `usePerpsDepositProgress`—both subscribe to `TransactionController:transactionStatusUpdated` and monitor `availableBalance` changes.

Given that we'll be supporting both the legacy deposit flow and the new "pay with any token" flow, having a dedicated hook for the order flow makes sense.

Would it be worth extracting the shared deposit detection logic (transaction subscription + balance monitoring) into a base hook that both can consume?

This could reduce duplication while keeping their distinct responsibilities:

`usePerpsDepositProgress` → returns state for UI components
`usePerpsOrderDepositTracking` → orchestrates toasts + callback execution

Not a blocker—just wanted to get your thoughts on whether that refactor would be valuable here or better suited for a follow-up.

Edit: this seems related to this comment here: https://github.com/MetaMask/metamask-mobile/pull/24964#discussion_r2718611813

app/components/UI/Perps/hooks/usePerpsDepositStatus.ts
gambinish: Ahh. I see that it's to dedupe the deposit tracking between `usePerpsDepositStatus` and `usePerpsOrderDepositTracking`. My initial instinct is that having this "module specific state" might cause hard to troubleshoot bugs.

Would a Context-based approach be cleaner? A `PerpsDepositTrackingProvider` could own the deposit state and allow consumers to register for toast ownership explicitly. This would:
- Centralize the subscription/balance monitoring logic
- Make toast ownership explicit and debuggable
- Eliminate the need for hidden coordination via a Set

We _would_ need to introduce a new `Provider`, but imo that might be worth the tradeoff

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -197,6 +220,14 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
   const { track } = usePerpsEventTracking();
   const { openTooltipModal } = useTooltipModal();
 
+  // Feature flag for trade with any token
+  const isTradeWithAnyTokenEnabled = useSelector(
+    selectPerpsTradeWithAnyTokenEnabledFlag,
+  );
+
+  // Check if there's an active transaction
+  const activeTransactionMeta = useTransactionMetadataRequest();
```
gambinish, line 254: I think that this might be causing infinite re-renders, since `activeTransactionMeta` is an object from a Redux selector.

If the selector returns a new object reference on each render (even with the same data), the effect re-runs continuously. I was able to reproduce this by logging this variable in the component's body.

There's a guard for showing toasts:
```
if (hasShownDepositToastRef.current !== activeTransactionMeta.id) {  
  showToast(...);
}
```

But the effect itself still re-runs on every `activeTransactionMeta` change.

Each re-run:
- Unsubscribes from TransactionController
- Re-subscribes to TransactionController

This causes performance "thrashing" even if no toast is shown.

I threw this into Claude and it proposed three possible fixes here:

Option 1: Only depend on the transaction ID, not the whole object
Option 2: Memoize the transaction meta reference
Option 3: Use a ref instead of dependency

Basically the root cause is that React is treating an object as a dependency when you really only care about specific primitive properties of that object.

gambinish (dismissed):
I pulled this down and functionally it works well! I opened a couple of different positions for `ETH` and `HYPE` by paying with `arb.usdc` and `bnb`

I left a couple of comments related to a potential refactor to consolidate some of our deposit tracking logic. I'm happy to discuss this further in Slack. I'm interested in what you think about those thoughts.

Additionally, I just wanted to flag that the designs for this just got finalized today. Were you planning on revisiting the designs in a follow up PR? I can see the value in getting this into Product's hands to test while we iterate on designs.

Lastly, it would be good to have higher unit test coverage here before merging, especially since this is a high visibility feature.

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
michalconsensys: Resolved in https://github.com/MetaMask/metamask-mobile/pull/24964/commits/cfac13570e981230203ce6573ecc8202e20e4d41

app/components/UI/Perps/hooks/usePerpsDepositStatus.ts
michalconsensys: Hello, 100% agree with you that this approach is not ideal. This is just a POC to get it to hands of the product team, and then the idea is to iterate and improve it

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
michalconsensys: Replied here https://github.com/MetaMask/metamask-mobile/pull/24964#discussion_r2719594089

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,295 @@
...
+      token.address?.toLowerCase() ===
+        primaryRequiredToken?.address.toLowerCase() &&
+      token.chainId === primaryRequiredToken?.chainId,
+  );
+
+  const assetId = toCaipAssetType(
+    'eip155',
+    Number(primaryRequiredToken?.chainId ?? '0x0').toString(),
+    'erc20',
+    asset?.assetId ?? '0x0',
```
abretonc7s: We need to avoid the default values! if not set it should throw

app/components/UI/Perps/hooks/usePerpsDepositStatus.ts
abretonc7s: Good catch @gambinish .                                                   
                                                                                                                                                              
I'd recommend extending `PerpsStreamManager` rather than adding a new provider:                                           

**Why not a new provider:**
- Current nesting: `PerpsConnectionProvider → PerpsStreamProvider → PerpsOrderProvider`
- Adding a 4th provider creates more re-render boundaries for state that updates frequently (tx status, 1s balance polling)

**Proposed: Simple coordination flag in `PerpsStreamManager` singleton:**

```typescript
// PerpsStreamManager - UI coordination only
private activeDepositHandler = false;

public setActiveDepositHandler(isActive: boolean): void
public hasActiveDepositHandler(): boolean
```

Note: Deposit details (amount, currency, status) already live in `PerpsController` state (`depositInProgress`, `lastDepositTransactionId`, `depositRequests[]`), so we don't duplicate that here - this is purely for UI toast coordination.

**Flow:**
1. Component managing deposit UI sets `setActiveDepositHandler(true)`
2. Global `usePerpsDepositStatus` checks `hasActiveDepositHandler()` → skips its toast
3. Component unmounts → cleanup sets `false`
4. Global monitor takes over if deposit still in progress

This handles the 60+ second deposit→trade flows gracefully - if user navigates away mid-deposit, the global monitor picks up and shows completion feedback.

**Layer responsibilities:**
- **PerpsController**: Business logic & data (deposit amount, currency, status, transaction history)
- **PerpsStreamManager**: UI coordination (WebSocket streams, toast ownership handoff)
- **Hooks/Components**: Consume both, render UI

This keeps the coordination concern (who shows the toast?) separate from the business data (what deposit is happening?).

**Benefits:**
- ✅ Debuggable via `getStreamManagerInstance().hasActiveDepositHandler()`
- ✅ No duplication - deposit details stay in controller
- ✅ Clear separation of concerns between layers
- ✅ Graceful handoff on navigation
- ✅ No additional provider nesting
- ✅ Just a boolean - simple mental model

abretonc7s (changes requested):

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,295 @@
...
+  disablePay?: boolean;
+  hasMax?: boolean;
+  preferredToken?: SetPayTokenRequest;
+  overrideContent?: (amountHuman: string) => ReactNode;
+  defaultValue?: string;
+  skipNavigation?: boolean;
+  onConfirmCallback?: (transactionMeta: TransactionMeta) => void;
+}
+
+export const PerpsInlineDeposit: React.FC<CustomAmountInfoProps> = memo(
```
matthewwalsh0: I'm afraid this is the inverse of the architecture I proposed.

What we seem to be doing here is rendering a partial confirmation inside the existing order component, however the core of the confirmation architecture across both clients is rendering a single generic confirmation component, inside which we change the contents based on the transaction type.

This enables us to maintain a consistent shape and style for all confirmations, but support modularity and custom rows per confirmation and use case.

This is also a major duplication of the `CustomAmountInfo` component that is already intended to be generic and extensible with custom components as needed.

Ideally we'd make a fresh info component such as `PerpsDepositTradeInfo` and route to it inside `InfoRoot` via a new transaction type.

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
matthewwalsh0, line 1: For visibility, we keep all the info components inside `app/components/Views/confirmations`.

But you can still own the file via a `external/perps` directory.

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,295 @@
...
+export interface CustomAmountInfoProps {
+  children?: ReactNode;
+  currency?: string;
+  disablePay?: boolean;
+  hasMax?: boolean;
+  preferredToken?: SetPayTokenRequest;
+  overrideContent?: (amountHuman: string) => ReactNode;
+  defaultValue?: string;
+  skipNavigation?: boolean;
+  onConfirmCallback?: (transactionMeta: TransactionMeta) => void;
```
matthewwalsh0: We shouldn't need to couple the resulting logic via props, since we can instead soley rely on messenger events from the `TransactionController` to detect the deposit transaction is `approved`, `submitted`, `confirmed` etc.

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
```diff
@@ -0,0 +1,295 @@
...
+      updateTokenAmount();
+      EngineService.flushState();
+      setIsKeyboardVisible(false);
+    }, [updateTokenAmount]);
+
+    return (
+      <Box>
+        <Box>
+          {children}
+          {disablePay !== true && hasTokens && <PayWithRow hideNetworkFilter />}
```
matthewwalsh0: Why don't we want the network filter?

The intent was to provide a consistent asset selection experience for all MetaMask Pay usages?

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
matthewwalsh0, line 1: Unit tests?

matthewwalsh0 (changes requested):

app/components/UI/Perps/Views/PerpsOrderView/PerpsInlineDeposit.tsx
michalconsensys: This was a requirement raised on Slack here https://consensys.slack.com/archives/C092T3GPHQD/p1769091779669089?thread_ts=1769087995.104139&cid=C092T3GPHQD

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
```diff
@@ -0,0 +1,285 @@
...
+        onCustomTokenSelected?.();
+      }
+    }
+  }, [payToken, onCustomTokenSelected]);
+
+  const handleClick = useCallback(() => {
+    if (!canEdit) return;
+    // Mark that user has interacted when they open the modal
+    setHasUserInteracted(true);
+    onCustomTokenSelected?.();
```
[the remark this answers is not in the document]
michalconsensys: This is intentional for the POC

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
```diff
@@ -0,0 +1,285 @@
...
+    setConfirmationMetric({
+      properties: {
+        mm_pay_token_list_opened: true,
+      },
+    });
+    navigation.navigate(Routes.CONFIRMATION_PAY_WITH_MODAL, {
+      hideNetworkFilter: true,
+    });
+  }, [canEdit, navigation, setConfirmationMetric, onCustomTokenSelected]);
```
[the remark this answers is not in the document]
michalconsensys: This is intentional for the POC

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -1551,6 +1770,31 @@ const PerpsOrderView: React.FC = () => {
...
+    isTradeWithAnyTokenEnabled,
+    activeTransactionMeta,
+    depositWithConfirmation,
+  ]);
+
+  // Don't render until transaction exists (if trade with any token is enabled)
+  // This prevents blank screen while transaction is being created
+  if (isTradeWithAnyTokenEnabled && !activeTransactionMeta) {
+    return null;
+  }
```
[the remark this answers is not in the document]
michalconsensys: Fixed in 694717894ac587f4d89db0c2fc24de63624d62b2

app/components/UI/Perps/hooks/usePerpsDepositStatus.test.ts
```diff
@@ -411,6 +421,25 @@ describe('usePerpsDepositStatus', () => {
         mockPerpsToastOptions.accountManagement.deposit.inProgress,
       ).toHaveBeenCalledWith(60, 'test-tx-id'); // 60 seconds for other tokens
     });
+
+    it('should skip showing toast when active deposit handler exists', () => {
```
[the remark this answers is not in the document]
michalconsensys: Resolved here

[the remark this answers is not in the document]
michalconsensys: f59e38a54a04e21939a6e68f6988074cb8792ff4

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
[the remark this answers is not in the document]
michalconsensys: This is intentional for the PoC until the modal is changed by the confirmations team

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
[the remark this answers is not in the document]
michalconsensys: Resolved in 694717894ac587f4d89db0c2fc24de63624d62b2

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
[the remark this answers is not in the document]
michalconsensys: this is intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -913,6 +1049,74 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
+    handlePlaceOrderRef.current = handlePlaceOrder;
+  }, [handlePlaceOrder]);
+
+  // Setup transaction custom amount for deposit
+  // Note: This hook requires activeTransactionMeta to exist, otherwise it will crash
+  // We create the transaction in useEffect below when depositAmount is set
+
+  const { updatePendingAmount, amountHuman } = useTransactionCustomAmount({
+    currency: PERPS_CURRENCY,
+  });
```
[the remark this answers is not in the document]
michalconsensys: The TX is created in the parent component

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -1551,6 +1777,89 @@ const PerpsOrderView: React.FC = () => {
...
+  }
+
+  // Don't render until transaction exists (if trade with any token is enabled)
+  // This prevents blank screen while transaction is being created
+  // Note: This should not happen in normal flow, but kept as fallback
+  if (!activeTransactionMeta) {
+    return (
+      <PerpsLoader message={strings('perps.order.initializing')} fullScreen />
+    );
+  }
```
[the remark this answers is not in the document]
michalconsensys: This is intentional, not to cause a crash in the child component

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -1551,6 +1777,89 @@ const PerpsOrderView: React.FC = () => {
...
+          // eslint-disable-next-line no-console
+          console.error('Failed to create deposit transaction:', error);
+        });
+    }
+  }, [
+    activeTransactionMeta,
+    depositWithConfirmation,
+    isCreatingTransaction,
+    transactionCreationError,
+  ]);
```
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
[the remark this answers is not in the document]
michalconsensys: This is intentional for the PoC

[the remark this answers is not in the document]
michalconsensys: Intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -713,10 +799,60 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
+      if (!activeTransactionMeta) {
+        console.error('No active transaction to confirm');
+        return;
+      }
+
+      // Show deposit toast and set up tracking before confirming
+      handleDepositConfirm(activeTransactionMeta);
+
+      await onDepositConfirm();
+      return;
```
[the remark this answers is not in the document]
michalconsensys: This is just a POC, will be handled later

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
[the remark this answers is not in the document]
michalconsensys: This is intentional to prevent children from throwing an error

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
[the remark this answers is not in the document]
gambinish: If this is something we want to implement after the PR gets merged, it probably makes sense to add a TODO comment so that we don't lose track of it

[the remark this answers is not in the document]
michalconsensys: List of known issues is in the PR title

app/components/UI/Perps/hooks/usePerpsPaymentTokens.test.ts
```diff
@@ -189,26 +189,26 @@ describe('usePerpsPaymentTokens', () => {
   });
 
   describe('Token filtering by minimum balance', () => {
-    it('should filter tokens by minimum order amount on mainnet', () => {
+    it('should show all tokens regardless of balance on mainnet', () => {
```
abretonc7s: no should in test names

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -713,10 +799,49 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
```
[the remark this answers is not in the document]
michalconsensys: This issue is raised in the PR description, and will be resolved later

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -713,10 +799,49 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
+    }
```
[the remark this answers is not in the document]
michalconsensys: This issue is raised in the PR description, and will be resolved later

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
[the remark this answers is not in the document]
michalconsensys: This is intentional for the PoC, and the issue is raised in the PR description, and will be resolved later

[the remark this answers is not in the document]
michalconsensys: This is intentional for the PoC, and the issue is raised in the PR description, and will be resolved later

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
```diff
@@ -16,6 +17,10 @@ import { hasTransactionType } from '../../../utils/transaction';
 import { useMusdConversionTokens } from '../../../../../UI/Earn/hooks/useMusdConversionTokens';
 
 export function PayWithModal() {
+  const route = useRoute();
+  const hideNetworkFilter =
+    (route.params as { hideNetworkFilter?: boolean })?.hideNetworkFilter ??
```
abretonc7s: do we need to cast here? shouldnt this by typed in the route declaration?

ios/MetaMask.xcodeproj/project.pbxproj
```diff
@@ -1281,7 +1281,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 3418;
```
abretonc7s, line 1284: is this supposed to be committed?

abretonc7s (changes requested):
Small changes, I think the routes should be typed correctly and not change the xcode config

ios/MetaMask.xcodeproj/project.pbxproj
michalconsensys: Reverted

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
michalconsensys: Moved the type up

michalconsensys: https://github.com/MetaMask/metamask-mobile/pull/24964/commits/70ccaaaa45ecdea0b088c73bd6b8a51f74e3e552

app/components/UI/Perps/hooks/usePerpsPaymentTokens.test.ts
michalconsensys: Fixed here https://github.com/MetaMask/metamask-mobile/pull/24964/commits/ba70e5d101bac969af0f1270a509319f1f4e4e30

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -1551,6 +1766,89 @@ const PerpsOrderView: React.FC = () => {
...
+
+  // Reset error state and creation attempt flag when transaction is successfully created
+  useEffect(() => {
+    if (activeTransactionMeta) {
+      if (transactionCreationError) {
+        setTransactionCreationError(null);
+      }
+      hasAttemptedCreationRef.current = false;
+    }
+  }, [activeTransactionMeta, transactionCreationError]);
```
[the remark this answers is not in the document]
michalconsensys: This is intentional for now

abretonc7s (dismissed):

infiniteflower (dismissed):

app/components/Views/confirmations/components/info/external/perps/perps-deposit-fees.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { TotalRow } from '../../../rows/total-row';
+import { InfoRowSkeleton } from '../../../UI/info-row/info-row';
+import {
+  useIsTransactionPayLoading,
+  useTransactionPayQuotes,
+  useTransactionPayRequiredTokens,
+  useTransactionPaySourceAmounts,
+} from '../../../../hooks/pay/useTransactionPayData';
+
+export const PerpsDepositFees = () => {
```
matthewwalsh0, line 15: Placeholder, pending architecture meeting.

matthewwalsh0 (changes requested):

Cal-L (dismissed):
LGTM

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
matthewwalsh0, line 1: As already discussed, out of confirmations scope and not blocking, but would definitely recommend modularising this into multiple components, including isolating the confirmation specific logic into a `PerpsDepositOrderInfo` for example.

matthewwalsh0, line 1: Not a blocker, but we also have an `external` directory inside the confirmations directory to keep all the confirmation components together for easier readability and maintenance, but to allow separate team ownership.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -1529,100 +1535,69 @@ export class PerpsController extends BaseController<
...
+      let result: Promise<string>;
+      let transactionMeta: { id: string };
+
+      if (depositAndPlaceOrder) {
+        // Use addTransaction to create transaction without navigating to confirmation screen
+        const { transactionMeta: addedTransactionMeta } = await addTransaction(
+          transaction,
+          {
+            networkClientId,
+            origin: 'metamask',
```
matthewwalsh0: Minor, `ORIGIN_METAMASK` from `@metamask/controller-utils`.

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -1529,100 +1535,69 @@ export class PerpsController extends BaseController<
...
+            type: TransactionType.perpsDepositAndOrder,
```
matthewwalsh0, line 1554: ❤️

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -1529,100 +1535,69 @@ export class PerpsController extends BaseController<
...
+          // Never resolves - transaction handled via confirmation flow
+        });
+      } else {
+        // submit shows the confirmation screen and returns a promise
+        // The promise will resolve when transaction completes or reject if cancelled/failed
+        const submitResult = await controllers.transaction.submit(transaction, {
           networkClientId,
           origin: 'metamask',
           type: TransactionType.perpsDeposit,
           skipInitialGasEstimate: true,
```
matthewwalsh0: Minor, could remove some duplication here when adding either transaction.

app/components/Views/confirmations/components/confirm/confirm-component.tsx
```diff
@@ -95,9 +102,17 @@ const ConfirmWrapped = ({
 
 interface ConfirmProps {
   route?: UnstakeConfirmationViewProps['route'];
+  /** Optional style applied to the SafeAreaView when confirmation is full screen */
+  safeAreaStyle?: StyleProp<ViewStyle>;
+  /** Optional edges for the SafeAreaView when confirmation is full screen. Defaults to ['right', 'bottom', 'left']. Use [] for no insets. */
+  safeAreaEdges?: Edges;
```
matthewwalsh0: Could we simplify these a little, maybe just a `disableSafeArea` boolean and `fullscreenStyle`?

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
```diff
@@ -15,7 +16,14 @@ import { TransactionType } from '@metamask/transaction-controller';
...
 import { useMusdConversionTokens } from '../../../../../UI/Earn/hooks/useMusdConversionTokens';
 
+interface PayWithModalRouteParams {
+  hideNetworkFilter?: boolean;
+}
+
 export function PayWithModal() {
+  const route =
+    useRoute<RouteProp<{ params: PayWithModalRouteParams }, 'params'>>();
+  const hideNetworkFilter = route.params?.hideNetworkFilter ?? false;
```
matthewwalsh0: Still would like to confirm the design choice here as the network filter is very useful and we want this consistent with the existing deposit?

Plus rather than passing this though the params and `PayWithRow` and `Asset`, could we add a `HIDE_NETWORK_FILTER_TYPES` constant and hide based on that?

As we already have some type specific behaviour in here.

app/components/Views/confirmations/hooks/transactions/useTransactionConfirm.ts
```diff
@@ -128,6 +129,11 @@ export function useTransactionConfirm() {
       log('Error confirming transaction', error);
     }
 
+    // Skip navigation if skipNavigation option is provided
+    if (skipNavigation) {
```
matthewwalsh0: Since we have lots of custom type routing below, could we do the same for our new type?

app/components/Views/confirmations/hooks/useConfirmationRedesignEnabled.ts
```diff
@@ -61,6 +61,10 @@ function isRedesignedTransaction({
   }
 
   if (REDESIGNED_CONTRACT_INTERACTION_TYPES.includes(transactionType)) {
+    // perpsDepositAndOrder (one-click trade) always uses redesigned full-screen flow
+    if (transactionType === TransactionType.perpsDepositAndOrder) {
```
matthewwalsh0: Could this just be added to the condition below, we could convert to a list with `includes`?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
michalconsensys: The network selector removal was confirmed here https://consensys.slack.com/archives/C092T3GPHQD/p1769091779669089?thread_ts=1769087995.104139&cid=C092T3GPHQD

app/components/Views/confirmations/hooks/useConfirmationRedesignEnabled.ts
michalconsensys: Addressed

app/components/Views/confirmations/hooks/transactions/useTransactionConfirm.ts
michalconsensys: Addressed

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
michalconsensys: Addressed

app/components/Views/confirmations/components/confirm/confirm-component.tsx
michalconsensys: Addressed

app/components/UI/Perps/controllers/PerpsController.ts
michalconsensys: Addressed

app/components/UI/Perps/controllers/PerpsController.ts
michalconsensys: Addressed

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
```diff
@@ -0,0 +1,283 @@
...
+        onCustomTokenSelected?.();
+      }
+    }
+  }, [payToken, onCustomTokenSelected]);
+
+  const handleClick = useCallback(() => {
+    if (!canEdit) return;
+    // Mark that user has interacted when they open the modal
+    setHasUserInteracted(true);
+    onCustomTokenSelected?.();
```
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,129 @@
...
+      };
+
+      Engine.controllerMessenger.subscribe(
+        'TransactionController:transactionFailed',
+        handleTransactionFailed,
+      );
+      Engine.controllerMessenger.subscribe(
+        'TransactionController:transactionStatusUpdated',
+        handleTransactionStatusUpdated,
+      );
```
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/Views/confirmations/constants/confirmations.ts
```diff
@@ -62,3 +64,12 @@ export const EARN_CONTRACT_INTERACTION_TYPES = [
   TransactionType.lendingDeposit,
   TransactionType.lendingWithdraw,
 ];
+
+/**
+ * Transaction types for which the Pay With modal hides the network filter.
+ * Used when pay token selection is constrained to a single network (e.g. Perps).
+ */
+export const HIDE_NETWORK_FILTER_TYPES = [
+  TransactionType.perpsDeposit,
```
matthewwalsh0: I'm not a fan of removing it from the perps deposit at the same time as it's existing functionality, is this change new?

matthewwalsh0 (dismissed):

app/components/Views/confirmations/constants/confirmations.ts
michalconsensys: 👍 , reverted this

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -712,218 +794,299 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
-        currentPrice: assetData.price,
-        leverage: orderForm.leverage,
-        // USD as source of truth (hybrid approach)
-        usdAmount: orderForm.amount, // USD amount (primary source of truth, provider calculates size from this)
-        priceAtCalculation: assetData.price, // Price snapshot when size was calculated (for slippage validation)
-        maxSlippageBps:
-          orderForm.type === 'limit'
-            ? ORDER_SLIPPAGE_CONFIG.DefaultLimitSlippageBps // 1% for limit orders
-            : ORDER_SLIPPAGE_CONFIG.DefaultMarketSlippageBps, // 3% for market orders
-        // Only add TP/SL/Limit if they are truthy and/or not empty strings
```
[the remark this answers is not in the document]
michalconsensys: PoC, will be handled later

matthewwalsh0 (dismissed):

app/components/UI/Perps/hooks/usePerpsNavigation.ts
```diff
@@ -124,11 +128,26 @@ export const usePerpsNavigation = (): PerpsNavigationHandlers => {
...
-      navigation.navigate(Routes.PERPS.ORDER, params);
+      depositWithOrder()
+        .then(() => {
+          navigation.navigate(
+            Routes.FULL_SCREEN_CONFIRMATIONS.REDESIGNED_CONFIRMATIONS,
+            params,
+          );
+        })
+        .catch((error: unknown) => {
+          Logger.error(ensureError(error), {
```
abretonc7s, line 143: maybe should keep a TODO here to show a toast on error?

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,129 @@
...
+import { useCallback, useContext } from 'react';
+import Engine from '../../../../core/Engine';
+import { ToastContext } from '../../../../component-library/components/Toast';
+import { strings } from '../../../../../locales/i18n';
+import { getStreamManagerInstance } from '../providers/PerpsStreamManager';
+import { usePerpsLiveAccount } from './stream/usePerpsLiveAccount';
+import usePerpsToasts from './usePerpsToasts';
+
+/**
+ * Hook to track deposit status for Perps order view
```
abretonc7s, line 15: is it just for the order view? Should it be part of the perps context so we can listen if we exit the ordeR?

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,129 @@
...
+
+          toastRef?.current?.closeToast();
+          showToast(
+            PerpsToastOptions.accountManagement.deposit.success(
+              account?.availableBalance?.toString() || '0',
+            ),
+          );
+          setTimeout(
+            () => getStreamManagerInstance().setActiveDepositHandler(false),
+            1000,
```
abretonc7s, line 102: magic num

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,129 @@
...
+          );
+          callback?.();
+        }
+      };
+
+      Engine.controllerMessenger.subscribe(
+        'TransactionController:transactionFailed',
```
abretonc7s, line 109: do we need to unsubscribe when unmount?

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,129 @@
...
+        handleTransactionFailed,
+      );
+      Engine.controllerMessenger.subscribe(
+        'TransactionController:transactionStatusUpdated',
```
abretonc7s, line 113: might be needed here too?

abretonc7s (dismissed):
Approved to unblock but we need to a follow up PR to cleanup the perps method signature and maintain object parameter consistency

aganglada (dismissed):

infiniteflower (dismissed):

Cal-L (dismissed):
Platform code lgtm. I'm seeing a handful of cursor bot comments, let's resolve the ones that are not applicable

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -712,218 +794,299 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
+    ],
+  );
+
+  // Setup transaction custom amount for deposit
+  // Note: This hook requires activeTransactionMeta to exist, otherwise it will crash
+  // We create the transaction in useEffect below when depositAmount is set
+
+  const { updatePendingAmount, amountHuman } = useTransactionCustomAmount({
+    currency: PERPS_CURRENCY,
+  });
```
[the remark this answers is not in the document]
michalconsensys: We do not use the old route anymore

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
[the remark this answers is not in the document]
michalconsensys: Will be handled in a subsequent PR

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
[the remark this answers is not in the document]
michalconsensys: This is intentional for the POC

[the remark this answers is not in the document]
michalconsensys: This is intentional for the POC

app/components/UI/Perps/hooks/usePerpsNavigation.ts
```diff
@@ -124,11 +128,26 @@ export const usePerpsNavigation = (): PerpsNavigationHandlers => {
...
+            feature: PERPS_CONSTANTS.FeatureName,
+            message:
+              'Failed to start one-click trade (deposit rejected or failed)',
+          });
+        });
     },
-    [navigation],
+    [navigation, depositWithOrder],
```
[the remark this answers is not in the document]
michalconsensys: This is intentional

app/components/UI/Perps/Views/PerpsOrderView/PerpsPayRow.tsx
```diff
@@ -0,0 +1,283 @@
...
+    setConfirmationMetric({
+      properties: {
+        mm_pay_token_list_opened: true,
+      },
+    });
+    navigation.navigate(Routes.CONFIRMATION_PAY_WITH_MODAL);
+  }, [canEdit, navigation, setConfirmationMetric, onCustomTokenSelected]);
```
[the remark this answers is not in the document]
michalconsensys: This is intentional for the POC

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
```diff
@@ -0,0 +1,129 @@
...
+          updatedTransactionMeta.status === TransactionStatus.confirmed
+        ) {
+          // Unmark active handler so usePerpsDepositStatus can handle it if needed
+
+          toastRef?.current?.closeToast();
+          showToast(
+            PerpsToastOptions.accountManagement.deposit.success(
+              account?.availableBalance?.toString() || '0',
+            ),
+          );
```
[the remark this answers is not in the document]
michalconsensys: The toast doesnt display balance

app/components/UI/Perps/hooks/usePerpsOrderDepositTracking.ts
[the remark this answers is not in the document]
michalconsensys: This is intentional for the POC

Cal-L (approved):
platform LGTM

javiergarciavera (approved):

vinnyhoward (approved):

aganglada (approved):

dan437 (approved):

PR #25148, opened 2026-01-24
(closed) fix: Multiple bugs when editing SL via Stop Loss Banner

gambinish:
## **Description**

The "Protect against further losses" stop loss banner on the detail page had several UX bugs:
1. After setting then removing a stop loss, returning to the page showed the toggle incorrectly in the "on" state
2. Reusing the toggle caused the screen to flash/flicker
3. Users were seeing "No position found" and "429 Too Many Requests" errors when attempting to set stop loss

See "before" recording for more details.

**Summary of Changes**

**1. Replaced toggle with stateful button for clearer UX**
- Replaced the `Switch` component that felt buggy with async state with a stateful `Button`
- Button shows: text → loader → checkmark → banner fades out

**2. Banner rendering flickering**
- Mostly due to incorrect success state, and boolean rendering conditions
- Added  `preservedBannerVariantRef` to cache the banner variant during fade, preventing the banner from briefly disappearing and reappearing before fading out
- Updated render condition to `(isBannerVisible || isStopLossSuccess) && bannerVariant` to keep banner mounted during fade
- Changed `handleUpdateTPSL` to return `{ success: boolean }` so callers can check actual API result before showing success state
- Prevents success state from triggering when API actually failed

**3. Prefer WebSocket data over REST fetches**
- Attempts to reduce API calls to mitigate rate limiting
- `usePerpsOrderFills` now uses `skipInitialFetch: true` to rely on pre-warmed WebSocket fills channel
- `updatePositionTPSL` accepts optional `position` parameter from WebSocket, skipping the `getPositions()` REST call when provided

**4. Exponential backoff retry for rate limiting resilience**
- Added retry logic wrapping `updatePositionTPSL` to handle HyperLiquid's rate limiting. Details in code comment. Basically this helps by "failing slower and less often"

__

**Known Issue**: Navigation Causes Excessive API Calls (429s)

Navigating rapidly between Asset Details → Perps Tab → Asset Details → Stop Loss Banner Set, it consistently triggers 429 “Too Many Requests” errors, often blocking TP/SL actions.

Even with the changes in `3` and `4`, we are still hitting the rate limit more often than we probably should. When `PerpsMarketDetailsView` mounts, several hooks fire simultaneous REST requests, particularly when fetching candle data. The retry logic helps, but doesn't solve the issue, and TP/SL can often feel "slow" as it works through it's exponential backoff.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-2435

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **Before**

https://github.com/user-attachments/assets/f32310f5-85dc-4cc3-ac73-466417ae2aec

### **After**

https://github.com/user-attachments/assets/9030cd2a-3dc4-4ae4-8001-ab2b488c5150

### DIFF
--- a/app/components/UI/Perps/constants/perpsConfig.ts
+++ b/app/components/UI/Perps/constants/perpsConfig.ts
@@ -509,20 +509,20 @@ export const STOP_LOSS_PROMPT_CONFIG = {
 
   // ROE (Return on Equity) threshold (percentage)
   // Shows "Set stop loss" banner when ROE drops below this value
-  ROE_THRESHOLD: -10,
+  ROE_THRESHOLD: -1, // TODO: TESTING ONLY - restore to -10
 
   // Minimum loss threshold to show ANY banner (percentage)
   // No banner shown until ROE drops below this value
-  MIN_LOSS_THRESHOLD: -10,
+  MIN_LOSS_THRESHOLD: -1, // TODO: TESTING ONLY - restore to -10
 
   // Debounce duration for ROE threshold (milliseconds)
   // User must have ROE below threshold for this duration before showing banner
   // Prevents banner from appearing during temporary price fluctuations
-  ROE_DEBOUNCE_MS: 60_000, // 60 seconds
+  ROE_DEBOUNCE_MS: 1_000, // TODO: TESTING ONLY - restore to 60_000 (60 seconds)
 
   // Minimum position age before showing any banner (milliseconds)
   // Prevents banner from appearing immediately after opening a position
-  POSITION_MIN_AGE_MS: 60_000, // 60 seconds
+  POSITION_MIN_AGE_MS: 1_000, // TODO: TESTING ONLY - restore to 60_000 (60 seconds)
 
   // Suggested stop loss ROE percentage
   // When suggesting a stop loss, calculate price at this ROE from entry
--- a/app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
+++ b/app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
@@ -3389,28 +3389,45 @@ export class HyperLiquidProvider implements IPerpsProvider {
     try {
       this.deps.debugLogger.log('Updating position TP/SL:', params);
 
-      const { symbol, takeProfitPrice, stopLossPrice } = params;
+      const {
+        symbol,
+        takeProfitPrice,
+        stopLossPrice,
+        position: livePosition,
+      } = params;
 
       // Explicitly ensure builder fee approval for trading
       await this.ensureReady();
       await this.ensureBuilderFeeApproval();
 
-      // Get current position to validate it exists
-      // Force fresh API data (not WebSocket cache) since we're about to mutate the position
-      let positions: Position[];
-      try {
-        positions = await this.getPositions({ skipCache: true });
-      } catch (error) {
-        this.deps.logger.error(
-          ensureError(error),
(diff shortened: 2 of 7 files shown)


app/components/UI/Perps/constants/perpsConfig.ts
```diff
@@ -509,20 +509,20 @@ export const STOP_LOSS_PROMPT_CONFIG = {
...
   // Debounce duration for ROE threshold (milliseconds)
   // User must have ROE below threshold for this duration before showing banner
   // Prevents banner from appearing during temporary price fluctuations
-  ROE_DEBOUNCE_MS: 60_000, // 60 seconds
+  ROE_DEBOUNCE_MS: 1_000, // TODO: TESTING ONLY - restore to 60_000 (60 seconds)
 
   // Minimum position age before showing any banner (milliseconds)
   // Prevents banner from appearing immediately after opening a position
-  POSITION_MIN_AGE_MS: 60_000, // 60 seconds
+  POSITION_MIN_AGE_MS: 1_000, // TODO: TESTING ONLY - restore to 60_000 (60 seconds)
```
gambinish, line 525: These are set to 1 second and 1% to make testing easier. This needs to be reverted before merging.

app/components/UI/Perps/hooks/usePerpsTPSLUpdate.ts
```diff
@@ -6,6 +6,24 @@ import type { Position, TPSLTrackingData } from '../controllers/types';
...
+ * - When rate limited: addresses receive 1 request every 10 seconds (drip recovery)
+ *
+ * Strategy: Retry with exponential backoff, but fail fast (~14s total) to avoid
+ * leaving the user waiting too long. Sequence: 2s → 4s → 8s = ~14s total.
+ *
+ * @see https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/rate-limits
+ */
+const TPSL_UPDATE_MAX_RETRIES = 3;
+const TPSL_UPDATE_BASE_DELAY_MS = 2000;
+const TPSL_UPDATE_MAX_DELAY_MS = 10000;
```
gambinish, line 26: This doesn't fully solve our issue, but it does help mitigate it. The root cause is that we exhaust our request limit, by making many http/socket requests to hyperliquid when navigating back and forth from Perps tab to Perps Detail.

So, in the case where a user is rapidly navigating between these two views, they could hit the 429 limit.

I think we should also look into how we can improve this by reducing the amounts of requests fired on the detail view somehow.

app/components/UI/Perps/controllers/providers/HyperLiquidProvider.ts
```diff
@@ -3389,29 +3389,46 @@ export class HyperLiquidProvider implements IPerpsProvider {
...
+          );
+          throw error;
+        }
+        position = positions.find((p) => p.symbol === symbol);
+      } else {
+        this.deps.debugLogger.log('Using live position from WebSocket', {
+          symbol: position.symbol,
+          size: position.size,
+        });
       }
```
gambinish, line 3430: There was a discrepancy between websocket position data and REST data. Where sometimes the REST data wouldn't return the position that already existed.

It discrepancy could happen for a few reasons:
1. 429 rate limit caused REST request to return empty positions, when we already had socket data cached
2. Position got was closed, socket data updated, but REST data wasn't refreshed

So, it would result in the "Position not found" error. By prioritizing socket data, we mitigate this problem.

PR #25159, opened 2026-01-26
fix: cp-7.63.0 Fix `mm_pay_quote_*` metrics

OGPoyraz:
## **Description**

This PR aims to fix `mm_pay_quote_requested` and `mm_pay_quote_loaded` on certain `Transaction*` events.

After this PR metrics will be setlled as noted below:

1. On `TransactionAdded, both `mm_pay_quote_requested` and `mm_pay_quote_loaded` are initialized to `false`.
2. If no quote is requested and the user rejects the transaction, both values stay `false`.
3. If the user views the quote and then rejects the transaction, both values are true in the `TransactionRejected` event.
4. If the user requests a quote but it hasn’t finished loading before they reject, `mm_pay_quote_requested` is true and `mm_pay_quote_loaded` is false.
5. If the user requests a quote, let it load then change the amount to get new quote but leave it while loading, both values remain `true` as they request and load before.
6. If the user views the quote and approves the transaction, both values are true in the finalized `TransactionConfirmed` event.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/25112

## **Manual testing steps**

```gherkin
Feature: my feature name

```

- [X] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [X] I've included tests if applicable
- [X] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [X] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

---

> [!NOTE]
> Improves accuracy of pay quote analytics and introduces explicit source-amount detection.
> 
> - Adds `useTransactionPayHasSourceAmount` hook (+ tests) to detect non-optional source amounts that require a quote and uses it in `custom-amount-info` readiness and in `useTransactionCustomAmount` to set `mm_pay_quote_requested`
> - Refactors `useTransactionPayMetrics` to read `mm_pay_quote_requested` from stored metrics via `selectConfirmationMetricsById` and set `mm_pay_quote_loaded` using an internal ref when quotes are present; removes loading-based inference
> - Updates `useTransactionCustomAmount` to emit `mm_pay_quote_requested` on `updateTokenAmount` only when source amounts exist (+ tests)
> - Adjusts related tests to new behavior and selectors
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5667ac2ba69c9f3601b14f36d220210e3de117ac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/components/Views/confirmations/components/info/custom-amount-info/custom-amount-info.tsx
+++ b/app/components/Views/confirmations/components/info/custom-amount-info/custom-amount-info.tsx
@@ -55,6 +55,7 @@ import EngineService from '../../../../../../core/EngineService';
 import { ConfirmationFooterSelectorIDs } from '../../../ConfirmationView.testIds';
 import { useTransactionPayToken } from '../../../hooks/pay/useTransactionPayToken';
 import { getNativeTokenAddress } from '@metamask/assets-controllers';
+import { useConfirmationMetricEvents } from '../../../hooks/metrics/useConfirmationMetricEvents';
 
 export interface CustomAmountInfoProps {
   children?: ReactNode;
@@ -87,6 +88,7 @@ export const CustomAmountInfo: React.FC<CustomAmountInfoProps> = memo(
     });
     useTransactionPayMetrics();
 
+    const { setConfirmationMetric } = useConfirmationMetricEvents();
     const { isNative: isNativePayToken } = useTransactionPayToken();
     const { styles } = useStyles(styleSheet, {});
     const [isKeyboardVisible, setIsKeyboardVisible] = useState(true);
@@ -118,7 +120,13 @@ export const CustomAmountInfo: React.FC<CustomAmountInfoProps> = memo(
       updateTokenAmount();
       EngineService.flushState();
       setIsKeyboardVisible(false);
-    }, [updateTokenAmount]);
+
+      setConfirmationMetric({
+        properties: {
+          mm_pay_quote_requested: true,
+        },
+      });
+    }, [setConfirmationMetric, updateTokenAmount]);
 
     const handleAmountPress = useCallback(() => {
       setIsKeyboardVisible(true);
--- a/app/components/Views/confirmations/components/info/custom-amount-info/custom-amount-info.tsx
+++ b/app/components/Views/confirmations/components/info/custom-amount-info/custom-amount-info.tsx
@@ -28,8 +28,8 @@ import {
   useIsTransactionPayLoading,
   useTransactionPayQuotes,
   useTransactionPayRequiredTokens,
-  useTransactionPaySourceAmounts,
 } from '../../../hooks/pay/useTransactionPayData';
+import { useTransactionPayHasSourceAmount } from '../../../hooks/pay/useTransactionPayHasSourceAmount';
 import { useTransactionPayMetrics } from '../../../hooks/pay/useTransactionPayMetrics';
 import { useTransactionPayAvailableTokens } from '../../../hooks/pay/useTransactionPayAvailableTokens';
 import Text, {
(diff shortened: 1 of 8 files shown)


dan437 (dismissed):

app/components/Views/confirmations/components/info/custom-amount-info/custom-amount-info.tsx
```diff
@@ -118,7 +120,13 @@ export const CustomAmountInfo: React.FC<CustomAmountInfoProps> = memo(
       updateTokenAmount();
       EngineService.flushState();
       setIsKeyboardVisible(false);
-    }, [updateTokenAmount]);
+
+      setConfirmationMetric({
```
matthewwalsh0: Could this go inside `updateTokenAmount` to minimise the logic in this component?

app/components/Views/confirmations/hooks/pay/useTransactionPayMetrics.ts
```diff
@@ -74,8 +79,9 @@ export function useTransactionPayMetrics() {
 
     properties.mm_pay_payment_token_list_size = availableTokens.length;
 
-    properties.mm_pay_quote_requested = hasRequestedQuoteRef.current;
-    properties.mm_pay_quote_loaded = hasQuotes;
+    properties.mm_pay_quote_requested =
```
matthewwalsh0, line 82: Why do we need to set this here if it's also set elsewhere?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/pay/useTransactionPayMetrics.ts
OGPoyraz: We want this to be explicitly settled to `false` for `TransactionAdded` and `TransactionRejected` events. Hence we either keep this here or do it in the `app/core/Engine/controllers/transaction-controller/metrics_properties/metamask-pay.ts`

matthewwalsh0: But it will always be `undefined` in those cases?

I'm nervous to add complexity to the code only to satisfy metrics that we could infer from our dashboards by checking the event type and `undefined` etc.

OGPoyraz: I agree this can be derived in dashboards, but that might be confusing for non-developers, that’s my main concern, it's more explicit way. On the other hand, we could simplify by using a single tri-state property:

`undefined`: user never engaged with the quote flow
`false`: quote requested but failed/pending
`true`: quote successfully loaded

This just need extra alignment, what do we think to go with this approach?

app/components/Views/confirmations/components/info/custom-amount-info/custom-amount-info.tsx
OGPoyraz: Done

app/components/Views/confirmations/hooks/pay/useTransactionPayMetrics.ts
matthewwalsh0: For the sake of time as it's a blocker, lets leave like this for now 👍

app/components/Views/confirmations/hooks/transactions/useTransactionCustomAmount.ts
```diff
@@ -152,7 +152,13 @@ export function useTransactionCustomAmount({
 
   const updateTokenAmount = useCallback(() => {
     updateTokenAmountCallback(amountHuman);
-  }, [amountHuman, updateTokenAmountCallback]);
+
+    setConfirmationMetric({
+      properties: {
+        mm_pay_quote_requested: true,
```
matthewwalsh0: Won't this also fire if token is Polygon USDCe meaning no quote is needed if Predict?

Should we first ensure `useTransactionPaySourceAmounts` isn't empty?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/transactions/useTransactionCustomAmount.ts
OGPoyraz: Good call - fixed

matthewwalsh0 (approved):

PR #25238, opened 2026-01-27
feat: MUSD-243 enforce same chain conversions in one click

Matt561:
## **Description**
Enforce same-chain-only mUSD conversions in the Pay flow by recreating the unapproved conversion transaction when the user selects a “Pay with” token on a different network.

Preserves the entered amount, sets the new payment token, and automatically rejects the old approval to prevent accidental cross-chain/bridge quotes.

## **Changelog**

CHANGELOG entry: add same-chain mUSD conversion enforcement

## **Related issues**

Fixes: [MUSD-243: Enforce same-chain conversion in one-click conversion](https://consensyssoftware.atlassian.net/browse/MUSD-243)

## **Manual testing steps**

```gherkin
Feature: Same-chain mUSD conversion enforcement

  Scenario: user selects a Pay With token on a different chain during mUSD conversion
    Given user is on the mUSD conversion confirmation screen with an entered amount
      And the current conversion transaction is on chain A

    When user opens "Pay with" and selects a payment token on chain B
    Then the app recreates the mUSD conversion transaction on chain B
      And the entered amount is preserved
      And the confirmation screen reflects the chain B conversion (no cross-chain/bridge conversion)
```

### **After**

https://www.loom.com/share/f799dd93f5404224a16ef11251a0074d

## **Pre-merge author checklist**

- [x] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I've included tests if applicable
- [x] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

---

> [!NOTE]
> **Medium Risk**
[...]

PR #25285, opened 2026-01-27
feat(earn): add earn-musd deeplink handler with navigation fixes

nickewansmith:
## **Description**

- Add handleEarnMusd deeplink handler for earn-musd universal links
- Fix ramp navigation by using goToAggregator instead of goToBuy to work around V2 ramps routing issue
- Fix wallet home navigation with proper nested navigation params
- Simplify deeplink routing logic in EarnMusdConversionEducationView
- Add 'Continue' button text for users routed home
- Add EarnMusdConversionEntryView component
- Add useMusdConversionFlowData hook
- Update tests for all navigation changes

## **Changelog**

CHANGELOG entry: Added `earn-musd` deeplink handler for direct navigation to mUSD conversion education flow

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUSD-248
Fixes: https://consensyssoftware.atlassian.net/browse/MUSD-249

## **Manual testing steps**

```gherkin
Feature: earn-musd Deeplink Navigation

  Scenario: Deeplink navigates geo-eligible user with convertible tokens to conversion flow
    Given user has geo-eligible location
      And user has USDC/USDT in wallet
    When user opens "metamask://earn-musd" deeplink
    Then mUSD education screen displays
    When user taps "Convert to mUSD" button
    Then conversion confirmation screen displays

  Scenario: Deeplink navigates geo-eligible user with empty wallet to buy flow
    Given user has geo-eligible location
      And user wallet is empty
      And mUSD is buyable in region
    When user opens "metamask://earn-musd" deeplink
    Then mUSD education screen displays
      And primary button shows "Buy mUSD"
    When user taps "Buy mUSD" button
    Then Ramp buy screen displays

  Scenario: Deeplink navigates geo-ineligible user to home
    Given user has geo-blocked location
    When user opens "metamask://earn-musd" deeplink
    Then mUSD education screen displays
      And primary button shows "Continue"
    When user taps "Continue" button
    Then wallet home screen displays

  Scenario: Normal flow (non-deeplink) continues to work
    Given user navigates to mUSD conversion via CTA
      And params include preferredPaymentToken and outputChainId
    When user taps "Convert to mUSD" button
    Then conversion flow initiates with provided params
```

### **After**

https://www.loom.com/share/07d771b761cb447cae7eadfb7de9005a

---

> [!NOTE]
> **Medium Risk**
> Medium risk: adds a new universal-link action and changes mUSD conversion education/CTA navigation (home vs buy vs convert) based on derived state, which could misroute users if edge cases are missed.
> 
> **Overview**
> Adds support for a new `earn-musd` deeplink by wiring a dedicated handler into `handleUniversalLink` and routing it to the mUSD conversion education screen with an `isDeeplink` flag (with fallback to `Routes.WALLET.HOME` on errors).
> 
> Updates `EarnMusdConversionEducationView` to handle deeplink-driven routing: it now derives a `deeplinkState` (convert, buy, or navigate home) using a new `useMusdConversionFlowData` hook, adjusts the primary button label (*Convert / Buy mUSD / Continue*), and records metrics with the correct `redirects_to` location.
> 
> Introduces `useMusdConversionFlowData` and `useMusdRampAvailability` to centralize network selection, geo/empty-wallet state, token selection, and Ramp buyability; refactors `useMusdCtaVisibility` and `MusdConversionAssetListCta` to use this unified state (including updated Ramp intent chain selection and improved missing-payment-token handling). Tests and new `EARN_TEST_IDS` are added/updated accordingly, plus a new `earn.musd_conversion.continue` i18n string.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0b5d2f677dbfbfb61ee4237e6ccc399d17446452. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Earn/hooks/useMusdConversionFlowData.ts
```diff
@@ -0,0 +1,190 @@
...
+  getPreferredPaymentToken: () => { address: Hex; chainId: Hex } | null;
+  getChainIdForBuyFlow: () => Hex;
+  getMusdOutputChainId: (inputChainId?: string) => Hex;
+  isMusdBuyableOnChain: Record<Hex, boolean>;
+  isMusdBuyableOnAnyChain: boolean;
+  isMusdBuyable: boolean;
+}
+
+/**
+ * Unified hook for mUSD conversion flow data.
```
Matt561, line 38: Quick question: was the intent for `useMusdConversionFlowData` to be a thin aggregator layer, or the long-term “source of truth” hook for mUSD flows? I’m asking because it spans multiple domains, and I want to make sure we don't create a catch-all/god-hook. At first glance, it seems easier to maintain several small and domain-scoped hooks instead of an aggregator hook.

I'm all for moving reusable logic out of components and into hooks/utils but think we can possibly organize this a bit better.

app/core/DeeplinkManager/handlers/legacy/handleEarnMusd.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Supported URL formats:
+ * - https://link.metamask.io/earn-musd
+ * - metamask://earn-musd
+ */
+export const handleEarnMusd = () => {
+  DevLogger.log('[handleEarnMusd] Starting deeplink handling');
+
+  try {
+    NavigationService.navigation?.navigate(Routes.EARN.ROOT, {
```
Matt561: Is the NavigationService guaranteed to be set when this deeplink handler is called? Mentioning it because `NavigationService.navigation?.navigate(` has the potential of failing silently if it isn't set.

app/components/UI/Earn/hooks/useMusdConversionFlowData.ts
```diff
@@ -0,0 +1,190 @@
...
+  ]);
+
+  const getPreferredPaymentToken = useCallback(() => {
+    if (conversionTokens.length === 0) return null;
+
+    const preferredTokenOnSelectedChain = selectedChainId
+      ? conversionTokens.find((token) => token.chainId === selectedChainId)
+      : undefined;
+
+    const paymentToken = preferredTokenOnSelectedChain ?? conversionTokens[0];
```
Matt561: This fallback to `conversionTokens[0]` seems like it could cause bridging scenarios to occur. We want to enforce same-chain swaps. We do not want users bridging since it won't leverage our Relay subsidy and we can't guarantee dollar parity.

For additional context, I have a PR open to introduce this [here](https://github.com/MetaMask/metamask-mobile/pull/25238)

app/components/UI/Earn/hooks/useMusdConversionFlowData.ts
```diff
@@ -0,0 +1,190 @@
...
+    // No networks selected or unknown state
+    return false;
+  }, [
+    isPopularNetworksFilterActive,
+    isMusdBuyableOnAnyChain,
+    selectedChainId,
+    isMusdBuyableOnChain,
+  ]);
+
+  const getPreferredPaymentToken = useCallback(() => {
```
Matt561: This function name is a bit ambiguous because it's not obvious what "preferred" means. What do you think about naming it something akin to `getPaymentTokenForSelectedNetwork` or something else more intuitive?

app/components/UI/Earn/components/Musd/MusdConversionAssetListCta/index.tsx
```diff
@@ -43,9 +41,14 @@ import { useNetworkName } from '../../../../../Views/confirmations/hooks/useNetw
 const MusdConversionAssetListCta = () => {
   const { styles } = useStyles(styleSheet, {});
 
-  const { goToBuy } = useRampNavigation();
+  const { goToAggregator } = useRampNavigation();
```
Matt561: How come we're calling `goToAggregator` now instead? My understanding is that the aggregator flow is [deprecated](https://github.com/MetaMask/metamask-mobile/blob/aa920ec8c7f3e543807e6fb8fdaa8a1d322502f9/app/components/UI/Ramp/hooks/useRampNavigation.ts#L126) and that `goToBuy` is the latest.

Matt561 (changes requested):

app/components/UI/Earn/components/Musd/MusdConversionAssetListCta/index.tsx
nickewansmith: I did this because of what was happening to me earlier where the buy CTA link was not working using `goToBuy` due to a change in codepath on the Ramps side. This change is what fixed it for me. I copied an existing pattern in the codebase as a stopgap for launch, but I will double check as the issue I thought was on main is not, and so this may have been a fluke in my local env. 

If it proves true, due to time I would probably opt to move forward with the deprecated method that works for now since other parts of the codebase are still using it as well.

app/components/UI/Earn/hooks/useMusdConversionFlowData.ts
nickewansmith: My original problem was that the `initiateConversion` hook was the gateway to a lot of the mUSD logic, by design. However, if working on a component that needs to use the logic inside of `initiateConversion` itself, then there would be a good bit of duplication in my particular case where I needed very slightly different logic than what the CTAs used.

I saw a commonality in the data that was encapsulated in that it is has been used for many of the different ways we are displaying mUSD data. 

I hear you on the catch-all/god-hook, and agree that we should steer clear of those. However, imho as it currently stands, it may be more confusing to figure out what data is needed to make mUSD ui decisions with it scattered in multiple domain specific hooks as we still need to piece the puzzle together each time.

app/core/DeeplinkManager/handlers/legacy/handleEarnMusd.ts
nickewansmith: The `NavigationService` is a class with static methods. I don't see a reason why it would be `undefined` outside of a circular dependency which isn't happening. If it was `undefined` it would throw. As for `navigation?.navigate(`, this is copied from existing pattern in other handlers.

There would be no silent failure in this case because the getter `navigation` on the class has a validation method that throws if the navigation reference cannot be returned. That being said I will update to remove the question mark.

app/components/UI/Earn/components/Musd/MusdConversionAssetListCta/index.tsx
nickewansmith: There is some nesting issue where redirect isn't working with `goToBuy` using ramps unified version. I'm not sure why but this solves the problem for now in the short term.

app/components/UI/Earn/hooks/useMusdConversionFlowData.ts
nickewansmith: I slimmed down the aggregator making one more domain specific hook that was pulled out of it.

app/components/UI/Earn/hooks/useMusdConversionFlowData.ts
nickewansmith: This is actually only selecting the first available conversion token when the view is popular filter networks

Matt561 (approved):

weitingsun (approved):
codeowners file handleUniversalLink.ts looks good to me

PR #25326, opened 2026-01-28
fix(confirmations): handle batch predict deposits and quote‑based alerts

pedronfigueiredo:
- Update pay token selection to use hasTransactionType for predict deposit detection (covers nested/batch txs), and tighten typing/guarding around transactionMeta when updating selectedGasFeeToken. Perps behavior remains unchanged.
- Suppress the native insufficient balance alert when the pay token matches the required token and pay‑controller quotes are loading/available, letting pay‑token balance/fee alerts drive blocking instead.
- Add unit coverage for the quote‑present case and expand mocks for quote/loading state in useInsufficientBalanceAlert tests.

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> **Medium Risk**
> Changes blocking alert behavior and updates how `selectedGasFeeToken` is set during pay-token selection for `predictDeposit`, which could impact confirmation gating and gas/token selection in edge cases.
> 
> **Overview**
> **Confirmation alerts now defer to transaction-pay UI when pay quotes/source amounts are in play.** `useInsufficientBalanceAlert` stops emitting the native insufficient-balance blocking alert whenever `useTransactionPayHasSourceAmount()` indicates pay source amounts are being used, replacing the prior pay-token/required-token matching logic.
> 
> **Pay-token selection for `predictDeposit` is more robust.** `useTransactionPayToken` uses `hasTransactionType` to detect `predictDeposit` (covering nested/batch transactions) and only updates `selectedGasFeeToken` when `transactionMeta` is present, with tighter typing.
> 
> Tests for `useInsufficientBalanceAlert` were updated to mock `useTransactionPayHasSourceAmount` and cover the new suppression behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 567382ea1d7ebe73e8de1f40c6adaa0f051ff9c6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
+++ b/app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
@@ -68,8 +68,8 @@ export const useInsufficientBalanceAlert = ({
     // Simulation is complete if it's disabled, or if enabled and gasFeeTokens is loaded
     const isSimulationComplete = !isSimulationEnabled || Boolean(gasFeeTokens);
 
-    // Check if user has selected a gas fee token (or we're ignoring that check)
-    const hasNoGasFeeTokenSelected = ignoreGasFeeToken || !selectedGasFeeToken;
+    // Check if user has selected a gas fee token (ignore when flag is set)
+    const hasNoGasFeeTokenSelected = !ignoreGasFeeToken && !selectedGasFeeToken;
 
     // Gasless check is complete AND one of:
     //  - Gasless is NOT supported (native currency needed for gas)
@@ -79,7 +79,7 @@ export const useInsufficientBalanceAlert = ({
       isGaslessCheckComplete &&
       (!isGaslessSupported ||
         isGasFeeTokensEmpty ||
-        (!isGasFeeTokensEmpty && !selectedGasFeeToken));
+        (!isGasFeeTokensEmpty && !selectedGasFeeToken && !ignoreGasFeeToken));
 
     const showAlert =
       hasInsufficientBalance &&
--- a/app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
+++ b/app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
@@ -11,12 +11,6 @@ import { useConfirmationContext } from '../../context/confirmation-context';
 import { useIsGaslessSupported } from '../gas/useIsGaslessSupported';
 import { TransactionType } from '@metamask/transaction-controller';
 import { hasTransactionType } from '../../utils/transaction';
-import { useTransactionPayToken } from '../pay/useTransactionPayToken';
-import {
-  useIsTransactionPayLoading,
-  useTransactionPayQuotes,
-  useTransactionPayRequiredTokens,
-} from '../pay/useTransactionPayData';
 import { useTransactionPayHasSourceAmount } from '../pay/useTransactionPayHasSourceAmount';
 import { selectUseTransactionSimulations } from '../../../../../selectors/preferencesController';
 import { useHasInsufficientBalance } from '../useHasInsufficientBalance';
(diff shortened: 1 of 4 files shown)


OGPoyraz (dismissed):

OGPoyraz (dismissed):

app/components/Views/confirmations/hooks/pay/useTransactionPayToken.ts
```diff
@@ -77,6 +85,8 @@ export function useTransactionPayToken(): {
         };
 
         updateTransaction(updatedTx, transactionMeta.id);
+      } else if (isPerpsDepositTransaction) {
```
matthewwalsh0: Why do we need this if the branch is empty?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/pay/useTransactionPayToken.ts
```diff
@@ -57,16 +61,20 @@ export function useTransactionPayToken(): {
...
-        transactionMeta?.type === TransactionType.predictDeposit;
+      const isPredictDepositTransaction = hasTransactionType(transactionMeta, [
+        TransactionType.predictDeposit,
+      ]);
+      const isPerpsDepositTransaction = hasTransactionType(transactionMeta, [
+        TransactionType.perpsDeposit,
+      ]);
 
-      if (isPredictDepositTransaction) {
+      if (isPredictDepositTransaction && transactionMeta) {
```
matthewwalsh0, line 68: Does this ever run without a transaction present? Was it a race condition on confirm?

pedronfigueiredo: This was needed to silence a linting warning as useTransactionMeta returns TransactionMeta | undefined and updateTransaction requires a transactionMeta and id

OGPoyraz (dismissed):

matthewwalsh0 (dismissed):

app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
```diff
@@ -45,6 +51,10 @@ export const useInsufficientBalanceAlert = ({
     payToken.address.toLowerCase() ===
       primaryRequiredToken?.address.toLowerCase();
 
+  const hasPayTokenQuotes = Boolean(quotes?.length);
```
matthewwalsh0: Great find!

Could we simplify even further and just replace `isPayTokenTarget` with:

`const isUsingPay = useTransactionPayHasSourceAmount();`

Then use that on line 62 instead to early return?

Since that would be true if we're relying on quotes, including for Perps, but not for Predict via Polygon USDCe?

https://github.com/MetaMask/core/blob/main/packages/transaction-pay-controller/src/utils/source-amounts.ts#L105-L108

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
pedronfigueiredo: Thanks Matt! I added useTransactionPayHasSourceAmount() and use it as isUsingPay for the quote‑based skip. I kept the existing payToken && !isPayTokenTarget early‑return behavior, but broadened it to && !isUsingPay so non‑target pay tokens still avoid the native‑gas alert. This keeps swap/bridge flows working while letting Perps rely on quotes.

app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
```diff
@@ -45,11 +53,15 @@ export const useInsufficientBalanceAlert = ({
...
+  const shouldSkipInsufficientBalanceForPayToken =
+    isUsingPay && (isQuotesLoading || hasPayTokenQuotes);
+
   return useMemo(() => {
     if (
       !transactionMetadata ||
       isTransactionValueUpdating ||
-      (payToken && !isPayTokenTarget)
+      (payToken && !isPayTokenTarget && !isUsingPay)
```
matthewwalsh0: Sorry for the confusion.

The goal here is to skip this alert if we're using MetaMask Pay with a quote.

So we'd want to return if `isUsingPay` is `true`.

And we shouldn't need `payToken && !isPayTokenTarget` since `isUsingPay` will only be true if there is a pay token, and the controller decided a quote is needed so generated a source amount.

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/alerts/useInsufficientBalanceAlert.ts
```diff
@@ -82,6 +94,7 @@ export const useInsufficientBalanceAlert = ({
         (!isGasFeeTokensEmpty && !selectedGasFeeToken));
 
     const showAlert =
+      !shouldSkipInsufficientBalanceForPayToken &&
```
matthewwalsh0: These changes shouldn't be needed as we can early return above with no alerts if `isUsingPay`.

matthewwalsh0 (approved):

PR #25574, opened 2026-02-03
feat: new token details button layour

bergarces:
## **Description**

Adds new button layout for token details v2, hidden behind a feature flag for now.

## **Changelog**

CHANGELOG entry: Added new token details button layout behind a feature flag

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/ASSETS-2546

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **Before**

image

### **After**

image
image
image

- [X] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [X] I've included tests if applicable
- [X] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [X] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

---

> [!NOTE]
> **Medium Risk**
> Touches token-details navigation and action flows (including token hiding, ramp navigation, and perps entry points) behind a new feature flag, so regressions are limited to enabled cohorts but could affect core token actions when on.
> 
> **Overview**
> Adds a new **TokenDetails V2 action button layout** gated by the new `tokenDetailsV2Buttons` feature flag, including `Cash Buy`, `Long/Short`, and a `More` overflow path that adapts to perps market + balance state and shows a skeleton while buyability/perps availability loads.
> 
> Introduces a new `MoreTokenActionsMenu` bottom sheet route for overflow actions (receive/cash buy/view on explorer/remove token) and wires it into app navigation and routes; also replaces the previous temporary "revamped" toggle with the selector-based flag.
> 
> Refactors `useTokenBuyability` to return `{ isBuyable, isLoading }` and updates call sites/tests, plus adds new test IDs and i18n strings for the new buttons.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d9681457b7b5b8852762aac92c90674c0feaa739. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/TokenDetails/components/AssetOverviewContent.tsx
```diff
@@ -179,6 +182,13 @@ const AssetOverviewContent: React.FC<AssetOverviewContentProps> = ({
 
   const isTokenTrustworthy = isTokenTrustworthyForPerps(token);
 
+  const isTokenDetailsV2ButtonsEnabled =
+    useSelector(selectTokenDetailsV2ButtonsEnabled) || true;
```
bergarces: Remove the constant before review.

app/components/UI/TokenDetails/Views/TokenDetails.tsx
```diff
@@ -128,8 +127,6 @@ const TokenDetails: React.FC<{ token: TokenI }> = ({ token }) => {
     ///: END:ONLY_INCLUDE_IF
   } = useTokenBalance(token);
 
-  const isTokenBuyable = useTokenBuyability(token);
```
bergarces, line 149: Moved this down to another component as it wasn't being used here.

app/components/UI/TokenDetails/hooks/usePerpsActions.ts
bergarces, line 1: This is just returning navigation to the perps market details page, which is what handles the connection to the perps stack.

Ideally, we want to navigate directly to long/short pages, but some setup is required first and we have asked perps team for advice.

Whilst we wait for a response, it's not worth writing tests for this.

app/components/UI/TokenDetails/components/TokenDetailsActions.tsx
bergarces, line 1: I was planning on adding a constan with their layout and the variables involved, as it would be more readable, but there are too many variables at play, and every variable would double the number of configs.

If on top of that we are asked to introduce A/B testing, that would make things rather long if done that way.

bergarces: Instead. I'm relying on having all those combinations as tests and verifying that the correct buttons are being displayed there.

app/components/UI/TokenDetails/components/TokenDetailsActions.tsx
```diff
@@ -0,0 +1,351 @@
...
+  ]);
+
+  // Determine which buttons to display based on perps market and balance
+  // IF no perps market:
+  //   IF buyable: ["Cash Buy", "Send", "Receive", "More"]
+  //   IF not buyable: ["Send", "Receive", "More"]
+  // IF has perps market:
+  //   IF has balance: ["Long", "Short", "Send", "More"]
+  //   IF no balance: ["Long", "Short", "Receive", "More"]
+  const buttons = useMemo(() => {
```
Prithpal-Sooriya, line 206: nit (fuuuuture) - button comparison is pretty complex, thoughts on pulling this out of the component so we can test this as a hook in isolation?

app/components/UI/TokenDetails/components/TokenDetailsActions.tsx
```diff
@@ -0,0 +1,351 @@
...
+    onReceive,
+  ]);
+
+  // Determine which buttons to display based on perps market and balance
+  // IF no perps market:
+  //   IF buyable: ["Cash Buy", "Send", "Receive", "More"]
+  //   IF not buyable: ["Send", "Receive", "More"]
+  // IF has perps market:
+  //   IF has balance: ["Long", "Short", "Send", "More"]
+  //   IF no balance: ["Long", "Short", "Receive", "More"]
```
Prithpal-Sooriya, line 205: (some thoughts to discuss, not a blocker and we can think about it in the future).
I'm thinking we can potentially move this into a config approach if we want to A/B test in the future.

```ts
// raw string literals union, or could be an enums??
const buttonKeysTokenBuyable = ["Cash Buy", "Send", "Receive", ....]
const buttonkeysTokenNonBuyable = ["Send", "Receive", "More"]
...

const ButtonConfig = {
  "Cash Buy": { ... },
  "Send": { ... },
  "Receive": { ... },
  ...
}
```

And when it comes to rendering side, we can keep it a generic map, any buttons that "spill over" are automatically added into the more section.
```tsx

const mainButtons = buttons.slice(0, 3)
const moreButtons = buttons.slice(3)

return (
  <>
    {mainButtons.map(...)}
    {moreButtons.length > 0 ? <MoreButton buttons={moreButtons}} />
  </>
)

```

app/components/UI/TokenDetails/components/TokenDetailsActions.test.tsx
```diff
@@ -0,0 +1,242 @@
...
+      expect(
+        getByTestId(TokenOverviewSelectorsIDs.MORE_BUTTON),
+      ).toBeOnTheScreen();
+      expect(
+        queryByTestId(TokenOverviewSelectorsIDs.RECEIVE_BUTTON),
+      ).not.toBeOnTheScreen();
+      expect(
+        queryByTestId(TokenOverviewSelectorsIDs.BUY_BUTTON),
+      ).not.toBeOnTheScreen();
+    });
```
Prithpal-Sooriya: Love these tests, nice and clear!

(nit - optional) I think we could extract this into an assert util to clean up the duplicate button assertions.

```ts
const assertButtonsVisibility = (opts: {
  visibleButtonIds: string[]
  hiddenButtonIds: string[]
  getByTestId,
  queryByTestId
}) => {
  visibleButtonIds?.map(id => expect(getByTestId(id).toBeOnTheScreen())
  hiddenButtonIds?.map(id => expect(queryByTestId(id).not.toBeOnTheScreen()))
}

```

bergarces:
> Nice one @bergarces ; did you remove the TokenDetails page from the flow all together? If so; can we also remove the 3dots (top right) asset options; if everything is now under the more button I see there is "View on Portfolio" which redirects to portfolio dashboard page; ( perhaps it should redirect to token page instead) we can get rid of for now; the back button also redirects to explore page XDD cc @Prithpal-Sooriya

Ah, good point. It was my plan to remove the menu but I forgot.

app/components/UI/TokenDetails/components/TokenDetailsActions.tsx
bergarces: I thought about it. I left a comment on the PR about it.

The main problem with the config is that you "need" 2^n config entries, where n are the variables that determine what it shows. So far we have:
- isPerpsEnabled
- isBuyable
- hasBalance
- isNative

And for the menu button, on top of those, there's also whether is non-evm.

So those are a lot of config entries. If then we add A/B testing, double that.

And let's face it, chances are some more variables will come into play soon (e.g. whether there's perps balance).

So the choices are "if madness" (current one), "very long config objects" and "clever config objects". The last one would involve having config objects based on only a few variables, but add simple branch logic when it does not change the layout too much. For example, we don't need to make a config based on isBuyable. We keep the existing config but we remove Buy button.

It is also possible that the amount of variables is reduced over time. isBuyable might not be needed once we have a buy + swap flow.

I'm happy to explore a clever config objects approach further down the line, as any other options become a maintenance nightmare in the long run.

Only reason I have chosen the if jungle is that we can easily validate every possible scenario with a test by checking if the correct buttons are rendered, since it's unlikely that a new variable doubles the number of different configs.

Prithpal-Sooriya: Yep understandable, we can def look into this later, not a blocker :)

Prithpal-Sooriya (approved):

PR #25773, opened 2026-02-06
feat(perps): add MYX provider Stage 1 implementation

abretonc7s:
## **Description**

Add MYX protocol as a second perpetuals trading provider (Stage 1 - read-only). MYX is a decentralized perpetuals exchange on BNB Chain that complements the existing HyperLiquid integration, giving users access to MYX-exclusive markets.

**Why?**
- Expand perps trading options beyond HyperLiquid
- Access MYX-exclusive markets (tokens not available on HyperLiquid)
- Build multi-provider infrastructure for perps trading

**What's included (Stage 1 - Read-Only):**
- `MYXProvider` implementing the `PerpsProvider` interface (market data only)
- `MYXClientService` wrapping `@myx-trade/sdk` (v0.1.265)
- Type adapters (`myxAdapter.ts`) for MYX ↔ MetaMask data transformation
- Provider selector UI (badge + bottom sheet) to switch between HyperLiquid and MYX
- Feature flag `perpsMYXProviderEnabled` with graceful fallback
- Refactored controllers to use object parameters for consistency

**Stage 1 Limitations (by design):**
- Read-only: markets and prices only, no trading
- REST polling for prices (WebSocket deferred to Stage 3)
- Positions/orders return empty state (auth not implemented)

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Part of MYX integration epic - Stage 1 implementation

## **Manual testing steps**

```gherkin
Feature: MYX Provider Stage 1

  Scenario: User views MYX markets
    Given user has perps feature enabled
    And perpsMYXProviderEnabled feature flag is true

    When user navigates to Perps home screen
    Then user sees provider selector badge in header
    And user can tap badge to open provider selection sheet

  Scenario: User switches to MYX provider
    Given user is on Perps home screen with HyperLiquid selected

    When user taps provider badge
    And user selects "MYX" from the provider list
    Then provider switches to MYX
    And MYX markets are displayed
    And prices update via REST polling
    And positions/orders sections show empty state

  Scenario: MYX feature flag disabled
    Given perpsMYXProviderEnabled is false

    When user navigates to Perps home screen
    Then provider selector is not shown
    And HyperLiquid remains the only provider
```

### **Before**

N/A - New feature

### **After**

https://github.com/user-attachments/assets/e955fac0-082a-446a-a6f3-a05802f4599d

---

> [!NOTE]
> **Medium Risk**
> Introduces a new perps provider and provider-switch reinitialization flow, which can impact initialization, caching, and navigation behavior even though MYX is read-only in this stage.
> 
> **Overview**
> Adds **Stage 1 read-only support for the MYX perps provider** (markets + price polling) alongside HyperLiquid, including new `MYXProvider`, `MYXClientService` (wrapping `@myx-trade/sdk`), MYX config/constants, and MYX↔Perps adapters.
> 
> Introduces **provider enablement and switching**: a new hybrid feature flag (`perpsMyxProviderEnabled` with `MM_PERPS_MYX_PROVIDER_ENABLED` fallback) controls MYX registration; `PerpsController.switchProvider` now performs a guarded full re-init with rollback on failure, and init falls back to HyperLiquid if MYX is selected but disabled.
> 
> Updates the **Perps UI/navigation** to show a provider badge in the home header (flagged) that opens a new modal bottom sheet for provider selection, and hardens stream/caching behavior by invalidating/discarding market data when the active provider changes mid-fetch. Tests and mocks are added/updated (including a Jest mock for `@myx-trade/sdk`), and ledger update types are generalized to `RawLedgerUpdate` for provider-agnostic returns.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7d3f85697cf129c949a85e73a1e356c92dbb5c44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Perps/components/PerpsProviderSelector/PerpsProviderSelectorBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+  testID,
+}) => {
+  const { styles } = useStyles(styleSheet, {});
+  const navigation = useNavigation();
+  const { activeProvider, isMultiProviderEnabled } = usePerpsProvider();
+
+  const handlePress = useCallback(() => {
+    navigation.navigate(Routes.PERPS.MODALS.ROOT, {
+      screen: Routes.PERPS.MODALS.SELECT_PROVIDER,
+    });
```
michalconsensys, line 41: Is it better to have it in usePerpsNavigation?

app/components/UI/Perps/components/PerpsProviderSelector/PerpsProviderSelectorBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+      ? PROVIDER_DISPLAY_INFO[activeProvider]
+      : PROVIDER_DISPLAY_INFO.hyperliquid;
+
+  return (
+    <TouchableOpacity
+      style={styles.badgeContainer}
+      onPress={handlePress}
+      testID={testID}
+      accessibilityRole="button"
+      accessibilityLabel={`Current provider: ${currentProvider.name}. Tap to change.`}
```
michalconsensys, line 61: Should it use translations?

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,712 @@
...
+      this.deps.debugLogger.log('[MYXProvider] Initialized successfully', {
+        totalPools: pools.length,
+        exclusivePools: this.poolsCache.length,
+      });
+
+      return { success: true };
+    } catch (error) {
+      const err = ensureError(error);
+      this.deps.logger.error(err, this.getErrorContext('initialize'));
+      return { success: false, error: err.message };
```
michalconsensys, line 199: I see some of the methods throw errors and some return an object with the operation status, was this intentional?

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -1275,16 +1295,19 @@ class MarketDataChannel extends StreamChannel<PerpsMarketData[]> {
         const data = await provider.getMarketDataWithPrices();
         const fetchTime = Date.now() - fetchStartTime;
 
-        // Update cache
+        // Update cache and track which provider this data came from
         this.cache.set('markets', data);
         this.lastFetchTime = Date.now();
+        this.cachedProviderId =
+          controller.state?.activeProvider || 'hyperliquid';
```
michalconsensys: Maybe worth saving it in the config file as the default provider

app/components/UI/Perps/services/MYXClientService.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  constructor(
+    deps: PerpsPlatformDependencies,
+    options: { isTestnet?: boolean } = {},
+  ) {
+    this.deps = deps;
+
+    // Stage 1: Force testnet - mainnet credentials not available
+    this.isTestnet = options.isTestnet ?? true;
+    this.chainId = this.isTestnet ? 97 : 56; // BNB testnet/mainnet
```
michalconsensys: Can we grab these numbers from some config file?

app/components/UI/Perps/components/PerpsProviderSelector/PerpsProviderSelectorBadge.tsx
abretonc7s: this will completely disappear currently everything is pretty temporary

app/components/UI/Perps/components/PerpsProviderSelector/PerpsProviderSelectorBadge.tsx
abretonc7s: not for now

app/components/UI/Perps/controllers/providers/MYXProvider.ts
abretonc7s: keeping for now but will make consistent as other methods get implemented

michalconsensys (dismissed):

michalconsensys (dismissed):

michalconsensys (dismissed):

app/components/UI/Perps/controllers/PerpsController.ts
```diff
@@ -2267,11 +2319,100 @@ export class PerpsController extends BaseController<
...
+
+    try {
+      this.debugLog('PerpsController: Provider switch initiated', {
+        from: previousProvider,
+        to: providerId,
+        timestamp: new Date().toISOString(),
+      });
+
+      // Disconnect current provider
+      await this.disconnect();
```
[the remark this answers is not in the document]
gambinish: Low severity but seems like an annoying bug here

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,712 @@
...
+      }
+    }
+
+    if (poolIds.length === 0) {
+      this.deps.debugLogger.log('[MYXProvider] No matching pools for symbols', {
+        symbols,
+      });
+      return () => {
+        /* noop */
+      }; // No-op unsubscribe
```
gambinish: HyperLiquid’s `subscribeToPrices` always calls back; MYX returns a no-op. Should we call `callback([])` once before returning?

If `initialize()` or `getMarkets()` has not run, poolsCache is empty and poolIds will always be empty. We probably want to make sure pools are loaded before subscribing (e.g. call `getMarkets()` or `initialize()` when `poolsCache.length === 0`).

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,712 @@
...
+    return () => {
+      /* noop */
+    };
+  }
+
+  subscribeToOICaps(_params: SubscribeOICapsParams): () => void {
+    return () => {
+      /* noop */
+    };
+  }
```
gambinish: I'm noticing that we also have a lot of other noops here. Wonder if we should instead `params.callback([])` for consistency per my other comment

app/components/UI/Perps/hooks/usePerpsProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * Get list of available providers based on feature flags
+   */
+  const availableProviders = useMemo((): PerpsProviderType[] => {
+    const providers: PerpsProviderType[] = ['hyperliquid'];
+
+    if (isMYXProviderEnabled) {
+      providers.push('myx');
+    }
```
gambinish, line 32: Small nit: but `hyperliquid` and `myx` are magic strings here. Seems like this could be a good candidate for a config string

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,712 @@
...
+
+  async getUserNonFundingLedgerUpdates(_params?: {
+    accountId?: string;
+    startTime?: number;
+    endTime?: number;
+  }): Promise<RawHyperLiquidLedgerUpdate[]> {
+    return [];
+  }
```
gambinish, line 525: This is returning `RawHyperLiquidLedgerUpdate`

We should probably create a separate type for MYX

app/components/UI/Perps/services/MYXClientService.ts
```diff
@@ -0,0 +1,355 @@
...
+// ============================================================================
+// Types
+// ============================================================================
+
+/**
+ * MYX Client Configuration
+ */
+export interface MYXClientConfig {
+  isTestnet: boolean;
+}
```
gambinish, line 30: This currently isn't being used. Should it?

The constructor is using `options: { isTestnet?: boolean }` instead. Might make sense to just start using the `MYXClientConfig` from the start if we're going to use it.

app/components/UI/Perps/services/MYXClientService.ts
```diff
@@ -0,0 +1,355 @@
...
+  // Market Operations
+  // ============================================================================
+
+  /**
+   * Get all available markets/pools
+   * Uses SDK markets.getPoolSymbolAll()
+   */
+  async getMarkets(): Promise<MYXPoolSymbol[]> {
+    // Return cache if valid
+    const now = Date.now();
```
gambinish, line 123: Shouldn't we set this timestamp happen after the fetch?

app/components/UI/Perps/controllers/providers/MYXProvider.ts
abretonc7s: This is already handled — `subscribeToPrices` calls `setTimeout(() => params.callback([]), 0)` when `poolIds.length === 0` (line 325). 

Regarding ensuring pools are loaded before subscribing: the normal flow always calls `initialize()` before subscribing (via `PerpsConnectionManager`), so pools will be loaded. Good enough for Stage 1.

app/components/UI/Perps/hooks/usePerpsProvider.ts
abretonc7s: These are already type-checked via the `PerpsProviderType` union type (`'hyperliquid' | 'myx'`). Any typo would be a compile error, so the type system enforces correctness here. Adding a separate config constant would add indirection without benefit since the union type is the single source of truth.

abretonc7s:
> It seems like `PerpsConnectionManager` clears these when the provider changes...

Good catch — `fills`, `topOfBook`, and `candles` should indeed be cleared on provider switch alongside the other channels. Tracking this as a follow-up since it is a separate concern from the MYX provider implementation.

app/components/UI/Perps/controllers/providers/MYXProvider.ts
abretonc7s: Addressed — all subscription no-ops now call `callback` with empty/default data for consistency. The last remaining one (`subscribeToOrderBook`) was fixed in commit 6a81766.

app/components/UI/Perps/controllers/providers/MYXProvider.ts
abretonc7s: Fixed in commit bfc0fe8 — created provider-agnostic `RawLedgerUpdate` type.

app/components/UI/Perps/services/MYXClientService.ts
abretonc7s: Fixed — constructor now takes `MYXClientConfig` directly.

app/components/UI/Perps/services/MYXClientService.ts
abretonc7s: Fixed — `marketsCacheTimestamp` is now set to `Date.now()` after the fetch completes.

app/components/UI/Perps/providers/PerpsStreamManager.tsx
abretonc7s: Good suggestion — will add default provider config support in a follow-up. For Stage 1 we default to HyperLiquid since it is the only fully-implemented provider.

app/components/UI/Perps/services/MYXClientService.ts
abretonc7s: These constants are already centralized in `myxConfig.ts` — the values here come from MYX SDK defaults. Happy to move to a shared config if we add more providers with similar patterns.

app/components/UI/Perps/controllers/PerpsController.ts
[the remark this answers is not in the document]
abretonc7s: It's intentional — `disconnect()` is a no-op during switch because `performInitialization()` handles the actual provider teardown. The comment above explains the pattern. Agreed it's a bit confusing though, will add a clearer inline comment.

app/components/UI/Perps/types/myx-types.ts
```diff
@@ -0,0 +1,93 @@
...
+ * These are excluded from MYX display in v1.0 to avoid confusion
+ * In Stage 7, we'll implement market collision handling
+ */
+export const MYX_HL_OVERLAPPING_MARKETS = [
+  'BTC',
+  'ETH',
+  'BNB',
+  'PUMP',
+  'WLFI',
+] as const;
```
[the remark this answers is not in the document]
abretonc7s: Acknowledged — `MYX_HL_OVERLAPPING_MARKETS` is a runtime const that belongs in a constants file. Will move it to `myxConfig.ts` in a follow-up.

app/components/UI/Perps/components/PerpsProviderSelector/PerpsProviderSelector.styles.ts
```diff
@@ -0,0 +1,57 @@
...
+      paddingHorizontal: 16,
+      marginVertical: 4,
+      borderRadius: 12,
+      backgroundColor: theme.colors.background.default,
+    },
+    optionRowSelected: {
+      backgroundColor: theme.colors.primary.muted,
+      borderWidth: 1,
+      borderColor: theme.colors.primary.default,
+    },
```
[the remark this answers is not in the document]
abretonc7s: Good catch — the base `optionRow` should have a transparent border of the same width so selection doesn't cause a layout shift. Will fix.

app/components/UI/Perps/utils/myxAdapter.ts
```diff
@@ -0,0 +1,230 @@
...
+ * Check if a symbol overlaps with HyperLiquid markets
+ *
+ * @param symbol - Market symbol to check
+ * @returns true if the symbol is available on both MYX and HyperLiquid
+ */
+export function isOverlappingMarket(symbol: string): boolean {
+  return MYX_HL_OVERLAPPING_MARKETS.includes(
+    symbol as (typeof MYX_HL_OVERLAPPING_MARKETS)[number],
+  );
+}
```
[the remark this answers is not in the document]
abretonc7s: These are utility functions exposed for later stages (`isOverlappingMarket` for Stage 7 collision handling, `buildPoolSymbolMap` is used internally in `buildSymbolPoolsMap`). `getAllTickers` will be used once we add portfolio view. Keeping them for now to avoid churn, but will prune any that remain unused by Stage 3.

app/components/UI/Perps/services/MYXClientService.ts
```diff
@@ -0,0 +1,354 @@
...
+      }
+    } catch (error) {
+      const err = ensureError(error);
+      this.deps.debugLogger.log('[MYXClientService] Price poll failed', {
+        error: err.message,
+      });
+      // Don't propagate error - polling continues
+    } finally {
+      this.scheduleNextPoll();
+    }
```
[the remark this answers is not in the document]
abretonc7s: Already handled — `pollPrices()` re-checks `this.pollingCallback` after the await (line 267). If polling was stopped and restarted during the fetch, the new callback is invoked, which is correct since the data is still fresh ticker data for the same symbols.

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,744 @@
...
+      success: false,
+      error: MYX_NOT_SUPPORTED_ERROR,
+    };
+  }
+
+  async updateMargin(_params: {
+    symbol: string;
+    amount: string;
+    isAdd: boolean;
+  }): Promise<MarginResult> {
```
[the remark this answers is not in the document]
abretonc7s: Correct — the stub uses `{symbol, amount, isAdd}` instead of `UpdateMarginParams`. This is intentional for Stage 1 since the method is a no-op, but will align the signature with the interface when implementing the actual margin update in Stage 2.

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,744 @@
...
+    return {
+      success: false,
+      error: MYX_NOT_SUPPORTED_ERROR,
+    };
+  }
```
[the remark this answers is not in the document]
abretonc7s: Duplicate of the comment above — will fix the signature to use `UpdateMarginParams` in Stage 2.

app/components/UI/Perps/constants/myxConfig.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (!myxPrice || myxPrice === '0') return 0;
+
+  try {
+    const bn = new BigNumber(myxPrice);
+    const divisor = new BigNumber(10).pow(MYX_PRICE_DECIMALS);
+    return bn.dividedBy(divisor).toNumber();
+  } catch {
+    return 0;
+  }
+}
```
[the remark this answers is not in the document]
abretonc7s: Fair point — BigNumber constructor doesn't throw for invalid inputs, it returns NaN. The `bn.isNaN()` check on line 130 is the real guard; the try/catch is defensive but effectively dead code. Will remove the unnecessary try/catch and keep just the NaN check.

app/components/UI/Perps/utils/myxAdapter.ts
```diff
@@ -0,0 +1,230 @@
...
+ */
+export function adaptPriceFromMYX(ticker: MYXTicker): {
+  price: string;
+  change24h: number;
+} {
+  // MYX ticker prices are in 30-decimal format
+  const priceNum = fromMYXPrice(ticker.price);
+
+  // Change is provided as a percentage string (e.g., "2.5" means 2.5%)
+  const change24h = ticker.change ? parseFloat(ticker.change) : 0;
```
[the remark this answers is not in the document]
abretonc7s: Good catch — `parseFloat` on a non-numeric string does return NaN. The fallback `|| 0` on that same line handles the falsy case (`undefined`, empty string) but NaN is truthy. Will add an explicit `isNaN` guard: `const change24h = ticker.change ? parseFloat(ticker.change) : 0` → add `|| 0` after parseFloat to coerce NaN to 0.

app/components/UI/Perps/controllers/providers/MYXProvider.ts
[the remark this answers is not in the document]
abretonc7s: Addressed in the earlier thread — will align with `UpdateMarginParams` in Stage 2.

app/components/UI/Perps/controllers/providers/MYXProvider.ts
```diff
@@ -0,0 +1,744 @@
...
+  private poolSymbolMap: Map<string, string> = new Map();
+
+  // Ticker cache for price data
+  private tickersCache: Map<string, MYXTicker> = new Map();
+
+  // Price subscription callback
+  private priceCallback?: (updates: PriceUpdate[]) => void;
+
+  // Track initialization state
+  private isInitialized = false;
```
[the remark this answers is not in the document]
abretonc7s: These fields (`tickersCache`, `priceCallback`, `isInitialized`) are scaffolding for Stage 2 write operations and price streaming. They're assigned during init but reads come in later stages. Will either remove them and re-add when needed, or add a `// Stage 2` annotation to make intent clear.

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -1275,16 +1300,19 @@ class MarketDataChannel extends StreamChannel<PerpsMarketData[]> {
         const data = await provider.getMarketDataWithPrices();
         const fetchTime = Date.now() - fetchStartTime;
 
-        // Update cache
+        // Update cache and track which provider this data came from
         this.cache.set('markets', data);
         this.lastFetchTime = Date.now();
+        this.cachedProviderId =
+          controller.state?.activeProvider || PROVIDER_CONFIG.DefaultProvider;
```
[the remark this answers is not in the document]
gambinish: https://github.com/MetaMask/metamask-mobile/pull/25773/commits/32bff82add27632dde015c0c3e202be521b81f54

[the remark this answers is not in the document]
gambinish: https://github.com/MetaMask/metamask-mobile/pull/25773/commits/32bff82add27632dde015c0c3e202be521b81f54

app/components/UI/Perps/controllers/types/index.ts
```diff
@@ -12,14 +12,24 @@ import type {
...
+/** Provider-agnostic raw ledger update. Fields match the common shape across providers. */
+export interface RawLedgerUpdate {
+  hash: string;
+  time: number;
+  delta: {
+    type: string;
+    usdc?: string;
+    coin?: string;
+  };
+}
```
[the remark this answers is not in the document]
gambinish: @abretonc7s we will need to address this in Stage 2

deeeed (approved):

aganglada (approved):

PR #25837, opened 2026-02-09
feat(claim-musd): open custom transaction request on Claim bonus click

PatrykLucka:
## **Description**

Implements a custom claim transaction request (bottom sheet confirmation) for `musdClaim` transactions, replacing the generic full-screen transaction request.

**What changed:**
- Added a new `MusdClaimInfo` component that renders a custom confirmation UI with "Claim bonus" title, network subtitle, hero row with token amount/fiat value, claiming address, network info, and gas fee details
- Removed `musdClaim` from `FULL_SCREEN_CONFIRMATIONS` so it renders as a bottom sheet
- The Title component now returns null for `musdClaim` since `MusdClaimInfo` renders its own title/subtitle
- Extended `useTokenAmount` to compute the unclaimed amount (`total - claimed`) by fetching the already-claimed amount from the contract
- Extended `useTokenAsset` to resolve the mUSD token address for `musdClaim` (since `txParams.to` is the Merkl distributor, not the token)
- Refactored `decodeMerklClaimAmount` into `decodeMerklClaimParams` to extract all claim parameters (amount, user, token) with a convenience wrapper
- Changed "Claim bonus" CTA in the token list to directly initiate the claim transaction instead of navigating to asset overview and scrolling
- Added a Spinner (from design system) on the token list item while the claim transaction is being prepared
- Hide "Claim bonus" CTA when there's a pending Merkl claim transaction in-flight
- Removed the entire `scrollToMerklRewards` mechanism (hook, utils, event listeners, tests) as it's no longer needed
- Added `secondaryBalanceElement` prop to `AssetElement` to support custom React nodes in the secondary balance area
- Changed useMerklClaim error handling from throwing to returning undefined with error state, simplifying callers (removed try-catch in ClaimMerklRewards)

**Why:**
The generic transaction request didn't provide enough context for mUSD claim transactions. The custom UI gives users clear information about what they're claiming, on which network, and to which address. The direct claim flow from the token list is simpler than the previous navigate-then-scroll approach.

## **Changelog**

CHANGELOG entry: Added custom claim transaction request screen for mUSD bonus claims with improved UX flow

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/MUSD-305

## **Manual testing steps**

```gherkin
Feature: Custom mUSD claim transaction request

  Scenario: User claims mUSD bonus from token list
    Given user has claimable mUSD rewards visible in the token list
    And "Claim bonus" is displayed as secondary text on the mUSD token row

    When user taps "Claim bonus"
    Then a loading spinner replaces the "Claim bonus" text
    And a bottom sheet confirmation appears with:
      | Element              | Expected Value                              |
      | Title                | Claim bonus                                 |
      | Subtitle             | Bonus payout will be on Linea Mainnet Network. |
      | Hero amount          | Unclaimed mUSD amount with fiat value       |
      | Claiming to          | User's formatted address                    |
      | Network              | Linea Mainnet with network badge            |
      | Network fee          | Gas fee estimate                            |
    And the loading spinner disappears from the token list

  Scenario: User approves the claim transaction
    Given the claim confirmation bottom sheet is displayed

    When user taps "Confirm"
    Then the transaction is submitted to the network
    And "Claim bonus" is no longer shown on the token row while tx is pending

  Scenario: User rejects the claim transaction
    Given the claim confirmation bottom sheet is displayed

    When user taps "Reject"
    Then the bottom sheet closes
    And "Claim bonus" reappears on the token row

  Scenario: Pending claim transaction completes or fails
    Given user previously submitted a claim transaction
    And "Claim bonus" is hidden on the token row

    When the transaction is confirmed or dropped/failed
    Then "Claim bonus" reappears if there are still claimable rewards
```

### **Before**

https://github.com/user-attachments/assets/cd90961d-52c2-473a-a82e-68d9bd09b9f6

### **After**
Full flow for test token (with some values hardcoded)

https://github.com/user-attachments/assets/2d2088a3-add6-4c88-b1e2-47a5b8485c33

Transaction init for Linea and mUSD

https://github.com/user-attachments/assets/90474987-76d9-44fe-89b8-2f452780d62b

---

> [!NOTE]
> **Medium Risk**
> Changes transaction confirmation rendering and mUSD claim submission flow, plus new async contract reads to compute claimable amounts; regressions could affect claim UX and transaction details but scope is limited to mUSD/Merkl paths.
> 
> **Overview**
> Adds a dedicated bottom-sheet confirmation experience for `musdClaim` transactions, including new `MusdClaimInfo` content and updated title/subtitle copy (with network name), and removes `musdClaim` from the full-screen confirmation list.
> 
> Refactors mUSD Merkl claim handling: claim submission now always targets Linea, `useMerklClaim` returns `undefined` instead of throwing and resets `isClaiming` once the request is shown, and new decoding/utils compute the *actual* payout as `total - alreadyClaimed` via a contract read (`useMerklClaimAmount`), with `useTokenAsset` mapping `musdClaim` to the mUSD token.
> 
> Updates the token list to show a "Claim bonus" CTA directly on eligible assets and initiate the claim on tap (showing a spinner via a new `AssetElement.secondaryBalanceElement` override and hiding the CTA when a claim is pending), and deletes the old navigate-and-scroll-to-MerklRewards mechanism (hooks, DeviceEventEmitter listener, and related tests).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f05da70094cc93cac83ac382730c16b1a9019912. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/Views/confirmations/components/title/title.tsx
```diff
@@ -209,6 +210,14 @@ const Title = () => {
     return null;
   }
 
+  // musdClaim renders its own title + subtitle in MusdClaimInfo
```
matthewwalsh0: For consistency, can we not do that here since we have title and subtitle support?

app/components/Views/confirmations/components/info/musd-claim-info/musd-claim-info.tsx
```diff
@@ -0,0 +1,84 @@
...
+          style={styles.subtitle}
+        >
+          {strings('earn.claim_bonus_subtitle', {
+            networkName: networkName ?? 'Linea',
+          })}
+        </Text>
+      </View>
+      <HeroRow />
+      <InfoSection>
+        <InfoRow label={strings('stake.claiming_to')}>
```
matthewwalsh0: Do we not render the account in any existing rows? Could we define a small row component such as `AccountRow`?

app/components/Views/confirmations/components/info/musd-claim-info/musd-claim-info.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Name
+            type={NameType.EthereumAddress}
+            value={address as string}
+            variation={MERKL_CLAIM_CHAIN_ID}
+          />
+        </InfoRow>
+        <InfoRowDivider />
+        <InfoRow label={strings('confirm.label.network')}>
```
matthewwalsh0: `NetworkAndOriginRow`?

app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -221,10 +240,18 @@ export const useTokenAmount = ({
       break;
     }
     case TransactionType.musdClaim: {
-      const claimAmountRaw = decodeMerklClaimAmount(txParams?.data as string);
-      if (claimAmountRaw) {
+      // Wait for the claimed amount to be fetched from the contract
```
matthewwalsh0: Why can we no longer decode this synchronously from the transaction data?

app/components/Views/confirmations/hooks/useTokenAmount.ts
```diff
@@ -221,10 +240,18 @@ export const useTokenAmount = ({
...
+      if (musdClaimedPending) break;
+
+      if (musdClaimParams) {
```
matthewwalsh0: Could we encapsulate all this into a `useMerklClaimAmount` and use that here?

app/components/Views/confirmations/hooks/useTokenAsset.ts
```diff
@@ -36,16 +39,31 @@ export const useTokenAsset = () => {
       return { displayName: strings('token.unknown') };
     }
 
+    // For musdClaim, txParams.to is the Merkl distributor contract, not the mUSD token
     const tokenAddress =
-      transactionType && TypesForNativeToken.includes(transactionType)
-        ? nativeTokenAddress
-        : safeToChecksumAddress(txParams?.to)?.toLowerCase();
+      transactionType === TransactionType.musdClaim
```
matthewwalsh0: Is it worth a small local util to avoid the double ternary?

app/components/Views/confirmations/hooks/useTokenAsset.ts
```diff
@@ -36,16 +39,31 @@ export const useTokenAsset = () => {
...
+          ? nativeTokenAddress
+          : safeToChecksumAddress(txParams?.to)?.toLowerCase();
 
     const asset = tokens[chainId]?.find(
       ({ address }) => address.toLowerCase() === tokenAddress,
     ) as TokenI;
 
     if (!asset) {
+      // For musdClaim, fall back to known mUSD constants when token isn't in user's wallet
+      if (transactionType === TransactionType.musdClaim) {
```
matthewwalsh0: Minor, is it worth speaking to assets team to add this token by default?

matthewwalsh0 (changes requested):

app/components/UI/AssetElement/index.tsx
```diff
@@ -131,34 +137,35 @@ const AssetElement: React.FC<AssetElementProps> = ({
...
+                style={
+                  secondaryBalanceColor
+                    ? styles.secondaryBalanceCustomColor
+                    : styles.secondaryBalance
+                }
+                color={secondaryBalanceColor}
+                isHidden={privacyMode && hideSecondaryBalanceInPrivacyMode}
+                length={SensitiveTextLength.Short}
+                testID={SECONDARY_BALANCE_TEST_ID}
+                // Remove onPress from here since it's on Pressable now
```
Matt561: Do we need this comment?

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
```diff
@@ -43,14 +43,21 @@ export const useMerklClaim = (asset: TokenI) => {
 
   // Get the chain ID where claims should be executed
   // For mUSD, claims always go to Linea regardless of which chain the user is viewing
-  const claimChainId = getClaimChainId(asset);
+  // asset may be undefined during state transitions; getClaimChainId requires a valid asset
+  const claimChainId = asset ? getClaimChainId(asset) : ('' as Hex);
```
Matt561: Your comment above mentions
>For mUSD, claims always go to Linea regardless of which chain the user is viewing

Based on this, could we hardcode Linea as the chainId? I see the `getClaimChainId` util falls back to the `asset.chainId` when it isn't an mUSD token but do we expect to have non-mUSD tokens being claimed? If not, we may be able to simplify this code and just pass in the Linea chainId and remove the `getClaimChainId` util altogether.

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
```diff
@@ -43,14 +43,21 @@ export const useMerklClaim = (asset: TokenI) => {
...
 
   const endpoint = useSelector((state: RootState) =>
     selectDefaultEndpointByChainId(state, claimChainId),
   );
   const networkClientId = endpoint?.networkClientId;
 
   const claimRewards = useCallback(async () => {
+    if (!asset) {
+      const errorMessage = 'No asset available for claiming';
```
Matt561: Do we expect this `errorMessage` to be user-facing? If yes, we'll want a locale string instead of hardcoding English for this and all other error messages in this hook.

app/components/UI/Earn/components/MerklRewards/hooks/useMerklRewards.ts
```diff
@@ -26,7 +26,7 @@ const MUSD_ADDRESS_MAINNET = MUSD_TOKEN_ADDRESS_BY_CHAIN[CHAIN_IDS.MAINNET];
 export const eligibleTokens: Record<Hex, Hex[]> = {
   [CHAIN_IDS.MAINNET]: [AGLAMERKL_ADDRESS_MAINNET, MUSD_ADDRESS_MAINNET], // mUSD and test token
   [CHAIN_IDS.LINEA_MAINNET]: [AGLAMERKL_ADDRESS_LINEA, MUSD_ADDRESS], // mUSD and test token
-  ['0xe709' as Hex]: [AGLAMERKL_ADDRESS_LINEA], // Linea fork
+  ['0xe709' as Hex]: [AGLAMERKL_ADDRESS_LINEA, MUSD_ADDRESS], // Linea fork
```
Matt561, line 28: Non-blocking/Minor: Would it be worth creating a constant for this Linea testnet fork? This way you wouldn't need the comment.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
```diff
@@ -175,10 +177,25 @@ export const TokenListItem = React.memo(
...
+
     const hasClaimableBonus = Boolean(
-      isMerklCampaignClaimingEnabled && claimableReward && isEligibleForMerkl,
+      isMerklCampaignClaimingEnabled &&
+        claimableReward &&
+        isEligibleForMerkl &&
+        !hasPendingClaim,
     );
 
+    const { claimRewards, isClaiming } = useMerklClaim(asset);
```
Matt561: Q: `useMerklClaim` and `usePendingMerklClaim` will be called for every token in the list but only few will need it. Is there a performance impact of doing this? I think only mUSD on Mainnet and Linea will use this right?

app/components/UI/Earn/utils/musd.ts
```diff
@@ -127,29 +127,61 @@ export function convertMusdClaimAmount({
...
     return null;
   }
 
   try {
     const contractInterface = new Interface(DISTRIBUTOR_CLAIM_ABI);
     const decoded = contractInterface.decodeFunctionData('claim', data);
-    // amounts is the 3rd parameter (index 2)
+    const users = decoded[0];
+    const tokens = decoded[1];
     const amounts = decoded[2];
```
Matt561: Minor/Non-blocking: Just sharing in case you weren't aware! It's fine as is 👍 


```suggestion
    const [users, tokens, amounts] = decoded;
```

app/components/UI/TokenDetails/components/AssetOverviewContent.tsx
```diff
@@ -380,16 +376,7 @@ const AssetOverviewContent: React.FC<AssetOverviewContentProps> = ({
...
-              ref={merklRewardsRef}
-              testID="merkl-rewards-section"
-              onLayout={(event) => {
-                // Store Y position relative to header (which is the scroll offset)
-                // This is more reliable than measureInWindow for FlatList scrolling
-                const { y } = event.nativeEvent.layout;
-                merklRewardsYInHeaderRef.current = y;
-              }}
-            >
+            <View ref={merklRewardsRef} testID="merkl-rewards-section">
```
Matt561: Is `merklRewardsRef` redundant now? I don't see it being called anywhere.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
```diff
@@ -175,10 +177,25 @@ export const TokenListItem = React.memo(
...
+
+    const handleClaimBonus = useCallback(async () => {
+      try {
+        await claimRewards();
+      } catch {
+        // Error is handled by useMerklClaim hook
```
Matt561: Is this error caught and handled within `claimRewards`? If yes, this try/catch may not be needed.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
```diff
@@ -381,6 +397,11 @@ export const TokenListItem = React.memo(
         privacyMode={privacyMode}
         hideSecondaryBalanceInPrivacyMode={false}
         onSecondaryBalancePress={secondaryBalanceDisplay.onPress}
+        secondaryBalanceElement={
+          isClaiming ? (
+            <ActivityIndicator size="small" color={colors.primary.default} />
```
Matt561: Non-blocking: I think we use spinners usually instead of the native `ActivityIndicator`. Non-blocking is DS is cool with it!

Matt561 (commented):
Tests well for me! Left some comments 👍

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
PatrykLucka: oh wow, I actually did it, but cursor brought it back! 🙃 Thanks for catching that!

app/components/UI/Earn/components/MerklRewards/hooks/useMerklClaim.ts
PatrykLucka: Not really, it's really just a very defensive mechanism if something will go very wrong. I don't think it is possible to get there realistically.

app/components/UI/Earn/components/MerklRewards/hooks/useMerklRewards.ts
PatrykLucka: I didn’t want to overcommit to this naming, as the chain ID is arbitrarily chosen for a local fork and could theoretically collide with a real chain.

app/components/Views/confirmations/components/info/musd-claim-info/musd-claim-info.tsx
PatrykLucka: We don't need Origin here and we would lose visual grouping with account row, but we could do the similar thing as above - create a small NetworkRow component and use it both here and in the NetworkAndOriginRow, let me propose implementation

app/components/Views/confirmations/hooks/useTokenAmount.ts
PatrykLucka: It was not always correct. The transaction data only contains the cumulative total reward amount (the merkle tree leaf value), not the per-claim payout. We compute `unclaimed` = `totalAmount` - `claimedAmount` to show the user what they'll actually receive. The `claimedAmount` value is stored in the contract's state, so there's no way to derive this from the transaction data alone. Especially if they already claimed a portion of it.

app/components/Views/confirmations/hooks/useTokenAsset.ts
PatrykLucka: Interesting idea — could you elaborate a bit? As I understand it, here we’re checking data for tokens already owned by the user. I don’t think we have any “default tokens” besides native ones (do we?).

The fallback is mainly for cases where a user has mUSD on Ethereum (so they can claim it), but doesn’t have it on Linea, in which case it wouldn’t appear in the token list for that chain.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
PatrykLucka: That's a great question, you're right, it shouldn't be a big impact for users with few tokens, but for huge list of tokens, it could be impactful. Let's try to narrow it to just eligible tokens

app/components/UI/Earn/components/MerklRewards/hooks/useMerklRewards.ts
Matt561: Sounds good!

Matt561 (approved):

vinnyhoward (approved):
UX owned files LGTM

bergarces (approved):

matthewwalsh0 (approved):

app/components/Views/confirmations/components/info/musd-claim-info/musd-claim-info.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const MusdClaimInfo = () => (
+  <View testID={ConfirmationInfoComponentIDs.MUSD_CLAIM}>
+    <HeroRow />
+    <InfoSection>
+      <AccountRow
+        label={strings('stake.claiming_to')}
+        chainId={MERKL_CLAIM_CHAIN_ID}
+      />
+      <InfoRowDivider />
+      <NetworkRow chainId={MERKL_CLAIM_CHAIN_ID} />
```
matthewwalsh0, line 23: Minor, why do we have to fix the chain ID, could we not theoretically support this same transaction on different chains?

PatrykLucka: Currently we support only Linea, and we want to force this chain regardless of which network the action was initiated from.

matthewwalsh0: But that would be a side effect of specifying the relevant `networkClientId` when creating the transaction?

Meaning the UI can't enforce or influence this so for the sake of decoupling and accuracy, should just reflect the transaction metadata?

PatrykLucka: Ah, I see what you mean, you're right. We can hardcode it on the transaction creation level, but the confirmation screen doesn't have to know that and should reflect what transaction data says. Noted, I'll include it in one of the future PRs!

PR #25885, opened 2026-02-10
fix(perps): include tp/sl on limit orders

geositta:
### Why this change

  Limit orders with TP/SL were not consistently represented in our orders UI before the parent limit filled. In those cases, users could miss TP/SL visibility in Market Details and see confusing trigger/execution price presentation.

  ### What this improves

  This PR fixes the limit-order TP/SL display gap and related order-row/detail semantics:

  - Shows TP/SL rows in Market Details Orders for limit orders when TP/SL is available (including synthetic display rows when child orders are not yet present).
  - Keeps full-position TP/SL behavior in Auto-close unchanged.
  - Hides reduce-only orders from Home/Tab lists (per requirement) while still showing relevant orders in Market Details.
  - Clarifies order details:
      - Trigger condition uses trigger price.
      - Price falls back to Market when execution price is unavailable; trigger price remains in Trigger Condition.
      
#### Note on order normalization
HyperLiquid payloads are not fully uniform across all TP/SL cases (child orders, parent metadata, and isPositionTpsl can vary by order flow). This PR adds a UI-side normalization layer (association + synthetic display rows + fallback logic) to ensure consistent behavior across supported and HL-originated variants. This is intentionally defensive; a future refactor can simplify this if/when upstream order contracts become strictly consistent.
  
## **Changelog**

CHANGELOG entry: Fixes the limit-order TP/SL display gap and related order-row/detail display.

## **Related issues**
https://consensyssoftware.atlassian.net/browse/TAT-2234

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```
Feature: Limit order TP/SL display in Perps

    Scenario: user views TP/SL rows for an untriggered limit order
      Given user has placed a limit order with both Take Profit and Stop Loss
      And the parent limit order is still open (not filled)
      When user opens that market and goes to Market Details > Orders
      Then user sees the parent limit order row
      And user sees a Take Profit row
      And user sees a Stop Loss row

    Scenario: user verifies parent limit order details fields
      Given user is on Market Details > Orders with an open parent limit order
      When user taps the parent limit order
      Then Order Details shows Date, Price, Size, Original size, Order value, Reduce only, Fee, Status
      And Order Details shows Cancel order button
      And top-left back arrow closes the screen

    Scenario: user verifies trigger order details fields (stop market)
      Given user opens a "Stop Market close short" order details screen
      When user checks Trigger condition and Price fields
      Then Trigger condition shows "Price above $X" (for short SL)
      And Price shows "Market"

   Scenario: user verifies trigger order details fields (take profit limit)
     Given user opens a "Take Profit Limit close short" order details screen
     When user checks Trigger condition and Price fields
     Then Trigger condition shows "Price below $triggerPrice" (for short TP)
     And Price shows execution limit price "$price" (from order price, not trigger price)

    Scenario: user verifies reduce-only orders are hidden on Home and Tab
      Given user has open reduce-only TP/SL orders
      When user views Perps Home or Perps Tab order sections
      Then reduce-only TP/SL orders are not shown there

    Scenario: user cancels a TP/SL order from details
      Given user opens a cancelable TP/SL order details screen
      When user taps "Cancel order"
      Then cancellation succeeds
      And user sees success feedback

### **After**

https://github.com/user-attachments/assets/14ed085b-368e-4940-8c85-de06c809b439

---

> [!NOTE]
> **Medium Risk**
> Touches order display/cancellation logic and introduces synthetic TP/SL rows plus new filtering flags, which could affect what orders users see and can cancel if edge cases are missed, but changes are scoped to perps UI/adapters with strong test coverage.
> 
> **Overview**
> Fixes TP/SL visibility for open limit orders by **normalizing Market Details order rows**: `PerpsMarketDetailsView` now uses `normalizeMarketDetailsOrders` to include standalone TP/SL orders and add synthetic TP/SL rows from parent order metadata, while filtering out full-position TP/SL (kept for Auto-close).
> 
> Updates order display semantics across components: `PerpsCompactOrderRow`/`PerpsCard` now derive consistent order labels and price labels (trigger vs limit vs market) via new `orderUtils` helpers; `PerpsOrderDetailsView` is refactored to render a richer details table (trigger condition, price fallbacks, original size/value, reduce-only, TP/SL prices) and to **hide the cancel action for non-cancelable synthetic placeholder orders**.
> 
> Adds a `hideReduceOnly` option to `usePerpsLiveOrders` and enables it in Home/Tab order lists to remove reduce-only orders there; extends HyperLiquid order adaptation and perps `Order` type to carry `isPositionTpsl`, parent TP/SL metadata, and synthetic-order fields. Tests are expanded broadly to cover filtering, synthetic row behavior, trigger-condition inference, and adapter/runtime edge cases, plus minor styling/i18n tweaks (e.g., new `market_price` label).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 74d883ab82f9345f9ee1dbc7543335186ac273ac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Perps/components/PerpsCard/PerpsCard.tsx
```diff
@@ -76,13 +79,38 @@ const PerpsCard: React.FC<PerpsCardProps> = ({
...
+      order.isTrigger || isTPSLOrder(order.detailedOrderType),
+    );
+    const isLimitOrder = Boolean(
+      order.orderType === 'limit' ||
+        normalizedDetailedOrderType.includes('limit'),
+    );
+    const validTriggerPrice = getValidTriggerPrice(order);
+    const triggerOrLimitPrice =
+      validTriggerPrice !== null ? order.triggerPrice : order.price;
+    const parsedOrderPrice = parseFloat(triggerOrLimitPrice ?? '0');
```
abretonc7s: we may want to avoid the '0' fallback, I know this is pattern we are trying to get away from in codebase.
It would need first to change the type to allow undefined instead of forcing a string, maybe the key should be ''. Otherwise we have FALLBACK_DATA constant in perpsConfig.ts

app/components/UI/Perps/components/PerpsCompactOrderRow/PerpsCompactOrderRow.tsx
```diff
@@ -49,15 +51,20 @@ const PerpsCompactOrderRow: React.FC<PerpsCompactOrderRowProps> = ({
...
 
-    // Format price - trigger orders already have triggerPx mapped to price by adapter
-    const priceValue = parseFloat(order.price || '0');
+    // For trigger orders, display trigger condition price when available.
+    // `order.price` may represent execution limit/slippage protection for SL/TP.
+    const validTriggerPrice = getValidTriggerPrice(order);
+    const priceValue =
+      order.isTrigger && validTriggerPrice !== null
+        ? validTriggerPrice
+        : parseFloat(order.price ?? '0');
```
abretonc7s: same as previous comment on defaulting to '0'. probably not blocker here but to keep in mind

app/components/UI/Perps/utils/orderUtils.ts
```diff
@@ -12,6 +13,344 @@ import { Position } from '../hooks';
...
+ * Use this instead of inline `parseFloat(order.triggerPrice ?? '')` + validity checks.
+ */
+export const getValidTriggerPrice = (order: Order): number | null => {
+  const parsed = parseFloat(order.triggerPrice ?? '');
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : null;
+};
+
+const getAbsoluteOrderSize = (order: Order): BigNumber | null => {
+  const size = order.originalSize || order.size;
+  const parsedSize = new BigNumber(size || '0');
```
abretonc7s: same with defaulting to 0

app/components/UI/Perps/utils/orderUtils.ts
abretonc7s, line 1: We need to think carefully on how to handle that logic in the UI once we truly work in a protocol agnostic way

app/controllers/perps/utils/hyperLiquidAdapter.ts
```diff
@@ -104,6 +111,11 @@ export function adaptOrderFromSDK(
   rawOrder: FrontendOrder,
   position?: Position,
 ): Order {
+  // TODO: Remove this cast when the SDK FrontendOrder type includes parent-level
+  // TP/SL metadata fields (takeProfitPrice/stopLossPrice and child IDs).
+  const rawOrderWithParentTpsl = rawOrder as FrontendOrderWithParentTpsl;
```
abretonc7s: should we add runtime guard here instead of casting?

abretonc7s (commented):
A few comments to consider before approving after re-reviewing once bugbots addressed and no linting issue

app/components/UI/Perps/components/PerpsCard/PerpsCard.tsx
geositta: @abretonc7s Agree we should move off the '0' fallback.

I kept scope to the order display code that I'm already touching. Wherever we had `parseFloat(... ?? '0')`, I replaced it with `getValidTriggerPrice` / `getValidOrderPrice`. Those return `number | null`, so UI can branch explicity on missing data (`--` or `market`) instead of treating missing as zero.

I left the Order field types unchanged (still required string). Making them optional/nullable would cascade through the adapter/controller/UI layers and is better as a refactor.

Happy to open a follow up to standardize numeric parsing at the adapter boundary (undefined for unavailable values) and migrate callers off the sentinel defaults.

Proposed flow:
API string -> adapter parses to number | undefined -> domain consumes typed values -> UI renders fallback (-- / market) where needed.

app/components/UI/Perps/utils/orderUtils.ts
geositta: Agree, I introduced `normalizeMarketDetailsOrders(...)` as an abstraction layer for detail view order rendering in this PR.

The normalizer is now the integration point where provider specific logic can be routed.

Follow-up for multi-provider: route normalization by providerId (or protocol id) behind this same normalizer, so provider specific logic stays out of `PerpsMarketDetailsView`.

app/controllers/perps/utils/hyperLiquidAdapter.ts
geositta: Good callout. We now widen once to a local boundary type where parent TP/SL fields are unknown, then parse them through `readOptionalString` and `readOptionalOrderId` into `string | undefined`. We still have runtime guards, but in one parser step.

app/components/UI/Perps/Views/PerpsOrderDetailsView/PerpsOrderDetailsView.tsx
```diff
@@ -25,37 +25,65 @@ import { usePerpsMeasurement } from '../../hooks/usePerpsMeasurement';
...
+      return {
+        validOrderPrice: null as number | null,
+        validTriggerPrice: null as number | null,
+        effectivePrice: null as number | null,
+      };
+    }
+
+    const validOrderPrice = getValidOrderPrice(order);
+    const validTriggerPrice = getValidTriggerPrice(order);
+    const effectivePrice = validOrderPrice ?? validTriggerPrice;
```
gambinish, line 80: This seems to be used for order value calculation and fee estimation. However, in `hyperLiquidAdapter.ts` we do the following:

`const price = rawOrder.limitPx || rawOrder.triggerPx || '0';`

So, for stop-market orders where `limitPx` is absent, price gets set to triggerPx. So `getValidOrderPrice(order)` returns the _trigger price_ for stop-market orders, even though it's semantically meant to be the _execution/limit price_.

The Price row correctly shows "Market" (because isMarketExecution is true), but `effectivePrice` ends up being the trigger price, which flows into order value and fee calculations.

I think that this is defensible as a best-available estimate, unless you can think of something else. The semantic coupling here is subtle and could break if the adapter changes.

Can we add a brief comment here explaining this data flow?

gambinish (dismissed):

app/components/UI/Perps/Views/PerpsOrderDetailsView/PerpsOrderDetailsView.tsx
geositta: Good callout! I kept current behavior as a best available estimate and pushed a documentation clarification update:
  - inline comment at the `effectivePrice` fallback in `PerpsOrderDetailsView`
  - clearer contract wording in existing `PerpsOrderDetailsView` and `hyperLiquidAdapter` tests
  
You're right that this can break if adapter semantics change; a good future safeguard would be to expose an explicit price field for estimation (such as `estimatedPriceForNotional` or `executionPrice`) so the UI doesn’t have to derive meaning from price fallback logic.

gambinish (approved):

Issue #7941, opened 2023-11-29
[Bug]: Missing Translation for Biometric Authentication Feature in Canadian English Localization


rmashate:
### Describe the bug

Users with the Canadian English (en-CA) locale experience a missing translation error when attempting to enable biometric authentication in the application. The error message displayed is "[missing "en-CA.biometrics.enable_fac translation]".

### Expected behavior

The application should display a Canadian English translation for enabling biometric authentication, allowing the user to unlock the wallet using facial recognition.

### Screenshots/Recordings

![Screenshot_20231128-210515](https://github.com/MetaMask/metamask-mobile/assets/5077049/329faee2-5628-4c9a-87a0-d87788f22a05)

### Steps to reproduce

Open the application.
Navigate to the login screen.
Observe the error message in the biometric authentication section.

### Error messages or log output

```shell
missing "en-CA.biometrics.enable_face translation
```

### Version

7.11

### Build type

None

### Device

Google Pixel 4 

### Operating system

Android

### Additional context

Ensure that all language keys are correctly implemented in the application's i18n (internationalization) files.
Add a missing translation string for "en-CA.biometrics.enable_fac" in the relevant localization file.

anaamolnar:
Hello, @rmashate. Thanks for reporting! I will pass this on to the team.

gauthierpetetin:
Even if translation is missing, it shall fallback to english.

NicolasMassart:
I tested it and was not able to reproduce with what we currently have on main:
![image](https://github.com/MetaMask/metamask-mobile/assets/4677568/ae186b0d-c17f-43fb-9744-982b05ec3d52)
![image](https://github.com/MetaMask/metamask-mobile/assets/4677568/da4e3d2b-d22b-4233-9b7e-97b31c31db52)

However, the only way to reproduce is to actually remove the string from the en.json file so that it's not available for any english variant and then it will display the error when your language is Canadian English (en-CA)
![image](https://github.com/MetaMask/metamask-mobile/assets/4677568/8e81cd62-56ea-47e6-b8bb-a5c3a1de6ae4)

[...]

PR #26110, opened 2026-02-16
fix(perps): address bugbot findings on Core PR #7941

abretonc7s:
## **Description**

Defensive hardening for edge cases in `PerpsController` surfaced by static analysis on Core PR [#7941](https://github.com/MetaMask/core/pull/7941). None of these bugs are currently triggerable in production — the app works correctly today — but all represent latent issues that could bite under future changes or unusual network conditions.

### What changed

**1. `toggleTestnet()` — false success on silent init failure**

`performInitialization()` catches errors internally and sets `InitializationState.Failed` instead of throwing. `toggleTestnet()` was not checking for this, so it could return `{ success: true }` even when the network switch failed. This patch:
- Adds the same `InitializationState.Failed` check after `await this.init()`
- Rolls back `isTestnet` to its previous value on failure

**2. `depositWithConfirmation()` — never-resolving promise when `placeOrder=true`**

The `placeOrder` path created `new Promise(() => {})` — a promise that can never be GC'd and would hang any consumer who awaits it. Replaced with `Promise.resolve(transactionMeta.id)` for proper fire-and-forget semantics.

**3. `depositWithConfirmation()` — failed deposits remain permanently pending**

`depositWithConfirmation` adds a pending entry to `state.depositRequests` before validating `networkClientId`. If the validation throws, the deposit request stays `status: 'pending'` forever. This patch marks the deposit request as `failed` in the outer catch when a pre-submission error occurs.

**4. Feature flag listener lost after disconnect**

`disconnect()` unsubscribed `RemoteFeatureFlagController:stateChange` but no reconnect path re-subscribed. After disconnect → reconnect, geo-blocking and HIP-3 flag changes stopped propagating. The feature-flag subscription is a controller-lifetime concern (not session-lifetime), so we removed the unsubscribe from `disconnect()` entirely — the subscription now lives for the full controller lifecycle.

**5. `depositWithConfirmation()` — deposit+order request stays pending forever**

When `placeOrder=true`, the `if (!placeOrder)` guard skips the `.then()/.catch()` lifecycle block that transitions the deposit request to `completed`/`failed`. No other handler exists for this path. Added an `else if` branch that attaches lifecycle tracking to `addResult.result` (the real transaction promise) for the deposit+order flow.

**6. Non-EVM account switch leaves stale cache**

The account-change handler only cleared cached user data when `currentAddress` is truthy. Switching to an account group with no EVM account (e.g., Bitcoin-only) skipped cache clearing, leaving stale positions/orders visible. Restructured the condition so cache is always cleared when the account group changes, with preload guarded separately behind `if (currentAddress)`.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: bugbot findings from Core PR [#7941](https://github.com/MetaMask/core/pull/7941)

## **Manual testing steps**

These are defensive fixes for edge cases not reachable through normal UI flows. Verification is via unit tests.

```gherkin
Feature: Perps network toggle resilience

  Scenario: toggleTestnet reports accurate status when init fails silently
    Given user is on mainnet with perps initialized
    When user toggles to testnet and initialization fails internally
    Then toggleTestnet returns { success: false }
    And isTestnet is rolled back to its previous value

Feature: Perps deposit promise contract

  Scenario: depositWithConfirmation result promise resolves when placeOrder is true
    Given user initiates a deposit with placeOrder=true
    When the transaction is submitted
    Then the result promise resolves with the transaction ID

Feature: Perps deposit request lifecycle

  Scenario: deposit request is marked failed on pre-submission error
    Given user initiates a deposit and the deposit request is added as pending
    When the networkClientId lookup fails
    Then the deposit request status is updated to 'failed'

  Scenario: deposit+order request transitions from pending
    Given user initiates a deposit with placeOrder=true
    When the transaction completes or fails
    Then the deposit request status is updated to 'completed' or 'failed'

Feature: Feature flag subscription resilience

  Scenario: geo-blocking flags propagate after disconnect/reconnect
    Given perps controller is initialized with feature flag subscription
    When disconnect() is called and controller reconnects
    Then feature flag changes still propagate correctly

Feature: Account switch cache clearing

  Scenario: switching to non-EVM account clears stale perps cache
    Given user has cached perps data from an EVM account
    When user switches to a Bitcoin-only account group
    Then cached positions, orders, and account state are cleared
```

## **Screenshots/Recordings**

N/A — no UI changes, logic-only hardening.

### DIFF
--- a/app/controllers/perps/PerpsController.ts
+++ b/app/controllers/perps/PerpsController.ts
@@ -881,6 +881,17 @@ export class PerpsController extends BaseController<
    */
   protected activeProviderInstance: PerpsProvider | null = null;
 
+  /**
+   * Cached standalone provider for pre-initialization discovery queries.
+   * Avoids creating a new HyperLiquidProvider (and potentially leaking WebSocket
+   * connections) on every standalone call from the preload cycle.
+   */
+  #standaloneProvider: HyperLiquidProvider | null = null;
+
+  #standaloneProviderIsTestnet: boolean | null = null;
+
+  #standaloneProviderHip3Version: number | null = null;
+
   // Store options for dependency injection (allows core package to inject platform-specific services)
   readonly #options: PerpsControllerOptions;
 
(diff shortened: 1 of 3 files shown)


app/controllers/perps/PerpsController.ts
```diff
@@ -978,15 +1026,22 @@ export class PerpsController extends BaseController<
...
-      this.refreshEligibilityOnFeatureFlagChange.bind(this),
+      featureFlagHandler,
     );
 
     this.providers = new Map();
 
     // Migrate old persisted data without accountAddress
     this.#migrateRequestsIfNeeded();
+
+    this.messenger.registerMethodActionHandlers(
```
michalconsensys, line 1041: Why do we need this?

abretonc7s: This exposes PerpsController methods (placeOrder, cancelOrder, getPositions, etc.) via the messenger pattern so other controllers/services can call them through action handlers instead of needing a direct reference to the controller instance. It's the standard BaseController approach for inter-controller communication - keeps things decoupled.

abretonc7s: basically we didnt need it outside of core, but when putting in core and want to share wiht other controllers we need it.

michalconsensys (approved):

PR #26125, opened 2026-02-16
chore: Update `@metamask/eslint-config-typescript` to v13

Gudahtt:
## **Description**

The ESLint configuration for TypeScript has been updated to prepare for ESLint v9 (this is the last major version before ESLint v9 is required). Various related libraries needed to be updated as well.

The most disruptive part is that in v13, `eslint-plugin-import` was replaced with `eslint-plugin-import-x`. This required widespread changes to any reference to an `import/` rule (it's now `import-x/`), but there should be no functional changes. `eslint-plugin-import-x` is a drop-in replacement for `eslint-plugin-import`.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

N/A

## **Manual testing steps**

N/A

## **Screenshots/Recordings**

N/A

---

> [!NOTE]
> **Low Risk**
> Primarily tooling/config changes (ESLint rules and disable comments) with no intended runtime behavior changes; risk is limited to potential CI/lint or developer workflow breakage if rules are misconfigured.
> 
> **Overview**
> Updates ESLint tooling to support `@metamask/eslint-config-typescript` v13 by switching from `eslint-plugin-import` to the drop-in `eslint-plugin-import-x`, including updating rule names, resolver settings, and inline `eslint-disable` comments across the codebase.
> 
> Adds temporary rule overrides (mostly turning off newer `@typescript-eslint`/`react-hooks` rules) to preserve pre-ESLint-v9 behavior during the migration, and updates `.depcheckrc.yml` ignores to treat ESLint-related packages as intentionally used.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a8b5ec9dafeb8910d09df5d783120eee6047dc14. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Gudahtt:
@SocketSecurity ignore npm/@unrs/resolver-binding-wasm32-wasi@1.11.1
@SocketSecurity ignore npm/napi-postinstall@0.3.4

These have been removed with a resolution

app/controllers/perps/services/AccountService.ts
```diff
@@ -1,5 +1,6 @@
 import { v4 as uuidv4 } from 'uuid';
 
+import type { ServiceContext } from './ServiceContext';
```
Gudahtt, line 3: This series of import re-orderings was caused by the switch from `import/` to `import-x/`. Rather than disable the rule completely (like I did for the other lint violations introduced by these dependency updates), I decided to fix these directly for three reasons:

* They were easy to fix, and
* Disabling the lint rule might cause problems with the automated script that synchronizes this code with `core`
* These changes appear to have been made on `core` already

package.json
```diff
@@ -156,7 +156,9 @@
   "resolutions": {
     "@appium/schema/json-schema": "^0.4.0",
     "@metamask/react-native-payments/validator": "^13.7.0",
+    "@unrs/resolver-binding-wasm32-wasi": "npm:npm-empty-package@1.0.0",
```
Gudahtt, line 159: This is the package used by `@unrs/resolver` in browsers. We only use this package in Node.js, so we don't need this WASM bundle. This was removed with a resolution to address a Socket warning

package.json
```diff
@@ -156,7 +156,9 @@
...
     "d3-color": "3.1.0",
+    "napi-postinstall": "npm:npm-empty-package@1.0.0",
```
Gudahtt, line 161: This package seemingly is only meant to support legacy npm clients, so it serves no purpose for us. Removed via resolution.

package.json
```diff
@@ -329,7 +331,7 @@
     "@react-native-firebase/messaging": "^20.5.0",
     "@react-native-masked-view/masked-view": "^0.3.1",
     "@react-native/babel-preset": "0.76.9",
-    "@react-native/eslint-config": "0.76.9",
+    "@react-native/eslint-config": "^0.82.0",
```
Gudahtt, line 334: This package uses ESLint TypeScript plugins as well, so I updated it to align with the versions used by our ESLint config package.

georgewrmarshall (dismissed):
LGTM! Approving on behalf of @MetaMask/design-system-engineers

tommasini (approved):

metro.transform.js
```diff
@@ -197,6 +197,10 @@ function getESLintInstance() {
     // the build type.
     eslintrc.rules['prefer-const'] = 0;
 
+    // This rule is broken when run under lockdown with `process.env.CI` set to `true`.
+    // Temporarily disabled until we can eliminate code fences.
+    eslintrc.rules['react/no-unescaped-entities'] = 0;
```
Gudahtt, line 202: Tracking this down was very tricky. Essentially it seems that there is a bug in the _parser_ we're using, such that the `react/no-unescaped-entities` rule ends up with two incompatible copies of the source code, which then throws an error when it tries to report a problem with a line that doesn't exist. And the issue only happens under lockdown when `CI` is `true`.

I never discovered the root cause, but it should be OK to disable the rule for now in builds. It's only set to `warn` anyway.

In the future we're hoping to enable tree shaking, which would let us eliminate code fences (and this lint step) altogether.

gambinish (approved):

.eslintrc.js
```diff
@@ -559,6 +621,14 @@ module.exports = {
     'react/prefer-es6-class': 'error',
     '@metamask/design-tokens/color-no-hex': 'warn',
     radix: 'off',
+
+    // These rule modifications are removing changes to our shared ESLint config made after
+    // version v9. This is a temporary measure to get us to ESLint v9 compatible versions,
+    // at which point we can restore the intended rules and use error suppression instead.
+    //
+    // TODO: Remove these modifications after the ESLint v9 update
+    'react-hooks/rules-of-hooks': 'off',
```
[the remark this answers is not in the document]
Gudahtt: True. This is not good. But this will be restored again in the next PR.

PR #26259, opened 2026-02-19
fix: Handle slashes in `StorageService` keys cp-7.66.0

FrederikBolding:
## **Description**

`redux-persist-filesystem-storage` has a number of problems with arbitrary keys. For one it doesn't properly handle slashes but treats them as part of the key. This means that stored data where the key includes slashes are stored in a subdirectory and not returned when using `getAllKeys`. This makes clearing data not work. This PR fixes that by replacing `/` with `%2F`, using URI encoding.

There are other problems, such as `-` being replaced with `:` by default, meaning that keys that include `-` will be misrepresented without recovery. This means there will always be a mismatch between the return value of `getAllKeys` and the actual keys used by the caller of `StorageService`.

Ideally we would URI encode the keys entirely, but this library has been used for so long that migrating would likely be complicated and its configured globally, so we can't even make the breaking change just for the `StorageService` without impacting the rest of the app.

## **Changelog**

CHANGELOG entry: null

### DIFF
--- a/app/core/Engine/controllers/storage-service-init.ts
+++ b/app/core/Engine/controllers/storage-service-init.ts
@@ -13,6 +13,20 @@ import Device from '../../../util/device';
 import Logger from '../../../util/Logger';
 
 /**
+ * Ensures the storage is configured properly and returns it.
+ */
+function getStorage() {
+  FilesystemStorage.config({
+    // Converting : to - is the default behavior of redux-persist-filesystem-storage
+    // This is flawed, but we need to keep that for backwards compatiblity even though it does not work properly for keys including -
+    // We use URI encoding to replace /
+    toFileName: (name) => name.split(':').join('-').split('/').join('%2F'),
+    fromFileName: (name) => name.split('-').join(':').split('%2F').join('/'),
+  });
+  return FilesystemStorage;
+}
+
+/**
  * Mobile-specific storage adapter using FilesystemStorage.
  * This provides persistent storage for large controller data.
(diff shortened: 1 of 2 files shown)


app/core/Engine/controllers/storage-service-init.ts
```diff
@@ -12,6 +12,20 @@ import {
...
+function getStorage() {
+  FilesystemStorage.config({
+    // Converting : to - is the default behavior of redux-persist-filesystem-storage
+    // This is flawed, but we need to keep that for backwards compatiblity even though it does not work properly for keys including -
+    // We use URI encoding to replace /
+    toFileName: (name) => name.split(':').join('-').split('/').join('%2F'),
+    fromFileName: (name) => name.split('-').join(':').split('%2F').join('/'),
+  });
+  return FilesystemStorage;
+}
```
GuillaumeRx: We need to apply the same thing on the migration

FrederikBolding: Hmm, maybe it should be applied globally somewhere during app init then. Because it will apply to all usage of it anyway: https://github.com/robwalkerco/redux-persist-filesystem-storage/blob/master/index.js#L36-L42

FrederikBolding: https://github.com/MetaMask/metamask-mobile/pull/26259/commits/19fe347d798eb415387c7b5f5526b9c872f39851

PR #26260, opened 2026-02-19
feat: Show allowed Predict Withdraw tokens in the select box, remove ENV variable

dan437:
## **Description**

Moves withdraw token filtering configuration from the flat `confirmations_pay` feature flag to a new version-gated `confirmations_pay_post_quote` flag with a `default`/`override` structure keyed by transaction type (e.g. `predictWithdraw`, `perpsWithdraw`).

Override entries fully replace matching default properties (via `??` fallback), so each transaction type can independently control `enabled` and `tokens` while inheriting unset properties from `default`.

The withdraw token picker now builds its list from the allowlist rather than filtering account tokens down — tokens the user doesn't hold are resolved from the `TokenListController` catalog (ERC20s) or network config (native tokens), ensuring allowlisted tokens appear even with a zero balance.

Both the `MM_PREDICT_WITHDRAW_ANY_TOKEN` env var and the remote feature flag `enabled` must be true for the feature to activate.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/6970

## **Manual testing steps**

```gherkin
Feature: Withdraw token filtering via remote feature flag

  Scenario: user sees allowlisted tokens in withdraw token picker
    Given the user has a Predict Withdraw transaction pending
    And the confirmations_pay_post_quote feature flag is enabled with a tokens allowlist
    And MM_PREDICT_WITHDRAW_ANY_TOKEN=true in .js.env

    When user opens the "Select receive token" modal
    Then only tokens matching the allowlist are shown
    And tokens from the allowlist that the user does not hold appear with zero balance

  Scenario: override config takes precedence for predictWithdraw
    Given the feature flag has a default tokens list and a predictWithdraw override with different tokens
    And the transaction type is predictWithdraw

    When user opens the "Select receive token" modal
    Then the override's token list is used instead of the default

  Scenario: feature is disabled when env var is not set
    Given MM_PREDICT_WITHDRAW_ANY_TOKEN is not set or false
    And the feature flag enabled is true

    When user opens a Predict Withdraw confirmation
    Then the "Select receive token" picker is not available
```

---

> [!NOTE]
> **Medium Risk**
> Changes token-listing and selection behavior for withdraw confirmations and adds side effects (auto-tracking tokens) plus new feature-flag plumbing; mistakes could hide tokens or misconfigure post-quote flows.
> 
> **Overview**
> Withdraw token selection is now driven by a new versioned remote flag, `confirmations_pay_post_quote`, with `default` + per-transaction-type `overrides` resolved via a new `selectPayQuoteConfig` selector and `getPostQuoteTransactionType` helper.
> 
> In the Pay-with/Receive token modal, withdraw flows now use a dedicated `useWithdrawTokenFilter` allowlist (fed from the full token catalog via `includeAllTokens`) instead of `getAvailableTokens`, and selecting a zero-balance ERC-20 during withdraw proactively tracks it in `TokensController` so metadata can be resolved.
> 
> This removes the `MM_PREDICT_WITHDRAW_ANY_TOKEN` env var toggle and adds/updates unit tests around override behavior, allowlist filtering (including native-token zero-address matching), and catalog-backed zero-balance tokens.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e6c66568da40d74ac235ddf0fcbad84205706361. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -50,12 +51,17 @@ export const selectMetaMaskPayFlags = createSelector(
...
+      | Record<Hex, Hex[]>
+      | undefined;
+
     return {
       attemptsMax,
       bufferInitial,
       bufferStep,
       bufferSubsequent,
       slippage,
+      allowedWithdrawTokens,
```
dan437: Checking if we can have one list (allowedWithdrawTokens) or split it into two lists (allowedPredictWithdrawTokens, allowedPerpsWithdrawTokens)

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
```diff
@@ -113,9 +119,17 @@ export function PayWithModal() {
...
     (tokens: AssetType[]): TokenListItem[] => {
-      // For withdrawal transactions, show all available tokens (any chain, popular tokens)
-      // The bridging service will handle the actual token conversion
+      // For predict withdrawal transactions, filter by the LD allowlist if set,
+      // otherwise show all tokens. Perps will use its own allowlist.
       if (isTransactionPayWithdraw(transactionMeta)) {
+        if (
+          hasTransactionType(transactionMeta, [
+            TransactionType.predictWithdraw,
+          ]) &&
```
OGPoyraz: Is second check needed if it's already checked in `isTransactionPayWithdraw`?

app/components/Views/confirmations/utils/transaction-pay.ts
```diff
@@ -159,6 +159,23 @@ export function getAvailableTokens({
...
+  allowlist: Record<Hex, Hex[]>,
+): AssetType[] {
+  const allowedSets = new Map<string, Set<string>>();
+  for (const [chainId, addrs] of Object.entries(allowlist)) {
+    allowedSets.set(
+      chainId.toLowerCase(),
+      new Set(addrs.map((a) => a.toLowerCase())),
+    );
+  }
+  return tokens.filter((token) => {
```
OGPoyraz: I think we shouldn't check against `tokens` because these tokens are coming from existing account tokens. So if I don't have USDC on mainnet but I want to withdraw USDC on mainnet (assuming mainnet USDC is allowed list) I won't be able to withdraw?

app/components/Views/confirmations/utils/transaction-pay.ts
```diff
@@ -159,6 +159,23 @@ export function getAvailableTokens({
     });
 }
 
+export function filterTokensByAllowlist(
```
OGPoyraz: Can we rename it to `withdrawTokenFilter` as we follow with other token filter naming conventions (`perpsBalanceTokenFilter`, `musdTokenFilter`)

OGPoyraz: Ideally we can make it hook like others and move it's tests.

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
dan437: We could just use the right allowlist based on if it's predictWithdraw or perpsWithdraw and then we don't need this second check.

app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -16,6 +16,7 @@ export interface MetaMaskPayFlags {
   slippage: number;
   predictWithdrawAnyToken: boolean;
   perpsWithdrawAnyToken: boolean;
+  allowedPredictWithdrawTokens: Record<Hex, Hex[]> | undefined;
```
matthewwalsh0: As discussed, better to decouple from features here and use an override mechanism such as:

```
postQuote?: {
  default?: {
      enabled?: boolean
      tokens?: Record<Hex, Hex[]>;
   },
   override?: {
       [transactionType: string]: {
           ...
       }
    }
}
```

matthewwalsh0 (changes requested):

app/selectors/featureFlagController/confirmations/index.ts
dan437: I've created a new flag (confirmations_pay_post_quote) that's also versioned and has default / override structure.

app/components/Views/confirmations/components/modals/pay-with-modal/pay-with-modal.tsx
```diff
@@ -113,10 +115,8 @@ export function PayWithModal() {
 
   const tokenFilter = useCallback(
     (tokens: AssetType[]): TokenListItem[] => {
-      // For withdrawal transactions, show all available tokens (any chain, popular tokens)
-      // The bridging service will handle the actual token conversion
       if (isTransactionPayWithdraw(transactionMeta)) {
-        return tokens;
+        return withdrawTokenFilter(tokens);
```
dan437, line 148: Should I save this into `filteredTokens` instead, so it would be passed into `wrapHighlightedItemCallbacks`?

app/components/Views/confirmations/hooks/pay/usePayPostQuoteConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+} from '../../../../../selectors/featureFlagController/confirmations';
+
+/**
+ * Resolves the effective post-quote config for the current transaction,
+ * applying any transaction-type-specific override.
+ */
+export function usePayPostQuoteConfig(): PayPostQuoteConfig {
+  const transactionMeta = useTransactionMetadataRequest();
+  const postQuoteFlags = useSelector(selectPayPostQuoteFlags);
+  const overrideKey = getPostQuoteOverrideKey(transactionMeta);
```
dan437: It is `predictWithdraw` when a user opens the Predict Withdraw page.

.js.env.example
```diff
@@ -175,9 +175,6 @@ export REWARDS_API_URL=""
 export MM_PERPS_ENABLED="true"
 
 ## Predict Withdrawal
-# When true, allow users to select any supported token/network for withdrawals
-# When false/unset, withdrawals go to Polygon USDC.E only (original behavior)
-export MM_PREDICT_WITHDRAW_ANY_TOKEN="false"
```
matthewwalsh0, line 180: Do we not want to keep this while transaction details and other PRs and testing are still in progress?

app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -58,14 +66,74 @@ export const selectMetaMaskPayFlags = createSelector(
...
 );
 
+const PAY_POST_QUOTE_DEFAULT: PayPostQuoteConfig = {
+  enabled: false,
+};
+
+export const selectPayPostQuoteFlags = createSelector(
+  selectRemoteFeatureFlags,
+  (featureFlags): PayPostQuoteFlags => {
+    const raw = featureFlags?.confirmations_pay_post_quote as
```
matthewwalsh0, line 82: Rather than casting as we go below, would it be cleanest to define a local type for the full flag, and cast once here?

app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -58,14 +66,74 @@ export const selectMetaMaskPayFlags = createSelector(
...
+      | Record<string, Json>
+      | undefined;
+
+    const rawDefault = raw?.default as Record<string, Json> | undefined;
+    const rawOverride = raw?.override as
```
matthewwalsh0: `overrides`?

app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -58,14 +66,74 @@ export const selectMetaMaskPayFlags = createSelector(
...
+      override,
+    };
+  },
+);
+
+/**
+ * Resolves the effective config for a given override key.
+ * If the key exists in override, the entire override entry is used (no merging with default).
+ */
+export function resolvePayPostQuoteConfig(
```
matthewwalsh0: Since it's just coming from state, would it be simpler to make a `selectPayQuoteConfig` with a `transactionType` argument so we can access it idiomatically via a `useSelector`?

Then we can encapsulate entirely in here?

Could do the same for `selectPayQuoteTokens` with `transactionType` and `chainId` arguments?

app/components/Views/confirmations/utils/transaction.ts
```diff
@@ -132,3 +132,21 @@ export function isTransactionPayWithdraw(
...
   );
 }
+
+/**
+ * Returns the override key for post-quote feature flags based on the withdraw
+ * transaction type. Maps the matching POST_QUOTE_TRANSACTION_TYPES entry to
+ * its string value (e.g. "predictWithdraw"), which is used to look up override
+ * config in confirmations_pay_post_quote flags.
+ */
+export function getPostQuoteOverrideKey(
```
matthewwalsh0: The key is the `transactionType` always so we could specify this when used via a standard selector, see other comment.

app/components/Views/confirmations/hooks/pay/usePayPostQuoteConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+  PayPostQuoteConfig,
+  selectPayPostQuoteFlags,
+  resolvePayPostQuoteConfig,
+} from '../../../../../selectors/featureFlagController/confirmations';
+
+/**
+ * Resolves the effective post-quote config for the current transaction,
+ * applying any transaction-type-specific override.
+ */
+export function usePayPostQuoteConfig(): PayPostQuoteConfig {
```
matthewwalsh0: We shouldn't need this if we expose via standard selector.

app/components/Views/confirmations/utils/transaction-pay.ts
```diff
@@ -159,6 +163,115 @@ export function getAvailableTokens({
...
 }
 
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+export interface BuildAllowlistDeps {
+  tokensChainsCache?: TokenListState['tokensChainsCache'];
+  networkConfigs?: Record<string, NetworkConfiguration>;
+}
+
+function isNativeAddress(address: string): boolean {
```
matthewwalsh0: There are alternate native token addresses for some chains like Polygon, so we already have this exported from the assets controller and use it throughout the UI.

app/components/Views/confirmations/hooks/pay/useWithdrawTokenFilter.ts
```diff
@@ -0,0 +1,45 @@
...
+    [tokensChainsCache, networkConfigs],
+  );
+
+  return useCallback(
+    (tokens: AssetType[]): AssetType[] => {
+      if (!isWithdraw || !allowlist) {
+        return tokens;
+      }
+
+      return buildAllowlistTokens(tokens, allowlist, deps);
```
matthewwalsh0: Why is this a util if only used here?

Would it be cleaner to encapsulate here and just search over `useSendTokens` result for tokens from feature flag as they are already pre-built?

matthewwalsh0 (changes requested):

.js.env.example
dan437: We want to remove this ENV variable and rely on a versioned feature flag, which will allow us to easily test a nightly experimental build.

app/components/Views/confirmations/utils/transaction.ts
dan437: I've renamed the function to `getPostQuoteTransactionType`. The reason why this function is useful is because it also checks nested transactions via hasTransactionType and it filters to post-quote types only.

app/components/Views/confirmations/utils/transaction-pay.ts
dan437: Sounds good. I suppose you meant bridge-controller? `import { isNativeAddress } from '@metamask/bridge-controller';`

app/components/Views/confirmations/hooks/pay/useWithdrawTokenFilter.ts
dan437: Nice, glad I can reuse `useSendTokens`, that simplifies things a lot.

app/components/Views/confirmations/hooks/pay/useWithdrawTokenFilter.ts
```diff
@@ -0,0 +1,80 @@
...
+ * and returns only tokens present in the feature-flag allowlist.
+ */
+export function useWithdrawTokenFilter(): (tokens: AssetType[]) => AssetType[] {
+  const transactionMeta = useTransactionMetadataRequest();
+  const isWithdraw = isTransactionPayWithdraw(transactionMeta);
+  const transactionType = getPostQuoteTransactionType(transactionMeta);
+  const config = useSelector((state: RootState) =>
+    selectPayQuoteConfig(state, transactionType),
+  );
+  const allTokens = useSendTokens({ includeNoBalance: true });
```
OGPoyraz: Not sure if I follow here, `useSendTokens` is existing account tokens, if user don't have USDC, this will end up not showing USDC on withdraw token list even if it's in the allowed list?

Shouldn't we search through `allTokens` from `AssetController` or use some selector to collect all `allowlist` tokens here?

matthewwalsh0: Isn't that why is `includeNoBalance` is used?

The goal here is just to have all the assets already in the right format, and then just remove what is not allowed to greatly simplify the code.

dan437: I saw Matt mentioned `useSendTokens` here: https://github.com/MetaMask/metamask-mobile/pull/26260#discussion_r2834187485
I changed it in this commit to `useSendTokens`: https://github.com/MetaMask/metamask-mobile/pull/26260/changes/e4807bfd37b5e53a94502b5adbb613054acf8746

Before that I used these 2 selectors to get all tokens + [custom](https://github.com/MetaMask/metamask-mobile/pull/26260/changes/e4807bfd37b5e53a94502b5adbb613054acf8746#diff-079ffad4f618d2f2acdee385a6c3d47c5cdd6733fa4b6ef549518babb2f3445aL37) logic to build the final list:
```
const tokensChainsCache = useSelector(selectERC20TokensByChain);
const networkConfigs = useSelector(selectEvmNetworkConfigurationsByChainId);
```

I could revert some of that code back, but maybe there is a more minimalistic solution for this. I could probably combine results from these 2 selectors into `const allTokens` and then I would just use it as I do currently.

app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -58,10 +68,55 @@ export const selectMetaMaskPayFlags = createSelector(
...
+      overrides: raw?.overrides,
+    };
+  },
+);
+
+/**
+ * Resolves the effective post-quote config for a given transaction type.
+ * If the type has an override entry, unset properties fall back to default.
+ */
+export const selectPayQuoteConfig = createDeepEqualSelector(
```
matthewwalsh0: We shouldn't need deep equality checks anymore as the controllers use Immer to update state references.

app/selectors/featureFlagController/confirmations/index.ts
```diff
@@ -58,10 +68,55 @@ export const selectMetaMaskPayFlags = createSelector(
...
+    };
+  },
+);
+
+interface RawPayPostQuoteFlag {
+  default?: PayPostQuoteConfig;
+  overrides?: Record<string, PayPostQuoteConfig>;
+}
+
+export const selectPayPostQuoteFlags = createSelector(
```
matthewwalsh0: Does this need to be exported, is this not just internal for `selectPayQuoteConfig`?

app/components/Views/confirmations/utils/transaction-pay.ts
matthewwalsh0: There is a `getNativeTokenAddress` from `@metamask/assets-controllers` that we use as standard.

app/components/Views/confirmations/hooks/pay/useWithdrawTokenFilter.ts
OGPoyraz: > Isn't that why is `includeNoBalance` is used?

No, that property is just eliminating 0 balance native tokens. The source of that token list selector is `selectAssetsBySelectedAccountGroup`, so it picks from account assets

https://github.com/MetaMask/metamask-mobile/blob/feat/onramp-ux-with-mmpay/app/components/Views/confirmations/hooks/send/useAccountTokens.ts#L20

I think we should revert as @dan437 mentioned above.

matthewwalsh0: That property checks fiat and raw balance of all ERC-20 tokens, not just native doesn't it?

https://github.com/MetaMask/metamask-mobile/blob/main/app/components/Views/confirmations/hooks/send/useAccountTokens.ts#L31

But I see your point about whether it's added to the specific account or not.

We don't want to duplicate the entire logic to build asset types if we already do it inside `useAccountTokens`.

Could we add a `includeAllTokens` option to `useAccountTokens` that iterates over `selectAllTokens` instead, but gets any balances from the original selector, then we rename the hook in future to be more general.

app/components/Views/confirmations/hooks/pay/useWithdrawTokenFilter.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+  getPostQuoteTransactionType,
+  isTransactionPayWithdraw,
+} from '../../utils/transaction';
+import { selectPayQuoteConfig } from '../../../../../selectors/featureFlagController/confirmations';
+import { useSendTokens } from '../send/useSendTokens';
+import { AssetType } from '../../types/token';
+import { RootState } from '../../../../../reducers';
+
+function isAllowlisted(
```
matthewwalsh0: Minor, exported functions at top?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
```diff
@@ -62,14 +71,66 @@ export function useAccountTokens({
         networkBadgeSource: getNetworkBadgeSource(asset.chainId as Hex),
         balanceInSelectedCurrency,
         standard: TokenStandard.ERC20 as const,
-      };
+      } as AssetType;
```
matthewwalsh0, line 74: Why do we have to cast here?

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
```diff
@@ -62,14 +71,66 @@ export function useAccountTokens({
...
+    if (includeAllTokens) {
+      let zeroFiat: string;
+      try {
+        zeroFiat = getIntlNumberFormatter(I18n.locale, {
+          style: 'currency',
+          currency: fiatCurrency,
+          minimumFractionDigits: 0,
+        }).format(0);
+      } catch {
+        zeroFiat = `0 ${fiatCurrency}`;
```
matthewwalsh0, line 86: Can this ever fail?

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
```diff
@@ -62,14 +71,66 @@ export function useAccountTokens({
...
+      );
+
+      for (const [chainId, cache] of Object.entries(tokensChainsCache ?? {})) {
+        for (const [address, entry] of Object.entries(cache?.data ?? {})) {
+          if (
+            existing.has(`${chainId.toLowerCase()}:${address.toLowerCase()}`)
+          ) {
+            continue;
+          }
+          processedAssets.push({
```
matthewwalsh0: Minor, worth a local function such as `buildNoBalanceAsset` for readability?

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
```diff
@@ -62,14 +71,66 @@ export function useAccountTokens({
...
+        flatAssets.map(
+          (a) =>
+            `${String(a.chainId).toLowerCase()}:${String('address' in a ? a.address : a.assetId).toLowerCase()}`,
+        ),
+      );
+
+      for (const [chainId, cache] of Object.entries(tokensChainsCache ?? {})) {
+        for (const [address, entry] of Object.entries(cache?.data ?? {})) {
+          if (
+            existing.has(`${chainId.toLowerCase()}:${address.toLowerCase()}`)
```
matthewwalsh0: Minor, worth a local `getAssetKey` function to avoid the duplication?

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
```diff
@@ -62,14 +71,66 @@ export function useAccountTokens({
...
+            decimals: entry.decimals ?? 18,
+            image: entry.iconUrl ?? '',
+            logo: entry.iconUrl ?? undefined,
+            balance: '0',
+            balanceInSelectedCurrency: zeroFiat,
+            isETH: false,
+            isNative: false,
+            networkBadgeSource: getNetworkBadgeSource(chainId as Hex),
+            standard: TokenStandard.ERC20,
+          } as AssetType);
```
matthewwalsh0: Is cast required?

matthewwalsh0 (dismissed):

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
dan437: Because we do this spread `...asset` and the Asset type is a union of EVM and Multichain, which doesn't match AssetType. I've removed casting from other places though.

app/components/Views/confirmations/hooks/send/useAccountTokens.ts
dan437: Yes, getIntlNumberFormatter can throw if fiatCurrency is invalid / unsupported currency code. We already have this try / catch pattern on line 58.

matthewwalsh0 (approved):

PR #26268, opened 2026-02-19
fix: Fix StorageService Key Encoding cp-7.66.0

andrepimenta:
### Problem

The `redux-persist-filesystem-storage` library has two problematic behaviors when handling arbitrary keys in `StorageService`:

1. **Slashes (`/`) create subdirectories**: Keys containing slashes are stored in subdirectories, making them unreachable via `getAllKeys`. This breaks the `clear` method for affected keys.

2. **Hyphens (`-`) get corrupted**: The library's internal `fromFileName` function converts hyphens to colons (`:`), meaning keys like `simple-key` are returned as `simple:key` by `getAllKeys`. This causes a permanent mismatch between stored keys and returned keys.

### Solution

This PR introduces URI-style encoding for problematic characters in **both namespace and key** portions of storage keys:

- `/` → `%2F` (prevents subdirectory creation)
- `-` → `%2D` (prevents hyphen-to-colon corruption)
- `%` → `%25` (prevents double-encoding issues)

The encoding is applied via shared utility functions (`encodeStorageKey`/`decodeStorageKey`) used by both:
- `mobileStorageAdapter` in `storage-service-init.ts`
- Migration 118 which stores snap source code

### Backward Compatibility

**This change is backward compatible and will NOT break existing production keys.**

- Existing keys like `storageService:TokenListController:tokensChainsCache:0x1` are **unaffected** because:
  - The namespace (`TokenListController`) has no special characters → encoding produces identical output
  - Colons (`:`) are **not encoded** - they pass through unchanged
  - The key portion (`tokensChainsCache:0x1`) contains no hyphens or slashes
- **Strings without `-`, `/`, or `%` characters produce identical output when encoded**
- This means all current production keys work exactly as before, while future keys with special characters will be handled correctly

### Examples

```
# No special characters → unchanged (backward compatible)
storageService:TokenListController:tokensChainsCache:0x1 → storageService:TokenListController:tokensChainsCache:0x1

# Snap ID with slashes and hyphens → encoded
storageService:SnapController:npm:@metamask/bip32-keyring-snap → storageService:SnapController:npm:@metamask%2Fbip32%2Dkeyring%2Dsnap

# Namespace with hyphen → encoded
storageService:Some-Controller:some-key → storageService:Some%2DController:some%2Dkey

# Key with slashes → encoded (prevents subdirectory creation)
storageService:TestController:nested/path/key → storageService:TestController:nested%2Fpath%2Fkey
```

### Files Changed

| File | Change |
|------|--------|
| `app/core/Engine/utils/storage-service-utils.ts` | New utility with `encodeStorageKey`/`decodeStorageKey` functions |
| `app/core/Engine/utils/storage-service-utils.test.ts` | 35 unit tests for the encoding utilities |
| `app/core/Engine/controllers/storage-service-init.ts` | Apply encoding in `mobileStorageAdapter` methods |
| `app/core/Engine/controllers/storage-service-init.test.ts` | 22 new tests for key and namespace encoding behavior |
| `app/store/migrations/119.ts` | Encode snap IDs when storing snap source code |

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Refs: StorageService key handling issues with `redux-persist-filesystem-storage`

## **Manual testing steps**

```gherkin
Feature: StorageService key encoding

  Scenario: Keys with hyphens are stored and retrieved correctly
    Given the app is running

    When StorageService stores a key containing hyphens (e.g., "npm:@metamask/bip32-keyring-snap")
    Then the key is encoded as "npm:@metamask%2Fbip32%2Dkeyring%2Dsnap" on disk
    And getAllKeys returns the original key "npm:@metamask/bip32-keyring-snap"
    And getItem with the original key returns the stored data

  Scenario: Keys with slashes are stored and retrieved correctly
    Given the app is running

    When StorageService stores a key containing slashes (e.g., "nested/path/key")
    Then the key is stored as a single file (not in subdirectories)
    And getAllKeys returns the original key "nested/path/key"
    And clear removes the key correctly

  Scenario: Existing keys with colons remain unchanged
    Given existing production keys like "storageService:TokenListController:tokensChainsCache:0x1"

    When the app starts with this fix
    Then the existing keys are still accessible
    And no migration is required for existing data
```

## **Screenshots/Recordings**

N/A - No UI changes

### **Before**

N/A

### **After**

N/A

---

> [!NOTE]
> **Medium Risk**
> Changes the on-disk key format for any namespaces/keys containing `%`, `/`, or `-`, which could impact access to previously stored data for those cases; behavior is covered by new unit tests but touches persistence and migration logic.
> 
> **Overview**
> Fixes StorageService filesystem key handling by **URI-encoding namespaces and keys** (encoding `%`, `/`, `-` while leaving `:` intact) before calling `redux-persist-filesystem-storage`, and decoding keys returned from `getAllKeys`.
> 
> Updates `mobileStorageAdapter` to apply this encoding in `getItem`/`setItem`/`removeItem`, use encoded namespace prefixes in `getAllKeys`/`clear`, and adds a shared `storage-service-utils` module with thorough unit coverage. Migration `119` is updated (and its test adjusted) to store snap source code under the **encoded snap ID** so snap IDs containing slashes/hyphens are retrievable/clearable.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 59b39f4c6a747402763a6641ec4b17e83039d2ba. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/core/Engine/utils/storage-service-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ * We use URI-style encoding for these characters.
+ *
+ * @param key - The key to encode.
+ * @returns The encoded key safe for filesystem storage.
+ */
+export const encodeStorageKey = (key: string): string =>
+   key
+    .replace(/%/g, '%25') // Encode % first to avoid double-encoding
+    .replace(/\//g, '%2F') // Encode slashes (would create subdirectories)
+    .replace(/-/g, '%2D') // Encode hyphens (would be converted to colons)
```
FrederikBolding: Hyphens are valid in URIs, how did you decide this encoding?

app/core/Engine/controllers/storage-service-init.ts
```diff
@@ -30,8 +34,11 @@ const mobileStorageAdapter: StorageAdapter = {
    */
   async getItem(namespace: string, key: string): Promise<StorageGetResult> {
     try {
-      // Build full key: storageService:namespace:key
-      const fullKey = `${STORAGE_KEY_PREFIX}${namespace}:${key}`;
+      // Build full key: storageService:encodedNamespace:encodedKey
+      const encodedNamespace = encodeStorageKey(namespace);
```
FrederikBolding, line 38: What's the argument for encoding namespaces too? I would expect those to be quite standardized?

app/core/Engine/utils/storage-service-utils.ts
GuillaumeRx: why aren't we using something like `encodeURI` or any util function available rather than creating our own ? 🤔

app/core/Engine/utils/storage-service-utils.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * Utility functions for StorageService key encoding/decoding.
+ *
+ * These functions handle the quirks of redux-persist-filesystem-storage:
+ * 1. `/` in keys creates subdirectories, making keys unreachable via getAllKeys
+ * 2. `-` gets converted to `:` by fromFileName, corrupting the key
+ *
+ * We use URI-style encoding for these characters.
+ * Note: We only encode the key part, not the namespace (for backward compatibility).
```
FrederikBolding: This is not accurate to the other changes in the PR

app/core/Engine/utils/storage-service-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ * 2. `-` gets converted to `:` by fromFileName, corrupting the key
+ *
+ * We use URI-style encoding for these characters.
+ *
+ * @param key - The key to encode.
+ * @returns The encoded key safe for filesystem storage.
+ */
+export const encodeStorageKey = (key: string): string =>
+   key
+    .replace(/%/g, '%25') // Encode % first to avoid double-encoding
```
FrederikBolding, line 24: If we are URI encoding anyway we could consider using `encodeURIComponent`?

app/core/Engine/utils/storage-service-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ * The library has two problematic behaviors:
+ * 1. `/` in keys creates subdirectories, making keys unreachable via getAllKeys
+ * 2. `-` gets converted to `:` by fromFileName, corrupting the key
+ *
+ * We use URI-style encoding for these characters.
+ *
+ * @param key - The key to encode.
+ * @returns The encoded key safe for filesystem storage.
+ */
+export const encodeStorageKey = (key: string): string =>
```
FrederikBolding, line 22: This doesn't account for `:` -> `-` which `redux-persist-filesystem-storage` will do under the hood.

But perhaps we don't care about that?

app/core/Engine/controllers/storage-service-init.ts
```diff
@@ -30,8 +34,11 @@ const mobileStorageAdapter: StorageAdapter = {
...
+      const encodedKey = encodeStorageKey(key);
+      const fullKey = `${STORAGE_KEY_PREFIX}${encodedNamespace}:${encodedKey}`;
```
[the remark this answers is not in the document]
FrederikBolding: This is accurate.

app/core/Engine/controllers/storage-service-init.ts
andrepimenta: Just because it's better safe than sorry. If someone uses "Some-Namespace" instead of "SomeNamespace" it will still work. And it's a simple implementation.

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: `-` is not normally encoded yeah

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: Sorry I implemented the namespace encoding after and forgot this. We actually do encode the namespace itself now, `storageService:[encoded]:[encoded]`.

Will fix

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: The reason I chose to encode - and `/` but not `:` is because we already have keys with colons in production (like tokensChainsCache:0x1), so encoding : would break existing data, while `-` and `/` are the characters that redux-persist-filesystem-storage actually corrupts. `:` is replaced by `-` and then back to `:` so that still works.

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: It doesn't encode `-`

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: Yeah we don't care. `:` -> `-` -> `:`. No corruption there.

As I mention before:
The reason I chose to encode - and `/` but not `:` is because we already have keys with colons in production (like tokensChainsCache:0x1), so encoding : would break existing data, while `-` and `/` are the characters that redux-persist-filesystem-storage actually corrupts. `:` is replaced by `-` and then back to `:` so that still works.

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: Fixed

app/store/migrations/119.ts
```diff
@@ -89,7 +90,9 @@ async function transformState(state: ValidState) {
     ).map(async (snap) => {
       const sourceCode = snap.sourceCode as string;
 
-      const fullKey = `${STORAGE_KEY_PREFIX}SnapController:${snap.id}`;
+      // Encode the snap ID to handle special characters (e.g., slashes and hyphens in npm:@metamask/bip32-keyring-snap)
+      const encodedSnapId = encodeStorageKey(snap.id as string);
+      const fullKey = `${STORAGE_KEY_PREFIX}SnapController:${encodedSnapId}`;
```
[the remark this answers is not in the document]
andrepimenta: It's not in production

app/core/Engine/utils/storage-service-utils.ts
FrederikBolding: That's not what I am asking, how did you choose the value for encoding `-`? It is valid in URIs so it isn't normally required to be encoded.

app/core/Engine/utils/storage-service-utils.ts
FrederikBolding: Sure, but what about other characters than `/` and `-`?

app/core/Engine/controllers/storage-service-init.ts
```diff
@@ -135,15 +155,18 @@ const mobileStorageAdapter: StorageAdapter = {
...
+      const encodedNamespace = encodeStorageKey(namespace);
+      const prefix = `${STORAGE_KEY_PREFIX}${encodedNamespace}:`;
 
-      await Promise.all(
-        keysToDelete.map((key) => FilesystemStorage.removeItem(key)),
+      const keysToDelete = allKeys.filter((rawKey) =>
+        rawKey.startsWith(prefix),
       );
 
-      Logger.log(
```
FrederikBolding, line 145: Why was this removed?

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: Since there is no normal encoding for `-`, I just followed the format `%XY` and picked `%2D`

app/core/Engine/utils/storage-service-utils.ts
andrepimenta: It encodes `:` as well I believe? So it would not be backwards compatible

app/core/Engine/controllers/storage-service-init.ts
andrepimenta: Removed by accident when removing other logs that I added for testing. Fixing now

app/core/Engine/utils/storage-service-utils.ts
FrederikBolding: True. I guess the worry is whether there are characters we haven't considered here that would be problematic in the future?

andrepimenta: I don't think so. I think only `-` because of the conversion, and `/` because it was creating folders. But I will create a ticket for us to do a bit more digging after but we could merge this for now to unblock the release if you are ok with it?

FrederikBolding (approved):

tommasini (approved):

PR #26418, opened 2026-02-23
feat: Show withdraw token + amount in Predict toast message

dan437:
## **Description**

The Predict withdraw completion toast was hardcoded to display "USDC" as the token symbol. This PR makes the toast dynamically resolve the actual withdrawn token and amount from the transaction metadata stored on-chain.

For post-quote withdrawals, the toast now uses `targetFiat` from the transaction's `metamaskPay` metadata (reflecting the actual received value after fees). The token symbol is resolved via `selectSingleTokenByAddressAndChainId`, falling back to the chain's native ticker via `selectTickerByChainId`.

The hardcoded `withdraw_completed_subtitle` i18n key (with "USDC") is replaced by `withdraw_any_token_completed_subtitle` which accepts a `{{token}}` placeholder. The old key is removed from all 15 locale files.

## **Changelog**

CHANGELOG entry: Updated Predict withdraw toast to show the actual token symbol and amount instead of hardcoded USDC

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/6986

## **Manual testing steps**

```gherkin
Feature: Predict withdraw toast shows correct token

  Scenario: user completes a post-quote withdraw of a non-USDC token
    Given user has a Predict position and initiates a withdrawal of USDT on BSC

    When the withdrawal transaction confirms
    Then the toast displays the fiat amount and "USDT" (not "USDC")

  Scenario: user completes a standard USDC withdraw
    Given user has a Predict position and initiates a USDC withdrawal

    When the withdrawal transaction confirms
    Then the toast displays the correct amount and "USDC"
```

---

> [!NOTE]
> **Medium Risk**
> Changes user-facing messaging for withdraw confirmations and introduces synchronous store reads + selector lookups, which could produce incorrect amounts/symbols if metadata is missing or malformed, though logic includes explicit fallbacks.
> 
> **Overview**
> Predict withdraw **success** toasts now optionally render a *post-quote* message that shows the resolved token symbol and amount, derived from the transaction’s `metamaskPay` metadata (via `store.getState()` + `selectTransactionMetadataById`), rather than being hardcoded to USDC.
> 
> Adds `getWithdrawConfirmedMessage` with fallbacks (non-post-quote/invalid `targetFiat`/missing token -> existing amount behavior and defaulting token to chain ticker or `USDC`), introduces a new i18n string `predict.withdraw.withdraw_any_token_completed_subtitle`, and extends the hook unit tests to cover these cases with mocked selectors/store.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a2a42371e2176f6361463aae7686c9f21dd9aa26. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


locales/languages/de.json
```diff
@@ -2261,7 +2261,6 @@
       "withdrawing": "Abhebung im Gange",
       "withdrawing_subtitle": "Verfügbar in etwa 1 Minute",
       "withdraw_completed": "Auszahlung abgeschlossen",
-      "withdraw_completed_subtitle": "{{amount}} USDC wurden Ihre Wallet hinzugefügt",
```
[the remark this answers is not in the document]
dan437: The i18n library falls back to en.json for missing keys. Non-English translations are handled by a separate localization team and don't need to be added in this PR.

app/components/UI/Predict/hooks/usePredictToastRegistrations.tsx
```diff
@@ -279,18 +284,37 @@ export const usePredictToastRegistrations = (): ToastRegistration[] => {
         }
 
         if (status === 'confirmed') {
-          const formattedWithdrawAmount = formatPrice(
-            amount ?? withdrawTransaction?.amount ?? 0,
-          );
+          const state = store.getState();
```
matthewwalsh0: If this is a React hook, should we be relying on `useSelector` rather than reading the store directly?

locales/languages/en.json
```diff
@@ -2275,7 +2275,7 @@
       "withdrawing": "Withdrawal in progress",
       "withdrawing_subtitle": "Available in about 1 minute",
       "withdraw_completed": "Withdrawal complete",
-      "withdraw_completed_subtitle": "{{amount}} USDC moved to your wallet",
```
matthewwalsh0, line 2278: As this is behind a feature flag, shouldn't this be an alternate label but we fallback to the original if no quote was used?

app/components/UI/Predict/hooks/usePredictToastRegistrations.tsx
matthewwalsh0: Given this extra complexity, is it worth a local `useWithdrawConfirmedMessage` or similar?

app/components/UI/Predict/hooks/usePredictToastRegistrations.tsx
```diff
@@ -279,18 +284,37 @@ export const usePredictToastRegistrations = (): ToastRegistration[] => {
...
 
           showSuccessToast({
             showToast,
             title: strings('predict.withdraw.withdraw_completed'),
             description: strings(
-              'predict.withdraw.withdraw_completed_subtitle',
-              {
-                amount: formattedWithdrawAmount,
-              },
+              'predict.withdraw.withdraw_any_token_completed_subtitle',
```
matthewwalsh0: As mentioned in label, should we fallback to original message if no metamask pay data?

matthewwalsh0 (changes requested):

app/components/UI/Predict/hooks/usePredictToastRegistrations.tsx
```diff
@@ -279,19 +281,17 @@ export const usePredictToastRegistrations = (): ToastRegistration[] => {
         }
 
         if (status === 'confirmed') {
-          const formattedWithdrawAmount = formatPrice(
-            amount ?? withdrawTransaction?.amount ?? 0,
+          const fallbackAmount = amount ?? withdrawTransaction?.amount ?? 0;
+          const { title, description } = getWithdrawConfirmedMessage(
+            store.getState(),
```
matthewwalsh0: We're still using the store directly which isn't ideal if we're in a React context so can use `useSelector` and pass whatever we need.

matthewwalsh0 (changes requested):

app/components/UI/Predict/utils/withdrawConfirmedMessage.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/**
+ * Derives the withdraw-confirmed toast message from transaction metadata.
+ *
+ * For post-quote withdrawals with valid metamaskPay data, displays the
+ * resolved token symbol and targetFiat amount. Otherwise falls back to the
+ * original USDC-only message.
+ */
+export function getWithdrawConfirmedMessage(
```
matthewwalsh0: Good that this is in a function, but is it excessive to export a util for it if it's very coupled to the toast?

I assumed it would be another `useX` hook inside the original file.

Will defer to the Predict team's preference.

dan437: I've moved it into usePredictToastRegistrations.tsx, but as a plain function, not a hook, because it's called inside an event callback, not during render.

app/components/UI/Predict/hooks/usePredictToastRegistrations.tsx
dan437: The transactionId comes from the event payload inside the callback, not at render time. And hooks (useSelector) can only run at the top level during render. Also, this function isn't a pure selector, it reads state via selectors, but then it constructs title and description for us. Ideally selectors should only derive data from state, not build UI strings.

matthewwalsh0: I don't mean put the whole thing inside a `useSelector`, but just access any state we need using a `useSelector` rather than the `store`.

So we could do a `useSelector` to get transactions, then we could find the relevant one in the callback using the ID?

caieu (dismissed):
LGTM

app/components/UI/Predict/hooks/usePredictToastRegistrations.tsx
dan437: I've discussed it with the Predict team and we agreed to keep store.getState() for now since it's used for 3 selectors inside the callback. They suggested to add a TODO comment until they refactor the hook.

matthewwalsh0 (dismissed):

caieu (approved):

PR #26430, opened 2026-02-23
feat(perps): add perps section to the new homepage

PatrykLucka:
## **Description**

Add a Perpetuals (Perps) positions section to the new wallet Homepage. When the perps feature flag is enabled, the section displays the user's open perps positions (up to 5) sourced from the PerpsController's cached data, which is populated by startMarketDataPreload() — a lightweight REST-based preload that doesn't require a full WebSocket connection or client initialization.

**Key changes:**

- **PerpsPositionRow** — new component displaying a single open position: token logo, direction (Long/Short) + symbol, leverage badge, TP/SL as percentage distance from entry, position value, and ROE.
- **PerpsPositionRowSkeleton** — skeleton placeholder (2 shimmer rows) shown while cachedPositions is loading.
- **PerpsSection** — updated to render position rows from Redux (selectCachedPositions), with skeleton loading state and navigation to market details (position tab) on row press.
- **selectCachedPositions / selectCachedMarketData** — new Redux selectors for cached controller state, necessary because the Homepage can't use the existing WebSocket-based hooks (usePerpsLivePositions, usePerpsHomeData) which require PerpsConnectionProvider/PerpsStreamProvider.
- All user-facing strings use existing i18n translations (perps.order.long_label, perps.order.short_label, perps.order.tp, perps.order.sl) plus a new homepage.positions.no_tp_sl key.
- Cached positions refresh every ~5 minutes via the controller's preload timer.

## **Changelog**

CHANGELOG entry: Added open perps positions section to the new wallet Homepage

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: Perps positions on Homepage

  Scenario: user sees open perps positions on the Homepage
    Given the user has the perps feature flag enabled
    And the user has open perps positions

    When user navigates to the new wallet Homepage
    Then the "Perpetuals" section is displayed
    And up to 5 open positions are shown with token logo, direction, symbol, leverage, TP/SL, value, and ROE

  Scenario: user sees skeleton loading while positions load
    Given the user has the perps feature flag enabled
    And cached positions have not loaded yet

    When user navigates to the new wallet Homepage
    Then 2 skeleton placeholder rows are displayed in the Perpetuals section

  Scenario: user taps a position row to navigate to market details
    Given the user has open perps positions on the Homepage

    When user taps on a position row
    Then the app navigates to the Market Details screen for that symbol
    And the "position" tab is selected

  Scenario: user taps section title to navigate to Perps home
    Given the Perpetuals section is visible on the Homepage

    When user taps the "Perpetuals" section title
    Then the app navigates to the Perps Home screen
```

### **After**
Screenshot 2026-02-23 at 13 24 05

https://github.com/user-attachments/assets/44442a7b-3a70-476c-959b-e2c8ac0a9e9a

---

> [!NOTE]
> **Medium Risk**
> Touches homepage rendering/navigation and introduces WebSocket-backed perps hooks + new animated UI components, which could impact performance or loading states if edge cases slip through; changes are well-covered by targeted unit tests.
> 
> **Overview**
> Adds a new **Homepage `Perpetuals` section** (replacing the previous placeholder `Sections/Perps`) that is gated by the perps feature flag and wrapped in `PerpsConnectionProvider`/`PerpsStreamProvider` keyed by selected account.
> 
> When data is available, the section shows up to **5 items** (positions first, then orders), with navigation to `MARKET_DETAILS` (position tab) from a position row and to perps home from the title. When the user has no positions/orders, it instead shows a **trending markets horizontal carousel** (top by volume) with sparkline charts, plus a "More perps" tile; loading uses a skeleton and avoids empty-content flashes.
> 
> Enhances `PerpsPositionCard` compact mode with a new `compactVariant="position"` that displays direction + symbol, a leverage badge, **TP/SL % from entry** (via new `buildTpSlLabel`), ROE coloring, and a TP/SL skeleton placeholder. Adds supporting UI components (`PerpsMarketTileCard`, `SparklineChart`, `FadingScrollContainer`, skeletons) plus tests and new i18n strings (`homepage.sections.more_perps`, `homepage.sections.positions.no_tp_sl`), and updates homepage `ErrorState` to use an image asset instead of a DS icon.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7b64e1c040a0d8886c99436aac13bdc9442f6946. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const title = strings('homepage.sections.perpetuals');
+  const [error, setError] = useState<boolean>(false);
+
+  const startPreload = useCallback(async () => {
+    const controller = Engine.context.PerpsController;
+    if (!controller) return;
+
+    setError(false);
+    try {
+      await controller.startMarketDataPreload();
```
abretonc7s: not async no need to await

app/components/Views/Homepage/Sections/Perpetuals/components/PerpsPositionRow/PerpsPositionRow.tsx
```diff
@@ -0,0 +1,127 @@
...
+import styles from './PerpsPositionRow.styles';
+import type { PerpsPositionRowProps } from './PerpsPositionRow.types';
+
+const ICON_SIZE = 36;
+
+/**
+ * Build a human-readable TP/SL label as percentage distance from entry.
+ * Returns null when neither TP nor SL is configured.
+ */
+export function buildTpSlLabel(
```
abretonc7s: This should probably go to positionCalculations.ts as an util to separate concerns

app/components/Views/Homepage/Sections/Perpetuals/components/PerpsPositionRow/PerpsPositionRow.tsx
```diff
@@ -0,0 +1,127 @@
...
+          color={isPositive ? TextColor.Success : TextColor.Error}
+        >
+          {roeDisplay}
+        </Text>
+      </View>
+    </TouchableOpacity>
+  );
+};
+
+export default React.memo(PerpsPositionRow);
```
abretonc7s: isnt this similar to PerpsPositionCard with compact mode? Probably we can reuse existing component.

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
abretonc7s: Also the try catch wouldnt do anything here, the method is designed as fire-and-forget

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
```diff
@@ -0,0 +1,143 @@
...
+              />
+            ))}
+          </View>
+        )}
+      </SectionRow>
+    </Box>
+  );
+});
+
+export default PerpsSection;
```
abretonc7s, line 249: I think we need to subscribe to PerpsCacheInvalidator to update the position after they change otherwise the user could have invalid state for up to 5min (eg: still see a position after they close it).
See pattern for usePerpsPositionForAsset.ts

abretonc7s (changes requested):

abretonc7s (changes requested):
First pass review — haven't tested locally yet but found a few things that need to be addressed before merging.

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
```diff
@@ -0,0 +1,143 @@
...
+    } catch {
+      setError(true);
+    }
+  }, []);
+
+  useEffect(() => {
+    if (!isPerpsEnabled) return;
+
+    startPreload();
+    return () => Engine.context.PerpsController?.stopMarketDataPreload();
```
abretonc7s: There may be a bug here where it could kill the preload timer for other pages. I think it should be handled by the wallet tab and not do anything around preloading stopPreload int his component.

abretonc7s (dismissed):

locales/languages/en.json
```diff
@@ -7707,8 +7707,12 @@
       "predictions": "Predictions",
       "defi": "DeFi",
       "nfts": "NFTs",
+      "trending_perps": "Trending Perps",
```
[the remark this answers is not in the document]
wachunei: Please remove this as the title remains the same in the section

app/images/wifi-off.png
gambinish, line 1: Just confirming whether or not that this icon is on-brand and provided from design team

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
```diff
@@ -0,0 +1,158 @@
...
+                onPress={() => handlePositionPress(position)}
+                testID={`perps-position-row-${position.symbol}`}
+              />
+            ))}
+          </View>
+        )}
+      </SectionRow>
+    </Box>
+  );
+});
```
gambinish, line 247: We should include an empty state here for when `positions.length` is 0.

gambinish (changes requested):
The PR uses `startMarketDataPreload()` (REST-based, 5-minute polling) instead of the WebSocket-based `PerpsConnectionProvider/PerpsStreamProvider` stack. This is probably fine for checking to see if a user has an active position for a given asset pair. In fact, this was the initial intent of this "standalone" lightweight fetch on the new token details page.

However, I don't think it is sufficient in this case, where live up-to-date data is critical to the user. This is because the `PerpsCacheInvalidator` only fires on explicit user actions (close position, modify TP/SL). It does NOT fire on external events like price movements or auto-liquidations.

My main concern, is that we are surfacing stale position/price data between refreshes, and that the 5-minute polling interval is a significant gap for trading, especially for high volume traders.

Between refreshes data on the homepage can become stale, in comparison to the live socket data we surface in the perps detail page. Note the ROE for BNB:

Screenshot 2026-02-24 at 3 19 58 PM
Screenshot 2026-02-24 at 3 19 48 PM



This discrepency happens because position values (positionValue, unrealizedPnl, returnOnEquity) are computed at preload time but won't reflect real-time mark price changes. A position showing +9.4% ROE could actually be at -5% or liquidated for a highly volatile asset. This means that liquidation risk is hidden -- if the mark price moves toward a user's liquidation threshold, the homepage will show stale values for up to 5 minutes.

My recommendation is to consider wrapping just the `PerpsSection` in the WebSocket provider. Since this new section is effectively replacing the `PerpsTab` we should implement the same data patterns to avoid regressions.

The `PerpsStreamManager` already supports reference counting and per-subscriber throttling, so we can always dial this in as needed. This would bring the homepage in line with the existing PerpsTab.

Regarding the missing TP/SL. This is somewhat related in that the standalone REST path explicitly skips TP/SL fetching to avoid getting 429 errors. This means position TP/SL will always be undefined on cached positions. Consequently: `buildTpSlLabel()` will always return null, and every single position row will show "No TP/SL" even when TP/SL orders are configured.

To fix this properly while only using the REST fetch, we'd need to extract TP/SL from the orders that are already fetched via `getOpenOrders({ standalone: true })` in `#performUserDataPreload()` and merge them into cached positions before storing them in state (this is similar to how the WebSocket subscription service does it with `#extractTPSLFromOrders` + `#mergeTPSLIntoPositions`). However, while this can be done with the standalone fetch, I recommend just using websocket subscription as mentioned above.

Lastly, it appears that this Perpetuals section only supports positions, but we should also be showing limit orders when they exist, in the same way that the Perps tab does today.

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
```diff
@@ -0,0 +1,158 @@
...
+              <PerpsPositionCard
+                key={position.symbol}
+                position={position}
+                compact
+                compactVariant="position"
+                iconSize={36}
+                onPress={() => handlePositionPress(position)}
+                testID={`perps-position-row-${position.symbol}`}
+              />
+            ))}
```
gambinish: We should also list orders here, in the perps tab we have a clear deliniation between market positions and limit orders

app/images/wifi-off.png
PatrykLucka: Thanks for double-checking! Yes, it is from this designs file: https://www.figma.com/design/z0panHXrMSMUSof2SaPkd4/Home-2026?node-id=3060-12825&t=1dzvFuB4lAO76GWF-0

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
PatrykLucka: I've added missing part - now we're showing trending perps cards if positions.length is 0.

app/components/Views/Homepage/components/FadingScrollContainer/FadingScrollContainer.tsx
wachunei, line 1: nit: in a next PR we should use this in the Predictions section too.

app/components/Views/Homepage/Sections/Perpetuals/components/PerpsMarketTileCard/PerpsMarketTileCard.tsx
```diff
@@ -0,0 +1,201 @@
...
+    if (!isSparklineLoading) {
+      shimmerOpacity.stopAnimation();
+      return;
+    }
+
+    const pulse = () => {
+      Animated.sequence([
+        Animated.timing(shimmerOpacity, {
+          toValue: 1,
+          duration: 900,
```
wachunei: nit: move duration to a constant at the top of the file for easier tweaks

wachunei (approved):

gambinish (approved):
Approving with one small nit (we can address in a separate PR). We are currently rounding the ROE percentage to 1 decimal place. However in Perps details we round to 2 decimal place precision. It appears we also do this on the Tokens section so it makes sense for us to align this rounding convention.

Image

PR #26459, opened 2026-02-24
feat: generic advanced charts component

sahar-fehri:
## **Description**

Introduces a reusable, generic `AdvancedChart` component built on [TradingView Advanced Charts](https://www.tradingview.com/charting-library-docs/latest/getting_started/) rendered inside a WebView. This component is designed to be consumed by multiple features (Token Details, Perps, etc.) with a composable props API, each consumer uses only the props it needs.

This PR is just for testing this new component: https://github.com/MetaMask/metamask-mobile/pull/26465

### What's included

**Core component (`AdvancedChart.tsx`)**
- WebView-based TradingView charting widget with full candlestick/OHLCV support
- Composable props API: `ohlcvData`, `indicators`, `positionLines`, `showVolume`, `chartType`, etc.
- Imperative ref handle for one-off actions (`addIndicator`, `removeIndicator`, `setChartType`, `reset`)
- Bidirectional message protocol between React Native and WebView (`RNToWebViewMessage` / `WebViewToRNMessage`)
- Error handling with retry-on-recovery and loading states
- Dynamic resolution detection: automatically adapts bar width based on incoming data interval

**WebView logic (`webview/chartLogic.js`)**
- Custom TradingView datafeed implementation (`onReady`, `getBars`, `subscribeBars`, `unsubscribeBars`)
- Widget lifecycle management: creates, destroys, and recreates the TradingView widget when the data resolution changes (e.g., switching from 1D to 1H)
- Volume study with themed colors
- Technical indicator support (MACD, RSI, MA200) via `createStudy` / `removeStudy`
- Position lines overlay (entry, take-profit, stop-loss, liquidation) for Perps use case
- Real-time bar updates via `realtimeCallback`
- Crosshair tracking with bar-snapping for OHLC data forwarding
- All unnecessary TradingView UI elements disabled via `disabled_features`

**HTML template (`AdvancedChartTemplate.ts`)**
- Generates WebView HTML with injected theme colors and feature flags
- Configurable `CHARTING_LIBRARY_URL` for local dev (`localhost:8000`) or S3 deployment

**Time range selector (`TimeRangeSelector.tsx`)**
- Time range options: 1H, 1D, 1W, 1M, YTD, ALL
- Each range maps to a Hyperliquid candle interval and count (e.g., 1H → 1m candles × 60)
- Exports `TIME_RANGE_CONFIGS` for consumer data fetching

**Indicator toggle (`IndicatorToggle.tsx`)**
- Toggle buttons for MACD, RSI, MA200 technical indicators
- Designed for Token Details chart header

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> **Medium Risk**
> Introduces a new WebView-based chart surface that loads the TradingView library from a configurable CDN and bridges messages between JS and React Native, which carries moderate risk around WebView security/CSP correctness and runtime stability.
> 
> **Overview**
> Adds a reusable `AdvancedChart` component that renders TradingView Advanced Charts in a `WebView`, with a typed RN↔WebView message protocol to sync OHLCV data, realtime ticks, indicators, chart type, volume visibility, and optional position-line overlays.
> 
> Introduces a generated WebView script pipeline (`webview/chartLogic.js` → `chartLogicString.ts` via `syncChartLogic.js`) plus an HTML template (`AdvancedChartTemplate.ts`) that injects theme/feature config, CSP, and a configurable charting-library base URL (`MM_CHARTING_LIBRARY_URL`).
> 
> Adds a `TimeRangeSelector` utility component with predefined range→interval/count mappings, updates `.eslintignore` for the WebView scripts, and includes unit tests covering chart message/ref behavior and the selector.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3785ec118c29361cfd4260fe91207784e41843eb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Charts/AdvancedChart/webview/chartLogicString.ts
```diff
@@ -0,0 +1,794 @@
...
+  try {
+    var message =
+      typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
+
+    if (!window.isChartReady && message.type !== 'SET_OHLCV_DATA') {
+      window.pendingMessages.push(message);
+      return;
+    }
+
+    switch (message.type) {
```
sahar-fehri, line 61: This file runs inside the WebView to drive the TradingView chart. 
It:
1- listens for messages exp TOGGLE_VOLUME from RN and routes them to handlers.
2- Feeds data to TradingView and implements the customDatafeed object.
3- Initializes the widget 
4-Sends events back to React Native (reports via postMessage).
it's the bridge between React Native and TradingView

app/components/UI/Charts/AdvancedChart/webview/chartLogicString.ts
```diff
@@ -0,0 +1,794 @@
+/**
+ * AUTO-GENERATED - DO NOT EDIT DIRECTLY
+ *
+ * This file is generated from chartLogic.js by syncChartLogic.js
+ * Edit chartLogic.js instead, then run:
+ *   node app/components/UI/Charts/AdvancedChart/webview/syncChartLogic.js
+ */
+
+// eslint-disable-next-line import/no-default-export
+export default `/**
```
sahar-fehri, line 10: This is an auto-generated file; it's the entire contents of chartLogic.js exported as a single template literal string. This lets the React Native side embed the chart logic directly into the WebView's HTML without needing to load a separate .js file.
It's generated by running syncChartLogic.js, which reads chartLogic.js and wraps it in export default \...\``. You never edit this file directly, you edit chartLogic.js and re-run the sync script.

app/components/UI/Charts/AdvancedChart/webview/syncChartLogic.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Run this after editing chartLogic.js:
+ *   node app/components/UI/Charts/AdvancedChart/webview/syncChartLogic.js
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const sourceFile = path.join(__dirname, 'chartLogic.js');
+const targetFile = path.join(__dirname, 'chartLogicString.ts');
```
sahar-fehri, line 14: This file is a dev-time build script only;  you run it locally after editing chartLogic.js to regenerate chartLogicString.ts.

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
```diff
@@ -0,0 +1,352 @@
...
+        },
+      }),
+      [addIndicator, removeIndicator, setChartTypeInternal],
+    );
+
+    // ---- Declarative prop syncing ----
+
+    useEffect(() => {
+      if (ohlcvData.length > 0 && webViewLoadedRef.current) {
+        sendOHLCVData(ohlcvData);
```
sahar-fehri, line 250: This pushes data and commands into the WebView via postMessage And it receives events back from the WebView via the onMessage prop on the <WebView> component

.eslintignore
```diff
@@ -1,5 +1,8 @@
 /scripts/inpage-bridge
 /app/core/InpageBridgeWeb3.js
+/app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
+/app/components/UI/Charts/AdvancedChart/webview/chartLogicString.ts
+/app/components/UI/Charts/AdvancedChart/webview/syncChartLogic.js
```
sahar-fehri, line 5: Ignored lint for these files
chartLogic.js this is code for the WebView, not React Native.
chartLogicString.ts is auto-generated.
syncChartLogic.js this is the Node.js script.

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+`;
+
+/**
+ * Creates the HTML template for TradingView Advanced Charts.
+ *
+ * @param theme - MetaMask theme for styling
+ * @param features - Optional feature flags forwarded to the WebView
+ */
+export const createAdvancedChartTemplate = (
```
sahar-fehri, line 77: AdvancedChart.tsx calls `createAdvancedChartTemplate` to build the HTML, then feeds it to the WebView. It generates the HTML, passes it to <WebView source={{ html }}>, and the WebView renders it as a local page

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
```diff
@@ -0,0 +1,104 @@
...
+import { chartLogicScript } from './webview';
+
+// AWS S3 - CORS configured to allow origin: null from WebView iframes
+// const CHARTING_LIBRARY_URL =
+//   'https://va-mmcx-terminal.s3.us-east-2.amazonaws.com/charting_library/charting_library/';
+
+// Local development server fallback:
+//   npx http-server --cors -p 8000 <dir-containing-charting_library/>
+export const CHARTING_LIBRARY_BASE_URL = 'http://localhost:8000/';
+const CHARTING_LIBRARY_URL = `${CHARTING_LIBRARY_BASE_URL}charting_library/`;
```
sahar-fehri, line 18: The url needs to be fixed and replaced here https://va-mmcx-terminal.s3.us-east-2.amazonaws.com/charting_library/charting_library/

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
```diff
@@ -0,0 +1,263 @@
...
+}
+
+/**
+ * Props for the IndicatorToggle component
+ */
+export interface IndicatorToggleProps {
+  activeIndicators: IndicatorType[];
+  onToggle: (indicator: IndicatorType) => void;
+  disabled?: boolean;
+}
```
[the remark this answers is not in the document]
sahar-fehri: This is fine; can be used in integration

app/components/UI/Charts/AdvancedChart/webview/syncChartLogic.js
```diff
@@ -0,0 +1,31 @@
...
+
+const jsContent = fs.readFileSync(sourceFile, 'utf8');
```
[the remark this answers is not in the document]
sahar-fehri: false issue the source file should always exist

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
Prithpal-Sooriya, line 1: OMG Webview JS injection 🙈 

I think we are mostly safe here, but it will be worth getting mobile experts view on this architecture.

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
```diff
@@ -0,0 +1,112 @@
+import type { Theme } from '../../../../util/theme/models';
+import { chartLogicScript } from './webview';
+
+// AWS S3 - CORS configured to allow origin: null from WebView iframes
+export const CHARTING_LIBRARY_BASE_URL =
+  'https://va-mmcx-terminal.s3.us-east-2.amazonaws.com/charting_library/';
```
Prithpal-Sooriya: This is public, so... okay?

But can you confirm? Otherwise we can move to env

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
geositta: Recommend using Content-Security-Policy to restrict the source to our hosted chart endpoint. We could also configure Cloudfront to send the Content-Security-Policy header automatically.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,804 @@
...
+            open: bar.open,
+            high: bar.high,
+            low: bar.low,
+            close: bar.close,
+            volume: bar.volume,
+          };
+        });
+        onResult(bars, { noData: bars.length === 0 });
+      } else {
+        onResult([], { noData: true });
```
geositta, line 652: This tells TradingView that no older candles exist, so left-scroll pagination never loads historical data. The TradingView docs (in `docs/charting-library-context.txt` around line 35640) are explicit that `noData` should only be set to `true` when there is genuinely no older data on the server, and that the library will call `getBars` multiple times to fill the visible area, returning empty on every non-first request short-circuits that entirely.

firstDataRequest: false is hit when TradingView asks again for more data, typically during:

  1. user pans left for older candles,
  2. TradingView does multi-pass fetches because first response didn’t satisfy countBack,
  3. indicators request extra historical bars.

Since the else branch is unconditional, all those follow-up calls are treated as "no more history exists," even when you intended to load more.

Related to history is my comment on `onRequestMoreHistory` in AdvancedChart.tsx

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
```diff
@@ -0,0 +1,371 @@
...
+            break;
+
+          case 'CROSSHAIR_MOVE':
+            onCrosshairMove?.(
+              (hasPayload ? (payload.data as CrosshairData) : null) ?? null,
+            );
+            break;
+
+          case 'NEED_MORE_HISTORY':
+            onRequestMoreHistory?.();
```
geositta: RN-side message protocol declares `NEED_MORE_HISTORY` as a valid `WebViewToRNMessageType` in `AdvancedChart.types.ts`. Howver, the webview never actually emits a `NEED_MORE_HISTORY` message , since `sendToReactNative('NEED_MORE_HISTORY', ...)` does not exist anywhere in `chartLogic.js`. So history loading is declared in the protocol but never triggered.

In the Perps integrations in metamask-mobile solves this by subscribing to `visibleLogicalRangeChange`, detecting when the user has scrolled within 5 candles of the left edge, throttling with a cooldown, and emitting `NEED_MORE_HISTORY`. The Terminal web app's `TradingViewDatafeed` takes the simpler approach of calling `dataSource.fetchHistoricalCandles({ symbol, resolution, from, to })` for every `getBars` request and only returning `noData: true` when the candle array is genuinely empty.

If this is intentionally deferred to a follow-up PR, that is completely reasonable. Can we add a brief comment in `chartLogic.js` near the `getBars` function (and a TODO in the types file next to the `NEED_MORE_HISTORY` declaration) noting that history pagination is not yet wired up? That way the next person who uses `onRequestMoreHistory` does not spend time debugging why their callback never fires.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,804 @@
...
+      }
+    } catch (error) {
+      onError(error.message);
+    }
+  },
+
+  subscribeBars: function (symbolInfo, resolution, onTick, listenerGuid) {
```
geositta, line 669: `subscribeBars` receives `listenerGuid` as its fourth argument but ignores it entirely, storing the callback in a single `window.realtimeCallback`. Then `unsubscribeBars` nulls out that global without checking which subscription it should remove.

TradingView's datafeed contract allows multiple concurrent subscriptions during symbol or resolution transitions. The library may call `subscribeBars` for a new resolution before calling `unsubscribeBars` for the old one. With a single global callback, the new subscription silently overwrites the previous one, and a subsequent `unsubscribeBars` for the old subscription clears the active callback, dropping all realtime updates until the next subscribe cycle. This is unlikely to cause visible issues today since the component uses a single symbol and resolution changes rebuild the widget, but it will become a real problem if the component evolves to support resolution switching without full widget recreation, which is exactly the Perps migration use case.

The Terminal web app's `MockTradingViewDatafeed` at `app/components/UI/Charts/tradingview/MockTradingViewDatafeed.ts` lines 199-226 already implements this correctly: it stores subscriptions in a `Map<string, { unsubscribe, onResetCacheNeededCallback }>` keyed by `subscriberUID`, and `unsubscribeBars` looks up by UID, calls `unsubscribe()`, and deletes the entry.

Would it work to follow the same pattern here? Something like `window.realtimeCallbacks = {}` at the top, `window.realtimeCallbacks[listenerGuid] = onTick` in `subscribeBars`, `delete window.realtimeCallbacks[listenerGuid]` in `unsubscribeBars`, and iterating over all active callbacks in `handleRealtimeUpdate`.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,804 @@
...
+
+  if (window.ohlcvData.length === 0) {
+    return;
+  }
+
+  try {
+    var theme = window.CONFIG.theme;
+    var features = window.CONFIG.features || {};
+
+    var disabledFeatures = [
```
geositta: The widget initialization at lines 651-674 unconditionally disables 22 TradingView features including `header_widget`, `timeframes_toolbar`, `header_symbol_search`, `legend_widget`, and `go_to_date`. Furthermore, the indicator system at lines 235-249 in this same file uses a hardcoded switch statement mapping exactly three names (`MACD`, `RSI`, `MA200`) to TradingView study constructors, and `AdvancedChart.types.ts` line 26 defines `IndicatorType` as `'MACD' | 'RSI' | 'MA200'`.

This is a sensible curated mobile UX for Token Details, but the component is named and documented as a "generic" AdvancedChart intended for Perps migration as well. TradingView already provides its own indicator picker and study management UI -  disabling it and reimplementing a subset is added maintenance surface. If Perps (or a future consumer) needs TradingView's native study picker, additional indicators like Bollinger Bands, or the timeframes toolbar, they would need to fork or significantly rework both the JS and the type layer.

I do not think this needs to block the PR. Would it be helpful to either rename the component to something that sets expectations accurately (like `CuratedChart` or `MobileChart`), or add a `disabledFeatures` override prop (defaulting to the current list) so future consumers can opt back into TradingView's native capabilities without modifying the core? Even just a comment at the top of the disabled features array noting "these defaults are optimized for the Token Details mobile UX; consumers needing TradingView native UI should override via props" would go a long way.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,804 @@
...
+    }
+    window.positionShapeIds = [];
+  } catch (error) {
+    sendToReactNative('ERROR', {
+      message: 'Failed to clear position lines: ' + error.message,
+    });
+  }
+}
+
+function handleSetPositionLines(payload) {
```
geositta, line 336: The `PositionLines` interface supports `entryPrice`, `takeProfitPrice`, `stopLossPrice`, and `liquidationPrice` but does not include `currentPrice`. The `handleSetPositionLines` checks for each of those four prices but has no branch for a current-price line. The existing Perps `TPSLLines` type at `TradingViewChart.tsx` line 32 includes `currentPrice?: string` and the Perps chart renders it as a distinct dashed horizontal line with axis label via a dedicated `updateCurrentPriceLine` function.

If Perps migrates to this generic component, it would need `currentPrice` support added to `PositionLines`, to `handleSetPositionLines`, and to the line-creation logic. Can we add `currentPrice?: number` to the `PositionLines` interface now, even if the rendering logic comes in a follow-up? That way the type contract is correct from the start and Perps migration does not require a breaking interface change later.

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /** External loading state */
+  isLoading?: boolean;
+}
+
+/**
+ * Imperative ref handle for AdvancedChart.
+ * Use props for declarative control; ref for one-off imperative actions.
+ */
+export interface AdvancedChartRef {
```
geositta, line 335: AdvancedChartRef is in a good place for the current scope, but I want to flag a Perps migration gap so we can plan for it.

In app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts:211, the ref currently exposes addIndicator, removeIndicator, setChartType, and reset. In the existing Perps flow, we rely on viewport oriented controls (zoomToLatestCandle on interval/fullscreen transitions and resetToDefault on refresh/interval reset) in app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx:382, app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx:533, and app/components/UI/Perps/components/PerpsChartFullscreenModal/PerpsChartFullscreenModal.tsx:84.

Advanced Charts can support those behaviors, so this is a wrapper/API gap rather than a platform limitation. I don’t think this should block this PR, but I recommend we avoid adding placeholder ref methods without implementation. A better follow-up is to add concrete ref methods (for example `zoomToLatestCandle(candleCount?)` and `resetViewport/resetToDefault`) with corresponding WebView message handlers so the interface remains truthful and migration ready.

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export interface RealtimeUpdatePayload {
+  bar: OHLCVBar;
+}
+
+export interface ToggleVolumePayload {
+  visible: boolean;
+}
+
+export interface RNToWebViewMessage {
```
geositta: `RNToWebViewMessage` pairs `type: RNToWebViewMessageType` (a plain string union) with `payload: SetOHLCVDataPayload | AddIndicatorPayload | ... | Record<string, never>`. Because these are independent fields rather than a discriminated union, TypeScript cannot narrow the payload based on the type in a switch branch.

A message like `{ type: 'ADD_INDICATOR', payload: { bar: someOHLCVBar } }` type-checks successfully even though that combination is semantically invalid.

The same pattern applies to `WebViewToRNMessage` at lines 153-161.

Modeling these as a discriminated union (`{ type: 'ADD_INDICATOR'; payload: AddIndicatorPayload } | { type: 'SET_OHLCV_DATA'; payload: SetOHLCVDataPayload } | ...`) would let TypeScript enforce valid pairings and eliminate the downstream casts.

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
```diff
@@ -0,0 +1,371 @@
...
+          type: 'SET_CHART_TYPE',
+          payload: { type },
+        });
+      },
+      [isChartReady, postMessage],
+    );
+
+    // ---- WebView message handling ----
+
+    const handleMessage = useCallback(
```
geositta, line 148: `handleMessage` parses `event.nativeEvent.data` with `JSON.parse`, then destructures `type` and `payload` from the untyped result. The switch branches use assertions; `payload.name as IndicatorType`, `payload.data as CrosshairData`, rather than runtime type guards to narrow the parsed data.

The manual `hasPayload && 'name' in payload` checks help, but they confirm structural presence without validating the value. Since this is a serialization boundary (data arrives from a WebView over `postMessage`), this is exactly where TypeScript's narrowing would be most valuable.

A small validator function (or even a discriminated union paired with `zod`/`io-ts` or a manual narrower) would turn these casts into safe narrowing and catch malformed messages early.

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
```diff
@@ -0,0 +1,225 @@
...
+  type: RNToWebViewMessageType;
```
geositta: Both `RNToWebViewMessage` and `WebViewToRNMessage` include `Record<string, never>` as a payload variant. The intent is to model message types that carry no payload (like `CHART_READY`), but `Record<string, never>` is structurally compatible with `{}`, which in TypeScript's structural type system is compatible with almost any object.

This means the union effectively becomes "any object," removing the constraint that the payload variants were meant to provide. A `null` payload (or omitting the field entirely on no-payload message types) would express the intent more precisely.

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
```diff
@@ -0,0 +1,225 @@
...
+  high: number;
+  low: number;
+  close: number;
+  volume?: number;
+}
+
+/**
+ * Chart type enum matching TradingView SeriesType
+ */
+export enum ChartType {
```
geositta: There have been [recent discussions](https://consensys.slack.com/archives/C1L7H42BT/p1770403410859309?thread_ts=1770400941.090849&cid=C1L7H42BT) around the removal of enums from the codebase.

`ChartType` is a numeric enum (`Candles = 1, Line = 2`). These values do correspond to TradingView's `SeriesType` constants, so the numbers are not arbitrary. That said, numeric enums in TypeScript have well-known ergonomic issues. They allow reverse lookups (`ChartType[1] === 'Candles'`), produce runtime code, and make debugging bridge messages harder since the raw message shows `{ type: 1 }` rather than `{ type: 'candles' }`.

An `as const` object (`const ChartType = { Candles: 1, Line: 2 } as const`) preserves the numeric values for TradingView while avoiding the enum pitfalls. This is a minor preference but worth considering since the bridge is the primary consumer.

app/components/UI/Charts/AdvancedChart/TimeRangeSelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+  BoxJustifyContent,
+} from '@metamask/design-system-react-native';
+import { useStyles } from '../../../../component-library/hooks';
+import { Theme } from '../../../../util/theme/models';
+
+export type TimeRange = '1H' | '1D' | '1W' | '1M' | 'YTD' | 'ALL';
+
+export interface TimeRangeConfig {
+  /** Hyperliquid candle interval */
+  hlInterval: string;
```
geositta: Nit: `TimeRangeConfig.hlInterval` is typed as `string`. On this branch, the only values are hardcoded literals (`'1m'`, `'15m'`, `'1h'`, `'4h'`, `'1d'`), so there is no practical risk today. However, when a consumer eventually passes `hlInterval` to a fetch layer that expects a specific interval union (as `useHyperliquidCandles` does on the test branch), the `string` type will silently widen past whatever validation the consumer enforces. Typing it as a literal union (or importing and reusing the interval type from the data layer) would catch mismatches at the call site.

geositta (changes requested):
The architectural direction here is strong, and I agree the generic TradingView WebView wrapper is the right foundation for serving both Token Details and the eventual Perps migration. The component's composable props API, declarative indicator syncing, position line overlay, and crosshair data forwarding are well-designed. The `syncChartLogic.js` dev workflow for keeping the JS and its string export in sync is a nice touch.

Before merge, I think the two datafeed contract issues are worth addressing or explicitly deferring with inline TODOs: `getBars` currently hard-stops history after the first request (which prevents left-scroll pagination), and `subscribeBars` uses a single global callback ignoring `listenerGuid` (which can drop realtime updates during resolution transitions). If these are intentionally deferred to a follow-up, documenting that in the code would prevent confusion.

I think that particularly in Perps that we would likely want the option to display the customary TradingView indicator set that users are accustomed to seeing on HyperLiquid, or on CEX mobile apps.

Having said that, the Perps migration gaps (missing `currentPrice` in `PositionLines`, missing zoom/reset in `AdvancedChartRef`, hardcoded feature/indicator set, and the "generic" naming question) are worth noting for planning but I don't think they block this PR.

On the TypeScript side, the message protocol types model `type` and `payload` as independent fields rather than a discriminated union, and the WebView message handler uses casts instead of narrowing on parsed JSON. These are not causing bugs today, but they create a false sense of type safety across the bridge. The types compile but do not actually constrain the runtime data. Strengthening these as the component matures (especially before other teams integrate against the message protocol) would reduce the surface for silent mismatches.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
sahar-fehri: [49c198d](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/49c198d78dab88df788aaf1bef092528a74a6937)
Fixed. getBars now implements proper asynchronous history loading:
On first request, returns all available bars from the local dataset
On subsequent requests (scroll-back)

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
sahar-fehri: [49c198d](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/49c198d78dab88df788aaf1bef092528a74a6937)
Fixed. chartLogic.js now emits NEED_MORE_HISTORY from within getBars whenever it cannot fulfill a request with the currently available data.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
sahar-fehri: [49c198d](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/49c198d78dab88df788aaf1bef092528a74a6937)
Fixed. subscribeBars now stores each callback in a window.realtimeCallbacks map keyed by listenerGuid. unsubscribeBars deletes the entry by guid

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
sahar-fehri: I like option 2: add a disabledFeatures override prop
Will fix

sahar-fehri: [d7dfdae](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/d7dfdaec3a9cfd96437f4061b460ca5085f55340)

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
sahar-fehri: [bd82b80](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/bd82b8025b4da4d952d968ba26c0382dd8192a8a)

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
```diff
@@ -0,0 +1,114 @@
...
+window.CONFIG = {
+  libraryUrl: '${libraryUrl}',
+  theme: {
+    backgroundColor: '${theme.colors.background.default}',
+    borderColor: '${stripHexAlpha(theme.colors.border.muted)}',
+    textColor: '${theme.colors.text.alternative}',
+    successColor: '${theme.colors.success.default}',
+    errorColor: '${theme.colors.error.default}',
+    primaryColor: '${theme.colors.primary.default}'
+  },
```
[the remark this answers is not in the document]
sahar-fehri: The five unwrapped colors currently resolve to standard 7-char hex values via brandColor references, so there's no active bug today.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,860 @@
...
+      try {
+        window.chartWidget
+          .activeChart()
+          .setResolution(newResolution, function () {});
+      } catch (e) {
+        window.chartWidget.remove();
+        window.chartWidget = null;
+        window.isChartReady = false;
+        initChart();
+      }
```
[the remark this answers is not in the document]
sahar-fehri: [a43cbb3](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/a43cbb3210ba9397598f921177c639c250cc2114)

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
sahar-fehri: I understand that this comment does not require any changes, right?
Advanced Charts supports these behaviors (e.g., chart.setVisibleRange(), chart.resetData()), so  when it's time, we add real implementations with corresponding WebView message handlers (e.g., a ZOOM_TO_LATEST message type handled in chartLogic.js.

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
sahar-fehri: [9c61f3a](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/9c61f3a0dd38176cebb8f11833d62229ea3219f3)

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
sahar-fehri: Will go with the first option; seems like a good middle ground without having to introduce new dependencies (zod/io-ts)

sahar-fehri: [ccf84a6](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/ccf84a6878ebdbd45c10b1b23cbd55f6d61f5874)

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
sahar-fehri: Fixed in previous commits and in here [b086c13](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/b086c13e5a41e354fbd2d4c2ac9623f0598c9abc)

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
sahar-fehri: Updated [b670303](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/b67030381d40cdaecf6227f5afe4ac78e5edb2ee) 🙏

app/components/UI/Charts/AdvancedChart/TimeRangeSelector.tsx
sahar-fehri: [595dc83](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/595dc83bf74e6102ccddd073f649c1e8c8f11732) 🙏

app/components/UI/Charts/AdvancedChart/TimeRangeSelector.tsx
```diff
@@ -0,0 +1,111 @@
...
+};
+
+export const TIME_RANGE_CONFIGS: Record<TimeRange, TimeRangeConfig> = {
+  '1H': { hlInterval: '1m', count: 60 },
+  '1D': { hlInterval: '15m', count: 96 },
+  '1W': { hlInterval: '1h', count: 168 },
+  '1M': { hlInterval: '4h', count: 180 },
+  get YTD() {
+    return { hlInterval: '1d' as const, count: Math.min(ytdDays(), 500) };
+  },
```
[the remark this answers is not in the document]
sahar-fehri: [771480d](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/771480db77d8f766b47443dd66d7dd944e1dd2a0)

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
sahar-fehri: Updated here [2120444](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/2120444e2848b4c808093a25f4692825db3234f7)
@geositta could you help me check what would the appropriate S3 config to make this more secure?
Will tag you again on the thread 🙏

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
sahar-fehri: I think related to previous comment;
We would need to make some updates on the S3 side?
cc @geositta

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,865 @@
...
+
+  if (window.activeStudies[indicatorName]) {
+    return;
+  }
+
+  try {
+    var chart = window.chartWidget.activeChart();
+    var studyName, inputs;
+
+    switch (indicatorName) {
```
geositta, line 241: The chart supports TradingView's full native indicator UI when header_widget or left_toolbar are re-enabled via the disabledFeatures override, but the React-driven programmatic path (indicators prop -> ADD_INDICATOR -> handleAddIndicator) is closed to three hardcoded names via this switch that silently returns on default.

These are two disconnected systems: studies added through TradingView's native UI are invisible to activeIndicatorsRef and won't survive chart recreation, while studies added through the props API are tracked and re-synced but limited to MACD, RSI, and MA200.

Perps needs both paths, the native TradingView indicator menus for user-driven exploration, and app-driven programmatic indicator actions (such as setting default studies on mount, restoring indicator state after chart recreation, or adding studies triggered from app-level controls like a dropdown). Can we add a fallback in the switch's default branch that calls `createStudy(payload.name, false, false, payload.inputs, payload.overrides)` for non-preset study names, and widen ADD_INDICATOR's payload to accept optional inputs and overrides fields?

The three existing presets can stay as convenience cases with their hardcoded parameters. On the type side, IndicatorType at AdvancedChart.types.ts:30 would need to become string (or a branded string with the three presets exported as named constants) so consumers are not artificially constrained at the type level.

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
geositta: Thanks for making that change, but we would ask for one more follow up: the webview now correctly emits NEED_MORE_HISTORY with `oldestTimestamp` and `parseWebViewMessage`, narrows it into a typed `NeedMoreHistoryPayload`, but `handleMessage` calls `onRequestMoreHistory?.()` without forwarding the payload, and the prop at AdvancedChart.types.ts:297 is typed as `() => void`.

This drops the pagination cursor (`oldestTimestamp`) that the webview just computed and forces consumers to track their own "what's my oldest candle" state independently.

Can we change the prop to `onRequestMoreHistory?: (params: { oldestTimestamp: number }) => void` and pass `message.payload` through at line 184? That makes the pagination contract explicit end-to-end: the webview says "I need data older than X", the consumer fetches it, and the deferred `getBars` callback resolves when the data arrives.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
geositta: Can we add a TODO comment near the other price checks in `handleSetPositionLines` noting that `currentPrice` rendering is not yet implemented? That way the gap between type and behavior is discoverable for the next person who sets `currentPrice` and sees nothing on the chart.

geositta (changes requested):
Thank you @sahar-fehri this is a strong improvement overall with datafeed pagination wiring, NEED_MORE_HISTORY emission, listenerGuid subscription handling, and disabledFeatures configurability now in place.

I see one remaining concrete gap and left a comment for it: oldestTimestamp is parsed but not forwarded to consumers.

My second request is based on expected UI requirements: we need React-driven controls that can programmatically add studies (via createStudy path) in addition to native TradingView UI.

I aligned with Product on that requirement; with those two changes, this should likely be ready to merge.

app/components/UI/Charts/AdvancedChart/AdvancedChart.types.ts
geositta: @sahar-fehri Correct, no change is required in this PR. I'm just flagging migration planning so we don’t miss parity later. I agree with your approach: when Perps migration starts, add real ref methods with implemented WebView handlers (not placeholders), e.g. zoomToLatestCandle / resetToDefault.

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
geositta: We should replace the hardcoded S3 URL here in AdvancedChartTemplate.ts with an environment/config driven CDN base URL (the CloudFront distribution URL once DevOps delivers it).

Keep trailing-slash semantics consistent with TradingView's library_path expectations, and make sure the CHARTING_LIBRARY_BASE_URL composition (appending charting_library/) remains correct.

The dev/prod URL strategy should be explicit: local http-server for development, CloudFront for production, and the default committed constant should always be the production value, not a dev only localhost URL.

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
geositta: @sahar-fehri  Thanks for the CSP update. I have a few follow-ups:

  - Add object-src 'none'.
  - Add base-uri 'none'.
  - Add frame-ancestors 'none'.
  - Ensure each source directive is explicit/minimal (use 'self', required data:/blob:, and the
    exact chart asset host; avoid wildcard sources).
  - Keep connect-src 'none' as is for now (no outbound calls needed today).

  Once these are in then the CSP change should be in good shape.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
sahar-fehri: Thnx @geositta ; fixed here [6e37ff8](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/6e37ff8ff5d4474ab5ca5ebe64e1d2867f1d1d76)

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
sahar-fehri: [b7b5472](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/b7b5472a7d13d1315e9278009b2b6924dc072df2)

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
sahar-fehri: updated [7df65d5](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/7df65d5fe978fa7b3f6d7d7017cc2a79a653dd84)

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
geositta: @sahar-fehri Thanks for making the above changes, however I realized the CSP currently setting `frame-src blob: only` is an issue:

With `iframe_loading_same_origin` enabled, TradingView loads `sameorigin.html` from the library host, so `frame-src` must allow that host.

Currently the library host is missing from `frame-src`, which can cause the `sameorigin.html` iframe load to be blocked by CSP.

Please update to include the library origin: `frame-src 'self' blob: <library-origin>;`

Note that <library-origin> means the scheme+host (and optional port) only, such as https://charts-assets.metamask.io and not a path like /charting_library/.

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -0,0 +1,870 @@
...
+  var newResolution = detectResolution(window.ohlcvData);
+  var hasPending = !!window.pendingGetBarsCallback;
+
+  if (hasPending) {
+    var pending = window.pendingGetBarsCallback;
+    window.pendingGetBarsCallback = null;
+    window.currentResolution = newResolution;
+    resolvePendingGetBars(pending);
+    return;
+  }
```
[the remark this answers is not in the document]
geositta: @sahar-fehri Please for this issue would you add this TODO above the if (hasPending) block at line 138?

```
// TODO: Early return bypasses resolution-change handling at lines 146–170.                                  
// If SET_OHLCV_DATA arrives at a different resolution while a history
// request is pending (if a user switches interval during a pending chart candle history navigation request),                                      
// the widget stays at the old resolution while window.currentResolution
// reflects the new one. Fix when wiring up history pagination for Perps.
```

app/components/UI/Charts/AdvancedChart/AdvancedChartTemplate.ts
sahar-fehri: Updated here [c9b3194](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/c9b3194321701777a917b773d315cfefd0407b2a)

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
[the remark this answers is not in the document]
sahar-fehri: [f81f427](https://github.com/MetaMask/metamask-mobile/pull/26459/commits/f81f42735c965f191ce9fac19bf6628a00b8b1f2)

geositta (approved):
Approved - impressive effort on this PR.

Thank you for the iterative fixes.

The main integration risks we flagged have been addressed: history pagination wiring is now end to end, listenerGuid subscription handling is correct, the CSP has been hardened (including same-origin iframe handling), and the indicator path now supports generic createStudy usage beyond the preset set.

I am comfortable with merging this as the shared AdvancedChart foundation, with remaining Perps parity items tracked as follow up work that Perps team can contribute.

Prithpal-Sooriya (approved):
LFG

PR #26526, opened 2026-02-24
feat: generate and publish Platform API docs

cryptodev-2s:
## **Description**

Wires up [`@metamask/platform-api-docs`](https://github.com/MetaMask/core/tree/main/packages/platform-api-docs) so Mobile's Platform API surface (actions and events exposed via the root messenger) is generated from source, previewable locally, and published to GitHub Pages.

**Docs generation**

* Adds `@metamask/platform-api-docs@^0.1.0` as a dev dependency
* Adds `docs:platform-api:build`, which generates docs from `GlobalActions` and `GlobalEvents` in `app/core/Engine/types.ts` and builds a static site under `.platform-api-docs/`
* Adds `docs:platform-api:serve`, same thing plus a local server (passes `--site-base-url /` so local serving works without the production path prefix)
* Ignores `.platform-api-docs/`

**Publishing**

* New reusable workflow `deploy-platform-api-docs.yml` builds the site and pushes it to `gh-pages` under `platform-api-docs/` via `peaceiris/actions-gh-pages`, using the existing token exchange pattern. `keep_files: true` so it does not clobber anything else already published there. (https://metamask.github.io/metamask-mobile/platform-api-docs)
* `ci.yml` calls it on pushes to `main` only. It runs after `check-workflows` and does not gate any other job.

## **Changelog**

CHANGELOG entry: null (dev tooling only)

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/WPC-1203

## **Manual testing steps**

1. `yarn install`
2. `yarn docs:platform-api:build` and confirm `.platform-api-docs/build` is produced
3. `yarn docs:platform-api:serve` and confirm the site lists actions and events from the root messenger

## **Screenshots/Recordings**

N/A — dev tooling only

### **Before**

N/A

### **After**

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -759,6 +759,15 @@ jobs:
         run: ${{ steps.download-actionlint.outputs.executable }} -color -config-file .github/actionlint.yaml
         shell: bash
 
+  deploy-platform-api-docs:
+    name: Deploy Platform API Docs
+    needs: check-workflows
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    permissions:
+      contents: read
+      id-token: write
+    uses: ./.github/workflows/deploy-platform-api-docs.yml
+
   # Warms the Namespace cache volume before test shards run. On non-Namespace
   # runners uploads a ci-js-deps artifact so consumers can skip their own install.
   prepare-ci-js-deps:
--- a/.github/workflows/deploy-platform-api-docs.yml
+++ b/.github/workflows/deploy-platform-api-docs.yml
@@ -1,0 +1,44 @@
+name: Deploy Platform API Docs
+
+on:
+  workflow_call:
+
+jobs:
+  deploy:
+    name: Build and deploy to GitHub Pages
+    runs-on: ubuntu-latest
+    environment: default-branch
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout and setup environment
+        uses: MetaMask/action-checkout-and-setup@v3
+        with:
+          is-high-risk-environment: true
+          persist-credentials: false
+
+      - name: Generate and build Platform API docs
+        env:
+          REPO_OWNER: ${{ github.repository_owner }}
+          REPO_NAME: ${{ github.event.repository.name }}
+        run: |
+          yarn docs:platform-api:build \
+            --site-url "https://${REPO_OWNER}.github.io" \
+            --site-base-url "/${REPO_NAME}/platform-api-docs/"
+
+      - name: Get access token
+        id: get-token
+        uses: MetaMask/github-tools/.github/actions/get-token@v1
+        with:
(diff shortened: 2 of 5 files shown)


.github/workflows/deploy-platform-api-docs.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          yarn docs:platform-api:build \
+            --site-url "https://${REPO_OWNER}.github.io" \
+            --site-base-url "/${REPO_NAME}/platform-api-docs/"
+
+      - name: Get access token
+        id: get-token
+        uses: MetaMask/github-tools/.github/actions/get-token@v1
+        with:
+          token-exchange-url: ${{ vars.TOKEN_EXCHANGE_URL }}
```
mcmire, line 34: It sounds like this is already on your todo list, but you will need to add a token to the Token Exchange Service to get this to work. I would use `job_workflow_ref_match` and not `workflow_ref_match` (so you can target this file instead of `ci.yml`), but up to you.

mcmire (approved):
This looks good, when you have the token set up.

.github/workflows/deploy-platform-api-docs.yml
cryptodev-2s: Yes done here https://github.com/consensys-vertical-apps/token-exchange-service/pull/169

.github/workflows/ci.yml
```diff
@@ -759,6 +759,15 @@ jobs:
         run: ${{ steps.download-actionlint.outputs.executable }} -color -config-file .github/actionlint.yaml
         shell: bash
 
+  deploy-platform-api-docs:
+    name: Deploy Platform API Docs
+    needs: check-workflows
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
```
NicolasMassart, line 765: suggestion (non-blocking) (test): run the Platform API docs build in PR CI

Right now the deployment job only runs on pushes to `main`, so `yarn docs:platform-api:build` isn't exercised by PR CI. This means changes that break docs generation would only be detected after merging.

Could we add a credential-free PR CI step/job that runs `yarn docs:platform-api:build`, while keeping the actual publishing step restricted to `main`? This would validate the main functionality introduced by this PR before merge.

Or if it's on purpose, please provide more info on the reasons of this choice.

cryptodev-2s: All the package does is go through the JSDoc on the root messenger and all its types. If there were any JSDoc or type issues, I believe they would already have been caught by the existing TypeScript (tsc) and lint workflows.

The purpose of republishing the docs after every merge to main is to keep them up to date, so that anyone starting a new feature or piece of work can easily discover all the actions and events currently available through the messenger.

But we could also add another step to run this on PRs. I can handle that as a follow-up.

NicolasMassart: its up to you, i just wanted to make sure it was a known behaviour. not a missed step.

NicolasMassart (approved):

PR #26582, opened 2026-02-25
test: added aproxy to track all the device requests happening

javiergarciavera:
## **Description**

This PR introduces a **Transparent HTTPS MITM Proxy** into the E2E test framework and uses it to implement **Solana swap E2E tests**.

### Why?

Solana swaps require real HTTPS/WSS traffic to external services (Bridge API, Solana RPC, price APIs) that cannot be intercepted by the existing mock server approach (`/proxy?url=…`). The current mock server only captures traffic that the app explicitly routes through it, but Solana-specific calls (particularly WSS `signatureSubscribe` for transaction confirmation) bypass it entirely.

### What changed?

**New: Transparent Proxy (`TransparentProxy.ts`)**
- A `mockttp`-based HTTPS MITM proxy that intercepts **all** device HTTP/HTTPS/WSS traffic at the OS level.
- Generates a self-signed CA certificate, installs it in the simulator/emulator trust store, and configures the device's system proxy.
- Test-specific mock rules take precedence; unmatched requests pass through to real servers.
- Handles `mockttp` internal "Aborted" rejections gracefully to prevent test failures.

**New: Device Proxy Configuration (`DeviceProxyConfig.ts`)**
- Platform-specific helpers to configure/remove system-level proxy settings.
- **iOS**: Sets macOS network proxy via `networksetup`, installs CA cert via `xcrun simctl keychain`.
- **Android**: Sets global HTTP proxy via `adb shell settings`, installs CA cert into system trust store.

**New: Solana Swap E2E Tests (`solana-swap.spec.ts`)**
- 3 smoke tests: SOL→USDC swap, USDC→SOL swap, and no-quotes error state.
- All tests use `useTransparentProxy: true` to intercept real Solana traffic.

**New: Solana Swap Mocks (`solana-swap-mocks.ts`)**
- Comprehensive mock responses for Bridge API (quotes, tokens, tx status, prices).
- Full Solana JSON-RPC mock handler (getBalance, sendTransaction, getTransaction, etc.).
- Local WebSocket server mimicking Solana's `signatureSubscribe` to deliver transaction confirmations.
- Dual-mode rules: works with both the regular mock server (`/proxy` pattern) and the transparent proxy (direct URL matching).

**Framework integration (`FixtureHelper.ts`, `types.ts`, `Constants.ts`, `PortManager.ts`, `FixtureUtils.ts`)**
- New `useTransparentProxy` option in `WithFixturesOptions`.
- Proxy starts **before** `launchApp()` but device proxy configuration is deferred until **after** `launchApp()` + `dismissDevScreens()` so Detox synchronization settles during app startup.
- Detox sync is disabled (`device.disableSynchronization()` + `setURLBlacklist(['.*'])`) only when the transparent proxy is active.
- Full cleanup on teardown: removes device proxy config, stops proxy server, releases port.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: MMQA-1450

## **Manual testing steps**

Feature: Solana Swap E2E Tests with Transparent Proxy

  Scenario: SOL to USDC swap completes successfully
    Given the app is launched with transparent proxy enabled
    And Solana swap API responses are mocked
    When user navigates to Solana token and opens swap
    And enters 1 SOL as source amount
    And confirms the swap
    Then the swap activity shows "SOL to USDC" as confirmed

  Scenario: USDC to SOL swap completes successfully
    Given the app is launched with transparent proxy enabled
    And Solana swap API responses are mocked
    When user navigates to Solana token and opens swap
    And selects USDC as source token and SOL as destination
    And enters 1 USDC as source amount
    And confirms the swap
    Then the swap activity shows "USDC to SOL" as confirmed

  Scenario: No route available when no quotes exist
    Given the app is launched with transparent proxy enabled
    And Solana swap API returns empty quotes
    When user navigates to Solana token and opens swap
    And selects USDC as destination and enters amount
    Then the "no route available" error banner is displayed

## **Changelog**

CHANGELOG entry:

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it adds OS-level proxy/CA certificate installation and Android emulator boot changes that can affect Detox stability and CI device behavior, though it’s gated behind `useTransparentProxy` and skipped on BrowserStack.
> 
> **Overview**
> Adds an optional **transparent HTTPS/WSS MITM proxy** for Detox E2E runs (`useTransparentProxy`) that installs a generated CA cert on Android/iOS, configures system proxy settings, disables Detox synchronization, and passes through unmatched traffic while allowing `testSpecificMock` rules to intercept real external requests.
> 
> Updates the E2E framework to allocate/clean up a new `TRANSPARENT_PROXY` port, tweak Android CI emulator boot args to allow `/system` writes, and wrap Detox Jest `globalSetup` to pre-run `adb root` to avoid mid-test daemon restarts.
> 
> Introduces **Solana swap smoke tests** plus comprehensive Solana swap mocks, including a local `ws`-based WebSocket server to emulate Solana subscription confirmations and dual-mode mocking rules that work with both the existing `/proxy` mock server and the new transparent proxy.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c2231556babc2f369ac8401177311f66532afbc7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


tests/framework/fixtures/FixtureHelper.ts
```diff
@@ -627,15 +652,39 @@ export async function withFixtures(
...
+    // This runs BEFORE proxy config so Detox sync is still active and reliable.
     if (process.env.CI !== 'true') {
       await dismissDevScreens();
     }
 
+    // Configure device proxy AFTER launchApp + dismissDevScreens so Detox sync
+    // can settle during startup. Then disable sync before the test callback.
+    if (useTransparentProxy && transparentProxyInstance && !isBrowserStack) {
+      logger.debug('Configuring device proxy after app launch...');
+      const proxyPort = transparentProxyInstance.getServerPort();
```
christopherferreira9: We should use the PortManager to get the port for this resource:
`PortManager.getInstance().getPort(ResourceType.TRANSPARENT_PROXY);`

tests/framework/fixtures/FixtureHelper.ts
```diff
@@ -576,6 +586,21 @@ export async function withFixtures(
...
     mockServerInstance = mockServerResult.mockServerInstance;
     mockServerPort = mockServerResult.mockServerPort;
+
+    // Step 5: (optional) Start transparent proxy server
+    // Only start the server here; device proxy configuration is deferred until
+    // after launchApp() so Detox synchronization can settle during app startup.
+    // Skip on BrowserStack — the tunnel handles routing there
+    const isBrowserStack =
+      process.env.BROWSERSTACK_LOCAL?.toLowerCase() === 'true';
+    if (useTransparentProxy && !isBrowserStack) {
```
christopherferreira9: We don't need to guard for `isBrowserStack` since `startResourceWithRetry` handles that when starting when setting up port forwarding. Also the tunnel handles routing but the TransparentProxy would still need to exist. We may need to adapt the logic to work with BrowserStack to some extent but we're not looking to use this with BrowserStack just yet.

tests/framework/fixtures/FixtureHelper.ts
```diff
@@ -627,15 +652,39 @@ export async function withFixtures(
...
+        const deviceId = (device as { id?: string }).id || '';
+        await installCACertAndroid(deviceId, certPem);
+        await configureAndroidProxy(deviceId, proxyPort);
+      } else {
+        const simulatorId = (device as { id?: string }).id || 'booted';
+        await configureIOSProxy(simulatorId, proxyPort, certPem);
+      }
+
+      await device.setURLBlacklist(['.*']);
+      await device.disableSynchronization();
```
christopherferreira9: Same as above

tests/framework/fixtures/FixtureHelper.ts
```diff
@@ -627,15 +652,39 @@ export async function withFixtures(
...
+      if (isAndroid) {
+        const deviceId = (device as { id?: string }).id || '';
+        await installCACertAndroid(deviceId, certPem);
+        await configureAndroidProxy(deviceId, proxyPort);
+      } else {
+        const simulatorId = (device as { id?: string }).id || 'booted';
+        await configureIOSProxy(simulatorId, proxyPort, certPem);
+      }
+
+      await device.setURLBlacklist(['.*']);
```
christopherferreira9: We should avoid framework specific mentions in this generic file. Would it be possible to set this on the Transparent server instead behind the FrameworkDetector?

tests/framework/fixtures/DeviceProxyConfig.ts
```diff
@@ -0,0 +1,154 @@
...
+  // Compute the OpenSSL subject_hash_old (the filename Android expects)
+  const { stdout: hashOutput } = await execAsync(
+    `openssl x509 -subject_hash_old -noout -in "${certPath}"`,
+  );
+  const certHash = hashOutput.trim();
+
+  const remotePath = `/system/etc/security/cacerts/${certHash}.0`;
+
+  await execAsync(`adb ${deviceFlag} root`);
+  await execAsync(`adb ${deviceFlag} remount`);
```
christopherferreira9, line 147: we may need to remove `-read-only` from the ci device in `detoxrc.js` for this to work otherwise the command will be revoked.

tests/framework/fixtures/DeviceProxyConfig.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/**
+ * Installs the proxy CA certificate into the Android emulator's system trust store.
+ *
+ * Requires the emulator to be booted with `-writable-system` (and without `-read-only`)
+ * so that `adb root` + `adb remount` can make `/system` writable.
+ * See `.detoxrc.js` — the CI emulator configs are set up for this.
+ */
+export async function installCACertAndroid(
```
FrederikBolding, line 67: Consider whether this should be automatically removed from the simulator after test run

tests/framework/fixtures/DeviceProxyConfig.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  return 'Wi-Fi';
+}
+
+/**
+ * Configures the iOS simulator to route HTTP and HTTPS traffic through the
+ * transparent proxy and installs the CA certificate into the simulator keychain.
+ *
+ * NOTE: The proxy is set on the Mac's network interface so the simulator (which
+ * shares the host network stack) picks it up automatically.
```
FrederikBolding: This seems like a drawback, is it not possible to configure at the simulator level?

tests/framework/fixtures/FixtureHelper.ts
```diff
@@ -706,6 +732,17 @@ export async function withFixtures(
       }
     }
 
+    // Clean up the transparent proxy (remove device proxy config first, then stop server)
+    if (transparentProxyInstance?.isStarted()) {
+      try {
+        await transparentProxyInstance.removeDeviceConfig();
```
FrederikBolding, line 759: We should ensure the device config is removed even if the proxy fails to start

tests/helpers/swap/solana-swap-mocks.ts
```diff
@@ -0,0 +1,852 @@
...
+        responseCode: 200,
+      });
+    }
+
+    // ── Direct URL rules for transparent proxy ──────────────────────────────
+    // When the TransparentProxy server calls this testSpecificMock, traffic arrives
+    // as real HTTPS requests (not via /proxy?url=…). The setupMockRequest calls
+    // above use the /proxy pattern and are no-ops on the transparent proxy, so we
+    // register equivalent rules here that match the actual request URLs.
+    // On the regular mock server these rules are no-ops (it only receives /proxy).
```
FrederikBolding, line 491: I don't fully understand this, why do we need to mock twice? Shouldn't we just be able to use the new proxy exclusively?

tests/framework/fixtures/FixtureHelper.ts
```diff
@@ -627,15 +645,23 @@ export async function withFixtures(
...
 
     // Dismiss dev screens if running locally (not in CI)
+    // This runs BEFORE proxy config so Detox sync is still active and reliable.
     if (process.env.CI !== 'true') {
       await dismissDevScreens();
     }
 
+    // Configure device proxy AFTER launchApp + dismissDevScreens so Detox sync
+    // can settle during startup. Then disable sync before the test callback.
+    if (useTransparentProxy && transparentProxyInstance && !isBrowserStack) {
```
christopherferreira9: We should add an extra guard to check if the transparentProxyInstance has started.

tests/helpers/swap/solana-swap-mocks.ts
christopherferreira9: @FrederikBolding we should phase this in and increase confidence in the new approach before migrating everything. I'm expecting huge volumes of calls to start showing up on this transparent proxy which the catch all mechanism would cause tests to fail. I think this is a good first approach which allows us to transition over time and not cause any breaking changes to the existing mocking mechanism.

tests/framework/fixtures/DeviceProxyConfig.ts
javiergarciavera: This is a real limitation. iOS simulators share the host macOS network stack -- they don't have their own isolated networking like Android emulators do. That's why the current code uses networksetup to set the proxy at the Mac level, which has the side effect of routing all host traffic through the proxy during the test, not just the simulator's.
We could investigate deeper on a follow up PR, but I guess this is not a blocker

PR #26583, opened 2026-02-25
chore: add sentry tracing to ai market insights

joaosantos15:
## **Description**

Adds Sentry performance tracing to Market Insights to measure load times for two key user flows:

MarketInsightsEntryCardLoad — Measures the time from when a token details page starts rendering to when the Market Insights entry card mounts. Uses the asset's CAIP-19 ID as the trace id so rapid token switching doesn't cause trace collisions.

MarketInsightsViewLoad — Measures the time from when the user taps the entry card to when the full Market Insights view renders with report data. Includes an unmount cleanup to end the trace if the user navigates away before the report loads.

Both traces use a new TraceOperation.MarketInsightsLoad (market_insights.load) operation for easy filtering in Sentry.

## **Changelog**

CHANGELOG entry:

---

> [!NOTE]
> **Low Risk**
> Low risk instrumentation-only change, but it touches token details rendering and navigation flows and could create noisy traces if the start/end pairing is missed in edge cases.
> 
> **Overview**
> Adds two new Sentry traces for Market Insights load performance: `MarketInsightsEntryCardLoad` (token details render -> entry card mount) and `MarketInsightsViewLoad` (entry card tap -> report displayed).
> 
> Starts traces from `AssetOverviewContent` using a new `TraceOperation.MarketInsightsLoad`, passes `caip19Id` into `MarketInsightsEntryCard` to end the matching trace on mount, and ends the view-load trace in `MarketInsightsView` once report data is available. Updates the entry card prop types and tests to include the required `caip19Id`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f6ebb8050d2869cf3ad1a87bed80d937a76d5c4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/TokenDetails/components/AssetOverviewContent.tsx
```diff
@@ -309,6 +310,23 @@ const AssetOverviewContent: React.FC<AssetOverviewContentProps> = ({
...
+    marketInsightsCaip19Id &&
+    entryCardTraceStartedRef.current !== marketInsightsCaip19Id
+  ) {
+    entryCardTraceStartedRef.current = marketInsightsCaip19Id;
+    trace({
+      name: TraceName.MarketInsightsEntryCardLoad,
+      op: TraceOperation.MarketInsightsLoad,
+      id: marketInsightsCaip19Id,
+    });
+  }
```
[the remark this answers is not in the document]
joaosantos15: false positive: After tracing the actual render flow — no, the race condition on initial render is not a real bug. Here's why:

The key insight is that MarketInsightsEntryCard never renders on the first render cycle.

Looking at [useMarketInsights.ts:34](vscode-webview://0k9t7vhio2jglqp3nrc78rthcnd2q5udjce0o7a6fic21aodk8ql/app/components/UI/MarketInsights/hooks/useMarketInsights.ts#L34):


const [report, setReport] = useState<MarketInsightsReport | null>(null);  // always null initially
And the render guard in [AssetOverviewContent.tsx:527](vscode-webview://0k9t7vhio2jglqp3nrc78rthcnd2q5udjce0o7a6fic21aodk8ql/app/components/UI/TokenDetails/components/AssetOverviewContent.tsx#L527):


{isMarketInsightsEnabled && marketInsightsReport ? (
  <MarketInsightsEntryCard ... />
) : null}
The actual timeline is:

Render 1: marketInsightsReport = null → MarketInsightsEntryCard is not rendered
Parent's useEffect fires → trace() registers the span
Async fetch completes → marketInsightsReport gets data
Render 2: MarketInsightsEntryCard now renders for the first time
Child's useEffect fires → endTrace() finds the span from step 2
Since the data always loads asynchronously, the parent effect always fires (on render 1) before the child even exists (render 2). There's no same-cycle child-before-parent ordering issue.

app/components/UI/MarketInsights/components/MarketInsightsEntryCard/MarketInsightsEntryCard.types.ts
```diff
@@ -7,6 +7,8 @@ export interface MarketInsightsEntryCardProps {
   timeAgo: string;
   /** Callback when the card is pressed to open the full view */
   onPress: () => void;
+  /** The CAIP-19 asset ID, used to match the trace started by the parent */
+  caip19Id: string;
```
xavier-brochard: I suggest we change this to 
```suggestion
  /** The CAIP-19 asset ID, used to match the trace started by the parent */
  caip19Id?: CaipAssetId;
```
This way
- better typing
- avoid the awkward `caip19Id={marketInsightsCaip19Id ?? ''}` from  `AssetOverviewComponent.tsx` line 533

xavier-brochard: Ideally if we can make it non-optional, that better. But let's avoid passing empty strings in place of caip19s. It's all over the place in the mobile + extension repo, and it's terrible

app/components/UI/TokenDetails/components/AssetOverviewContent.tsx
```diff
@@ -508,6 +530,7 @@ const AssetOverviewContent: React.FC<AssetOverviewContentProps> = ({
                 report={marketInsightsReport}
                 timeAgo={marketInsightsTimeAgo}
                 onPress={handleMarketInsightsPress}
+                caip19Id={marketInsightsCaip19Id ?? ''}
```
xavier-brochard: Let's avoid this `?? ''`.

The fact that `marketInsightsCaip19Id` can be underfined is self-inflicted because we define it only when `isMarketInsightsEnabled=true`.  
When in the view, we always have the asset's caip19 defined, so we should be able to pass it all the time without brittle empty strings

app/components/UI/MarketInsights/components/MarketInsightsEntryCard/MarketInsightsEntryCard.tsx
```diff
@@ -104,10 +105,20 @@ const MarketInsightsEntryCard: React.FC<MarketInsightsEntryCardProps> = ({
...
   const tw = useTailwind();
 
+  useEffect(() => {
+    // End the trace started by the parent (AssetOverviewContent) to measure
+    // how long it takes for the entry card to mount after navigation.
+    endTrace({
+      name: TraceName.MarketInsightsEntryCardLoad,
+      id: caip19Id,
+    });
+  }, [caip19Id]);
```
aganglada, line 120: is this ok? Can we check in here when do we have a marketInsightReport data?

app/components/UI/MarketInsights/Views/MarketInsightsView/MarketInsightsView.tsx
```diff
@@ -485,6 +488,17 @@ const MarketInsightsView: React.FC = () => {
...
+  // End the ViewLoad trace on unmount if the report never loaded,
+  // preventing an orphaned Sentry span when the user navigates away early.
+  useEffect(
+    () => () => {
+      if (!hasTrackedViewRef.current) {
+        endTrace({ name: TraceName.MarketInsightsViewLoad });
+      }
+    },
+    [],
+  );
```
aganglada: can you explain this as well?

aganglada (changes requested):
I can be wrong, but I think we need to base our tracing on how long does it take to have data on the marketInsightReport

app/components/UI/MarketInsights/components/MarketInsightsEntryCard/MarketInsightsEntryCard.tsx
joaosantos15: @aganglada not sure I understand, this component `<MarketInsightsEntryCard/>` is only rendered by the parent component `<AssetOverviewContent/>` and the use effect runs after the component is mounted, which will only happen once the data is loaded. 

> I can be wrong, but I think we need to base our tracing on how long does it take to have data on the marketInsightReport

Oh, so you're suggesting to start the trace inside the hook which does the data fetching? I think with the current flow it makes sense, but maybe I'm thinking about it wrong:

1. `< AssetOverviewContent />` renders, we start the trace at the same time when the `useMarketInsights` hook is called
2. `< MarketInsightsEntryCard />` is rendered only when the report becomes available, after rendering the `useEffect` runs and stops the trace.

app/components/UI/MarketInsights/Views/MarketInsightsView/MarketInsightsView.tsx
joaosantos15: This is just a cleanup function in case the user clicks the report card but goes back before the report has had time to load. If you prefer I can add it to the other useEffect above?

zone-live: Seems unnecessary 🤔 since there won't be a report card without a report, the report has to exist 1st in order to show the market insights card. Also if the user clicks the card the inside info exists, all comes in the same payload.

app/components/UI/MarketInsights/components/MarketInsightsEntryCard/MarketInsightsEntryCard.types.ts
joaosantos15: Great points, I'll fix it

app/components/UI/MarketInsights/Views/MarketInsightsView/MarketInsightsView.tsx
joaosantos15: That's true for what we have today, however if we create deep links to the AI summary page, the page will load without the report in the cache and `useMarketInsights` will make a network request.

joaosantos15: But if you guys insist, I'll take it out, no problem! 😂 I agree that it won't cause problems for now.

zone-live: True, feel free to keep it, doesn't hurt.

joaosantos15: removed

zone-live (dismissed):

app/components/UI/TokenDetails/components/AssetOverviewContent.tsx
joaosantos15: agreed, done!

xavier-brochard (approved):

aganglada (approved):

salimtb (approved):
assets changes lgtm

PR #26596, opened 2026-02-25
feat(ramps): refactors order state into the controller

georgeweiler:
## **Description**

Moves V2 ramp order state management from Redux (`FiatOrder`) into the `RampsController`, eliminating the conversion layer where clean `RampsOrder` API data was mangled into a legacy shape and immediately cast back.

This PR will allow for future PRs such as a headless ramps experience where callers can subscribe to order updates:
https://github.com/MetaMask/metamask-mobile/pull/26728

or implementing websocket connections for real-time order status updates:
https://github.com/MetaMask/core/pull/8075

Please reference loom videos below for a demo of this PR.

**What changed:**

- **Order creation** (Checkout + Transak native): `addOrder(rampsOrder)` stores directly in controller state. No more `createInitialFiatOrder()`, `rampsOrderToFiatOrder()`, or `dispatch(addFiatOrder())`. Transak deposit orders merge `paymentDetails` from the `TransakDepositOrder` onto the `RampsOrder` before storing, preserving inline bank transfer details.

- **Order detail screen**: `OrderDetails` + `OrderContent` read `RampsOrder` from controller state via `useRampsOrders` hook. No Redux, no `processFiatOrder`, no `useInterval` polling. Pull-to-refresh calls `refreshOrder()`. Controller polling keeps data fresh automatically.

- **Bank details screen**: `V2BankDetails` reads order lifecycle from controller state and fetches deposit-specific data (payment details) from `TransakService`. Cancel and confirm use `providerOrderId` directly.

- **Orders list**: `OrdersList` merges legacy `FiatOrder[]` from Redux with V2 `RampsOrder[]` from controller state via `DisplayOrder` projection. Legacy orders route to the aggregator detail screen; V2 orders route to the new controller-native detail screen.

- **Side effects**: `ramps-controller-init.ts` subscribes to `RampsController:orderStatusChanged` for notification and analytics handlers, and starts controller order polling on init. Follows the same pattern as `TransactionController:transactionConfirmed` handlers.

- **Messenger**: `RampsControllerInitMessenger` delegates the `orderStatusChanged` event so init-level subscriptions work.

**What stays untouched:** Legacy `AGGREGATOR`, `DEPOSIT`, and older orders remain in Redux. Their processors, detail screens, types, and polling are unchanged.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Core PR that added orders https://github.com/MetaMask/core/pull/8045

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **After**

Image

Loom video of the controller state being polled, created and updated for Aggregator order (Moonpay)
https://www.loom.com/share/70658e66e55c444cadc02aad407c3da2

Loom video of the controller state being polled, created and updated for Native order (Transak)
https://www.loom.com/share/2c6d5941cbfa48c4be7cc36000e40e89

---

> [!NOTE]
> **Medium Risk**
> Updates unified ramp checkout/order list navigation and switches V2 order creation/storage to controller state, which could impact order visibility and detail routing for buy/deposit flows. Risk is mitigated by updated unit tests, but regressions could affect transaction history UX.
> 
> **Overview**
> **V2 ramp orders are now stored and read from `RampsController` instead of being converted into legacy `FiatOrder` Redux state.** `Checkout` uses `useRampsOrders` (`getOrderFromCallback`, `addOrder`) and always `reset`s into `Routes.RAMP.RAMPS_ORDER_DETAILS` with `showCloseButton`, even when order IDs are `null`.
> 
> **Orders list UI and navigation were reworked to merge legacy Redux orders with controller V2 orders.** `OrdersList` now projects both into a unified `DisplayOrder` list (`mergeDisplayOrders`), renders new list rows, and routes presses by source/provider (legacy aggregator -> `OrderDetails`, V2 -> `RampsOrderDetails`, deposit -> new `DepositOrderDetails`, created deposit -> deposit flow).
> 
> Adds `Routes.DEPOSIT.ORDER_DETAILS` to the main navigator and updates provider selection to consider completed controller orders when computing previously-used providers. Tests/snapshots were updated accordingly, and V2 toast wiring now passes `status` (typed as `RampsOrderStatus`) instead of legacy `state`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7c7f7351086e1cbb65408846a625f82696c001d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Ramp/Views/NativeFlow/BankDetails.tsx
```diff
@@ -42,33 +36,48 @@ import { useTransakController } from '../../hooks/useTransakController';
...
+  const { getOrderById, refreshOrder, removeOrder } = useRampsOrders();
 
   const regionIsoCode = userRegion?.country?.isoCode || '';
 
   const { orderId, shouldUpdate = true } = useParams<BankDetailsParams>();
-  const order = useSelector((state: RootState) => getOrderById(state, orderId));
+  const order = getOrderById(orderId);
 
+  const [depositOrder, setDepositOrder] = useState<TransakDepositOrder | null>(
+    null,
```
AxelGes, line 80: depositOrder starts as null here and only gets populated inside handleOnRefresh, which is gated by shouldUpdate. When navigating with shouldUpdate: false (from useTransakRouting after bank order creation and from OrderContent's Manage Bank Transfer button), this refresh never runs, so depositOrder stays null forever. hasData at line 318 requires both order and depositOrder, so the screen just shows a loader indefinitely. The paymentDetails are already on the controller order at that point since useTransakRouting merges them before calling addOrder, so reading paymentDetails from the controller order when shouldUpdate is false would fix it.

app/components/UI/Ramp/Views/NativeFlow/BankDetails.tsx
```diff
@@ -124,31 +131,34 @@ const V2BankDetails = () => {
...
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   useEffect(() => {
-    if (!order?.state) return;
-    if (order.state === FIAT_ORDER_STATES.CANCELLED) {
+    if (!order?.status) return;
+    if (order.status === RampsOrderStatus.Cancelled) {
+      removeOrder(order.providerOrderId);
```
AxelGes: This permanently deletes the cancelled order from controller state. The old code just navigated away and left the order in Redux with a Cancelled status, so it was still visible in the orders list. With removeOrder here, the user loses any record of the cancelled order entirely. I think we should navigate away but keep the order, and let the list display it with the Cancelled status like it does for other terminal states.

app/components/UI/Ramp/hooks/useTransakRouting.ts
```diff
@@ -269,66 +268,49 @@ export const useTransakRouting = (_config?: UseTransakRoutingConfig) => {
...
+        const rampsOrder = await refreshOrder(
+          providerCode,
+          depositOrder.providerOrderId,
+          walletAddress || depositOrder.walletAddress,
+        );
 
-        await handleNewOrder(processedOrder);
+        addOrder({
+          ...rampsOrder,
+          paymentDetails: depositOrder.paymentDetails,
```
AxelGes, line 290: paymentDetails is a Transak-specific deposit field that is not part of the standard RampsOrder API response, so when the controller's background polling fetches an updated order, the API response won't include it. The controller will overwrite the stored order and paymentDetails will disappear after the first poll cycle. This breaks OrderContent's bank detail rows and the Manage Bank Transfer button for anyone who comes back to check their order after the first polling interval. You probably need to persist paymentDetails separately, or at least re-merge it whenever the order is refreshed.

app/components/UI/Ramp/Aggregator/Views/OrdersList/OrdersList.tsx
```diff
@@ -25,29 +24,158 @@ import {
...
+  );
+
+  const title = item.providerName
+    ? `${item.providerName}: ${strings(
+        isBuy
+          ? 'fiat_on_ramp_aggregator.purchased_currency'
+          : 'fiat_on_ramp_aggregator.sold_currency',
+        { currency: item.cryptoCurrencySymbol },
+      )}`
+    : strings('fiat_on_ramp_aggregator.purchased_currency', {
```
AxelGes: The fallback when providerName is empty always uses purchased_currency regardless of the isBuy flag. A V2 sell order where provider is undefined would end up with an empty providerName and show 'Sold ETH' in the truthy branch but 'Purchased ETH' in this fallback. Should apply the same isBuy ternary here.

app/components/UI/Ramp/utils/displayOrder.ts
```diff
@@ -0,0 +1,79 @@
...
+    account: order.walletAddress,
+  };
+}
+
+export function mergeDisplayOrders(
+  legacyOrders: FiatOrder[],
+  v2Orders: RampsOrder[],
+): DisplayOrder[] {
+  const legacy = legacyOrders
+    .filter((o) => o.provider !== FIAT_ORDER_PROVIDERS.RAMPS_V2)
```
AxelGes: Any V2 orders that were stored in Redux under FIAT_ORDER_PROVIDERS.RAMPS_V2 before this PR lands get silently dropped here. During the transition, a user who already has an in-progress V2 purchase won't see it in the orders list anymore since it no longer appears in either Redux (filtered out) or the controller (not added yet). Might be worth thinking about a migration path or at least making sure the controller polling picks those orders up.

app/components/UI/Ramp/hooks/useTransakRouting.ts
```diff
@@ -270,66 +268,49 @@ export const useTransakRouting = (_config?: UseTransakRoutingConfig) => {
...
+        });
 
         navigateToOrderProcessingCallback({
-          orderId: processedOrder.id,
+          orderId: rampsOrder.providerOrderId,
```
AxelGes, line 301: The old handleNewOrder called showV2OrderToast after adding the order, so the user would see a pending notification right after checkout. That call is gone here. Checkout.tsx still shows the toast in its equivalent flow, but the Transak redirect path through useTransakRouting does not. Same issue applies to the bank transfer addOrder call a bit further down. Probably just needs a showV2OrderToast call after addOrder in both places, same as Checkout.tsx does.

app/components/UI/Ramp/hooks/useTransakRouting.ts
georgeweiler: I've updated the controller to account for this, it should not be a problem anymore

app/components/UI/Ramp/Views/NativeFlow/BankDetails.tsx
georgeweiler: fixed

app/components/UI/Ramp/Views/NativeFlow/BankDetails.tsx
georgeweiler: fixed

app/components/UI/Ramp/Aggregator/Views/OrdersList/OrdersList.tsx
georgeweiler: done

app/components/UI/Ramp/utils/displayOrder.ts
georgeweiler: 👍 yeah these would only be from QA testing, but yeah we should show them

app/components/UI/Ramp/Aggregator/Views/OrdersList/OrdersList.tsx
```diff
@@ -25,29 +24,161 @@ import {
...
+    case 'CREATED':
+    default:
+      statusText = strings('fiat_on_ramp_aggregator.order_status_pending');
+      break;
+  }
+
+  return [statusColor, statusText];
+}
+
+function DisplayOrderListItem({ item }: { item: DisplayOrder }) {
```
georgeweiler, line 95: Image

georgeweiler: 👆 V2 and Legacy orders in harmony

app/components/UI/Ramp/Aggregator/Views/OrdersList/OrdersList.tsx
```diff
@@ -73,44 +204,54 @@ function OrdersList() {
...
-        );
+        handleNavigateToAggregatorTxDetails(orderId);
       }
     },
-    [navigation, orders, goToDeposit],
+    [allLegacyOrders, goToDeposit, handleNavigateToAggregatorTxDetails],
   );
 
-  const renderItem = ({ item }: { item: FiatOrder }) => (
+  const handleItemPress = useCallback(
```
georgeweiler, line 233: V2 orders go to new page. legacy orders go to old page. This can/should be fixed by migrating users from legacy to V2 orders and dropping the concept of legacy orders in the future

Image

georgeweiler:
@cursoragent review this PR

app/components/UI/Ramp/Aggregator/Views/OrdersList/OrdersList.tsx
```diff
@@ -73,44 +205,67 @@ function OrdersList() {
...
+        handleNavigateToAggregatorTxDetails(item.id);
       }
     },
-    [navigation, orders, goToDeposit],
+    [
+      allLegacyOrders,
+      handleNavigateToAggregatorTxDetails,
+      handleNavigateToRampsTxDetails,
+      handleNavigateToDepositTxDetails,
+    ],
```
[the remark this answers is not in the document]
georgeweiler: I will skip this.  it is only for orders created in development/QA withhin the last week, and the impact is very low. Not worth refactoring the clean V2 Order details page to have backwards compatibility as it would defeat the entire purpose of this PR.

AxelGes (approved):

PR #26835, opened 2026-03-02
fix: Updated Input to conditionally apply lineheight 0 when placeholder is shown

brianacnguyen:
## **Description**

This PR fixes input layout by applying **conditional lineHeight** on the component-library **Input** component: when the input is empty (placeholder visible), `lineHeight: 0` is applied to avoid extra vertical space; when the input has a value, lineHeight is omitted so typography renders normally.

**Reason for change:** Empty TextInputs can show inconsistent vertical alignment or extra space due to default line height. Applying `lineHeight: 0` only when the placeholder is visible keeps the cursor/placeholder aligned without affecting filled-in text.

**What changed:**

1. **Input component** (`app/component-library/components/Form/TextField/foundation/Input/`)
   - **Input.tsx**
     - Supports controlled and uncontrolled usage: accepts `value`, `defaultValue`, and `onChangeText` (forwarded from TextInputProps).
     - Derives `isPlaceholderVisible` from current value (`currentValue === '' || currentValue == null`); uses internal state for uncontrolled inputs and syncs with `defaultValue` via `useEffect`.
     - Passes `isPlaceholderVisible` into the style sheet and wires `onChangeText` so internal state stays in sync when uncontrolled.
   - **Input.types.ts**
     - Extends `InputStyleSheetVars` with `isPlaceholderVisible: boolean`.
   - **Input.styles.ts**
     - Reads `isPlaceholderVisible` from vars and applies `...(isPlaceholderVisible && { lineHeight: 0 })` to the base text style so lineHeight is only set when the placeholder is visible.

2. **Input tests** (`Input.test.tsx`)
   - Migrated from Enzyme (shallow) to React Native Testing Library with `ThemeContext` for theme.
   - Removed snapshot test; added behavior-focused tests: render with testID, TextVariant typography, disabled/state-styles-disabled, **lineHeight 0 when value is empty**, **lineHeight omitted when value is non-empty**, and `onChangeText` callback.
   - Deleted `__snapshots__/Input.test.tsx.snap`.

No other app or Storybook code is changed; only the Input foundation and its tests are updated.

## **Changelog**

CHANGELOG entry: Fixed the occasional Input placeholder misalignments

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-499

## **Manual testing steps**

```gherkin
Feature: Input conditional lineHeight

  Scenario: Lint and type check pass
    Given the branch is checked out
    When the author runs yarn lint and yarn lint:tsc
    Then both complete without errors

[...]

PR #26921, opened 2026-03-03
fix(input): simplify iOS placeholder lineHeight fix

georgewrmarshall:
## **Description**

The original PR (#26835) introduced controlled/uncontrolled state tracking (`isControlledRef`, `internalValue`, `useEffect`) and an `onChangeText` interception solely to compute `isPlaceholderVisible` — all to gate a single `lineHeight: 0` style on iOS.

This PR simplifies that approach:

1. **Remove state machinery** — `internalValue`, `isControlledRef`, the `useEffect` for `defaultValue` sync, and `onChangeTextHandler` are all removed. The component returns to its original uncontrolled-friendly shape.
2. **Pass `value` and `placeholder` as style vars** — already available as props, so no extra state is needed.
3. **Compute `isPlaceholderVisible` in the stylesheet** — `!!placeholder && !value`, keeping the logic co-located with where it's used.
4. **Guard with `Platform.OS === 'ios'`** — `lineHeight: 0` is applied only on iOS, preventing the Android text-disappearing regression reported in the Slack thread (text content became invisible while typing because `lineHeight: 0` collapsed the text area on Android).

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: #26835 (addresses Android regression introduced by that approach)

## **Manual testing steps**

```gherkin
Feature: TextField Input placeholder alignment

  Scenario: user views a text field with placeholder on iOS
    Given the app is running on an iOS device
    When a TextField is rendered with a placeholder and no value
    Then the placeholder text is vertically aligned correctly

  Scenario: user types into a text field on iOS
    Given a TextField with a placeholder is visible
    When the user types text into the field
    Then the typed text is visible and not hidden

  Scenario: user types into a text field on Android
    Given a TextField with a placeholder is visible on Android
    When the user types text into the field
    Then the typed text remains visible and does not disappear
```

### **After**

https://github.com/user-attachments/assets/39b301a3-0c5b-475d-83f3-bbdd78ec2c7a

---

> [!NOTE]
> **Medium Risk**
> Changes `Input` controlled/uncontrolled behavior by no longer defaulting `value` to `''` and conditionally omitting the `value` prop, which could affect consumers relying on the previous always-controlled semantics. Styling logic also changes for placeholder visibility, so cross-platform rendering should be spot-checked (especially iOS placeholder and Android typing).
> 
> **Overview**
> Simplifies the iOS placeholder alignment workaround by deriving `isPlaceholderVisible` in `Input.styles.ts` from `placeholder` + `value === ''`, and applying `lineHeight: 0` *only* on iOS.
> 
> Updates `Input.tsx` to stop coercing `value` to `''` and to only pass the `value` prop when it’s defined (preserving uncontrolled inputs), with tests added to ensure omitting `value` doesn’t force controlled mode or apply the placeholder `lineHeight` style.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 53f40d458a60565ee12a20350842fa76b1aae133. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/component-library/components/Form/TextField/foundation/Input/Input.styles.ts
+++ b/app/component-library/components/Form/TextField/foundation/Input/Input.styles.ts
@@ -25,13 +25,14 @@ const styleSheet = (params: { theme: Theme; vars: InputStyleSheetVars }) => {
     isDisabled,
     isStateStylesDisabled,
     isFocused,
-    value = '',
+    value,
     placeholder,
   } = vars;
 
-  const hasPlaceholder = placeholder != null && placeholder !== '';
-  const isPlaceholderVisible =
-    hasPlaceholder && (value === '' || value == null);
+  // Only apply placeholder-mode styling for controlled empty inputs.
+  // For uncontrolled inputs (value is undefined), we avoid forcing lineHeight
+  // since style vars cannot track typing state transitions.
+  const isPlaceholderVisible = !!placeholder && value === '';
 
   const stateObj = isStateStylesDisabled
     ? {
@@ -56,6 +57,11 @@ const styleSheet = (params: { theme: Theme; vars: InputStyleSheetVars }) => {
         fontWeight: theme.typography[textVariant].fontWeight,
         fontSize: theme.typography[textVariant].fontSize,
         letterSpacing: theme.typography[textVariant].letterSpacing,
+        // iOS-only workaround: when a placeholder is visible, the native
+        // TextInput renders it misaligned (vertically offset) due to how iOS
+        // applies lineHeight to placeholder text. Setting lineHeight: 0 resets
+        // this. We guard with Platform.OS === 'ios' because lineHeight: 0 on
+        // Android collapses the text area, making typed content invisible.
         ...(Platform.OS === 'ios' && isPlaceholderVisible && { lineHeight: 0 }),
       },
       style,
--- a/app/component-library/components/Form/TextField/foundation/Input/Input.tsx
+++ b/app/component-library/components/Form/TextField/foundation/Input/Input.tsx
@@ -36,7 +36,6 @@ const Input = React.forwardRef<TextInput, InputProps>(
       autoFocus = true,
       value,
       placeholder,
-      onChangeText,
       ...props
     },
     ref,
@@ -49,7 +48,7 @@ const Input = React.forwardRef<TextInput, InputProps>(
       isStateStylesDisabled,
       isDisabled,
       isFocused,
-      value: value ?? '',
+      value,
       placeholder,
     });
 
@@ -79,8 +78,7 @@ const Input = React.forwardRef<TextInput, InputProps>(
         placeholderTextColor={theme.colors.text.alternative}
         {...props}
         placeholder={placeholder}
-        value={value}
-        onChangeText={onChangeText}
+        {...(value !== undefined ? { value } : {})}
         style={styles.base}
         editable={!isDisabled && !isReadonly}
         autoFocus={autoFocus}
--- a/app/component-library/components/Form/TextField/foundation/Input/Input.types.ts
+++ b/app/component-library/components/Form/TextField/foundation/Input/Input.types.ts
@@ -28,6 +28,12 @@ export interface InputProps extends Omit<TextInputProps, 'editable'> {
    * @default false
    */
   isStateStylesDisabled?: boolean;
+  /**
+   * The value of the text input. Defaults to empty string when not provided,
+   * ensuring the iOS placeholder alignment workaround (lineHeight: 0) can
+   * safely determine whether the placeholder is visible without internal state.
+   */
+  value?: string;
 }
 
 /**
--- a/app/component-library/components/Form/TextField/foundation/Input/Input.test.tsx
+++ b/app/component-library/components/Form/TextField/foundation/Input/Input.test.tsx
@@ -104,4 +104,13 @@ describe('Input', () => {
 
     expect(onChangeText).toHaveBeenCalledWith('a');
   });
+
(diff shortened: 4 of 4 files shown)


app/component-library/components/Form/TextField/foundation/Input/Input.types.ts
```diff
@@ -7,7 +7,7 @@ import { TextVariant } from '../../../../Texts/Text';
 /**
  * Input component props.
  */
-export interface InputProps extends Omit<TextInputProps, 'editable'> {
+export interface InputProps extends Omit<TextInputProps, 'editable' | 'value'> {
```
georgewrmarshall: `value` is explicitly omitted from TextInputProps here so it can be re-declared as a required string below. TextInputProps defines value as string | undefined, but requiring string enforces that this component is always controlled — which is the contract the iOS placeholder fix relies on.

app/component-library/components/Form/TextField/foundation/Input/Input.types.ts
```diff
@@ -28,16 +28,22 @@ export interface InputProps extends Omit<TextInputProps, 'editable'> {
    * @default false
    */
   isStateStylesDisabled?: boolean;
+  /**
+   * The value of the text input. Required — Input is a controlled-only
+   * component. This constraint ensures the iOS placeholder alignment workaround
+   * (lineHeight: 0) can safely determine whether the placeholder is visible
+   * without needing internal state tracking.
+   */
+  value: string;
```
georgewrmarshall: Making `value` required enforces a controlled-only contract at the type level. A codebase audit confirmed all existing call sites already pass `value` this constraint prevents future uncontrolled usage without any migration burden today. We shouldn't ever need an uncontrolled input

app/component-library/components/Form/TextField/foundation/Input/Input.styles.ts
```diff
@@ -25,9 +25,15 @@ const styleSheet = (params: { theme: Theme; vars: InputStyleSheetVars }) => {
...
     isFocused,
-    isPlaceholderVisible,
+    value,
+    placeholder,
   } = vars;
 
+  // `value` is a required prop (Input is controlled-only), so `!value` safely
+  // resolves to true only when value is an empty string — i.e. the placeholder
+  // is actually being shown.
+  const isPlaceholderVisible = !!placeholder && !value;
```
georgewrmarshall: Because value is a required string (never undefined), !value is only true when value is an empty string. This means isPlaceholderVisible reliably reflects whether the placeholder is actually rendered, without needing internal state to track what the user has typed.

app/component-library/components/Form/TextField/foundation/Input/Input.styles.ts
```diff
@@ -51,7 +57,12 @@ const styleSheet = (params: { theme: Theme; vars: InputStyleSheetVars }) => {
         fontWeight: theme.typography[textVariant].fontWeight,
         fontSize: theme.typography[textVariant].fontSize,
         letterSpacing: theme.typography[textVariant].letterSpacing,
-        ...(isPlaceholderVisible && { lineHeight: 0 }),
+        // iOS-only workaround: when a placeholder is visible, the native
+        // TextInput renders it misaligned (vertically offset) due to how iOS
+        // applies lineHeight to placeholder text. Setting lineHeight: 0 resets
+        // this. We guard with Platform.OS === 'ios' because lineHeight: 0 on
+        // Android collapses the text area, making typed content invisible.
+        ...(Platform.OS === 'ios' && isPlaceholderVisible && { lineHeight: 0 }),
```
georgewrmarshall, line 65: The Platform.OS === 'ios' guard is load-bearing. Without it, lineHeight: 0 on Android collapses the text area — typed content becomes invisible even though the value is still present in state. This was an active regression reported on main affecting Android users.

app/component-library/components/Form/TextField/foundation/Input/Input.tsx
```diff
@@ -31,40 +31,21 @@ const Input = React.forwardRef<TextInput, InputProps>(
...
 
     const { styles, theme } = useStyles(styleSheet, {
       style,
       textVariant,
       isStateStylesDisabled,
       isDisabled,
       isFocused,
-      isPlaceholderVisible,
+      value,
+      placeholder,
```
georgewrmarshall, line 52: value and placeholder are passed as style vars rather than a pre-computed isPlaceholderVisible boolean. This keeps the visibility logic inside the stylesheet where it belongs — the component itself has no awareness of the iOS rendering quirk.

app/component-library/components/Form/TextField/foundation/Input/Input.tsx
```diff
@@ -31,40 +31,21 @@ const Input = React.forwardRef<TextInput, InputProps>(
       onFocus,
       autoFocus = true,
       value,
-      defaultValue,
```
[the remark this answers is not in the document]
georgewrmarshall: uncontrolled inputs aren't used in the codebase and likely never will be see https://github.com/MetaMask/metamask-mobile/pull/26921#discussion_r2879893382

app/component-library/components/Form/TextField/foundation/Input/Input.tsx
```diff
@@ -82,8 +78,7 @@ const Input = React.forwardRef<TextInput, InputProps>(
         placeholderTextColor={theme.colors.text.alternative}
         {...props}
         placeholder={placeholder}
-        value={value}
-        onChangeText={onChangeText}
```
georgewrmarshall, line 83: onChangeText is passed via `...props`

PR #26651, opened 2026-02-26
test: enforce color-no-hex

georgewrmarshall:
## **Description**

Removes static hex color values from tests and replaces them with theme-aware color lookups.

Why: hardcoded hex values are brittle.
- In production UI, hex literals can drift from the active theme, become incorrect as the color system evolves, and regress when token/theme mappings change.
- In tests, hex literals in mocks/assertions can fail just because a token’s underlying value changed (noise during token upgrades; see PR #27825).

Goal: no static hex colors anywhere inside app/components/ or app/component-library/ (including their tests and Storybook stories).

Enforcement:
- ESLint: @metamask/design-tokens/color-no-hex is turned on as error only for app/components/** and app/component-library/**.
- Cursor rule: .cursor/rules/unit-testing-guidelines.mdc explicitly calls out “Never hardcode design token hex values” (backed by the eslint rule).

Notable changes in this PR:
- Tests: update theme mocks to reference mockTheme/design tokens instead of literals.
- UI: replace a couple of production hex literals with token-backed values (e.g. SecurityTrustScreen, MarketInsights gradient).

Related: https://github.com/MetaMask/metamask-mobile/pull/26639
Related: https://github.com/MetaMask/metamask-mobile/pull/27825

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: static hex prevention in tests

  Scenario: tests pass after design token upgrade
    Given the design tokens package is upgraded
    When the test suite runs
    Then tests that check colors should not fail due to hex value changes
```

### **After**

Screenshot 2026-03-25 at 10 05 41 PM

## **Pre-merge author checklist**

- [x] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I've included tests if applicable
- [ ] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [ ] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
[...]

PR #12345, opened 2024-11-20
chore: bump walletconnect/* deps

abretonc7s:
## **Description**

Update to latest walletconnect dependencies.

## **Related issues**

Fixes:

## **Manual testing steps**

1. Go to this page...
2.
3.

## **Pre-merge author checklist**

- [x] I’ve followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I’ve included tests if applicable
- [x] I’ve documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I’ve applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

## **Pre-merge reviewer checklist**

- [ ] I've manually tested the PR (e.g. pull and build branch, run the app, test code being changed).
- [ ] I confirm that this PR addresses all acceptance criteria described in the ticket it closes and includes the necessary testing evidence such as recordings and or screenshots.

PR #27149, opened 2026-03-06
test: color-no-hex confirmations batch

georgewrmarshall:
## **Description**

Split from #26651 to reduce CODEOWNERS fanout.

Batch: confirmations
Source branch: \`origin/remove-static-hex-from-tests\`

When \`@metamask/design-tokens\` upgrades change color values, tests that hardcode hex literals (e.g. \`'#ffffff'\`) break because the component renders the new token value while the test still asserts the old one — even though nothing is actually wrong. This PR fixes that brittleness across the confirmations scope.

**Two strategies are applied:**

1. **Replace hardcoded hex with token references** — test mocks and assertions use \`mockTheme.colors.*\` (or \`brandColor.*\` for production code) so the test always agrees with whatever the current token value is. Package upgrades no longer cause spurious failures.

2. **Add targeted \`eslint-disable\` comments** — strings like \`'#12345'\` or \`'#456'\` are NFT token IDs that happen to look like hex colors. The \`color-no-hex\` rule can't distinguish them from CSS colors, so suppress comments are added inline to keep the rule enforced everywhere else without false positives.

Together these changes make the \`@metamask/design-tokens/color-no-hex\` lint rule enforceable across this scope, so future hex literals won't accidentally creep in.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

\`\`\`gherkin
Feature: color-no-hex lint compliance (confirmations batch)

  Scenario: user runs lint
    Given the confirmations test files have been updated
    When user runs the lint checks
    Then no color-no-hex violations are reported
\`\`\`

### DIFF
--- a/app/components/Views/confirmations/components/UI/bottom-modal/bottom-modal.tsx
+++ b/app/components/Views/confirmations/components/UI/bottom-modal/bottom-modal.tsx
@@ -6,6 +6,7 @@ import { useTheme } from '../../../../../../util/theme';
 import { useStyles } from '../../../../../hooks/useStyles';
 import styleSheet from './bottom-modal.styles';
 
+// eslint-disable-next-line @metamask/design-tokens/color-no-hex
 const OPAQUE_GRAY = '#414141';
 interface BottomModalProps {
   avoidKeyboard?: boolean;
--- a/app/components/Views/confirmations/components/UI/bottom-modal/bottom-modal.tsx
+++ b/app/components/Views/confirmations/components/UI/bottom-modal/bottom-modal.tsx
@@ -2,12 +2,13 @@ import React, { ReactChild } from 'react';
 import Modal from 'react-native-modal';
 import { View } from 'react-native';
 
+import { brandColor } from '@metamask/design-tokens';
+
 import { useTheme } from '../../../../../../util/theme';
 import { useStyles } from '../../../../../hooks/useStyles';
 import styleSheet from './bottom-modal.styles';
 
-// eslint-disable-next-line @metamask/design-tokens/color-no-hex
-const OPAQUE_GRAY = '#414141';
+const OPAQUE_GRAY = brandColor.grey600;
 interface BottomModalProps {
   avoidKeyboard?: boolean;
   children: ReactChild;
--- a/app/components/Views/confirmations/components/UI/info-row/divider/divider.test.tsx
+++ b/app/components/Views/confirmations/components/UI/info-row/divider/divider.test.tsx
@@ -1,14 +1,14 @@
 import React from 'react';
 import { InfoRowDivider } from './divider';
 import renderWithProvider from '../../../../../../../util/test/renderWithProvider';
-import { lightTheme } from '@metamask/design-tokens';
+import { mockTheme } from '../../../../../../../util/theme';
 describe('InfoRowDivider', () => {
   it('should use correct styles from useStyles hook', () => {
     const wrapper = renderWithProvider(<InfoRowDivider />);
 
     expect(wrapper.root.props.style).toEqual({
       height: 1,
-      backgroundColor: lightTheme.colors.border.muted,
+      backgroundColor: mockTheme.colors.border.muted,
       marginHorizontal: -8,
     });
   });
--- a/app/components/Views/confirmations/components/UI/nft/nft.test.tsx
+++ b/app/components/Views/confirmations/components/UI/nft/nft.test.tsx
@@ -29,6 +29,7 @@ describe('Nft', () => {
     const mockNft = createMockNft({
       standard: 'ERC721',
       collectionName: 'Bored Apes',
+      // eslint-disable-next-line @metamask/design-tokens/color-no-hex
       name: 'Ape #456',
       tokenId: '456',
       balance: '0',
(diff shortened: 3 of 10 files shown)


app/components/Views/confirmations/components/UI/bottom-modal/bottom-modal.tsx
```diff
@@ -2,11 +2,13 @@ import React, { ReactChild } from 'react';
...
 import { View } from 'react-native';
 
+import { brandColor } from '@metamask/design-tokens';
+
 import { useTheme } from '../../../../../../util/theme';
 import { useStyles } from '../../../../../hooks/useStyles';
 import styleSheet from './bottom-modal.styles';
 
-const OPAQUE_GRAY = '#414141';
+const OPAQUE_GRAY = brandColor.grey600;
```
georgewrmarshall, line 11: We should avoid using colors outside of our design tokens. `brandColor.grey600` (`#4b505c`) is the closest token to `#414141`, so we should use that instead of introducing a new hex value.

We also do not currently have a fully opaque overlay token. I am curious what `hideBackground` is intended for. From a quick look, `hideBackground` does not appear to be used. Should it be removed? We should likely be using `BottomSheet` instead of `BottomModal` anway?

@MetaMask/confirmations

```
backdropColor={hideBackground ? OPAQUE_GRAY : colors.overlay.default}
```

app/components/Views/confirmations/hooks/ui/useEmptyNavHeaderForConfirmations.test.tsx
```diff
@@ -4,27 +4,31 @@ import { Theme } from '../../../../../util/theme/models';
...
+  const { mockTheme } = jest.requireActual('../../../../../util/theme');
 
   const mockNavbarOptions = {
     headerTitle: () => <></>,
     headerLeft: () => <></>,
     headerRight: () => <></>,
     headerTitleAlign: 'center' as const,
     headerStyle: {
-      backgroundColor: '#ffffff',
+      backgroundColor: mockTheme.colors.background.default,
```
georgewrmarshall, line 31: When `'#ffffff'` is hardcoded in a test assertion, any `@metamask/design-tokens` upgrade that changes `background.default` will cause this test to fail even though the component is rendering the correct token value. By referencing `mockTheme.colors.background.default`, the assertion always matches whatever the current token resolves to, making this test resilient to package upgrades.

app/components/Views/confirmations/components/modals/edit-spending-cap-modal/edit-spending-cap-modal.test.tsx
```diff
@@ -10,23 +10,20 @@ jest.mock('../../../../../../../locales/i18n', () => ({
...
+        container: {},
+        title: {},
+        description: {},
+        balanceInfo: {},
+        buttonsContainer: {},
+        button: {},
       },
-    },
-  })),
+      theme: mockTheme,
```
georgewrmarshall, line 24: The previous mock returned `theme: { colors: { text: { alternative: '#6a737d' } } }` — a partial, hardcoded copy of a design token value. If `@metamask/design-tokens` upgrades that color, the component renders the new value but the test mock still returns the old hex, producing a spurious failure. Replacing it with `mockTheme` (the full canonical test theme) means the mock always stays in sync with the token package automatically.

app/components/Views/confirmations/components/UI/info-row/divider/divider.test.tsx
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
 import { InfoRowDivider } from './divider';
 import renderWithProvider from '../../../../../../../util/test/renderWithProvider';
-import { lightTheme } from '@metamask/design-tokens';
+import { mockTheme } from '../../../../../../../util/theme';
```
georgewrmarshall, line 4: Using `lightTheme` directly here achieves the same thing, we will not see a test failure if the design tokens are updated. `mockTheme` uses `lightTheme` and is the convention we are using to mock colors.

app/components/Views/confirmations/components/UI/nft/nft.test.tsx
```diff
@@ -29,6 +29,7 @@ describe('Nft', () => {
     const mockNft = createMockNft({
       standard: 'ERC721',
       collectionName: 'Bored Apes',
+      // eslint-disable-next-line @metamask/design-tokens/color-no-hex
```
georgewrmarshall, line 32: Strings like `'#456'`, `'#12345'`, `'#22222'` are NFT token IDs — `#` is a display prefix (e.g. `#<tokenId>`), not a CSS hex color. The `color-no-hex` rule can't distinguish between them, so these targeted suppressions are intentional. The rule is being enforced across the codebase to prevent real static hex values from creeping into tests (which would make them brittle on package upgrades), so the inline disables are the right way to allow these legitimate domain strings without weakening the broader rule.

app/components/Views/confirmations/components/UI/bottom-modal/bottom-modal.tsx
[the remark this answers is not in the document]
georgewrmarshall: It exists... chill

Screenshot 2026-03-10 at 10 09 47 AM
Screenshot 2026-03-10 at 10 09 51 AM

OGPoyraz (approved):

PR #27164, opened 2026-03-06
fix(confirmations): use address from tx and handle empty claimable positions

matallui:
## **Description**

`PredictClaimFooter` was internally reading the selected account address via `selectSelectedInternalAccountAddress` selector, which could be stale or mismatched with the actual transaction address. This change:

1. Accepts `address` as a prop so the parent passes the address directly from the transaction context, ensuring consistency.
2. Adds an `onError` callback prop that fires when the component encounters no won positions to claim, allowing the parent to handle the error appropriately instead of silently rendering nothing.
3. Removes the internal `selectSelectedInternalAccountAddress` selector dependency and the `?? '0x0'` fallback.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://github.com/MetaMask/MetaMask-planning/issues/7093

## **Manual testing steps**

```gherkin
Feature: Predict claim confirmation

  Scenario: user claims winning prediction positions
    Given user has winning prediction positions for their account

    When user opens the claim confirmation
    Then the claim footer displays the correct positions and claim button

  Scenario: user attempts to claim with no winning positions
    Given user has no winning prediction positions for the given address

    When user opens the claim confirmation
    Then the onError callback is invoked with an error message
```

### **Before**

N/A - logic-only change

### **After**

N/A - logic-only change

---

> [!NOTE]
> **Medium Risk**
> Touches confirmation flow for `predictClaim` by changing how the claim footer derives the address and how errors trigger rejection, which could affect whether a user can confirm or gets auto-rejected in edge cases.
> 
> **Overview**
> Updates the `predictClaim` confirmation footer to derive the wallet address from the current transaction (`txParams.from`) instead of the selected account, avoiding mismatches when the selected account is stale.
> 
> Adds an `onError` callback to `PredictClaimFooter` and triggers it (while rendering `null`) when there is no transaction address or no won/claimable positions; the parent `Footer` wires this to `onReject` so empty claims are handled explicitly. Tests are updated to cover the new `onError` behavior and the no-positions path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 76bc72f54b704b006edfd4683c625333e753713c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/Views/confirmations/components/footer/footer.tsx
```diff
@@ -194,7 +194,13 @@ export const Footer = () => {
     transactionMetadata &&
     hasTransactionType(transactionMetadata, [TransactionType.predictClaim])
   ) {
-    return <PredictClaimFooter onPress={onConfirm} />;
+    return (
+      <PredictClaimFooter
+        address={transactionMetadata.txParams.from}
```
matthewwalsh0: Minor, we could read this inside the component since it's confirmation specific.

app/components/Views/confirmations/components/predict-confirmations/predict-claim-footer/predict-claim-footer.tsx
```diff
@@ -24,22 +23,33 @@ import ButtonHero from '../../../../../../component-library/components-temp/Butt
...
+      address,
     }),
   );
 
-  if (!wonPositions?.length) {
+  const hasNoPositions = !wonPositions?.length;
+
+  useEffect(() => {
+    if (hasNoPositions) {
+      onError(new Error('Tried to claim but no positions were won'));
```
matthewwalsh0, line 50: This is a logical fallback, but ideally we'd check the state before we even add the transaction, meaning this shouldn't be possible?

matthewwalsh0 (dismissed):

app/components/Views/confirmations/components/predict-confirmations/predict-claim-footer/predict-claim-footer.tsx
matallui: We do check the state before we send the TX, that is why I can't figure out why this bug is happening. Either way, we should never let the user get to this state, so IMO it's still a good fallback to error the tx so we can catch that and show an error to the user.

app/components/Views/confirmations/components/footer/footer.tsx
matallui: Updated the PR to grab this info inside the component rather than passing `address` as a prop

matthewwalsh0 (approved):

PR #27324, opened 2026-03-11
fix(perps): prevent action button flash when navigating to market with open position [codex]

abretonc7s:
WIP — TAT-2236 (Codex session).


abretonc7s:
Pausing Codex experiment — need to configure approval policy first before running autonomously.

PR #26258, opened 2026-02-19
fix(oauth): add RefreshTokenHttpError and guard refreshAuthTokens in MaxKeyChainLength recovery

himanshuchawla009:
## Explanation

Companion to MetaMask/core#7989.

### 1. `RefreshTokenHttpError` in `AuthTokenHandler`

The core `SeedlessOnboardingController` now distinguishes permanent token failures (401) from transient ones to apply the right recovery strategy. It does this by checking whether the error thrown by `refreshJWTToken` is a `RefreshTokenHttpError` with `statusCode === 401`.

Previously `AuthTokenHandler` threw a plain `Error('Failed to refresh JWT token')` for all non-2xx responses, making this distinction impossible. This PR adds a `RefreshTokenHttpError` class that carries the HTTP `statusCode`, giving the controller the information it needs

### 2. try/catch around `refreshAuthTokens` in the `MaxKeyChainLengthExceeded` recovery path

`refreshAuthTokens` is called fire-and-forget in the `MaxKeyChainLengthExceeded` recovery path before `rehydrateSeedPhrase`. If the refresh token is already stale or revoked (which is precisely the scenario that triggers this path), the call would throw and abort rehydration entirely — leaving the user stuck.

Wrapping in try/catch logs the error and lets rehydration proceed regardless.

### 3. encryptorAdapter for SeedlessOnboardingController
KeyringController has a generic EncryptionResult type that accepts our mobile Encryptor's cipher property. However, SeedlessOnboardingController uses a fixed VaultEncryptor type (from ExportableKeyEncryptor) that strictly requires data instead of cipher.

## **Changelog**

CHANGELOG entry: null

## References

- Companion PR: MetaMask/core#7989 — fixes JWT token refresh race conditions in `SeedlessOnboardingController`

## Checklist

- [ ] I've updated the test suite for new or updated code as appropriate
- [ ] I've updated documentation (JSDoc, Markdown, etc.) for new or updated code as appropriate
- [ ] I've manually tested the PR (required)

---

> [!NOTE]
> **Medium Risk**
> Touches OAuth token refresh and seedless recovery flows; incorrect error typing/handling could change recovery behavior for revoked or transient refresh-token failures.
> 
> **Overview**
> Improves seedless OAuth token-refresh error handling so callers can distinguish auth-server HTTP failures by status code.
> 
[...]

PR #27382, opened 2026-03-11
fix(seedless-onboarding): handle both vault formats in encryptorAdapter cp-7.69.0

himanshuchawla009:
## Explanation

Fixes a critical regression introduced by PR #26258 (`encryptorAdapter` for SeedlessOnboardingController) that permanently locks affected users out of their wallets.

### Root cause

The `encryptorAdapter` overrides `encryptWithKey` to write vaults using a `data` field (browser-passworder standard), and overrides `decryptWithKey` to read from `data`. However, `decrypt` and `decryptWithDetail` were left as spreads from `...encryptor` — the mobile Encryptor — which internally reads `cipher`.

The only passwordless `#updateVault` call (in `#doRefreshAuthTokens`, line 2193 in the controller) fires automatically in the background whenever the wallet is unlocked and a JWT token refresh succeeds. After this runs:

1. Vault is written to AsyncStorage in **`data` format** (adapter's `encryptWithKey`)
2. Wallet locks (auto-lock or restart)
3. `vaultEncryptionKey` is cleared (not persisted)
4. User tries to unlock → `decryptWithDetail` (mobile Encryptor) reads `JSON.parse(vault).cipher` → **`undefined`**
5. `Buffer.from(undefined, 'base64')` → **crash** → user is permanently locked out

Every subsequent unlock attempt fails the same way — there is no recovery path without a fix.

### Fix

Three changes to `encryptorAdapter`:

1. **`decrypt`** — overridden to call `normalizeVaultFormat` before delegating to `encryptor.decrypt`, mapping `data` → `cipher` for vaults written by the adapter.
2. **`decryptWithDetail`** — same normalization before delegating to `encryptor.decryptWithDetail`.
3. **`decryptWithKey`** — falls back to `encryptedObject.cipher` when `encryptedObject.data` is absent, so pre-adapter vaults (stored with `cipher`) remain readable.

The `normalizeVaultFormat` helper adds a `cipher` field equal to `data` when a vault has `data` but not `cipher`. Vaults already in `cipher` format pass through unchanged.

### Impact

- Only **seedless onboarding users** are affected
- Triggered automatically after any background JWT token refresh while the wallet is unlocked

## References

- Sentry issue: METAMASK-MOBILE-5K9N
- Regression introduced by: https://github.com/MetaMask/metamask-mobile/commit/0134ffd0ce4c29706e7b88ba1d848535b8c5884f

## Changelog

CHANGELOG entry: Fixed a bug that permanently locked some seedless onboarding users out of their wallets after a background token refresh

## Checklist

- [x] I've updated the test suite for new or updated code as appropriate
- [ ] I've manually tested the PR (required)

---

> [!NOTE]
> **High Risk**
> Touches seedless onboarding vault encryption/decryption paths; mistakes here can break unlock/decrypt behavior and potentially lock users out. Although scoped and well-tested, it modifies security- and data-critical logic.
> 
> **Overview**
> Fixes `seedlessOnboardingController`’s `encryptorAdapter` to correctly handle both vault serialization formats (`data` vs `cipher`). It adds `normalizeVaultFormat()` and overrides `decrypt`/`decryptWithDetail` to map `data`→`cipher` before delegating to the mobile `Encryptor`, and updates `decryptWithKey` to accept legacy `cipher` objects (and throw a clear error if neither field exists).
> 
> Expands `index.test.ts` to mock the new decrypt methods and cover normalization, legacy fallback behavior, and the new error case.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7e270d211f6564f85b5fedebcdaf60d2b9e97f59. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/core/Engine/controllers/seedless-onboarding-controller/index.test.ts
+++ b/app/core/Engine/controllers/seedless-onboarding-controller/index.test.ts
@@ -68,11 +68,15 @@ const getAuthTokenHandlerMocks = () => {
 jest.mock('../../../Encryptor', () => {
   const mockEncryptWithKeyFn = jest.fn();
   const mockDecryptWithKeyFn = jest.fn();
+  const mockDecryptFn = jest.fn();
+  const mockDecryptWithDetailFn = jest.fn();
 
   return {
     Encryptor: jest.fn().mockImplementation(() => ({
       encryptWithKey: mockEncryptWithKeyFn,
       decryptWithKey: mockDecryptWithKeyFn,
+      decrypt: mockDecryptFn,
+      decryptWithDetail: mockDecryptWithDetailFn,
     })),
     LEGACY_DERIVATION_OPTIONS: {
       algorithm: 'PBKDF2',
@@ -80,6 +84,8 @@ jest.mock('../../../Encryptor', () => {
     },
     __mockEncryptWithKey: mockEncryptWithKeyFn,
     __mockDecryptWithKey: mockDecryptWithKeyFn,
+    __mockDecrypt: mockDecryptFn,
+    __mockDecryptWithDetail: mockDecryptWithDetailFn,
   };
 });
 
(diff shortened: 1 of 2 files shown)


ieow (approved):
lgtm

PR #27409, opened 2026-03-12
fix(perps): pass position in TP/SL onConfirm to avoid No position found errors

michalconsensys:
## **Description**

When editing Take Profit / Stop Loss (TP/SL) for an existing Perps position, the `onConfirm` callback relied on a ref (`currentPositionRef.current`) to get the position at execution time. If the position was updated during navigation (e.g. via WebSocket), the ref could be stale, causing "No position found" errors when the user confirmed TP/SL.

This change extends the TP/SL `onConfirm` signature to accept an optional first argument `position`. The TPSL view now passes the position from route params into `onConfirm`, and the market-details flow uses that passed position when present, falling back to the ref. Order flow and market-tabs callbacks were updated to match the new signature (they ignore the position argument). This ensures the callback always receives the correct position and avoids the stale-ref error.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TAT-2369

## **Manual testing steps**

```gherkin
Feature: Perps TP/SL confirmation

  Scenario: user confirms TP/SL for an existing position without "No position found" error
    Given user has an open Perps position and navigates to edit TP/SL (market details or market tabs)
    And the position data may have been updated (e.g. via WebSocket) before user taps Confirm

    When user sets Take Profit and/or Stop Loss and taps Confirm

    Then TP/SL is updated successfully and no "No position found" error is shown
```

## **Screenshots/Recordings**

N/A (behavioral fix; no UI change).

---

> [!NOTE]
> **Medium Risk**
> Changes the `PerpsTPSL` route `onConfirm` callback signature/return type and updates multiple callers, so mismatches could break TP/SL confirmation flows if any usage is missed.
> 
> **Overview**
> Fixes a TP/SL confirmation race by **passing the `position` from `PerpsTPSLView` into the route `onConfirm` callback**, and having `PerpsMarketDetailsView` prefer that position (falling back to `currentPositionRef`) before calling `handleUpdateTPSL`.
> 
> Updates the `PerpsTPSL` navigation type and all known callers (`PerpsOrderView`, `PerpsMarketTabs`) to match the new `(position?, tp?, sl?, trackingData?)` signature, adjusts TPSL view tests accordingly, and adds a market-details test asserting `{ success: false }` when confirm runs after the position is cleared.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ee47f9c91b254caa11ef1ee8a482f7c108c38380. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


gambinish (dismissed):
Approving, but if you have time, could we increase test coverage here?

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.test.tsx
```diff
@@ -1927,6 +1927,69 @@ describe('PerpsMarketDetailsView', () => {
...
+        error: null,
+        existingPosition: null,
+        refreshPosition: jest.fn(),
+        positionOpenedTimestamp: undefined,
+      });
+
+      rerender(viewTree);
+
+      await act(async () => {
+        const result = await onConfirm(undefined, undefined, undefined);
```
abretonc7s, line 1989: This test calls `onConfirm(undefined, undefined, undefined)` — simulating both position sources being null — and asserts `{ success: false }`. That failure path also existed before the fix (the ref would have been null in the same scenario). The primary fix scenario is untested: when `positionFromRoute` is the captured position object and `currentPositionRef.current` is cleared, `handleUpdateTPSL` should be called with `positionFromRoute`. Consider adding a second `act` block that calls `onConfirm(existingPosition, undefined, undefined, undefined)` after clearing the ref and asserts that the update succeeds (mock `handleUpdateTPSL` and verify it was called with the route-captured position).

app/components/UI/Perps/Views/PerpsTPSLView/PerpsTPSLView.tsx
```diff
@@ -383,7 +383,13 @@ const PerpsTPSLView: React.FC = () => {
...
       };
-      await onConfirm(parseTakeProfitPrice, parseStopLossPrice, trackingData);
+      // Pass position from route params so the callback always has the correct position (avoids "No position found" when parent ref is stale)
+      await onConfirm(
+        position,
+        parseTakeProfitPrice,
+        parseStopLossPrice,
+        trackingData,
+      );
       navigation.goBack();
```
abretonc7s, line 393: `navigation.goBack()` is called unconditionally after `await onConfirm(...)`. If `onConfirm` returns `{ success: false }` (both position sources null), the user is navigated back silently with no error feedback — `handleUpdateTPSL` is never called, so no error toast fires either. Pre-existing behavior, but now more visible with the explicit `{ success: boolean }` return type. Consider checking the result: `const result = await onConfirm(...); if (!result || result.success !== false) { navigation.goBack(); }`

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -761,9 +761,12 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
...
       amount: orderForm.amount,
       szDecimals: marketData?.szDecimals,
-      onConfirm: async (takeProfitPrice?: string, stopLossPrice?: string) => {
-        // Use the same clearing approach as the "Off" button
-        // If values are undefined or empty, ensure they're cleared properly
+      onConfirm: async (
+        _position?: Position,
+        takeProfitPrice?: string,
+        stopLossPrice?: string,
+      ) => {
```
abretonc7s, line 768: Minor inconsistency: this callback lists 3 params (`_position`, `takeProfitPrice`, `stopLossPrice`) and omits `_trackingData`, while `PerpsMarketTabs` explicitly lists all 4. TypeScript allows fewer params, but for readability and future-proofing consider adding `_trackingData?: TPSLTrackingData` as the 4th ignored param to match the full type signature.

abretonc7s (dismissed):
**Review: PR #27409 — fix(perps): pass position in TP/SL onConfirm**

Tier: standard | Tests: all pass | TSC: clean | Device: validated on iOS sim

Sound, minimal fix for the stale-ref race. All three PerpsTPSL callers updated correctly. 2 suggestions (test the primary fix path, check onConfirm result before goBack), 1 nitpick.

app/components/UI/Perps/Views/PerpsTPSLView/PerpsTPSLView.tsx
michalconsensys: This was there before

abretonc7s (approved):

PR #27419, opened 2026-03-12
chore: adds market insights to perps

zone-live:
## **Description**

Integrates the Market Insights feature into the Perps Market Details view behind a new remote feature flag, displaying the `MarketInsightsEntryCard`. Tapping the card opens the existing MarketInsightsView, extended with an isPerps route param that switches the Trade button to navigate to the Perps order view instead of Swaps and checks the new perps-specific feature flag. The useMarketInsights hook's parameter is renamed from `caip19Id` to `assetIdentifier` to match the Core changes, allowing perps market symbols (ETH or BTC) to be passed directly without a new hook.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> **Medium Risk**
> Adds a new perps-specific remote flag and integrates Market Insights into Perps market details plus shared navigation/analytics; risk is mainly around conditional routing and event property changes for an existing screen.
> 
> **Overview**
> **Adds Market Insights to Perps market details behind a new remote flag.** `PerpsMarketDetailsView` now fetches insights by perps market symbol and shows a `MarketInsightsEntryCard`; tapping it traces load and opens the existing `MarketInsightsView` in a perps context.
> 
> **Extends `MarketInsightsView` to support perps context.** Route params rename `caip19Id` to `assetIdentifier` (token CAIP-19 *or* perps symbol) and add `isPerps`, switching CTAs from Swap/Buy to Long/Short that deep-link to `Routes.PERPS.ORDER_REDIRECT`.
> 
> **Updates gating + telemetry + tests.** The Market Insights screen is registered when either token or perps insights flags are enabled, analytics uses `caip19` vs `perps_market` depending on context, and new selectors/tests are added for `selectMarketInsightsPerpsEnabled` plus updated hook/view tests (including new button test IDs).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 943bd0ca410bf27de162cdf3bab579aeb30da3b0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


sahar-fehri (dismissed):

Prithpal-Sooriya (dismissed):

app/components/UI/MarketInsights/Views/MarketInsightsView/MarketInsightsView.tsx
```diff
@@ -219,14 +240,36 @@ const MarketInsightsView: React.FC = () => {
...
+      )
+        .addProperties({
+          ...assetIdProperty,
+          interaction_type: direction,
+        })
+        .build();
+      trackEvent(event);
+
+      navigation.navigate(
+        Routes.PERPS.ROOT as never,
```
geositta: Nit: the `as never` casts remove route/param validation on the new Long/Short navigation path. The repo already uses typed Perps navigation with `NavigationProp<PerpsNavigationParamList>` (for example in PerpsSection and TrendingView).

Using that same approach here and on line 268 in this file would keep `PerpsOrderRedirect` params compile time checked as route contracts change in future.

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
```diff
@@ -1005,6 +1015,20 @@ const PerpsMarketDetailsView: React.FC<PerpsMarketDetailsViewProps> = () => {
...
   }, [handleBannerDismissComplete]);
 
+  // Handler for market insights card tap - navigates to full market insights view
+  const handleMarketInsightsPress = useCallback(() => {
+    if (!market?.symbol) return;
+    trace({
+      name: TraceName.MarketInsightsViewLoad,
+      op: TraceOperation.MarketInsightsLoad,
+    });
+    navigation.navigate(Routes.MARKET_INSIGHTS.VIEW, {
```
geositta, line 1025: Potential rollout bug: If Perps insights is turned on before token insights, users can see this card and tap it, but we send them to a screen that isn’t actually registered in the navigator. In that flag state, the tap path is broken. Can we either require both flags before rendering the card, or add a guard before calling navigate?

geositta (commented):
The feature implementation and tests all are solid. There is one blocking issue IMO: the new Perps insights card can navigate to an unmounted screen under what would be a valid flag combination. I would prefer we gate merge on that fix. The `as never` item is nonblocking but I think worth cleaning up for this new navigation path.

app/components/UI/Perps/Views/PerpsMarketDetailsView/PerpsMarketDetailsView.tsx
zone-live: We will enable the perps at a later stage, likely, otherwise it will be at the same time, but I get your POV and is totally valid. Its better to be safe than sorry, I'll update it 💪🏼

sahar-fehri (approved):

geositta (approved):
Thanks for addressing both review items. One possible follow up I noticed is that `useNavigation` in `MarketInsightsView` is still untyped, so `PerpsOrderRedirect` params are not compile time validated yet. If you want to close that gap, we can align with the `PerpsSection` pattern and use `useNavigation<NavigationProp<PerpsNavigationParamList>>()`.

However, I am not blocking on this PR for that change.

More broadly, I see that route typing is inconsistent across the app today, so this is a good candidate for a wider cleanup pass.

PR #27448, opened 2026-03-12
feat(ramp): Converts payment methods and quotes to use react-query for requ…

imyugioh:
## **Description**

Fixes the "Token Not Available" modal and payment methods loading behavior in the Buy flow. Three issues are addressed:

### 1. React-query migration for request status tracking

Previously, the controller only exposed `isLoading` (boolean) and `data` (array). This made it impossible to distinguish between:
- **Never fetched yet** (idle) — `isLoading: false`, `data: []`
- **Fetched successfully with no results** (token genuinely unavailable) — `isLoading: false`, `data: []`

This ambiguity caused the app to flash the "Token Not Available" modal immediately when switching providers, before payment methods had even been fetched.

By using react-query, we now have explicit status tracking (`idle` | `loading` | `success` | `error`) plus `isFetching` for background refetches. The "Token Not Available" modal only shows when `paymentMethodsStatus === 'success' && !paymentMethodsFetching && paymentMethods.length === 0`.

### 2. Fix token unavailability detection across all Buy entry points

The `isTokenUnavailable` check in `BuildQuote` previously relied solely on `params?.assetId` (route navigation params). This worked for the **token buy** flow (`Home → Tokens → Token Info → Buy`) which passes `assetId` via `createBuildQuoteNavDetails`, but failed for the **home buy** flow (`Home → Buy → Token Selection → BuildQuote`) where `params.assetId` could be missing or stale.

The fix: treat `route.params.assetId` as bootstrapping input only. Once on the BuildQuote screen, derive the effective asset ID from the controller state (`selectedToken?.assetId`) as the source of truth, falling back to `params?.assetId`.

### 3. Fix navigation behavior for three distinct Buy flow origins

The "Token Not Available" modal previously had a single dismiss/change-token behavior regardless of how the user entered the Buy flow. This caused a crash (`cannot read property 'address' of undefined`) when navigating back via `navigation.navigate('Asset')` without params from the token info flow.

Introduced `BuyFlowOrigin` type (`'tokenInfo' | 'homeTokenList' | undefined`) to distinguish the three entry points and handle navigation correctly:

- **Token info flow** (`tokenInfo`): "Change token" → Tokens Full View; dismiss → `goBack()` twice (exits ramp flow, preserves Asset screen params)
- **Home token list flow** (`homeTokenList`): "Change token" → Home; dismiss → Home
- **Default/home buy flow** (`undefined`): "Change token" → Token Selection; dismiss → Token Selection

Additional fixes:
- 600ms debounce on modal navigation to prevent flashing when `isTokenUnavailable` is briefly true due to stale cached data
- `lastShownUnavailableKeyRef` dedup pattern (`providerId:assetId`) to prevent duplicate modal navigations
- `focusTrigger` counter via `useFocusEffect` to force modal re-evaluation on screen re-focus (e.g., after returning from provider picker)
- `isOnBuildQuoteScreen` guard to prevent effects firing when screen is in background
- Payment pill disabled when token is unavailable (`onPress={isTokenUnavailable ? undefined : handlePaymentPillPress}`)

### Changes

- **New `queries/paymentMethods.ts`** — react-query `queryOptions` for fetching payment methods via `RampsController.getPaymentMethods`, keyed by region/fiat/assetId/providerId
- **New `queries/quotes.ts`** — react-query `queryOptions` for fetching quotes via `RampsController.getQuotes`, keyed by assetId/amount/wallet/paymentMethod/provider
- **New `queries/index.ts`** — barrel export combining both query definitions
- **Updated `useRampsPaymentMethods.ts`** — rewired to use `useQuery` instead of reading from Redux store; exposes `status`, `isFetching`, `isSuccess`, `isLoading`, and `error`
- **Updated `useRampsQuotes.ts`** — rewired to use `useQuery` with proper `enabled` gating; exposes `status`, `isSuccess`, `loading`, and `error`
- **Updated `useRampsController.ts`** — passes through `paymentMethodsStatus` and `paymentMethodsFetching` from the payment methods hook
- **Updated `BuildQuote.tsx`** — derives `effectiveAssetId` from controller state (source of truth) with route params as fallback; uses `paymentMethodsStatus === 'success' && !paymentMethodsFetching` check before determining token unavailability; 600ms debounced modal navigation with dedup; `BuyFlowOrigin` passthrough; disabled payment pill when unavailable
- **Updated `TokenNotAvailableModal.tsx`** — handles three flows via `buyFlowOrigin` param with correct navigation for each
- **Updated `useRampNavigation.ts`** — passes `buyFlowOrigin` through to `createBuildQuoteNavDetails`
- **Updated `useTokenActions.ts`** — passes `{ buyFlowOrigin: 'tokenInfo' }` to `goToBuy`
- **Updated `PopularTokenRow.tsx`** — passes `{ buyFlowOrigin: 'homeTokenList' }` to `goToBuy`
- **Unit tests** — added/updated tests for `BuildQuote`, `TokenNotAvailableModal`, `useRampNavigation`, `useRampsPaymentMethods`, `useRampsQuotes`, `useRampsController`, and query definitions

## **Changelog**

CHANGELOG entry: Fixed false "Token Not Available" errors during Buy flow when payment methods are still loading after provider change; fixed missing "Token Not Available" modal in home buy flow; fixed crash when navigating back from "Token Not Available" modal in token info buy flow

## **Related issues**

Refs: [TRAM-3314](https://consensyssoftware.atlassian.net/browse/TRAM-3314)

### Step-by-step walkthrough for testing

**Test 1 — Home buy flow (the fixed path):**
1. Launch the app in the iOS
2. Tap **Buy** on the home screen
3. In the Token Selection screen, pick a niche token (e.g. **TRX on Tron**)
4. On the BuildQuote screen, verify the **"Token Not Available" modal** appears after loading
5. Press "Change token" — should go back to token selection
6. Press "Change provider" — should open provider picker

**Test 2 — Token buy flow (existing path, verify no regression + crash fix):**
1. From the Home screen, tap on a token in the Tokens list (e.g. **TRX**)
2. On the Token Info screen (with graph), tap **Buy**
3. Verify the **"Token Not Available" modal** appears after loading
4. Press "Change token" — should go to Tokens Full View
5. Dismiss the modal — should go back to Token Info (NOT crash)

**Test 3 — Home token list buy flow:**
1. From the Home screen, tap the **Buy** button next to a token in the token list
2. If the token is unavailable, verify the modal appears
3. Press "Change token" — should go to Home screen
4. Dismiss — should go to Home screen

**Test 4 — Re-selecting tokens:**
1. Enter via token buy flow (Token Info → Buy) with a supported token
2. Go back to token selection
3. Select a token that is unavailable with the current provider
4. Verify the "Token Not Available" modal appears (previously it did NOT in this path)

**Test 5 — Modal re-appears after provider change:**
1. Enter Buy flow with an unavailable token
2. After modal appears, tap "Change provider"
3. Select another provider that also doesn't support the token
4. Verify the modal appears again (previously it only showed once due to dedup)

### **Before**

https://github.com/user-attachments/assets/efff2d42-698c-4084-8e64-8857852bd34a

### **After**

https://github.com/user-attachments/assets/e8803901-9d83-4803-a324-1787c0a630ea

https://github.com/user-attachments/assets/e2f8b51a-4e12-44c6-8d75-d83d8083929e

https://github.com/user-attachments/assets/37a56f6c-af24-4cf4-85a6-00929312b41f

### **After rebasing 7.71.0**

https://github.com/user-attachments/assets/db1da8ea-3641-4302-95cf-39735b6d9260

https://github.com/user-attachments/assets/d9598eda-2f5a-4788-b523-6e3b0738a189

https://github.com/user-attachments/assets/930866b3-0a75-454c-b5cc-96e132a9df7e

---

> [!NOTE]
> **Medium Risk**
> Touches the core Buy flow by changing how payment methods/quotes are fetched and how the Token Unavailable modal is triggered/navigated, which can affect ramp availability and navigation behavior across entry points.
> 
> **Overview**
> **Improves Buy-flow reliability by migrating `paymentMethods` and `quotes` fetching to `@tanstack/react-query`** (new `rampsQueries` with stable keys/options) and exposing richer request state (`status`, `isFetching`, `isSuccess`) through `useRampsPaymentMethods`, `useRampsQuotes`, and `useRampsController`.
> 
> **Fixes Token Unavailable handling in `BuildQuote`** by deriving the effective token from controller state (not just route params), gating the modal on settled payment-methods state, and adding focus-aware, de-duped, 600ms-debounced modal navigation; it also disables the payment-method pill when the token is unavailable.
> 
> **Adds `BuyFlowOrigin` plumbing (`tokenInfo` | `homeTokenList`)** through `useRampNavigation`/callers and updates `TokenNotAvailableModal` to return the user to the correct screen on change-token/dismiss for each entry path. Tests are updated/added to cover the new query state and navigation behaviors, and the Ramp routes are wrapped in a `QueryClientProvider`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45f4c0af8dd0173c6773ff1ac9e4c25d62fb92d3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/UI/Ramp/queries/quotes.ts
```diff
@@ -0,0 +1,53 @@
...
+  }: RampsQuotesQueryParams) =>
+    [
+      ...rampsQuotesKeys.all(),
+      assetId ?? '',
+      amount,
+      walletAddress,
+      paymentMethodId,
+      providerId,
+    ] as const,
+};
```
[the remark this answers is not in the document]
imyugioh: that is intentional - `forceRefresh` controls how the controller fetch internally and that is not what data is being requested. so with `staleTime: 0` react-query always refetches on mount, so the queryFn always runs with `forceRefresh: true` and that hits the api directly.

app/components/UI/Ramp/hooks/useRampsPaymentMethods.ts
```diff
@@ -38,12 +59,27 @@ export interface UseRampsPaymentMethodsResult {
...
+
+  const paymentMethodsQuery = useQuery({
+    ...rampsQueries.paymentMethods.options({
+      regionCode: userRegion?.regionCode ?? '',
+      fiat: userRegion?.country?.currency ?? '',
+      assetId: selectedToken?.assetId ?? '',
+      providerId: selectedProvider?.id ?? '',
+    }),
+    enabled: queryEnabled,
+  });
```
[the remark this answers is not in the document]
imyugioh: QueryClientProvider is already set up at `app/components/Views/Root/index.tsx:81`, wrapping the entire app. All ramp screens are children of that tree. The only real instance of this error was in custom-amount-info.test.tsx (a test without the root provider), which was fixed by mocking useRampsPaymentMethods.

app/components/UI/Ramp/Views/BuildQuote/BuildQuote.tsx
```diff
@@ -174,33 +192,104 @@ function BuildQuote() {
...
-      !!(
-        selectedProvider &&
-        params?.assetId &&
-        selectedProvider.supportedCryptoCurrencies &&
-        !selectedProvider.supportedCryptoCurrencies[params.assetId]
-      ),
-    [selectedProvider, params?.assetId],
+  const tokenStateIsSettled =
+    !params?.assetId || selectedToken?.assetId === params.assetId;
+
```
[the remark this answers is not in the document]
imyugioh: Screenshot 2026-03-17 173913

app/components/UI/Ramp/Views/BuildQuote/BuildQuote.tsx
```diff
@@ -174,33 +192,104 @@ function BuildQuote() {
...
+  // Controller state is the source of truth for the active token;
+  // route params are only used as bootstrapping input.
+  const effectiveAssetId = selectedToken?.assetId ?? params?.assetId;
```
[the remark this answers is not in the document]
imyugioh: Screenshot 2026-03-17 174007

app/components/UI/Ramp/hooks/useRampsController.ts
```diff
@@ -145,6 +147,8 @@ export function useRampsController(): UseRampsControllerResult {
     selectedPaymentMethod,
     setSelectedPaymentMethod,
     isLoading: paymentMethodsLoading,
+    isFetching: paymentMethodsFetching,
```
georgeweiler, line 150: What is the difference between `isLoading` and `isFetching`? could we combine this state? Does it represent different statuses?

app/components/UI/Ramp/hooks/useRampsPaymentMethods.ts
```diff
@@ -38,12 +59,27 @@ export interface UseRampsPaymentMethodsResult {
...
-    selected: selectedPaymentMethod,
-    isLoading,
-    error,
-  } = useSelector(selectPaymentMethods);
+  const { selected: selectedPaymentMethod } = useSelector(selectPaymentMethods);
+  const { selected: selectedProvider } = useSelector(selectProviders);
+  const { selected: selectedToken } = useSelector(selectTokens);
+  const userRegion = useSelector(selectUserRegion);
+
+  const queryEnabled = Boolean(
```
georgeweiler, line 73: Should we also check for `selectedProvider.supportedCryptoCurrencies` here? Otherwise we will fetch tokens for a provider who we expect to return `[]`.


We only show the "Token Unavailable" modal when  this hook returns `success`
https://github.com/MetaMask/metamask-mobile/pull/27448/changes#diff-dd73417979ddc55353f3601fa4a46d46d6d10830fd7bcd70cc78545529610ab0R212


and in the controller side we do not request payment methods if the provider does not support the token
https://github.com/MetaMask/core/blob/main/packages/ramps-controller/src/RampsController.ts#L1209-L1213

georgeweiler: The controller comment there says why: for an unsupported token, the payment-methods request would “fail or return empty for the wrong reason,” and the UI should show the Token Not Available modal instead. So I think we should avoid fetching.

app/components/UI/Ramp/Views/Modals/TokenNotAvailableModal/TokenNotAvailableModal.tsx
```diff
@@ -72,11 +76,25 @@ function TokenNotAvailableModal() {
         .build(),
     );
     sheetRef.current?.onCloseBottomSheet(() => {
```
georgeweiler, line 77: I love this idea of `goToBuy` taking a return/origin route so the user gets back to the familiar page when they close the modal.

However I think it could be more extendable if we just accept a redirection _**route**_ instead of this enum.  The `goToBuy` method is called in a few places in the app (more than we've added here).  Each of them could pass their specific route (or routing callback?) instead of us having to keep this record of entry points and make sure we are sending then back to the right place.

what do you think?

Image

app/components/UI/Ramp/hooks/useRampsController.ts
imyugioh: @georgeweiler yes, those are react-query concepts with different purposes - isLoading is isPending internally : that is true only on the initial fetch when there is no cached data yet. and that is used for showing loading spinners.
but isFetching is true whenever any fetch is in-flight, like including background refetches when cached data already exists. I thought we need this specifically in the BuildQuote to guard the "Token Not Available " modal

imyugioh: without it - the modal could flash during background refetches (the dedup ref clears too early while stale data shows  paymentMethods: [] as empty)

app/components/UI/Ramp/hooks/useRampsPaymentMethods.ts
imyugioh: Good catch - that matches what the controller does at RampsController. we can add an early-exit check in queryEnabled

app/components/UI/Ramp/Views/Modals/TokenNotAvailableModal/TokenNotAvailableModal.tsx
imyugioh: OK @georgeweiler - I also like the idea in principle - a generic redirect route would be more extensible, i think. currently goToBuy  is called from several places  - but most of them don't pass any origin/route at all. 
The trade-off is that the current enum approach keeps redirection logic centralized in tokennotavailableModal for 3 cases of buy flow. 
But for this PR's scope, i'd prefer to keep the enum and open a follow-up to explore the generic redirect approach
we can refactor goToBuy to accept an optional onDismissroute without touching the core fix. 
What do you think?

georgeweiler (dismissed):

Prithpal-Sooriya (dismissed):

app/components/UI/Ramp/hooks/useRampsPaymentMethods.ts
```diff
@@ -38,12 +59,34 @@ export interface UseRampsPaymentMethodsResult {
...
+  const { selected: selectedPaymentMethod } = useSelector(selectPaymentMethods);
+  const { selected: selectedProvider } = useSelector(selectProviders);
+  const { selected: selectedToken } = useSelector(selectTokens);
+  const userRegion = useSelector(selectUserRegion);
+
+  const tokenSupportedByProvider = selectedProvider?.supportedCryptoCurrencies
+    ? selectedProvider.supportedCryptoCurrencies[
+        selectedToken?.assetId ?? ''
+      ] === true
+    : true;
```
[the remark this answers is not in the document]
imyugioh: False positive - the type is Record<string, boolean> (RampsService.ts:94 in core). values  are always boolean so never truthy non-booleans like 1. 
The ===true vs truthness distinction is irrelevant here.

Prithpal-Sooriya (approved):

OGPoyraz (approved):

amitabh94 (approved):

PR #27527, opened 2026-03-17
fix: correct onboarding button colors in dark mode

tylerc-consensys:
## **Description**

The design system's `isInverse` prop on `Button` (Primary and Secondary variants) resolves to theme-aware tokens in dark mode:

- **Primary + isInverse**: `bg-default` → dark grey background, `text-default` → muted text
- **Secondary + isInverse**: `text-primary-inverse` → muted white text

The onboarding screen uses a custom purple background (`gettingStartedTextColor`) that doesn't participate in the standard light/dark theme system. As a result, the buttons appear washed out in dark mode instead of crisp white/black.

### Fix

Override the design system token colors with explicit values via `style` and `textProps` to ensure correct contrast on the custom purple background:

- **Primary "Create a new wallet"**: Force `white` background + `applePayBlack` text
- **Secondary "I have an existing wallet"**: Force `white` text

These use existing `importedColors` constants, not arbitrary hex values.

### Why `style` instead of `twClassName`

The `style` prop is used intentionally here rather than `twClassName` for the following reasons:

1. **`twClassName` can't reliably override the Button's internal Tailwind classes.** The design system `Button` internally computes its own `twClassName` from `isInverse`/`isDanger` state (e.g. `bg-default`). Passing an additional `twClassName` would concatenate with the internal classes, and last-wins resolution in Tailwind is unpredictable. React Native `style` always takes precedence because it's applied after `tw.style()` in `ButtonBase`.

2. **No theme-agnostic design token exists for this use case.** All design tokens are theme-aware by design. The onboarding screen's custom purple background is a non-themeable surface — it doesn't change between light and dark mode — so the buttons on it need non-themeable colors. There is no "always white" or "always black" semantic token.

3. **`style` is a first-class prop on the design system `Button`.** It's part of the `ButtonBase` API and is explicitly merged via `computedStyle.push(additionalStyle)` in the component source. This is not a hack — it's the supported escape hatch for cases where tokens don't apply.

## **Related issues**

Refs: Design system migration for onboarding screens (TO-554)

## **Manual testing steps**

```gherkin
Feature: Onboarding button contrast in dark mode

  Scenario: buttons are readable in dark mode
    Given the device is set to dark mode
    And the app is freshly installed

    When the onboarding screen appears
    Then "Create a new wallet" has a white background with black text
    And "I have an existing wallet" has a black background with white text

  Scenario: buttons are readable in light mode
    Given the device is set to light mode
    And the app is freshly installed

    When the onboarding screen appears
    Then "Create a new wallet" has a white background with black text
    And "I have an existing wallet" has a black background with white text
```

### Dark mode

| Before (bug) | After (fix) |
|:---:|:---:|
| Primary button appears dark grey; secondary button text is barely readable | Both buttons now have correct white/black contrast |
| Before - dark mode | After - dark mode with legacy comparison |

> _After screenshot includes legacy buttons at the bottom for visual comparison — they are not part of the final change._

### Light mode (no regression)

| After (light mode) |
|:---:|
| Buttons render identically to before — no visual change in light mode |
| After - light mode |

- [x] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md)
- [x] I've applied the right labels on this PR
- [x] I've manually tested this PR
- [x] I've confirmed there are no breaking changes

## **Changelog entry**

CHANGELOG entry: Fixed onboarding button colors appearing washed out in dark mode


grvgoel81 (approved):
lgtm

PR #27139, opened 2026-03-06
fix: try fixing X deeplinks by adding $deeplink_path param [GE-139] cp-7.69.0

baptiste-marchand:
## **Description**

Fix branch deeplinks not working with the X (Twitter) app by adding a `$deeplink_path` param in branch.io and consuming it in MetaMask Mobile.

### Context

- Platform: happens only on iOS. Android behaves differently (no Deepview in our flow).
- User taps a t.co link in X (Twitter) → redirects to https://metamask.app.link/1WkF6GmE40b (which is a link generated by X/Branch.io on Tweet posting) → Branch Deepview is shown → user taps “Get the app” → opens MetaMask via Universal Link: https://metamask-alternate.app.link/1WkF6GmE40b?__branch_flow_type=viewapp&__branch_flow_id=...&__branch_mobile_deepview_type=1.
- Issue: The app opens correctly, but in-app we show a “page not found” (404). So the OS and Branch open the app, but we don’t know which in-app route corresponds to link ID 1WkF6GmE40b.

### What we found

1. No resolved URI from the SDK
- The Branch iOS SDK (and react-native-branch) give us:
  - The raw URL that opened the app (e.g. https://metamask-alternate.app.link/1WkF6GmE40b?...), and
  - The params from the session (e.g. from getLatestReferringParams() / subscribe callback).
- The SDK does not build a custom deeplink URI (e.g. metamask://swap) from link data; it only returns the params. So we have to do routing ourselves.

2. Implemented fix:
- Added a $deeplink_path in Branch deeplinks (this must be added to every existing and future deeplinks that we want to post on X)
- Read it from the params and build our deeplink ourselves (e.g. metamask://${params.$deeplink_path}) and then route inside the app.

## **Changelog**

CHANGELOG entry: Fixed a bug that was causing deeplinks opened from X app to fail

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/GE-139
Fixes: https://github.com/MetaMask/metamask-mobile/issues/27140

## **Manual testing steps**

```gherkin
Feature: my feature name

  Scenario: user [verb for user action]
    Given [describe expected initial app state]

    When user [verb for user action]
    Then [describe expected outcome]
```

## **Pre-merge author checklist**

[...]

PR #27645, opened 2026-03-18
feat: Clean up incoming deeplink Logic for Branch SDK and Linking API

MarioAslau:
## **Description**

Cleans up the incoming deeplink logic for the Branch SDK and React Native Linking API, and fixes the X (Twitter) iOS deeplink bug where tapping a deeplink from X lands on "this page doesn't exist" or does nothing instead of navigating to the correct screen.

**Why:** The deeplink entry pipeline had redundancies, dead code, a bug in `rewriteBranchUri`, and a race condition where both the Linking API and Branch SDK could fire for the same incoming link. Additionally, when deeplinks arrive from X (Twitter) via Branch's Deepview intermediate page, the Branch SDK fails to resolve the short link — returning `+clicked_branch_link=false` and no `$deeplink_path`. This left the app with an unroutable Branch short link URL (e.g. `metamask-alternate.app.link/1WkF6GmE40b`) that matched no supported action. For links with query params (e.g. `link.metamask.io/buy?chainId=59144&amount=25`), X wraps them in a generic Branch link whose Deepview returns `$deeplink_path: "open"` — losing both the correct path and all query params.

**What changed:**

1. **Fix `rewriteBranchUri` return value** — Returns `undefined` instead of the original URI when it cannot rewrite (when `+clicked_branch_link` is false or `$deeplink_path` is absent). Previously returned a truthy, unresolvable Branch short link that was dispatched as a deeplink and caused 404s.

2. **Remove `.filter(Boolean)` dead code** — Removed unnecessary `.filter(Boolean)` from the `METAMASK_HOSTS` array in `util/deeplinks/index.ts`. Every entry already has an `||` fallback to a string literal, so no entry can ever be falsy.

3. **Eliminate redundant `getLatestReferringParams` call** — `handleUniversalLink` was calling `branch.getLatestReferringParams()` a second time (with a 500ms timeout/Promise.race) for analytics context. Now Branch params are cached on the `DeeplinkManager` instance when first fetched and read directly in `handleUniversalLink`. Removed the `react-native-branch` and `Logger` imports from `handleUniversalLink.ts`.

4. **Resolve Linking API / Branch SDK race condition** — Added `isBranchDomainUrl()` helper to identify Branch domain URLs (`metamask.app.link`, `metamask-alternate.app.link`). Both `Linking.getInitialURL()` and `Linking.addEventListener('url')` now skip Branch domain URLs, since those are exclusively handled by the Branch SDK. This prevents duplicate processing where both entry points would fire for the same link.

5. **Fix `branch.subscribe` fallback for unresolved short links** — Previously, when `rewriteBranchUri` returned `undefined` in the `branch.subscribe` handler, the code fell back to `opts.uri` (the raw Branch short link). This caused the app to try routing a short link path like `/1WkF6GmE40b` as an action, which failed. Now the fallback only uses `opts.uri` when it's NOT a Branch domain URL.

6. **Add `resolveBranchShortLink` fallback for Deepview flows** — When the Branch SDK returns `+non_branch_link` pointing to a Branch domain (which happens with X/Twitter Deepview links), the SDK has failed to resolve the short link. A new `resolveBranchShortLink()` function fetches the short link URL directly (using a crawler User-Agent so Branch returns the metadata page), follows redirects, and resolves the original deep link using a tiered fallback chain. This handles both the `branch.subscribe` warm-start path and the `getBranchDeeplink` cold-start path. Includes a 3-second `AbortController` timeout, domain validation on `+non_branch_link`, and `.catch()` on the promise chain.

7. **Add `extractBranchUrlFromDeepview` for reliable Deepview resolution** — Branch Deepview HTML embeds the full original URL in the `al:ios:url` meta tag as a base64-encoded JSON blob inside the `link_click_id` parameter. This new function decodes that blob to extract the `+url` field (e.g. `https://link.metamask.io/buy`), which is the authoritative original URL configured in the Branch dashboard. This is the primary resolution method and is more reliable than regex-matching `$deeplink_path` or parsing CTA button hrefs from the Deepview HTML. Query params from the original Branch URL (e.g. `chainId`, `address`, `amount`) are merged onto the resolved `+url`.

8. **Add `stripBranchDeepviewParams` helper** — Strips Branch Deepview query params (`__branch_flow_type`, `__branch_flow_id`, `__branch_mobile_deepview_type`, `_referrer`) from URLs before resolution. Preserves MetaMask's own signed deeplink params (`sig`, `sig_params`) and user-facing params. These Deepview params are appended by Branch's intermediate page and can interfere with link resolution.

9. **Add `extractDeepviewPath` fallback** — Parses the Deepview CTA button href or `window.top.location` assignment for the deeplink path. Handles `null`-prefixed paths (e.g. `nulltrending`), `metamask://` scheme, and full `https://link.metamask.io/` URLs. Used as a last-resort fallback when `link_click_id` decoding and `$deeplink_path` regex both fail.

10. **Use crawler User-Agent for Branch metadata fetch** — Branch serves different HTML based on User-Agent. Browser/app UAs get the visual Deepview page (no `al:ios:url`), while crawler UAs get the metadata page with `al:ios:url` containing the `link_click_id`. Changed the fetch User-Agent to `facebookexternalhit/1.1 (MetaMask-LinkResolver)` to ensure we receive the metadata page needed for `link_click_id` decoding.

## **Changelog**

CHANGELOG entry: Fixed deeplinks from X (Twitter) on iOS not landing on the correct screen by decoding Branch Deepview metadata to recover the original deep link URL with full path and query params; resolved duplicate processing race condition between Branch SDK and Linking API; added 3-second timeout on Branch link resolution fetch

## **Related issues**

Fixes: #27140
Refs: #27139
Jira Ticket: https://consensyssoftware.atlassian.net/browse/MCWP-386?atlOrigin=eyJpIjoiNjUyYjFmMmVkYWM3NDk1MDliNTNiYzFhYTE3YTcwZWUiLCJwIjoiaiJ9

## **Manual testing steps**

```gherkin
Feature: Deeplinks from external apps

  Scenario: User taps a trending deeplink from X (Twitter) on iOS cold start
    Given the MetaMask app is not running
    When user taps the MetaMask trending deeplink shared on X (e.g. from @MetaMask tweet)
    And the Branch Deepview intermediate page is shown in X's in-app browser
    And user taps the "Open" / "Get the app" button on the Deepview
    Then the app opens and navigates to the Explore/Trending screen

  Scenario: User taps a trending deeplink from X (Twitter) on iOS warm start
    Given the MetaMask app is running in the background
    When user taps the MetaMask trending deeplink shared on X
    And the Branch Deepview intermediate page is shown
    And user taps the "Open" / "Get the app" button
    Then the app navigates to the Explore/Trending screen

  Scenario: User taps a buy deeplink with query params from X (Twitter) on iOS
    Given the MetaMask app is installed
    When user taps a link like https://link.metamask.io/buy?chainId=59144&address=0x...&amount=25 shared on X
    And the Branch Deepview intermediate page is shown
    And user taps the "Open" / "Get the app" button
    Then the app opens and navigates to the Buy screen with chainId, address, and amount pre-filled

  Scenario: User taps a deeplink from Slack on iOS cold start
    Given the MetaMask app is not running
    When user taps a MetaMask deeplink (e.g. https://link.metamask.io/buy) from Slack
    Then the app opens and navigates to the correct screen (Buy)

  Scenario: User taps a deeplink from Slack on iOS warm start
    Given the MetaMask app is running in the background
    When user taps a MetaMask deeplink from Slack
    Then the app navigates to the correct screen

  Scenario: User taps a deeplink from Slack on Android cold start
    Given the MetaMask app is not running
    When user taps a MetaMask deeplink from Slack
    Then the app opens and navigates to the correct screen

  Scenario: User taps a deeplink from Slack on Android warm start
    Given the MetaMask app is running in the background
    When user taps a MetaMask deeplink from Slack
    Then the app navigates to the correct screen

  Scenario: User taps a deeplink from X (Twitter) on Android cold start
    Given the MetaMask app is not running
    When user taps a MetaMask deeplink shared on X
    Then the app opens and navigates to the correct in-app destination

  Scenario: User taps a deeplink from X (Twitter) on Android warm start
    Given the MetaMask app is running in the background
    When user taps a MetaMask deeplink shared on X
    Then the app navigates to the correct in-app destination

  Scenario: Custom scheme deeplinks still work
    Given the MetaMask app is installed
    When user opens a metamask:// or ethereum: deeplink from any source
    Then the app handles it correctly via Linking API

  Scenario: Push notification deeplinks still work
    Given the MetaMask app receives a push notification with a deeplink
    When user taps the notification
    Then the app navigates to the correct screen
```

### **Before**

N/A (logic-only changes, no UI)

### **After**

https://github.com/user-attachments/assets/44de4e4d-a8de-43d7-9f7f-7e59392511bb

N/A (logic-only changes, no UI)

---

> [!NOTE]
> **Medium Risk**
> Changes core deeplink entry logic (Branch + React Native Linking) and adds network-based short-link resolution, which could affect navigation/attribution and introduce edge-case regressions despite extensive tests.
> 
> **Overview**
> Improves incoming deeplink processing by **preventing unroutable Branch short links from being dispatched**: `rewriteBranchUri` now returns `undefined` when it can’t rewrite, and `DeeplinkManager.start` no longer falls back to handling raw Branch-domain URLs.
> 
> Adds Branch-specific utilities to **avoid duplicate processing and recover Deepview/X (Twitter) links**: `isBranchDomainUrl` filters Branch domains out of the React Native `Linking` pipeline, and `resolveBranchShortLink` fetches/decodes Deepview metadata (including `link_click_id` → `+url`) with param merging, Deepview-param stripping, and timeouts.
> 
> Branch params are now **cached on `DeeplinkManager`** and consumed by `handleUniversalLink` for analytics (removing extra Branch SDK calls). Tests were expanded significantly to cover cold/warm start flows, error handling, and Deepview resolution; `bitrise.yml` build numbers were bumped.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e7ecd631aa0aca1380135099b9286dfd4034d080. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


bitrise.yml
```diff
@@ -3534,13 +3534,13 @@ app:
       VERSION_NAME: 7.71.0
     - opts:
         is_expand: false
-      VERSION_NUMBER: 3911
```
tommasini, line 3537: Can we revert here as well?

app/core/DeeplinkManager/DeeplinkManager.ts
```diff
@@ -9,6 +9,219 @@ import FCMService from '../../util/notifications/services/FCMService';
...
+ * that can confuse the Branch SDK's link resolution.
+ */
+export function stripBranchDeepviewParams(url: string): string {
+  try {
+    const parsed = new URL(url);
+    const paramsToStrip = [
+      '__branch_flow_type',
+      '__branch_flow_id',
+      '__branch_mobile_deepview_type',
+      '_referrer',
```
tommasini, line 29: I probably am missing context, but why do we need to strip this params?

app/core/DeeplinkManager/DeeplinkManager.ts
```diff
@@ -9,6 +9,219 @@ import FCMService from '../../util/notifications/services/FCMService';
...
+  try {
+    const cleanUrl = stripBranchDeepviewParams(shortLinkUrl);
+
+    const controller = new AbortController();
+    const timeout = setTimeout(() => controller.abort(), 3000);
+
+    const response = await fetch(cleanUrl, {
+      redirect: 'follow',
+      headers: {
+        'User-Agent': 'facebookexternalhit/1.1 (MetaMask-LinkResolver)',
```
tommasini: What is this header about?

app/core/DeeplinkManager/DeeplinkManager.ts
MarioAslau: These params are injected by Branch's Deepview intermediate page (shown when opening links from X/Twitter's in-app browser). They are not part of the original link.

When a user taps a Branch link from X, the Deepview appends these tracking params onto the URL before handing it to the OS to open the app:

__branch_flow_type: Branch internal Deepview flow type
__branch_flow_id: Branch internal session ID for this Deepview instance
__branch_mobile_deepview_type: which Deepview template was used
_referrer: which app the user came from (e.g. twitter)
We strip them because resolveBranchShortLink fetches this URL server-side to resolve the short link. Including these Deepview-injected params would interfere with Branch's server-side resolution and pollute the final resolved URL with irrelevant internal tracking data.

For example, the original short link arrives as:

https://metamask-alternate.app.link/DudG79nFJ0b?__branch_flow_type=viewapp&__branch_flow_id=1563308585802406280&__branch_mobile_deepview_type=1&chainId=59144&address=0xABC&amount=25&_referrer=twitter
After stripping, only the user-facing params remain:

https://metamask-alternate.app.link/DudG79nFJ0b?chainId=59144&address=0xABC&amount=25
Those preserved params (chainId, address, amount) are then merged onto the resolved destination URL.

app/core/DeeplinkManager/DeeplinkManager.ts
MarioAslau: Branch serves different HTML responses depending on the User-Agent header:

* Browser/app UAs (e.g. Safari, Chrome, MetaMask-Mobile/1.0) → returns the visual Deepview landing page, which does not include the al:ios:url meta tag we need
* Crawler/bot UAs (e.g. facebookexternalhit, Googlebot, Twitterbot) → returns the metadata/App Links page, which does include the al:ios:url meta tag containing the link_click_id with the full original URL
We use facebookexternalhit/1.1 because it's a well-known social media crawler UA that Branch is designed to serve App Links metadata to, it's the same UA Facebook uses to scrape link previews. The (MetaMask-LinkResolver) suffix identifies our usage.

We verified this behavior with curl:

curl -H "User-Agent: MetaMask-Mobile/1.0" → 0 matches for al:ios:url
curl -H "User-Agent: facebookexternalhit/1.1" → 1 match for al:ios:url
curl (default UA) → 1 match for al:ios:url

Without this UA, extractBranchUrlFromDeepview would never find the al:ios:url meta tag and the resolution would fall through to the less reliable $deeplink_path regex / extractDeepviewPath fallbacks.

app/core/DeeplinkManager/DeeplinkManager.ts
```diff
@@ -9,6 +9,219 @@ import FCMService from '../../util/notifications/services/FCMService';
...
+      },
+      signal: controller.signal,
+    });
+
+    clearTimeout(timeout);
```
NicolasMassart: suggestion (non-blocking): Clear the abort timer when fetch fails

If fetch rejects before clearTimeout(timeout), the 3s timer still runs until it fires. That is a small leak and extra work on failure paths. Using try/finally (or finally after try/catch) around the fetch + clearTimeout keeps behavior obvious.

Try something like:
```ts
const controller = new AbortController();
const timeout = setTimeout(() => controller.abort(), 3000);
try {
  const response = await fetch(cleanUrl, {
    redirect: 'follow',
    headers: {
      'User-Agent': 'facebookexternalhit/1.1 (MetaMask-LinkResolver)',
    },
    signal: controller.signal,
  });
  // ... existing body handling ...
} finally {
  clearTimeout(timeout);
}
```

NicolasMassart (commented):
Strong unit coverage. I’d clear the `AbortController` timeout in a finally though.
Also I’d still skim Sonar’s new issues on the PR even though the quality gate passed.

tommasini (approved):

tommasini (changes requested):
It seems that we should put this on hold for now https://consensys.slack.com/archives/C02U025CVU4/p1774278680610519?thread_ts=1772742579.747479&cid=C02U025CVU4

PR #27735, opened 2026-03-20
perf: Speed up the creation of the token list for Predict withdraw

dan437:
## **Description**

This speeds up the withdraw token-selection path by moving the allowlist filter earlier in the pipeline.

Before this change, withdraw flows loaded the full token catalog, built `AssetType`s for all matching catalog entries, sorted the full result, and only then filtered down to the small allowlisted set we actually care about. This change adds a `tokenFilter` hook parameter so `useAccountTokens` can skip irrelevant catalog entries before building them, and `useWithdrawTokenFilter` passes the withdraw allowlist down to that lower-level loop.

It also keeps the existing post-filter in place for correctness, so allowlisted tokens already owned by the user and native-token address mapping still work as before.

Local dev measurements while opening **Predict Withdraw**:

| Metric | Before | After |
| --- | ---: | ---: |
| `useAccountTokens` | 954-1,143ms | 41-48ms |
| Catalog tokens built | 18,524 | 0 |
| Total tokens sorted | 18,618 | 15 |

This brings the measured token-loading path from roughly `~1.05s` down to `~44ms` (`~24x` faster).

## **Changelog**

CHANGELOG entry: Improved performance when loading Predict withdraw token selection

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/CONF-1054

## **Manual testing steps**
1. Open Predict Withdraw, it will load much faster now

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how withdraw token lists are constructed and filtered, which could inadvertently hide valid tokens if chainId/address normalization or `assetId` handling is wrong. Scope is limited to token-selection hooks and is covered by updated/added unit tests.
> 
> **Overview**
> Speeds up Predict Withdraw token selection by pushing the withdraw allowlist filter down into token collection, avoiding building/sorting the full token catalog before narrowing to allowlisted entries.
> 
> This introduces an optional `tokenFilter(chainId, address)` plumbed from `useWithdrawTokenFilter` → `useSendTokens` → `useAccountTokens`, and replaces the previous post-filtering in `useWithdrawTokenFilter` with a precomputed allowlist lookup that also maps zero-address entries to chain-specific native token addresses.
> 
> Tests are updated to assert `useSendTokens` is called with the new filtering options and to validate `tokenFilter` behavior (case-insensitive matching, chain gating, and native-token mapping), plus new `useAccountTokens` coverage ensuring both owned assets and catalog token building respect `tokenFilter`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b8e1cc795bd49e82c1b2bdb15225bfd71f94f695. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/components/Views/confirmations/hooks/send/useAccountTokens.ts
+++ b/app/components/Views/confirmations/hooks/send/useAccountTokens.ts
@@ -35,6 +35,17 @@ export function useAccountTokens({
     const flatAssets = Object.values(assets).flat();
 
     const assetsWithBalance = flatAssets.filter((asset) => {
+      if (tokenFilter) {
+        const address = asset.assetId;
+        if (
+          !asset.chainId ||
+          !address ||
+          !tokenFilter(asset.chainId, address)
+        ) {
+          return false;
+        }
+      }
+
       if (includeNoBalance) {
         return true;
       }
--- a/app/components/Views/confirmations/hooks/send/useAccountTokens.ts
+++ b/app/components/Views/confirmations/hooks/send/useAccountTokens.ts
@@ -19,9 +19,11 @@ const selectEmptyCache = () => EMPTY_CACHE;
 export function useAccountTokens({
   includeNoBalance = false,
   includeAllTokens = false,
+  tokenFilter,
 }: {
   includeNoBalance?: boolean;
   includeAllTokens?: boolean;
+  tokenFilter?: (chainId: string, address: string) => boolean;
 } = {}): AssetType[] {
   const assets = useSelector(selectAssetsBySelectedAccountGroup);
   const fiatCurrency = useSelector(selectCurrentCurrency);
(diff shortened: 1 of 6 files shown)


app/components/Views/confirmations/hooks/send/useAccountTokens.ts
```diff
@@ -100,6 +102,9 @@ export function useAccountTokens({
           if (existing.has(getAssetKey(chainId, address))) {
             continue;
           }
+          if (tokenFilter && !tokenFilter(chainId, address)) {
```
matthewwalsh0, line 116: Can we also filter `assetsWithBalance` above?

matthewwalsh0 (dismissed):

pedronfigueiredo (approved):

PR #27816, opened 2026-03-23
feat(error-boundary): move ErrorBoundary above providers to catch all render errors

tommasini:
---

## **Description**

Previously, `ErrorBoundary` was mounted as a child of `ThemeProvider`, `NavigationProvider`, `FeatureFlagOverrideProvider`, and all other providers in `Root/index.tsx`. This meant that any error thrown during the initialization or render of those providers would bypass the error boundary entirely, resulting in an unrecoverable blank crash.

This PR moves `ErrorBoundary` to wrap the entire `PersistGate` content tree — above all providers — so that errors from any part of the tree are caught and the user sees the error screen with recovery options rather than an unrecoverable crash.

Two companion fixes were required to make `ErrorBoundary` work correctly outside provider context:

- `useStyles.ts`: falls back to `mockTheme` when `useAppThemeFromContext()` returns `null` or `undefined` (e.g. when a component using `useStyles` is rendered inside `ErrorBoundary` before `ThemeProvider` is available).
- `ErrorBoundary/index.js`: uses optional chaining (`this.context?.colors`) to safely access theme colors when the theme context is absent; also changes the `ScrollView` import back to `react-native` (instead of `react-native-gesture-handler`) since `ErrorBoundary` now renders outside any gesture-handler context.

## **Changelog**

CHANGELOG entry: Fixed unrecoverable crash when a core provider fails to render by moving ErrorBoundary to wrap the entire component tree

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: Error boundary catches provider-level render errors

  Background:
    Given I have the MetaMask Mobile app installed

  Scenario: app displays error screen instead of blank crash when a provider throws
    Given the app is launching for the first time
    And a provider in the component tree throws a render error

    When the app finishes loading
    Then I should see the ErrorBoundary error screen
    And I should be able to tap "Reset app" to recover

  Scenario: normal app launch is unaffected
    Given the app is launching normally
    And no providers throw errors

    When the app finishes loading
    Then I should land on the wallet home screen
    And the app should function normally
```

## **Screenshots/Recordings**
After intentionally throwing an error on `FeatureFlagOverrideProvider`

### **Before**
image

N/A

### **After**

image (6)

N/A

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes the root component hierarchy and error-handling scope, which can affect startup/render order and recovery behavior across the app.
> 
> **Overview**
> **Expands crash coverage at app startup** by moving `ErrorBoundary` in `Root/index.tsx` to wrap all providers (and the snaps webview), so errors thrown during provider initialization/render are caught and surfaced via the error UI instead of a blank crash.
> 
> **Hardens theming and UI dependencies outside provider context** by falling back to `mockTheme` in `useStyles` and `ErrorBoundary` when theme context is missing, and switching `ErrorBoundary`’s `ScrollView` import to `react-native` (updating the Wallet snapshot accordingly).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2f344cfce62d656f846a63bdc2f12937b5054edd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


NicolasMassart (approved):

georgewrmarshall (approved):
LGTM! Approving on behalf of @MetaMask/design-system-engineers.

app/component-library/hooks/useStyles.ts
```diff
@@ -28,7 +28,7 @@ export function useStyles<R, V>(
     | ((params: { theme: Theme }) => R),
   vars?: V,
 ): { styles: R; theme: Theme } {
-  const theme = useAppThemeFromContext();
+  const theme = useAppThemeFromContext() || mockTheme;
```
brianacnguyen, line 31: wait why are we defaulting to mockTheme here?

tommasini: To be consistent with [useTheme()](https://github.com/MetaMask/metamask-mobile/blob/bc83b72c6c39af792884b45145ee8271e3b33511/app/util/theme/index.ts#L190)
```
const colors = this.context?.colors || mockTheme.colors;
```

Although I can totally agree that we should cleanup this fallbacks and find only a solution from when the context is not available? Maybe it will be the same end result

georgewrmarshall: Falling back to `mockTheme` seems like a reasonable safeguard when `themeContext` is unexpectedly `undefined`.

@brianacnguyen raises a good point though. Are we comfortable using `mockTheme` in production, given it’s primarily intended for tests?

tommasini: Error Boundary component will indeed use it if the app crashes before context is defined! The component above on after reflects that since the crash was on `FeatureFlagOverrideProvider` which was one level higher in the tree than `ThemeProvider`! Do you agree?

brianacnguyen: @tommasini so what does it mean for the app theme when the app crashes before context is defined? and what other case do you think might end up defaulting to the mockTheme?

tommasini: No other case unless the theme provider fails to initialise for some reason, then everywhere?

PR #27877, opened 2026-03-24
fix: history source amount when swap/bridge native

maxime-oe:
## **Description**

In the specific gas of a bridge/swap involving the **gas** token **as source** token,
the `quote.srcTokenAmount` value is incorrect because the MetMask fee is already taken out from it.
Ex: For a 10 MON swap, the value for `quote.srcTokenAmount` will be the base units of `9.9125` MON instead of `10`.
Meanwhile, the field `quote.feeData.metabridge.amount` will contain the missing `0.0875` MON value.
There is currently no parameter representing "amount the user requested the swap for".

From a technical point of you. The `quote` object is correct - the MetaMask fee is taken out directly from the source asset, and then only the swap is performed. However from a UX perspective, the user swapped 10 MON in.

This consumer-side solution is to:
- When `bridgeTxHistoryItem.pricingData.amountSent`, use it since it already contains the right information.
- If not (fallback) use `quote.srcTokenAmount` but re-incorporate the fee in the history items using `quote.feeData.metabridge.amount`.
Ex: Before the change, swapping 10 MON for USDC would show as `9.9125` MON in history because of the fee of 0.0875 MON. After the change it correctly shows `10` MON.

## **Changelog**

CHANGELOG entry: re-adds MetaMask fee to `srcAmount` in swap/bridge history when source is native

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **Before**

image
image

### **After**

image
image

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how source amounts are computed/displayed in swap/bridge history and details, which is user-visible and affects fiat calculations; logic is localized and covered by added unit tests.
> 
> **Overview**
> Fixes native-token swap/bridge history amounts that previously appeared reduced by the MetaMask fee by preferring `bridgeTxHistoryItem.pricingData.amountSent` when available.
> 
> Adds `getSourceAmountBaseUnitFromBridgeSwapQuote` in `quoteUtils` to reconstruct the true base-unit source amount by re-adding `feeData.metabridge.amount` *only* when both source and fee assets are the native token on the same chain, and updates `TransactionDetails` plus `decodeSwapsTx`/`decodeBridgeTx` to use this fallback. Unit tests were expanded to cover the new helper and to ensure behavior remains stable when `pricingData` is missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a38f5a08c12c34e8a300148ca08f0061a40d520c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/components/UI/Bridge/utils/transaction-history.ts
khanti42, line 1: Seems `app/components/UI/Bridge/utils/transaction-history.test.ts` needs test to be updated

khanti42 (changes requested):

app/components/UI/Bridge/utils/quoteUtils.ts
```diff
@@ -67,3 +69,31 @@ export const isQuoteExpired = (
...
+  let sourceAmountBaseUnit = quote.srcTokenAmount;
+  // In the specific gas of a bridge/swap involving the gas token as source token,
+  // the `quote.srcTokenAmount` value is incorrect because the MetMask fee is taken out from it.
+  // This consumer-side solution re-incorporates the MetaMask fee when it was paid directly from source token.
+  // Ex: Swapping 10 MON for USD would show as "9.9125 MON" in history because of the fee of 0.0875 MON.
+  // A cleaner solution would have to have 'quote.srcTokenAmount' containing the correct amount
+  // (but risks breaking existing components relying on that field), or create a new quote.originalSrcTokenAmount field.
+  if (
+    isZeroAddress(quote.srcAsset.address) &&
+    isZeroAddress(quote.feeData.metabridge.asset.address)
```
stanleyyconsensys: only native address is handle?  i remember we can take fund from non native token, if we config it from bridge api?

nit:
isZeroAddress is not a good choice to verify if it is native, may be isNativeAddress from bridge controller

app/components/UI/Bridge/components/TransactionDetails/TransactionDetails.tsx
```diff
@@ -227,10 +233,14 @@ export const BridgeTransactionDetails = (
     chainId: sourceChainId,
   };
 
-  const sourceTokenAmount = calcTokenAmount(
-    quote.srcTokenAmount,
-    quote.srcAsset.decimals,
-  ).toFixed(5);
+  const rawSourceAmount = parseFloat(
+    pricingData?.amountSent ||
+      ethers.utils.formatUnits(
```
stanleyyconsensys: i think we can keep using `calcTokenAmount`?

parseFloat will have issue when the number is very big

```suggestion
      calcTokenAmount(
```

app/components/UI/Bridge/utils/quoteUtils.ts
maxime-oe: > only native address is handle? i remember we can take fund from non native token, if we config it from bridge api?
Yes for now my intent is to have a "least impact fix" so since I've only observed the issue specifically with sent native tokens + fee native token, the idea is to limit the fix to that scope in case it would affect other flows negatively.

> isZeroAddress is not a good choice to verify if it is native
I can change it to something that actually checks if it is the native token :+1:

app/components/UI/Bridge/components/TransactionDetails/TransactionDetails.tsx
maxime-oe: Fixed

app/components/UI/Bridge/utils/quoteUtils.ts
maxime-oe: Fixed:
- Using `isNativeAddress` from `@metamask/bridge-controller` for consistency with the feature/controller.
- Added an extra criteria to target the fix even more: fee token and sent token need to have the same chainId (in case in the future we allow the user to pay with a native token from a different chainId)

app/components/UI/Bridge/components/TransactionDetails/TransactionDetails.tsx
```diff
@@ -227,9 +234,10 @@ export const BridgeTransactionDetails = (
     chainId: sourceChainId,
   };
 
-  const sourceTokenAmount = calcTokenAmount(
-    quote.srcTokenAmount,
-    quote.srcAsset.decimals,
+  const sourceTokenAmount = (
+    pricingData?.amountSent
+      ? BigNumber(String(pricingData.amountSent))
+      : calcTokenAmount(quote.srcTokenAmount, quote.srcAsset.decimals)
```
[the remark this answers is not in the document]
maxime-oe: I actually need to fix that

PR #28034, opened 2026-03-27
feat(perps): compliance on ui

aganglada:
## Description

Adds a `gate()` helper to `useComplianceGate` and applies it across every geo-eligibility-gated action in the Perps and TokenDetails flows. Also removes the unused `useAccountGroupCompliance` hook and updates the compliance documentation.

### Motivation

Every flow that needed OFAC compliance checking was copy-pasting the same boilerplate:

```ts
const result = await checkCompliance();
if (isBlocked) {
  showAccessRestrictedModal();
  return;
}
```

This had two problems:
1. **Repetition** — the same 4-line pattern appeared in every action handler.
2. **Stale closure bug** — `isBlocked` was read from the Redux closure at render time, not from the fresh API response. So a wallet could be blocked mid-session and the check would pass if the Redux state hadn't re-rendered yet.

### Changes

**`useComplianceGate` — new `gate()` helper**

`gate(action)` wraps any async action:
1. If compliance is disabled (feature flag off) → runs the action directly (fast path).
2. Calls `checkCompliance()` and reads `blocked` from the **API return value** directly (no stale closure).
3. If blocked → shows `AccessRestrictedModal` and returns `undefined`.
4. If not blocked → runs the action and returns its result.

Existing geo-eligibility guards belong **inside** the `gate` callback — compliance runs first, then your logic:

```ts
const handleTrade = useCallback(
  () =>
    gate(async () => {
      if (!isGeoEligible) {
        showGeoBlockModal();
        return;
      }
      navigateToOrder();
    }),
  [gate, isGeoEligible],
);
```

**Applied to all gated action handlers:**

| File | Gated actions |
|------|---------------|
| `usePerpsHomeActions.ts` | `handleAddFunds` |
| `PerpsMarketDetailsView.tsx` | `handleTradeAction`, `handleClosePosition`, `handleModifyPress`, `handleAutoClosePress`, `handleMarginPress`, `handleAddMarginFromBanner`, `handleSetStopLossFromBanner` |
| `PerpsOrderBookView.tsx` | `handleLongPress`, `handleShortPress`, `handleClosePosition`, `handleModifyPress` |
| `PerpsOpenOrderCard.tsx` | `handleCancelPress` |
| `AssetOverviewContent.tsx` | `handleLongPress`, `handleShortPress` |

**`useAccountGroupCompliance` removed**

This hook had zero usages in the app — only its own definition, test, and export. All blocking is action-gated via `gate()`. Removed to reduce surface area.

**`useWalletCompliance.ts` → `useComplianceGate.ts`, internal helper inlined**

The file is renamed to match its only public export. The internal `useWalletCompliance` helper was subsequently inlined directly into `useComplianceGate` (it was called exactly once, private, and the intermediate `useMemo` it returned was immediately destructured — eliminating it removes a redundant object allocation per render).

Additional simplifications made during this cleanup:
- **Always-batch**: removed the `isSingle` branch that routed single addresses to `checkWalletCompliance`. All checks now go through `checkWalletsCompliance`, which accepts both single-element and multi-element arrays identically.
- **`normalizeAddresses` removed**: addresses are derived from the stable `addressKey` scalar via `addressKey.split(',')`, eliminating the helper function and the `eslint-disable` comment on the dependency array.
- **`selectIsWalletBlocked` removed from hook**: replaced by the single `selectAreAnyWalletsBlocked(addresses)` selector call.

**Haptic feedback on block modal**

`notificationAsync(NotificationFeedbackType.Warning)` is now fired inside `showAccessRestrictedModal` in `AccessRestrictedContext.tsx`, consistent with the `expo-haptics` pattern used throughout Perps. The haptic fires automatically for every call site with no changes required in `useComplianceGate` or any action handler.

**`docs/compliance.md` updated**

- Promotes `gate()` as the recommended pattern (Option 1).
- Adds version-gating callout (`react-native-device-info`'s `getVersion()` reads the native binary version, not `package.json` — a stale build will fail the version check even if the flag is enabled in LaunchDarkly).
- Adds testing patterns for `gate()` mocking.
- Adds consumer flows file reference table.
- Removes the `useAccountGroupCompliance` and `useWalletCompliance` public-usage options.

## Changelog

- **Added**: `gate()` helper to `useComplianceGate` for single-call compliance + modal integration
- **Fixed**: Stale closure bug where `isBlocked` was read from Redux at render time instead of from the fresh API response
- **Fixed**: SonarQube violations — hooks were called after an early return, violating the Rules of Hooks
- **Added**: Haptic feedback (`NotificationFeedbackType.Warning`) when the access-restricted modal appears
- **Removed**: Unused `useAccountGroupCompliance` hook
- **Removed**: Internal `useWalletCompliance` helper (inlined into `useComplianceGate`)
- **Removed**: Single-address `checkWalletCompliance` path — always uses batch `checkWalletsCompliance`
- **Updated**: `useWalletCompliance.ts` renamed to `useComplianceGate.ts`
- **Updated**: Compliance documentation (`docs/compliance.md`)

CHANGELOG entry: Gate perps actions in compliance checks

## Related issues

Fixes: 

## Manual testing steps

**Feature**: OFAC compliance gate on Perps actions

1. Enable the `complianceEnabled` feature flag via the local override screen (Settings → Developer → Feature Flags).
2. Open the Perps home screen and tap **Add Funds** → modal should appear if wallet is blocked, action should proceed if not.
3. Open a market and tap **Long / Short / Close / Modify / Add Margin / Set Stop Loss** → same gate applies.
4. Open the Order Book and tap **Long / Short / Close / Modify** → same gate applies.
5. Open an open order card and tap **Cancel** → same gate applies.
6. Open Token Details for a perpetuals-eligible asset and tap **Long / Short** → same gate applies.
7. Disable `complianceEnabled` flag and repeat — all actions should proceed without any compliance check.

## Screenshots/Recordings

https://github.com/user-attachments/assets/066fa2b0-382f-4757-b300-cdcfce2ebd48

## Checklist

- [x] I've added unit tests for `gate()` covering: disabled flag fast-path, not-blocked pass-through, blocked modal display, error fail-open, and multi-address array handling.
- [x] I've manually tested on a clean native build (required for feature flag version check to resolve correctly).
- [x] I've updated `docs/compliance.md` with the new patterns and removed outdated content.
- [x] SonarQube Rules of Hooks violations resolved — no `eslint-disable` comments remain in the compliance hooks.

---

> [!NOTE]
> **Medium Risk**
> Touches compliance enforcement and perps trading entrypoints by introducing a new `gate()` flow that can block user actions and changing controller initialization/messaging to align with `@metamask/compliance-controller@2.0.0`; mistakes could cause unintended blocks or missed checks.
> 
> **Overview**
> Introduces a new `useComplianceGate` hook with a `gate(async action)` helper that *prefetches* `checkWalletsCompliance`, waits for any in-flight check to avoid tap-race issues, and blocks actions by showing `AccessRestrictedModal` (fail-open on API errors).
> 
> Updates Perps and TokenDetails flows to wrap all previously geo-eligibility-gated actions (trade, order book, position management, deposits, etc.) with `gate()`, and adjusts unit tests to mock/verify the new gating behavior.
> 
> Migrates compliance infrastructure to `@metamask/compliance-controller@2.0.0` by removing the controller `init()`/blocklist fetch path and its init messenger/actions, simplifying selectors/mocks/docs to rely solely on `walletComplianceStatusMap`, and adds warning haptics when the access-restricted modal is shown.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2b9cd79e8c0bdd18b32dc866ed1b01cb7e3bb55. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/components/UI/Compliance/hooks/useComplianceGate.ts
```diff
@@ -0,0 +1,126 @@
...
+          ? results.some((r) => r.blocked)
+          : false;
+      })
+      .catch(() => {
+        prefetchBlockedRef.current = false; // fail-open on error
+        DevLogger.log('[useComplianceGate] Prefetch compliance check failed');
+      });
+  }, [isComplianceEnabled, checkCompliance]);
+
+  const gate = useCallback(
```
michalconsensys, line 92: Given that this API creates significant diff, would it not be better to have something like

```tsx
const handleWithdraw = useCallback(async () => {

    await ensureCompliance(); // displays modal, and throws if user is blocked

   // withdraw logic

}, [...]);

```

eventually just a hook that would block the UI if the user is blocked, and run on render?

aganglada: We will need to add `try/catch` blocks to the functions and will likely be the same diff

app/components/UI/TokenDetails/components/AssetOverviewContent.tsx
```diff
@@ -290,38 +292,50 @@ const AssetOverviewContent: React.FC<AssetOverviewContentProps> = ({
     useState(false);
   const { track } = usePerpsEventTracking();
 
+  // Compliance gate
+  const selectedAddress = useSelector(selectSelectedInternalAccountAddress);
```
bergarces, line 296: I can see it's being used all over the place, but this selector is a bit out of date.

- It uses `AccountsController` state, which is deprecated.
- It assumes there's only one `accountId` present at any time (not necessarily the case since bip44) that somehow depends on the selected chain (another deprecated concept since we removed the global network selector).

Can we guarantee that the returned address here is the one from the selected asset? Because, if not, we should be using a selector that takes the chainId of the asset as an input and returns the address from the accountId of the selected account group that matches the chain.

I'm not going to block approval of the PR because of this, but we ought to start deprecating these selectors.

bergarces (approved):
Assets approval

michalconsensys (approved):
LGTM, what needs to be blocked is TBD and can be updated in other PR if needed

tommasini (approved):

PR #28085, opened 2026-03-30
feat: add segment events for ab trending sectinos

PatrykLucka:
### **Summary**
This PR refactors homepage trending-sections A/B analytics so exposure / business-event context and transaction attribution have clearer ownership and less accidental cross-flow leakage.

### **Changes**
**Auto-injected active_ab_tests via registry**
Removed manual active_ab_tests inlining from useHomeViewedEvent and useHomeSessionSummary. Registered HOMEPAGE_TRENDING_SECTIONS_AB_TEST_ANALYTICS_MAPPING in abTestAnalyticsRegistry so HOME_VIEWED is auto-enriched by enrichWithABTests (see docs/ab-testing.md).

**No useABTest inside useSwapBridgeNavigation**
The swap/bridge navigation hook no longer resolves the homepage experiment. That avoided Experiment Viewed firing from non-homepage surfaces (Token Details, navbar, etc.). Assignment stays in homepage-owned code: Homepage.tsx + HomepageTrendingAbTestContext.

**Explicit transactionActiveAbTests threading**
useSwapBridgeNavigation takes an optional transactionActiveAbTests and puts it on Bridge route params only when the caller passes a non-empty array. Homepage sections build the payload with useHomepageTrendingTransactionActiveAbTests (from context + abTestConfig). Token Details does not read this from Redux — it’s carried on TokenDetailsRouteParams / navigation params from entry points (e.g. TrendingTokenRowItem → asset screen) and forwarded through useTokenActions / usePerpsActions / Predict routes as needed.

**Transaction Added attribution (tx-scoped)**
When the user actually submits, attribution is bound to transactionMeta.id (and batch IDs where applicable): pending payload is set around submit via withPendingTransactionActiveAbTests (wraps the former stash/finally pattern), TransactionController is patched to register pending tests onto new transaction IDs, and getSwapTransactionActiveAbTestProperties reads from that registry — not from global bridge state for this experiment.

**Homepage trending tokens UI**
TrendingTokenRowItem accepts optional transactionActiveAbTests; TokensSection (trending-only) passes trendingTransactionActiveAbTests into rows so dedicated homepage trending taps carry the same attribution payload as Perps/Predict.

**LaunchDarkly-style flag key**
homeTMCU470AbtestTrendingSections follows {team}{TICKET}Abtest{Name} and is documented in abTestConfig.ts.

## **Changelog**

CHANGELOG entry: Added an A/B-tested homepage layout that separates trending sections and tracks swap analytics attribution for actions launched from homepage trending sections.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TMCU-471

## **Manual testing steps**

```gherkin
Feature: Homepage trending sections AB attribution

  Scenario: Control variant keeps default homepage composition
    Given a build with LaunchDarkly flag "homeTMCU470AbtestTrendingSections" set to variant "control"
    When the user opens Home
    Then the default section composition is rendered (no dedicated trending-only sections)
    And HOME_VIEWED events include active_ab_tests with key "homeTMCU470AbtestTrendingSections" and value "control"

  Scenario: TrendingSections variant renders dedicated trending sections
    Given a build with LaunchDarkly flag "homeTMCU470AbtestTrendingSections" set to variant "trendingSections"
    When the user opens Home
    Then dedicated Trending Tokens and Trending Predict sections are rendered
    And Trending Perps is rendered when Perps is enabled for the user
    And HOME_VIEWED section_viewed events use section_name trending_tokens, trending_perps, and trending_predict for those sections when they are enabled and visible
    And HOME_VIEWED events include active_ab_tests with key "homeTMCU470AbtestTrendingSections" and value "trendingSections"

  Scenario: Session summary includes AB attribution
    Given any active assignment for "homeTMCU470AbtestTrendingSections"
    When the user navigates away from Home
    Then a HOME_VIEWED session_summary event fires
    And active_ab_tests includes the homepage trending-sections flag (auto-injected via analytics mapping for HOME_VIEWED)

  Scenario: Swap AB attribution from homepage trending token details
    Given variant "trendingSections" is active for "homeTMCU470AbtestTrendingSections"
    And the user opens Token Details from the homepage Trending Tokens section (source homepage-trending / dedicated row)
    When the user taps Swap and completes a transaction that emits TRANSACTION_ADDED
    Then TRANSACTION_ADDED includes active_ab_tests with key "homeTMCU470AbtestTrendingSections"
    And the value matches the active variant name (e.g. "trendingSections")

  Scenario: Swap from non-homepage surface does not carry homepage attribution
    Given the user previously visited a homepage trending section while the experiment was active
    When the user opens Swap from the navbar or another entry that does not pass transactionActiveAbTests
    Then TRANSACTION_ADDED does not include active_ab_tests for "homeTMCU470AbtestTrendingSections"

  Scenario: No Experiment Viewed side effects from non-homepage callers
    Given any assignment for "homeTMCU470AbtestTrendingSections"
    When useSwapBridgeNavigation is invoked from Token Details or the navbar without homepage trending transactionActiveAbTests
    Then no Experiment Viewed event fires attributable solely to opening that swap flow for "homeTMCU470AbtestTrendingSections"
```

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes transaction analytics enrichment and threads new route params through swap/bridge, perps, and predict submission flows; mistakes could drop or misattribute `active_ab_tests` on `TRANSACTION_ADDED`. Scope is mostly analytics/attribution with limited impact on core transaction execution paths.
> 
> **Overview**
> Refactors homepage trending-sections A/B analytics to **separate exposure tracking from transaction attribution**: `HOME_VIEWED` now gets `active_ab_tests` via the central `abTestAnalyticsRegistry`, and homepage provides an explicit context for the experiment assignment.
> 
> Introduces a new `transaction-active-ab-test-attribution-registry` and a `swap-transaction-ab-tests` metrics builder so `TRANSACTION_ADDED` can attach `active_ab_tests` **per transaction ID** (registered synchronously before metrics builders run). Swap/bridge, perps, and predict flows now thread optional `transactionActiveAbTests` via navigation params and wrap submit/deposit/order actions with `withPendingTransactionActiveAbTests` to bind assignments at tx creation time.
> 
> Updates homepage trending sections to pass this attribution payload into Token Details/trending rows and into perps/predict navigations, and adds `enabled` gating to Predict homepage data hooks to avoid fetching when the feature flag is off. Includes targeted test updates covering param threading, attribution registration, and focus-effect stability.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e470bf2c64eb057be89cc922c46fbf169eb036ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sahar-fehri (dismissed):

pedronfigueiredo (dismissed):

app/components/Views/Homepage/hooks/useHomeViewedEvent.ts
```diff
@@ -100,6 +103,14 @@ const useHomeViewedEvent = ({
           is_empty: isEmpty,
           item_count: itemCount,
           entry_point: entryPoint,
+          ...(isActive && {
+            active_ab_tests: [
```
bfullam: We don't have to manually inline `active_ab_tests`, we can just have it auto-injected using the registry. See `/docs/ab-testing.md` for examples

app/components/Views/Homepage/hooks/useHomeSessionSummary.ts
```diff
@@ -61,9 +68,17 @@ const useHomeSessionSummary = ({
               interaction_type: 'session_summary',
               location: 'home',
               total_sections_viewed: getViewedSectionCount(),
-              total_sections_loaded: totalSectionsLoadedRef.current,
-              entry_point: entryPointRef.current,
+              total_sections_loaded: snap.totalSectionsLoaded,
+              entry_point: snap.entryPoint,
               session_time: sessionTime,
+              ...(snap.isActive && {
+                active_ab_tests: [
```
bfullam: Same as other comment, we can use the registry to avoid this manual inlining 🤝

app/components/UI/Bridge/hooks/useSwapBridgeNavigation/index.ts
```diff
@@ -144,6 +159,14 @@ export const useSwapBridgeNavigation = ({
   );
   const currentNetworkInfo = useCurrentNetworkInfo();
 
+  const {
+    variantName: homepageTrendingVariantName,
+    isActive: isHomepageTrendingAbActive,
+  } = useABTest(
```
bfullam: I think this hook is doing two different jobs right now: resolving the homepage experiment and carrying attribution into later flows. Since `useSwapBridgeNavigation` is used from non-homepage surfaces too, putting `useABTest` here creates exposure side effects from unrelated callers. I’d rather keep experiment resolution in homepage-owned code and pass explicit attribution forward from homepage entry points.

A cleaner model would be:

- homepage owns useABTest
- homepage-owned entry points pass `activeAbTests` forward as props
- Token Details / Perps / Predict just carry that context
- submit/transaction code reads that explicit context, rather than reading sticky global Redux state

That keeps exposure tied to the homepage, and keeps transaction attribution tied to the specific navigation path that produced the transaction.

sahar-fehri (dismissed):

bfullam (commented):
I like the direction here, especially the goal of attributing downstream actions back to homepage trending entry points. The one thing I’d push on is the current ownership boundary.

Right now the PR is mixing:
- experiment ownership/exposure
- downstream attribution

For homepage-owned events, using the registry pattern in the docs makes sense.

For transaction attribution, the registry isn't enough since we need to only apply the ab test data for specific transactions. I think the cleaner long-term shape is to pass explicit attribution context forward from homepage entry points, instead of resolving the homepage AB inside a generic hook and storing it in global bridge Redux state. That would avoid accidental homepage exposures from non-homepage callers and avoid stale transaction attribution leaking across flows.

If you want to keep this PR smaller, I’d at least avoid `useABTest` inside the generic swap-navigation hook and keep the experiment resolution in homepage-owned code.

Happy to sync on a minimal version for this PR vs a follow-up cleanup, because I think the intent is right, it’s mostly the placement/ownership that worries me.

pedronfigueiredo (dismissed):

bfullam (commented):
This looks much better!

Two remaining concerns:
- The LD flag name isn't following the naming convention from the docs: `{team}{TICKET}Abtest{Name}`. We should fix it here so the experiment ID stays unique and future homepage-trending tests don’t end up overlapping in analytics.
- The transaction attribution still lives in global Redux state.

The stale-state path I’m worried about with this Redux pattern is:
- user taps a dedicated homepage trending token / perps / predict entry, which sets `transactionActiveAbTests`
- user leaves without creating a transaction
- later, the user opens a regular token from the wallet token list and taps Swap
- `Transaction Added` can still pick up the old homepage A/B tag, because the metrics code still reads from Redux

The new `clearTransactionAbTests()` calls definitely reduce the risk, but correctness still depends on cleanup happening before the next transaction is created, which is a bit path-heavy.

I think the cleaner fix would be to make the attribution explicit to the flow that should own it, rather than storing it globally in Redux.

Concretely, I think that would look like:
- homepage-owned trending entry points pass `activeAbTests` explicitly through navigation / flow params
- Token Details / Perps / Predict carry that explicit attribution forward
- when a transaction or batch is actually created (`util/transaction-controller/index.ts`, `bridge-status-controller-init.ts`), bind that explicit attribution to the created `transactionMeta.id` / `batchId` in a small registry
- make `getSwapTransactionActiveAbTestProperties()` read that tx-specific registry instead of global Redux state

That would make the attribution belong to the transaction that was actually created, instead of whichever homepage flow last wrote to Redux.

bfullam (dismissed):
Great work, this is looking fantastic! Approved and much appreciated 🙏

pedronfigueiredo (dismissed):

app/components/UI/TokenDetails/constants/constants.ts
```diff
@@ -28,4 +31,6 @@ export enum TokenDetailsSource {
 export interface TokenDetailsRouteParams extends TokenI {
   source?: TokenDetailsSource;
   securityData?: TokenSecurityData;
+  /** Carried into swap / perps / predict flows for tx-scoped `active_ab_tests` */
+  transactionActiveAbTests?: TransactionActiveAbTestEntry[];
```
Prithpal-Sooriya, line 35: Hmm, do we need to inject and carry over AB tests across screens?
I thought the new config approach makes it so that you do not need to manage you're AB test across the app.

For example here is our recent assets AB test, we are injecting the AB test properties across multiple events we are interested in.
https://github.com/MetaMask/metamask-mobile/blob/9c635c680a136824a40fd29ccf43bbc493a738ab/app/components/UI/TokenDetails/components/abTestConfig.ts#L27-L35

app/components/UI/Predict/views/PredictMarketDetails/PredictMarketDetails.tsx
```diff
@@ -210,6 +211,7 @@ const PredictMarketDetails: React.FC<PredictMarketDetailsProps> = () => {
           outcomeToken: token,
           entryPoint:
             entryPoint || PredictEventValues.ENTRY_POINT.PREDICT_MARKET_DETAILS,
+          ...(transactionActiveAbTests?.length && { transactionActiveAbTests }),
```
matallui, line 214: Not sure I understand. Why are we manually injecting these directly into each feature?

I thought the new AB testing framework would allow us to define active AB tests and what events they need to be automatically injected into, so I'm not sure I follow what this is about.

Prithpal-Sooriya (dismissed):

app/components/UI/Predict/views/PredictMarketDetails/PredictMarketDetails.tsx
PatrykLucka: Yea, I see there is confusion around it. The main issue is that we want to have it mapped 1-1 for each transaction, so we can't really rely on global state, as in some edge cases something might change between starting the flow and the point where we trigger addTransaction.

There is a comment above from Bryan where he correctly points that out: https://github.com/MetaMask/metamask-mobile/pull/28085#pullrequestreview-4082092194

Prithpal-Sooriya (approved):

app/components/UI/TokenDetails/hooks/usePerpsActions.ts
```diff
@@ -60,10 +64,11 @@ export const usePerpsActions = ({
           direction,
           asset: marketData.symbol,
           fromTokenDetails,
+          ...(transactionActiveAbTests?.length && { transactionActiveAbTests }),
```
michalconsensys, line 67: If the `transactionActiveAbTests` is an empty array, then it will spread the number `0`. Can we change it to something like this? `...((transactionActiveAbTests?.length ?? 0 > 1) && { transactionActiveAbTests }),`

app/components/Views/Homepage/Sections/Perpetuals/PerpsSection.tsx
```diff
@@ -103,9 +113,17 @@ const usePerpsNavigationHandlers = () => {
       navigateToTutorialOrScreen(Routes.PERPS.MARKET_DETAILS, {
         market,
         source: PERPS_EVENT_VALUE.SOURCE.HOME_SECTION,
+        ...(isDedicatedTrendingSection &&
+          trendingTransactionActiveAbTests?.length && {
```
michalconsensys, line 117: Same as in the comment above

bfullam (approved):

michalconsensys (approved):

pedronfigueiredo (approved):

matallui (approved):

PR #28157, opened 2026-03-31
feat: add SDKConnectV2 connection cap

ffmcgee725:
## **Description**

SDKConnectV2 had no upper bound on the number of active connections. Every deeplink created and persisted a new connection unconditionally, and expiry was only enforced at deserialization time (`ConnectionStore.list()`). During the 30-day `DEFAULT_SESSION_TTL` window, connections accumulated freely — amplifying startup reconnection bursts, chain-change fan-out, and stale `BackgroundBridge` listeners.

This PR introduces a `MAX_CONNECTIONS` cap (20) with deterministic oldest-first eviction:

- **`evictIfAtCapacity()`** — called in `handleConnectDeeplink()` before `Connection.create()`. When at capacity, the connection with the earliest `expiresAt` (oldest) is disconnected, deleted from the store, and has its permissions revoked.
- **`trimToCapacity()`** — called in `initialize()` after all persisted connections are resumed on cold start. Evicts excess connections that may have been stored before this cap was introduced.

Both methods use `expiresAt` as a proxy for creation time, which is reliable because all connections share the same TTL.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/WAPI-1369

## **Manual testing steps**

```gherkin
Feature: SDKConnectV2 connection cap enforcement

  Scenario: new connection evicts oldest when at capacity
    Given the wallet has 20 active SDK connections
    When a new dApp initiates a connection via deeplink
    Then the oldest connection is disconnected and removed
    And the new connection is established successfully

  Scenario: new connection succeeds when below capacity
    Given the wallet has fewer than 20 active SDK connections
    When a new dApp initiates a connection via deeplink
    Then the new connection is established without evicting any existing connection

  Scenario: excess persisted connections are trimmed on cold start
    Given the wallet has more than 20 persisted SDK connections in storage
    When the app performs a cold start
    Then only the 20 most recent connections are resumed
    And the oldest excess connections are disconnected and cleaned up
```

### **Before**

N/A

### **After**

N/A

---

> [!NOTE]
> **Medium Risk**
> Adds automatic eviction/trimming of SDKConnectV2 connections, which can disconnect existing sessions and revoke permissions if the cap is hit or on cold-start cleanup. Logic is contained to connection lifecycle management but affects persisted state and startup behavior.
> 
> **Overview**
> Introduces a hard cap (`MAX_CONNECTIONS = 20`) for SDKConnectV2 connections, evicting the *oldest* connection (by earliest `expiresAt`) before establishing a new deeplink connection.
> 
> On cold start, initialization now trims any previously-persisted excess connections down to the cap, continuing best-effort even if some disconnects fail. Tests are expanded to cover eviction on new connects, startup trimming behavior, and failure tolerance during trimming.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dde766cf61ab08c11976a3d290e1bdeed532b6f9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -187,8 +225,9 @@ export class ConnectionRegistry {
           message: 'External transactions cannot use internal origins',
         });
       }
-      connInfo = this.toConnectionInfo(connReq);
+      await this.evictIfAtCapacity();
```
jiexi, line 228: `trimToCapacity()` to can be used instead of `evictIfAtCapacity` since they pretty much do the same thing.

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -75,9 +88,34 @@ export class ConnectionRegistry {
...
+    await this.trimToCapacity();
+
     this.hostapp.syncConnectionList(Array.from(this.connections.values()));
   }
 
+  /**
+   * Trims connections down to {@link MAX_CONNECTIONS} after a cold-start
+   * resume. Oldest connections (earliest `expiresAt`) are evicted first.
+   */
+  private async trimToCapacity(): Promise<void> {
```
jiexi, line 100: Thoughts on changing the name of this method to something that describes what we are trimming? On the WCv2 side, we have an equivalent method called `enforceSessionLimit()`. We don't have to use that name here, but I feel this is a bit too vague as-is

ffmcgee725: It's local to the `connection-registry` file, and documentation states it's trimming connections, so I don't think it's all that vague that it's handling connections

app/core/SDKConnectV2/services/connection-registry.ts
ffmcgee725: Logic is actually slightly different, `evictIfAtCapacity` only removes one while `trimToCapacity` removes multiple until it respects the limit of connections established.

We could unify them by calling `trimToCapacity` in both places —but we'd pay for the `sort` on every deeplink when a single `reduce` suffice. Knowing this, would you still suggest using `trimToCapacity()` in both places ? I'm okay with whichever choice we make.

app/core/SDKConnectV2/services/connection-registry.ts
jiexi: that's a fair point. we are in the ConnectionRegistry class.

app/core/SDKConnectV2/services/connection-registry.ts
jiexi: fair. O(n) vs O(n^2) at worst case. Perhaps we can remove `trimToCapacity()` on startup then if we just assume that MMC is still fairly new that pretty much nobody is at >20 connections already

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -210,6 +249,28 @@ export class ConnectionRegistry {
...
+   * Because every connection is created with the same TTL, the smallest
+   * `expiresAt` reliably identifies the connection that was established
+   * first.
+   */
+  private async evictIfAtCapacity(): Promise<void> {
+    if (this.connections.size < MAX_CONNECTIONS) return;
+
+    const oldest = Array.from(this.connections.values()).reduce((a, b) =>
+      a.info.expiresAt <= b.info.expiresAt ? a : b,
+    );
```
adonesky1, line 265: I've never used reduce like this 🤔 🆒  as in to return one value rather than an obj/array... genuinely didn't know you could do that

app/core/SDKConnectV2/services/connection-registry.ts
```diff
@@ -210,6 +249,28 @@ export class ConnectionRegistry {
...
+
+    logger.debug(
+      'Connection cap reached, evicting oldest connection:',
+      oldest.id,
+    );
+    await this.disconnect(oldest.id);
```
adonesky1:
```suggestion
try {
    logger.debug('Connection cap reached, evicting oldest connection:', oldest.id);
    await this.disconnect(oldest.id);
  } catch (error) {
    logger.error('Failed to evict oldest connection:', oldest.id, error);
  }
```

app/core/SDKConnectV2/services/connection-registry.ts
adonesky1, line 288: Just a note that in trimming cases this will get spammed a bit unnecessarily

app/core/SDKConnectV2/services/connection-registry.ts
adonesky1: I think we can keep the `trimToCapacity()` on startup. Its defensive and costs nothing when there are no prior connections.

app/core/SDKConnectV2/services/connection-registry.ts
ffmcgee725: https://github.com/MetaMask/metamask-mobile/pull/28157/commits/7bdda2c3a1425919e1835c4e5c9ac172d11ae432

adonesky1 (approved):
LGTM!

PR #27864, opened 2026-03-24
feat: implement wallet-side analytics for SDKConnectV2 / MWP

ffmcgee725:
## **Description**

- Fire `DEEP_LINK_USED` (MetaMetrics) for every MWP deeplink received, with `route: 'mmc-mwp'` and Branch.io app-installation detection
- Track wallet-side MWP connection lifecycle events (`Remote Connection Request Received`, `Remote Connection Request Failed`) through the standard MetaMetrics / AnalyticsController pipeline
- Track reconnect attempts (`handleSimpleDeeplink`) with `found_in_store` to surface silent store-miss failures

## Motivation

Dapp-side events (`mmconnect_connection_initiated` with `transport_type: 'mwp'`) tell us how many dapps _tried_ to connect via MWP, but not how many wallets actually received them. The drop-off between dapp-initiated and wallet-received (silent failures, iOS routing issues) is completely invisible without wallet-side instrumentation. MWP adoption rate — the key signal for deprecating SDKv1 — can't be tracked from the wallet's perspective.

### Part A — `DEEP_LINK_USED` for MWP deeplinks

| File | Change |
|------|--------|
| `app/core/DeeplinkManager/types/deepLinkAnalytics.types.ts` | Add `MMC_MWP = 'mmc_mwp'` to `DeepLinkRoute` enum |
| `app/core/DeeplinkManager/util/deeplinks/deepLinkAnalytics.ts` | Register `MMC_MWP` in the route-extractor map |
| `app/core/DeeplinkManager/handlers/legacy/handleDeeplink.ts` | Fire `DEEP_LINK_USED` asynchronously before `SDKConnectV2.handleMwpDeeplink()` — never blocks the WebSocket handshake |

### Part B — Wallet-side connection events via MetaMetrics

All wallet-side MWP events are routed through the standard `analytics.trackEvent()` / `AnalyticsController` pipeline. This means:
- Events appear in the mobile app's Segment source under the user's `analyticsId`
- Opt-in / opt-out is enforced by the controller — no manual consent check needed
- "Delete MetaMetrics Data" covers MWP events (same source + same identity)
- Events automatically include `metamask-mobile-globals` default properties

The per-session `remote_session_id` (previously `anon_id`) is preserved as an event property so dapp↔wallet join analysis continues to work.

| File | Change |
|------|--------|
[...]

PR #28290, opened 2026-04-01
fix: detect SDKConnectV2 connections in useOriginSource via v2Connections store

adonesky1:
## Summary

- **Fixes a bug** where `useOriginSource` misclassified all SDKConnectV2 (MWP) connections as `in-app browser`, causing the `source` property on `CONNECT_REQUEST_STARTED`, `CONNECT_REQUEST_COMPLETED`, and `CONNECT_REQUEST_CANCELLED` analytics events to be incorrect for V2 connections.
- The hook's existing check for the `MMSDKCONNECTV2::` prefix was dead code — the prefix is stripped by `RPCMethodMiddleware` before it reaches the permission system. V2 connections arrive with a bare UUID as their origin.
- Adds a lookup against the `v2Connections` Redux store (already populated by `HostApplicationAdapter.syncConnectionList`) to correctly identify V2/MWP connections and return `SourceType.SDK_CONNECT_V2`.

## Root cause

The `RPCBridgeAdapter` constructs `middlewareHostname = MMSDKCONNECTV2::<connId>`, but:
1. `RPCMethodMiddleware.ts` immediately strips the prefix (`hostname.replace(SDK_CONNECT_V2_ORIGIN, '')`)
2. For the CAIP path (`wallet_createSession`), `BackgroundBridge.channelIdOrOrigin` returns the bare `channelId` (UUID) when `isMMSDK=true`
3. The `PermissionController` approval request therefore carries a bare UUID as `metadata.origin`
4. `useOriginSource` couldn't match this UUID to V2 — it only checked the V1 `SDKConnect` store

CHANGELOG entry: `null`

## Test plan

- [x] Unit tests: 7 passing (2 new V2-specific tests added)
- [ ] Manual: Open an MWP deeplink, verify `CONNECT_REQUEST_STARTED` event has `source: 'sdk_connect_v2'` instead of `source: 'in-app browser'`
- [ ] Verify no regression for V1 SDK connections (`source: 'sdk'`)
- [ ] Verify no regression for WalletConnect connections (`source: 'walletconnect'`)
- [ ] Verify no regression for in-app browser connections (`source: 'in-app browser'`)

## Related

- Fixes: [WAPI-1380](https://consensyssoftware.atlassian.net/browse/WAPI-1380)
- Related PR: #27864 (wallet-side analytics for SDKConnectV2/MWP — this fix reduces the scope of new events needed there)

---

> [!NOTE]
> **Medium Risk**
> Changes the source-classification logic used for connection-request analytics by introducing a new SDK v2 detection path and reordering precedence, which could shift event attribution if the v2Connections store is incomplete or stale.
> 
> **Overview**
> Fixes `useOriginSource` to correctly detect **SDK Connect v2 (MWP)** permission origins by treating a *bare UUID* as v2 when it exists in the Redux `sdk.v2Connections` map (instead of relying on a removed/stripped v2 prefix).
> 
> Refactors the hook to use an explicit priority order (**v2 → v1 → WalletConnect → in-app browser**) and adds unit coverage for v2 detection (including the case where the same UUID exists in both v1 and v2, with v2 winning). Tests that build mock Redux `sdk` state were updated to include an empty `v2Connections` field.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5a7473ef9506ce229a290df0fbbd925d7c0f270b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/components/hooks/useOriginSource.test.ts
```diff
@@ -67,6 +69,58 @@ describe('useOriginSource', () => {
...
+
+    mockSelector.mockImplementation((selector: (state: RootState) => unknown) =>
+      selector(v2State),
+    );
+
+    const { result } = renderHook(() => useOriginSource({ origin: v2Uuid }));
+    expect(result.current).toBe(SourceType.SDK_CONNECT_V2);
+  });
+
+  it('should prefer SDK_CONNECT_V2 over SDK when UUID exists in both stores', () => {
```
jiexi, line 92: "if the almost impossible happens..." 😂 . Doesn't hurt to keep it though i guess

app/components/hooks/useOriginSource.ts
```diff
@@ -11,39 +11,59 @@ interface UseOriginSourceProps {
...
     return undefined;
   }
 
-  // Check for V2 connections
-  if (origin.startsWith(AppConstants.MM_SDK.SDK_CONNECT_V2_ORIGIN)) {
+  // --- SDK v2 (MWP) ---
+  // V2 connections use the bare session UUID as the permission-system origin.
+  // Look it up in the v2Connections store (populated by
+  // HostApplicationAdapter.syncConnectionList, keyed by connection ID).
+  if (isUUID(origin) && v2Connections?.[origin]) {
```
jiexi, line 40: i mean, if it isn't UUID then it shouldn't match anything in v2Connections, right? I guess this doesn't hurt to have the extra check though

app/components/hooks/useOriginSource.ts
```diff
@@ -11,39 +11,59 @@ interface UseOriginSourceProps {
...
 
-  // Check if origin matches WalletConnect metadata
-  const isWalletConnect = wc2Metadata?.id && wc2Metadata.id.length > 0;
-  if (isWalletConnect) {
+  // --- WalletConnect ---
+  // wc2Metadata is a single Redux slot holding the *most recent* WC proposal
+  // metadata (set on session_proposal, cleared after approval/rejection).
+  // It is not keyed by origin — we rely on the WC proposal flow being
+  // serialized (via proposalLock in WalletConnectV2) so that during the
+  // approval window, a non-empty id implies *this* origin is from WC.
```
jiexi, line 65: 😭 😭 😭

adonesky1: Created a ticket to improve this here: https://consensyssoftware.atlassian.net/browse/WAPI-1383

jiexi (approved):

PR #28542, opened 2026-04-08
fix: use event subscription for timeframe changing resolution for AC

sahar-fehri:
## **Description**

When switching between time ranges (e.g. 1M and 1Y) on the advanced chart, the charts appeared visually identical despite the API returning different data for each period. Two root causes were identified and fixed:

1. **Stale-data race condition**: When switching timeframes, the `useEffect` syncing OHLCV data to the WebView would immediately send the *previous* timeframe's data (still in React state) before the fresh fetch completed. If both timeframes happened to return similar data point counts, the chart would never receive the correct data at all.

2. **Missing visible range constraint**: TradingView auto-fits to *all* available data by default. Since the API returns similar amounts of daily-resolution data for 1M and 1Y, both timeframes displayed the full dataset — making them visually indistinguishable.

**Solution:**
- Defer sending stale data during timeframe switches by detecting when `ohlcvSeriesKey` changes but data hasn't refreshed yet. A skeleton is shown until fresh data arrives.
- Compute the expected viewport start timestamp (`visibleFromMs = lastBar.time - durationMs`) and pass it through to the WebView, where `chart.setVisibleRange()` constrains the initial viewport to only the selected period. Uses TradingView's `onDataLoaded()` event subscription to apply the range at the correct time.

## **Changelog**

CHANGELOG entry: Fixed advanced chart timeframe switching (1H/1D/1W/1M/1Y) showing identical charts for different time periods

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/ASSETS-3040

## **Manual testing steps**

```gherkin
Feature: my feature name

```

---

> [!NOTE]
> **Medium Risk**
> Touches RN↔WebView chart syncing and TradingView viewport logic; regressions could cause missing/incorrect chart data or loading skeletons during range switches.
> 
> **Overview**
> Fixes AdvancedChart timeframe switching issues where different ranges could appear identical.
> 
> On RN, `AdvancedChart` now *avoids pushing stale OHLCV* to the WebView when `ohlcvSeriesKey` changes by treating the first render after a key switch as a loading/skeleton state until fresh bars arrive.
> 
> Adds a `visibleFromMs` signal from the token price view (computed from the selected range duration) through `SET_OHLCV_DATA`, and updates the WebView `chartLogic` to apply `chart.setVisibleRange()` via a `onDataLoaded()` subscription so the initial viewport matches the selected timeframe while still allowing left-scroll/pagination. Tests were updated to assert the stale-data suppression behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f03b6d83bef824d5d5e923c5c4280b972d656636. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
+++ b/app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
@@ -91,6 +91,7 @@ const AdvancedChart = forwardRef<AdvancedChartRef, AdvancedChartProps>(
       onChartTradingViewClicked,
       isLoading = false,
       lineChrome,
+      visibleFromMs,
     },
     ref,
   ) => {
@@ -181,11 +182,18 @@ const AdvancedChart = forwardRef<AdvancedChartRef, AdvancedChartProps>(
     );
     paginationRef.current = ohlcvPagination;
 
+    const visibleFromMsRef = useRef<number | undefined>(visibleFromMs);
+    visibleFromMsRef.current = visibleFromMs;
+
     const sendOHLCVData = useCallback(
       (data: OHLCVBar[]) => {
         postMessage({
           type: 'SET_OHLCV_DATA',
-          payload: { data, pagination: paginationRef.current },
+          payload: {
+            data,
+            pagination: paginationRef.current,
+            visibleFromMs: visibleFromMsRef.current,
+          },
         });
       },
       [postMessage],
@@ -390,6 +398,17 @@ const AdvancedChart = forwardRef<AdvancedChartRef, AdvancedChartProps>(
         ohlcvSeriesKey !== undefined &&
         ohlcvSeriesKey !== prevOhlcvSeriesKeyRef.current
       ) {
+        if (prevOhlcvSeriesKeyRef.current !== undefined) {
+          // Time range switch: ohlcvData is still stale from the previous
+          // period (fetch is in progress). Show skeleton, mark the key, and
+          // clear prevData so the fresh data triggers the length-diff branch
+          // on arrival — avoiding sending stale data to the WebView which
+          // causes a resolution race condition in TradingView.
+          beginFullOhlcvLayoutSettle();
+          prevOhlcvSeriesKeyRef.current = ohlcvSeriesKey;
+          prevOhlcvDataRef.current = [];
+          return;
+        }
         beginFullOhlcvLayoutSettle();
         sendOHLCVData(ohlcvData);
         prevOhlcvDataRef.current = ohlcvData;
--- a/app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
+++ b/app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
@@ -428,9 +428,14 @@ function handleSetOHLCVData(payload) {
 
   function scheduleVisibleRangeAfterDataLoad(chart) {
     if (visibleFromMs == null) return;
+    var capturedGeneration = window.ohlcvGeneration;
     var sub = chart.onDataLoaded();
     sub.subscribe(null, function onLoaded() {
       sub.unsubscribe(null, onLoaded);
+      // Discard stale callback if user switched timeframes before load completed
+      if (capturedGeneration !== window.ohlcvGeneration) {
+        return;
+      }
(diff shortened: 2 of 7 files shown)


app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -38,6 +38,8 @@ window.ohlcvPagination = {
 };
 /** Bumped on each `SET_OHLCV_DATA` so in-flight fetches from a previous series are discarded. */
 window.ohlcvGeneration = 0;
+/** Visible-range start (ms) from `SET_OHLCV_DATA`. Applied once after the first `getBars` load. */
+window.pendingVisibleFromMs = null;
```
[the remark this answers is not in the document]
sahar-fehri: cleaned up here [a6ed8d4](https://github.com/MetaMask/metamask-mobile/pull/28542/commits/a6ed8d443aa7f073688f59c4d43d799c2daef52a)

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
```diff
@@ -390,6 +398,17 @@ const AdvancedChart = forwardRef<AdvancedChartRef, AdvancedChartProps>(
         ohlcvSeriesKey !== undefined &&
         ohlcvSeriesKey !== prevOhlcvSeriesKeyRef.current
       ) {
+        if (prevOhlcvSeriesKeyRef.current !== undefined) {
```
geositta, line 401: The new ohlcvSeriesKey switch path changes the core behavior to "show skeleton and wait for fresh data" instead of immediately sending SET_OHLCV_DATA, but the branch does not add replacement coverage for that flow.

I ran the focused AdvancedChart test for series key changes and it fails on this branch, so I would update the test to match the intended behavior and add a regression case for "stale data is not sent, fresh data is."

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
```diff
@@ -421,8 +421,32 @@ function handleSetOHLCVData(payload) {
     };
   }
 
+  var visibleFromMs =
+    payload.visibleFromMs != null ? payload.visibleFromMs : null;
+
   var newResolution = detectResolution(window.ohlcvData);
 
+  function scheduleVisibleRangeAfterDataLoad(chart) {
```
geositta, line 429: This adds a new chart.onDataLoaded() subscription on every full reload, but it does not cancel or guard the previous one if the user switches timeframes again before the load finishes.

That means a quick sequence like 1D -> 1W -> 1M can leave the older callback alive, and when the later load completes the older callback can still call setVisibleRange(...) with the wrong visibleFromMs.

I would either clear the previous subscription when a new range load starts, or gate the callback on the current OHLCV generation before applying the viewport.

geositta (changes requested):
Requesting changes. The overall direction makes sense and the stale data guard looks like the right fix for the timeframe switch race, but there are two follow-ups I have requested.

app/components/UI/Charts/AdvancedChart/AdvancedChart.tsx
sahar-fehri: [c88dc67](https://github.com/MetaMask/metamask-mobile/pull/28542/commits/c88dc67166301d88120cc444a746356baf799d6b)

app/components/UI/Charts/AdvancedChart/webview/chartLogic.js
sahar-fehri: [f03b6d8](https://github.com/MetaMask/metamask-mobile/pull/28542/commits/f03b6d83bef824d5d5e923c5c4280b972d656636)

geositta (approved):
Thank you for adding the focused regression test and the guard on the viewport callback. With this, the older timeframe callbacks are discarded if the user switches again before load completes.

Prithpal-Sooriya (approved):

juanmigdr (commented):
Tested locally, works well

juanmigdr (approved):

PR #28545, opened 2026-04-08
fix(perps): extract payment token init into useInitPerpsPaymentToken hook and move it to the parent cp-7.73.0

### LINKED ISSUE #28539: [Bug]: Order size slider broken after user change payment token
### Describe the bug

After user change payment token, the order size slider is unresponsive.
If user change order size with keyboard, it switch back payment token to 'perps balance'

### Expected behavior

Slider works as expected. Payment token isn't changed with order size changes

### Screenshots/Recordings

_No response_

### Steps to reproduce

1. Go to perp market (ex: BTC)
2. Tap long
3. Select payment token other than perp balance 
4. Try to move the slider -- notice it doesn't move
5. Input a different order size manually -- notice it switch back to perp balance as payment method

```shell

```

### Where was this bug found?

Live version (from official store)

### Version

7.73.0

### Build number

4353

### Build type

None

### Device

iphone 11

### Operating system

iOS

### Additional context

_No response_

### Severity

_No response_

michalconsensys:
## **Description**

The `PerpsPayRow` component contained ~110 lines of payment token initialization logic (pending config syncing, default token selection, asset-change resets) mixed in with its rendering concerns. This made the component harder to follow and impossible to unit-test the initialization behavior in isolation.

This PR extracts that logic into a dedicated `useInitPerpsPaymentToken` hook and calls it from `PerpsOrderViewContentBase` instead. The hook encapsulates:
- Syncing the pay token from `pendingTradeConfiguration` when resuming an order
- Falling back to the highest-balance allowlist token when the user has no perps balance
- Resetting state when the traded asset changes
- Cleaning up the controller's selected payment token on unmount

A `key` prop keyed on `payToken.symbol` was also added to the `PerpsSlider` so it recomputes width when the payment token changes, keeping the slider range in sync, allowing the user to use the slider

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://github.com/MetaMask/metamask-mobile/issues/28539

## **Manual testing steps**

```gherkin
Feature: Perps payment token initialization

  Scenario: user opens a perps order with some perps balance and other tokens
    Given the user selects USDC on Arbitrum
    When user navigates changes the amount to a different amount
    Then the USDC on Arbitrum should be still selected
    And user should be able to use slider if they have enough balance to cover the trade
```

### **Before**

N/A

### **After**

N/A

---

> [!NOTE]
> **Medium Risk**
> Touches Perps payment-token selection/init logic and its interaction with `PerpsController`/confirmation pay-token state, which could affect default token selection and persisted pending configs if incorrect.
> 
> **Overview**
> Moves Perps “pay with” token initialization/sync logic out of `PerpsPayRow` into a new `useInitPerpsPaymentToken` hook, and runs it from `PerpsOrderView` so pending trade config, default-token fallback, asset-change resets, and unmount cleanup are handled at the screen level.
> 
> Simplifies `PerpsPayRow` to be a pure display/navigation row (dropping the `initialAsset` prop) and updates tests accordingly, while adding focused unit tests for `useInitPerpsPaymentToken`.
> 
> Adds a `key` to `PerpsSlider` based on `payToken.symbol` to force a re-mount when the payment token changes, keeping the slider’s range/width in sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3a7d10624018350f91a6322ce77eb808e87e508. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
+++ b/app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
@@ -149,6 +149,8 @@ import { useUpdateTokenAmount } from '../../../../Views/confirmations/hooks/tran
 import { useConfirmActions } from '../../../../Views/confirmations/hooks/useConfirmActions';
 import { useInsufficientPayTokenBalanceAlert } from '../../../../Views/confirmations/hooks/alerts/useInsufficientPayTokenBalanceAlert';
 import { useNoPayTokenQuotesAlert } from '../../../../Views/confirmations/hooks/alerts/useNoPayTokenQuotesAlert';
+import { useDefaultPayWithTokenWhenNoPerpsBalance } from '../../hooks/useDefaultPayWithTokenWhenNoPerpsBalance';
+import { useInitPerpsPaymentToken } from './useInitPerpsPaymentToken';
 
 // Navigation params interface
 interface OrderRouteParams {
@@ -1229,6 +1231,8 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
     setSelectedTooltip(null);
   }, []);
 
+  useInitPerpsPaymentToken(orderForm.asset ?? '');
+
   // Use the same calculation as handleMaxAmount in usePerpsOrderForm to avoid insufficient funds error
   const amountTimesLeverage = Math.floor(availableBalance * orderForm.leverage);
 
@@ -1330,6 +1334,7 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
                 const amount = Math.floor(value).toString();
                 setAmount(amount);
               }}
+              key={payToken?.symbol ?? ''}
               minimumValue={0}
               maximumValue={maxPossibleAmount}
               step={1}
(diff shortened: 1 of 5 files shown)


gambinish (dismissed):

app/components/UI/Perps/Views/PerpsOrderView/PerpsOrderView.tsx
```diff
@@ -1229,6 +1230,8 @@ const PerpsOrderViewContentBase: React.FC<PerpsOrderViewContentProps> = ({
     setSelectedTooltip(null);
   }, []);
 
+  useInitPerpsPaymentToken(orderForm.asset ?? '');
```
geositta, line 1233: I am concerned that this changes the timing: initialization can now run before a transaction exists. Since this hook is called unconditionally from the parent view. On main, the same initialization path only existed inside PerpsPayRow, which is still gated.

Once that happens, both initialization effects in useInitPerpsPaymentToken.ts and useInitPerpsPaymentToken can call setPayToken() before useTransactionMetadataRequest() has produced an id.

In that case useTransactionPayToken() falls back to transactionId: '' in useTransactionPayToken.ts:26, so updatePaymentToken() is invoked with an empty transaction id in useTransactionPayToken.

After that, appliedPendingTokenRef is already set in the hook, so the initialization path does not retry when the real transaction arrives.

I think we should keep the main behavior here and guard initialization on transaction readiness. The simplest fix is to pass activeTransactionMeta or transactionId into useInitPerpsPaymentToken() and return early until it exists.

geositta (changes requested):
Requesting changes on the extracted payment token initialization.

The refactor itself is a good cleanup, but moving the initialization hook to the parent changed when it runs. That now allows the setup path to execute before the transaction exists, which can send `setPayToken()` through the empty id path and then suppress the retry once the real transaction is available.

I think this is worth fixing before merge by restoring the original transaction readiness constraint from main, either by gating useInitPerpsPaymentToken() on activeTransactionMeta or by passing a valid transactionId into the hook and returning early until it exists.

geositta (approved):
Thanks for helping me understand the flow. Looks good.

PR #28722, opened 2026-04-13
fix: tempo disable confirm button until gas autoselect

maxime-oe:
## **Description**

The goal of the change is to prevent a premature "enablement" for the Confirm button when a transaction is loading. There are pre-existing race conditions that allowed users to confirm a transaction before everything is fully loaded. Those conditions could make the tx skip gasless flows, falling back to native token for gas payment.

The impact is bigger on Tempo because using a native token is not an option (there is no native token on Tempo), which means that such race condition will always result in a failed tx (the tx won't be sent to the RPC or relayer). 

The changes solve the race conditions cases, as well as adds some extra conditions for the loading when on a chain that has no native token.

## **Changelog**

CHANGELOG entry: add new `isGaslessLoading` guard to Confirmation button - consistently with Extension.
CHANGELOG entry: `useIsGaslessSupported` to use `useQuery` instead of `useAsync` to avoid race conditions.

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/NEB-944?atlOrigin=eyJpIjoiYWFiY2U0NjU1ZjEzNGNmODlmOTZkNTQwNzM4N2MxMTUiLCJwIjoiaiJ9

## **Manual testing steps**

**Repeating instructions shared for Extension since Mobile flow will be very similar (just "tap" instead of click ;))**

- On Tempo mainnet, find `pathUSD` and click on "Send".
- Set amount to 0.01 and select the account to send to (any account).
- As soon as you clicked on the recipient account, start "spam clicking" on "Confirm" or the location where you know "Confirm" will appear.
- 
image

image

What should be observed is that despite spam-clicking, it should not be possible to "Confirm" the transaction "too soon" and it should always succeed no matter how "early" you click on "Confirm".

---

> [!NOTE]
> **Medium Risk**
> Changes confirmation disablement logic around gasless support and switches gasless eligibility checks to React Query; could affect when users can confirm transactions and may introduce edge-case loading states on chains without native gas.
> 
> **Overview**
> Prevents premature transaction confirmation by **disabling the Confirm button while gasless support/fee-token auto-selection is still loading** (`Footer` now gates on new `useIsGaslessLoading` in addition to transaction pay loading).
> 
> Adds `useIsGaslessLoading` to detect gasless-related loading (including Tempo-style *no native token* cases and mismatched `selectedGasFeeToken`), and refactors `useIsGaslessSupported` to use React Query (`useQuery`) for relay support checks to reduce race conditions.
> 
> Test infrastructure is updated to wrap `renderWithProvider`/`renderHookWithProvider` with a `QueryClientProvider`, and relevant tests are adjusted/mocked accordingly (including new unit tests for `useIsGaslessLoading`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 164017bd9d68db64fcc375b2d9e4f0f4608ecc4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/components/UI/Predict/components/PredictPositionsHeader/PredictPositionsHeader.test.tsx
+++ b/app/components/UI/Predict/components/PredictPositionsHeader/PredictPositionsHeader.test.tsx
@@ -71,6 +71,7 @@ jest.mock('../../hooks/usePredictBalance', () => ({
 
 const mockInvalidateQueries = jest.fn();
 jest.mock('@tanstack/react-query', () => ({
+  ...jest.requireActual('@tanstack/react-query'),
   useQueryClient: () => ({
     invalidateQueries: mockInvalidateQueries,
   }),
--- a/app/components/Views/confirmations/components/footer/footer.tsx
+++ b/app/components/Views/confirmations/components/footer/footer.tsx
@@ -39,6 +39,7 @@ import { PredictClaimFooter } from '../predict-confirmations/predict-claim-foote
 import { useIsTransactionPayLoading } from '../../hooks/pay/useTransactionPayData';
 import { Skeleton } from '../../../../../component-library/components-temp/Skeleton';
 import { useQRHardwareContext } from '../../context/qr-hardware-context';
+import { useIsGaslessLoading } from '../../hooks/gas/useIsGaslessLoading';
 
 const HIDE_FOOTER_BY_DEFAULT_TYPES = [
   TransactionType.moneyAccountDeposit,
@@ -71,6 +72,7 @@ export const Footer = () => {
     TRANSFER_TRANSACTION_TYPES.includes(transactionType) &&
     transactionMetadata?.origin === MMM_ORIGIN;
   const isPayLoading = useIsTransactionPayLoading();
+  const { isGaslessLoading } = useIsGaslessLoading();
 
   const { isFooterVisible: isFooterVisibleFlag, isTransactionValueUpdating } =
     useConfirmationContext();
@@ -158,7 +160,8 @@ export const Footer = () => {
     needsCameraPermission ||
     hasBlockingAlerts ||
     isTransactionValueUpdating ||
-    isPayLoading;
+    isPayLoading ||
+    isGaslessLoading;
 
   const buttons = [
     {
--- a/app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
+++ b/app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
@@ -1,7 +1,8 @@
-import { useTransactionMetadataRequest } from '../transactions/useTransactionMetadataRequest';
-import { useAsyncResult } from '../../../../hooks/useAsyncResult';
-import { isRelaySupported } from '../../../../../util/transactions/transaction-relay';
+import { useQuery } from '@tanstack/react-query';
 import { Hex } from '@metamask/utils';
+
+import { useTransactionMetadataRequest } from '../transactions/useTransactionMetadataRequest';
+import { isRelaySupported } from '../../../../../util/transactions/transaction-relay';
 import { isHardwareAccount } from '../../../../../util/address';
 import { useGaslessSupportedSmartTransactions } from './useGaslessSupportedSmartTransactions';
 
@@ -20,7 +21,7 @@ import { useGaslessSupportedSmartTransactions } from './useGaslessSupportedSmart
 export function useIsGaslessSupported() {
   const transactionMeta = useTransactionMetadataRequest();
 
-  const { chainId, txParams } = transactionMeta ?? {};
+  const { txParams, chainId } = transactionMeta ?? {};
 
   const {
     isSmartTransaction,
@@ -31,15 +32,11 @@ export function useIsGaslessSupported() {
   const shouldCheck7702Eligibility =
     !smartTransactionPending && !isSmartTransactionAndBundleSupported;
 
-  const { value: relaySupportsChain, pending: relayPending } =
-    useAsyncResult(async () => {
-      if (!shouldCheck7702Eligibility) {
-        return undefined;
-      }
-
-      return isRelaySupported(chainId as Hex);
-    }, [chainId, shouldCheck7702Eligibility]);
-
+  const { data: relaySupportsChain, isFetching: relayPending } = useQuery({
+    queryKey: ['relaySupportsChain', chainId],
+    queryFn: () => isRelaySupported(chainId as Hex),
+    enabled: shouldCheck7702Eligibility && Boolean(chainId),
+  });
   const is7702Supported = Boolean(
     relaySupportsChain &&
       // contract deployments can't be delegated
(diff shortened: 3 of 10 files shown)


app/util/test/renderWithProvider.tsx
```diff
@@ -17,6 +17,7 @@ import { Theme } from '../theme/models';
 import configureStore from './configureStore';
 import { RootState } from '../../reducers';
 import { FeatureFlagOverrideProvider } from '../../contexts/FeatureFlagOverrideContext';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
```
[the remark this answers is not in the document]
maxime-oe: I indeed had to fix a number of unit tests.
For the `CardHome.test.tsx` case, I see:
```
jest.mock('@tanstack/react-query', () => ({
  ...jest.requireActual('@tanstack/react-query'),
  useQueryClient: jest.fn(() => ({
    removeQueries: mockRemoveQueries,
  })),
}));
```
And the test runs fine, maybe I am misunderstanding Cursor's comment ?

nikolastoimenovski-consensys:
Tested locally, fixed, adding the "qa passed" label.

app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
```diff
@@ -31,15 +32,11 @@ export function useIsGaslessSupported() {
...
-  const { value: relaySupportsChain, pending: relayPending } =
-    useAsyncResult(async () => {
-      if (!shouldCheck7702Eligibility) {
-        return undefined;
-      }
-
-      return isRelaySupported(chainId as Hex);
-    }, [chainId, shouldCheck7702Eligibility]);
-
+  const { data: relaySupportsChain, isFetching: relayPending } = useQuery({
```
matthewwalsh0, line 35: What's the benefit of switching to `useQuery` here? Is this required?

Any caching should be handled by the server and browser / os based on caching headers.

app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
```diff
@@ -55,12 +52,12 @@ export function useIsGaslessSupported() {
     !isHardwareWallet &&
     Boolean(isSmartTransactionAndBundleSupported || is7702Supported);
 
-  const isPending =
-    smartTransactionPending || (shouldCheck7702Eligibility && relayPending);
+  const is7702SupportedPending = shouldCheck7702Eligibility && relayPending;
+  const pending = smartTransactionPending || is7702SupportedPending;
```
matthewwalsh0, line 56: Is this needed? If not, does this file need changing at all?

app/components/Views/confirmations/components/footer/footer.tsx
```diff
@@ -158,7 +160,8 @@ export const Footer = () => {
     needsCameraPermission ||
     hasBlockingAlerts ||
     isTransactionValueUpdating ||
-    isPayLoading;
+    isPayLoading ||
+    isGaslessLoading;
```
matthewwalsh0, line 164: Does this have larger implications if mobile wasn't previously disabled while gas station was loading?

app/components/UI/Predict/components/PredictPositionsHeader/PredictPositionsHeader.test.tsx
matthewwalsh0, line 1: We wouldn't need any changes in Predict code if we didn't change to `useQuery`?

matthewwalsh0 (changes requested):

app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
maxime-oe: The problem was that `useAsync` was not properly managing the `{ enabled }` mode with the `shouldCheck7702Eligibility && Boolean(chainId)` condition. The hook go through cycles where `chainId` has some value, then hasn't, then has some value again.
I first tried to change `useAsync` so it would behave more that `useQuery`, then decided it was too risky cause `useAsync` is used in multiple places.
Then I tried to wrap the `shouldCheck7702Eligibility && Boolean(chainId)` condition manually in `useIsGaslessSupported` but I kept getting race conditions.

`useQuery` on the other hand worked immediately and exacted as one would expect when reading the code.  So I went for it - after 1/2 hrs of die and retry.

app/components/UI/Predict/components/PredictPositionsHeader/PredictPositionsHeader.test.tsx
maxime-oe: No but I couldn't find any clean alternative.
I modified those 2 test files after making the changes in [renderWithProvider.tsx](https://github.com/MetaMask/metamask-mobile/pull/28722/changes#diff-e3a666b130a17f8ece53c99f20c6ceceee27ac30e013ab4ed0948b2cf320139e)

app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
matthewwalsh0: Then the deeper issue is why the `chainId` is ever not set for an active transaction?

It should be available on the transaction meta for its entire lifetime, why doesn't the same behaviour happen in extension?

app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
maxime-oe: No I can roll back those two lines if reducing the LoCs / amount of changes is import. I still want to fight for my `useQuery` though, unless you see a clear alternative: Only one I'd see otherwise is making the hook more complex by not using either `useQuery` or `useAsync`.

app/components/Views/confirmations/components/footer/footer.tsx
maxime-oe: Yes it might actually feel less flaky on slower devices - the "Confirm" button blinks much more on Mobile than on Extension, maybe due to `isTransactionValueUpdating`. If `isGaslessLoading` remains "true" longer than the time for `isTransactionValueUpdating` to go `true` then `false` for example.
But on Tempo this "blinking" is more than UX gimmick, it makes the tx fail if the user taps during one of those blinks.

app/components/Views/confirmations/hooks/gas/useIsGaslessSupported.ts
maxime-oe: > Then the deeper issue is why the chainId is ever not set for an active transaction?

I think this never happens, but that `transactionMeta` is not always set. Maybe I was wrong about that one re-becoming `undefined` but at least one variable was doing back-and-forth - I should have noted this one.

I'll start over and try to redo it but it might not be as bullet-proof as the `useQuery` solution. The current solutions works (as opposed as before). It is just not the original code.

maxime-oe:
Dropped and replaced by https://github.com/MetaMask/metamask-mobile/pull/29188

PR #28864, opened 2026-04-15
feat: migrate `TextField` component (web3auth)

kirillzyusko:
## **Description**

Use `TextField` from DSRN package (pilot migration).

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/DSYS-278

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **Before**

https://github.com/user-attachments/assets/11c9a637-7019-49a0-ac93-46b16c6be72c

### **After**

https://github.com/user-attachments/assets/ab75abad-fed4-4b9a-b562-61a1aef4a9c5

- [ ] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Swaps the password inputs in the SRP import onboarding flow to a new `TextField` implementation and updates tests to match, which could affect input behavior (secure entry/disabled state) in a critical onboarding path.
> 
> **Overview**
> Updates `ImportFromSecretRecoveryPhrase` to use the design-system `TextField` instead of the legacy component-library `TextField` for the new/confirm password fields (including end-accessory visibility toggles and disabled/error states).
> 
> Adjusts the associated Jest tests to assert against the nested React Native `TextInput` inside the new `TextField` wrapper (e.g., `secureTextEntry`, `editable`, and `value`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5e1618890f3ae8f04ee8dc84b075fd0f5b851aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/components/Views/ImportFromSecretRecoveryPhrase/index.test.tsx
+++ b/app/components/Views/ImportFromSecretRecoveryPhrase/index.test.tsx
@@ -13,7 +13,7 @@ import Clipboard from '@react-native-clipboard/clipboard';
 import { MIN_PASSWORD_LENGTH } from '../../../util/password';
 import { BIOMETRY_TYPE } from 'react-native-keychain';
 import AUTHENTICATION_TYPE from '../../../constants/userProperties';
-import { Alert, InteractionManager } from 'react-native';
+import { Alert, InteractionManager, TextInput } from 'react-native';
 import { QRTabSwitcherScreens } from '../QRTabSwitcher';
 import { createStackNavigator } from '@react-navigation/stack';
 import { NavigationContainer } from '@react-navigation/native';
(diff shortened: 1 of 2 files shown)


app/components/Views/ImportFromSecretRecoveryPhrase/index.test.tsx
```diff
@@ -1280,16 +1280,24 @@ describe('ImportFromSecretRecoveryPhrase', () => {
       );
 
       // Initially passwords should be hidden
-      expect(passwordInput.props.secureTextEntry).toBe(true);
-      expect(confirmPasswordInput.props.secureTextEntry).toBe(true);
+      expect(passwordInput.findByType(TextInput).props.secureTextEntry).toBe(
```
kirillzyusko: @brianacnguyen @georgewrmarshall is it the expected way of writing unit tests now? I think it is is fragile and tedious. Should we modify `design-system` repo and assign ID like this:

```tsx
          {inputElement ?? (
            <Input
              {...props}
              ref={inputRef}
              testID={testID ? `${testID}-input` : undefined}
              textVariant={TextVariant.BodyMd}
              isDisabled={isDisabled}
              autoFocus={autoFocus}
              onBlur={onBlurHandler}
              onFocus={onFocusHandler}
              isStateStylesDisabled
```

brianacnguyen: @kirillzyusko the testID it should be passed through the props

georgewrmarshall: Yes, I think that’s right. Looking at `packages/design-system-react-native/src/components/TextField/TextField.tsx`, `testID` is pulled out of props and attached to the outer `Pressable`, so it never reaches the inner `Input` via the spread. That means consumers lose a direct selector on the actual `TextInput` and end up having to traverse the rendered tree in tests.

So this looks like a component bug / API gap rather than something mobile tests should normalize around. I think we should keep the wrapper `testID` if we need it, but also expose a stable way to target the inner input. Your `testID ? `${testID}-input` : undefined` suggestion seems reasonable as a first pass.

georgewrmarshall: I think the root issue here is a bug in our current `@metamask/design-system-react-native` `TextField` API.

Right now `TextField` destructures `testID` and applies it to the wrapping `Pressable`, but it does not provide a way for callers to target the inner `Input` / `TextInput` with a test id via the spread props. So the selector ends up on the chrome instead of the actual input element, which is why these tests have to reach into the rendered tree.

I’m inclined to hold off on this PR until the `TextField` refactor is complete in `metamask-design-system#1081`: https://github.com/MetaMask/metamask-design-system/pull/1081

That refactor will introduce breaking changes, but I think it gives us a better component API overall and aligns better with our conventions. In particular, it moves toward a cleaner separation where top-level props belong to the outer field container and inner input props are passed intentionally, instead of the current situation where the layering is effectively the other way around.

PR #29076, opened 2026-04-20
feat(predict): add Polymarket CLOB v2 support cp-7.73.1

matallui:
## **Description**

This PR implements Polymarket CLOB v2 support for MetaMask Predict (7.74 target), introducing a clean coexistence architecture that keeps v1 working while v2 is toggled on via `predictClobV2`. During the temporary CLOB host migration window, a companion rollout flag (`predictClobV2UseLegacyClobHost`) can force v2 traffic onto the legacy host for internal RC testing without turning v2 on for everyone by default.

The implementation is designed with deletion in mind — after the v1/v2 coexistence window, v1 can be cleanly removed with minimal surgery. The work follows the "data-first protocol definition" architecture from the internal plan: a private `protocol/` module owns all version-specific differences and is resolved once at the provider entry point. Lower-level helpers never read feature flags directly.

**Key design decisions:**
- Primary remote flag: `predictClobV2` (default `false`) enables the v2 protocol
- Temporary rollout flag: `predictClobV2UseLegacyClobHost` forces v2 to use `https://clob-v2.polymarket.com` during the migration window
- Canonical/default v2 host remains `https://clob.polymarket.com`
- Host selection is resolved once into `protocol.transport.clobBaseUrl`; lower-level helpers never read raw flags
- v2 API key caching is host-aware to avoid reusing credentials across host changes
- v2 `getBalance()` always returns `Safe USDC.e + Safe pUSD` — no branching on upgrade state
- Trade path is **relayer-only**, **Permit2-only** fee collection on pUSD, always runs preflight before submission
- Canonical v2 allowance requirement set defined once; inspector + compiler are small pure modules
- Deposit/withdraw use **release-time code choices** for preferred vs fallback variant — not runtime flags
- Wrap: always wraps **entire current Safe USDC.e balance** (never `MaxUint256`) when maintenance tx is already being emitted
- Unwrap: always unwraps **exact deficit** needed — no over-unwrapping

---

## **Commits**

This PR is now structured as 12 focused commits for easier review:

### 1. `feat(predict): add CLOB v2 feature flag plumbing`
Wires the `predictClobV2` boolean through the Predict feature-flag infrastructure:
- Adds `predictClobV2Enabled` to `PredictFeatureFlags` interface
- Adds `selectPredictClobV2EnabledFlag` selector
- Extends `resolvePredictFeatureFlags` to resolve the new flag
- Refactors `resolveVersionGatedBooleanFlag` helper to reduce duplication across the existing flag resolution logic
- Adds selector and resolver tests

### 2. `feat(predict): add CLOB v2 protocol and preflight foundation`
Introduces the two new private modules that everything else builds on:

**`protocol/`** — data-first protocol definitions:
- `definitions.ts`: v1 and v2 `PolymarketProtocolDefinition` objects; protocol resolution; builder code env config (`MM_PREDICT_BUILDER_CODE`); deposit/withdraw execution mode types
- `orderCodec.ts`: v2-aware order build, EIP-712 typed data, and relayer payload serialization
- `transport.ts`: shared CLOB transport helpers parameterized by protocol endpoint; collapses v1/v2 endpoint differences

**`preflight/`** — private readiness inspection and Safe plan construction:
- `v2AllowanceRequirements.ts`: canonical declarative list of all 10 v2 allowance requirements (7 ERC-20, 3 ERC-1155)
- `inspectMissingRequirements.ts`: reads on-chain state and returns the subset of requirements not yet satisfied
- `compileRequirementTransactions.ts`: compiles missing requirements into `SafeTransaction[]`
- `core.ts`: shared raw-fact readers, wrap/unwrap builders, and signed Safe execution helpers

All modules covered by unit tests.

### 3. `feat(predict): add CLOB v2 buy and sell flow`
Implements the full v2 trade path inside `PolymarketProvider`:
- `preflight/trade.ts`: `planTradePreflight` + `buildTradeAllowancesTx` — inspects missing v2 allowances and Safe USDC.e balance; compiles optional maintenance tx (allowances-only / wrap-only / allowances+wrap / none)
- `PolymarketProvider`: `placeOrder` now resolves the protocol once, runs preflight under v2, builds the optional `allowancesTx`, and uses the protocol's `orderCodec` for order construction, EIP-712 signing, and relayer payload
- `utils.ts`: adds `encodeWrapUsdceTransaction` and `encodeUnwrapTransaction` helpers (+ tests)
- v2 preview keeps `feeRateBps = '0'` until the upstream fee endpoint is confirmed

### 4. `feat(predict): add CLOB v2 deposit flow`
Adds the v2 deposit maintenance planner and wires it into `PolymarketProvider.prepareDeposit`:
- `preflight/deposit.ts`: `planDepositMaintenance` + `compileDepositMaintenanceTransactions` — inspects missing v2 allowances and **pre-existing** Safe USDC.e balance (does not incorporate the just-entered deposit amount); emits optional maintenance tx
- `PolymarketProvider.prepareDeposit`: under v2, resolves protocol, runs the maintenance planner, and attaches the optional maintenance tx to the deposit plan — the 3-step shape (optional deploy → transfer funding asset → optional maintenance tx) is preserved
- Currently wired to the `usdce-transfer` fallback mode; flip `depositMode` in the protocol definition to switch to pUSD-native when that dependency lands

### 5. `feat(predict): add CLOB v2 withdraw flow`
Adds the v2 withdraw planner and wires both the `prepareWithdraw` / `signWithdraw` contract:
- `preflight/withdraw.ts`: `planWithdraw` — reads missing v2 allowances and Safe USDC.e balance; computes the exact USDC.e deficit; compiles the final MultiSend (allowance repair → exact-deficit unwrap-to-Safe → USDC.e transfer to EOA)
- `PolymarketProvider.signWithdraw`: under v2, parses the requested amount from the original template calldata, calls `planWithdraw`, and returns the user-requested amount (not any larger intermediate amount)
- `safe/utils.ts`: adds `parseTransactionCalldata` helper to extract the withdraw amount from the stored template
- Currently wired to the `usdce-deficit-unwrap` fallback mode; flip `withdrawMode` in the protocol definition to switch to pUSD-native

### 6. `feat(predict): add CLOB v2 claim flow`
Adds the v2 claim planner and wires it into `PolymarketProvider.claimWinnings`:
- `preflight/claim.ts`: `planClaim` — reads EOA USDC.e directly (not Safe balances or provider `getBalance()`); computes `gasStationDeficit = max(0, MIN_GAS_STATION_USDCE_BALANCE - eoaUsdceBalance)`; proactively wraps the **entire current Safe USDC.e balance**; compiles the MultiSend in the required order: (1) missing allowance/operator repair, (2) wrap-all current Safe USDC.e, (3) claim subcalls, (4) optional exact-deficit unwrap to EOA
- `PolymarketProvider.claimWinnings`: under v2, resolves protocol and delegates to `planClaim`; uses protocol-owned claim targets (pUSD as collateral) rather than per-position collateral metadata

### 7. `test(predict): add CLOB v2 integration coverage`
Adds integration-level tests and reorganizes the provider test suite:
- `preflight/workflows.test.ts`: end-to-end workflow planner tests covering all four preflight paths (trade, deposit, claim, withdraw) with concrete on-chain mock scenarios — verifies MultiSend ordering invariant (approvals first, then wraps/claims/transfers)
- `PolymarketProvider.test.ts`: reorganized into `v1` / `v2` describe blocks; covers protocol routing (flag off → v1, flag on → v2), v2 `getBalance()` aggregation, all four trade preflight outcomes (none / allowances-only / wrap-only / allowances+wrap), deposit/withdraw preferred vs fallback shape, and claim ordering

### 8. `fix: preserve EIP-712 field order in buildProtocolUnsignedOrder`
Fixes the v2 order-signing codec so the generated typed data preserves the expected field ordering for EIP-712 signing.

### 9. `codex: address PR review feedback (#29076)`
Addresses review feedback in the allowance inspector path:
- forwards `requirement.tokenAddress` into the ERC-1155 operator approval read instead of relying on a hardcoded contract address
- adds assertions proving the token address is propagated correctly through the read path

### 10. `refactor(predict): tighten CLOB v2 preflight workflow seams`
Addresses follow-up review feedback without changing behavior:
- makes lower-level transaction compilers private where possible
- keeps `getClaimRequirements` exported for focused pure testing
- tightens `preflight/workflows.test.ts` to assert through planner entry points instead of private helpers
- renames claim gas-top-up terminology for clarity (`MIN_GAS_STATION_USDCE_BALANCE_RAW`, `gasStationDeficit`) while keeping `eoaUsdceBalance` unchanged

### 11. `feat(predict): add configurable CLOB v2 host override`
Adds temporary host-migration plumbing while keeping canonical host resolution centralized:
- introduces canonical and legacy CLOB host constants
- resolves the selected host into `protocol.transport.clobBaseUrl`
- threads the resolved host through API key creation, order-book reads, and preview flow
- isolates API key cache entries by protocol + host + address

### 12. `refactor(predict): gate legacy CLOB v2 host with boolean flag`
Simplifies the rollout shape for internal RC testing:
- replaces the nested raw `predictClobV2.clobBaseUrl` rollout control with a second version-gated boolean flag: `predictClobV2UseLegacyClobHost`
- keeps the lower-level host plumbing from commit 11 intact
- allows internal RC builds to enable the legacy host remotely while internal users locally toggle `predictClobV2`

---

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/PRED-817

## **Manual testing steps**

```gherkin
Feature: Polymarket CLOB v2 trading

  Scenario: user trades with CLOB v2 enabled
    Given the predictClobV2 remote flag is enabled
    And the user has an active Polymarket Safe

    When user places a buy or sell order
    Then the provider runs preflight and submits any needed allowance/wrap tx first
    And the order is submitted via the relayer with X-Clob-Version: 2

  Scenario: internal RC uses the legacy v2 CLOB host
    Given the predictClobV2UseLegacyClobHost remote flag is enabled
    And predictClobV2 is enabled via local override

    When user places a buy or sell order
    Then v2 CLOB requests use https://clob-v2.polymarket.com
    And API keys are cached separately from canonical-host credentials

  Scenario: user deposits with CLOB v2 enabled
    Given the predictClobV2 remote flag is enabled

    When user initiates a deposit
    Then the deposit plan includes an optional maintenance tx (allowance repair + wrap pre-existing USDC.e)
    And the newly deposited amount is not included in the wrap

  Scenario: user withdraws with CLOB v2 enabled
    Given the predictClobV2 remote flag is enabled
    And the user requests a specific withdrawal amount

    When user confirms the withdraw
    Then only the exact USDC.e deficit is unwrapped (not all pUSD)
    And the reported amount matches the user-requested amount

  Scenario: user claims winnings with CLOB v2 enabled
    Given the predictClobV2 remote flag is enabled
    And the user has winning positions

    When user claims
    Then all missing v2 allowances are repaired first
    And the entire pre-existing Safe USDC.e is wrapped
    And claim subcalls execute after
    And only the exact EOA gas-top-up deficit is unwrapped to USDC.e

  Scenario: CLOB v2 flag is disabled (v1 behavior)
    Given the predictClobV2 remote flag is disabled

    When user interacts with any predict flow
    Then v1 protocol is used unchanged
```

## **Screenshots/Recordings**

https://www.loom.com/share/155120bd46c44723a8b838172b4fd45b

- [ ] I've tested on Android

---

> [!NOTE]
> **High Risk**
> High risk because it changes Polymarket trade/claim/deposit/withdraw transaction construction and relayer submission paths, including new Safe preflight logic and EIP-712 signing for a new protocol version.
> 
> **Overview**
> Adds **Polymarket CLOB v2** support behind new Predict feature flags (`predictClobV2Enabled` plus optional legacy host override), with host-aware API key caching and endpoint selection.
> 
> Refactors `PolymarketProvider` to resolve a protocol definition once and route **preview + order submission** through a new `protocol/` module (v2 uses new order schema/typed data, zero preview fee rate, and a relayer request header for v2 routing).
> 
> Introduces a `preflight/` layer that inspects on-chain allowance/approval requirements and builds signed Safe executions for **trade allowances**, **deposit maintenance**, **claim**, and **withdraw** (including wrap/unwrap flows and balance aggregation across USDC.e + pUSD for v2). Adds extensive unit/integration tests and updates Safe utilities (raw USDC amount decoding, token-address-aware approvals, Permit2 token selection) plus CI env wiring for `MM_PREDICT_BUILDER_CODE`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5df04d5b269e607c431158ab87aac21a835afed6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/components/UI/Predict/providers/polymarket/protocol/transport.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function normalizeRelayerHeaders(headers: ClobHeaders): Record<string, string> {
+  const normalizedHeaders: Record<string, string> = { ...headers };
+
+  for (const [key, value] of Object.entries(headers)) {
+    normalizedHeaders[key.replace(/_/gu, '-')] = value;
+  }
+
+  return normalizedHeaders;
+}
```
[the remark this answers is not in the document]
matallui: This is fine, it is the same behavior we had before.

app/components/UI/Predict/providers/polymarket/protocol/orderCodec.ts
```diff
@@ -0,0 +1,378 @@
...
+      tokenId,
+      makerAmount,
+      takerAmount,
+      expiration: '0',
+      timestamp: `${nowInSeconds}`,
+      metadata: protocol.order.metadata,
+      builder,
+      side,
+      signatureType,
+    };
```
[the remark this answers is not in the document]
matallui: Not a bug — `expiration` is declared as an optional field in the `OrderDataV2` interface by design. The Polymarket v2 relayer accepts it in the request body; it simply isn't part of the EIP-712 signed struct (v2 uses `timestamp` for order timing instead). This exact shape was already present and working in the codebase before this PR.

app/components/UI/Predict/utils/resolvePredictFeatureFlags.test.ts
```diff
@@ -184,6 +186,55 @@ describe('resolvePredictFeatureFlags', () => {
 
     expect(result.fakOrdersEnabled).toBe(true);
     expect(result.predictWithAnyTokenEnabled).toBe(false);
+    expect(result.predictClobV2Enabled).toBe(false);
```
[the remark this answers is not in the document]
matallui: Not a bug — the test passes. The bot did not account for extendedSportsFlag (DEFAULT_EXTENDED_SPORTS_MARKETS_FLAG) being unconditionally present, which consumes mock slot #1 in both old and new code. With the refactored short-circuit (no predictMarketHighlights in the test), the call sequence is: (1) extendedSports=false, (2) fakOrders=true, (3) predictWithAnyToken=false, (4) predictUpDown=false, (5) predictClobV2 gets no mock left, falls back to undefined from beforeEach -> false. All three assertions pass correctly. Verified locally: all 18 tests in this file are green.

caieu (dismissed):

app/components/UI/Predict/providers/polymarket/preflight/deposit.ts
```diff
@@ -0,0 +1,88 @@
...
+}): SafeTransaction[] {
+  return compileAllowanceMaintenanceTransactions({
+    protocol,
+    safeAddress,
+    missingRequirements,
+    usdceBalance: preExistingSafeUsdceBalance,
+  });
+}
+
+export async function buildDepositMaintenanceTransaction({
```
matthewwalsh0, line 71: Minor, I like the modularity of these preflight util files, but why export all the functions?

Would it make the tests less brittle if we only use a single entrypoint?

Plus better encapsulation in the utils.

app/components/UI/Predict/providers/polymarket/preflight/withdraw.ts
```diff
@@ -0,0 +1,167 @@
...
+      operation: OperationType.Call,
+      value: '0',
+    });
+
+    return transactions;
+  }
+
+  const unwrapTransaction = buildUnwrapTransaction({
+    recipientAddress: safeAddress,
+    amount: deficit,
```
matthewwalsh0, line 119: Just confirming the asymmetry that deposit wraps everything it can (excluding the new value), but withdraw adds the approvals, but no `wrap` and only unwraps the deficit?

Did we discuss wrapping everything (minus the withdraw amount) in every operation to optimise the performance of future orders?

app/components/UI/Predict/providers/polymarket/preflight/claim.ts
```diff
@@ -0,0 +1,220 @@
...
+        address: safeAddress,
+        tokenAddress: protocol.collateral.legacyUsdceToken,
+      }),
+      getRawTokenBalance({
+        address: signer.address,
+        tokenAddress: protocol.collateral.legacyUsdceToken,
+      }),
+    ]);
+
+  const deficit =
```
matthewwalsh0: Minor, could we name these variables and constants to add more clarity this is for gas station?

app/components/UI/Predict/providers/polymarket/preflight/claim.ts
```diff
@@ -0,0 +1,220 @@
...
+  compileAllowanceMaintenanceTransactions,
+  getRawTokenBalance,
+} from './core';
+import { inspectMissingRequirements } from './inspectMissingRequirements';
+import {
+  getCanonicalV2AllowanceRequirements,
+  type V2AllowanceRequirement,
+} from './v2AllowanceRequirements';
+
+const MIN_CLAIM_BALANCE_RAW = BigInt(
```
matthewwalsh0: Minor, should this be a feature flag if gas changes in future or we want to reduce dust?

matthewwalsh0 (dismissed):

app/components/UI/Predict/providers/polymarket/preflight/deposit.ts
matallui: Tightened the workflow-module surface here. I kept the real entrypoints exported and made the internal compile helpers private, then moved the workflow coverage up to the planner/output seam in the tests. I left `getClaimRequirements` exported because it is the small pure requirement seam used by the claim-specific workflow test.

app/components/UI/Predict/providers/polymarket/preflight/claim.ts
matallui: Kept this as a constant rather than adding another feature flag. If we need to tune the threshold later, config would be the better abstraction than rollout flagging, since this is a numeric policy value rather than a behavior toggle.

app/components/UI/Predict/providers/polymarket/preflight/withdraw.ts
matallui: Yes — that asymmetry is intentional. Deposit/trade/claim are the maintenance opportunities that proactively converge the Safe toward the pUSD steady state. Withdraw is the egress path, so the fallback behavior stays minimal: repair approvals if needed, unwrap only the exact USDC.e deficit, then transfer the exact requested amount. I kept that behavior unchanged rather than adding extra wrap/cleanup work on the exit path.

app/components/UI/Predict/providers/polymarket/preflight/claim.ts
matallui: Renamed the claim-side threshold/deficit naming to make the gas-station intent explicit while keeping `eoaUsdceBalance` as-is, since that name already reflects the full EOA USDC.e balance being read.

app/components/UI/Predict/providers/polymarket/PolymarketProvider.ts
```diff
@@ -272,6 +283,399 @@ export class PolymarketProvider implements PredictProvider {
...
+    preview,
+  }: {
+    signer: Signer;
+    preview: OrderPreview;
+  }) {
+    const protocol = this.#getProtocol();
+
+    if (protocol.key !== 'v1') {
+      throw new Error('Invalid Polymarket protocol for v1 order submission');
+    }
```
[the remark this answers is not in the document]
matallui: Adjusted the v1 path to consume the protocol already resolved in `placeOrder()` instead of re-reading it inside `#submitOrderV1`.

Also added a regression test in `PolymarketProvider.test.ts` with a stateful `getFeatureFlags()` implementation that flips `predictClobV2Enabled` after the initial resolution; `placeOrder()` now stays on the already-selected v1 submission path and succeeds.

builds.yml
```diff
@@ -242,6 +244,7 @@ builds:
       METAMASK_ENVIRONMENT: 'e2e'
       METAMASK_BUILD_TYPE: 'main'
       MM_PORTFOLIO_URL: 'https://portfolio.dev-api.cx.metamask.io'
+      MM_PREDICT_BUILDER_CODE: '0xd48300a99deac0f23265dad1f59d32920be33b75919201a88ed80f457f97b924'
```
Gudahtt, line 247: Generally it's better for the E2E builds to match production if possible. Network requests get mocked regardless

I know we set a bunch of dev APIs here already, but those are also a mistake

matallui: So my main concern here is that we might want to avoid using the prod builder code to submit any orders outside of the prod environment, since we're going to be collecting Predict stats directly from on-chain data, and I would not want to mix test data with prod data. And if they mock the requests, then it shouldn't matter to them anyway, right?

app/components/UI/Predict/providers/polymarket/PolymarketProvider.ts
```diff
@@ -272,6 +283,395 @@ export class PolymarketProvider implements PredictProvider {
...
+
+    const signedOrder = {
+      ...order,
+      signature,
+    };
+    const signerApiKey = await this.getApiKey({
+      address: signer.address,
+      protocol,
+    });
+    const { feeCollection, fakOrdersEnabled } = this.#getFeatureFlags();
```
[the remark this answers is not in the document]
matallui: not an issue

builds.yml
Gudahtt: Hmm OK. Still I would argue that our ideal for E2E builds is for them to be as close to prod as possible. And we should be confident in our mocks, I would hope? Maybe not yet though.

I know we have a fallback case to intercept and block any unmocked network calls in extension E2E tests, but I don't know if it's setup on mobile yet (probably not).

Going with the test value for now seems fine; we can revisit later when we're more confident in our mock strategy.

matallui: Fair enough! And yeah, we can always change back. I do not think this is a big deal for E2E testing if we're mocking the requests, since it's not really testing anything other than the variable being set.

tommasini (approved):

app/components/UI/Predict/providers/polymarket/PolymarketProvider.ts
```diff
@@ -1855,6 +2037,44 @@ export class PolymarketProvider implements PredictProvider {
...
+
+      if (maintenanceTransaction) {
+        transactions.push(maintenanceTransaction);
+      }
+
+      return {
+        chainId: CHAIN_IDS.POLYGON,
+        transactions,
+      };
+    }
```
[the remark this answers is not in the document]
matallui: that's fine

MarioAslau (approved):
LGTM

PR #28084, opened 2026-03-30
feat: wallet-side analytics to SDKConnect v1 deeplink paths 

wenfix:
## **Description**

Adds wallet-side analytics for the deeplink protocol path (`comm=deeplinking`), which previously produced zero events on the wallet side. Also fixes a gap where `DEEP_LINK_USED` never fired for SDK action deeplinks (`connect`, `mmsdk`, `bind`) because they early-exited the universal link handler before any analytics code ran.

**Part A — `DEEP_LINK_USED` for SDK deeplinks**

`handleUniversalLink.ts` intercepts SDK actions and routes them to `handleMetaMaskDeeplink` before the rest of the handler runs. This meant the existing `DEEP_LINK_USED` analytics (which tracks every other deeplink type) never fired for SDK deeplinks. This PR fires `trackDeepLinkAnalytics` before the early-exit, using:
- `route: 'sdk_connect'` or `'sdk_mmsdk'` (new `DeepLinkRoute` enum values)
- `signatureStatus: MISSING` (SDK deeplinks carry no signature)
- `interstitial: NOT_SHOWN` (SDK deeplinks bypass the modal)
- `branchParams: {}` — see note below

Required additions: `SDK_CONNECT` and `SDK_MMSDK` to the `DeepLinkRoute` enum, `ACTIONS.CONNECT/MMSDK/ANDROID_SDK` to `SUPPORTED_ACTIONS`, and corresponding cases in `mapSupportedActionToRoute`.

**Part B — wallet-side `sdk-analytics` events for deeplink protocol**

The socket-relay path already fires `wallet_connection_request_received` and `wallet_action_received` via `@metamask/sdk-analytics`. The deeplink protocol path (`DeeplinkProtocolService`) fired nothing. This PR adds the same events at the service layer:

| Method | Event | Key properties |
|--------|-------|----------------|
| `handleConnection()` | `wallet_connection_request_received` | `transport: 'deeplink_protocol'`, `connection_type: 'new_session' \| 'reconnect'`, `sdk_version` |
| `processDappRpcRequest()` | `wallet_action_received` | `transport: 'deeplink_protocol'`, `rpc_method` |
| `handleMessage()` / `handleEventAsync()` | `wallet_action_received` | `transport: 'deeplink_protocol'`, `rpc_method` |

All calls are guarded by `if (anonId)` and include only non-PII properties (`anon_id`, `transport`, `connection_type`, `rpc_method`). No addresses, RPC params, or `originatorInfo` content is included.

[...]

PR #29186, opened 2026-04-22
feat(analytics): fire DEEP_LINK_USED for SDK deeplinks

wenfix:
## **Description**

SDK action deeplinks (`connect`, `mmsdk`, `bind`) previously early-exited `handleUniversalLink` before `trackDeepLinkAnalytics` ran, so they never produced a `DEEP_LINK_USED` event. This PR fires the event before the early-exit so SDK deeplinks are measured alongside the rest of the deeplink surface.

**Scope — which "connect" this is:**

This covers the **MetaMask SDK** deeplink surface only — `@metamask/sdk` / `sdk-communication-layer`, i.e. what's now referred to as **"SDKv1"**:

- `connect` (iOS/universal) and `bind` (`ACTIONS.ANDROID_SDK`, the Android SDK) → `route: 'sdk-connect'`
- `mmsdk` (the SDK's RPC message channel) → `route: 'sdk-mmsdk'`

It is **not** MetaMask Connect product, a.k.a. **"SDKv2"**, internally still named `SDKConnectV2`. MMC connects over the Mobile Wallet Protocol (`metamask://connect/mwp?...`), which is intercepted earlier in `handleDeeplink.ts` via `SDKConnectV2.isMwpDeeplink`, already fires its own `DEEP_LINK_USED` with `route: 'mmc-mwp'`, and `return`s before reaching this SDK path. The two surfaces are mutually exclusive, and the `sdk-` prefix (rather than `mm-`/`mmc-`) keeps them distinct in the `DeepLinkRoute` enum.

**Event properties:**

* `route: 'sdk-connect' | 'sdk-mmsdk'` — new `DeepLinkRoute` enum values (kebab-case to match the rest of the enum)
* `signatureStatus: MISSING` — SDK deeplinks carry no signature
* `interstitialAction: NOT_SHOWN` — SDK deeplinks bypass the modal

**Known deviation from AC — Branch.io attribution not fetched (`was_app_installed` always `true`):**

The ticket AC says Branch.io attribution should be captured "on the same basis as all other `DEEP_LINK_USED` events." It isn't — intentionally. The SDK early-exit in `handleUniversalLink.ts` fires _before_ the Branch.io fetch, which has a 500ms timeout guard. Moving the fetch above the early-exit would add up to 500ms of latency to every SDK connect deeplink before `handleMetaMaskDeeplink` runs — a visible delay to connection.

Instead, we pass `branchParams: {}` to skip the fetch. `detectAppInstallation({})` returns `true`, so `was_app_installed` is always `true` for this path. SDK deeplinks (`metamask://connect?comm=deeplinking`) are not distributed through Branch.io campaigns, so `was_app_installed: true` is the correct default anyway.

Part A of WAPI-1347, split from #28084. Part B (wallet-side deeplink events) will move to MetaMetrics + `transport_type: 'deeplink_protocol'` in a follow-up, now that `@metamask/sdk-analytics` has been removed.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: WAPI-1347

## **Manual testing steps**

Feature: DEEP_LINK_USED event for MetaMask SDK deeplinks

  Scenario: SDK connect deeplink fires analytics
    Given the MetaMask app is installed and MetaMetrics is enabled
    When a dapp opens metamask://connect?comm=deeplinking&... (or an https://metamask.app.link/connect equivalent)
    Then a DEEP_LINK_USED event is emitted with route='sdk-connect'
    And signature_status='missing', interstitial_action='not shown', was_app_installed=true
    And the existing connection flow in handleMetaMaskDeeplink still runs without added latency

  Scenario: MMSDK deeplink fires analytics
    When a dapp opens metamask://mmsdk?... (or the https equivalent)
    Then a DEEP_LINK_USED event is emitted with route='sdk-mmsdk'

  Scenario: MetaMask Connect (MMC / MWP) is unaffected
    When an MMC dapp opens metamask://connect/mwp?...
    Then it is still tracked separately as route='mmc-mwp' (not sdk-connect), via the existing SDKConnectV2 path

  Scenario: non-SDK deeplinks are unaffected
    When a user opens any non-SDK deeplink (dapp, wc, card-home, etc.)
    Then the existing DEEP_LINK_USED event is emitted with the same route/properties as before

### **Before**

N/A — no event was emitted for SDK deeplinks.

- [x] I've tested on Android


app/core/DeeplinkManager/util/deeplinks/deepLinkAnalytics.ts
```diff
@@ -599,6 +601,11 @@ export const mapSupportedActionToRoute = (
       return DeepLinkRoute.CARD_HOME;
     case ACTIONS.NFT:
       return DeepLinkRoute.NFT;
+    case ACTIONS.CONNECT:
+    case ACTIONS.ANDROID_SDK:
+      return DeepLinkRoute.SDK_CONNECT;
+    case ACTIONS.MMSDK:
+      return DeepLinkRoute.SDK_MMSDK;
     default:
```
adonesky1, line 640: The naming conventions here are so confusing

app/core/DeeplinkManager/handlers/legacy/__tests__/handleUniversalLink.test.ts
```diff
@@ -145,6 +149,111 @@ describe('handleUniversalLink', () => {
...
+      await handleUniversalLink({
+        instance,
+        handled: jest.fn(),
+        urlObj: extractURLParams(url).urlObj,
+        url,
+        source: 'test',
+      });
+
+      expect(mockCreateBuilder).toHaveBeenCalledWith(
+        expect.objectContaining({
```
adonesky1, line 207: this assertion (and the two below) doesn't actually verify the route — `mapSupportedActionToRoute` is mocked to `() => 'test-route'` at line 79, so the new SDK route values (`sdk-connect`/`sdk-mmsdk`) are never checked. anyone refactoring the mapping could break it without these tests catching it.

suggest `mockReturnValueOnce(DeepLinkRoute.SDK_CONNECT)` (and `SDK_MMSDK` for the mmsdk case) per test, then assert that same value here.

wenfix: good catch, addressed here: 404b4842bb

adonesky1 (approved):
LGTM!

jiexi (approved):

tommasini (approved):

PR #29193, opened 2026-04-22
feat: hub page discovery tabs A/B test 

vinnyhoward:
## **Description**

Implements the treatment UI for the `coreMCU589AbtestHubPageDiscoveryTabs` A/B test

When the treatment variant is active, the homepage replaces the standard scrollable layout with a top-tab navigation bar exposing three verticals:

- **Portfolio:** existing homepage sections with balance header and pull-to-refresh
- **Perpetuals:** `PerpsHomeView` wrapped in connection + stream providers
- **Predictions:** `PredictFeed` wrapped in preview sheet provider

When the control variant is active (or the flag is absent), the existing homepage layout is fully preserved.

Key implementation details:
- `HomepageDiscoveryTabs`: new component that owns tab layout, per-tab gradient crossfade, and wallet header hide/show coordination across all three tabs
- `TabsList` — design-system tab primitive with `keepMounted` support. Perpetuals and Predictions both use `keepMounted={false}`:
  - **Performance**: both screens have a heavy hydration cost on first mount; keeping them alive while invisible wastes resources on tabs the user may never visit
  - **Open connections**: `PerpsHomeView` establishes WebSocket channels for live market data via its stream providers on mount; leaving these running in the background wastes bandwidth and server-side connection slots
  - **Memory**: `PredictFeed` and its preview sheet provider hold feed state that can safely be discarded between visits
  - Portfolio uses `keepMounted={true}` (the default) since it is the landing tab and its scroll position and section state should survive switching away and back

- Wallet header animates up/down on scroll via Reanimated shared values; icon collapse and gradient opacity are synced via `TabIconAnimationContext`
- Scroll event forwarding keeps existing `HOME_VIEWED` section analytics working in the Portfolio tab
- A/B gating is handled by `useABTest(HUB_PAGE_DISCOVERY_TABS_AB_KEY)`

[Figma Design](https://www.figma.com/design/z0panHXrMSMUSof2SaPkd4/Home-2026?m=auto&node-id=4280-62214&t=AAKr2hzmyPx57F4Y-1) for reference. Only difference between this UI and the Figma is that the tabs take up the entire space 

## **Changelog**

CHANGELOG entry:null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TMCU-591

## **Manual testing steps**

```gherkin
Feature: Hub Page Discovery Tabs

  Scenario: treatment variant shows top tabs
    Given the coreMCU589AbtestHubPageDiscoveryTabs flag is set to "treatment"

    When user opens the app to the Wallet screen
    Then three tabs are visible: Portfolio, Perpetuals, Predictions
    And switching between tabs is smooth with no layout shift
    And the wallet header hides on scroll and restores on tab switch
    And pull-to-refresh works on the Portfolio tab

  Scenario: control variant preserves existing layout
    Given the coreMCU589AbtestHubPageDiscoveryTabs flag is set to "control"

    When user opens the app to the Wallet screen
    Then the existing homepage layout is shown with no tabs
```

### iOS

Dark Mode

https://github.com/user-attachments/assets/7df6a46d-b3b3-44cc-a697-b796581dd759

Light Mode (No Gradient)

https://github.com/user-attachments/assets/97b1f901-6092-42f9-926c-e8e6785c6f4e

### Android

Dark Mode

https://github.com/user-attachments/assets/4232f3fc-a47c-4516-93c2-104e4a3fb5cb

Light Mode (No Gradient)

https://github.com/user-attachments/assets/3f11dd8f-696f-438e-9867-1b08dc69200d

### **Before**

`~`

### **After**

`~`

- [x] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Introduces a new A/B-test-gated navigation structure and shared header animation/scroll handling on the Wallet home screen, which could affect core navigation and scroll/refresh behavior. Risk is mitigated by control-path preservation and added unit coverage, but Reanimated worklet interactions and tab switching edge cases remain sensitive.
> 
> **Overview**
> When the Hub Page Discovery Tabs A/B test is in *treatment* (and homepage sections are enabled), Wallet home now renders a new top-level `HomepageDiscoveryTabs` experience with three tabs: Portfolio, Perpetuals (`PerpsHomeView`), and Predictions (`PredictFeed`). The control path preserves the existing scroll layout.
> 
> Adds `useDiscoveryScrollManager`, a Reanimated-backed hook that hides/shows the shared Wallet header based on scroll threshold, restores per-tab header state on tab entry, forwards scroll events back to JS (for analytics), and emits `onHeaderHiddenChange` to sync sibling animations.
> 
> Updates `Wallet` to measure/animate the header via shared values, pass refresh/portfolio header/scroll callbacks into the new tabs view, and crossfade a dark-mode gradient overlay per active tab. `PerpsHomeView` and `PredictFeed` gain `hideHeader` and header-sync props, and `useFeedScrollManager` can now notify header hidden/show changes; tests are added/updated across these components and the new hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38e679d20c800a000ba1755ca4ddf60092e33859. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -0,0 +1,358 @@
...
+};
+
+/**
+ * Thin wrapper that exposes a `tabLabel` prop consumed by TabsList to build
+ * the tab bar. The children are rendered as the tab's content.
+ */
+interface DiscoveryTabViewProps {
+  tabLabel: string;
+  tabIcon?: IconName;
+  keepMounted?: boolean;
```
gambinish, line 67: I suspect that this is at least part of the issue. `TabList` defaults to `keepMounted` true. So if we omit this, each tab remains mounted (including perps and predict tabs)

I know that the perps views call several websocket channels to fetch data, so we are certainly wasting bandwidth by keeping these views mounted even if not visible.

What do you think about making this `keepMounted` property required, and passing it in as `false`?

app/components/UI/Predict/hooks/useDiscoveryScrollManager.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  const scrollHandler = useAnimatedScrollHandler({
+    onScroll: (event) => {
+      'worklet';
+
+      // Forward to the section-visibility tracking subscribers
+      runOnJS(callPortfolioScroll)();
+
+      const currentY = event.contentOffset.y;
+      runOnJS(callScrollEvent)(currentY, event.layoutMeasurement.height);
```
gambinish: Should this only run on the Portfolio tab? This seems like it might be getting called unconditionally, regardless of the tab that the user is on. It seems like this method is only really useful on the portfolio tab.

Might be contributing to some of the lagginess we're seeing.

Somewhat related, but would it be possible to somehow combine these two `runOnJS` calls? I think that would help reduce the pressure on the js thread.

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -0,0 +1,358 @@
...
+            onChangeTab={handleChangeTab}
+            tabsListContentTwClassName="px-0"
+            style={tw.style('mt-2')}
+            tabsBarProps={{ fillWidth: true, collapseAnim: tabBarCollapseAnim }}
+          >
+            <DiscoveryTabView tabLabel="Portfolio" tabIcon={IconName.Portfolio}>
+              <Reanimated.ScrollView
+                showsVerticalScrollIndicator={false}
+                onScroll={scrollHandler}
+                scrollEventThrottle={16}
```
gambinish, line 291: One easy thing to tinker with could be this. Does the FPS improve at all if you bump this to say `32`?

app/components/UI/Predict/hooks/useDiscoveryScrollManager.ts
gambinish: One quick test you could try is to comment both of these out and compare FPS perf. I did this and it didn't see the UI drop below 60FPS

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: I don't think I see a difference between the two

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: The tradeoff is that Perps takes ~1 second to fully mount and render on each visit, which would be noticeable every time the user switches to that tab. `keepMounted={true}` was intentional, it trades bandwidth for a faster tab switch UX. The sluggishness in dev mode is mostly JS thread overhead that isn't present in release builds, so the current behavior feels acceptable there. I do like the idea of opting out of websocket connections though, once we solve the 1 second mounting issue I think we should do this. We currently do it for the Predict screen.

app/components/UI/Predict/hooks/useDiscoveryScrollManager.ts
vinnyhoward: I'm currently investigating this and will get back to you

vinnyhoward: I've combined the `runOnJS` good callout on that. As for the unconditional runs of this function call. It only runs for the active tab, while other tabs have its own instances and only are in use if they tab is active

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
gambinish: Hmm, we have a preloader that "warms" the perps tab to prevent this kind of remount if it happens within a given grace period. Is this still an issue after your changes to optimize the `useWithdrawRequests` selector?

vinnyhoward: Yeah doesn't seem to be an issue anymore, I'm testing it now. I've added `keepMounted={false}` for Perps and things seem fast. We won't have to worry about idle listeners

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -0,0 +1,367 @@
...
+    },
+    ref,
+  ) => {
+    const tabsRef = useRef<TabsListRef>(null);
+    const homepageRef = useRef<SectionRefreshHandle>(null);
+    const perpsTabEnterRef = useRef<(() => void) | null>(null);
+    const tw = useTailwind();
+    const { styles } = useStyles(styleSheet, {});
+    const { themeAppearance } = useTheme();
+    const isDarkMode = themeAppearance === 'dark';
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [18aee86](https://github.com/MetaMask/metamask-mobile/pull/29193/commits/18aee8602ccb94910ebbde21af4f4c7902e54bfa)

app/component-library/components-temp/Tabs/TabsBar/TabsBar.tsx
```diff
@@ -304,55 +316,73 @@ const TabsBar: React.FC<TabsBarProps> = ({
...
-                })}
-              />
-            )}
-          </Box>
-        </Box>
+                },
+              )}
+            />
+          )}
+        </Animated.View>
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [18aee86](https://github.com/MetaMask/metamask-mobile/pull/29193/commits/18aee8602ccb94910ebbde21af4f4c7902e54bfa)

app/components/UI/Perps/Views/PerpsHomeView/PerpsHomeView.tsx
```diff
@@ -124,6 +142,34 @@ const PerpsHomeView = () => {
...
+      onScrollEvent: handleScrollEvent,
+      onHeaderHiddenChange,
+    });
+
+  // Expose onTabEnter to the parent so it can restore this tab's header state on switch.
+  useEffect(() => {
+    if (tabEnterCallbackRef) {
+      tabEnterCallbackRef.current = perpsOnTabEnter;
+    }
+  }, [tabEnterCallbackRef, perpsOnTabEnter]);
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [18aee86](https://github.com/MetaMask/metamask-mobile/pull/29193/commits/18aee8602ccb94910ebbde21af4f4c7902e54bfa)

app/component-library/components/Icons/Icon/assets/candlestick.svg
```diff
@@ -1 +1,4 @@
-<svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><g><path d="m9.49932 5.33142h-1.66714v1.66714h-1.66714v10.00284h1.66714v1.6672h1.66714v-1.6672h1.66718v-10.00284h-1.66718zm0 10.00288h-1.66714v-6.66859h1.66714z"/><path d="m17.835 8.66571h-1.6671v-3.33429h-1.6671v3.33429h-1.6672v5.83499h1.6672v4.1679h1.6671v-4.1679h1.6671zm-1.6671 4.16789h-1.6671v-2.5008h1.6671z"/></g></svg>
\ No newline at end of file
+<svg width="24" height="24" viewBox="3 2 18 20" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <path fill="currentColor" d="m9.49932 5.33142h-1.66714v1.66714h-1.66714v10.00284h1.66714v1.6672h1.66714v-1.6672h1.66718v-10.00284h-1.66718zm0 10.00288h-1.66714v-6.66859h1.66714z"/>
```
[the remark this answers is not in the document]
vinnyhoward: This was intended, the previous icon was too small

app/components/Views/Wallet/index.tsx
```diff
@@ -1016,6 +1041,24 @@ const Wallet = ({
     selectHomepageSectionsV1Enabled,
   );
 
+  const { variantName: discoveryTabsVariantName } = useABTest(
+    HUB_PAGE_DISCOVERY_TABS_AB_KEY,
+    HUB_PAGE_DISCOVERY_TABS_VARIANTS,
+  );
+
+  const isDiscoveryTabsTreatment = false;
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [c7e8605](https://github.com/MetaMask/metamask-mobile/pull/29193/commits/c7e86055ce4b66ff0b9559b6f08af16c3e08d5a5)

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+            <DiscoveryTabView
+              tabLabel="Predictions"
+              tabIcon={IconName.Predict}
+              keepMounted={false}
+            >
+              <PredictPreviewSheetProvider>
+                <PredictFeed hideHeader onHeaderHiddenChange={animateIcons} />
+              </PredictPreviewSheetProvider>
+            </DiscoveryTabView>
```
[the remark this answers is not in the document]
vinnyhoward: This was intentional. The predict screen has its own tabs and when you scroll down the collapses the new home tabs while preserving the header. This is the original behavior and we are going to preserve that.

caieu (dismissed):

app/components/UI/Perps/Views/PerpsHomeView/PerpsHomeView.tsx
```diff
@@ -124,6 +142,38 @@ const PerpsHomeView = () => {
...
+    (scrollY: number, viewportHeight: number) => {
+      handleScroll({
+        nativeEvent: {
+          contentOffset: { x: 0, y: scrollY },
+          layoutMeasurement: { width: 0, height: viewportHeight },
+        } as NativeScrollEvent,
+      });
+    },
+    [handleScroll],
+  );
```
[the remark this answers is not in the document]
gambinish: This might be worth looking into

app/components/UI/Perps/Views/PerpsHomeView/PerpsHomeView.tsx
```diff
@@ -72,7 +74,23 @@ import PerpsNavigationCard, {
...
+  /** Forwarded to useDiscoveryScrollManager to sync icon animations with header hide/show. */
+  onHeaderHiddenChange?: (hidden: boolean) => void;
+}
+
+const PerpsHomeView = ({
+  hideHeader = false,
+  walletHeaderTranslateY,
+  walletHeaderHeight = 0,
+  tabEnterCallbackRef,
+  onHeaderHiddenChange,
```
gambinish, line 92: nit: It would be nice if we didn't need to pass these props. I wonder if it would be cleaner to extract these into a hook.

Maybe the exception would be the `tabEnterCallbackRef`

Non-blocking though, it seems to be pretty stable as is

app/components/UI/Perps/Views/PerpsHomeView/PerpsHomeView.tsx
```diff
@@ -124,6 +142,38 @@ const PerpsHomeView = () => {
...
+  // Expose onTabEnter to the parent so it can restore this tab's header state on switch.
+  useEffect(() => {
+    if (tabEnterCallbackRef) {
+      tabEnterCallbackRef.current = perpsOnTabEnter;
+      return () => {
+        tabEnterCallbackRef.current = null;
+      };
+    }
+    return undefined;
+  }, [tabEnterCallbackRef, perpsOnTabEnter]);
```
gambinish, line 175: Not against having these here, but my first though is that these are conceptually separate from the business logic. One thing that came to mind as a possible cleanup:

You could create a component that lives at the top of the tsx, something like: `<PerpsTabDiscoverScrollHelper/>` that houses these callback.

This could help keep the layout related effects separate from the business logic

app/components/UI/Perps/Views/PerpsHomeView/PerpsHomeView.tsx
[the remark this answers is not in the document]
vinnyhoward: Cursor's cousin says its not a valid issue:

> Not a valid bug. `usePerpsHomeSectionTracking` only destructures `contentOffset` and `layoutMeasurement.height` from the scroll event (line 101-103) — it never reads `contentSize` or `layoutMeasurement.width`. The synthesized event provides both of those correctly (`contentOffset` comes from the real event, `layoutMeasurement.height` is `viewportHeight`). The zero width and missing `contentSize` are unused fields.

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -0,0 +1,368 @@
...
+    'rgba(25, 0, 102, 0.9)',
+    'rgba(25, 0, 102, 0.2)',
+    'transparent',
+  ],
+  [TAB_INDEX.PREDICTIONS]: [
+    'rgba(61, 6, 95, 0.9)',
+    'rgba(61, 6, 95, 0.2)',
+    'transparent',
+  ],
+};
```
gambinish, line 58: I'm guessing we don't have design system values for these?

gambinish (dismissed):
I'm tentatively approving this as is, but would like to do more testing on my end. It appears that there could be some different behaviors between dev builds and rc builds

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: No we don't, if the test is successful we plan on working with the design and design system teams to refine this component and its values. The DS team is aware of this

caieu (dismissed):

app/component-library/components-temp/Tabs/hooks/useTabsBarLayout.ts
```diff
@@ -0,0 +1,254 @@
...
+
+          if (containerWidth > 0) {
+            const totalWidth = tabLayouts.current.reduce(
+              (sum, l) => sum + (l?.width || 0),
+              0,
+            );
+            const gapsWidth = (tabs.length - 1) * 24;
+            const shouldScroll = totalWidth + gapsWidth > containerWidth - 32;
+            setScrollEnabled(shouldScroll);
+          }
```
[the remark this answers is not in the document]
vinnyhoward: This is fixed and added the `!fillWidth` guard to `handleTabLayout`

gambinish (approved):

caieu (approved):

wachunei (approved):

PR #29195, opened 2026-04-22
feat: React Native Upgrade / 0.81.5 

adnxy:
## **Description**

  Upgrades the core mobile stack from React Native 0.76.9 to 0.81.5,
  React 18.3 to 19.1, and Expo SDK 52 to 54. Includes all necessary
  migration fixes for breaking changes in the new architecture, test
  infrastructure updates, and dependency alignment.

### Core stack
- **React Native** 0.76.9 → 0.81.5
- **React** 18.3.1 → 19.1.0 (and `react-test-renderer` to match)
- **Expo SDK** 52 → 54 (every `expo-*` package bumped to its SDK 54 line)
- **Hermes** updated alongside RN 0.81
- **`@react-native-community/cli` (+ `cli-platform-android`/`-ios`)** 15.0.1 → 20.0.0
- **`@react-native/metro-config`** 0.76.9 → 0.81.5
- **`@types/react`** ^18.2.6 → ^19.1.0
- **iOS app entry point migrated from Objective-C (`main.m`) to Swift (`AppDelegate.swift`)** as required by the RN 0.81 iOS template

### React Native ecosystem libraries
- **`react-native-screens`** 3.37.0 → 4.16.0 (major)
- **`react-native-safe-area-context`** ^5.4.0 → ~5.6.0
- **`react-native-reanimated`** ^3.17.2 → 3.19.0
- **`react-native-gesture-handler`** ^2.25.0 → ~2.28.0
- **`react-native-mmkv`** ^3.2.0 → ^4.1.2 (major; resolves to 4.3.1)
- **`react-native-pager-view`** ^6.7.0 → 6.9.1
- **`react-native-nitro-modules`** ^0.29.6 → 0.35.5
- **`react-native-svg`** ^15.11.1 → 15.12.1
- **`react-native-video`** ^6.10.1 → ^6.19.0
- **`react-native-vision-camera`** ^4.6.4 → ^4.7.3
- **`react-native-view-shot`** ^3.1.2 → 4.0.3 (major; legacy `patch-package` patch dropped)
- **`react-native-qrcode-svg`** 5.1.2 → 6.3.21 (major; legacy patch dropped)
- **`react-native-performance`** ^5.1.2 → ^6.0.0 (major; legacy patch dropped)
- **`react-native-release-profiler`** ^0.4.0 → ^0.4.4
- **`@react-native-community/slider`** ^4.4.3 → 5.0.1 (major)
- **`@react-native-async-storage/async-storage`** ^1.23.1 → 2.2.0 (major)
- **`@react-native-community/viewpager`** new direct dep (patched) to keep legacy paged-view consumers building under RN 0.81
- **`lottie-react-native`** 6.7.2 → ~7.3.1 (major)
- **`rive-react-native`** 9.3.4 → ^9.8.0 (legacy patch dropped)

### Native / platform integrations
- **`@sentry/react-native`** ~6.15.0 → ~7.2.0
- **`@segment/analytics-react-native`** ^2.20.3 → ^2.21.4
- **`@solana-mobile/mobile-wallet-adapter-protocol`** ^2.2.5 added (replaces the removed Solana shim path)

### Tooling / test infra
- **`detox`** ^20.35.0 → ^20.43.0
- **`@expo/fingerprint`** ^0.15.0 → ~0.15.4
- **`pretty-format/react-is`** pinned to ^19.0.0 (resolution; required for React 19 test compat)
- New global mock: `tests/module-mocking/sentry/react-native.ts` (E2E-only Sentry shim selected by `metro.config.js` when `IS_TEST=true`)
- `tsconfig.json` path aliases for selected `@metamask/*` dist subpaths

### React 19
- Removed `defaultProps` from all function components (deprecated in React 18.3, removed in 19)
- `useRef` now requires an explicit initial value — every `useRef<T>()` updated to `useRef<T>(null)`
- `RefObject<T>` typing tightened — call sites switched to `RefObject<T | null>` where they were assuming non-null
- `React.cloneElement` generics tightened — added explicit type parameters where inference no longer narrows
- `ReactChild` removed — switched to `ReactNode`
- Migrated hook tests from `@testing-library/react-hooks` (unmaintained on React 19) to `@testing-library/react-native` `renderHook`
- Updated mocked ref parameters from `{}` to `undefined` to match the new `forwardRef` signature
- Updated test `rerender()` call sites to the new return shape

### React Native 0.81
- Added a `BackHandler.removeEventListener` shim (the static method was removed; `addEventListener` now returns a subscription)
- Migrated iOS entry point to **Swift `AppDelegate.swift`** + new `LaunchScreen.xib` (deleted `main.m`)
- `MMKV` is now a type-only export and `.remove()` was renamed to `.delete()` (`react-native-mmkv` v4)
- `NavigationContainerRef` requires a generic parameter (`@react-navigation/native` v7)
- `BlurEvent` / `FocusEvent` typing changes on `TextInput` callbacks
- `refreshControl` prop typing on virtualized lists
- New `MessageEvent` (private) usage in `WebSocket` is internal — Snaps continue to boot via the existing `patches/@metamask+post-message-stream+10.0.0.patch` (no shim needed)
- Patched `@metamask/react-native-webview`, `@metamask/react-native-button`, `@metamask/react-native-payments`, `@metamask/react-native-acm` for RN 0.81 / React 19 compatibility (see TODO list below)
- Fixed `expo-modules-core` `EventEmitter` mock, `Linking` mock path (now `.default.openURL`), safe-area / screens snapshots

### Snaps
- `AbstractExecutionService` renamed to `ExecutionService` in `@metamask/snaps-controllers`
- Allowed `XMLSerializer` and `DOMParser` inside the snap iframe via the new yarn patch on `@metamask/snaps-execution-environments` (replaces the runtime HTML mutation we previously carried in `SnapsExecutionWebView.tsx`)
- Restored upstream `pump` usage in snap/background bridges (rolled back local stream-cleanup defensive code in `MobilePortStream.js` and `execution-service-init.ts`)

### Test infrastructure
- Fixed wallet adapter tests (dynamic import race with `@expensify/react-native-wallet`)
- Restored the `isClosingRef` guard in `BottomSheetDialog` to prevent double `goBack` (MUSD-406) and rolled back a faulty `lodash.debounce` attempt
- Fixed `PolymarketProvider` fetch mocks (missing `response.text()`)
- Fixed `useTokenSearch` throttle timing, `EarnLendingWithdrawal` mock ordering
- Updated `initial-background-state.json` fixture, `useTailwind` mock, `reanimatedLogger` config
- E2E: fixed URL editor dismissal and snap error assertions across platforms; raised default Detox timeout

### Build / CI
- iOS `Podfile.lock` regenerated for RN 0.81's new Pod graph
- iOS `PrivacyInfo.xcprivacy` updated for the SDK 54 surface
- Android: `CMAKE_VERSION` pinned for RN 0.81's NDK build
- E2E CI workflows append `${{ github.run_id }}` to native-build cache keys (Android APKs/Gradle, iOS DerivedData) to force fresh builds during the upgrade rollout
- `.depcheckrc.yml` ignores extended for short-name Babel plugins

## Dependencies Patched (`.yarn/patches/`)

**MetaMask-owned (see TODO below — should be upstreamed and the patches dropped):**
- `@metamask/react-native-acm@1.2.0`
- `@metamask/react-native-button@3.0.0`
- `@metamask/react-native-payments@2.0.2`
- `@metamask/react-native-webview@14.6.0`
- `@metamask/snaps-execution-environments@11.0.2`

**Third-party (need upstream PRs or version bumps as they ship RN 0.81-compatible releases):**
- `react-native@0.81.5` (RN itself — small follow-up patch on top of 0.81.5)
- `@react-native-community/viewpager@3.3.0`
- `@braze/react-native-sdk@19.1.0` (ReactModuleInfo signature change)
- `react-native-aes-crypto-forked` (we maintain the fork; can be folded into the fork repo)
- `react-native-fast-crypto@2.2.0`
- `react-native-gzip@1.1.0`
- `react-native-i18n@2.0.15` (deprecated upstream — long-term replace)
- `react-native-os@1.2.6`
- `react-native-sensors@5.3.0`
- `reactotron-core-client@2.9.7`
- `expo-web-browser@14.0.2`

**Legacy `patches/` (patch-package) entries removed in this PR**
because the upstream packages or our own forks now cover them, or
because the patched version is no longer installed:
`@metamask+react-native-button+3.0.0`,
`@metamask+react-native-payments+2.0.0`,
`react-native+0.76.9`,
`react-native-aes-crypto-forked+1.2.1`,
`react-native-fast-crypto+2.2.0`,
`react-native-performance+5.1.2`,
`react-native-qrcode-svg+5.1.2`,
`react-native-view-shot+3.8.0`,
`expo-updates-npm-0.27.4`. Net result: no orphan patches carried forward.

## TODO — follow-ups after this PR is merged

Several Yarn patches added by this upgrade target packages we own under
`@metamask/*`. They unblock RN 0.81 today, but we should publish fixed
releases and drop the local patches in follow-up PRs:

- **`@metamask/react-native-acm@1.2.0`** — RN 0.81's Kotlin rewrite of
  `ReactContextBaseJavaModule` removed the `currentActivity` synthetic
  accessor (now `getCurrentActivity()`) and made `onNewIntent(intent: Intent)`
  non-nullable. Patch updates the 4 affected sites in `GoogleAcmModule.kt`.
  → Publish a 1.x compatible release that compiles against RN 0.81.
- **`@metamask/react-native-button@3.0.0`** — `TouchableOpacity.propTypes`,
  `Text.propTypes` and `ViewPropTypes.style` no longer exist on RN 0.81 /
  React 19. Patch falls back to `PropTypes.any`/`PropTypes.bool` defensively.
  → Replace `propTypes` with TypeScript types or PropTypes.shape literals
  upstream.
- **`@metamask/react-native-payments@2.0.2`** — Bumps Android
  `compileSdk`/`buildTools`/`targetSdk` from 28 → 36 and drops the removed
  `ReactBridge` import. → Publish a release with the SDK bump (and
  optionally migrate the module to Kotlin / TurboModules while we're there).
- **`@metamask/react-native-webview@14.6.0`** — Adds
  `codegenConfig.ios.componentProvider`/`modulesProvider` entries needed by
  RN 0.81's Fabric codegen so `RNCWebView` and `RNCWebViewModule` register
  correctly. → Cherry-pick into the next published release; this is purely
  a `package.json` metadata fix.
- **`@metamask/snaps-execution-environments@11.0.2`** — Adds `XMLSerializer`
  and `DOMParser` to the LavaMoat `scuttleGlobalThis.exceptions` list in
  `dist/webpack/webview/index.html` so snaps can run inside the iframe
  after lockdown. (This replaces the runtime HTML mutation we previously
  carried in `SnapsExecutionWebView.tsx`.) → Land the equivalent change in
  `snaps` upstream so the scuttle exceptions are baked into the published
  HTML.

Other notable patches added here target third-party packages we don't own
(see the "Third-party" list above); they'll need separate upstream PRs or
version bumps as those projects publish RN 0.81-compatible releases.

CHANGELOG entry: null

## **Manual testing steps**

Because this PR moves the entire native floor (RN 0.81 / React 19 / Expo 54),
the surfaces most likely to regress are the ones that cross the JS↔native
bridge or rely on WebViews, animations, and platform integrations. Run the
scenarios below on **both iOS and Android** (release-style build preferred so
LavaMoat lockdown and Hermes kick in).

```gherkin
Feature: RN 0.81 / React 19 / Expo 54 upgrade smoke

  Scenario: In-place upgrade from the latest production build
    Given a device with the latest store version of MetaMask Mobile installed
    And a wallet that has been used (multiple accounts, custom networks, contacts, NFTs, recent txs, push enabled, biometrics enabled)
    When the user installs this build over the existing one (without uninstalling)
    Then the app launches without crashing on cold start
    And redux-persist migrations run cleanly (no migration errors in logs)
    And accounts, custom networks, contacts, hidden tokens, NFT collections, transaction history, and address book entries are preserved
    And MMKV-backed storage opens (no "Failed to open MMKV" or version mismatch errors)
    And biometric unlock still works without re-prompting for a password reset
    And push-notification token registration still resolves
    And no data needs to be re-imported

  Scenario: In-place upgrade from an older minor (n-3 versions back)
    Given a device with a 3-versions-old build of MetaMask Mobile and a used wallet
    When the user installs this build over it
    Then the chained migrations execute in order with no errors
    And the same continuity assertions as above hold

  Scenario: Onboarding — create new wallet
    Given a fresh install
    When the user creates a new wallet, sets a password, and skips backup
    Then onboarding completes and the wallet screen renders without animation glitches
    And the BackHandler shim does not trigger duplicate goBacks on Android

  Scenario: Onboarding — import via SRP
    Given a fresh install
    When the user imports an existing 12-word SRP via "Import using Secret Recovery Phrase"
    Then accounts populate and the home screen loads with balances

  Scenario: Manual backup flow (Steps 1, 2, 3)
    Given a wallet that has not yet completed manual backup
    When the user walks through ManualBackupStep1 → 2 → 3
    Then the SRP is shown, verification succeeds, and Step 3 navigates to HomeNav without re-registering the Android back handler twice

  Scenario: Send ETH (legacy + EIP-1559)
    Given a funded EOA on Sepolia and Mainnet
    When the user sends ETH to another address using both legacy and 1559 gas
    Then the confirmation sheet renders, the user signs, and the tx appears in Activity → confirmed

  Scenario: Token send + ERC-20 approval
    Given a funded EOA with USDC on Mainnet
    When the user sends USDC and (separately) approves a spender via dApp
    Then both flows render their confirmation sheets correctly and the txs land

  Scenario: Swap (single-chain) + Bridge (cross-chain)
    Given a funded EOA
    When the user performs a Swap and a Bridge from the wallet UI
    Then quotes load, approvals (if needed) sign, the user can switch routes, and the txs land

  Scenario: Smart transactions (STX) on Mainnet
    Given an account with STX enabled
    When the user submits a swap eligible for STX
    Then the STX status sheet renders, polls, and resolves to confirmed without flicker

  Scenario: Transaction batching (EIP-7702 / 5792)
    Given a smart account that supports batched txs
    When the user signs a batch confirmation
    Then the batch appears in Activity with the correct sub-txs

  Scenario: Hardware wallet — Ledger BLE
    Given a Ledger device paired via BLE
    When the user signs an EVM transaction using the Ledger account
    Then the BLE connection succeeds, the user approves on-device, and the tx is broadcast

  Scenario: WalletConnect / SDK dApp connection
    Given a sample dApp (e.g. Uniswap testnet) on desktop
    When the user scans the QR with the in-app camera and connects via WalletConnect
    Then the connection request renders, methods (eth_signTypedData_v4, personal_sign, eth_sendTransaction) work end-to-end, and the dApp shows the wallet address

  Scenario: Deeplinks — universal links + custom scheme + cold/warm start
    Given the app is killed (cold) and separately, in the background (warm)
    When the user taps:
      | metamask://send?recipient=0x...                          | from another app |
      | metamask://buy?address=0x...&chainId=0x1                 | from another app |
      | metamask://swap?from=ETH&to=USDC                         | from another app |
      | metamask://wc?uri=wc:abcd...                             | from a dApp QR  |
      | https://metamask.app.link/wallet                         | universal link  |
      | https://metamask.app.link/buy?...                        | universal link  |
    Then each link routes to the correct screen on both cold and warm starts
    And in-app `Linking.openURL(...)` (e.g. ramp Buy widget, support links) opens the external browser correctly

  Scenario: In-browser dApp via the built-in Browser tab
    Given the wallet is unlocked
    When the user opens the in-app Browser, navigates to a dApp, and signs a tx
    Then the WebView renders, the provider injects, and signing flows work
    # exercises the @metamask/react-native-webview Fabric codegen patch

  Scenario: Snaps — install + invoke
    Given a Flask build with snaps enabled
    When the user installs a permissioned snap and invokes a method that opens a snap UI dialog
    Then the snap iframe loads under LavaMoat lockdown without "XMLSerializer is not a function" or scuttle errors and the dialog renders
    # exercises the @metamask/snaps-execution-environments LavaMoat patch

  Scenario: Buy/Sell ramps (in-app + external browser)
    Given a wallet unlocked and a quote available
    When the user purchases via Transak (in-app widget) and via an external-browser provider
    Then both routes complete; on the external path Linking.openURL or InAppBrowser is invoked correctly and Activity reflects the precreated order

  Scenario: Apple Pay / Google ACM
    Given the user has an Apple Pay card (iOS) or a Google account on device (Android)
    When the user triggers a flow that hits @metamask/react-native-payments (iOS) and Google ACM sign-in (Android)
    Then both native sheets present, complete, and return values to JS without crashing
    # exercises the @metamask/react-native-payments and @metamask/react-native-acm patches

  Scenario: Biometrics / Keychain unlock
    Given a wallet protected by biometrics
    When the user backgrounds the app, returns, and unlocks
    Then Face ID / Touch ID / Android biometric prompts present and unlock works on cold start, warm resume, and after switching networks

  Scenario: Push notifications (transaction notifications)
    Given push notifications enabled
    When an inbound tx confirms
    Then a notification arrives, deeplink tap-through navigates to the right activity row, and Sentry breadcrumbs do NOT show "[E2E Sentry Mock]" (verifies dev/prod env separation)

  Scenario: Token detail chart (TradingView WebView)
    Given the user opens a token detail (e.g. ETH) with MM_CHARTING_LIBRARY_URL set
    When the chart loads
    Then the TradingView library loads from CDN, the OHLC data renders, the crosshair pills work, and there is no "Failed to load TradingView library" error

  Scenario: BottomSheets, modals, and gestures
    Given any flow with a bottom sheet (account selector, network picker, confirmation, settings)
    When the user opens a sheet and rapidly taps the overlay 3-5 times
    Then the sheet closes once, the screen behind is unchanged (no double goBack — MUSD-406 regression check), and the swipe-to-dismiss gesture also works

  Scenario: Animations and Reanimated surfaces
    Given the user navigates onboarding → wallet → token → bridge → settings
    When the fox loader, onboarding success animation, carousel cards, and Rive-driven loaders play
    Then all animations render at 60fps with no warning spam and no crash on iOS 18 / Android 14+

  Scenario: Background restoration after long suspend
    Given the app is backgrounded for >15 minutes
    When the user resumes
    Then accounts, balances, and websocket connections (price feed, notifications) reconnect without leaking timers or duplicate listeners

  Scenario: SDK / WebSocket reconnection
    Given an active dApp connection and an active price-feed websocket
    When the user toggles airplane mode off→on→off
    Then both connections recover and txs/quotes flow again

  Scenario: Settings → Reveal SRP / Private Key
    Given an unlocked wallet
    When the user reveals the SRP and a private key from settings
    Then biometrics gate works, the secrets render correctly, and the "Hold to reveal" + scroll-on-iPhone-SE behaviors are unchanged

  Scenario: Power-user wallet (many accounts / tokens / NFTs)
    Given a power-user SRP imported (see internal SRPs)
    When the user navigates Wallet, Activity, NFTs, and switches networks
    Then list rendering, scrolling, and network switching remain smooth (FlashList 2 / RN 0.81 perf check)
```

The `[E2E Sentry Mock]` log line should NOT appear in any of the dev/release
builds above; if it does, `IS_TEST=true` is leaking into the bundler env.

- [ ] I've tested on Android

---

> [!NOTE]
> **High Risk**
> High risk because it updates core React Native/Android build toolchain and introduces multiple Yarn patches to native dependencies, which can affect app startup, native compilation, and E2E reliability across both platforms.
> 
> **Overview**
> Upgrades the RN stack to `react-native@0.81.5` and aligns native build tooling around it: Android moves to `compileSdk/targetSdk 36`, Kotlin `2.1.20`, NDK `27.1`, Gradle `8.14.3`, updates app initialization to `loadReactNative(this)`, and adds an MLKit dependency override plus Detox flavor handling.
> 
> Stabilizes CI/E2E builds during the upgrade by forcing fresh native caches (cache keys include `${{ github.run_id }}`), pinning CMake for Android E2E builds, bumping iOS cache versions, increasing E2E timeout, and restoring iOS app/framework execute permissions after artifact download.
> 
> Introduces/updates multiple `.yarn/patches/*` to keep third-party and MetaMask native modules building under RN 0.81 (Gradle/SDK bumps, Kotlin/ObjC signature fixes, codegen metadata, safer JS snippets), removes obsolete patches, and adjusts JS/TS code for React 19 typing/runtime changes (BackHandler subscription cleanup, stricter `RefObject<T | null>`/`useRef` init, safer prop spreads/cloneElement generics, default props moved to parameters, and test expectation updates).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3c6e6eaa9290246947bcb21a5d3085609540ba0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tommasini:
@SocketSecurity ignore-all All the package are needed for expo or react native to work

app/selectors/transactionController.ts
```diff
@@ -42,17 +42,22 @@ function matchesTransactionType(
...
-  (transactionControllerState) => transactionControllerState.transactions,
+  (transactionControllerState) =>
+    transactionControllerState?.transactions ?? EMPTY_TRANSACTIONS,
 );
 
 const selectTransactionBatchesStrict = createSelector(
   selectTransactionControllerState,
-  (transactionControllerState) => transactionControllerState.transactionBatches,
+  (transactionControllerState) =>
+    transactionControllerState?.transactionBatches ?? EMPTY_TRANSACTION_BATCHES,
```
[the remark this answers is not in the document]
andrepimenta: Reverted

app/components/UI/Ramp/hooks/useContinueWithQuote.test.ts
```diff
@@ -98,7 +98,7 @@ const mockDeviceIsAndroid = jest.requireMock('../../../../util/device')
   .isAndroid as jest.Mock;
 const mockLinkingOpenURL = jest.requireMock(
   'react-native/Libraries/Linking/Linking',
-).openURL as jest.Mock;
+).default.openURL as jest.Mock;
```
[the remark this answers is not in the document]
andrepimenta: false-positive

app/component-library/components/BottomSheets/BottomSheet/foundation/BottomSheetDialog/BottomSheetDialog.tsx
```diff
@@ -191,6 +189,21 @@ const BottomSheetDialog = forwardRef<
       );
     };
 
+    const onDebouncedCloseDialog = useMemo(
+      // Prevent hide from being called multiple times. Potentially caused by taps in quick succession.
+      () => debounce(onCloseDialog, 2000, { leading: true }),
+      [onCloseDialog],
+    );
```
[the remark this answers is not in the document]
andrepimenta: fixed

aganglada (dismissed):

tests/smoke/snaps/test-snap-background-events.spec.ts
```diff
@@ -120,20 +120,16 @@ describe(SmokeSnaps('Background Events Snap Tests'), () => {
...
-            'scheduleBackgroundEventResultSpan',
-            'Cannot schedule an event in the past.',
-            { timeout: 30000 },
-          );
-        }
+        // Both iOS and Android surface the snap error as a native WebView
+        // alert (`window.alert`). The previous Android-specific branch read
+        // the error from the inline result span, but post RN 0.81 / new
+        // react-native-webview the Android WebView routes `alert()` through
+        // the native dialog handler too.
```
FrederikBolding: Then we can remove this comment

.yarn/patches/@metamask-snaps-execution-environments-npm-11.0.2-83df58c486.patch
FrederikBolding, line 1: This PR should be blocked until we don't require an impossible to review patch to this package.

app/lib/snaps/SnapsExecutionWebView.tsx
```diff
@@ -10,13 +10,23 @@ import {
...
 const styles = createStyles();
 
+// The bundled `index.html` ships a LavaMoat scuttling allow-list that does
+// not include `XMLSerializer` / `DOMParser`, which several first-party
+// snaps (`bitcoin-wallet-snap`, `tron-wallet-snap`, …) need at startup.
+// We extend that allow-list via a Yarn patch on
+// `@metamask/snaps-execution-environments` (see
+// `.yarn/patches/@metamask-snaps-execution-environments-npm-*.patch`).
+// TODO: drop the patch once the upstream package adds those globals to the
+// scuttling exceptions list.
```
FrederikBolding: This reasoning seems strange, the Snaps have never had access to these globals before.

Is there some change to the underlying WebView here?

FrederikBolding (changes requested):
Blocking on Snaps changes.

app/components/Snaps/SnapUIFooterButton/SnapUIFooterButton.test.tsx
```diff
@@ -61,7 +61,7 @@ describe('SnapUIFooterButton', () => {
   it('shows loading state', () => {
     render(<SnapUIFooterButton {...defaultProps} loading />);
     const button = screen.getByRole('button', { name: 'Test Button' });
-    expect(button.findByType(ActivityIndicator)).toBeTruthy();
+    expect(button.findByType(ActivityIndicator as never)).toBeTruthy();
```
FrederikBolding: ?

tests/smoke/snaps/test-snap-background-events.spec.ts
FrederikBolding: The same goes for all the other E2Es

VGR-GIT (dismissed):

android/build.gradle
```diff
@@ -61,10 +83,6 @@ subprojects { subproject ->
 				force "org.jetbrains.kotlin:kotlin-stdlib:${brazeKotlinStdlibVersion}"
 				force "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${brazeKotlinStdlibVersion}"
 				force "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${brazeKotlinStdlibVersion}"
```
baptiste-marchand, line 82: Considering the above comment, can we go further and remove the whole `subprojects` block?
It's not a blocker though, we can polish it in a follow up commit

ios/MetaMask/AppDelegate.swift
baptiste-marchand, line 1: Shouldn't AppDelegate.m be deleted too?

baptiste-marchand (dismissed):

wachunei (dismissed):

infiniteflower (dismissed):

christopherferreira9 (dismissed):
Approving for QA to unblock but if we have the chance it would be good to avoid the extensive comments related to the upgrade in spec files/page-objects otherwise we can clean up afterwards.

wachunei (dismissed):

Cal-L (dismissed):
LGTM

georgewrmarshall (dismissed):

caieu (dismissed):

brianacnguyen (dismissed):

jeffsmale90 (dismissed):
Delegators are go 🐊🚀

javiergarciavera (dismissed):

OGPoyraz (dismissed):

ccharly (dismissed):
LGTM for accounts (new patterns seem consistent with other similar files)!

bergarces (dismissed):
Ok from @MetaMask/metamask-assets

ieow:
lgtm on behalf metamask/onboarding

ieow (dismissed):

.yarn/patches/react-native-fast-crypto-npm-2.2.0-57b8a8a01a.patch
ccharly, line 1: Compared to the previous patch we had, it seems this one is much much lighter.

We should probably re-import the same changes there and update the SDK version to use similar version than: `.yarn/patches/react-native-aes-crypto-forked-https-9ebec3d485.patch`?

Or was this causing some compilation issues?

NOTE: The indentation on this patch is completely off too!

.yarn/patches/react-native-aes-crypto-forked-https-9ebec3d485.patch
ccharly, line 1: Nit: any reason why the patch is named that way? we usually have the `-npm-<version>-<checksum>.patch` suffix?

ccharly (changes requested):

.yarn/patches/react-native-fast-crypto-npm-2.2.0-57b8a8a01a.patch
adnxy: I just double-check everything in the git history around the patch

This patch has only ever included the register keyword removal (commit #883ec0c51a) there wasn't any heavier previous version for this package.         

Are you thinking of the react-native-aes-crypto-forked patch? This is a different package that required SDK/Gradle changes. 

react-native-fast-crypto doesn’t need those changes. It already uses implementation (not compile), and its SDK versions are inherited from the root project via safeExtGet().

The only compilation issue here was the C++ register storage class specifier (removed in C++17), which breaks with newer NDK/clang in RN 0.81.

Good catch on the indentation, I’ll fix that.

.yarn/patches/react-native-aes-crypto-forked-https-9ebec3d485.patch
adnxy: This package is installed from a GitHub URL (not npm registry), so Yarn auto-generates the patch filename based on the URL hash instead of the usual npm-<version>-<checksum> pattern. Not something we named manually.

.yarn/patches/react-native-fast-crypto-npm-2.2.0-57b8a8a01a.patch
ccharly: Sorry, I meant this patch: https://github.com/MetaMask/metamask-mobile/blob/0b8d63871a09ac8dad7e9a4392a186c0d57c1cd0/patches/react-native-fast-crypto%2B2.2.0.patch#L1-L121

But maybe this package is correctly setup and the patch we have on `main` was already wrong? There were some CMake patches as well as gradle

adnxy: You're right, the old patch-package patch (patches/react-native-fast-crypto+2.2.0.patch) had Gradle, CMake, and SDK changes too.
  
When we migrated to Yarn patches, only the register fix was carried over. The Gradle/CMake changes were dropped during that migration. 

I'll check if we need to bring them back: safeExtGet() handles SDK versions, but the CMake and Gradle plugin changes might still be needed. I'll double-check and get back to you! Y

adnxy:
## react-native-fast-crypto patch — updated

Addressed all review feedback:

1. **Re-imported Gradle/CMake/SDK changes** from the old `patch-package` patch that were dropped during Yarn patch migration:
   - Gradle plugin `3.6.0` → `8.1.0`
   - Compile/Build SDK `28` → `34`, minSdk `19` → `24`
   - Removed old AGP < 4.0 `jniLibs` workaround
   - Modernized CMakeLists.txt (project declaration, `target_include_directories`, quoted paths, ARM64 compile options)

2. **Fixed duplicate `CMAKE_CXX_STANDARD`** — removed the C++17 line, keeping only C++14 (the code uses no C++17 features; the `register` keyword removal is the only compilation fix needed)

3. **Fixed indentation** — `register` keyword removal lines now have correct leading whitespace

Verified: `yarn install` applies the patch cleanly. Rebuilding iOS and Android locally to confirm compilation.

.yarn/patches/react-native-fast-crypto-npm-2.2.0-57b8a8a01a.patch
adnxy: @ccharly  Done in **c6101dc**

Re-imported all Gradle/CMake/SDK changes from the old patches/react-native-fast-crypto+2.2.0.patch, fixed the indentation.. I verified everything locally..

Please let me know if you notice anything else

package.json
```diff
@@ -331,7 +335,7 @@
     "@metamask/slip44": "^4.2.0",
     "@metamask/smart-transactions-controller": "^24.0.0",
     "@metamask/snaps-controllers": "^20.0.3",
-    "@metamask/snaps-execution-environments": "^11.0.2",
+    "@metamask/snaps-execution-environments": "11.0.2",
```
FrederikBolding:
```suggestion
    "@metamask/snaps-execution-environments": "^11.0.2",
```
Shouldn't be needed anymore

babel.config.js
```diff
@@ -69,32 +134,30 @@ module.exports = {
...
-      test: [
-        './node_modules/**/@metamask/rpc-errors/**',
-        './node_modules/@metamask/rpc-errors/**',
-      ],
+      test: (f) => !!f?.includes('/node_modules/@metamask/rpc-errors'),
       plugins: [['@babel/plugin-transform-classes', { loose: true }]],
     },
     {
-      test: './app/lib/snaps',
+      test: (f) => !!f?.includes('/app/lib/snaps'),
```
FrederikBolding: This doesn't seem like it gives the same guarantees as the previous configuration?

FrederikBolding: If this is the new pattern, perhaps this would be safer?
```suggestion
      test: (f) => !!f?.includes('/app/lib/snaps/SnapsExecutionWebView.tsx'),
```

FrederikBolding (dismissed):

jiexi (dismissed):

georgewrmarshall (dismissed):

Matt561 (dismissed):
Approving for Earn. I've successfully tested the pooled-staking, stablecoin lending, mUSD conversion, and Money Hub flows.

ccharly (dismissed):
LGTM for accounts, we have re-checked the crypto patches with @danroc and @gantunesr, the only "minor" thing that's different is the `minSdkVersion 19` that used to be `minSdkVersion 21` for the `aes-crypto-forked` patch.

Was (21):
```diff
-        minSdkVersion 19
-        targetSdkVersion 22
+        minSdkVersion 21
+        targetSdkVersion 33
```

Now (is not patched anymore, and is set to 19):
```diff
         minSdkVersion 19
-        targetSdkVersion 22
+        targetSdkVersion 36
```

That should be fine. But feel free to fix the patch again if you wish!

tommasini:
https://socket.dev/dashboard/org/MetaMask/diff-scan/9ebfed4b-e986-4ab8-a5f9-bce06c038588/alert/Q0MAsuJrCWDqvq2spRorVjP02TsC8rCgL3Pwcu1LcEvQ -> `lan-network` 0.1.7 maintained by a expo tool contributor - https://github.com/kitten/lan-network

tommasini:
gem/benchmark@0.5.0 -> high score in [socket.io ](https://socket.dev/dashboard/org/MetaMask/diff-scan/29631f02-8c5b-4e16-96a9-5e4825dba6e1/alert/QwIuNQOfFES6KcFhtr7aX2PISCR-go8Uoho_lmStsYD0)
Shell access identified is part of the developers workflow https://github.com/ruby/benchmark/blob/70c5aeb5c6faa15e67928ae083d250f9cc4c1d6d/bin/setup#L6
Should be harmless

naugtur (dismissed):
approving on behalf of @MetaMask/supply-chain 

future use of `socket ignore-all` won't be approved and will require closing and opening a new PR

what I looked at:
- publisher changes in dependencies
- new dependencies with shell access - do they really use it and need it
- all changes to package.json
- patches

package.json
```diff
@@ -458,10 +463,10 @@
     "qs": "6.14.1",
     "query-string": "^6.12.1",
     "randomfill": "^1.0.4",
-    "react": "18.3.1",
-    "react-native": "patch:react-native@patch%3Areact-native@npm%253A0.76.9%23~/.yarn/patches/react-native-npm-0.76.9-1c25352097.patch%3A%3Aversion=0.76.9&hash=9739da#~/.yarn/patches/react-native-patch-d76d50a92f.patch",
+    "react": "19.1.0",
+    "react-native": "patch:react-native@npm%3A0.81.5#~/.yarn/patches/react-native-npm-0.81.5-d8232ef145.patch",
     "react-native-aes-crypto": "3.0.3",
-    "react-native-aes-crypto-forked": "MetaMask/react-native-aes-crypto-forked#397d5db5250e8e7408294807965b5b9fd4ca6a25",
+    "react-native-aes-crypto-forked": "patch:react-native-aes-crypto-forked@https%3A//github.com/MetaMask/react-native-aes-crypto-forked.git%23397d5db5250e8e7408294807965b5b9fd4ca6a25#~/.yarn/patches/react-native-aes-crypto-forked-https-9ebec3d485.patch",
```
naugtur, line 470: warning: This will no longer generate warnings from `git-safe-dependencies` if the hash is updated to one that doesn't belong to the original repo, because the patch syntax is getting in the way.
the patch (or preferably all git dependencies) should be eliminated. (not a blocker)

package.json
```diff
@@ -530,7 +535,7 @@
     "redux-saga": "^1.3.0",
     "redux-thunk": "^2.4.2",
     "reselect": "^5.1.1",
-    "rive-react-native": "patch:rive-react-native@npm%3A9.3.4#~/.yarn/patches/rive-react-native-npm-9.3.4-8082feca90.patch",
+    "rive-react-native": "^9.8.0",
```
naugtur, line 539: the patch file was not removed. is that intentional?

package.json
```diff
@@ -739,6 +744,7 @@
       "chromedriver": false,
       "detox": true,
       "detox>bunyan>dtrace-provider": false,
+      "eciesjs>secp256k1": true,
```
naugtur, line 747: legitimate

package.json
```diff
@@ -755,6 +761,7 @@
       "@metamask/sdk-communication-layer>utf-8-validate": false,
       "detox>ws>bufferutil": false,
       "@metamask/notification-services-controller>firebase>@firebase/firestore>@grpc/proto-loader>protobufjs": false,
+      "@metamask/sdk-communication-layer>eciesjs>secp256k1": false,
```
naugtur, line 764: any chance we could deduplicate that?

package.json
```diff
@@ -728,7 +733,7 @@
       "@sentry/react-native>@sentry/cli": true,
       "@storybook/manager-webpack5>@storybook/core-common>webpack>watchpack>watchpack-chokidar2>chokidar>fsevents": false,
       "@storybook/addon-controls>@storybook/core-common>esbuild": false,
-      "appium-adb>@appium/support>sharp": false,
+      "appium-adb>@appium/support>sharp": true,
```
naugtur, line 736: legitimate build script. Why was it not needed before?

package.json
adnxy: No, that wasn't intentional, good catch! The patch was for 9.3.4 and is no longer needed with ^9.8.0. I'll remove it.

package.json
tommasini: Awesome catch! We didn't update appium, so I'm assuming that this is out of scope, will talk with the team

adnxy: The dependency upgrade bumped @appium/support from v4 to v6. v4 didn't depend on sharp at all, so the    allowScripts entry was a no-op. v6 added sharp (as an optional dep), so now the install script actually needs to be allowed to run.    

 @appium/support@4.1.6 - no sharp in its dependency list                                                 
 @appium/support@6.1.1 - has sharp: "npm:0.34.2" with dependenciesMeta: sharp: optional: true

package.json
adnxy: done, removed! ty

package.json
adnxy: Already checked and packages seem to be deduplicated. The separate allowScripts entry is required by LavaMoat's policy resolution. I can look into consolidating further as a follow-up

package.json
adnxy: Acknowledged, the patch updates Android SDK to 36 and fixes iOS header search paths for RN 0.81. Will address this as a follow-up.

andrepimenta (approved):

PR #29247, opened 2026-04-23
ci: reuse native E2E builds across commits and PRs

tommasini:
## **Description**

Replaces the branch-keyed `cirruslabs/cache` reuse layer for native E2E builds with a content-addressable cross-run lookup, driven by a `build-source-hash` commit status that is computed once on a stable runner.

### How it works

```mermaid
flowchart TD
    A["post-build-source-hash<br/>(ubuntu-latest, pinned to PR head SHA)<br/>fingerprint = yarn fingerprint:generate"] -->|posts build-source-hash status<br/>+ exposes as job output| B["Build iOS / Android job"]
    B --> C{"check-force-builds<br/>label OR [force-builds] in commit?"}
    C -- yes --> H["Heavy native-build path<br/>setup-e2e-env + Gradle/Xcode + compile"]
    C -- no --> D["find-reusable-build<br/>(GitHub Actions REST API)"]
    D --> D1["Tier 1: same-branch runs"]
    D1 -- miss --> D2["Tier 2: base-branch (main) runs"]
    D2 -- miss --> D3["Tier 3: cross-PR runs<br/>(event=pull_request, no branch filter)"]
    D1 -- hit --> E["actions/download-artifact@v4<br/>by run-id"]
    D2 -- hit --> E
    D3 -- hit --> E
    D3 -- miss --> H
    E --> F["Lean repack path<br/>Node + yarn install + repack:{ios,android}"]
    H --> I["Upload .app / release.apk / androidTest.apk"]
    F --> I
    I -.->|feeds future runs| D
```

The probe runs **before** any heavy setup. On a hit, iOS skips Ruby/Bundler/CocoaPods/`pod install`; Android skips `setup-e2e-env`/Gradle. Only the JS bundle is rebuilt and re-packed into the cached native shell.

### Benefits vs the previous `cirruslabs/cache` layer

| | Before | After (this PR) |
|---|---|---|
| **Reuse keyspace** | `${ref_name}` baked into key → every PR isolated; only `main → PR` fallback crossed branches | Content-addressable by fingerprint; any PR can reuse any other PR's build |
| **When the lookup runs** | *After* `setup-e2e-env` (cache key needs `node_modules` to compute fingerprint) | Before any heavy setup |
| **Fingerprint stability** | Recomputed per runner → drifts across macOS / Linux build / Linux CI runners; shifts on every `main` push | Computed once on `ubuntu-latest`, pinned to `pull_request.head.sha` |
| **iOS reuse-hit wall time** | ~6m setup + ~5m repack | ~30-60s setup + ~1-2m repack |
| **Android reuse-hit wall time** | ~5m23s | ~1-2m warm / ~2-3m cold-cache |
| **Cold build (no reuse)** | ~20-25m | unchanged |

### What changed

- **`ci.yml`** — new `post-build-source-hash` job emits the canonical fingerprint and exposes it as a job output consumed by both build jobs.
- **New composite actions** — `find-reusable-build` (3-tier scan), `check-force-builds` (label + `[force-builds]` commit-message escape hatch; reads commit message via REST API to survive shallow checkout), `post-build-source-hash` (factored out for reuse).
- **`build-{ios,android}-e2e.yml`** — probe → gate → lean-repack vs heavy-native paths. Lean paths skip the work `yarn build:repack:*` doesn't need.
- **`setup-e2e-env`** — new `install-foundry` input (default `true`, opt-out for build workflows where `yarn setup:github-ci` already runs `install:foundryup`).
- **Repack throughput** — `METRO_MAX_WORKERS` bumped `2 → 6` on the lean path (no Gradle/Xcode competing for RAM).
- **Removed** — branch- and main-scoped `cirruslabs/cache` for `MetaMask.app` / `release.apk` / `release-androidTest.apk`. Gradle, Xcode DerivedData, and `.metamask` caches kept (gated to the heavy path).

### Safety

- Fingerprint job failure or forked PR (no `statuses: write`) → empty `source-fingerprint` → all reuse steps skipped → fresh build runs. Never "no build".
- The same `build-source-hash` status gates the OTA path in `push-eas-update.yml`, so a JS-only OTA cannot ship after native code changes.
- Full architecture, decision tables, and failure-mode catalog: `docs/ci-build-reuse.md` (new in this PR).

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Refs: Prior art in MetaMask extension — https://github.com/MetaMask/metamask-extension/pull/41435

No issue: CI infrastructure improvement; no Jira ticket.

## **Manual testing steps**

```gherkin
Feature: Reusable native E2E builds on CI

  Scenario: Fresh run populates the reuse pool
    Given this branch has no prior completed CI run with a matching fingerprint
    When CI runs on the latest commit
    Then post-build-source-hash posts a build-source-hash commit status on the PR head SHA
    And Build iOS E2E Apps runs setup-e2e-env and the full Xcode native compile
    And Build Android E2E APKs runs setup-e2e-env and the full Gradle build
    And both jobs upload artifacts named ${build_type}-${env}-MetaMask.app, ${build_type}-${env}-release.apk, and ${build_type}-${env}-release-androidTest.apk

  Scenario: Empty commit reuses the prior native build (same-branch tier)
    Given the branch has a prior successful run with non-expired artifacts
    When an empty commit is pushed
    Then post-build-source-hash computes and posts the same fingerprint as before
    And find-reusable-build reports a hit via the same-branch tier
    And Build iOS E2E Apps skips Ruby/Bundler/CocoaPods/Xcode setup and only runs yarn build:repack:ios
    And Build Android E2E APKs skips setup-e2e-env and only runs yarn build:repack:android
    And E2E tests run against the repacked artifacts

  Scenario: Cross-PR reuse finds an unrelated PR with matching fingerprint
    Given PR A on branch feature-a completed a fresh native build and uploaded artifacts
    And PR B on branch feature-b touches only JS/tests/docs (no native-affecting files)
    When PR B's CI runs
    Then find-reusable-build misses the same-branch tier (feature-b runs)
    And misses the base-branch tier (no recent main run with the matching fingerprint)
    And matches PR A's run via the cross-PR tier (event=pull_request, no branch filter)
    And PR B downloads PR A's artifacts by run-id and runs yarn build:repack:*

  Scenario: force-builds label bypasses reuse
    Given a PR that would otherwise hit the reuse path
    When the force-builds label is added and CI re-runs
    Then check-force-builds reports force=true
    And find-reusable-build does not run
    And both iOS and Android run a fresh native compile
    And when the label is removed and CI re-runs, reuse resumes normally

  Scenario: [force-builds] commit tag bypasses reuse under shallow checkout
    Given a PR whose head commit message contains [force-builds]
    When CI runs
    Then check-force-builds reads the commit message via the GitHub REST API (not git show)
    And reports force=true even with actions/checkout's default fetch-depth: 1
    And both iOS and Android run a fresh native compile

  Scenario: Fingerprint job failure degrades gracefully to fresh build
    Given post-build-source-hash fails (e.g. broken fingerprint.config.js, transient yarn install error, or missing statuses: write on a fork)
    When the build workflows run
    Then inputs.source-fingerprint is empty
    And every fingerprint-keyed step is skipped
    And both iOS and Android run a fresh native compile
    And no E2E build is blocked

  Scenario: Metro transform cache persists across repack runs
    Given a reuse hit on the repack path
    When yarn build:repack:{ios,android} runs
    Then Metro uses METRO_CACHE_DIR as its FileStore root
    And the persisted transform cache is honored (no --reset-cache)
    And the E2E build runs with updates disabled (no app.manifest generation)
```

### **Before**

N/A — CI-only change; no UI impact.

### **After**

N/A — CI-only change; no UI impact.

- [x] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Changes CI build/reuse logic and caching paths for iOS/Android E2E artifacts, which could cause unexpected stale/invalid artifacts or missed rebuilds if the fingerprint/status or artifact lookup is wrong. Scoped to CI/workflows, but failures can block or slow builds across PRs.
> 
> **Overview**
> Enables **cross-run, cross-PR reuse** of native E2E build artifacts by introducing a canonical `@expo/fingerprint` published as a `build-source-hash` commit status and using it to locate/download matching artifacts from prior workflow runs.
> 
> Build workflows now **gate between a full native build vs a lean repack path**: they first check a `force-builds` override (PR label or `[force-builds]` commit token), then attempt to find/download reusable artifacts; on a hit they skip heavy setup (Gradle/Xcode) and only run repack + lightweight dependency setup.
> 
> CI wiring is updated to add the `post-build-source-hash` job and pass `source-fingerprint` into iOS/Android build jobs (with added `actions/statuses/pull-requests` read permissions), `setup-e2e-env` gains an `install-foundry` toggle to avoid redundant installs, and repack throughput is tuned (e.g., `METRO_MAX_WORKERS` increased on reuse paths) with clearer guidance when repack detects a broken cached iOS app.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89e6bc59b84493e1bb5598e6c866a08979588db7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fingerprint.config.js
```diff
@@ -8,6 +8,17 @@
 const config = {
   /**
    * Track files and directories under `extraSources` if they affect native code changes.
+   *
+   * Intentionally NOT tracked:
```
Gudahtt: Should `ci.yml` be tracked as well? That's where the build workflows are invoked, and parameters/environment variables passed in.

Gudahtt: Also, this change doesn't seem strictly tied to native E2E build reuse, it could be done in a separate PR

package.json
```diff
@@ -656,6 +656,7 @@
     "lint-staged": "10.5.4",
     "listr2": "^8.0.2",
     "liveline": "0.0.7",
+    "metro-cache": "^0.81.1",
```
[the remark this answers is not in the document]
Gudahtt: This seems like a good point, I'm not sure why it was resolved

metro.config.js
```diff
@@ -9,6 +9,8 @@
 const { getDefaultConfig } = require('expo/metro-config');
 const { mergeConfig } = require('@react-native/metro-config');
 const { lockdownSerializer } = require('@lavamoat/react-native-lockdown');
+// eslint-disable-next-line import-x/no-extraneous-dependencies
```
Gudahtt: Why is this ignore directive here? Is this an 'extraneous dependency'?

app.config.js
```diff
@@ -10,6 +10,15 @@ const OTA_ENV_MAP = {
...
 const OTA_ENV = OTA_ENV_MAP[process.env.METAMASK_ENVIRONMENT] ?? 'exp';
 
+// Disable OTA updates for E2E builds. We only ship OTA updates to real users
+// on production/RC builds; flipping this off for E2E lets the native runtime
+// skip reading `app.manifest` entirely (see expo-updates'
+// `EmbeddedManifestUtils`), which in turn lets `scripts/repack.js` replace
+// the redundant Metro-backed manifest generation with a cheap stub and cut
+// ~2m off every repack.
+const IS_E2E_BUILD =
+  process.env.IS_TEST === 'true' || process.env.METAMASK_ENVIRONMENT === 'e2e';
```
Gudahtt: `IN_TEST` and `METAMASK_ENVIRONMENT` are environment variables we inject into the application itself. But this file is a configuration file, it's not part of the application.

Are these always set when this file is read? Where do they get set?

.github/actions/post-build-source-hash/action.yml
```diff
@@ -0,0 +1,58 @@
...
+        echo "fingerprint=$FINGERPRINT" >> "$GITHUB_OUTPUT"
+        echo "Current fingerprint: ${FINGERPRINT}"
+
+    - name: Post commit status
+      uses: actions/github-script@v7
+      continue-on-error: true
+      env:
+        FINGERPRINT: ${{ steps.generate-fingerprint.outputs.fingerprint }}
+        STATUS_CONTEXT: ${{ inputs.status-context }}
+        TARGET_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
```
Gudahtt: `pull_request.head.sha` and `github.sha` are two completely different commits; which one is it supposed to be in this case?

It's a little confusing here because this action doesn't perform the checkout, it seems to expect that it was done already prior to this action being called. But there is no documented expectation of this, nor of which commit exactly is expected to be checked out.

.github/workflows/ci.yml
```diff
@@ -78,6 +78,41 @@ jobs:
             echo "No changes detected"
           fi
 
+  post-build-source-hash:
+    name: Post build-source-hash commit status
+    runs-on: ubuntu-latest
+    # Always publish the commit status (even for merge_group or when builds are
```
Gudahtt: Nit: This comment doesn't match the line that it's above.

The condition is ensuring we _skip_ this on forks, but the comment suggests the line ensures this _always runs_ (which it does not).

.github/actions/post-build-source-hash/action.yml
Gudahtt: It looks like currently this action is being used when `github.sha` is the current commit, but this would set the status on `github.event.pull_request.head.sha` instead.

It's setting the status on _the wrong commit_.

.github/workflows/ci.yml
```diff
@@ -538,6 +573,9 @@ jobs:
     name: 'Build Android APKs'
     if: >-
       ${{
+        !cancelled() &&
```
Gudahtt, line 829: I see that this ensures the job still runs when `post-build-source-hash` fails. Is this scenario expected to happen? Why would `post-build-source-hash` fail?

Gudahtt: Also I believe this would cause _later_ jobs to fail as well (e.g. `e2e-smoke-tests-android`) because the default status check function `success()` [looks at _all previous jobs_](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#success) (including dependencies of dependencies)

.github/workflows/build-android-e2e.yml
```diff
@@ -25,6 +25,22 @@ on:
...
+        description: >-
+          Canonical @expo/fingerprint hash for this commit, computed once in
+          the `post-build-source-hash` job. Used for cache keys and cross-run
+          artifact lookups. When empty (e.g. forked PR where the hash job is
+          skipped), the build compiles fresh with no cache reuse.
+        required: false
+        default: ''
+        type: string
+
+permissions:
```
Gudahtt: Why are permissions being declared redundantly both here, and in `ci.yml` where this is being invoked?

.github/workflows/ci.yml
Gudahtt: Likewise for `build-ios-apps`

.github/workflows/build-ios-e2e.yml
```diff
@@ -75,8 +87,113 @@ jobs:
...
+        uses: cirruslabs/cache/restore@bba69c6578b863ad0398ad40567bd2ef70290fe0 # v4
+        with:
+          path: |
+            ios/build/Build/Products/Release-iphonesimulator/MetaMask.app
+          key: ios-app-main-v${{ env.IOS_APP_CACHE_VERSION }}-${{ inputs.source-fingerprint }}
+
+      # Cross-PR reuse: before building, try to find a prior workflow run whose
+      # build-source-hash commit status matches our fingerprint AND whose
+      # MetaMask.app artifact is still available. Falls back to a fresh build.
+      - name: Find reusable build from prior run
```
Gudahtt, line 97: In which scenario would this step succeed where `Restore iOS app matching fingerprint from branch cache` and `Restore iOS app matching fingerprint from main cache` failed?

.github/actions/check-force-builds/action.yml
```diff
@@ -0,0 +1,81 @@
...
+  steps:
+    - name: Compute force-builds flag
+      id: compute
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        LABEL_NAME: ${{ inputs.label-name }}
+        COMMIT_TAG: ${{ inputs.commit-tag }}
+        EVENT_NAME: ${{ github.event_name }}
+        HEAD_COMMIT_HASH: ${{ github.event.pull_request.head.sha || github.sha }}
```
Gudahtt: Again these are two different commits. I think the one we want here is always `github.event.pull_request.head.sha`, since we only use this value for `pull_request` triggers, and since we're looking for a specific commit message (in cases where `github.sha`differs, it would be the auto-generated merge commit, which is never what we want)

scripts/repack.js
```diff
@@ -16,6 +17,159 @@ const logger = {
   warn: (msg) => console.warn(`⚠️  ${msg}`),
 };
 
+// Repack only runs for E2E-style builds; we use this to safely stub
+// `expo-updates` manifest generation (see `makeCustomSpawnAsync`).
+const IS_E2E_REPACK =
+  process.env.IS_TEST === 'true' ||
+  process.env.E2E === 'true' ||
+  process.env.METAMASK_ENVIRONMENT === 'e2e';
```
Gudahtt: Why three conditions? Is this run in different contexts, where different environment variables are set?

scripts/repack.js
```diff
@@ -16,6 +17,159 @@ const logger = {
...
+
+/**
+ * Implements the `SpawnProcessAsync` contract expected by `@expo/repack-app`
+ * using Node's `child_process.spawn`. Resolves with `{ pid, output, stdout,
+ * stderr, status, signal }` and attaches `.child` to the returned promise.
+ *
+ * Kept dependency-free so repack keeps working even if `@expo/spawn-async`
```
Gudahtt: :/

Why are we reimplementing `@expo/spawn-async` here? Why would it be "pruned"?

scripts/repack.js
```diff
@@ -16,6 +17,159 @@ const logger = {
...
+ *    consulted. Skipping this second Metro bundle saves ~2m per repack.
+ *    The stub writes an empty JSON manifest so the downstream `copyFile`
+ *    call still succeeds.
+ *
+ * @param {string} workingDirectory - Passed through so the stub can write
+ *   `app.manifest` at the expected path.
+ */
+function makeCustomSpawnAsync(workingDirectory) {
+  return function customSpawnAsync(command, args, options = {}) {
+    // 1) Drop --reset-cache from `npx expo export:embed` invocations so
```
Gudahtt: I certainly _hope_ we're not using `npx` still, that is forbidden for security reasons.

scripts/repack.js
```diff
@@ -16,6 +17,159 @@ const logger = {
...
+      command === 'npx' &&
+      Array.isArray(args) &&
+      args[0] === 'expo' &&
+      args[1] === 'export:embed' &&
+      args.includes('--reset-cache')
+    ) {
+      args = args.filter((a) => a !== '--reset-cache');
+    }
+
+    // 2) Skip `createUpdatesResources.js` for E2E repacks — `expo-updates`
```
Gudahtt: Could you elaborate on why we'd want to skip the manifest here, replacing it with a stub?

The comment implies that this manifest is related to OTA updates somehow, and that it's not needed if OTA updates are disabled. Maybe that's true. But is it also causing some problem for the repack process?

tommasini:
@Gudahtt Thanks for the awesome review!! Will address each comment and get back to you 🚀

scripts/repack.js
tommasini: This customSpawnAsync was created to save some time on the repack process that it was alone taking around 5 minutes, therefore the goal   was to improve it. 
This morning I thought about splitting this into another PR so we could discuss more around it. I will perform that, so we can get the cache in, and therefore the repack optimization on other PR 🙏

package.json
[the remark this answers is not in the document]
tommasini: This was removed on this commit: https://github.com/MetaMask/metamask-mobile/pull/29247/commits/f628dbc7bf15b0bed9ad66decd212638ef5b393a

fingerprint.config.js
tommasini: Argh I forgot to remove this comment, super fair! I thought, wrongly, that I would need to remove it to test the native cache, but the truth is that it was not changing commit to commit or cross PR because of github/workflows already changed

metro.config.js
tommasini: This was removed on this commit: https://github.com/MetaMask/metamask-mobile/commit/f628dbc7bf15b0bed9ad66decd212638ef5b393a

app.config.js
tommasini: This was removed on this commit: https://github.com/MetaMask/metamask-mobile/commit/f628dbc7bf15b0bed9ad66decd212638ef5b393a

scripts/repack.js
tommasini: This was removed on this commit: https://github.com/MetaMask/metamask-mobile/commit/f628dbc7bf15b0bed9ad66decd212638ef5b393a

scripts/repack.js
tommasini: Yeah the goal on re-implementing it was ambitious, it was to making it more efficient the repack process, maybe too ambitious, but I will get a PR with only this change opened so we can analyze deeper, it will be fun

This was removed on this commit: https://github.com/MetaMask/metamask-mobile/commit/f628dbc7bf15b0bed9ad66decd212638ef5b393a

scripts/repack.js
tommasini: This was removed on this commit: https://github.com/MetaMask/metamask-mobile/commit/f628dbc7bf15b0bed9ad66decd212638ef5b393a

scripts/repack.js
tommasini: No, this was referring to a usage of `npx` inside of expo/repack package

scripts/repack.js
Gudahtt: Oh interesting, OK sounds good

scripts/repack.js
Gudahtt: That would be equally problematic. If we're invoking expo and it's internally using `npx`, then we are still using `npx`

scripts/repack.js
Gudahtt: To clarify, I was asking here about the spawn-async utility function. Not about why we're writing a custom spawn async function.

I follow why we created a custom spawn async function, but we could have reused the utility function underneath rather than inlining it.

But sure, something to discuss in the future PR

scripts/repack.js
tommasini: Hmm marking here a todo, I need to look for this usage at the expo/repack node modules code

.github/workflows/ci.yml
tommasini: >I see that this ensures the job still runs when post-build-source-hash fails. Is this scenario expected to happen? Why would post-build-source-hash fail?

No, I'm not finding any expectation for it to fail. It's a safeguard for something that probably will not happen, and if it happens, we just need to force the builds to re run the e2e!

Niceeee!

.github/workflows/build-android-e2e.yml
tommasini: Amazing catch, it's indeed redudant, since it's already in ci.yml

.github/workflows/build-ios-e2e.yml
tommasini: Working on a fix to only use artifacts, we lost +/- 30 seconds comparing it to cirrus cache, but that's okay, since it reduces burden of maintainability if we shift runners.

.github/workflows/ci.yml
```diff
@@ -546,12 +582,16 @@ jobs:
     permissions:
       contents: read
       id-token: write
-    needs: [needs_e2e_build, smart-e2e-selection]
+      actions: read
+      pull-requests: read
+      statuses: read
+    needs: [needs_e2e_build, smart-e2e-selection, post-build-source-hash]
```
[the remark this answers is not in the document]
tommasini: Yeah this seems accurante, and on a fingerprint job failure we should run the E2E freshly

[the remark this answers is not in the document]
tommasini: Uh, it seems that it was what `!isCancelled()` was doing, I really need to understand this part better

.github/actions/post-build-source-hash/action.yml
```diff
@@ -0,0 +1,71 @@
...
+            );
+            return;
+          }
+          await github.rest.repos.createCommitStatus({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            sha: TARGET_SHA,
+            state: 'success',
+            context: STATUS_CONTEXT,
+            description: FINGERPRINT.slice(0, 140),
```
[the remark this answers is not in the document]
tommasini: This is not an issue a fingerprint should only have 40 chars. Example: "ba21f28bf6b8d33fbaeb6b8dbf0e6fbce36820a8"

scripts/repack.js
tommasini: @Gudahtt it's indeed there
image

Cal-L (dismissed):
LGTM

.github/workflows/build-ios-e2e.yml
tommasini: This latest version uses only github artifacts 🚀

Cal-L (dismissed):
Lgtm

Cal-L (approved):
Lgtm

PR #27978, opened 2026-03-26
fix: dedup chain change relay notifications over WalletConnect

wenfix:
## **Description**

WalletConnect reported 7,818 anomalous relay sessions from MetaMask Mobile. A relay complexity limit takes effect April 15, 2026. Chain change amplification is a major contributor — a single chain switch with 20 sessions generates ~80 relay messages instead of the expected ~40.

The root cause is two independent notification paths for chain changes:
1. **Store subscription path**: each `WalletConnect2Session` subscribes to Redux and calls `handleChainChange()` → sends `updateSession` + `emitSessionEvent` (2 messages)
2. **WalletConnectPort path**: `BackgroundBridge` emits `chainChanged` → `WalletConnectPort.postMessage` intercepts and calls `updateSession()` → sends another `updateSession` + `emitSessionEvent` (2 more messages)

This PR:
- Removes the duplicate WalletConnectPort `chainChanged` forwarding (path 2 above)
- Adds a `lastEmittedChainId` dedup guard in `handleChainChange` to prevent sequential duplicates from the `approveRequest` + store subscription race
- Captures the `store.subscribe()` unsubscribe function and calls it in `removeListeners()` to prevent leaked subscriptions

**Result**: 1 chain switch × 20 sessions = ~40 relay messages (2 per session), down from ~80.

## **Changelog**

CHANGELOG entry: Fixed duplicate WalletConnect relay messages when switching chains

## **Related issues**

Fixes: [WAPI-1358](https://consensyssoftware.atlassian.net/browse/WAPI-1358)

## **Manual testing steps**

```gherkin
Feature: WalletConnect chain change notifications

  Scenario: dapp requests chain switch
    Given the user has a dapp connected via WalletConnect on Mainnet

    When user approves a wallet_switchEthereumChain request from the dapp
    Then the dapp shows the new chain once (no duplicate notification)

  Scenario: chain round-trip A to B to A
    Given the user has a dapp connected via WalletConnect on Mainnet

    When user switches to Polygon then back to Mainnet
    Then the dapp reflects Polygon, then Mainnet (both transitions work)

  Scenario: account change still works
    Given the user has a dapp connected via WalletConnect

    When user switches the active account in MetaMask
    Then the connected dapp receives the account change notification
```

## **Pre-merge author checklist**

[...]

PR #29275, opened 2026-04-23
refactor(walletconnect): refactor wallet connect

Olivier-BB:
## **Description**

Refactored WalletConnect v2 session handling into a multichain adapter/routing architecture and updated permission handling to support Tron-aware flows.  
This improves separation of chain-specific logic (especially EIP-155 routing) from generic WalletConnect namespace/session code, and includes follow-up fixes from review for routing error handling and namespace resolution.

## **Changelog**

CHANGELOG entry: null

## **Related issues**

Fixes: TBD (no confirmed linked issue provided)

## **Manual testing steps**

```gherkin
Feature: WalletConnect multichain routing and Tron-aware permissions

  Scenario: user connects to a WalletConnect session using supported EIP-155 chains
    Given the app is running with WalletConnect enabled
    And a dapp provides a valid WalletConnect v2 URI

    When user approves the WalletConnect connection
    Then the session is established successfully
    And chain-specific routing is handled by the multichain adapter layer

  Scenario: user attempts session flow with delegated namespace keys
    Given a WalletConnect proposal/session includes delegated namespace keys

    When user proceeds through the WalletConnect connection flow
    Then the app resolves delegated namespace keys correctly
    And network selection remains consistent

  Scenario: routing encounters an EIP-155 handler error
    Given a WalletConnect request is routed through EIP-155 handling
    And the routing handler throws an error

    When the request is processed
    Then error details are captured in the catch path for diagnostics

  Scenario: Tron-aware permissions are requested
    Given a flow requires multichain permissions including Tron-related handling

    When permissions are evaluated
    Then Tron-aware permission logic is applied as expected

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Refactors WalletConnect v2 session approval/routing and permission-derived namespaces, which can affect request handling, chain switching, and event emission across dapps. Risk is moderated by added unit coverage but touches core connectivity flows.
> 
> **Overview**
> **WalletConnect v2 request/session handling is refactored into a multichain adapter architecture.** Session proposal approval now builds namespaces via a new `multichain` registry/adapters layer (including `wallet:eip155` aliasing), and non-EVM requests are routed through `MultichainRoutingService` while EVM requests move into `multichain/eip155` routing.
> 
> **Session lifecycle behavior changes.** `WalletConnect2Session` now normalizes persisted namespaces, subscribes to store chain changes to update the WC session + emit events, centralizes chain switching in `WalletConnect2Session.chain`, and tightens `chainChanged`/event emission to pick an appropriate CAIP chain and skip unsupported emissions.
> 
> **Permissions and network selection become Tron/delegation-aware.** Default CAIP-25 optional scopes now include Tron (feature-flagged), `getPermittedChains` returns all non-`wallet` scopes (not just EVM), and `MultichainAccountConnect` resolves delegated scope keys (e.g. `wallet:eip155`) when selecting default networks; related tests are updated/added, and WC logging shifts from `console.warn` to `DevLogger`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae57bdd43cb7f7dfb142a5f2b0ba26af4693b718. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### CODE REVIEW

File: app/core/Permissions/index.ts [hunk 1]
```diff
@@ -297,18 +300,25 @@ export const getPermittedCaipChainIdsByHostname = (
...
-  isMultichainOrigin: false,
-});
+export const getDefaultCaip25CaveatValue = (): Caip25CaveatValue => {
+  const optionalScopes: Caip25CaveatValue['optionalScopes'] = {
+    'wallet:eip155': { accounts: [] },
+  };
+
+  ///: BEGIN:ONLY_INCLUDE_IF(tron)
+  optionalScopes[TrxScope.Mainnet] = { accounts: [] };
+  ///: END:ONLY_INCLUDE_IF
```
adonesky1, line 313: [nit] not blocking but should we include in the tron PR?

app/core/Permissions/index.ts
```diff
@@ -297,18 +300,25 @@ export const getPermittedCaipChainIdsByHostname = (
...
+
+  return {
+    requiredScopes: {},
+    optionalScopes,
+    sessionProperties: {},
+    isMultichainOrigin: false,
```
jiexi, line 319:
```suggestion
    isMultichainOrigin: true,
```

@adonesky1 doesn't exactly map to the original purpose, but this for sure isn't EVM only

app/core/Permissions/index.ts
```diff
@@ -640,14 +651,16 @@ export const getPermittedChains = async (
...
-        )}` as CaipChainId,
+    return getAllScopesFromCaip25CaveatValue(caveat.value).filter(
+      (caipChainId: CaipChainId) => {
+        try {
+          const { namespace } = parseCaipChainId(caipChainId);
+          return namespace !== KnownCaipNamespace.Wallet;
+        } catch {
+          return false;
+        }
+      },
```
adonesky1, line 662: Are we sure we've accounted for all cases where this was expected to be EVM only?

adonesky1: I guess this is only used by WC... weird. I wonder if we should move it into the WalletConnect directory then?

app/core/WalletConnect/multichain/eip155/routing.ts
```diff
@@ -0,0 +1,256 @@
...
+    DevLogger.log(`WC::handleEvmRequest switching to chainId=${caip2ChainId}`);
+    await host.switchToChain(caip2ChainId, host.channelId);
+    await new Promise((resolve) => setTimeout(resolve, 100));
+    // Chain change notification is handled by BackgroundBridge → WalletConnectPort
+  }
+
+  if (!isAllowedChainId) {
+    DevLogger.log(`WC::checkWCPermissions chainId is not permitted`);
+    throw rpcErrors.invalidParams({
+      message: `Invalid parameters: active chainId is different than the one provided.`,
```
adonesky1, line 222: This doesn't seem to be the right error for `!isAllowedChainId`. I'm also not clear on when we'd ever want to throw an error if the "active chainId is different than the one provided".  Wouldn't it be more like `the chainId on which the request was submitted was not permitted in the connection. If the intended chain for submission ${hexChainId} is correct, try to reconnect and make sure that permissions are granted for this chain`

app/core/WalletConnect/WalletConnect2Session.chain.ts
```diff
@@ -0,0 +1,136 @@
...
+  // indicate the claimed source of the request. They MUST NOT be treated as
+  // equivalent to a verified origin/hostname.
+  const unverifiedOrigin = originFromRequest ?? selfReportedUrl;
+
+  const { allowed, existingNetwork, hexChainIdString } =
+    await hasPermissionsToSwitchChainRequest(caip2ChainId, channelId);
+
+  if (!allowed && !allowSwitchingToNewChain) {
+    throw rpcErrors.invalidParams({
+      message: `Invalid parameters: active chainId is different than the one provided.`,
```
adonesky1, line 80: This also seems like the wrong error to throw for this case... These errors should be way clearer about what the root cause is and how can the user fix or get around it.

app/core/WalletConnect/multichain/eip155/adapter.ts
```diff
@@ -0,0 +1,111 @@
...
+    const chains = permittedChains.filter((chain) =>
+      chain.startsWith(EIP155_PREFIX),
+    );
+
+    if (chains.length === 0) {
+      return undefined;
+    }
+
+    const approvedAccounts = getPermittedAccounts(channelId);
+    const accounts = Array.isArray(approvedAccounts)
```
jiexi, line 76: if this isn't an array, we have other problems. I don't think we should continue silently

app/core/WalletConnect/multichain/eip155/routing.ts
adonesky1: `invalid parameters` also doesn't feel like the right tag. Its more like a 500 internal error

adonesky1: Its also unclear if this error propogates in a way that will actually make it to the app/user. Have you attempted to validate these flows manually? Unless I'm misreading it - which is totally possible - it seems this may get stuck in [this catch block ](https://github.com/MetaMask/metamask-mobile/blob/ae57bdd43cb7f7dfb142a5f2b0ba26af4693b718/app/core/WalletConnect/WalletConnectV2.ts#L871-L875) where it will just logged to the console?

app/core/WalletConnect/multichain/eip155/adapter.ts
```diff
@@ -0,0 +1,111 @@
...
+      .some((ns) => ns?.chains?.includes(`${WALLET_PREFIX}eip155`));
+
+    if (!hasWalletEip155) return;
+
+    namespaces[KnownCaipNamespace.Wallet] = {
+      chains: [`${WALLET_PREFIX}eip155`],
+      methods: eip155.methods,
+      events: eip155.events,
+      accounts: eip155.accounts.map((account) => {
+        const address = account.split(':').slice(2).join(':');
```
jiexi, line 106: parseCaipAccountId might be good to use here

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -114,7 +123,36 @@ class WalletConnect2Session {
...
 
-    this.backgroundBridge = backgroundBridgeFactory.create({
+    this.backgroundBridge = this.createBackgroundBridge(
+      backgroundBridgeFactory,
+      { url, name, icons, channelId },
+    );
+
+    this.checkPendingRequests();
+    this.lastChainId = this.getCurrentChainId();
+    this.storeUnsubscribe = store.subscribe(this.onStoreChange.bind(this));
```
adonesky1, line 133: This reintroduces a chain-change relay amplification bug [WAPI-1358](https://consensyssoftware.atlassian.net/browse/WAPI-1358?atlOrigin=eyJpIjoiM2M5YTlkMGIyODkyNGUxNThiNDgwYTY2Yzg5YzU0OWUiLCJwIjoiaiJ9) / #27978 fixed three weeks ago. that PR deliberately deleted storeUnsubscribe / onStoreChange / handleChainChange and routed chain changes exclusively through WalletConnectPort (which is still wired up — see NOTIFICATION_NAMES.chainChanged there). with this store subscription back, every chain switch fires twice again: once via onStoreChange → handleChainChange → updateSession + emitSessionEvent, and once via BackgroundBridge.onStateUpdate → WalletConnectPort → wcRequestActions.updateSession → syncAndEmitChainChanged. isHandlingChainChange only guards re-entry within the store path, it can't see the port path. also note the lastEmittedChainId dedup guard from #27978 isn't brought into this PR.

PR #29428, opened 2026-04-28
feat(walletconnect): add multichain routing and namespace approval for first non-EVM chain (Tron)

Olivier-BB:
## **Description**

This PR refactors WalletConnect multichain handling by introducing a per-chain `ChainAdapter` registry and moving non-EVM behavior behind that adapter layer (with Tron as the first implementation). The goal is to reduce coupling in WalletConnect session code and make additional non-EVM chains easier to add.

It also improves namespace approval and routing behavior for mixed EVM + Tron sessions by approving only requested namespaces/chains, preserving EVM chain selection, and normalizing Tron request/response mapping for compatibility with dapp expectations.

## **Changelog**

CHANGELOG entry: Improved WalletConnect multichain compatibility by refining namespace approval and request routing for mixed EVM and Tron sessions.

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: WalletConnect multichain routing and namespace approval

  Scenario: user connects to a dapp requesting mixed EVM and Tron namespaces
    Given the user has at least one EVM account and one Tron account available
    And a dapp initiates a WalletConnect v2 proposal with EVM and Tron requirements

    When user approves the WalletConnect connection
    Then the approved namespaces include only namespaces and chains requested by the dapp
    And existing EVM chain/account selection is preserved

  Scenario: user signs a Tron request through WalletConnect
    Given an active WalletConnect session for Tron
    When user approves a Tron signing request from the dapp
    Then the request is mapped through the Tron chain adapter
    And the response is normalized to WalletConnect format expected by the dapp

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **High Risk**
> High risk because it changes WalletConnect v2 session approval, chain permission checks, namespace updates, and request routing, including new Snap-based handling for non-EVM requests; mistakes could break existing EVM sessions or incorrectly permit/deny chains.
> 
> **Overview**
> Adds a new WalletConnect `multichain` adapter/registry layer and implements the first non-EVM adapter for **Tron**, including CAIP chain-id normalization (hex/decimal), request mapping to Snap RPC shapes, and response normalization for Tron dapp expectations.
> 
> Updates WalletConnect v2 flows to **approve and update only requested namespaces/chains**, preserve previously approved non-EVM namespaces during updates, normalize/validate session namespaces, and route non-EVM requests through `MultichainRoutingService` (with improved unauthorized-chain errors and redirect-methods keyed by namespace).
> 
> Extends permissions defaults and lookups to be **namespace-agnostic** (adds Tron optional scope behind feature flag; `getPermittedChains` now returns all non-`wallet` scopes), tightens network-selection expansion logic to avoid broadening explicit `eip155` chain requests, and adjusts logging/error handling plus related unit/e2e test coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 72d4ea9ea37aa4d69f6efb36ea5ce3ea61393f36. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### CODE REVIEW

File: app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.test.tsx [hunk 1]
```diff
@@ -699,6 +704,108 @@ describe('MultichainAccountConnect', () => {
...
+            },
+            permissionRequestId: 'test-mixed-request',
+          },
+        }}
+      />,
+      { state: createMockState() },
+    );
+
+    const requestedChainIds =
+      mockUseAccountGroupsForPermissions.mock.calls.at(-1)?.[2] ?? [];
```
jiexi, line 735: guessing a plain ole `toHaveBeenCalledWith` wouldn't work here for some reason? Is there some other value in the rendered component we could query for instead here that would imply that `mockUseAccountGroupsForPermissions` was called with the expected chains?

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -215,6 +216,32 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
+    // Expand wallet:<namespace> scopes so default network selection still includes
+    // those concrete namespaces in multichain WalletConnect flows.
+    rawScopeKeys.forEach((scope) => {
+      if (!scope.startsWith(`${KnownCaipNamespace.Wallet}:`)) {
+        return;
+      }
+
+      const delegatedNamespace = scope.slice(
+        `${KnownCaipNamespace.Wallet}:`.length,
+      );
```
jiexi: Thoughts on something like this?
```
const { namespace, reference } = parseScopeString(scope);
if (namespace !== KnownCaipNamespace.Wallet || !reference) {
  return;
}
namespaces.add(reference as CaipNamespace);

```

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -322,14 +349,23 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
+      requestedCaipChainIds.map(
+        (caipChainId) => parseCaipChainId(caipChainId).namespace,
+      ),
+    );
+
+    // Expand only namespaces that were requested without explicit chains
+    // (e.g. wallet:eip155 delegated requests in mixed namespace proposals).
+    const additionalChains = nonTestNetworkCaipChainIds.filter(
+      (caipChainId) => {
+        const namespace = parseCaipChainId(caipChainId).namespace;
```
jiexi:
```suggestion
        const { namespace } = parseCaipChainId(caipChainId);
```

jiexi: nit

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -322,14 +349,23 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
+        return (
+          requestedNamespacesForNetworkSelection.includes(namespace) &&
+          !namespacesWithExplicitChainRequests.has(namespace)
```
jiexi, line 365: so if `wallet:eip155` and `eip155:1` are requested together, then no additional evm caip chain ids would be included in `addtionalChains`. Not saying that is right or wrong, just pointing it out. Not sure what we want here

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -71,6 +89,36 @@ class WalletConnect2Session {
 
   public session: SessionTypes.Struct;
 
+  private normalizeSessionNamespaces(
+    namespaces: SessionTypes.Struct['namespaces'] | undefined,
```
jiexi: `SessionTypes.Namespaces` is the same, right?

jiexi: ahh i see. You're doing it this way because we use the `SessionTypes.Struct` type elsewhere. Still feel like `SessionTypes.Namespaces` is more to the point

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -71,6 +89,36 @@ class WalletConnect2Session {
...
+  ): SessionTypes.Struct['namespaces'] {
+    const normalizedNamespaces: SessionTypes.Struct['namespaces'] = {};
+
+    Object.entries(namespaces ?? {}).forEach(([namespace, config]) => {
+      const accounts = Array.isArray(config?.accounts) ? config.accounts : [];
+      const derivedChains = Array.from(
+        new Set(
+          accounts
+            .map((account) => account.split(':').slice(0, 2).join(':'))
+            .filter((chain) => chain.includes(':')),
```
jiexi:
```
accounts.map => parseCaipAccountId(account).chainId
```

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -71,6 +89,36 @@ class WalletConnect2Session {
...
+        ),
+      );
+      const chains =
+        Array.isArray(config?.chains) && config.chains.length > 0
+          ? config.chains
+          : derivedChains;
```
jiexi, line 114: I'm unfortunately not familiar enough with the WCv2 api here. is `config.chains` guaranteed to contain all chains derived from the `config.accounts` value? or is it possible for `config.chains` to include less chains?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -269,21 +320,51 @@ class WalletConnect2Session {
   isHandlingRequest = () => this._isHandlingRequest;
 
   emitEvent = async (eventName: string, data: unknown) => {
```
jiexi, line 327: i believe emitEvent is only ever called with chainChanged and chain id (number). Is this no longer true?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -269,21 +320,51 @@ class WalletConnect2Session {
...
+    try {
+      return await getScopedPermissions({
+        channelId: sessionTopic,
+      });
+    } catch {
+      return await getScopedPermissions({
+        channelId: this.channelId,
+      });
+    }
+  }
```
jiexi: shouldn't this always use `this.channelId`? I'm not entirely sure, but `session.topic` isn't used to key the permissions. The channel id is, which is the same value as s`ession.pairingTopic`

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -370,18 +452,25 @@ class WalletConnect2Session {
...
 
       if (accounts.length === 0) {
-        const approvedAccounts = getPermittedAccounts(this.channelId);
-        if (approvedAccounts.length > 0) {
+        const sessionTopicApprovedAccounts = getPermittedAccounts(sessionTopic);
+        const fallbackApprovedAccounts = getPermittedAccounts(this.channelId);
+        const effectiveApprovedAccounts =
+          sessionTopicApprovedAccounts.length > 0
+            ? sessionTopicApprovedAccounts
+            : fallbackApprovedAccounts;
```
jiexi: i'm confused. why is sessionTopic relevant here? Aren't the permissions keyed by the channelId/pairingTopic?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -370,18 +452,25 @@ class WalletConnect2Session {
...
         } else {
-          console.warn(
-            `WC2::updateSession no permitted accounts found for topic=${this.session.topic} selfReportedUrl=${this.selfReportedUrl}`,
+          DevLogger.log(
+            `WC2::updateSession no permitted accounts found for sessionTopic=${sessionTopic} and channelId=${this.channelId} selfReportedUrl=${this.selfReportedUrl}`,
           );
-          return;
+          // Even if EVM accounts aren't permitted, Tron sessions can still
+          // succeed because `getScopedPermissions` adds the Tron namespace
+          // from account discovery. Avoid aborting here.
```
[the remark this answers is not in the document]
jiexi: wasn't `getPermittedAccounts` updated in this PR to return accounts for all namespaces, not just eip155?

app/core/WalletConnect/WalletConnect2Session.ts
Olivier-BB: Today it is effectively used for `chainChanged`, yes. The signature was kept generic because this class is already the session event dispatch point and we wanted to avoid hard-coding assumptions if additional event types are added.

app/core/Permissions/index.ts
ffmcgee725, line 22: remove if indeed it's an unused import

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
ffmcgee725: ➕

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+    }
+    if (typeof message === 'string') {
+      mappedParams.message = message;
+    }
+
+    return {
+      method: 'signMessage',
+      params: mappedParams,
+    };
+  }
```
ffmcgee725, line 297: to reduce cognitive complexity of function, as per https://sonarcloud.io/?pr=29428, perhaps we could extract this logic to a separate smaller helper function ? 

same for logic inside `if (method === 'tron_signTransaction')`

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.test.tsx
Olivier-BB: Good point. We used hook-call inspection because the regression is specifically in the derived chain list passed into useAccountGroupsForPermissions. We also added coverage for delegated expansion vs explicit-chain restriction to validate behavior.

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
Olivier-BB: Correct, intentional. When explicit eip155 chains are present, we treat that list as authoritative and avoid namespace-wide expansion to prevent over-approving chains. Delegated expansion still applies when no explicit chain IDs are provided.

app/core/WalletConnect/WalletConnect2Session.ts
Olivier-BB: Good suggestion, we switched this to SessionTypes.Namespaces since it’s more direct/explicit for this helper. Functionally equivalent, and validated in testing.

app/core/WalletConnect/WalletConnect2Session.ts
Olivier-BB: Addressed. We keep config.chains as authoritative when present, and only derive chains from accounts when config.chains is missing/empty. This avoids dropping valid session scope during restore/legacy cases and avoids narrowing to account-derived chains when config.chains is broader.

app/core/WalletConnect/WalletConnect2Session.ts
Olivier-BB: Addressed

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
Olivier-BB: Addressed

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
Olivier-BB: Addressed

app/core/WalletConnect/WalletConnect2Session.ts
Olivier-BB: Addressed: we removed the fallback helper and now use the canonical channelId directly for scoped-permissions lookup. This simplifies the flow and matches the actual permission keying model.

app/core/WalletConnect/WalletConnect2Session.ts
Olivier-BB: Addressed: good call, we removed the sessionTopic account lookup fallback and now read permitted accounts directly from the canonical channelId/pairing topic. This keeps permission reads aligned with the actual keying model and simplifies the logic.

app/core/WalletConnect/WalletConnect2Session.ts
jiexi: sure, but we're the only ones calling this method. There's also a hard baked assumption that this is being called for EVM accountsChanged given the eip155 caip chain id. I feel that this method can be significantly simplified if we assume `data` is already a `Number`

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -567,6 +745,110 @@ class WalletConnect2Session {
...
+      new Set([...permittedChainsFromChannel, ...permittedChainsFromSession]),
+    );
+    const compatibleRequestChainIds =
+      getCompatibleCaipChainIdsForWalletConnect(requestChainId);
+    const isPermittedByPermissionController = compatibleRequestChainIds.some(
+      (chainId) => permittedChains.includes(chainId as CaipChainId),
+    );
+    const activeSessionChains = Object.values(this.session.namespaces ?? {})
+      .flatMap((namespaceSlice) => namespaceSlice?.chains ?? [])
+      .filter(Boolean);
```
jiexi, line 820: is the filter needed? can there be falsy values in `chains`?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -567,6 +745,110 @@ class WalletConnect2Session {
...
+    );
+    const permittedChainsFromSession =
+      this.session.topic !== this.channelId
+        ? await getPermittedChainsSafe(this.session.topic)
+        : [];
+    const permittedChains = Array.from(
+      new Set([...permittedChainsFromChannel, ...permittedChainsFromSession]),
+    );
+    const compatibleRequestChainIds =
+      getCompatibleCaipChainIdsForWalletConnect(requestChainId);
```
jiexi, line 814: since `handleRequest` is the inbound direction, we only need to handle converting hex `tron:0xa` to decimal `tron:10` here, right?

jiexi: requestChainId is already in `tron:10` inbound format here and can be directly compared against permittedChains. No?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -370,18 +445,28 @@ class WalletConnect2Session {
       );
 
       if (accounts.length === 0) {
-        const approvedAccounts = getPermittedAccounts(this.channelId);
-        if (approvedAccounts.length > 0) {
+        const effectiveApprovedAccounts = getPermittedAccounts(this.channelId);
```
jiexi, line 448: thoughts on undoing this rename since the added ternary has been removed?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -370,18 +445,28 @@ class WalletConnect2Session {
...
+          DevLogger.log(
+            `WC2::updateSession no permitted accounts found for sessionTopic=${sessionTopic} and channelId=${this.channelId} selfReportedUrl=${this.selfReportedUrl}`,
           );
-          return;
+          // Keep legacy behavior for EVM-only sessions: do not update a session
+          // with empty account lists. Non-EVM adapter sessions may still build
+          // valid namespaces from adapter-managed permissions/accounts.
+          if (referencedAdapterNamespaces.length === 0) {
+            return;
+          }
```
jiexi, line 469: why is this additional check on `referencedAdapterNamespaces` needed? If there are no permitted accounts, is there still anything to really connect?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -395,22 +480,115 @@ class WalletConnect2Session {
         );
       }
 
-      const namespaces = await getScopedPermissions({
-        channelId: this.channelId,
-      });
+      const currentNamespaces = this.normalizeSessionNamespaces(
+        this.session.namespaces,
+      );
```
jiexi, line 485: proposedNamespaces? or some other naming for either this and/or `namespaces` below?

Also. isn't `this.session.namespaces` already normalized because we normalized it ourselves before setting `this.session`?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -395,22 +480,115 @@ class WalletConnect2Session {
...
         `🔴🔴 WC2::updateSession updating with namespaces`,
         namespaces,
       );
 
+      // Preserve already-approved namespaces (e.g. tron) if the freshly
+      // computed permissions payload is temporarily partial (e.g. only eip155).
+      const mergedNamespaces = {
+        ...currentNamespaces,
+        ...namespaces,
+      };
```
jiexi, line 499: is it possible for there to be overlaps in keys here? Is this being merged properly?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -395,22 +480,115 @@ class WalletConnect2Session {
...
+      // crash when they hit unexpected chain ids. Delegate the slice
+      // construction to each adapter.
+      const adapterSlices = buildAdapterNamespaces({
+        proposal: {
+          requiredNamespaces: this.session.requiredNamespaces,
+          optionalNamespaces: this.session.optionalNamespaces,
+        },
+        existingNamespaces: mergedNamespaces,
+      });
+      Object.assign(mergedNamespaces, adapterSlices);
```
jiexi, line 513: it's not clear what this does on the surface level. Is this effectively filtering out namespaces without accounts?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -395,22 +480,115 @@ class WalletConnect2Session {
...
+      // If the original proposal referenced any non-EVM namespace but no
+      // adapter slice was produced (e.g. wallet has no Tron account), do
+      // not emit a chainChanged with an EVM fallback. Non-EVM dapps don't
+      // recognise EVM hex chain ids and WalletKit may reject the emit.
+      const requiredAdapterNamespaces = proposalReferencedAdapterNamespaces({
+        requiredNamespaces: this.session.requiredNamespaces,
+        optionalNamespaces: {},
+      });
+      const missingRequiredAdapterNamespaces = requiredAdapterNamespaces.filter(
+        (ns) => !mergedNamespaces[ns]?.chains?.length,
```
jiexi, line 524: I'm having trouble understanding what is going on here

`requiredAdapterNamespaces` returns adapter generated namespaces from this.session.requiredNamespaces that have a matching adapter, which effectively is requiredNamespaces with an adapter and accounts.

`mergedNamespaces` contains this.session.namespace, which contains this.requiredNamespace

`missingRequiredAdapterNamespaces` returns this.session.requiredNamespaces with matching adapter and non-empty accounts but are missing entirely or have no chains defined in `mergedNamespaces`. 

Is this right? Can this be explained in simpler terms?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -395,22 +480,115 @@ class WalletConnect2Session {
...
+      // session update/rehydration.
+      const allowedNamespaceKeys = new Set<string>([
+        ...Object.keys(this.session.requiredNamespaces ?? {}),
+        ...Object.keys(this.session.optionalNamespaces ?? {}),
+      ]);
+      // Preserve whatever was in the currently-approved session as well
+      // (a session previously approved eip155 should keep eip155 on update).
+      Object.keys(currentNamespaces).forEach((key) =>
+        allowedNamespaceKeys.add(key),
+      );
```
jiexi, line 546: currentNamespaces is `this.session.namespaces`. Is this not the same set of keys as `this.session.requiredNamespaces` and `this.session.optionalNamespaces`?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -567,6 +745,110 @@ class WalletConnect2Session {
...
+          error: { code: 4902, message: ERROR_MESSAGES.INVALID_CHAIN },
+        },
+      });
+    }
+
+    // Mark redirect before any routing so all namespaces benefit from it.
+    const redirectNamespace =
+      requestNamespace === KnownCaipNamespace.Wallet
+        ? KnownCaipNamespace.Eip155
+        : requestNamespace;
```
jiexi, line 782: i don't think WC supports targeting the `wallet` or `wallet:eip155` scopes. `requestNamespace` may be safe as is

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -567,6 +745,110 @@ class WalletConnect2Session {
...
+    ): Promise<CaipChainId[]> => {
+      try {
+        return await getPermittedChains(subject);
+      } catch (error) {
+        if (error instanceof PermissionDoesNotExistError) {
+          return [];
+        }
+        throw error;
+      }
+    };
```
jiexi, line 801: if the session exists, then a permission should exist for it. Are you seeing this not be the case?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -567,6 +745,110 @@ class WalletConnect2Session {
...
+
+    const permittedChainsFromChannel = await getPermittedChainsSafe(
+      this.channelId,
+    );
+    const permittedChainsFromSession =
+      this.session.topic !== this.channelId
+        ? await getPermittedChainsSafe(this.session.topic)
+        : [];
```
jiexi, line 809: won't this always be empty array because permissions are not key'ed by `this.session.topic`, but rather `channelId` / `this.session.pairingTopic`?

app/core/WalletConnect/WalletConnect2Session.ts
```diff
@@ -567,6 +745,110 @@ class WalletConnect2Session {
...
+      (chainId) => permittedChains.includes(chainId as CaipChainId),
+    );
+    const activeSessionChains = Object.values(this.session.namespaces ?? {})
+      .flatMap((namespaceSlice) => namespaceSlice?.chains ?? [])
+      .filter(Boolean);
+    const isPermittedByActiveSession = compatibleRequestChainIds.some(
+      (chainId) => activeSessionChains.includes(chainId),
+    );
+    const isPermittedRequestChain =
+      isPermittedByPermissionController || isPermittedByActiveSession;
```
jiexi, line 825: the result from the PermissionController should be the source of truth ultimately. Both the activeSession and the permission are kept in sync - ish, so at minimum only one of these needs to be checked, not both. Unless you're seeing some reason otherwise?

app/core/WalletConnect/multichain/index.ts
```diff
@@ -0,0 +1,183 @@
...
+  mappedRequest,
+}: {
+  connectedAddresses: CaipAccountId[];
+  scope: CaipChainId;
+  requestId: number;
+  mappedRequest: SnapMappedRequest;
+}): Promise<unknown> =>
+  Engine.controllerMessenger.call('MultichainRoutingService:handleRequest', {
+    connectedAddresses,
+    origin: 'metamask',
```
jiexi, line 173: this is an unsafe internal origin. The value used here should probably be the channelId

app/core/WalletConnect/WalletConnectV2.ts
```diff
@@ -223,7 +277,8 @@ export class WC2Manager {
           accounts: approvedAccounts,
         });
       } catch (err) {
-        console.warn(`WC2::init can't update session ${sessionKey}`);
+        DevLogger.log(`WC2::init can't update session ${sessionKey}`);
+        await this.cleanupBrokenRestoredSession(session, err);
```
jiexi, line 281: why is this needed now?

app/core/WalletConnect/WalletConnectV2.ts
```diff
@@ -599,10 +656,17 @@ export class WC2Manager {
     // Get the current chain ID to include in permissions
     const walletChainIdHex = selectEvmChainId(store.getState());
     const walletChainIdDecimal = parseInt(walletChainIdHex, 16);
+    const referencedAdapterNamespaces = proposalReferencedAdapterNamespaces(
+      proposal.params,
+    );
+    const isMultichainOrigin = referencedAdapterNamespaces.length > 0;
```
jiexi, line 662: @adonesky1 . does it really matter what value this flag is now?

app/core/WalletConnect/WalletConnectV2.ts
```diff
@@ -678,6 +746,100 @@ export class WC2Manager {
...
 
+      const requiredNamespaces = proposal.params.requiredNamespaces ?? {};
+      const optionalNamespaces = proposal.params.optionalNamespaces ?? {};
+      const allProposalNamespaces = {
+        ...optionalNamespaces,
+        ...requiredNamespaces,
+      } as Record<
+        string,
+        { chains?: string[]; methods?: string[]; events?: string[] }
+      >;
```
jiexi, line 757: keys can overlap here

app/core/WalletConnect/WalletConnectV2.ts
```diff
@@ -678,6 +746,100 @@ export class WC2Manager {
...
+          ? namespaces.eip155.accounts
+          : [];
+        namespaces.wallet = {
+          chains: walletChains,
+          methods: namespaces.eip155.methods ?? [],
+          events: namespaces.eip155.events ?? [],
+          accounts: eip155Accounts.map((account) => {
+            const [, , address] = account.split(':');
+            return `wallet:eip155:${address}` as `${string}:${string}:${string}`;
+          }),
```
jiexi, line 800: is this wallet namespace used later? I feel like i've already seen special handling of `wallet:eip155` that would imply that namespaces['wallet'] is not used, but maybe I'm wrong

app/core/WalletConnect/WalletConnectV2.ts
```diff
@@ -678,6 +746,100 @@ export class WC2Manager {
...
+      // providers for the namespaces they requested and crash during session
+      // rehydration when they encounter approved namespaces they don't know
+      // about (e.g. `Cannot read properties of undefined (reading 'chains')`
+      // when iterating `session.namespaces` and looking up an unknown rpc
+      // provider). Per WalletConnect spec, only namespaces that were present
+      // in required/optional should be approved.
+      const allowedNamespaceKeys = new Set<string>([
+        ...Object.keys(proposal.params.requiredNamespaces ?? {}),
+        ...Object.keys(proposal.params.optionalNamespaces ?? {}),
+      ]);
```
jiexi, line 815: is key collision possible here?

app/core/WalletConnect/WalletConnectV2.ts
```diff
@@ -823,14 +996,28 @@ export class WC2Manager {
...
+            session = new WalletConnect2Session({
+              web3Wallet: this.web3Wallet,
+              channelId: activeSession.pairingTopic,
+              navigation: this.navigation,
+              deeplink: true,
+              session: activeSession,
+            });
+            this.sessions[sessionTopic] = session;
+          }
+        }
```
jiexi, line 1020: why is this needed now?

app/core/WalletConnect/wc-utils.ts
```diff
@@ -427,3 +442,116 @@ export const getUnverifiedRequestOrigin = (
...
+ * Any non-EVM namespace (Tron, Solana, Bitcoin, ...) takes priority when
+ * present — its CAIP-2 chain id is used as both the `chainId` envelope and
+ * the `data` payload, since non-EVM dapps don't recognise EVM hex chain
+ * ids. Otherwise we fall back to EVM, using the wallet's active EVM chain
+ * when the session doesn't expose any.
+ */
+export const getChainChangedEmissionForWalletConnect = ({
+  namespaces,
+  fallbackEvmDecimal,
+  fallbackEvmHex,
```
jiexi, line 483: it feels odd for the caller to be responsible to provide the evm chain id formatted in both decimal and hex

tests/page-objects/Browser/TestSnaps.ts
```diff
@@ -28,6 +28,21 @@ import SolanaTestDApp from './SolanaTestDApp';
...
+  if (raw == null) {
+    return '';
+  }
+  try {
+    return JSON.stringify(raw);
+  } catch {
+    return String(raw);
+  }
+}
+
```
jiexi, line 45: confused on how the WC changes would warrant this change here

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.test.tsx
```diff
@@ -699,6 +704,174 @@ describe('MultichainAccountConnect', () => {
     });
   });
 
+  it('passes EVM chains to permissions hook for mixed wallet:eip155 + tron requests', () => {
```
adonesky1, line 707: right?
```suggestion
  it('passes EVM chains and Tron to permissions hook for mixed wallet:eip155 + tron requests', () => {
```

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -215,6 +216,27 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
   );
 
+  const requestedNamespacesForNetworkSelection = useMemo(() => {
+    const namespaces = new Set(requestedNamespacesWithoutWallet);
+    const rawScopeKeys = [
+      ...Object.keys(requestedCaip25CaveatValue.requiredScopes ?? {}),
+      ...Object.keys(requestedCaip25CaveatValue.optionalScopes ?? {}),
+    ];
+
+    // CAIP-25 may encode delegated namespace requests (for example wallet:eip155).
```
adonesky1, line 226: I'm not sure what `encode delegated namespace requests` means?

app/components/Views/MultichainAccounts/MultichainAccountConnect/MultichainAccountConnect.tsx
```diff
@@ -322,14 +344,28 @@ const MultichainAccountConnect = (props: AccountConnectProps) => {
...
+      ),
+    );
+
+    // Expand only namespaces that were requested without explicit chains
+    // (e.g. wallet:eip155 delegated requests in mixed namespace proposals).
+    // Intentionally restrictive: if explicit eip155 chains are requested, treat
+    // that list as authoritative and do not expand to all eip155 networks. This
+    // prevents approving more chains than requested. Delegated namespace
+    // expansion still applies when no explicit chain ids are provided.
+    // TODO: check if this restrictive logic has already been applied in the past or if this modifies previous behavior.
```
adonesky1, line 359: I think this makes sense for this pass. Just noting that we are probably going to be loosening chain permissions quite a bit in the near future: i.e. you ask for 1 EVM chain you get all EVM chains. But TBD

app/core/Permissions/index.ts
```diff
@@ -297,18 +300,25 @@ export const getPermittedCaipChainIdsByHostname = (
 
 /**
  * Returns a default CAIP-25 caveat value.
+ * Each chain appends its optional scope below (feature-flagged).
```
adonesky1, line 303: [nit]
```suggestion
 * Each ecosystem (i.e. EVM, Tron) appends its optional scope below (feature-flagged).
```

app/core/Permissions/index.ts
```diff
@@ -297,18 +300,25 @@ export const getPermittedCaipChainIdsByHostname = (
...
-  isMultichainOrigin: false,
-});
+export const getDefaultCaip25CaveatValue = (): Caip25CaveatValue => {
+  const optionalScopes: Caip25CaveatValue['optionalScopes'] = {
+    'wallet:eip155': { accounts: [] },
+  };
+
+  ///: BEGIN:ONLY_INCLUDE_IF(tron)
+  optionalScopes[TrxScope.Mainnet] = { accounts: [] };
+  ///: END:ONLY_INCLUDE_IF
```
adonesky1, line 313: not sure I'm following why this would be added as a default namespace for permissions? I'm struggling to figure out how to test this properly and I'm not totally remembering or understanding how this helper function is used but its name implies that this addition will mean Tron network permissions will be granted by default? TBH I'm not sure how that would be the case... but either way I'm not sure I understand the purpose of this addition.

app/core/Permissions/index.ts
```diff
@@ -624,10 +634,11 @@ export const getPermittedAccounts = (
...
 /**
- * Get permitted chains for the given the host.
+ * Get permitted chains for the given the host, across all namespaces.
+ * Returns all non-wallet scopes stored in the CAIP-25 caveat.
  *
  * @param hostname - Subject to check if permissions exists. Ex: A Dapp is a subject.
- * @returns An array containing permitted chains for the specified host.
+ * @returns An array containing permitted CAIP chain IDs for the specified host.
  */
 export const getPermittedChains = async (
```
adonesky1, line 643:
```suggestion
export const getPermittedCaipChainIds = async (
```

Yes the typing gives this context, but the sibling `getPermittedAccounts` is currently only returning EVM accounts... so this could get confusing

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+/** WalletConnect events the wallet may emit for the Tron namespace. */
+const TRON_EVENTS: readonly string[] = [];
+
+/** CAIP-2 prefix used to identify Tron chain ids in proposals. */
+const TRON_PREFIX = 'tron:';
+const DEFAULT_TRON_CHAIN_ID = 'tron:0x2b6653dc';
+
+export const normalizeCaipChainIdInboundForWalletConnectTron = (
+  caipChainId: string,
+): string => {
```
adonesky1, line 50: [nit] shouldn't the return type also be `caipChainId`?

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+const TRON_METHODS: string[] = ['tron_signTransaction', 'tron_signMessage'];
+
+/** WalletConnect events the wallet may emit for the Tron namespace. */
+const TRON_EVENTS: readonly string[] = [];
+
+/** CAIP-2 prefix used to identify Tron chain ids in proposals. */
+const TRON_PREFIX = 'tron:';
+const DEFAULT_TRON_CHAIN_ID = 'tron:0x2b6653dc';
+
+export const normalizeCaipChainIdInboundForWalletConnectTron = (
```
adonesky1, line 48: We should add a comment explaining why this normalization is required? I asked Claude to explain why this was necessary because it's nonobvious when just looking at it. WalletConnect tends to encode the Tron chainIds with hex values for the `reference` i.e. `tron:0x2b6653dc` while we use decimals for the reference `tron:728126428`

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+  if (!chainRef.startsWith('0x')) {
+    if (!/^\d+$/.test(chainRef)) {
+      return caipChainId;
+    }
+    return `tron:0x${parseInt(chainRef, 10).toString(16)}`;
+  }
+  return caipChainId;
+};
+
+export const getCompatibleTronCaipChainIdsForWalletConnect = (
```
adonesky1, line 77: Lets please add some JSdocs to explain why this utility function is necessary?

adonesky1: Looking at where this is used, I'm not understanding why we need this encoded in both hex and decimal formats side by side ever?

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+  accounts: `${string}:${string}:${string}`[];
+}
+
+/**
+ * List Tron EOA addresses currently managed by the wallet.
+ */
+const listTronEoaAddresses = (): string[] =>
+  Engine.context.AccountsController.listAccounts()
+    .filter((account: { type: string }) => account.type === TrxAccountType.Eoa)
+    .map((account: { address: string }) => account.address);
```
adonesky1, line 121: Just so I'm clear do we support Tron SCAs?

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+    string,
+    { chains?: string[]; methods?: string[]; events?: string[] } | undefined
+  >;
+}
+
+/**
+ * Shape of a single namespace slice in WalletConnect's approved
+ * namespaces map. Mirrors `SessionTypes.Namespace` but kept loose.
+ */
+export interface TronNamespaceSlice {
```
adonesky1, line 108: hmmm this seems tied to a mixed version of CAIP-25. Can you point me to the docs where this type is described?

app/core/WalletConnect/multichain/types.ts
```diff
@@ -0,0 +1,115 @@
...
+  chains: string[];
+  methods: string[];
+  events: string[];
+  /**
+   * CAIP-10 account ids: `${namespace}:${chainRef}:${address}`.
+   * Kept as `string[]` (rather than a template-literal type) so this
+   * matches `SessionTypes.Namespace.accounts` which is a plain string[].
+   */
+  accounts: string[];
+}
```
adonesky1, line 28: Do we have two different versions of this same interface? https://github.com/MetaMask/metamask-mobile/pull/29428/changes#diff-09c1410947d91a7bf8ff2bd7a39a10ad19c1b4d40083abfdc2c8b86d08580f73R47

Also `TronNamespaceSlice` is awfully similar...?

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+    typeof (transactionContainer as Record<string, unknown>).transaction ===
+      'object' &&
+    (transactionContainer as Record<string, unknown>).transaction !== null
+      ? ((transactionContainer as Record<string, unknown>)
+          .transaction as Record<string, unknown>)
+      : typeof transactionContainer === 'object' &&
+          transactionContainer !== null &&
+          !Array.isArray(transactionContainer)
+        ? (transactionContainer as Record<string, unknown>)
+        : undefined;
```
adonesky1, line 410: ummm this can't be reasoned about in any meaningful way

app/core/WalletConnect/multichain/tron/index.ts
```diff
@@ -0,0 +1,608 @@
...
+
+/**
+ * List Tron EOA addresses currently managed by the wallet.
+ */
+const listTronEoaAddresses = (): string[] =>
+  Engine.context.AccountsController.listAccounts()
+    .filter((account: { type: string }) => account.type === TrxAccountType.Eoa)
+    .map((account: { address: string }) => account.address);
+
+export const buildTronScopedPermissionsNamespace = ({
```
adonesky1, line 123: When a function gets this long I find it can help to add some inline comments explain what is going on at each step

0xEdouardEth:
Closed in favor of: https://github.com/MetaMask/metamask-mobile/pull/30519/changes

PR #29509, opened 2026-04-29
feat: add token list security badges

sahar-fehri:
## **Description**

Adds security/trust indicators on main wallet token list rows when the remote flag **`tokenListSecurityBadges`** is enabled 

## **Changelog**

CHANGELOG entry: Adds security trust badges in tokenlist

## **Related issues**

Fixes:

## **Manual testing steps**

```gherkin
Feature: my feature name

```

### **After**

https://github.com/user-attachments/assets/813e570a-c420-4c02-8068-4cee5c072a40

- [ ] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Introduces new remote-flag-controlled network fetching and caching in a high-traffic UI (token list), which could impact performance or render timing. Logic is guarded by `basicFunctionalityEnabled`/feature flag and includes batching plus tests to reduce risk.
> 
> **Overview**
> Adds *remote-flagged* inline security/trust badges to main wallet token list rows via the new `tokenListSecurityBadges` LaunchDarkly/client-config flag (version-gated) and registry entry.
> 
> Implements a batched TanStack Query data path to resolve a token to a CAIP-19 asset id and fetch minimal token security payloads (`fetchTokenAssets` with `includeTokenSecurityData`), then renders Verified/Risky/Malicious indicators using a new shared `SecurityTrustInlineBadge` component (also refactoring Trending rows to reuse it). Includes unit tests for the selector, query keys, CAIP resolution/normalization, batching/deduping behavior, and adjusts TokenListItem tests to mock Query usage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6fd29b86b1a2669fe00fb1a46ed79f8afd3ba8d8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/components/UI/Tokens/queries/tokenSecurityBadgeKeys.ts
```diff
@@ -0,0 +1,10 @@
+import type { CaipAssetType } from '@metamask/utils';
+
+/**
+ * Query key factory for token list security badges (TanStack Query).
+ */
+export const tokenListSecurityBadgeKeys = {
+  all: () => ['tokenList', 'securityBadge'] as const,
+  byAsset: (caipAssetId: CaipAssetType) =>
+    [...tokenListSecurityBadgeKeys.all(), caipAssetId] as const,
```
sahar-fehri, line 16: Similar-ish pattern exists in **predictFeaturedCarouselKeys** which only has all() because there is one carousel query for the whole screen ; one cache entry: ['predict', 'featuredCarousel'].
Token list shows many assets at once. Each row needs TokenSecurityData for that caipAssetId. TanStack Query keys must differ per asset.

app/components/UI/Tokens/hooks/useTokenListSecurityBadgeQuery.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Fetches minimal security payload for list badges; results are batched via
+ * {@link requestTokenSecurityForAsset}.
+ */
+export function useTokenListSecurityBadgeQuery(caipAssetId: CaipAssetType) {
+  return useQuery<TokenSecurityData | null, Error>({
+    queryKey: tokenListSecurityBadgeKeys.byAsset(caipAssetId),
+    queryFn: async () => requestTokenSecurityForAsset(caipAssetId),
+    staleTime: STALE_TIME_MS,
```
sahar-fehri, line 18: We will refetch data after 12 hours as some sort of periodic scan

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,104 @@
...
+ * Used by TanStack `useQuery` for token list security badges.
+ */
+export function requestTokenSecurityForAsset(
+  assetId: CaipAssetType,
+): Promise<TokenSecurityData | null> {
+  return new Promise((resolve) => {
+    const list = waitersByAsset.get(assetId) ?? [];
+    list.push({ resolve });
+    waitersByAsset.set(assetId, list);
+    scheduleFlush();
```
sahar-fehri, line 122: batching with short flush delay; splits fetchTokenAssets into chunks of at most 100 assetIds

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,104 @@
...
+        for (const a of assets) {
+          byAssetId.set(a.assetId as CaipAssetType, a.securityData ?? null);
+        }
+      } catch {
+        // Fail-open per chunk so large flushes still get partial data.
+      }
+    }
+    resolveAll(
+      (id) => byAssetId.get(normalizeCaipAssetIdForTokenApi(id)) ?? null,
+    );
```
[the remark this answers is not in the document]
sahar-fehri: [cf6d83b](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/cf6d83b3cce8983eeaa3bf941fcb53cc228d0bb0)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,107 @@
...
+  );
+  if (match) {
+    return `eip155:${match[1]}/erc20:${match[2].toLowerCase()}` as CaipAssetType;
+  }
+  return assetId as CaipAssetType;
+}
+
+interface Waiter {
+  resolve: (data: TokenSecurityData | null) => void;
+}
```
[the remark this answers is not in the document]
sahar-fehri: this is ok; will retry in 12H;

app/components/UI/Tokens/hooks/useTokenListSecurityBadgeQuery.ts
```diff
@@ -0,0 +1,20 @@
...
+  });
```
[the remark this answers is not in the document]
sahar-fehri: The “5 minutes” note applies to TanStack Query’s default cacheTime when the QueryClient doesn’t override it. Our app sets cacheTime: 24h globally in ReactQueryService, so inactive queries here are not garbage-collected after 5 minutes.
We still set cacheTime on this hook to match staleTime so caching is explicit and stays consistent unless global defaults change;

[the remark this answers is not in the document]
sahar-fehri: [045898c](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/045898c08e999873b91218f00e34b962d2a0ce21)

app/components/UI/Tokens/queries/tokenSecurityBadgeKeys.ts
bergarces: Even though we are adding 'tokenList' as part of the key, I assume we will use the same key if we were to fetch the security badge for TokenDetails or any other features, right?

app/components/UI/Tokens/util/getCaipAssetIdForToken.ts
```diff
@@ -0,0 +1,89 @@
...
+    return null;
+  }
+
+  try {
+    if (asset.address && isCaipAssetType(asset.address)) {
+      return asset.address;
+    }
+
+    if (asset.isNative || asset.isETH) {
+      const id = await getNativeAssetId(asset.chainId as Hex);
```
sahar-fehri, line 69: @bergarces using this approach to get proper caip asset id for natives

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
```diff
@@ -181,6 +184,52 @@ export const TokenListItem = React.memo(
...
+        })
+        .catch(() => {
+          if (!cancelled) {
+            setCaipAssetIdForSecurity(null);
+          }
+        });
+      return () => {
+        cancelled = true;
+      };
+    }, [asset, shouldResolveCaipForSecurityBadge]);
```
bergarces:
```suggestion
    const { data: caipAssetIdForSecurity } = useQuery({
      queryKey: ['caipAssetId', asset?.chainId, asset?.address],
      queryFn: () => getCaipAssetIdForToken(asset),
      enabled: shouldResolveCaipForSecurityBadge,
      staleTime: Infinity, // CAIP mapping doesn't change
    });
```

bergarces: With useQuery, the useEffect + useState combo can be avoided. Only advantages IMO.

Basically, anything that needs to be fetched asynchronously is almost always the perfect candidate for useQuery.

bergarces: If this is also needed elsewhere (getting the caip19 from the chain and address), this hook and key can be extracted and then the cache is used by the whole app. But for now it can be left here.

app/components/UI/Tokens/util/getCaipAssetIdForToken.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (hasProperty(SPOT_PRICES_SUPPORT_INFO, chainId)) {
+    return SPOT_PRICES_SUPPORT_INFO[
+      chainId as keyof typeof SPOT_PRICES_SUPPORT_INFO
+    ] as string;
+  }
+
+  const decimalChainId = convertHexToDecimal(chainId);
+  const slip44CoinType = await Slip44Service.getEvmSlip44(
+    Number(decimalChainId),
+  );
```
bergarces, line 32: Something very annoying of this method is that it returns 60 when the slip44 is unknown.

I think we can live with that in this case, if the native asset for a chain is malicious, we have bigger problems 😓

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
sahar-fehri: Thnx @bergarces updated here [8865e0b](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/8865e0b160fe31fb4f0af6b7665aae38fc1a4676)

app/components/UI/Tokens/util/getCaipAssetIdForToken.ts
sahar-fehri: Agree; i dnt see another alternative for now

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,109 @@
...
+  const byAssetId = new Map<CaipAssetType, TokenSecurityData | null>();
+  try {
+    for (const chunk of chunkArray(assetIds, MAX_ASSETS_PER_REQUEST)) {
+      try {
+        const assets = await fetchTokenAssets(chunk, {
+          includeTokenSecurityData: true,
+        });
+        for (const a of assets) {
+          byAssetId.set(
+            normalizeCaipAssetIdForTokenApi(String(a.assetId)),
```
bergarces:
```suggestion
            normalizeCaipAssetIdForTokenApi(a.assetId),
```

I've seen LLMs converting types that are already a string to String before. Not sure why it does that, but I don't think it's needed.

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,109 @@
...
+ * requests of at most {@link MAX_ASSETS_PER_REQUEST} asset ids when a flush is large.
+ * Used by TanStack `useQuery` for token list security badges.
+ */
+export function requestTokenSecurityForAsset(
+  assetId: CaipAssetType,
+): Promise<TokenSecurityData | null> {
+  return new Promise((resolve) => {
+    const list = waitersByAsset.get(assetId) ?? [];
+    list.push({ resolve });
+    waitersByAsset.set(assetId, list);
```
bergarces: Couldn't `assetId` be normalized here so that duplicate entries with different case for the same asset are added to the same list?

By doing it here, it also means it won't be necessary to do so in line 88, when calling `resolveAll`.

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,109 @@
...
+  try {
+    for (const chunk of chunkArray(assetIds, MAX_ASSETS_PER_REQUEST)) {
+      try {
+        const assets = await fetchTokenAssets(chunk, {
+          includeTokenSecurityData: true,
+        });
+        for (const a of assets) {
+          byAssetId.set(
+            normalizeCaipAssetIdForTokenApi(String(a.assetId)),
+            a.securityData ?? null,
```
bergarces, line 96: I don't know if resolving to `null` instead of rejecting is a good idea.

If we resolve to null, the result will get cached for the next 12 hours with useQuery. If we reject, however, it will try to refetch (configurable in useQuery).

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,109 @@
...
+  assetId: string,
+): CaipAssetType {
+  const match = /^eip155:(\d+)\/erc20:(0x[0-9a-fA-F]{40})$/.exec(
+    String(assetId),
+  );
+  if (match) {
+    return `eip155:${match[1]}/erc20:${match[2].toLowerCase()}` as CaipAssetType;
+  }
+  return assetId as CaipAssetType;
+}
```
bergarces, line 33:
```suggestion
export function normalizeCaipAssetIdForTokenApi(
  assetId: CaipAssetType,
): CaipAssetType {
  const match = /^eip155:(\d+)\/erc20:(0x[0-9a-fA-F]{40})$/.exec(assetId);
  if (match) {
    return `eip155:${match[1]}/erc20:${match[2].toLowerCase()}`;
  }
  return assetId;
}
```

If we force the input to be CaipAssetType (which it is everywhere), we do not need all the casting.

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
bergarces: Nevermind, since we are not fetching here, but inside the controller, the retry logic should probably live there.

Otherwise we risk spamming the API if a token returns no results.

sahar-fehri: Agree with your point; just thought resolving null is simpler and handling rejection might not worth it;
I thought about this; from what i tested it seems whenever api fails its a legit failure; retrying wont make it work and i though it is fine to retry after cache is invalid;
wdyt 🤔

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: [4de8791](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/4de879167cc605bd5c07965812793d32273fc825)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: [4de8791](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/4de879167cc605bd5c07965812793d32273fc825)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: [4de8791](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/4de879167cc605bd5c07965812793d32273fc825)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
bergarces: I think it's ok for now. We can revisit in future if it becomes an issue, but I don't think it will.

bergarces (dismissed):

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
```diff
@@ -181,6 +185,40 @@ export const TokenListItem = React.memo(
...
+        'caipAssetIdForSecurityBadge',
+        asset?.chainId,
+        asset?.address,
+        asset?.isNative,
+        asset?.isETH,
+      ],
+      queryFn: () => getCaipAssetIdForToken(asset),
+      enabled: shouldResolveCaipForSecurityBadge && Boolean(asset?.chainId),
+      staleTime: Infinity,
+    });
```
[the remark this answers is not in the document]
bergarces: I have no idea if this is true or not. Maybe worth setting cacheTime to Infinity as well.

app/components/UI/Tokens/hooks/useTokenListSecurityBadgeQuery.ts
```diff
@@ -0,0 +1,23 @@
+import { useQuery } from '@tanstack/react-query';
+import type { TokenSecurityData } from '@metamask/assets-controllers';
+import type { CaipAssetType } from '@metamask/utils';
+
+import { tokenListSecurityBadgeKeys } from '../queries/tokenSecurityBadgeKeys';
+import { requestTokenSecurityForAsset } from '../util/tokenSecurityBadgeBatch';
+
+const STALE_TIME_MS = 12 * 60 * 60 * 1000; //12 hours
```
juanmigdr, line 8: Shouldnt all the code in app/components/UI/Tokens/queries/tokenSecurityBadgeKeys.test.ts live here?

.github/scripts/known-feature-flag-constants.ts
```diff
@@ -12,6 +12,7 @@ const FILE_SOURCES: Array<{ key: string; file: string; exportName: string }> = [
   { key: 'BITCOIN_REWARDS_FLAG_NAME', file: REWARDS_FILE, exportName: 'BITCOIN_REWARDS_FLAG_NAME' },
```
juanmigdr, line 12: Is there any chance we can replace the spinner with a skeleton? Looks more professional I believe

.github/scripts/known-feature-flag-constants.ts
```diff
@@ -12,6 +12,7 @@ const FILE_SOURCES: Array<{ key: string; file: string; exportName: string }> = [
...
   { key: 'TRON_REWARDS_FLAG_NAME', file: REWARDS_FILE, exportName: 'TRON_REWARDS_FLAG_NAME' },
```
juanmigdr, line 13: Badge is not aligned vertically centered with the title 

Image

.github/scripts/known-feature-flag-constants.ts
```diff
@@ -12,6 +12,7 @@ const FILE_SOURCES: Array<{ key: string; file: string; exportName: string }> = [
...
   { key: 'SNAPSHOTS_REWARDS_FLAG_NAME', file: REWARDS_FILE, exportName: 'SNAPSHOTS_REWARDS_FLAG_NAME' },
```
juanmigdr, line 14: When opening the token details page I see we are still making the API call to include token security dta, I would expect this to be removed since we can obtain it from the cache:

Image

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,106 @@
...
+
+function chunkArray<T>(items: T[], chunkSize: number): T[][] {
+  const chunks: T[][] = [];
+  for (let i = 0; i < items.length; i += chunkSize) {
+    chunks.push(items.slice(i, i + chunkSize));
+  }
+  return chunks;
+}
+
+function scheduleFlush() {
```
juanmigdr, line 56: Why are we scheduling a flush, shouldnt tanstack take care of this?

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,106 @@
+import {
+  fetchTokenAssets,
+  type TokenSecurityData,
+} from '@metamask/assets-controllers';
+import type { CaipAssetType } from '@metamask/utils';
+
+export function normalizeCaipAssetIdForTokenApi(
```
juanmigdr, line 12: Cant we reuse any existing function?

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
```diff
@@ -0,0 +1,106 @@
...
+  if (flushTimer !== null) {
+    return;
+  }
+  flushTimer = setTimeout(() => {
+    flushTimer = null;
+    flush().catch(() => undefined);
+  }, FLUSH_MS);
+}
+
+async function flush() {
```
juanmigdr, line 66: A bit lost into all this flushing logic, not sure why its needed and how it all works together

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
bergarces: Does Tasnstack query have a way of batching multiple calls and making a single query with all of them?

If so we definitely should use it, but I'm not aware of it.

.github/scripts/known-feature-flag-constants.ts
sahar-fehri: I thought about this too; i can attempt to do skeleton; the thing with skeleton is that we need the perfect width and height.. I can do something and if needs fixes we can cherry pick

.github/scripts/known-feature-flag-constants.ts
sahar-fehri: Nice catch; already fixed here

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
bergarces: Tanstack query just fetches one asset id at a time, and stores it as a single key.

The function exported here exposes that interface (receives one assetId, returns the result for it) but, internally, debounces the fetch by a set amount of time, batches all the assetIds together and makes a single fetch for all of them.

This way, when we open the home page, it makes one single call with 5 assetIds, but every call to useQuery only receives and caches its own.

.github/scripts/known-feature-flag-constants.ts
sahar-fehri: Agree; i left this intentionally; we have a 12H cache (which we might change in future depending on feedback we get regarding api load/ necessity to refresh this more or less often) 
So i thought i can keep the on demand call for now; and remove after
Happy to remove now if you think otherwise

app/components/UI/Tokens/hooks/useTokenListSecurityBadgeQuery.ts
sahar-fehri: [073cbf4](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/073cbf4c6dc876ff3050c30819ff37c141c68649)

.github/scripts/known-feature-flag-constants.ts
sahar-fehri: [35cf03f](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/35cf03f56e4dd2e85e37476541828167f151aea1)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: My understanding is that React Query runs one queryFn per key. It will dedupe the mounts of the same key, but it will still call queryFn separately for each token row e.g: USDC, DAI, WETH, etc. That would mean many `fetchTokenAssets` calls unless something else batches them

sahar-fehri: what the schedule flush is doing essentially;
We have first requestTokenSecurityForAsset that queues each queryFn’s resolve; then scheduleFlush waits a few milliseconds so all the queries that start in the same scroll/render enqueue together, then flush runs one (or multiple) `fetchTokenAssets` for the whole set

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: I think related to the other comment here https://github.com/MetaMask/metamask-mobile/pull/29509#discussion_r3168037301

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: Hmmmm not sure we have a utility that does this already, might have missed it

app/components/UI/Tokens/queries/tokenSecurityBadgeKeys.ts
Prithpal-Sooriya: Agreed, keys specifically used for caching and invalidation.

E.g. above if invalidation is used on "tokenList", all useQueries that contain tokenList will invalidate.

If platforms/feature teams adopt tanstack query, then it might be worth consolidating keys somewhere.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
```diff
@@ -181,6 +186,40 @@ export const TokenListItem = React.memo(
...
+
+    const { data: caipAssetIdForSecurity } = useQuery({
+      queryKey: [
+        'tokenList',
+        'caipAssetIdForSecurityBadge',
+        asset?.chainId,
+        asset?.address,
+        asset?.isNative,
+        asset?.isETH,
+      ],
```
Prithpal-Sooriya: nit (not a blocker) - Ideally I would move query key creation all to a shared config.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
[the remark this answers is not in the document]
sahar-fehri: the CacheTime is 24H in our client, not 5mins as it claims; i think yes can add same ad staleTime

app/components/UI/Tokens/util/getCaipAssetIdForToken.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    balance: '0',
+    logo: undefined,
+    isETH: false,
+    chainId: '0x1',
+    ...partial,
+  } as TokenI;
+}
+
+describe('getCaipAssetIdForToken', () => {
+  it('returns null when asset is null', async () => {
```
Prithpal-Sooriya: nit - tests follow a similar pattern. Can use a test table / `it.each` with defined test cases.

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
Prithpal-Sooriya: Agreed, don't really want us to manage and control regex here. Lets reuse some @metamask/utils - I think there is a parseAssetId or similar to split out regex properties; and toAssetId to build the caip-19 string.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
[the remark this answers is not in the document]
sahar-fehri: [be29b3e](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/be29b3e2779f62cb9aeeac8ddc463cbf1fb6600a)

app/components/UI/Tokens/util/getCaipAssetIdForToken.test.ts
sahar-fehri: [2af8043](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/2af8043ea580968464e48b913bc27f0ca99ea6a6)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
Prithpal-Sooriya: Okay, the flush is acting as a batch mechanism.
If mobile is using lodash, can we reuse lodash utils?

<details>
  <summary>sync ver</summary>

```ts
const createBatchedRequestTokenSecurityForAsset = () => {
  let pendingAssetIds: CaipAssetType[] = [];

  const processBatch = _.throttle(() => {
    if (pendingAssetIds.length > 0) {
      const batch = [...pendingAssetIds];
      pendingAssetIds = [];

      // BATCH CALL GOES HERE
    }
  }, FLUSH_MS);

  // single call
  return (assetId: CaipAssetType) => {
    pendingAssetIds.push(assetId);
    // can add optional processBatch.flush if we have max size
  }
}

export const requestTokenSecurityForAsset = createBatchedRequestTokenSecurityForAsset()
```
</details>

<details>
  <summary>async ver</summary>

This is fed to chat, might requires some tweaking

```ts
const createBatchedRequestTokenSecurityForAsset = () => {
  let pending: {
    assetId: CaipAssetType;
    resolve: (data: TokenSecurityData | null) => void;
    reject: (error: Error) => void;
  }[] = [];

  const processBatch = _.throttle(
    async () => {
      if (pending.length === 0) return;

      const batch = [...pending];
      pending = [];

      const uniqueAssetIds = _.uniq(batch.map((p) => p.assetId));
      const resultsByAssetId = new Map<CaipAssetType, TokenSecurityData | null>();

      // Chunk requests to respect API limit, fail-open per chunk
      for (const chunk of _.chunk(uniqueAssetIds, MAX_ASSETS_PER_REQUEST)) {
        try {
          const assets = await fetchTokenAssets(chunk, {
            includeTokenSecurityData: true,
          });
          for (const a of assets) {
            resultsByAssetId.set(
              normalizeCaipAssetIdForTokenApi(a.assetId),
              a.securityData ?? null,
            );
          }
        } catch {
          // Fail-open per chunk so large flushes still get partial data.
        }
      }

      // Resolve every pending request (including duplicates)
      batch.forEach(({ assetId, resolve }) => {
        resolve(resultsByAssetId.get(assetId) ?? null);
      });
    },
    FLUSH_MS,
    { leading: false },
  );

  /**
   * Batches concurrent `fetchTokenAssets` calls (DataLoader-style), splitting into
   * requests of at most {@link MAX_ASSETS_PER_REQUEST} asset ids when a flush is large.
   * Used by TanStack `useQuery` for token list security badges.
   */
  return (assetId: CaipAssetType): Promise<TokenSecurityData | null> =>
    new Promise((resolve, reject) => {
      pending.push({
        assetId: normalizeCaipAssetIdForTokenApi(assetId),
        resolve,
        reject,
      });
      processBatch();
    });
};

export const requestTokenSecurityForAsset =
  createBatchedRequestTokenSecurityForAsset();
```

</details>

app/components/UI/Tokens/util/getCaipAssetIdForToken.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Native CAIP-19 asset id for an EVM chain (spot-price constants first, then Slip44).
+ * Expects `chainId` as canonical hex from network state (e.g. `0x1`).
+ */
+export async function getNativeAssetId(chainId: Hex): Promise<string> {
+  if (hasProperty(SPOT_PRICES_SUPPORT_INFO, chainId)) {
+    return SPOT_PRICES_SUPPORT_INFO[
+      chainId as keyof typeof SPOT_PRICES_SUPPORT_INFO
+    ] as string;
+  }
```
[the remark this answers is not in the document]
sahar-fehri: dnt think this is an issue; SPOT_PRICES_SUPPORT_INFO is a fixed map of hex chain IDs => non-null CAIP strings. There are no null values to return. If a chain isn’t in the map, the code never reads that map; it uses the Slip44 fallback instead.

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: thnx @Prithpal-Sooriya ! pushed an update here [1b66c56](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/1b66c56887926ccfcd54b4fdbba6e334d02f85e2)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: heya @Prithpal-Sooriya ! thanks for this ❤️  is there a perf or behavior win I’m missing (vs setTimeout + Map deduping), or is it mainly consistency / readability?
just wanna be mindful that would love to get this in for the RC unless there are major concerns which i currently dnt see 🙏

.github/scripts/known-feature-flag-constants.ts
bergarces: I think we ought to do it now. It will prove that useQuery is handling the cached data and won't do another fetch.

It's literally just calling `useTokenListSecurityBadgeQuery` there.

app/components/UI/Tokens/TokenList/TokenListItem/TokenListItem.tsx
sahar-fehri: Great idea! Can use the already existing tokenListSecurityBadgeKeys

sahar-fehri: [6fd29b8](https://github.com/MetaMask/metamask-mobile/pull/29509/commits/6fd29b86b1a2669fe00fb1a46ed79f8afd3ba8d8)

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
Prithpal-Sooriya: Yeah this is mostly for consistency and readability.
The main change was to encapsulate most of this logic inside a closure - hides away complexity on how flushing and stuff works.
- Ideally we should have a generic shared async batching util, but I cannot find one

Prithpal-Sooriya (approved):
LGTM

bergarces (approved):

.github/scripts/known-feature-flag-constants.ts
sahar-fehri: This will be in a followup

app/components/UI/Tokens/util/tokenSecurityBadgeBatch.ts
sahar-fehri: will followup with this

davibroc (approved):

Cal-L (approved):
LGTM

PR #29653, opened 2026-05-04
feat: Implement UI messengers

Mrtenz:
## **Description**

Initial implementation of the UI messenger. This sets up the UI messenger, a route messenger, and refactors a Snap page to demonstrate how to use the new route-level messenger.

More details can be found in [the UI messenger ADR](https://github.com/MetaMask/decisions/blob/6760a1616df98f5b8bb7c876fadd7f255a2b8957/decisions/core/0020-integrate-messengers-into-ui.md).

## **Changelog**

CHANGELOG entry:

## **Related issues**

Refs: n/a

## **Manual testing steps**

n/a

## **Screenshots/Recordings**

n/a

- [x] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Changes how UI invokes controller actions (snap removal now goes through delegation) and introduces a broad new access-control surface; mitigated by explicit per-route capabilities and KeyringController exclusions, but more routes will need careful capability review as adoption grows.
> 
> **Overview**
> Introduces a **UI messenger layer** so screens can call controller actions through typed, capability-scoped messengers instead of reaching into `Engine` directly.
> 
> **Root** now creates a singleton `UIMessenger` and wraps the app in `UIMessengerProvider`. The UI messenger delegates allowed actions to `Engine.controllerMessenger` and forwards events from the background; a config list (starting with **KeyringController** unsafe/keyring APIs) blocks actions that must not be exposed to the UI.
> 
> **Per-route messengers** are wired via `withMessenger` / `RouteWithMessenger`: on mount they delegate only the declared actions/events, expose them through `RouteMessengerContext`, and revoke on unmount. **Snap Settings** is the first adopter—the navigator wraps `SnapSettings` with `SnapController:removeSnap`, and removal uses `useMessenger().call('SnapController:removeSnap', …)` instead of `Engine.context.SnapController.removeSnap`. Tests and `renderWithProvider` gain optional `uiMessenger` / `routeMessenger` mocks to match this pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbbcad3605177adcf32e57408b9e937a3e2a3359. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/components/Nav/Main/MainNavigator.js
+++ b/app/components/Nav/Main/MainNavigator.js
@@ -171,7 +171,7 @@ import BenefitFullView from '../../UI/Rewards/Views/BenefitFullView';
 import BenefitsFullView from '../../UI/Rewards/Views/BenefitsFullView';
 import { getDeFiProtocolPositionDetailsNavbarOptions } from '../../UI/Navbar';
 import MoneyTabPressTracker from '../../UI/Money/components/MoneyTabPressTracker';
-import { createRouteWithMessenger } from '../../../messengers/helpers/route-messenger-helpers';
+import { withMessenger } from '../../../messengers/helpers/route-messenger-helpers';
 
 const Stack = createStackNavigator();
 const NativeStack = createNativeStackNavigator();
@@ -434,8 +434,7 @@ const ExploreHome = () => {
 };
 
 ///: BEGIN:ONLY_INCLUDE_IF(snaps)
-const SnapSettingsWithMessenger = createRouteWithMessenger({
-  Component: SnapSettings,
+const SnapSettingsWithMessenger = withMessenger(SnapSettings, {
   capabilities: SNAPS_SETTINGS_ROUTE_ALLOWED_CAPABILITIES,
 });
 
--- a/app/components/Nav/Main/MainNavigator.js
+++ b/app/components/Nav/Main/MainNavigator.js
@@ -86,7 +86,10 @@ import SendTransaction from '../../UI/Ramp/Aggregator/Views/SendTransaction';
 import TabBar from '../../../component-library/components/Navigation/TabBar';
 ///: BEGIN:ONLY_INCLUDE_IF(snaps)
 import { SnapsSettingsList } from '../../Views/Snaps/SnapsSettingsList';
-import { SnapSettings } from '../../Views/Snaps/SnapSettings';
+import {
+  SnapSettings,
+  ALLOWED_CAPABILITIES as SNAPS_SETTINGS_ROUTE_ALLOWED_CAPABILITIES,
+} from '../../Views/Snaps/SnapSettings';
 import { CAN_INSTALL_THIRD_PARTY_SNAPS } from '../../../constants/snaps';
 ///: END:ONLY_INCLUDE_IF
 import Routes from '../../../constants/navigation/Routes';
@@ -168,6 +171,7 @@ import BenefitFullView from '../../UI/Rewards/Views/BenefitFullView';
 import BenefitsFullView from '../../UI/Rewards/Views/BenefitsFullView';
 import { getDeFiProtocolPositionDetailsNavbarOptions } from '../../UI/Navbar';
 import MoneyTabPressTracker from '../../UI/Money/components/MoneyTabPressTracker';
+import { createRouteWithMessenger } from '../../../messengers/helpers/route-messenger-helpers';
 
 const Stack = createStackNavigator();
 const NativeStack = createNativeStackNavigator();
@@ -430,6 +434,11 @@ const ExploreHome = () => {
 };
 
 ///: BEGIN:ONLY_INCLUDE_IF(snaps)
+const SnapSettingsWithMessenger = createRouteWithMessenger({
+  Component: SnapSettings,
+  capabilities: SNAPS_SETTINGS_ROUTE_ALLOWED_CAPABILITIES,
+});
+
 const SnapsSettingsStack = () => {
   const { colors } = useTheme();
   return (
@@ -445,7 +454,7 @@ const SnapsSettingsStack = () => {
       />
       <NativeStack.Screen
         name={Routes.SNAPS.SNAP_SETTINGS}
-        component={SnapSettings}
+        component={SnapSettingsWithMessenger}
       />
     </NativeStack.Navigator>
   );
--- a/app/components/Views/Snaps/SnapSettings/messenger.ts
+++ b/app/components/Views/Snaps/SnapSettings/messenger.ts
@@ -2,7 +2,11 @@ import { defineAllowedRouteCapabilities } from '../../../../messengers/helpers/r
 import type { RouteMessengerFromCapabilities } from '../../../../messengers/route-messenger';
 
 export const ALLOWED_CAPABILITIES = defineAllowedRouteCapabilities({
-  actions: ['SnapController:disconnectOrigin', 'SnapController:installSnaps'],
+  actions: [
+    'SnapController:disconnectOrigin',
+    'SnapController:installSnaps',
+    'SnapController:removeSnap',
+  ],
   events: [],
 });
 
--- a/app/components/Views/Snaps/SnapSettings/messenger.ts
+++ b/app/components/Views/Snaps/SnapSettings/messenger.ts
@@ -1,0 +1,11 @@
+import { defineAllowedRouteCapabilities } from '../../../../messengers/helpers/route-messenger-helpers';
+import type { RouteMessengerFromCapabilities } from '../../../../messengers/route-messenger';
+
+export const ALLOWED_CAPABILITIES = defineAllowedRouteCapabilities({
+  actions: ['SnapController:disconnectOrigin', 'SnapController:installSnaps'],
+  events: [],
+});
+
+export type RouteMessengerInstance = RouteMessengerFromCapabilities<
+  typeof ALLOWED_CAPABILITIES
+>;
--- a/app/components/Views/Snaps/SnapSettings/messenger.ts
+++ b/app/components/Views/Snaps/SnapSettings/messenger.ts
@@ -2,11 +2,7 @@ import { defineAllowedRouteCapabilities } from '../../../../messengers/helpers/r
 import type { RouteMessengerFromCapabilities } from '../../../../messengers/route-messenger';
 
 export const ALLOWED_CAPABILITIES = defineAllowedRouteCapabilities({
-  actions: [
-    'SnapController:disconnectOrigin',
-    'SnapController:installSnaps',
-    'SnapController:removeSnap',
-  ],
+  actions: ['SnapController:removeSnap'],
   events: [],
 });
 
--- a/app/core/Engine/types.ts
+++ b/app/core/Engine/types.ts
@@ -517,7 +517,7 @@ type SnapsGlobalEvents =
   | PhishingControllerEvents;
 ///: END:ONLY_INCLUDE_IF
 
-type GlobalActions =
+export type GlobalActions =
   ///: BEGIN:ONLY_INCLUDE_IF(sample-feature)
   | SamplePetnamesControllerActions
   ///: END:ONLY_INCLUDE_IF
@@ -607,7 +607,7 @@ type GlobalActions =
   | ChompApiServiceActions
   | MoneyAccountUpgradeControllerActions;
 
-type GlobalEvents =
+export type GlobalEvents =
   ///: BEGIN:ONLY_INCLUDE_IF(sample-feature)
   | SamplePetnamesControllerEvents
   ///: END:ONLY_INCLUDE_IF
--- a/app/messengers/configs/index.ts
+++ b/app/messengers/configs/index.ts
@@ -1,0 +1,22 @@
+/* eslint-disable import-x/no-namespace */
+
+import * as keyringController from './keyring-controller';
+import { GlobalActions, GlobalEvents } from '../../core/Engine';
+
+interface MessengerExclusions {
+  // This is named like this since it's intended to be defined as a top-level
+  // property of a messenger configuration.
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  EXCLUDED_CAPABILITIES: {
+    actions: readonly GlobalActions['type'][];
+    events: readonly GlobalEvents['type'][];
+  };
+}
+
+// If you need to exclude an action or event from being accessible in the UI,
+// add a file to this directory and then add the module to this list.
+// Note: A file does not need to exist for every controller or service, just
+// those that need to exclude certain actions and/or events.
+export const MESSENGERS_WITH_EXCLUSIONS = [
+  keyringController,
+] satisfies MessengerExclusions[];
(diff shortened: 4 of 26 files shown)


app/components/Nav/Main/MainNavigator.js
```diff
@@ -413,6 +417,11 @@ const ExploreHome = () => {
 };
 
 ///: BEGIN:ONLY_INCLUDE_IF(snaps)
+const SnapSettingsWithMessenger = createRouteWithMessenger({
+  Component: SnapSettings,
+  capabilities: SNAPS_SETTINGS_ROUTE_ALLOWED_CAPABILITIES,
+});
```
Mrtenz: Unfortunately we cannot have a `RouteWithMessenger` component to wrap `Stack.Screen`, as `Stack.Screen` _must_ be a direct descendant of `Stack.Navigator`. This seems like the cleanest way to create route messengers and avoid rerendering issues (if we were to call `createRouteWithMessenger` inside `SnapsSettingsStack`).

app/components/Views/Snaps/SnapSettings/messenger.ts
```diff
@@ -0,0 +1,15 @@
+import { defineAllowedRouteCapabilities } from '../../../../messengers/helpers/route-messenger-helpers';
+import type { RouteMessengerFromCapabilities } from '../../../../messengers/route-messenger';
+
+export const ALLOWED_CAPABILITIES = defineAllowedRouteCapabilities({
+  actions: [
+    'SnapController:disconnectOrigin',
```
FrederikBolding: All of these are not used on mobile (yet)

app/messengers/configs/keyring-controller.ts
```diff
@@ -0,0 +1,14 @@
+import { defineExcludedCapabilities } from './helpers';
+
+// By default, all actions and events are allowed. If there an action or event
+// you think should NOT be accessible from the UI, update this.
+export const EXCLUDED_CAPABILITIES = defineExcludedCapabilities({
+  actions: [
+    'KeyringController:addNewKeyring',
```
FrederikBolding, line 7: Why does this differ from the extension implementation?

app/messengers/ui-messenger.ts
```diff
@@ -0,0 +1,151 @@
...
+  async subscribe<EventType extends UIMessengerEvents['type']>(
+    eventType: EventType,
+    handler: (
+      ...payload: ExtractEventPayload<UIMessengerEvents, EventType>
+    ) => void,
+  ): Promise<() => Promise<void>> {
+    // @ts-expect-error: Type of handler is not compatible.
+    Engine.controllerMessenger.subscribe(eventType, handler);
+
+    return async () => {
```
FrederikBolding: How/when is this called?

app/messengers/helpers/route-messenger-helpers.tsx
```diff
@@ -0,0 +1,140 @@
...
+      ParamList,
+      RouteName,
+      NavigatorID,
+      State,
+      ScreenOptions,
+      EventMap
+    >,
+  ) {
+    return (
+      <RouteWithMessenger path={props.route.name} capabilities={capabilities}>
```
FrederikBolding, line 120: I don't understand how this works, where does `props.route` come from?

app/messengers/ui-messenger.ts
```diff
@@ -0,0 +1,151 @@
...
+type WithJsonParams<Action extends ActionConstraint> =
+  Action extends ActionConstraint
+    ? Parameters<Action['handler']> extends Json[]
+      ? Action
+      : never
+    : never;
+
+/**
+ * Keeps only events whose payload is JSON-serializable, since all events come
+ * through the background connection.
```
FrederikBolding: Should we update these JSDocs for mobile?

app/components/Views/Snaps/SnapSettings/messenger.ts
Mrtenz: `SnapController:removeSnap` is used actually. Will remove the others.

app/messengers/ui-messenger.ts
Mrtenz: Hmm, we do the same thing in the extension, but I realise now that this doesn't make sense as messengers have a separate `unsubscribe` function. Need to rethink this a bit.

app/messengers/helpers/route-messenger-helpers.tsx
Mrtenz: It's provided by the navigator library's `Stack.Screen`.

app/messengers/configs/keyring-controller.ts
Mrtenz: I think the version of `keyring-controller` was different when I started implementing this and it flagged a lot of actions as non-existent, but that seems to be fixed now. Will address.

Also looks like the list in the extension isn't properly type checked, since `KeyringController:addKeyring` doesn't exist. 😕

app/messengers/ui-messenger.ts
FrederikBolding: Yeah, I missed this during my extension review. In extension the UI messenger probably needs to keep track of the cleanup functions somehow. In mobile, that doesn't seem necessary, we can just forward the calls.

app/messengers/helpers/route-messenger-helpers.tsx
```diff
@@ -0,0 +1,140 @@
...
+ * a route messenger with the specified capabilities.
+ *
+ * @param options - Options bag.
+ * @param options.capabilities - Capabilities to delegate from the UI messenger
+ * to the route messenger.
+ * @param options.Component - The component to render for this route. This will
+ * be wrapped in a {@link RouteWithMessenger} component that provides the route
+ * messenger.
+ */
+export function createRouteWithMessenger<
```
mcmire: I'm worried it may not be obvious at first glance that this is a HOC.

I don't know that we have any HOCs in mobile, but in extension I notice they tend to be named `with*`, e.g. [`withRouterHooks`](https://github.com/MetaMask/metamask-extension/blob/33771e29480f81a10552005a7878d20f8341ecf1/ui/helpers/higher-order-components/with-router-hooks/with-router-hooks.tsx). Should we follow suit here and name this `withMessenger`?

Alternatively, should we at least take the component as a positional argument and take an options bag or the capabilities directly as the second argument?

app/components/Nav/Main/MainNavigator.js
mcmire: This will result in a lot of new variables created over time in this file or any any other routes file. But, I can't think of any other way to do this. So it seems fine for now.

app/core/Engine/types.ts
```diff
@@ -504,7 +504,7 @@ type SnapsGlobalEvents =
   | PhishingControllerEvents;
 ///: END:ONLY_INCLUDE_IF
 
-type GlobalActions =
+export type GlobalActions =
```
mcmire, line 520: It would be good to align on terminology at some point. I believe that from a Wallet Framework perspective we have settled on the name "root messenger" instead of "global messenger". Oh well, maybe later...

app/messengers/ui-messenger.ts
mcmire: Yes, would be good to update this type, `MakeActionsAsynchronous`, and `getAction` below.

app/messengers/ui-messenger.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (excludedActions.includes(anyActionType)) {
+      throw new Error(
+        `The action "${actionType}" has not been exposed to the UI.`,
+      );
+    }
+
+    return (...args: unknown[]) =>
+      // @ts-expect-error: `unknown[]` is not assignable to `args`, but the type
+      // here is checked in `call`, so this is safe.
+      Engine.controllerMessenger.call(actionType, ...args);
```
mcmire: Should we wrap this in a promise? 

``` suggestion
    return async (...args: unknown[]) =>
      // @ts-expect-error: `unknown[]` is not assignable to `args`, but the type
      // here is checked in `call`, so this is safe.
      Engine.controllerMessenger.call(actionType, ...args);
```

app/messengers/ui-messenger.ts
```diff
@@ -0,0 +1,151 @@
...
+  WithJsonParams<Exclude<GlobalActions, { type: ExcludedActionTypes }>>
+>;
+
+export type UIMessengerEvents = WithJsonPayload<
+  Exclude<GlobalEvents, { type: ExcludedEventTypes }>
+>;
+
+const UI_MESSENGER_NAMESPACE = 'UI';
+
+export class UIMessenger extends Messenger<
```
mcmire: Would it be worth mentioning why we are creating a special subclass of Messenger and what advantage it has, especially since in Mobile we have direct access to the root messenger? Something like this:

``` suggestion
// `UIMessenger` is a subclass of `Messenger` rather than a type alias because
// it functions as a denylist rather than an allowlist.
//
// Although it is a "child" of the root messenger, it does not expect to be set
// up using the conventional delegation pattern. Rather, it delegates all
// actions and events to the root messenger by default, using an exclusion list
// to guard action delegation in particular.
export class UIMessenger extends Messenger<
```

app/messengers/ui-messenger.ts
mcmire: I just realized that `subscribe` on the UI messenger acts differently than a conventional messenger, since it's using `subscribeToMessengerEvent` under the hood. Is there _any_ way we could get `unsubscribe` to work there? Then we could remove the `subscribe` override here.

Mrtenz: How would we remove the override? The subscribe call still needs to be forwarded to the engine messenger somehow.

mcmire: Yeah, you're right. I was thinking that since we have direct access to the root messenger in Mobile, there ought to be some simplifications we can make. But `UIMessenger` doesn't work like other messengers do, so overriding `subscribe` makes sense.

Mrtenz: I think we just need to override unsubscribe as well, right? Instead of returning an unsubscribe function.

FrederikBolding: Yeah, but we will need to do something else on extension.

Mrtenz: Yes, on the extension I think we can create a map of handlers to unsubscribe functions, e.g.:

```ts
#unsubscribeFunctions: Map<EventName, Map<HandlerFunction, UnsubscribeFunction>>;
```

Will create a pull request to fix that too.

app/messengers/helpers/route-messenger-helpers.tsx
Mrtenz: Makes sense!

app/messengers/configs/index.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable import-x/no-namespace */
+
+import * as keyringController from './keyring-controller';
+import { GlobalActions, GlobalEvents } from '../../core/Engine';
+
+interface MessengerExclusions {
```
mcmire, line 6: I just realized this is using an `interface` not a `type`. I guess the preference on whether to use `interface` vs. `type` is flipped in this repo vs. how we have it set up in other repos? Or can this use `type`?

mcmire (commented):
One minor thing I noticed, but everything else looks good!

app/messengers/configs/index.ts
Mrtenz: ESLint was complaining about using `type`, so it looks like the preference is flipped.

Gudahtt: Yeah the lint rule here is backwards from most places. Though it's very frequently overridden with eslint-ignore directives.

mcmire: Ah, gotcha. Well, I guess this is okay then.

app/messengers/route-with-messenger.tsx
```diff
@@ -0,0 +1,79 @@
...
+      .then(() => setDelegated(true))
+      .catch(captureException);
+
+    return () => {
+      uiMessenger
+        .revoke({ messenger, actions, events })
+        .then(() => setDelegated(false))
+        .catch(captureException);
+    };
+  }, [uiMessenger, capabilities]);
```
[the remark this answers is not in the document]
Mrtenz: While accurate, we shouldn't use `RouteWithMessenger` directly. I will add a note to the JSDoc for this.

app/messengers/route-with-messenger.tsx
```diff
@@ -0,0 +1,84 @@
...
+  children: ReactNode;
+}) => {
+  const uiMessenger = useUIMessenger();
+  const routeMessengerRef = useRef<RouteMessenger | null>(null);
+  const [isDelegated, setDelegated] = useState(false);
+
+  // `useMemo` doesn't work here because `capabilities` is an object, so we use
+  // a ref instead to ensure that we only create the route messenger once.
+  if (!routeMessengerRef.current) {
+    routeMessengerRef.current = createRouteMessenger({ path });
```
Mrtenz, line 47: Can maybe remove this ref now actually, and just create it in the `useEffect` below?

app/messengers/route-with-messenger.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  useEffect(() => {
```
Mrtenz, line 50: This is different from the extension because (thanks to Cursor) I realised we never performed the cleanup/revocation. Rather than handling delegation inside `createRouteMessenger`, it's now handled in the effect here, and we can revoke in the return function of the effect. This shouldn't cause any additional renders as long as `capabilities` is a stable reference (which it should always be if we use it with the `withMessenger` HOC).

We probably want to port this to the extension too.

app/messengers/configs/keyring-controller.ts
mcmire: Is there anything we need to do here? Maybe `defineExcludedCapabilities` doesn't work how it should?

Mrtenz: Seems to be working as intended on mobile, but I need to double check the extension implementation.

app/messengers/route-with-messenger.tsx
```diff
@@ -0,0 +1,84 @@
...
+        .catch(captureException);
+    };
+  }, [uiMessenger, capabilities]);
+
+  const routeMessenger = routeMessengerRef.current;
+
+  // Wait for delegation to complete before rendering children, to ensure that
+  // the route messenger has access to the delegated capabilities when children
+  // are rendered.
+  if (!isDelegated) {
```
mcmire, line 75: From a type perspective do we also need to verify that `routeMessenger` is set here? 

``` suggestion
  if (routeMessenger && !isDelegated) {

app/messengers/route-with-messenger.tsx
mcmire: Seems fine to do it here.

mcmire (dismissed):
LGTM!

app/messengers/route-with-messenger.tsx
Mrtenz: `routeMessenger` should always be defined at this point, no? The type seems fine too.

mcmire: It should be yes. The type on `routeMessengerRef.current` is `RouteMessenger | null` so I didn't know if TypeScript was okay with that. I see now that `RouteMessengerContext` has the same type though, so no problems.

tommasini (dismissed):

FrederikBolding (approved):

cryptodev-2s (approved):

tommasini (approved):

PR #29757, opened 2026-05-05
feat: hoist perps providers for homepage hub tabs UI cp-7.77.0 

vinnyhoward:
## **Description**

Optimizes the hub page discovery tabs for Perps WebSocket efficiency and eliminates a double provider instance introduced by the tabs feature.

Provider hoisting: `PerpsConnectionProvider` and `PerpsStreamProvider` were instantiated twice under the hub tabs flag, once inside the Perps tab and again inside Homepage (for PerpsSection on the Portfolio tab). Both consumed the same singleton stream manager but mounted separate React provider trees. Providers are now hoisted to the `HomepageDiscoveryTabs` root so both tabs share one instance under the Tabs UI feature flag. `Homepage` accepts a `perpsProvidersHoisted` prop to skip its own providers when a parent already provides them.

WS pause/resume: Added `pauseAllChannels()` and `resumeAllChannels()` to `PerpsStreamManager`. When the user navigates to the Predictions tab , the only tab with no Perps consumers. All stream channels are paused, stopping subscriber notifications while keeping the WebSocket subscribed and the cache warm. Channels resume immediately on return to Portfolio or Perpetuals. A cleanup useEffect ensures channels are never left paused if the component unmounts mid-transition.

Extensibility: Tab indices that require live Perps data are declared in a `PERPS_WS_TABS` Set, making future tab additions a one-line change.

## **Changelog**

CHANGELOG entry:null

## **Related issues**

Fixes: https://consensyssoftware.atlassian.net/browse/TMCU-737

## **Manual testing steps**

```gherkin
Feature: Perps WebSocket efficiency on hub page tabs

  Scenario: navigating to Predictions pauses Perps WS
    Given the hub page discovery tabs UI is enabled
    And the user is on the Portfolio or Perpetuals tab
    When user taps the Predictions tab
    Then Perps stream channels are paused (no active polling)
    And the WebSocket connection remains open

  Scenario: returning from Predictions resumes Perps WS
    Given the user is on the Predictions tab
    When user taps Portfolio or Perpetuals
    Then Perps stream channels resume immediately

  Scenario: switching between Portfolio and Perpetuals
    Given the user is on the Portfolio tab
    When user taps the Perpetuals tab
    Then Perps stream channels are not paused or resumed
    And PerpsSection on Portfolio and PerpsHomeView both reflect current data

  Scenario: Perps data visible on Portfolio tab
    Given the hub tabs flag is enabled
    When the user views the Portfolio tab
    Then PerpsSection renders with live data (single provider instance, no double-mount)
```

## **Screenshots/Recordings**

`~`

### **Before**

`~`

### **After**

`~`

- [ ] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes Perps WebSocket stream emission behavior (new pause reference counting + global pause/resume) and restructures provider mounting in `HomepageDiscoveryTabs`, which could affect live Perps data updates and tab navigation edge cases.
> 
> **Overview**
> Hoists `PerpsConnectionProvider` + `PerpsStreamProvider` to `HomepageDiscoveryTabs` so the Portfolio and Perpetuals tabs share a single Perps provider tree, and adds a `perpsProvidersHoisted` prop to `Homepage` to skip creating nested Perps providers.
> 
> Updates Perps streaming to support **reference-counted pausing** (`pauseCount`) and adds `pauseAllChannels()`/`resumeAllChannels()`; `HomepageDiscoveryTabs` now pauses all Perps stream channels when switching to tabs that don’t consume Perps WS data (e.g. Predictions) and resumes on return, with unmount cleanup guarded to avoid unmatched resumes.
> 
> Adds/extends unit tests covering pause reference counting, pause-all/resume-all behavior, and homepage/perps feature-flag rendering when providers are hoisted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 084ec3d313f6d7c4c18358dce294a42cfd7514b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -234,6 +246,23 @@ const HomepageDiscoveryTabs = forwardRef<
...
+          const isPerpsVisible =
+            i === TAB_INDEX.PERPETUALS || i === TAB_INDEX.PORTFOLIO;
+
+          if (wasPerpsVisible && !isPerpsVisible) {
+            getStreamManagerInstance().pauseAllChannels();
+          } else if (!wasPerpsVisible && isPerpsVisible) {
+            getStreamManagerInstance().resumeAllChannels();
+          }
+        }
+
```
gambinish: This seems fine for now. It would be nice if we could come up with a more scalable solution

app/components/Views/Homepage/Homepage.tsx
```diff
@@ -46,69 +46,118 @@ import { HomepageTrendingAbTestContext } from './context/HomepageTrendingAbTestC
...
-  const isDeFiEnabled = useSelector(selectDeFiPositionsSectionEnabled);
-  const isWhatsHappeningEnabled = useSelector(selectWhatsHappeningEnabled);
-  const isTopTradersEnabled = useSelector(selectSocialLeaderboardEnabled);
-  const isMusdConversionEnabled = useSelector(
-    selectIsMusdConversionFlowEnabledFlag,
-  );
-  const { isEligible: isGeoEligible } = useMusdConversionEligibility();
-  const isCashSectionEnabled = isMusdConversionEnabled && isGeoEligible;
+const Homepage = forwardRef<SectionRefreshHandle, HomepageProps>(
+  ({ perpsProvidersHoisted = false }, ref) => {
```
gambinish, line 59: Can you re-clarify to me what this boolean property guards against? Is this to prevent us from mounting the perps provider twice in certain A/B scenarios?

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: I totally agree. I can look into it today

app/components/Views/Homepage/Homepage.tsx
vinnyhoward: Yeah exactly. You can see it [here](https://github.com/MetaMask/metamask-mobile/pull/29757/changes#diff-ceec5da9c62daad0aaab903a6b30a0df64ba2f3847447ac41a3f0b3d5cdb819eR322). When `perpsProvidersHoisted=true` the treatment, the providers are already mounted higher up in the tree so we skip wrapping here to avoid mounting them twice. In the control they aren't hoisted so we still wrap perps content with `PerpsConnectionProvider`/`PerpsStreamProvider` as before.

app/components/UI/Perps/providers/PerpsStreamManager.tsx
```diff
@@ -1734,6 +1734,37 @@ export class PerpsStreamManager {
...
+    this.prices.resume();
+    this.orders.resume();
+    this.positions.resume();
+    this.fills.resume();
+    this.account.resume();
+    this.oiCaps.resume();
+    this.topOfBook.resume();
+    this.candles.resume();
+  }
+
```
vinnyhoward, line 1770: Would we need possibly need to clear the cache on resume [here](https://github.com/MetaMask/metamask-mobile/blob/0b6a6a0fe44c8fc684805ee28475156e1264c425/app/components/UI/Perps/providers/PerpsStreamManager.tsx#L324-L326)? Something like:

```typescript
  public resume(): void {
    this.isPaused = false;
    const cached = this.getCachedData();
    if (cached !== null) this.notifySubscribers(cached);
  }
```

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: Refactored [ce37b0b](https://github.com/MetaMask/metamask-mobile/pull/29757/commits/ce37b0bf8cad99e89770fe85669e64bcd22a5197). Uses `Set`to create a unique hash where we can just see if this tab needs WS connections

app/components/UI/Perps/providers/PerpsStreamManager.tsx
vinnyhoward: When you returning from the Predictions tab, subscribers immediately receive the last known state from the cache rather than showing stale data until the next WebSocket push. Not sure how frequent the polling is but if its fast already then maybe we don't need it

app/components/Views/Homepage/Homepage.tsx
```diff
@@ -296,97 +358,55 @@ const Homepage = forwardRef<SectionRefreshHandle>((_, ref) => {
...
-                  {renderTreatmentNonPerpsSections(trendingPerpsSection)}
-                </Box>
-              </PerpsStreamProvider>
-            </PerpsConnectionProvider>
+          <PerpsSectionWithProvider
+            ref={perpsSectionRef}
+            sectionIndex={getSectionIndex(HomeSectionNames.PERPS)}
+            totalSectionsLoaded={totalSectionsLoaded}
+            mode={sectionMode}
+          />
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [6d10073](https://github.com/MetaMask/metamask-mobile/pull/29757/commits/6d10073e5c73f72f3336114a47d33e8ed39d9c7a)

app/components/Views/Homepage/Homepage.tsx
```diff
@@ -296,97 +358,64 @@ const Homepage = forwardRef<SectionRefreshHandle>((_, ref) => {
...
-                  sectionIndex={getSectionIndex(HomeSectionNames.PERPS)}
-                  totalSectionsLoaded={totalSectionsLoaded}
-                  mode={sectionMode}
-                />
-                <Box gap={8}>
-                  {renderTreatmentNonPerpsSections(trendingPerpsSection)}
-                </Box>
-              </PerpsStreamProvider>
-            </PerpsConnectionProvider>
+          {perpsProvidersHoisted ? (
```
geositta: The provider hoist makes sense here, but this path also bypasses the Perps feature-flag gate that used to live in `PerpsSectionWithProvider`.

Before this change, `PerpsSectionWithProvider` returned `null` when `selectPerpsEnabledFlag` was false. With `perpsProvidersHoisted=true`, we now render `PerpsSectionBase` directly, which correctly avoids mounting duplicate providers but does not check the flag.

Can we keep the hoisted-provider behavior while preserving the old flag behavior, e.g. only render this when `isPerpsEnabled` is true? A small regression test for `perpsProvidersHoisted + Perps disabled` would cover it.

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
geositta: The tab-visibility pause/resume behavior makes sense for the Predictions tab. My concern is that it uses the same unscoped channel pause state as `PerpsController.#withStreamPause()`.

Each channel has one boolean pause state. If a Perps action such as canceling orders or closing a position has paused `orders` / `positions` / `account`, and the user switches to Predictions and back before that action finishes, the `resumeAllChannels()` call resumes those channels before the controller's `finally` runs.

That allows WebSocket updates to reach subscribers during the protected operation that `#withStreamPause()` is meant to isolate.

Could we make this source aware or reference counted so tab visibility only releases pauses it created?

geositta (changes requested):
Requesting two changes, both with inline comments:

1. The `perpsProvidersHoisted` path on the Portfolio homepage bypasses the Perps feature flag.

2. The tab-visibility WebSocket optimization is a good direction, but the current `pauseAllChannels()` / `resumeAllChannels()` implementation shares unscoped pause state with `PerpsController.#withStreamPause()` and can release a controller owned pause while a Perps action is still running.

The provider hoist is a reasonable direction. I would be ready to approve once these two issues are addressed.

app/components/Views/Homepage/Homepage.tsx
vinnyhoward: Fixed [d1f8f4b](https://github.com/MetaMask/metamask-mobile/pull/29757/commits/d1f8f4b5c2e2b46797e99e5c942429c237c18628). The `isPerpsEnabled` guard is now hoisted above the `perpsProvidersHoisted` ternary so it gates both `PerpsSectionBase` and `PerpsSectionWithProvider` from a single point

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: Let me know what you think of this fix [d1f8f4b](https://github.com/MetaMask/metamask-mobile/pull/29757/commits/d1f8f4b5c2e2b46797e99e5c942429c237c18628). I'm open to other solutions @geositta 

Swapped the boolean for a reference counter. Now `pause()` increments and `resume()` decrements, so the channel only unblocks when every holder has released. The tab switching back no longer releases an in-flight controller pause. Added tests for the race condition and the underflow guard.

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -202,6 +212,13 @@ const HomepageDiscoveryTabs = forwardRef<
       },
     }));
 
+    // Ensure streams are never left paused if the component unmounts mid-transition.
+    useEffect(
+      () => () => {
+        getStreamManagerInstance().resumeAllChannels();
```
geositta: @vinnyhoward Thank you for adding the ref counted pause to fix the main tab-switch/controller overlap. One small cleanup request: this unmount cleanup calls `resumeAllChannels()` even when the tab layer has not acquired a pause.

The initial tab is Portfolio, so the tab layer starts with no pause held. If this component unmounts while still on Portfolio or Perpetuals, this cleanup issues an unmatched resume and can decrement a controller owned pause count.

Could we track whether the tab layer currently owns a pause, then only call `resumeAllChannels()` from cleanup when that ref is true? 

That keeps the cleanup behavior without releasing a pause created by another caller.

vinnyhoward: Thats a good callout. Let me look into this

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -213,6 +223,13 @@ const HomepageDiscoveryTabs = forwardRef<
...
+      },
+      [],
+    );
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [c77dfd8](https://github.com/MetaMask/metamask-mobile/pull/29757/commits/c77dfd86c93b8b5dfcec7aea6c326f996ac75c35). Bug introduced by merge conflict

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
```diff
@@ -26,7 +26,10 @@ import Homepage from '../../Homepage';
 import PerpsHomeView from '../../../../UI/Perps/Views/PerpsHomeView/PerpsHomeView';
 import PredictFeed from '../../../../UI/Predict/views/PredictFeed';
 import { PerpsConnectionProvider } from '../../../../UI/Perps/providers/PerpsConnectionProvider';
-import { PerpsStreamProvider } from '../../../../UI/Perps/providers/PerpsStreamManager';
+import {
+  PerpsStreamProvider,
+  getStreamManagerInstance,
+} from '../../../../UI/Perps/providers/PerpsStreamManager';
```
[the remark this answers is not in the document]
vinnyhoward: Fixed [c77dfd8](https://github.com/MetaMask/metamask-mobile/pull/29757/commits/c77dfd86c93b8b5dfcec7aea6c326f996ac75c35). Bug introduced by merge conflict

app/components/Views/Homepage/components/HomepageDiscoveryTabs/HomepageDiscoveryTabs.tsx
vinnyhoward: I've added `tabLayerHoldsPauseRef` to track whether the tab layer currently owns a pause, so the unmount cleanup only calls `resumeAllChannels()` when the ref is `true`. This prevents any unmatched resume against a controller-owned pause count.

That said, I know this isn't the most scalable solution, since the hub tabs are behind an A/B test flag I didn't want to commit to a larger refactor here. I've put together a separate PR with the named key ownership approach that addresses this more robustly, would love your review on it since you know this layer best @geositta 

PR: https://github.com/MetaMask/metamask-mobile/pull/29935

geositta (dismissed):
This looks good to me.

The counter fixes the main issue; when the tab pause is released, it no longer turns streams back on if the controller still has its own pause active. The new test covers that case.

The `tabLayerHoldsPauseRef` guard also fixes the cleanup, because unmount only calls `resumeAllChannels()` if the tabs component actually paused the streams first.

The source-aware approach in the other PR is a stronger long-term API, but I don't think this PR needs that bigger refactor? This fix covers the regressions we found here.

vinnyhoward:
> This looks good to me.
> 
> The counter fixes the main issue; when the tab pause is released, it no longer turns streams back on if the controller still has its own pause active. The new test covers that case.
> 
> The `tabLayerHoldsPauseRef` guard also fixes the cleanup, because unmount only calls `resumeAllChannels()` if the tabs component actually paused the streams first.
> 
> The source-aware approach in the other PR is a stronger long-term API, but I don't think this PR needs that bigger refactor? This fix covers the regressions we found here.

@geositta Thank you for reviewing and approving. I agree, the work here does not require the bigger refactor. Maybe we can revisit if the tabs UI does become the default UI on the hompage? I'll close the other PR for now. Happy Friday!

wachunei (dismissed):

geositta (approved):

wachunei (approved):

gambinish (approved):

PR #29943, opened 2026-05-10
chore: align custom network logos to square format

jeremytsng:
## Summary

Polish pass on previously base-enabled chain PNG logos so their visual treatment in the Custom networks list matches the square format used by Popular networks (Mainnet, Linea, Base, etc.) and recent enablements (Mantle, Tempo).

CHANGELOG entry: Aligned previously base-enabled custom network logos (Stable, Flow, XDC, Fraxtal, Hemi, Plasma, Lukso, Rootstock, MSU, Lens, Plume) to a square format consistent with Popular networks

## Test plan

- [ ] iOS: build and open the network selector → Custom networks tab → confirm each updated chain renders as a square icon at the correct size
- [ ] Android: same as above
- [ ] Confirm chain glyphs are unchanged (only background is different)
- [ ] Sanity-check token list, send flow, and activity views that surface these network icons


maxime-oe (approved):
lgtm :)

Battambang (approved):

PR #30005, opened 2026-05-11
feat(predict): add crypto up/down details UI

ghgoodreau:
## **Description**
This PR adds the crypto up/down market details UI for Predict. It introduces the crypto up/down chart and details screen, wires crypto markets into the existing market details route, adds time-slot selection, price summary, target/current price display, and buy/claim actions.

It also includes focused UI test coverage and a small Bugbot cleanup so outcome status checks use the outcome status type rather than the market status enum.

# TODO
1. This PR is stack item 3/3 and builds on #30004, which builds on #30003.
2. Keep this PR as draft until the stacked PRs are validated and all required checks are passing.
3. Add screenshots or recordings before marking ready for review if reviewers want visual evidence for the new details UI.

## **Changelog**

CHANGELOG entry: Added crypto up/down market details with live charting in Predict

## **Related issues**

Fixes: N/A - stacked Predict crypto up/down implementation split from #29436.

## **Manual testing steps**

```gherkin
Feature: Crypto up/down market details

  Scenario: user views and interacts with a crypto up/down market
    Given the Predict feature is enabled and a crypto up/down market is available

    When user opens the crypto up/down market details screen
    Then user sees the live chart, current price, target price, and available time slots

    When user selects a different time slot
    Then the details screen updates to the selected market window

    When user taps an available Yes or No action
    Then the Predict buy flow opens for the selected outcome
```

### **Before**

N/A

### **After**

Screenshot 2026-05-15 at 4 34 36 PM

## **Pre-merge author checklist**

- [x] I've followed [MetaMask Contributor Docs](https://github.com/MetaMask/contributor-docs) and [MetaMask Mobile Coding Standards](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/CODING_GUIDELINES.md).
- [x] I've completed the PR template to the best of my ability
- [x] I've included tests if applicable
- [x] I've documented my code using [JSDoc](https://jsdoc.app/) format if applicable
- [x] I've applied the right labels on the PR (see [labeling guidelines](https://github.com/MetaMask/metamask-mobile/blob/main/.github/guidelines/LABELING_GUIDELINES.md)). Not required for external contributors.

#### Performance checks (if applicable)

- [ ] I've tested on Android
  - Ideally on a mid-range device; emulator is acceptable
- [ ] I've tested with a power user scenario
[...]

PR #30006, opened 2026-05-11
fix(predict): stabilize crypto up/down e2e tap

ghgoodreau:
## Summary
- Makes the wallet page-object prediction section tap more resilient by retrying in the opposite scroll direction.

## Stack
4/4. Stacked on `predict/crypto-updown-details-ui`.

## Test plan
- yarn eslint tests/page-objects/wallet/WalletView.ts (passed with existing deprecation warning only)
- yarn prettier --check tests/page-objects/wallet/WalletView.ts


ghgoodreau:
Folded this one-file stabilization slice into #30005 so the stack stays lean.

PR #29888, opened 2026-05-07
feat(rewards): show vip menu

sophieqgu:
## **Description**
https://consensyssoftware.atlassian.net/browse/RWDS-1241
Show VIP icon menu to users with VIP enabled
Depend on backend feature branch https://github.com/consensys-vertical-apps/va-mmcx-rewards/pull/546

## **Changelog**

CHANGELOG entry: Show menu for users with VIP feature enabled

## **Screenshots/Recordings**

---

> [!NOTE]
> **Medium Risk**
> Adds new VIP-gated navigation routes and UI plus a new API fetch/caching path in `RewardsController` that mutates persisted state based on backend responses. Moderate risk due to new controller action, persisted state shape changes, and navigation gating/redirect behavior.
> 
> **Overview**
> Adds a VIP experience to Rewards: new `REWARDS_VIP_VIEW` and `REWARDS_VIP_TIERS_VIEW` routes/screens backed by a new `useVipDashboard` hook, plus VIP UI components (tier progress card, fee tiles, volume/points sections, tier rows) with loading/error states.
> 
> Updates the Rewards dashboard header to conditionally show a crown icon that navigates to VIP when `subscription.features.vip.enabled` is true, and adds a hidden 5-tap title “unlock” that calls `RewardsController:getVIPDashboard` to flip the VIP flag when the backend returns a non-null payload.
> 
> Extends the rewards engine/data layer with a new `RewardsController:getVIPDashboard` action and `RewardsDataService:getVIPDashboard` endpoint (`/vip/me`), adds persisted `vipDashboard` state and VIP feature fields on `SubscriptionDto`, and ensures VIP cache is cleared on subscription invalidation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db490291becab1e54eae172102e738affecc87b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #30024, opened 2026-05-12
feat(rewards): expose VIP perps fee discount via getPerpsDiscountForAccount

VGR-GIT:
## Summary

Keep the existing `RewardsController.getPerpsDiscountForAccount` contract (still returns a discount in basis points; perps still applies it via `metamaskFeeRate = base * (1 - discount/10000)`), and only swap *how* the discount is derived. When the account's active subscription has `features.vip.enabled === true`, the controller calls the new authenticated `GET /vip/fees`, reads `hyperliquid.builderFeeBips`, and converts it to a discount fraction using the base fee the caller passes in. Non-VIP accounts now receive a 0 discount; the legacy `/public/rewards/perps-fee-discount` endpoint and its data-service method are removed.

Targets [#29888 (`vip-feature-icon-menu`)](https://github.com/MetaMask/metamask-mobile/pull/29888) because it relies on the new `SubscriptionDto.features.vip.enabled` flag and the per-subscription VIP state introduced there.

- Adds `RewardsDataService.getVipFees` (auth'd, mirrors [va-mmcx-rewards#546](https://github.com/consensys-vertical-apps/va-mmcx-rewards/pull/546) DTOs)
- Adds per-subscription cache slot `RewardsControllerState.vipPerpsFees` (5-min TTL; stores raw bips so the cache is independent of `baseFeeBips`)
- Extends `getPerpsDiscountForAccount(account, baseFeeBips)` — perps callers pass `BUILDER_FEE_CONFIG.MaxFeeDecimal * BASIS_POINTS_DIVISOR` so the controller stays a pure transformer (no perps constant import inside rewards)
- Removes `getPerpsDiscount` data service method, `RewardsDataServiceGetPerpsDiscountAction`, `#getPerpsFeeDiscountData`, `GetPerpsDiscountDto`, `PerpsDiscountData`
- Simplifies `getHasAccountOptedIn` to rely on cached state only
- Updates the three perps call sites (`RewardsIntegrationService`, `usePerpsOrderFees`, `usePerpsCloseAllCalculations`) and the mobile infrastructure adapter

## **Changelog**

CHANGELOG entry: null

## Screenshots

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes the perps fee-discount source and contract (adds `baseFeeBips`, introduces `null` for unhydrated/failed VIP lookups), which can affect fee calculations and caching behavior across perps UI flows.
> 
> **Overview**
> Perps fee discounts are now derived from VIP subscription data: `RewardsController.getPerpsDiscountForAccount(account, baseFeeBips)` calls a new authenticated `GET /vip/fees`, converts the returned absolute VIP builder fee into a discount fraction vs the caller-provided base fee, and caches the raw VIP builder fee per subscription (`vipPerpsFees`, 5‑min TTL).
> 
> The legacy public perps-discount endpoint and related data-service/action/types/caching logic are removed, and `getHasAccountOptedIn` no longer falls back to perps-discount API calls.
> 
> All perps call sites (`RewardsIntegrationService`, `usePerpsOrderFees`, `usePerpsCloseAllCalculations`, `mobileInfrastructure`) are updated to pass `baseFeeBips` and to treat `null` as *unhydrated/unavailable* (don’t cache; retry on subsequent calculations), with new tests covering these cases and VIP fee edge-condition handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb9ccbcfc8afabfa1094599262eb70a48749fa5f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/core/Engine/controllers/rewards-controller/RewardsController.ts
+++ b/app/core/Engine/controllers/rewards-controller/RewardsController.ts
@@ -4,7 +4,6 @@ import {
   type RewardsControllerState,
   type RewardsAccountState,
   type LoginResponseDto,
-  type PerpsDiscountData,
   type EstimatePointsDto,
   type EstimatedPointsDto,
   type SeasonDtoState,
@@ -51,6 +50,8 @@ import {
   type SubscriptionBenefitsState,
   type VipDashboardDto,
   type VipDashboardState,
+  type VipFeesResponseDto,
+  type VipPerpsFeesState,
   BASE32_REGEX,
   CampaignType,
 } from './types';
@@ -99,9 +100,6 @@ export const DEFAULT_BLOCKED_REGIONS = ['UK', 'GB', 'GI'];
 
 const controllerName = 'RewardsController';
 
-// Perps discount refresh threshold
-const PERPS_DISCOUNT_CACHE_THRESHOLD_MS = 1000 * 60 * 5; // 5 minutes
-
 // Season status cache threshold
 const SEASON_STATUS_CACHE_THRESHOLD_MS = 1000 * 60 * 1; // 1 minute
 
@@ -116,6 +114,10 @@ const BENEFITS_DETAILS_CACHE_THRESHOLD_MS = 1000 * 60 * 1; // 1 minutes
 
 // VIP dashboard cache threshold — disabled while backend still serves hardcoded data
 const VIP_DASHBOARD_CACHE_THRESHOLD_MS = 0;
+
+// VIP perps fees cache threshold — read on every perps trade UI render, so
+// cache for the same 5-minute window as the legacy public-discount path.
+const VIP_PERPS_FEES_CACHE_THRESHOLD_MS = 1000 * 60 * 5;
 
 // Active boosts cache threshold
 const ACTIVE_BOOSTS_CACHE_THRESHOLD_MS = 1000 * 60 * 1; // 1 minute
@@ -323,6 +325,12 @@ const metadata: StateMetadata<RewardsControllerState> = {
     persist: true,
     includeInDebugSnapshot: false,
     usedInUi: true,
+  },
+  vipPerpsFees: {
+    includeInStateLogs: true,
+    persist: true,
+    includeInDebugSnapshot: false,
+    usedInUi: false,
   },
 };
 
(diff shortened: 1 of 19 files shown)


sophieqgu (dismissed):

sophieqgu (dismissed):

app/core/Engine/controllers/rewards-controller/RewardsController.ts
```diff
@@ -1684,15 +1611,117 @@ export class RewardsController extends BaseController<
...
+   */
+  async #getVipPerpsDiscountBips(
+    account: CaipAccountId,
+    baseFeeBips: number,
+  ): Promise<number | null> {
+    if (!Number.isFinite(baseFeeBips) || baseFeeBips <= 0) return null;
+
+    const accountState = this.#getAccountState(account);
+    const subscriptionId = accountState?.subscriptionId;
+    if (!subscriptionId) return null;
```
geositta, line 1656: This path returns `0` for both "account has no subscription" and "subscription state has not hydrated yet."

That distinction matters now because perps providers mount earlier on the homepage hub tabs: `usePerpsOrderFees` can call this before silent auth has populated `accountState.subscriptionId`, then cache the returned `0` for 5 minutes in `feeDiscountCache`.

After hydration completes, the VIP fee is available, but the hook keeps using the cached no discount value, so VIP users can place trades with the full MetaMask builder fee during that window.

Please make the unhydrated state distinct from an explicit non VIP result.

For example, return `null`/`undefined` when `accountState` or `subscriptionId` is absent and only return `0` after the controller knows the subscription is non VIP.

Then callers can skip caching the unhydrated result and retry on the next fee calculation.

geositta (changes requested):
The new VIP fee path is close but from what I can tell there is a startup timing bug that can make VIP users lose their discount for a short time after launch.

Once that state readiness case is handled, the rest of the fee conversion approach looks good.

app/core/Engine/controllers/rewards-controller/RewardsController.ts
VGR-GIT: @geositta valid, see https://github.com/MetaMask/metamask-mobile/pull/30024/changes/cb9ccbcfc8afabfa1094599262eb70a48749fa5f

app/core/Engine/controllers/rewards-controller/RewardsController.ts
```diff
@@ -1684,15 +1611,118 @@ export class RewardsController extends BaseController<
...
+        Logger.log(
+          'RewardsController: VIP fees fetch failed; returning no discount:',
+          error instanceof Error ? error.message : String(error),
+        );
+        return null;
+      }
+    }
+
+    const builderFeeBips = parseFloat(builderFeeBipsRaw);
+    if (!Number.isFinite(builderFeeBips) || builderFeeBips <= 0) {
```
geositta, line 1699: Non-blocking: this rejects `builderFeeBips === 0`, but `0` is a valid value in the conversion formula and should produce a 100% discount.

Tier 0 is already represented by `fees=null`, so the numeric guard only needs to reject non-finite or negative values. 

The nearby comment also has the meanings reversed: `builderFeeBips = 0` means zero VIP builder fee, while
`builderFeeBips = baseFeeBips` means no discount.

This likely is not user facing today if all current VIP tiers have a non-zero builder fee, but changing `<= 0` to `< 0` keeps the transformer correct for a future zero-fee tier.

geositta (approved):
@VGR-GIT Thanks for fixing the startup cache race.

I left one small followup on the builder fee guard. It likely does not affect current users, but allowing `builderFeeBips = 0` keeps the conversion correct if a future VIP tier has a zero builder fee.

jbblanc (approved):

PR #30055, opened 2026-05-12
feat: add Universal dev mode feature flag

Jwhiles:
### The problem
It's currently very awkward to point the mobile app at non-prod backends. `AuthenticationController` is pinned to `Env.PRD`, so to test against dev you have to hand-edit its init, plus every downstream service's URL. Any that you don't update will throw errors for all upstream requests.

To get around this, a number of teams have configured their Dev APIs to work with Prod JWTs. This makes testing against Dev versions of these APIs easy. One can simply update the base URL in the API's configuration and everything will “Just work”.

We considered doing the same with Chomp, but we rely on the new newly introduced Authenticated User Storage service - which makes this approach impossible.

The reason that this becomes difficult is that Authenticated User Storage uses the same JWT for authentication, but *is* environment scoped. So if we attempt to use a Prod JWT in the Dev deployment of the Chomp API, when we attempt to interact with authenticated User Storage, we will hit authentication errors.

image

Thus we've decided that it's time to start properly treating dev as a separate environment, and have all our APIs respect dev JWTs. This PR introduces the basic pattern that we will use for pointing this client at those APIs. Initially it only wires up the authentication-controller, and the services that are required for testing chomp. We hope that other teams will update their own services as they update their APIs to support dev JWTs

### Implementation
This PR adds a build time switch — `MM_DEV_API_ENV=dev | prod` in `.js.env` that the auth controller reads at init time. This switch is also use to control the config passed into the chomp API service and authenticated user storage.

Additionally this PR adds a button to the dev settings menu which allows clearing the client's stored JWT - which may be needed when switching environments.

## **Changelog**

CHANGELOG entry: Add dev API testing switch

## **Related issues**

Fixes:

## **Manual testing steps**

echo 'export MM_DEV_API_ENV="dev"' >> .js.env
yarn watch:clean && yarn start:ios

Sign in, then confirm that outbund requests target
`*.dev-api.cx.metamask.io` - and that the JWT issuer claim (`iss`) should be
`https://oidc.dev-api.cx.metamask.io`.

```gherkin
Feature: my feature name

```

- [ ] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes how auth/user-storage/profile-metrics and Chomp select environments based on a build-time env var; misconfiguration could point builds at the wrong backend or cause auth failures.
> 
> **Overview**
> Introduces a build-time env switch `MM_DEV_API_ENV` (defaulting to `prod`) to consistently route **JWT minting and JWT-consuming services** to the same backend environment.
> 
> Adds `app/core/devApiEnv.ts` and wires it into `AuthenticationController` (OIDC env), `UserStorageController`, `AuthenticatedUserStorageService`, and `ProfileMetricsService`; `ChompApiService` now prefers an env-specific base URL override (with logging) over the remote feature flag to avoid token/env mismatches.
> 
> Adds an **Identity** section to Developer Options with a button to clear the persisted auth session after switching envs, plus new i18n strings and tests; updates test/Babel config to avoid inlining env vars for the new helper.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 56c8c14ca4e96d13a9368943904c537da48d4ab3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MoMannn (commented):
One problem is that when you switch this "ON" the JWT being returned is still PROD since its in a persistant storage until it get invalidated. I think we will also need to update `AuthenticationController` so that it clears storage if `MM_DEV_API_ENV` is on or something like this. Or like a button to clear persistant storage.

PatrykLucka (dismissed):

shane-t (dismissed):

mathieuartu (changes requested):
Looks good overall but this PR leaves `UserStorageController` untouched. This would render all things profile-sync related broken when `MM_DEV_API_ENV=dev`.
Fix would be to also pass `authEnv()` to `UserStorageController` constructor 🙂 

```ts
const controller = new UserStorageController({
  // ...
  config: { env: authEnv() },
});
```

Jwhiles:
> Looks good overall but this PR leaves `UserStorageController` untouched. This would render all things profile-sync related broken when `MM_DEV_API_ENV=dev`. Fix would be to also pass `authEnv()` to `UserStorageController` constructor 🙂
> 
> ```ts
> const controller = new UserStorageController({
>   // ...
>   config: { env: authEnv() },
> });
> ```

Hey @mathieuartu - we're trying to just touch the services that we directly need for chomp. The expectation is that this becomes the new standard and that other teams will update their services, and that in the meantime we'll see other API requests fail when this is enabled

app/core/devApiEnv.ts
```diff
@@ -0,0 +1,32 @@
...
+ * no ordering between controllers required. Each consumer reads the same
+ * source independently.
+ */
+
+export type DevApiEnv = 'dev' | 'prod';
+
+/**
+ * Read at call time (not module load) so tests can set/unset
+ * `process.env.MM_DEV_API_ENV` without juggling the module cache.
+ */
```
Prithpal-Sooriya, line 23: Are we able to add a disclaimer here or elsewhere?
> Disclaimer: Enabling dev will break all other authenticated prod endpoints

IIRC there are a few APIs from top of mind that are authenticated, but are still prod in this PR:
- Accounts API
- E2EE Encrypted Storage API
- Notification Services (trigger API, notification API, push-notification API)
- Perp Alerts API

mathieuartu:
> > Looks good overall but this PR leaves `UserStorageController` untouched. This would render all things profile-sync related broken when `MM_DEV_API_ENV=dev`. Fix would be to also pass `authEnv()` to `UserStorageController` constructor 🙂
> > ```ts
> > const controller = new UserStorageController({
> >   // ...
> >   config: { env: authEnv() },
> > });
> > ```
> 
> Hey @mathieuartu - we're trying to just touch the services that we directly need for chomp. The expectation is that this becomes the new standard and that other teams will update their services, and that in the meantime we'll see other API requests fail when this is enabled

After discussing this on slack, I'm happy to approve with or without my requested changes 🙂. If it doesn't land in this PR, I'll do a tiny follow up myself! Thanks for doing this

mathieuartu (dismissed):
Awesome, thanks for addressing my comment!

PatrykLucka (dismissed):

mathieuartu (approved):

PatrykLucka (approved):

shane-t (approved):

PR #30154, opened 2026-05-13
fix(predict): hide child more-market cards from feed

caieu:
# **Description**

Filters Polymarket "more market" child events out of Predict feed surfaces without changing raw pagination behavior.

Polymarket marks child/more-market events with `parentEventId`. This PR maps that API field to `PredictMarket.parentMarketId`, then filters parented markets from feed and featured carousel render data. Pagination still uses the raw page size before filtering, so infinite scroll can continue even when a page contains hidden child markets.

## **Changelog**

CHANGELOG entry: Fixed child prediction markets appearing as standalone cards in the Predict feed.

## **Related issues**

Fixes: PRED-865

## **Manual testing steps**

```gherkin
Feature: Predict feed more-market filtering

  Scenario: user scrolls the Predict feed
    Given Polymarket returns feed events that include child events with parentEventId

    When user opens the Predict feed
    Then child more-market cards are not rendered as standalone cards

    When user scrolls to the end of the visible feed page
    Then the feed continues loading the next raw page when available
```

## **Screenshots/Recordings**

N/A. Logic-only feed filtering change covered by unit tests.

### **Before**

N/A

### **After**

N/A

- [ ] I've tested on Android

---

> [!NOTE]
> **Medium Risk**
> Changes which markets are surfaced in the Predict feed/carousel by filtering out events with `parentMarketId`, which could inadvertently hide valid markets or impact pagination edge-cases despite added tests.
> 
> **Overview**
> Filters Polymarket “more-market” child events out of Predict feed surfaces.
> 
> This maps Polymarket `parentEventId` onto `PredictMarket.parentMarketId` (now nullable) during event parsing, then adds `filterStandaloneMarkets` and applies it in `usePredictMarketData` and `useFeaturedCarouselData` so only standalone markets render. Pagination/`hasMore` behavior continues to use the *raw* page size before filtering, with new unit tests covering filtering and offset handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c7379a0577da43ff628640661009f56a5bad438. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


MarioAslau (commented):
## Medium severity

### 1. `isStandaloneMarket` treats numeric `0` as a child market

```ts
// metamask-mobile/app/components/UI/Predict/utils/feed.ts
export function isStandaloneMarket(market: PredictMarket): boolean {
  const { parentMarketId } = market;
  return (
    parentMarketId === undefined ||
    parentMarketId === null ||
    String(parentMarketId).trim() === ''
  );
}
```

The `PredictMarket.parentMarketId` type is `string | number | null`. If a future Polymarket payload ever uses `0` as a sentinel for "no parent" (or if the field defaults to `0` for new rows), `String(0).trim() === '0'`, so the market is treated as a child and silently filtered out. Same applies to `NaN` (`String(NaN) === 'NaN'`).

The current tests cover string `''`, `null`, string parent, and numeric `123`, but do not cover the `0` case. Two cheap options:

```ts
// Option A: treat any number as a real id (matches the Polymarket convention of positive ids).
if (typeof parentMarketId === 'number') return false;
if (parentMarketId == null) return true;
return parentMarketId.trim() === '';

// Option B: be explicit about which numeric values are "no parent".
if (parentMarketId === undefined || parentMarketId === null) return true;
if (typeof parentMarketId === 'number') return parentMarketId === 0;
return parentMarketId.trim() === '';
```

Either is unambiguous and dodges the `String(0) === '0'` quirk. Worth one sentence in the PR confirming which numeric values Polymarket actually emits.

### 2. E2E fixture schema is out of date

From the bot comment on the PR:

> E2E Fixture Validation, Structural changes detected. Missing keys 15, Value mismatches 12 (informational).

This is the `parentMarketId` field being added to `PredictMarket`. Run `@metamaskbot update-mobile-fixture` in a PR comment before merge so the fixture stays in sync with the type.

---

## Low severity

### 3. Architectural inconsistency: carousel filters in `useMemo`, feed filters inside the fetch callback

`useFeaturedCarouselData` applies `filterStandaloneMarkets` reactively in `useMemo`. `usePredictMarketData` applies it once at fetch time and stores filtered data in state. Both work, but if a future change adds another feed filter, the two will need to be touched in different places. Not blocking; just worth noting that a single shared "transform pipeline" helper would be cleaner if more filters land.

### 4. Polymarket API type widened to allow `null` without auditing other consumers

```ts
// metamask-mobile/app/components/UI/Predict/providers/polymarket/types.ts
-  parentEventId?: string | number;
+  parentEventId?: string | number | null;
```

`PolymarketProvider.ts` line 303 reads `event.parentEventId ?? event.id` (nullish coalescing handles both `null` and `undefined`, safe). `utils.ts` lines 1383 to 1407 use `parentEventId` for child-event aggregation (`!e.parentEventId` falsy check, also safe). So the widening looks benign with current consumers, but worth a one-line note in the PR that this was checked, otherwise it reads as a quiet API contract change.

### 5. `refine` is now called on filtered data, not raw

```ts
// metamask-mobile/app/components/UI/Predict/hooks/usePredictMarketData.tsx
setMarketData(refine ? refine(visibleMarkets) : visibleMarkets);
```

Current callers (`PredictFeed.tsx` lines 271 and 527) only ever pass `deduplicateSeriesMarkets` via `refine`, and that helper does not depend on seeing children, so this change is safe today. If a future caller passes a refine function that does cross-market analysis (e.g., aggregating parent + children stats), the behavior would silently change. Low because the surface is small and well-known, but a JSDoc note on the `refine` option in `UsePredictMarketDataOptions` saying "called on the already-filtered, user-visible markets" would future-proof it.

matallui (approved):

